Issue #1034, opened 2018-09-14
read_only and write_only decorators in GPU kernels cause tests to fail on windows


SteveBronder:
## Description

The Windows OpenCL driver reserves read_only, write_only, and read_write argument decorators for pipes and image type arguments so our tests are failing on Rok's Windows test machine. The decorators are mostly a 'nice-ity' so we can remove them without much impact. The problem occured because the NVIDIAs OpenCL JIT compilers and older AMD compiler are not strict with this and ignore the use of access qualifiers for types that are not image or pipes.

## Example

On recent Windows AMD implementations the AMDs JIT OpenCL compiler outputs the following error: error: access qualifier can only be used for pipe and image type.

## Additional Information

Just going to remove the decorators. This should have no material effect on the kernels.

## Current Math Version
v2.18.0

rok-cesnovar:
It seems that we both posted the same issue :) You pick one and close the other :)

SteveBronder:
lol, I merged your comments into mine above so will close the other.

PR #1035, opened 2018-09-14
Remove argument decorators from GPU kernel signatures

SteveBronder:
## Summary

Removes read_only, write_only, and read_write from the arguments in the GPU kernel signature. The Windows OpenCL driver reserves read_only, write_only, and read_write argument decorators for pipes and image type arguments so our tests are failing on Rok's Windows test machine. The decorators are mostly a 'nice-ity' so we can remove them without much impact.

## Tests

Same tests as before

```cpp
./runTests.py -j8 test/unit/math/gpu/
```
## Checklist

- [x] Math issue #1034

Steve Bronder, Rok Češnovar, and Erik Štrumbelj

### DIFF
--- a/stan/math/gpu/kernels/add.hpp
+++ b/stan/math/gpu/kernels/add.hpp
@@ -22,10 +22,9 @@ const char *add_kernel_code = STRINGIFY(
      * <code>add_kernel_code.</code>
      * This kernel uses the helper macros available in helpers.cl.
      */
-    __kernel void add(__global write_only double *C,
-                      __global read_only double *A,
-                      __global read_only double *B, read_only unsigned int rows,
-                      read_only unsigned int cols) {
+    __kernel void add(__global double *C, __global double *A,
+                      __global double *B, unsigned int rows,
+                      unsigned int cols) {
       int i = get_global_id(0);
       int j = get_global_id(1);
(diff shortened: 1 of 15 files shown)


rok-cesnovar:
We can maybe just use `const __global` instead of `__global read_only`. The OpenCL standard states: 
> The const qualifier can also be used with the __global qualifier to specify a read-only buffer memory object.

We can also use const with all non-global parameters, since we never change them.

rok-cesnovar:
I think using the `const` qualifier will suffice for the kernels that we currently have. I dont see a case where `__constant` would be useful. 

It makes no sense to use caching for the simple parameters like `rows` and such, they all get loaded in registers anyways.

Using `__constant` for the simple kernels has no real benefit. In the only non-trivial kernels we currently have (matrix multiply), we use local memory instead of caching, which is a lot faster, if you can use it.
For future kernels we can just evaluate each case to see if it could be useful.

SteveBronder:
Word that makes sense. Let's leave const out for now and revisit it later

rok-cesnovar:
Agreed. I will open up an issue for a argument refactor (use size_t where applicable etc) that I will take care of this week. I have re-checked this PR and it looks fine.  Thanks Steve!

PR #1044, opened 2018-09-28
Revert "makefile fixes: setting DEPFLAGS_OS only if it hasn't been set by the user; adds O option to building Sundials"

syclik:
Reverts stan-dev/math#1042


syclik:
This reverts the PR that just went in. This should make Math build again.

seantalts:
If develop passes (I think it will now that I added more /tmp to the EC2 node) I think we should just close this.

syclik:
Thanks. I agree.

(If it doesn’t and this does pass, we’ll revert and I’ll figure it out)

syclik:
Looks like `develop` passed. Closing this PR.

Issue #951, opened 2018-07-24
Math repo runs no tests on Windows


seantalts:
Ideally we'd run the unit tests there. I think they might be broken, though.

#### Current Version:
v2.18.0

syclik:
Thanks. I've created a branch: feature/issue-951-windows-ci-tests

If anyone wants to help, much appreciated.

@seantalts, I don't have direct access to a windows box anymore. Can I log
into gelman-group-win periodically to fix this?

syclik:
Thanks! I think I still have credentials. I'll let you know if I don't.

rok-cesnovar:
I am not sure if this helps in any way, my knowledge of CI and all that is negligible :) I hope it does.

I did a clean clone of the math repo on my windows machine. The tests run when CC=g++ but fail with clang++

```
clang++ -Wall -I . -isystem lib/eigen_3.3.3 -isystem lib/boost_1.66.0 -isystem lib/sundials_3.1.0/include -std=c++1y -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION -Wno-unused-function -Wno-uninitialized -DNO_FPRINTF_OUTPUT -pipe   -shared -fPIC -o test/unit/libmultiple.so test/unit/multiple_translation_units1.o test/unit/multiple_translation_units2.o
test/unit/multiple_translation_units2.o: file not recognized: File truncated
clang++.exe: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [test/unit/libmultiple.so] Error 1
```

bob-carpenter:
> On Jul 24, 2018, at 11:45 PM, Rok Češnovar <notifications@github.com> wrote:
> 
> I am not sure if this helps in any way, my knowledge of CI and all that is negligible :) I hope it does.
> 
> I did a clean clone of the math repo on my windows machine. The tests run when CC=g++

Great news.

> but fail with clang++
> 
[...]

PR #1046, opened 2018-09-29
[WIP] Jenkins tests on Windows.

syclik:
## Summary
This PR enables Windows unit tests. (This is being submitted now to kick off the build on Jenkins.)

The goal is to fix all issues with Windows before this is merged.

## Checklist

- [x] Math issue #951

### DIFF
--- a/stan/math/prim/scal/meta/ad_promotable.hpp
+++ b/stan/math/prim/scal/meta/ad_promotable.hpp
@@ -34,51 +34,12 @@ struct ad_promotable<
     typename boost::enable_if<boost::is_arithmetic<T>, T>::type, T> {
   enum { value = true };
 };
-/**
- * A long double may be promoted to a double.
- */
-template <>
-struct ad_promotable<long double, double> {
-  enum { value = true };
-};
 
 /**
- * A double may be promoted to a double.
+ * A bool may be promoted to a double.
  */
 template <>
-struct ad_promotable<double, double> {
-  enum { value = true };
-};
-
-/**
- * A float may be promoted to a double.
- */
-template <>
-struct ad_promotable<float, double> {
-  enum { value = true };
-};
-
-/**
- * A long may be promoted to a double.
- */
-template <>
-struct ad_promotable<long, double> {  // NOLINT(runtime/int)
-  enum { value = true };
-};
-
-/**
- * An int may be promoted to a double.
- */
-template <>
-struct ad_promotable<int, double> {
-  enum { value = true };
-};
-
-/**
- * A short may be promoted to a double.
- */
-template <>
-struct ad_promotable<short, double> {  // NOLINT(runtime/int)
+struct ad_promotable<bool, double> {
   enum { value = true };
 };
 
@@ -91,26 +52,18 @@ struct ad_promotable<char, double> {
 };
 
 /**
- * A bool may be promoted to a double.
+ * An unsigned char may be promoted to a double.
  */
 template <>
-struct ad_promotable<bool, double> {
+struct ad_promotable<unsigned char, double> {
   enum { value = true };
 };
 
 /**
- * An unsigned long may be promoted to a double.
+ * A short may be promoted to a double.
  */
 template <>
-struct ad_promotable<unsigned long, double> {  // NOLINT(runtime/int)
-  enum { value = true };
-};
-
-/**
- * An unsigned int may be promoted to a double.
- */
-template <>
-struct ad_promotable<unsigned int, double> {
+struct ad_promotable<short, double> {  // NOLINT(runtime/int)
   enum { value = true };
 };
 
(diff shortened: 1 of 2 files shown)


rok-cesnovar:
[These ](https://github.com/stan-dev/math/pull/1046/commits/cf1a9e0a6893107008cd66459edebd72c141692a)are the changes I had to do in order to make things run and pass on Windows. It seems that `matrix_exp_computeUV::run` is having problems with stan::math::var Eigen matrices on Windows. The current solution is just a quick fix to see if this also works on CI. We need to find a more general solution probably.

@yizhang-cae can you please verify that this fix is actually OK.

syclik:
The first problem I'm seeing is:
- [ ] `test/unit/math/prim/mat/err/check_pos_definite_test.exe`

rok-cesnovar:
Oh, yeah. 

This test causes the Windows Error Reporting popup on the EXPECT_DEATH, because the child process crashes as part of this death test. You need to manually cancel the popup and then the test continues and passes the tests. I did a quick fix based on [this](https://github.com/google/googletest/pull/1120) just to see if this fixes it. Should we upgrade the gtest lib?

syclik:
Wow, you're fast! I was trying to dig into how to deal with the problem and
still searching whether it works on Windows.

Yes, I think we should update Google Test. It makes sense to be an
independent issue + PR. Want to get that started? Or I can write up the
issue if that helps.

rok-cesnovar:
Yeah, no problem. I will do the issue tonight. The PR will have to wait till tomorrow, I have to finish up the docs stuff for the latest GPU PR.

rok-cesnovar:
After merging the gtest changes, tests now fail at the matrix_exp_multiply test. 
I have written down the stuff I narrowed down so far on the seg fault in the discourse thread: https://discourse.mc-stan.org/t/the-math-library-is-not-passing-tests-on-windows-this-is-the-next-thing-im-working-on/5989/12?u=rok_cesnovar

syclik:
Thanks! I'll set aside time to jump in this week.

syclik:
Ok... there's a fix for the segfault in `matrix_exp_multiply`. Hopefully that's it and we can get things testing on Windows again.

seantalts:
Thanks all! Taking a look and trying to get my bearings. Have a few questions:
> ad_promotable had to be updated to include long double
Seems like it had long double before, here:
https://github.com/stan-dev/math/pull/1046/files#diff-0b6a929f9f467e93bc45dd154b930bfaL41
A lot of the promotable things seem to have been reordered, potentially? Just to double check, the order doesn't matter for semantics does it? Could someone possibly summarize what that bit is about for me? The rest of the changes look good.

syclik:
@seantalts... I was incorrect. We had to add: `long long` and `unsigned long long`. (I just combed through all the changes and those were the two new ones.) I'll update the top-level comment.

I'm not sure why it was reordered, but the order shouldn't matter for semantics. Since we're being specific, the ADL rules should guarantee that it is able to promote.

It looks like there are 0 tests for `ad_promotable`. If you think it'd help, we should add them.

stan/math/prim/scal/meta/ad_promotable.hpp
```diff
@@ -107,26 +108,42 @@ struct ad_promotable<unsigned long, double> {  // NOLINT(runtime/int)
 };
 
 /**
- * An unsigned int may be promoted to a double.
+ * A long long may be promoted to a double.
```
syclik: @seantalts: `long long` and `unsigned long long` are the two new specializations. (If you remove these two and try to run on Windows, it'll fail to compile.)

test/unit/math/rev/mat/prob/bernoulli_logit_glm_lpmf_test.cpp
```diff
@@ -37,7 +38,7 @@ TEST(ProbDistributionsBernoulliLogitGLM,
   for (size_t ii = 0; ii < 200; ii++) {
     Matrix<int, Dynamic, 1> y(3, 1);
     for (size_t i = 0; i < 3; i++) {
-      y[i] = Matrix<uint, 1, 1>::Random(1, 1)[0] % 2;
+      y[i] = abs(Matrix<int, 1, 1>::Random(1, 1)[0] % 2);
```
seantalts: Does this not work with `unsigned int` (instead of `uint`)? If it does, maybe we should test that instead?

seantalts:
Gotcha. Tests would be helpful but I think that this PR will leave the codebase better than we found it, so I'm happy to let it go through as-is - just one more question about the abs vs. unsigned int in the tests.

syclik:
@seantalts. I just added tests for `ad_promotable`. (It would have helped to have this test prior to the PR to isolate the issue, so I think it's good we add it now.)

test/unit/math/rev/mat/prob/bernoulli_logit_glm_lpmf_test.cpp
syclik: I just checked... it does work with `unsigned int` and I changed all the instances with that so it looks like the old code.

syclik:
@seantalts: updates since your review:
1. Added tests for `ad_promotable` for prim, rev, fwd, and mix.
2. Fixed a small typo for `ad_promotable` in prim
3. Changed the `abs(Matrix<int>)` to `Matrix<unsigned int>`. I agree that it's cleaner that way (and it mirrors the old code better.

syclik:
@seantalts, mind reviewing again? The last changes were due to a mistake I made when updating from `unit` to `unsigned int` (then dealing with linting after that).

syclik:
This commit fixes the error I introduced: https://github.com/stan-dev/math/pull/1046/commits/2c618de4022a03ae818a3d7da2dff295b540be24

Issue #1056, opened 2018-10-27
Add Operator Overloading for GPU functions


SteveBronder:
## Description

I want to add operator overloading for add, subtract, and multiply for the `matrix_gpu` type so that developers can use the functions in a more standard pattern. 

## Example
```cpp
  stan::math::matrix_gpu m_gpu(m);
  m_gpu = m_gpu * 2.0;

matrix_gpu rv11(rv1);
matrix_gpu rv22(rv2);
auto v33 = v11 + v22;
```
## Expected Output

Same as now

## Current Math Version
v2.18.0

PR #1057, opened 2018-10-27
Adds operators for addition, subtraction, and multiplication on the GPU

SteveBronder:
## Summary

This adds operators `*`, `+`, and `-` for `matrix_gpu` types.

## Tests

I updated the current tests we have to use the new operators, is that alright since they call the underlying function?
```
./runTests.py test/unit/math/gpu/subtract_test.cpp test/unit/math/gpu/multiply_test.cpp test/unit/math/gpu/add_test.cpp
```
## Side Effects

None to my knowledge

## Checklist

- [x] Math issue #1056 

Steve Bronder


SteveBronder:
I also noticed that we were missing 
```cpp
#include <stan/math/prim/mat/err/constraint_tolerance.hpp>
#include <stan/math/gpu/err/check_square.hpp>
```

in `check_symmetric` so I added those to this branch if that's cool. We should probably have `STAN_OPENCL` defined in the header tests so that doesn't get missed by Jenkins in the future

stan/math/gpu/multiply.hpp
```diff
@@ -91,6 +91,23 @@ inline matrix_gpu multiply(const matrix_gpu& A, const matrix_gpu& B) {
...
+ *
+ * @param A A matrix or scalar
+ * @param B A matrix or scalar
+ * @return the product of the first and second arguments
+ *
+ * @throw <code>std::invalid_argument</code> if the
+ *   number of columns in A and rows in B do not match
+ */
+template <typename T1, typename T2>
+inline matrix_gpu operator*(const T1& A, const T2& B) {
```
seantalts: Why is this templated? Aren't they always matrix_gpus?

test/unit/math/gpu/add_test.cpp
```diff
@@ -23,7 +23,8 @@ TEST(MathMatrixGPU, add_v_exception_pass_zero) {
   stan::math::matrix_gpu d11(d1);
   stan::math::matrix_gpu d22(d2);
   stan::math::matrix_gpu d33(0, 1);
-  EXPECT_NO_THROW(d33 = stan::math::add(d11, d22));
+  EXPECT_NO_THROW(d33 = d11 + d22);
+  EXPECT_NO_THROW(d33 = d11 + d22);
```
seantalts: double paste

test/unit/math/gpu/add_test.cpp
```diff
@@ -13,7 +13,7 @@ TEST(MathMatrixGPU, add_v_exception_pass) {
   stan::math::matrix_gpu d11(d1);
   stan::math::matrix_gpu d22(d2);
   stan::math::matrix_gpu d33(3, 1);
-  EXPECT_NO_THROW(d33 = stan::math::add(d11, d22));
+  EXPECT_NO_THROW(d33 = d11 + d22);
```
seantalts: A different path here might be to just add a test that shows that `+` calls `add` for all the right types...  There are pros and cons to each. Cons to this approach replacing all `add`s with `+`:
1) the Stan language actually generates code that calls `add`, not `+`, so in some sense these are testing a non-primary API
2) No longer mirrors the other `add_test.cpp`s
3) You had to change more code, but that's basically done now.

stan/math/gpu/multiply.hpp
SteveBronder: For this particular operator, you can have the left or right side be a double. I was being lazy, but I think it's fine because it will only template when the return type is `matrix_gpu`?

If not then writing these signatures out is fine by me.

seantalts: I don't think that C++ is capable of using the return type in template deduction: https://stackoverflow.com/questions/8165659/why-cant-c-deduce-template-type-from-assignment

So I think this could get pulled in at weird times... not sure, might need to run some tests and make sure it doesn't mix poorly with normal matrix operations. Could you perhaps add some of those? I would say something that imports both gpu/multiply and rev/mat/multiply and tries operator overloading with a variety of types.

SteveBronder: Oh whoops totally my bad, I assumed so. I'll add some test types. Totally fine with adding the particular signatures here as well

seantalts: I think some simple tests of this will be good anyway! But yeah, they might reveal that we need to add the particular signatures (or more specific templated versions).

test/unit/math/gpu/add_test.cpp
SteveBronder: > the Stan language actually generates code that calls add, not +, so in some sense these are testing a non-primary API

While it's testing a non-primary api, the GPU stuff is also a non-primary API right? As in, all of the GPU functions will actually be called in their prim and rev counterparts, so even generated code (as of now) would not directly touch these APIs.

> expressing more of the API a Math library programmer would like to use

Yeah the main reason for adding these is so it will be a little easier to write out other things in prim and rev so they can use the GPU

stan/math/gpu/multiply.hpp
SteveBronder: Sorry for my delay on this, busy work week. I removed the templating and added particular signatures for each one.

seantalts:
This one definitely seems to have forgotten a ifdef STANCL

stan/math/gpu/multiply.hpp
```diff
@@ -104,10 +104,15 @@ inline auto multiply(const matrix_gpu& A, const matrix_gpu& B) {
...
-template <typename T1, typename T2>
-inline matrix_gpu operator*(const T1& A, const T2& B) {
+inline matrix_gpu operator*(const matrix_gpu& A, const matrix_gpu& B) {
   return multiply(A, B);
 }
+inline matrix_gpu operator*(const matrix_gpu& A, const double scalar) {
+  return multiply(A, scalar);
+}
+inline matrix_gpu operator*(const double scalar, const matrix_gpu& B) {
+  return multiply(scalar, B);
```
seantalts: You should probably just make this also be `multiply(B, scalar)` to avoid confusion.

stan/math/gpu/add.hpp
```diff
@@ -21,7 +21,7 @@ namespace math {
  * input matrices do not have matching dimensions
  *
  */
-inline matrix_gpu add(const matrix_gpu& A, const matrix_gpu& B) {
+inline auto add(const matrix_gpu& A, const matrix_gpu& B) {
```
seantalts: This is fine but in general I think it's fine to leave return types as non-auto especially here where it's not a lot of typing to specify. Types are also documentation even if the compiler can infer them.

SteveBronder:
@seantalts from Jenkins
```
Cannot contact ec2-spot-c5d18x: hudson.remoting.ChannelClosedException: Channel "unknown": Remote call on JNLP4-connect connection from ec2-52-14-54-225.us-east-2.compute.amazonaws.com/52.14.54.225:47546 failed. The channel is closing down or has closed down
```

seantalts:
Restarting the node.

SteveBronder:
Passing tests!

Issue #1062, opened 2018-11-07
make `coupled_ode_system` more efficient


wds15:
## Description
Solving ODEs in Stan is slow... and currently the `coupled_ode_system` creates `var` instances for the parameters in each iteration while these can be left on the outer AD stack.

## Expected Output
Same output, just faster.

## Additional Information
Any additional information goes here.

## Current Math Version
v2.18.0

wds15:
fixed with #1066

PR #1066, opened 2018-11-17
Feature/issue 1062 ode speedup

wds15:
Ok, I found a way which I think is totally fine with our AD logic. We are still getting a nice 18% or so speedup on the sir example.

## Summary

The intent of the PR is to speedup the ODE integration. The approach is to avoid repeated allocation of the parameters vector `theta_` on the nested AD stack. The nested AD is carried out when calling the `coupled_ode_system` functor. Since the parameters will never change during ODE integration we do not need to have these on the nested AD stack. The speedup of the SIR performance benchmark when running in 5 repetitions is 19%:

```
develop: stat_comp_benchmarks/benchmarks/sir/sir.stan,98.8624837399
this PR: stat_comp_benchmarks/benchmarks/sir/sir.stan,80.3885923386

[10:11:37][sebi@sebastians-macbook-pro-1:~/work/performance-tests-cmdstan]$ ./comparePerformance.py performance.csv develop_performance.csv
('stat_comp_benchmarks/benchmarks/sir/sir.stan', 0.81)
```

## Tests

```
test/unit/math/rev/mat/functor/integrate_ode_adams_prim_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_adams_rev_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_bdf_rev_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_cvodes_grad_rev_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_bdf_prim_test.cpp
test/unit/math/rev/arr/functor/integrate_ode_rk45_grad_test.cpp
test/unit/math/rev/arr/functor/integrate_ode_rk45_tooMuchWork_test.cpp
test/unit/math/rev/arr/functor/integrate_ode_rk45_test.cpp
test/unit/math/prim/arr/functor/integrate_ode_rk45_test.cpp
test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp
test/unit/math/prim/arr/functor/coupled_ode_system_test.cpp
```

## Side Effects

Faster ODE integration.

## Checklist

- [X] Math issue #1062 

- [X] the new changes are tested (no behavior as tested so far is changed)


wds15:
Bump. Once tests finish this can be reviewed and merged I think.

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -72,7 +72,7 @@ struct coupled_ode_system<F, double, var> {
       : f_(f),
         y0_dbl_(y0),
         theta_(theta),
-        theta_dbl_(value_of(theta)),
+        theta_copy_(theta),
```
syclik: I discussed with @wds15. This is not the right copy; we need to create a copy that is disconnected from `theta` and is on our `no_chain` stack.

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -338,7 +336,7 @@ struct coupled_ode_system<F, var, double> {
   std::vector<double> initial_state() const {
     std::vector<double> initial(size_, 0.0);
     for (size_t i = 0; i < N_; i++)
-      initial[i] = y0_dbl_[i];
+      initial[i] = value_of(y0_[i]);
```
syclik: This makes sense. This is only used once, so no need to stash the initial conditions as a separate vector.

stan/math/rev/arr/functor/coupled_ode_system.hpp
syclik: Please test that there's nothing added to the stack on construction.

wds15: Ok, I am using the nonstack thing now (really ugly to get things there!). However, I am not sure how to test and why to test that nothing gets on construction onto the stack. Ok, I could use `ChainableStack::instance().var_stack.size()`... but that is a not an exported API and is used nowhere in any test which I could find such that I am not sure that this can be tested given our API. If you think it is important to test, can you suggest a scheme to do so?

wds15: Never mind... I figured that we can test the stack size changes if we nest things.

... this is one of the annoyances of our nesting system - things are slightly inconsistent wrt to the non-nested stack; at least to me.

So the coupled_ode_system is now tested to not put stuff on the AD stack upon construction.

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -45,7 +45,7 @@ struct coupled_ode_system<F, double, var> {
   const F& f_;
   const std::vector<double>& y0_dbl_;
   const std::vector<var>& theta_;
-  const std::vector<double> theta_dbl_;
+  std::vector<var> theta_copy_;
```
syclik: Could we please rename this to something descriptive? This is meant to be a copy of the `theta`, but placed on the global stack, but essentially treated as temporaries where we can reset the adjoints. I don't know what a good name is, but just calling it `theta_copy_` really isn't that great.

The old name was ok because it implies we've lost the connections between `theta_` and `theta_dbl_` by demoting it to a `double`.

syclik: (For the top-level comment... I can't comment through the PR interface on lines of code that aren't really affected.)

Can you document what this implementation does? I think it'd help anyone else trying to maintain the code.
- that the `theta` is copied and put on the stack, but are not meant to be autodiffed
- that there's use of nested autodiff to compute the sensitivities

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -443,15 +442,16 @@ struct coupled_ode_system<F, var, var> {
...
-        theta_dbl_(value_of(theta)),
         x_(x),
         x_int_(x_int),
         N_(y0.size()),
         M_(theta.size()),
         size_(N_ + N_ * (N_ + M_)),
-        msgs_(msgs) {}
+        msgs_(msgs) {
+    for (auto& p : theta)
+      theta_copy_.emplace_back(var(new vari(value_of(p), false)));
```
syclik: If we know `p` is a `stan::math::var`, why not use `v.val()` directly? I think it'd be a little clearer.

stan/math/rev/core/var.hpp
```diff
@@ -21,7 +21,7 @@ static void grad(vari* vi);
  * Independent (input) and dependent (output) variables for gradients.
  *
  * This class acts as a smart pointer, with resources managed by
- * an agenda-based memory manager scoped to a single gradient
+ * an arena-based memory manager scoped to a single gradient
```
syclik: Thanks!

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -130,6 +130,8 @@ struct coupled_ode_system<F, double, var> {
         }
 
         set_zero_all_adjoints_nested();
+        for (size_t j = 0; j < M_; ++j)
```
syclik: Can you write out a comment that explains why this happens? It doesn't have to be here; it might be in describing the autodiff variable.

Something to consider... what if we double-nested?

Also, just cause I'm not certain, is this safe? If an exception is thrown while this is nested, what happens? Will a recover memory actually do the right thing if this is left in a nested state?

stan/math/rev/arr/functor/coupled_ode_system.hpp
wds15: To me the `value_of` is actually clearer as I don't need to recal the details of the `var` class. Calling `value_of` always gives me a double - no matter what. The `val` method feels like an internal function whereas the `value_of` is a clear API.

stan/math/rev/arr/functor/coupled_ode_system.hpp
wds15: I think it is safe, yes.

If an exception is thrown during the nested AD sweep then we catch that, recover the nested memory and re-throw. The `theta_copy_` is placed on the outer AD stack and will get cleaner by the outer recovery operation (whatever that is).

stan/math/rev/arr/functor/coupled_ode_system.hpp
syclik: I'll try to explain why I find this so much more confusing than if this was:
```
for (var&& p : theta) 
  theta_copy_.emplace_back(var(new vari(p.val(), false));
```

And this is coming from me having to read this code to figure it out.
1. `value_of()` is a function that's designed to work on `double`, `var`, `fvar<double>`, `fvar<var>`, and `fvar<fvar<var>>`. So I'm looking at `value_of(p)` and trying to figure out... "what is p?"
2. `auto& p` instead of `var&& p` doesn't help me at all. I know that `p` is one of `theta`, so `theta` is a collection, but what is it? Since I'm flipping between base template classes that can handle all autodiff types, it's not always easy to keep track that this is a `rev` only implementation. It'd really just help to have this be `var&& p` so the type is obvious.
3. `theta`... now I need to go and find where this is defined. C++ is a structured language, but it actually allows the declaration of struct and class member variables either at the top, below, or anywhere in between. And now I've got to search. I was looking at this using the GitHub interface, so the actual declaration was off of the interface; in order for me to go look at this, I had to go and check out the branch, open the file, all to find that `theta` is declared as `std::vector<var>`.

I think the thing that we're not seeing eye-to-eye on is the use of `value_of`. I find it much easier to reason about something that's only valid for `var` instead of having to work to figure out whether the object passed in is one of many different types. (It's just a little more mental work the reader has to do when it can all be made explicit as an alternative.)

(By the way, your statement is incorrect -- `value_of` does not always give a `double`. It'll give the type of the `value`, which for `fvar<var>` will be `var`.)

Hopefully you've gotten some context on why this is harder to read for a reviewer or anyone else coming to the code later on. This is minor, but the idea still stands -- make it easier for the next person to read. It takes effort now, but it's so worth it later on.

wds15: Oh... `value_of` is defined differently in my head... I always forget about higher order things. With the correct definitions the `val` call is better, yeah. Sorry, but I am not used to meaning of `value_of`.

Honestly, I do expect the reviewer to look at the file and not just the diff from git. It is really hard to expect our changes to be self-explanatory on a single diff line. The auto vs actual type ... I read the opposite elsewhere. Your third point I do not quite understand. I am looping over theta which is the argument to the constructor.

charlesm93:
I can hop in and complete the review. I'll do a review of the code and make sure all of @syclik 's requests have been addressed.

stan/math/rev/mat/functor/cvodes_ode_data.hpp
```diff
@@ -150,11 +150,11 @@ class cvodes_ode_data {
    */
   inline void rhs(double t, const double y[], double dy_dt[]) const {
     const std::vector<double> y_vec(y, y + N_);
-    const std::vector<double> dy_dt_vec
+    const std::vector<double>& dy_dt_vec
         = f_(t, y_vec, theta_dbl_, x_, x_int_, msgs_);
     check_size_match("cvodes_ode_data", "dz_dt", dy_dt_vec.size(), "states",
                      N_);
-    std::copy(dy_dt_vec.begin(), dy_dt_vec.end(), dy_dt);
+    std::move(dy_dt_vec.begin(), dy_dt_vec.end(), dy_dt);
```
charlesm93: Could you explain why you are changing `std::copy` to `std::move`? This is not quite clear to me (based on comments and description of the PR). Is there a difference between the two if `dy_dt` is a vector of primitives?

stan/math/rev/mat/functor/cvodes_ode_data.hpp
```diff
@@ -165,14 +165,13 @@ class cvodes_ode_data {
...
    */
   inline int jacobian_states(double t, const double y[], SUNMatrix J) const {
-    const std::vector<double> y_vec(y, y + N_);
     start_nested();
-    std::vector<var> y_vec_var(y_vec.begin(), y_vec.end());
+    const std::vector<var> y_vec_var(y, y + N_);
     coupled_ode_system<F, var, double> ode_jacobian(f_, y_vec_var, theta_dbl_,
                                                     x_, x_int_, msgs_);
-    std::vector<double> jacobian_y(ode_jacobian.size(), 0);
+    std::vector<double>&& jacobian_y = std::vector<double>(ode_jacobian.size());
```
charlesm93: Could you explain why you're using R-value references here? I'm not very familiar with `&&`, but my understanding is that you are creating a new object on the R.H.S of the `=` operator anyway. In that case, the old and the new statement would be doing the same thing, i.e. compiled into the same code.

stan/math/rev/mat/functor/cvodes_ode_data.hpp
charlesm93: This comment applies to all the places where `std::copy` has been changed to `std::move`.

stan/math/rev/arr/functor/coupled_ode_system.hpp
charlesm93: This still needs a comment explaining why this for loop happens, see initial comment.

charlesm93:
> I discussed with @wds15. This is not the right copy; we need to create a copy that is disconnected from `theta` and is on our `no_chain` stack.

I get the basic idea based on the PR description but I'm not sure what the glitch was here. This makes me worry I'm missing important subtleties. @wds15 maybe we can discuss over a quick video call.

stan/math/rev/mat/functor/cvodes_ode_data.hpp
wds15: The intention is to allow the compiler further optimisations here. If we copy then we tell the compiler that we want two distinct copies... which we do not need. Changing this to a move tells the compiler that we don't need the stuff in the original structure any more. This way the compiler can hopefully inline better and avoid a few copies.

stan/math/rev/mat/functor/cvodes_ode_data.hpp
wds15: This is again an attempt to let the compiler optimise better. We do not need the Jacobian stored in this function at all. Telling the compiler that the vector `Jacobean_y` can be treated as r-value should allow the compiler to figure out that we actually want things to be written into `J`, because that where the r-value ends up. Thus, this avoids copying.

wds15:
@charlesm93 The idea is to have the parameters on a stack which is disconnected to overall AD tree. Such a place is the nonstack AD tape which I never knew before talking to @syclik about it (this explains the comment you quote).

I added some comments as you requested in the code. Maybe we shortly discuss this at the meeting later in case you have more questions?

stan/math/rev/mat/functor/cvodes_ode_data.hpp
charlesm93: I get that. I'm merely suggestion that the compiler already does all the optimization -- the resulting machine language code should be the same. Of course, it's not easy to test; recently I've started paying with https://godbolt.org/ to look into these things.

In any case, the new code is at least better than the old one, and for what it's worth, give us a better sense of what is going on. So I'm happy with the current commit.

charlesm93:
Ok, this all looks good to me. The only minor point missing is a explanatory comment on line 133 of `coupled_ode_system.hpp`. Once this is done, we should be good to go.

stan/math/rev/arr/functor/coupled_ode_system.hpp
syclik: I just realized that we might be able to do this a little cleaner, in a more-understandable way. Would it be possible to do this by `start_nested()` once, putting the `theta_no_chain_` on there, then `start_nested()` repeatedly like we currently do?

wds15: This would work, yes. But then we would instantiate a coupled ode system object and change the ad stack. That’s rather unclean to me, but this was an earlier implementation I did which I dismissed. Instantiating an object and messing with the ad stack seems not good to me.

syclik: I'm not sure I understand what you're saying (or I'm missing the nuance).

If I think I understand what you're saying, then this shows that it's not what you think: (start nested) https://github.com/stan-dev/math/blob/502487c511594ccb93eb979df5b8fe163becb417/stan/math/rev/core/start_nested.hpp

I'm really saying that we should not leave the `theta_copy_` on the stack anyway. We should clean up after ourselves. So... on construction, we'd `start_nested()`, then add the `theta_copy_` to the nested stack, do the same thing for each Jacobian call by nesting and unnesting. And then... when we construct the final vars to return, we should `stop_nested()` first, removing `theta_copy_` from the stack. It sounds like that's not what happens and instead, we do everything the same, but leave `theta_copy_` on the stack. I don't think we need to mess with the ad stack any more or less than what we already do.

stan/math/rev/arr/functor/coupled_ode_system.hpp
syclik: Thanks! The third point wasn't actually there... you're right. I should have looked at the constructor signature, not the actual type of the variable.

stan/math/rev/arr/functor/coupled_ode_system.hpp
wds15: I am afraid that we cannot clean up `theta_nochain_` ... and I think that is ok to me at least (we are using the AD stack here, so there can be side-effects).

If you make a `start_nested` call in the constructor of `coupled_ode_system` then you can get quite weird behaviour if you use `coupled_ode_system` on it's own:

```cpp

var v1 = 2;
var v2 = 4;
 {
   // construction of coupled ode will globally modify the AD stack
    coupled_ode_system mysystem(...);
   // how will this now work??
   v2 *= v1;
   // destruction of the coupled_ode_system will unnest the stack globally
}
// ... I am not sure what happens for v1 & v2 ...
```

So constructing a `coupled_ode_system` globally modifies the AD stack...I don't think this is a good thing to me. Leaving a few bits behind on the AD stack is the better choice, I think.

If you want to avoid leaving things behind, then we can put inside the `integrate*` function the `coupled_ode_system` into a nested block. That can be cleaned up.

wds15:
It would be really cool to see this change in the forthcoming stan release (which may happen soon given GPU code is complete)... so if we can come to a conclusion on this that would be great. ~18% performance bump for non-stiff ODE integration is really nice for our users. I understood that @charlesm93 is fine with the changes, but right now the former review from @syclik is still blocking this PR from merging.

In case I haven't answered any request (or things are unclear), let me know. Thanks.

stan/math/rev/arr/functor/coupled_ode_system.hpp
syclik: This last comment you made really scares me. It really makes me think that it's just setting us up for a disaster. I understand that we modify the global stack, but to have a side-effect where things are left in an improper state is a bit too risky for developer error.

> If you want to avoid leaving things behind, then we can put inside the integrate* function the coupled_ode_system into a nested block. That can be cleaned up.

This makes it sound like you know how to remove that behavior? Let's do it if possible! If it's not, let's reevaluate it. (To me, this is new information that this change really leaves the autodiff stack in an unsafe state between different operations and is the first one I've heard of in the Math library.)

syclik:
@wds15, can we set up a microbenchmark? I want to verify that it is indeed faster and by how much on the three OSes / compilers.

wds15:
nesting the coupled_ode_system instantiation does not work for integrate_ode_rk45 as it is living in prim (and there I cannot nest).

I pushed a version now which saves the adjoints of the theta parameter vector as a double vector in the constructor of coupled_ode_system and then restores the adjoints in its destructor. That will work I think.

I have no clue on microbenchmarks and how to set them up... we can [use this example](https://github.com/stan-dev/math/blob/1178f5f05e831b34eb53a837a38d9e6847837050/test/unit/math/rev/mat/functor/integrate_ode_rk45_stress_test.cpp)... ideally you turn the initials into double's to get the largest relative increase in performance. Could you set this up?

The doc is still for the old approach and I will update that once you are fine with the approach taken.

(To me the nicest approach is by leaving things behind on the stack - it is a really tough criterion to disallow that all stuff needs to be cleared from the tape).

syclik:
@wds15, I'll set it up.

Was there any doc you needed to complete?

> (To me the nicest approach is by leaving things behind on the stack - it is a really tough criterion to disallow that all stuff needs to be cleared from the tape).

I think it's worse the other way in terms of having side effects. The function calls aren't atomic and carry around state. If we're trying to push more towards concurrency, this will be a lot more difficult to deal with.

That said, we can do this if we carefully document it.

wds15:
What I really do not like about this approach now is that we mess with the global AD tape.

You are raising a good point about concurrency. From this perspective this PR can represent a risk which I am not sure if we want to take that given we embark on concurrency.

syclik:
Let’s think about whether this is actually worse than the original code.

From what I can tell, before this PR, each function call is atomic. Meaning
that if anyone changes the autodiff stack during the function call other
than directly from the function, we would have problems.

This PR, as written, does it require nothing to change the autodiff stack
while the object is constructed? Or is it something else?

Are we at risk of anyone trying to modify the autodiff stack while this
object is in scope? Would it be possible for someone in RStan to do this
with expose Stan functions?

If we’re not adding more risk, we should be able to do this if we add
enough doc. I’m not certain I’ve thought through everything yet.

wds15:
Here is my view:

- The current form... let's call it "backup-and-restore-adjoints": This form saves the adjoints of the theta's when constructing the `coupled_ode_system` and it restores the adjoints to their original values when that instance goes out of scope. During the use of the object the adjoints of theta (part of the global AD tree) get continuously changed as we use these to calculate gradients.
   - Pro: No left-overs on the global AD stack
   - Con: During use of the `coupled_ode_system` we modify the global AD tree!

- The old form "make-nochain-copy": This form makes a "deep" copy of theta into the nochain stack on the global AD tree. Only during creating of this copy things are concurrency unsafe, but thereafter everything is fine as the global AD tree is not being modified
  - Pro: Global AD tree is left unchanged during use of the `coupled_ode_system`
  - Con: A left-over in the nochain part of the outer global AD tree

Both operations are safe if nothing interacts with the AD tree during the `integrate*` calls... still I would much prefer the "make-nochain-copy" approach. The left-over should be OK, I think (it's not a lot of left-overs).

My view. Does that make sense? Your thoughts?

wds15:
2.19 is coming... any thoughts on this or do we need more time?

syclik:
Thanks for the summary. I think you've characterized the side effects
pretty well.

I suggest we move forward with the current design, but we must document it
carefully with the information we know. It's not a matter of if we hit
these side effects, it's now just a matter of when.

I'll try to help with it, but I'm focused on the other PR first. As you're
writing, please keep in mind the abstraction. So the coupled ode class
should state the conditions under which The autodiff stack stays valid.
Then the integrate functions should state that it's meant to be done as an
atomic unit.

wds15:
Ok. So I will add some doc to the `coupled_ode_system` and the `integrate*` functions. Great.

syclik:
Thanks! Try to think of the info you'd want months from now if you've
violated the conditions necessary for this to work. When I get back to
reviewing it, that's what I'm going to be looking for in the doc.

wds15:
Ok... heres my attempt to doc this. I was rather extensive in describing it and I hope it's clear what is going on.

wds15:
I don’t think we can merge this. We have to use the other approach. Will explain later.

syclik:
Ok. We can hold off on merging it.

If there's something new you learned about this approach, please share. For
now, I'll just mark as unapproved. We can close after there's a clearer
explanation. (I think it should work?)

syclik:
@wds15. I read your other note. Edits don't trigger notifications.

I agree. I'll make sure to read the doc before merging.

stan/math/prim/arr/functor/integrate_ode_rk45.hpp
```diff
@@ -44,6 +44,13 @@ namespace math {
  * method</a> as implemented in Boost's <code>
  * boost::numeric::odeint::runge_kutta_dopri5</code> integrator.
  *
+ * During ODE integration the global autodiff tape is continuously
```
syclik: I don't understand what you're trying to say here. Can you write that in plain words and maybe I can help?

It really seems like we should warn that this function cannot be used in parallel. Meaning that nothing can touch the autodiff stack while this function is in use. (Do I have that condition right?)

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -38,14 +38,32 @@ namespace math {
  * <p>The final M states correspond to the sensitivities with respect
  * to the second base system equation, etc.
  *
+ * <p>Note: For efficiency reasons the parameter vector is used as
```
syclik: I think this doc doesn't describe the conditions clearly.

Shouldn't this state that nothing else can touch the autodiff stack for the duration of this object? We need to start being careful about this because we're moving to parallel computation. (This particular condition requires dealing with this object with care and the user should be trying to limit the lifetime of the objects as much as possible.)

wds15:
Given that we are heading into the direction of more concurrency, we should probably find a way which does not modify the outer AD tree, really.

At least I would like to take one more moment to maybe find a way which does not add this additional complexity we are introducing here. If the assumption can not be made that nesting makes things independent, then I think we are very likely in trouble in the future. If nesting doesn't make things independent, then the only way to handle things is doing it like `map_rect` which would be very restrictive.

(I have the parallel reduce function in my mind which I just wrote and this one would not work with this design so that I would be hesitant)

Sorry for the back and forth, but I never felt good about modifying the outer AD tree. We probably should agree on not doing this which will make our "threaded life" a lot simpler.

I have something in mind which I will attempt to put into the code.

wds15:
Ok. So here is an approach which can work and has the best of both approaches. The idea is:

1. in the constructor of coupled_ode_system we call `start_nested` and then put the `theta` onto the nested no chain AD stack.
2. you would expect to put the `recover_memory_nested` into the destructor of coupled_ode_system, but that will not work as then the decoupling operation places things on the nested stack which gets cleared away.

Right now I address the 2. thing by doing the `recover_memory_nested` in the decoupled states operation.... which should not stay like this. So if this approach has merits to you, then we should turn the decouple_states operation into a static function of coupled_ode_system and call it after getting rid of the coupled_ode_system instance. This could work and give us

- full independence of the outer AD tree
- nothing left behind

If that makes sense to you, I would refactor accordingly. 

(I still think that the approach to leave things behind on the no chain outer AD stack is cleanest and the best compromise... but you feel obvious strongly against about that while I am strongly against modifying the outer AD tree given we head into a parallel world)

syclik:
> (I still think that the approach to leave things behind on the no chain outer AD stack is cleanest and the best compromise... but you feel obvious strongly against about that while I am strongly against modifying the outer AD tree given we head into a parallel world)

I'm not strongly opposed to the approach of leaving things behind on the AD stack. It is a design decision and I think that's probably the cleanest and we should go with that. I think that should be the safest thing to do given the circumstances.

In general, we shouldn't just leave things behind. If we do, we should clearly state what's going on and why we've done it this way. That's what I feel strongly about, not the fact that we do it. By questioning whether we can remove it, we've actually come a long way in the design discussion that we wouldn't have.

Are you ok with going back to putting things on the global stack on construction? We can't remove them on destruction. That means the constructor and the functor are the functions that need to be locked, but the AD stack can be modified outside of those times without a problem.

wds15:
Great. I much prefer to go back to the design which uses the no chain approach (but leave things behind).

I will revert the code and add some doc around that as to what is safe / what not / what the side-effect is.

At least in my head having the guarantee that starting a nested AD tree gives me an independent sub-tree is a very good one (I think @bob-carpenter said this once if I am not wrong). I would expect that this holds right now in Stan-math - and we should keep it like this.

bob-carpenter:
> On Apr 1, 2019, at 9:15 AM, wds15 <notifications@github.com> wrote:
> 
> Great. I much prefer to go back to the design which uses the no chain approach (but leave things behind).
> 
> I will revert the code and add some doc around that as to what is safe / what not / what the side-effect is.
> 
> At least in my head having the guarantee that starting a nested AD tree gives me an independent sub-tree is a very good one (I think @bob-carpenter said this once if I am not wrong).

That was certainly the intention.

> I would expect that this holds right now in Stan-math - and we should keep it like this.

I agree.

wds15:
I reverted to the make-nochain-copy approach described above which we agreed on. I feel much safer with this!

The only side-effect of this approach is that instances of the parameter vector theta will be left on the nochain AD stack (even after the `coupled_ode_system` instance is gone). Other than that I don't see how anything can break this approach in a parallel world where we are heading.

I added a note to the respective specialization which do this trick which explains what is done and what side-effects are. So this is ready for final review.

Tagging @syclik and @charlesm93 who are well familiar with the context here.

stan/math/prim/arr/functor/integrate_ode_rk45.hpp
syclik: What is meant by "the adjoints of the parameter vector are used for Jacobian calculations"? And how does that interact with concurrency?

test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp
```diff
@@ -39,15 +41,21 @@ TEST_F(StanAgradRevOde, coupled_ode_system_dv) {
   z0.push_back(1.0);
   z0.push_back(2.0);
 
+  std::size_t stack_size = stan::math::nested_size();
+
   coupled_ode_system<harm_osc_ode_fun, double, stan::math::var> system(
       harm_osc, y0, theta, x, x_int, &msgs);
 
+  EXPECT_EQ(stan::math::nested_size(), stack_size);
```
syclik: What exactly is this testing? It looks like you're checking to see that the constructor doesn't add anything to the autodiff stack. If that's it, for clarity, could you add a note to the test? Something like:

```
EXPECT_EQ(stack_size, stan::math::nested_size()) << "expecting no new things on the stack";
```

(And switch the order of the arguments. In Google Test, the first is what you expect to be true, the second is what is being tested.)

syclik: Additional question: is there a reason you didn't do this with just the stack size instead? Why use the nested autodiff?

wds15: Yes, I am testing for the stack size not increasing by the constructor. I think you asked for this a while ago.

I am happy to add that cout message and swap order (I did not know about order here).

I used nested AD, because there is no way to get the stack size of the AD tape which is not nested; at least the last time I checked.

wds15:
@syclik : The confusing doc you found leading to your question

> What is meant by "the adjoints of the parameter vector are used for Jacobian calculations"? And how does that interact with concurrency?

Is still a left-over from the old approach which used the outer AD tape. The approach which we implement now using the no chain stack is completely safe wrt to concurrency in the usual sense. I removed that comment there from the doc (which was also put into the cvodes_integrator file).

Once passes tests this should then be fine to merge.

Looks like there is once more the need for your approval. Sorry for that (I haven't actively dismissed your approval; probably that was automatic due to the change set).

Thanks!

(BTW, with the forthcoming independent AD in the parallel design docs we will be able to write this optimisation in a way such that nothing will be left behind, I think ... but that will still take some time until it lands, of course)

syclik:
Thanks for removing that comment -- I was scratching my head at that for a little bit, but I'm glad it was cleaned up.

Yes, the way the approvals work now, it needs to be re-approved once a commit has been made. It's a pessimistic view of the world, but it's to prevent someone maliciously committing whatever they want after getting an approval and merging that. I think it's the right approach to take for an open-source project like ours.

test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp
syclik: Thanks for the answers! That makes sense. And it does make sense to check for this... we don't want a future implementation to just stomp on this behavior accidentally.

wds15:
It makes sense to re-approve if there are changes...no objection to that. I just wasn't aware of it and always feel guilty to use peoples time if not actually needed.

wds15:
Whow! 6 month PR being around / 4 month under review or so / finally merged. Cool.

I know this PR bends our usual conventions... but it is worth it and at least I learned a lot about our AD reverse mode things.

Thanks @syclik for bearing with me on this one!

mattfidler:
@wds15 doesn't this imply that time-varying covariates (ie. thetas that change) won't work correctly?

Issue #1010, opened 2018-08-28
Replace boost::isfinite, etc. with std:: versions


bob-carpenter:
## Description

We should move functionality from boost to std libraries wherever possible.  There are a bunch of runtime numerical tests like isfinite and isnan that are supported in std now and we should move to those versions.

1.  Figure out which functions are now supported in std.

2.  Change the Stan code to use the std versions.

## Example

Line 18 of `prim/err/check_finite.hpp`:

```
 if (!(boost::math::isfinite)(value_of_rec(y)))
```

Also, we can get rid of a lot of the type promotion in Boost and just use `auto` as the return type. For example, stan/math/prim/mat/fun/add.hpp could just be:
```
#ifndef STAN_MATH_PRIM_MAT_FUN_ADD_HPP
#define STAN_MATH_PRIM_MAT_FUN_ADD_HPP

#include <stan/math/prim/mat/fun/Eigen.hpp>
#include <stan/math/prim/mat/err/check_matching_dims.hpp>

namespace stan {
namespace math {

/**
 * Return the sum of the specified matrices.  The two matrices
 * must have the same dimensions.
 * @tparam T1 Scalar type of first matrix.
 * @tparam T2 Scalar type of second matrix.
 * @tparam R Row type of matrices.
 * @tparam C Column type of matrices.
 * @param m1 First matrix.
 * @param m2 Second matrix.
 * @return Sum of the matrices.
 * @throw std::invalid_argument if m1 and m2 do not have the same
 * dimensions.
 */
template <typename T1, typename T2, int R, int C>
inline auto
add(const Eigen::Matrix<T1, R, C>& m1, const Eigen::Matrix<T2, R, C>& m2) {
  check_matching_dims("add", "m1", m1, "m2", m2);
  return (m1 + m2).eval();
}

/**
 * Return the sum of the specified matrix and specified scalar.
 *
 * @tparam T1 Scalar type of matrix.
 * @tparam T2 Type of scalar.
 * @param m Matrix.
 * @param c Scalar.
 * @return The matrix plus the scalar.
 */
template <typename T1, typename T2, int R, int C>
inline auto
add(const Eigen::Matrix<T1, R, C>& m, const T2& c) {
  return (m.array() + c).matrix().eval();
}

/**
 * Return the sum of the specified scalar and specified matrix.
 *
 * @tparam T1 Type of scalar.
 * @tparam T2 Scalar type of matrix.
 * @param c Scalar.
 * @param m Matrix.
[...]

PR #1067, opened 2018-11-18
Feature/issue 1010 - Replace boost::math:: functions with cmath (Part 1)

andrjohns:
## Summary

Addresses part 1 of issue #1010, replacing ```boost::math::``` functions that don't need input checking with their c++11 equivalents. 

The following functions have been replaced:
- asinh
- cbrt
- erf
- erfc
- expm1
- isinf
- isnan
- round
- trunc

## Tests
No major changes to tests, only replacing ```boost::math::foo()``` with ```stan::math::foo()``` where relevant

## Side Effects

N/A

## Checklist

- [x] Math issue #1010 


bob-carpenter:
> I get that we have to sort out the thread-safe thing, but why persist in Boost's convention of returning NaN when the input is zero when std::lgamma correctly returns infinity?

My PR for just `lgamma`, which I have closed without merging, just went with the `std::lgamma` values.  I agree with Ben for two reasons---returning infinity makes the most sense mathemaitcally and we should be following the standard library conventions wherever possible.

andrjohns:
The difficulty with the standard library and errors is that it never actually throws. There's a clearer writeup [here](https://codeyarns.com/2015/01/05/math-errors-in-c-and-how-to-prevent-or-catch-them/), but essentially the function returns ```-nan``` and the ```errno``` variable is set to ```EDOM``` or ```ERANGE``` depending on whether a range or domain error occurs. The user then has to check this ```errno``` variable after calling the function to determine whether an error occurred.

As far as my google-fu can tell, there's no way to 'enable' throwing on domain/range errors int he standard library functions.

bob-carpenter:
> As far as my google-fu can tell, there's no way to 'enable' throwing on domain/range errors int he standard library functions.

I also want to continue to follow the standard library's throw protocol.  This is why we explicitly specified Boost error policies before---to make sure they did *not* throw.

If you want to throw, the way to do it is:

```
...
double y = std::lgamma(x);
if (...error code is on...)
  throw ...;
return y;
```

The tricky part there is thread safety for error codes.

andrjohns:
@bob-carpenter just confirming the implementation for throwing here before I update all functions. 

Using acosh as an example:
```
inline double acosh(double x) {
  errno = 0;
  double y = std::acosh(x);
  if (errno != 0)
    throw std::domain_error(strerror(errno));
  return y;
}
```
The ```errno``` variable has to be reset to 0 before each function call, otherwise an error code from a previous call could remain. In terms of thread-safety, since c++11 each thread has its own ```errno``` variable that gets written/called (https://en.cppreference.com/w/cpp/error/errno).

The alternative to this is manually checking for valid inputs (consistent with the standard library) and throwing prior to calling the function:
```
inline double acosh(double x) {
  check_greater_or_equal("acosh", "x", x, 1.0);
  return std::acosh(x);
}
```

Is there a preference?

bob-carpenter:
Sorry for not being clearer.  Let me try again so there's no ambiguity:

*** We only want to throw for C++ standard library functions when those functions throw (i.e., never). ***

We can change that behavior in Stan 3, but not before then, because it will actually change the behavior of programs.  Someone could apply a function then test for nan and proceed in the current setup, but there's no way to recover from a throw in the Stan language.

The first form below will not work without expensive thread-local synchronization, at least when Stan's set for multi-threading mode.   Whenever we do throw, we want informative message that mentioned "acosh" and the argument and why it was out of bounds.  So the second version would be the way to proceed.  But see the above---***we do not want acosh to throw ever***.

bgoodri:
No, if you validate the input beforehand then neither Boost nor the
standard library functions will throw or fail.

andrjohns:
Right, but validating the input beforehand (i.e. ```check_greater_or_equal("acosh", "x", x, 1.0);```) results in throwing, and Bob's saying that it should not throw at all?

bgoodri:
Throwing from Stan functions is fine (at least for now) because we can
control the exception type and message.

sakrejda:
@andrjohns I'm glad you find this as confusing as I do (I've had a few very similar conversations).  It's confusing for everyone.  I don't think there's a clear answer for what to do when switching from boost::lgamma to std::lgamma because 1) we don't want to change the behavior of current programs; and 2) we are (currently) following what the std library does when using std functions.  We can have both so somebody has to pick one.    

@syclik is (is going to be?) the lead for the math library so it makes sense for him to decide what to do with this.

bob-carpenter:
> Ah I understand what you mean now. The main reason that I was confused is that the current tests expect that the function throws, and so I was trying to maintain the current behaviour.

I don't think so.  Here's the current version (`stan/math/prim/scal/fun/acosh.hpp`):

```
inline double acosh(double x) {
  if (unlikely(is_nan(x)))
    return x;
  else
    return boost::math::acosh(x, boost_policy_t());
}
```

> Won't implementing the standard library math functions in this way lead to inconsistencies with the boost math functions (i.e. the boost functions will throw while the standard functions will silently fail)?

No, we use policies on the boost functions so that they don't throw.  Here's the policy (`stan/math/prim/scal/fun/boost_policy.hpp`):

```
typedef boost::math::policies::policy<
    boost::math::policies::overflow_error<
        boost::math::policies::errno_on_error>,
    boost::math::policies::pole_error<boost::math::policies::errno_on_error> >
    boost_policy_t;
```

> Throwing from Stan functions is fine (at least for now) because we can control the exception type and message.

The math library policy right now is to follow the throw behavior of the C++ standard library.

> @syclik is (is going to be?) the lead for the math library so it makes sense for him to decide what to do with this.

Is.  He hasn't expressed an opinion yet.
It's confusing for everyone. I don't think there's a clear answer for what to do when switching from boost::lgamma to std::lgamma because 1) we don't want to change the behavior of current programs; and 2) we are (currently) following what the std library does when using std functions. We can have both so somebody has to pick one.

I don't know why everyone's confused.  The intended behavior is to follow the standard library for standard library functions.  The current calls to Bost are

bob-carpenter:
And just to be super explicit, here's the reverse-mode implementation:

```
inline var acosh(const var& a) {
  return var(new acosh_vari(stan::math::acosh(a.val()), a.vi_));
}
```

It calls the basic math version so it also won't throw.

andrjohns:
> I don't think so. Here's the current version (stan/math/prim/scal/fun/acosh.hpp):

It's not defined in the function header, but the throwing is expected nonetheless. This is from ```acosh_test.cpp``` in the vanilla math library:
```
TEST(MathFunctions, acosh_exception) {
  using stan::math::acosh;
  EXPECT_THROW(acosh(0.5), std::domain_error);
}
```
> No, we use policies on the boost functions so that they don't throw. Here's the policy (stan/math/prim/scal/fun/boost_policy.hpp):

That policy sets the behaviour for pole and overflow errors, not domain errors. If you update the policy to add the same behaviour for domain errors:
```
typedef boost::math::policies::policy<
    boost::math::policies::overflow_error<
        boost::math::policies::errno_on_error>,
    boost::math::policies::pole_error<boost::math::policies::errno_on_error>,
    boost::math::policies::domain_error<boost::math::policies::errno_on_error> >
    boost_policy_t;
```
Then re-run the test for acosh:
```
test/unit/math/prim/scal/fun/acosh_test --gtest_output="xml:test/unit/math/prim/scal/fun/acosh_test.xml"
[==========] Running 4 tests from 1 test case.
[----------] 4 tests from MathFunctions
[ RUN      ] MathFunctions.acosh
[       OK ] MathFunctions.acosh (0 ms)
[ RUN      ] MathFunctions.acosh_exception
test/unit/math/prim/scal/fun/acosh_test.cpp:17: Failure
Expected: acosh(0.5) throws an exception of type std::domain_error.
  Actual: it throws nothing.
[  FAILED  ] MathFunctions.acosh_exception (0 ms)
[ RUN      ] MathFunctions.acosh_inf_return
[       OK ] MathFunctions.acosh_inf_return (1 ms)
[ RUN      ] MathFunctions.acosh_nan
[       OK ] MathFunctions.acosh_nan (0 ms)
[----------] 4 tests from MathFunctions (1 ms total)

[==========] 4 tests from 1 test case ran. (1 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] MathFunctions.acosh_exception

 1 FAILED TEST
test/unit/math/prim/scal/fun/acosh_test --gtest_output="xml:test/unit/math/prim/scal/fun/acosh_test.xml" failed
```

bob-carpenter:
Hmm.  That's bad.  That's a bug in the traits and the tests, in my opinion, given the goal of matching the standard library throw behavior.    At the same time, I've been thinking that in Stan 3 we want to change behavior and have everything throw.  As is, we're in this intermediate situation where things like `sqrt` and `log` that have been implemented in `std::` all along don't throw, e.g.,

```
inline double sqrt(int x) { return std::sqrt(static_cast<double>(x)); }
```

but the functions like `lgamma` and `acosh` that were in C99 (and hence some C++ compilers but not all of them) were implemented with Boost in such a way that they didn't match the C99 (and now C++11) behavior.

The question is now what we want to do about it.  Let's see if @syclik has an opinion.  I'll try to ping him via email.  

My inclination is to not try to go and change all of our behavior back to matching the standard library only to change it up again.

andrjohns:
That sqrt code also brings up another point with the use of c++11 - the static_cast for integers is no longer needed. According to the spec, all integral types can be used directly with the std:: functions. [Using the sqrt ref as an example:](http://www.cplusplus.com/reference/cmath/sqrt/)

> Additional overloads are provided in this header (cmath) for the integral types: These overloads effectively cast x to a double before calculations (defined for T being any integral type).

syclik:
@andrjohns, sorry about the confusion. You're digging right into code that touches a lot of edge cases. I'll try to summarize what people are asking me to decide on, although I don't think there's a decision here.

I believe what people are asking is for Stan's math functions that are implemented in the standard library, what behavior should the functions have with the error conditions?

The answer is: first follow the existing Math library's implementation. Then fall back to the standard library's implementation. Then Boost. If the implementation throws, that's fine, but we would like to throw with a reasonable message. We currently check prior to calling a function to form an error message, but we don't have to do it that way (given the code is clean, the patterns are reasonable, and there isn't some sort of performance hit by doing it that way).

Hopefully that explains it a bit. To get into motivation, it's as Bob mentioned: changing behavior at these edge cases does change the behavior of the Stan programs. We won't do that at the Math library level unless we have a good reason and we announce that we will do this. At some point we should, but I don't think this PR should be changing a lot of behavior. I'd rather roll all of that up together in one release that is behavior-changing to happen all at once. This is why testing for known edge cases matter... turns out that Boost has changed its strategy for some functions over they years. We want to be consistent independent of what the standard library implementation on different compilers do and what Boost does.

Before this PR gets merged, it really needs to be tested on Windows. I know the standard library is supposed to implement a specification, but sometimes implementations don't behave the same way in either unspecified areas or edge cases.

@bob-carpenter, does that answer what you were asking? I'm saying we should match the current Math implementation first, no matter if that's inconsistent with either the standard library or Boost (or both). And then use the standard library's rules.

syclik:
> That sqrt code also brings up another point with the use of c++11 - the static_cast for integers is no longer needed. According to the spec, all integral types can be used directly with the std:: functions. [Using the sqrt ref as an example:](http://www.cplusplus.com/reference/cmath/sqrt/)

@andrjohns, feel free to remove the `static_cast`. As long as the behavior is the same, that's great. (We'll have to empirically verify that it's the same because doc isn't always correct.)

bob-carpenter:
> @bob-carpenter, does that answer what you were asking? I'm saying we should match the current Math implementation first, no matter if that's inconsistent with either the standard library or Boost (or both). And then use the standard library's rules.

Yes.  Thanks.

bob-carpenter:
> That sqrt code also brings up another point with the use of c++11 - the static_cast for integers is no longer needed. According to the spec, all integral types can be used directly with the std:: functions.

Simplifying the code is great.  The only downside is that it's more code changes to implement and to review.

andrjohns:
> Before this PR gets merged, it really needs to be tested on Windows

@syclik  I'll run the tests locally on windows before I push anything. Is the Rtools 3.5 compiler the right target (gcc 4.9.3, I believe)? Any other compilers I should test against?

syclik:
@bob_carpenter: no, it's currently not being tested. That's what I'm
working on (the heisenbug that I've posted a bit on has shown a real
problem with one of our functions, as far as I can tell). I'm definitely
prioritizing that over anything else on the math lib... it's just
super-slow cause it's deep in the autodiff stack and there's a lot less
expertise there in dealing with memory issues.

syclik:
Thanks! That's it. I'll take your word that it passes when it gets to it.

andrjohns:
Testing under windows was a good idea. For no reason that I can determine acosh (and only acosh) is returning a nan for an infinity input when it should be returning infinity:
```
int main() {
  double inf = std::numeric_limits<double>::infinity();
  std::cout << "acosh(inf) = " << std::acosh(inf) << std::endl;
  return 0;
}
```
On Linux:
```
andrew@VMLinux:~$ g++ --version
g++ (Ubuntu 8.2.0-7ubuntu1) 8.2.0

andrew@VMLinux:~$ g++ -std=c++11 test.cpp

andrew@VMLinux:~$ ./a.out 
acosh(inf) = inf
```
On Windows:
```
C:\Users\Andrew Johnson\Documents>g++ --version
g++ (x86_64-posix-seh, Built by MinGW-W64 project) 4.9.3

C:\Users\Andrew Johnson\Documents>g++ -std=c++11 test.cpp

C:\Users\Andrew Johnson\Documents>a.exe
acosh(inf) = nan
```

On Linux, I also tested with g++-4.8, g++-6, and g++-7 and all returned the correct output.

bob-carpenter:
Our goal's to have Stan behave the same way on all platforms, if that's what you mean by policy.

It might be that everything else is succeeding due to undefined behavior, which means that the check needs to be there for everything.  For instance, acosh's doc says this:  "If a domain error occurs, an implementation-defined value is returned (NaN where supported)"  I'm not even sure how to read this---does the NaN parenthetical supersede the implementation-defined value bit?  

https://en.cppreference.com/w/cpp/numeric/math/acosh

But it does look like the value for +infinity should return +infinity, as the doc says "if the argument is +∞, +∞ is returned"  It's silent about -infinity, though.

sakrejda:
> Our goal's to have Stan behave the same way on all platforms, if that's what you mean by policy.

I was trying to remember if platform-specific ifdef blocks were permited or if we should just add the pre-call check to all platforms.

bob-carpenter:
> trying to remember if platform-specific ifdef blocks were permited or if we should just add the pre-call check to all platforms.

Yes, we have lots of platform-specific and even compiler-specific ifdefs.  It's the nature of portable C++.  We'd rather avoid them if possible, but sometimes they're necessary.

andrjohns:
So for this case should a platform-specific ifdef or a blanket (all platform) input check be used?

bob-carpenter:
I'm not sure about the details, but the general rule should be that
if there's a bug on a platform, you code around it with a platform-specific
ifdef.  So if the standard says return infinity and Windows doesn't, then
a platform-specific ifdef would be in order.

andrjohns:
Sounds good to me, I'll make the changes and ping you when they're ready

syclik:
If you wanted to make this easier for the reviewer, it would help to split
the PR into one that has all the easy cases and one that has all the
special cases. But not necessary if it's a lot of effort to do.

For this to stick, we also need to have something in common place that will
prevent new PRs using Boost. We should handle that in a separate PR and tie
that to our continuous integration.

andrjohns:
That should be fine. I'm do one PR for the cases that are a straight swap from ```boost::math::foo()``` to ```stan::math::foo()```, and another for the functions that need error-checking/edge-case handling

syclik:
Thanks! That first one is going to be easy to review, even if it's lots of
changes. The second one is the one we need to look at more carefully, but
still straightforward.

andrjohns:
@bob-carpenter This is ready for review again. To simplify the review process, the changes have been broken into three parts (covered in more detail in [this comment](https://github.com/stan-dev/math/issues/1010#issuecomment-445527864). This pull replaces the boost functions that don't need input checking and throwing.

bob-carpenter:
Thanks for sticking with this.  I should be able to get through these early next week.

Issue #1068, opened 2018-11-19
test requires STAN_NUM_THREADS but doesn't recognize it


bob-carpenter:
## Description

Makefile for test `test/unit/math/prim/mat/functor/num_threads_test.cpp` doesn't work with default or with setting number of threads directly.  Here's what happens.  This is on Mac OSX High Sierra, with homebrew install of g++7.3.  Here's the `make/local`

```
CXX=/usr/local/Cellar/gcc@7/7.3.0/bin/g++-7
O=0
```

and here's where `g++-7` points:

```
~/cmdstan/stan/lib/stan_math(bugfix/1063-std-lgamma)$ /usr/local/Cellar/gcc@7/7.3.0/bin/g++-7 --version
g++-7 (Homebrew GCC 7.3.0) 7.3.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

I clean, set the environment variable, then try to make and it fails because it can't find the environment variable.

```
~/cmdstan/stan/lib/stan_math(bugfix/1063-std-lgamma)$ make clean-all
  removing test executables
rm -f lib/gtest_1.8.1/src/gtest-all.o
  removing generated test files
rm -f 
  removing doxygen
rm -f -r doc/api
  removing dependency files
  cleaning sundials targets
rm -f 
```

```
~/cmdstan/stan/lib/stan_math(bugfix/1063-std-lgamma)$ export STAN_NUM_THREADS=2
```

```
~/cmdstan/stan/lib/stan_math(bugfix/1063-std-lgamma)$ ./runTests.py -j5 test/unit/math/prim/mat/functor/num_threads_test.cpp 
------------------------------------------------------------
make -j5 test/unit/math/prim/mat/functor/num_threads_test
[...]

PR #1069, opened 2018-11-19
replace boost lgamma with std, fixes #1063

### LINKED ISSUE #1063: GoodGammaP Test broken on GCC 7.3
This came up on [this other issue](https://github.com/stan-dev/math/issues/943) - I think @bgoodri knew what was going on in these tests. He mentioned that the implementation of the function is fine, it's just our tests that are broken.

Here is a link to a test run with that (it happens to be using MPI, but these errors also exist without MPI): http://d1m1s1b1.stat.columbia.edu:8080/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1059/8/pipeline

bob-carpenter:
##  Summary

Replace Boost implementation of `lgamma` in `prim/scal/fun` with standard library version.

The goal is to solve #1063 so this works with g++ 7.3.  

## Tests

The makefiles are currently broken;  see #1068.   I don't believe our CI tests 7.3 and I can't test it without resolving #1068, so I'm tossing it up in a PR.

### DIFF
--- a/stan/math/prim/scal/fun/lgamma.hpp
+++ b/stan/math/prim/scal/fun/lgamma.hpp
@@ -1,41 +1,22 @@
 #ifndef STAN_MATH_PRIM_SCAL_FUN_LGAMMA_HPP
 #define STAN_MATH_PRIM_SCAL_FUN_LGAMMA_HPP
 
-#include <stan/math/prim/scal/fun/boost_policy.hpp>
-#include <boost/math/special_functions/gamma.hpp>
+#include <cmath>
 
 namespace stan {
 namespace math {
 
 /**
- * Return the natural logarithm of the gamma function applied to
- * the specified argument.
+ * Return the natural logarithm of the gamma function applied to the
+ * specified argument.  If the input is a non-positive integer, the
+ * result is positive infinity.  If the input is not-a-number, the
+ * result and gradient are not-a-number.
  *
-   \f[
-   \mbox{lgamma}(x) =
-   \begin{cases}
-     \textrm{error} & \mbox{if } x\in \{\dots, -3, -2, -1, 0\}\\
-     \ln\Gamma(x) & \mbox{if } x\not\in \{\dots, -3, -2, -1, 0\}\\[6pt]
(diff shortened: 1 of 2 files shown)


sakrejda:
The description of the std::lgamma implementations is pretty clear, there's a shared global so concurrent calls will get non-deterministic garbled results.  Unless we want to chase down whether the implementations relevant to us are thread-safe we should stick to boost::lgamma and just set the relevant boost policies to fix issue #1063

andrjohns:
I actually just made this change as part of #1067 with the diff for the lgamma function here: https://github.com/stan-dev/math/pull/1067/files#diff-6d2e0c5e89ad16a152ce65335ca2e93c

From reading the issues above it looks like the NaN return for lgamma(0) is not wanted, and the return should be the default (inf)? Or leave the boost implementation as-is (with the change to policy) for thread safety?

bgoodri:
In my opinion, Boost's default policy of returning NaN for lgamma(0) is
crazy. But particularly in statistical applications, the argument to lgamma
is usually constrained to be non-negative anyway, in which case returning
Inf is definitely the right limiting behavior.

bob-carpenter:
@andrjohns --- please don't stack multiple, independent functionality changes into a single PR.  I already have [a PR addressing `lgamma`](https://github.com/stan-dev/math/pull/1069). 

I'm still unclear on the POSIX implications of `std::lgamma` not being thread safe.  I thought @bgoodri would know how this all fits together.

andrjohns:
I should have been clearer, the replacement of ```boost::math::lgamma``` with ```std::lgamma``` was already a [part of my PR](#1067) replacing ```boost::math``` functions with ```std::``` functions, when I submitted it (only beat you by a day though). I didn't change an existing PR to make this replacement.

If you'd prefer to handle ```lgamma``` separately to the other ```boost::math::``` functions, I can remove it from my PR

bob-carpenter:
No, if you're dealing with them all at once, that's fine.

In the future, if you could add a note that you're going to be creating a PR for an issue, it'll cut down on duplicated work.  Luckily, this was only an hour or so's worth of work.

Before this gets merged, we need to answer the threading question about lgamma.

Issue #1073, opened 2018-12-04
matrix_exp_multiply seg faults


syclik:
## Description

`matrix_exp_multiply`, as written, currently seg faults.

## Example

This test current seg faults on my machine. I think it may be memory dependent, but not certain.
```
#include <stan/math/rev/mat.hpp>
#include <gtest/gtest.h>

TEST(MathRev, matrix_multiply_exp__vd) {
  Eigen::Matrix<stan::math::var, -1, -1> A(5, 5);
  A << -0.96871, 0.398827, 0.241306, 0.741373, 0.108926,
    0.888077, -0.915624, -0.373344, 0.255238, 0.717304,
    -0.0899219, -0.898862, -0.800546, -0.222652, -0.271382,
    0.683227, 0.827031, -0.780702, -0.104228, 0.885106,
    -0.996585, -0.097802, 0.739617, 0.235266, -0.0247717;


  Eigen::Matrix<double, -1, -1> B(5, 5);
  B << -0.96871, 0.398827, 0.241306, 0.741373, 0.108926, 0.888077, -0.915624,
    -0.373344, 0.255238, 0.717304, -0.0899219, -0.898862, -0.800546,
    -0.222652, -0.271382, 0.683227, 0.827031, -0.780702, -0.104228, 0.885106,
    -0.996585, -0.097802, 0.739617, 0.235266, -0.0247717;

  Eigen::Matrix<stan::math::var, -1, -1> result = stan::math::matrix_exp_multiply(A, B);

  std::vector<double> gradients;
  stan::math::print_stack(std::cout);

  std::vector<stan::math::var> vars = stan::math::to_array_1d(A);
  result(0, 0).grad(vars, gradients);

  std::cout << "gradients = " << std::endl;
  for (size_t j = 0; j < gradients.size(); ++j)
    std::cout << gradients[j] << " ";
  std::cout << std::endl;

  FAIL() << "There are no asserts in the tests." << std::endl
         << "This test seg faults with the existing implementation." <<	std::endl;
}
```

## Expected Output
Not seg fault. Return the same result as doing the actions separately.

## Additional Information
Thank you, @rok-cesnovar, for pushing the Windows tests to the point where we were able to find this. 

@yizhang-cae, can you look at the implementation and help identify why it's seg faulting? I'm going to patch this by using the naive implementation.

[...]

PR #1074, opened 2018-12-04
Bugfix/issue 1073 matrix exp mulitply: fixes segfault in matrix_exp_multiply

syclik:
## Summary

Fixes a seg fault.

## Tests

Includes a test that seg faults with the current implementation.

## Side Effects

This will add a lot more elements on the autodiff stack, but is correct. (~2600 vs 1.)

## Checklist

- [x] Math issue #1073


bob-carpenter:
This seems like it's going to introduce  a serious performance regression by replacing `matrix_exp_action(A, B);` with  `multiply(matrix_exp(A), B);`.    

I'm fully supportive of correctness taking precedence over efficiency, and the call about whether to take the performance hit on this is @syclik's.  

Nevertheless, I'd like to hear from  @charlesm93 and/or @yizhang-cae about why they didn't fix the bug and whether they care about the performance hit---the move to remove the efficient implementation in this PR may be enough to spur them to fix the original code.  If not, then by all means get someone to review this and merge in the fix for correctness.

syclik:
Yes, it's a performance regression, but the seg fault takes priority. I'd
prefer to see a safe, slow version merged first, then have a new PR for a
safe and fast implementation.

I'm going to try to run this branch on Windows to make sure we can get past
this failure.

yizhang-cae:
I'll take a look later today.

syclik:
Thanks. Would you mind reviewing the PR and approving if it looks ok.

Then we can go back and merge a fix that's efficient when that's available.

syclik:
@yizhang-cae: just FYI, that's not the only time it seg faults. That was
just the one I added an explicit test for that causes a failure.

yizhang-cae:
Thanks. I'm also curious why it doesn't fail on the rest unit tests.

syclik:
Yeah... I really want to find out what's causing it because I think we
might find the problem in other spots. Hopefully you get further than I do.

Issue #1075, opened 2018-12-04
Bug in map_rect_concurrent(); depends on number of threads and number of shards


syclik:
## Description
This is the bug that's causing stan-dev/stan#2708.

This bug occurs when threading is turned on and the number of threads is some function of the number of shards.

It occurs in the [`map_rect_concurrent()`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/functor/map_rect_concurrent.hpp#L71) function and this function returns an Eigen vector that is short one element. (`Eigen::Matrix<T, -1, 1>`, but the runtime size is off by one.)

## Example

As we dig, we need a reproducible example in C++, but for now, please see the Stan program in the Stan bug report: stan-dev/stan#2708.

## Additional Information
Thanks, @ermeel86, for uncovering this bug!

## Current Math Version
v2.18.0

ahartikainen:
Are these correct

    const int num_big_threads
        = (num_jobs - num_jobs_per_thread) % (num_threads - 1);

and

    for (int i = 1, job_start = num_jobs_per_thread, job_size = 0;
                i < num_threads; ++i, job_start += job_size) {
         job_size = i >= first_big_thread ? num_jobs_per_thread + 1
                                       : num_jobs_per_thread;
        futures.emplace_back(
             std::async(std::launch::async, execute_chunk, job_start, job_size));
    }

Should the first block be

    const int num_big_threads
        = num_jobs - num_jobs_per_thread * num_threads;

And should the second loop start from 0? TBH not sure about what is going on here, but I think the `job_size` is off-by-one for some chunks if the `num_big_threads` is wrong.

Some tests with Python

    num_threads = 6
    for num_jobs in range(num_threads, 3*num_threads):
        num_jobs_per_thread = num_jobs // num_threads
        print(num_jobs - num_jobs_per_thread * num_threads, (num_jobs - num_jobs_per_thread) % (num_threads - 1))

```
0 0
1 1
2 2
3 3
4 4
5 0
0 0
1 1
2 2
3 3
4 4
5 0
```


    num_threads = 2
    for num_jobs in range(num_threads, 3*num_threads):
        num_jobs_per_thread = num_jobs // num_threads
        print(num_jobs - num_jobs_per_thread * num_threads, (num_jobs - num_jobs_per_thread) % (num_threads - 1))

```
0 0
1 0
0 0
1 0
```

So it might just be a play between  the `num_jobs` and `num_threads` which is easier to hit with lower number of threads.

[...]

PR #1079, opened 2018-12-07
Bugfix/issue 2708 map rect fail

sakrejda:
Proposed fix to map_rect_concurrent.  @wds15 I tried updating the tests to do a range of shard sizes but I'm missing something about the combine step.  Currently I don't feel comfortable saying this is fixed unless we test a range of shard sizes.  It might make more sense to me if you explain the two checks in this test:

```
TEST_F(map_rect, concurrent_ragged_return_size_dd) {
  Eigen::VectorXd res1 = stan::math::map_rect<0, hard_work>(
      shared_params_d, job_params_d, x_r, x_i);

  EXPECT_EQ(res1.size(), 2 * N);

  x_i[1] = std::vector<int>(1, 1);

  Eigen::VectorXd res2 = stan::math::map_rect<1, hard_work>(
      shared_params_d, job_params_d, x_r, x_i);

  EXPECT_EQ(res2.size(), 2 * N + 1);
}
```

## Checklist

- [x] Math issue #1075 

- [x] Copyright holder: Krzysztof Sakrejda

### DIFF
--- a/stan/math/prim/mat/functor/map_rect_concurrent.hpp
+++ b/stan/math/prim/mat/functor/map_rect_concurrent.hpp
@@ -103,19 +103,17 @@ map_rect_concurrent(
 #ifdef STAN_THREADS
   if (num_threads > 1) {
     const int big_job_size = small_job_size + 1;
-    const int num_big_threads = num_threads - (num_jobs % num_threads); 
+    const int num_big_threads = num_threads - (num_jobs % num_threads);
     const int num_small_threads = num_threads - 1 - num_big_threads;
     int job_start = small_job_size;
     for (int i = 0; i < num_big_threads; ++i) {
-      futures.emplace_back(
-          std::async(std::launch::async, execute_chunk, job_start,
-	      big_job_size));
+      futures.emplace_back(std::async(std::launch::async, execute_chunk,
+                                      job_start, big_job_size));
       job_start += big_job_size;
     }
     for (int i = 0; i < num_small_threads; ++i) {
-      futures.emplace_back(
-          std::async(std::launch::async, execute_chunk, job_start,
-	      big_job_size));
+      futures.emplace_back(std::async(std::launch::async, execute_chunk,
+                                      job_start, big_job_size));
       job_start += small_job_size;
     }
   }
--- a/stan/math/prim/mat/functor/map_rect_concurrent.hpp
+++ b/stan/math/prim/mat/functor/map_rect_concurrent.hpp
@@ -96,21 +96,27 @@ map_rect_concurrent(
   };
 
   int num_threads = get_num_threads(num_jobs);
-  int num_jobs_per_thread = num_jobs / num_threads;
+  int small_job_size = num_jobs / num_threads;
   futures.emplace_back(
-      std::async(std::launch::deferred, execute_chunk, 0, num_jobs_per_thread));
+      std::async(std::launch::deferred, execute_chunk, 0, small_job_size));
 
 #ifdef STAN_THREADS
   if (num_threads > 1) {
-    const int num_big_threads
-        = (num_jobs - num_jobs_per_thread) % (num_threads - 1);
-    const int first_big_thread = num_threads - num_big_threads;
-    for (int i = 1, job_start = num_jobs_per_thread, job_size = 0;
-         i < num_threads; ++i, job_start += job_size) {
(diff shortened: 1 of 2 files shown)


bob-carpenter:
I don't see math issue 1708.  Is this addressing #1075?

sakrejda:
Fixed thanks.

wds15:
The idea of these tests is to test if the returned ragged vector has the correct size. The first test is for a case where each job returns 2 elements and the second test is to check if things work if one job (the first one) returns 3 instead of 2 elements.

stan/math/prim/mat/functor/map_rect_concurrent.hpp
```diff
@@ -96,21 +96,25 @@ map_rect_concurrent(
...
+    int job_start = small_job_size;
+    for (int i = 0; i < num_big_threads; ++i) {
+      futures.emplace_back(std::async(std::launch::async, execute_chunk,
+                                      job_start, big_job_size));
+      job_start += big_job_size;
+    }
+    for (int i = 0; i < num_small_threads; ++i) {
+      futures.emplace_back(std::async(std::launch::async, execute_chunk,
+                                      job_start, big_job_size));
+      job_start += small_job_size;
```
wds15: I like how you code it, but I think that you intended to schedule in the first for loop all the small jobs and in the second for loop all the big jobs. Also note that right now you schedule in both for loops `big_job_size` - but that should be different in the two loops (first the small, then the big ones).

sakrejda:
@wds15 can you answer my question up in the main PR description about the 'sizes' test? Why does changing the integer parameters (or running the command a second time with another id) increment the return size by 1? I'll look at it again now but if you have a quick answer to that question it would save me time with adding tests.

sakrejda:
re: joint testing with MPI, I agree that's a good idea, would also be good to separate the job counting out into shared code between the two and use a simple thread pool here (it's not bad with c++1y).. I'm going to leave those for later as this is a bugfix PR.  Definitely want more tests, we should've tested for varying shard/thread numbers to begin with. :/

sakrejda:
re: functor ok, no idea how I missed that, thanks.  I think I can do the tests now.

sakrejda:
@syclik how is the number of threads set for testing? I think we should expand the tests by running the current tests with a range of STAN_NUM_THREADS values but that doesn't seem to be in the testing framework itself.  Is that set via Jenkins or something like that?

wds15:
Have a look into the get num threads test. There the variable is set to different values.

wds15:
@sakrejda I am very close to have a PR ready which would allow us to use the parallel STL from Intel as backend in addition to the async facility. This will give us nice speedups - in particular when jobs have unequal cpu needs as the Parallel STL from Intel can deal with this.

This PR would, of course, contain a fix for this issue as well. The downside that this entails a few more lines to review, but not very hard.

The Parallel STL from Intel is rather easy to install via their binary distribution on Mac...and there is even hope for Windows, I think.

wds15:
uh... sorry to hear, but while not blessed with fever, I also do enjoy right now the season...

sakrejda:
@wds15 I think we should keep this branch as a bugfix rather than letting it creep.  I'm happy to review the parallel STL updates on a separate feature PR, just trying to get this in asap esp. since I've been out.

sakrejda:
There's still some segfault in here, with 5 shards and 4 threads.

wds15:
This PR is not quote right yet, I think.

We can work on this one, but the other parallel STL one has all fixes and adequate tests.

If we continue this one, we should at least take over some pieces (like the test) from the parallel STL. However, the parallel STL PR is not too massive ... all I am doing there is to refactor `map_rect_concurrent` to use `for_each` from C++17 (and providing a C++11 implementation of that).

syclik:
@wds15, please work on a fix first before any new improvements. Given that
this feature has lead to a major bug, this is high priority. And new
features will have to be evaluated carefully.

sakrejda:
@wds15 @syclik I'm closing this.  @wds15 Can you just take the relevant parts of the STL PR and do a new bugfix PR?

syclik:
@sakrejda, thanks for letting me know. Feel better.

@wds15, agreed. If you've fixed it on a branch that has a new feature, please separate out the fix so we can patch the issue immediately.

PR #1080, opened 2018-12-09
issue 1010 - Replacing Boost functions with cmath (Part 2 - error checking)

andrjohns:
## Summary

Addresses part 2 of issue #1010, replacing ```boost::math::``` functions that need input checking and platform-specific return handling

The following functions have been replaced:
- acosh
- atanh
- lgamma
- log1p
- tgamma

## Tests
Only major change is that ```lgamma(0)``` now expects an ```inf``` return rather than ```nan```

## Side Effects

N/A

## Checklist

- [x] Math issue #1010 

### DIFF
--- a/stan/math/prim/scal/fun/acosh.hpp
+++ b/stan/math/prim/scal/fun/acosh.hpp
@@ -1,11 +1,10 @@
 #ifndef STAN_MATH_PRIM_SCAL_FUN_ACOSH_HPP
 #define STAN_MATH_PRIM_SCAL_FUN_ACOSH_HPP
 
-#include <stan/math/prim/scal/fun/constants.hpp>
 #include <stan/math/prim/scal/fun/is_nan.hpp>
-#include <stan/math/prim/scal/meta/likely.hpp>
-#include <stan/math/prim/scal/fun/boost_policy.hpp>
-#include <boost/math/special_functions/acosh.hpp>
+#include <stan/math/prim/scal/fun/is_inf.hpp>
+#include <stan/math/prim/scal/err/check_greater_or_equal.hpp>
+#include <cmath>
 
 namespace stan {
 namespace math {
@@ -19,10 +18,16 @@ namespace math {
  * @throw std::domain_error If argument is less than 1.
  */
 inline double acosh(double x) {
-  if (unlikely(is_nan(x)))
+  if (is_nan(x)) {
     return x;
-  else
-    return boost::math::acosh(x, boost_policy_t());
+  } else {
+    check_greater_or_equal("acosh", "x", x, 1.0);
+    #ifdef _WIN32
+      if (is_inf(x))
+        return x;
+    #endif
+    return std::acosh(x);
+  }
 }
 
 /**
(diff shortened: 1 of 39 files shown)


stan/math/prim/scal/fun/acosh.hpp
```diff
@@ -32,7 +37,18 @@ inline double acosh(double x) {
  * @return Inverse hyperbolic cosine of the argument.
  * @throw std::domain_error If argument is less than 1.
  */
-inline double acosh(int x) { return acosh(static_cast<double>(x)); }
+inline double acosh(int x) {
+  if (is_nan(x)) {
+    return x;
+  } else {
+    check_greater_or_equal("acosh", "x", x, 1);
+#ifdef _WIN32
```
syclik: Do you remember what this was about? Why do we need this `ifdef`?

test/unit/math/prim/scal/fun/lgamma_test.cpp
```diff
@@ -11,5 +11,5 @@ TEST(MathFunctions, lgammaStanMathUsing) { using stan::math::lgamma; }
 TEST(MathFunctions, lgamma_nan) {
   double nan = std::numeric_limits<double>::quiet_NaN();
   EXPECT_PRED1(boost::math::isnan<double>, stan::math::lgamma(nan));
-  EXPECT_PRED1(boost::math::isnan<double>, stan::math::lgamma(0));
+  EXPECT_PRED1(boost::math::isinf<double>, stan::math::lgamma(0));
```
syclik: Was this the only change in behavior? I didn't see anything else.

stan/math/prim/scal/fun/acosh.hpp
andrjohns: Yep. acosh(inf) is expected to return inf, but under windows acosh(inf) returns nan. More background here: https://github.com/stan-dev/math/pull/1067#issuecomment-444145018

test/unit/math/prim/scal/fun/lgamma_test.cpp
andrjohns: Yes, lgamma(0) was the only change in behaviour, everything else remains the same

seantalts:
@syclik is this good to go? We could use it to fix #1063 which will let us start getting Math PRs through again while we don't have any EC2 nodes (spot pricing went way up).

seantalts:
I'm digging into why this didn't fix #1063 and just came across a comment there that `std::lgamma` isn't thread-safe? Is that right? I see a note that the POSIX implementation of `lgamma` isn't thread-safe, but I can't tell if the C++ `std::lgamma` is using the POSIX implementation of `lgamma` under the hood or not.

wds15:
uh... thread safety would be very good (if not required) to have...

seantalts:
Agreed. @bob-carpenter, I think it was you who originally wrote that `lgamma` wasn't thread safe - can you confirm that that's true of the one in this PR (`std::lgamma`)?  We may need to switch back to the boost one

bob-carpenter:
Nobody knows what that means.  At least nobody answered when I asked a few times.  I couldn't find any answers online of how POSIX relates to C++.

bob-carpenter:
See:  https://en.cppreference.com/w/cpp/numeric/math/lgamma

It says:

>  The POSIX version of lgamma is not thread-safe: each execution of the function stores the sign of the gamma function of arg in the static external variable signgam. Some implementations provide lgamma_r, which takes a pointer to user-provided storage for singgam as the second parameter, and is thread-safe.

bgoodri:
I don't think anywhere in Stan Math we are touching signgam even if it is
being written to by multiple threads. In other words, we are defining
lgamma to be the logarithm of the absolute value of the gamma function so
it does not matter what the sign of the gamma function might have been.

bob-carpenter:
Thanks!  And thanks for actually thinking through what's actually going on mathematically.  I was just staring gormlessly at the lgamma doc wondering how it was relevant.

Issue #1081, opened 2018-12-13
Upgrade to Boost 1.69.0


bgoodri:
## Description

Math should use Boost 1.69.0 because

https://github.com/boostorg/math/issues/134

fixes a build error with `integrate_1d` on Solaris.

## Example

Try to use `integrate_1d` on Solaris

## Expected Output

A value of the integral

## Additional Information

None

## Current Math Version
v2.18.0

andrjohns:
closed via #1082

PR #1082, opened 2018-12-13
Feature/boost 1.69.0

bgoodri:
## Summary

Upgrade Boost to 1.69.0, which was  released today. This includes

https://github.com/boostorg/math/issues/134

which fixes a build failure with `integrate_1d` on Solaris.

## Tests

No new ones

## Side Effects

## Checklist

- [X] Math issue #1081

- [X] Copyright holder: Trustees of Columbia University

- [X] the basic tests are passing; The math unit tests in develop go into an infinite compile loop for me but test-headers (now) passes

- [ ] the code is written in idiomatic C++ and changes are documented in the doxygen . This does not apply to lib/


wds15:
Can you have a look here:

https://github.com/stan-dev/math/blob/develop/stan/math/prim/arr/functor/integrate_ode_rk45.hpp#L14

there  is an ifdef for a specifc boost version... I have not really a clue which this ifdef is there... but it looks mood.

bgoodri:
Where do I find boostrap.log ?
```
./bootstrap.sh: 1: ./bootstrap.sh: ./tools/build/src/engine/build.sh: not found
Building Boost.Build engine with toolset ... 
Failed to build Boost.Build build engine
Consult 'bootstrap.log' for more details
make/libraries:85: recipe for target 'lib/boost_1.69.0/stage/lib/mpi.so' failed
```

wds15:
hmm... the problem is that you added everything under `tools` *except* the build directory. I am not sure what the intend here is.... so I added the build directory, but we can add all the others from what I can tell.

Also, you seem to include a lot of libraries from boost which we don't need (I mean the bits and pieces from libraries which need to be build). I used `bcp` at the time to extract only stuff which is needed for MPI.

There are still some failures with MPI at the moment though. I can look into these later.

syclik:
@bgoodri, it looks like the unit tests for `integrate_1d` is failing. It looks like the new version of Boost has different behavior then what was there before.

bgoodri:
Does anyone have a clue about this? Boost 1.69.0 added a line
https://github.com/boostorg/mpi/blob/develop/include/boost/mpi/collectives/scatterv.hpp#L32
that fails when we use `map_rect` with `mpirun` but passes if we use `map_rect` with threading. Also, it passes if I call `mpirun` via `gdb` which precludes debugging it. If I add print statements, I get stuff like
```
mpirun -np 2 test/unit/math/prim/mat/functor/map_rect_mpi_test --gtest_output="xml:test/unit/math/prim/mat/functor/map_rect_mpi_test.xml"
--------------------------------------------------------------------------
[[22136,1],1]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: Stanubuntu16

Another transport will be used instead, although this may result in
lower performance.

NOTE: You can disable this warning by setting the MCA parameter
btl_base_warn_component_unused to 0.
--------------------------------------------------------------------------
[==========] Running 3 tests from 2 test cases.
[----------] 2 tests from MpiJob
[ RUN      ] MpiJob.hard_work_dd
Here is sizes:
-892631376
Here is in_values:
0
map_rect_mpi_test: lib/boost_1.69.0/boost/mpi/collectives/scatterv.hpp:36: void boost::mpi::detail::scatterv_impl(const boost::mpi::communicator&, const T*, T*, int, const int*, const int*, int, mpl_::true_) [with T = double; mpl_::true_ = mpl_::bool_<true>]: Assertion `bool(sizes) == bool(in_values)' failed.
```
where the number printed for `sizes` changes from run to run. So, it looks like we have yet another instance of undefined behavior lurking somewhere, but I don't really understand the MPI code.

wds15:
Does your hello world MPI program work (which you find in the boost docs)? The MPI system is complaining here which is why I am suspicious at this.

bgoodri:
Yes (I think). The example from https://www.boost.org/doc/libs/1_69_0/doc/html/mpi/getting_started.html yields
```
bgoodri@Stanubuntu16:/opt/stan-dev/stan/lib/stan_math$ mpirun -np 2 ./mpi-test
--------------------------------------------------------------------------
[[20286,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: Stanubuntu16

Another transport will be used instead, although this may result in
lower performance.

NOTE: You can disable this warning by setting the MCA parameter
btl_base_warn_component_unused to 0.
--------------------------------------------------------------------------
Rank 0 OK!
Rank 1 OK!
[Stanubuntu16:12618] 1 more process has sent help message help-mpi-btl-base.txt / btl:no-nics
[Stanubuntu16:12618] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

wds15:
Ok...i can take a look this weekend...

wds15:
I think the new assertion can be considered a bug in boost. This change forces me to use the API in a way which is not as intended to be used. This should be solved now as I found a work-around. Let me know if things are still trouble-some.

bgoodri:
This is failing only because of 
```
Output written on _formulas.dvi (31 pages, 7608 bytes).
Transcript written on _formulas.log.
error: Problems running latex. Check your installation or look for typos in _formulas.tex and check _formulas.log!
```
which has nothing to do with this Boost PR. So, I think someone should merge this in order for it to be used on https://github.com/stan-dev/math/pull/1121/ .

wds15:
You can try to restart the testing and cross fingers that we get a green light...or save energy and just merge, yes.

rok-cesnovar:
The fix for this doxygen is already in #1140, just needs to pass the tests. make doxygen currently fails on develop.

bgoodri:
It is good that the LaTeX thing is getting fixed, but that should not be holding up this PR from being merged or other PRs that need an updated Boost.

wds15:
I merged the doxygen fix into develop and merged back develop into this PR. Lights should turn green now.

bgoodri:
This now appears to be failing because of
```
In file included from lib/boost_1.69.0/boost/random/detail/integer_log2.hpp:19:
lib/boost_1.69.0/boost/pending/integer_log2.hpp:7:1: warning: This header is deprecated. Use <boost/integer/integer_log2.hpp> instead. [-W#pragma-messages]
BOOST_HEADER_DEPRECATED("<boost/integer/integer_log2.hpp>");
```
and setting `-Werror`. That is coming from `boost/random/detail/` so I think we can ignore it and assume Boost will fix it before that header gets removed.

wds15:
While I think this can be merged as things will work, I guess the problem is that this error will stop any other Jenkins runs for all other PRs. So we need to work around this error message by ignoring it in the Jenkins runs somehow or we find some other solution here.

(We have super high standards in terms of not allowed warnings and all that...I had to dismiss some sundials stuff for these reasons as well)

wds15:
Finally good to merge once we have a review approving.

Issue #1040, opened 2018-09-25
Add missing _rng functions


VMatthijs:
## Description
In an ideal world, the Stan Math library would have both a random number generator (an _rng file) and a density (a _log, _lpmf or _lpdf) implemented for all its distributions. This will be particularly important if we want to (try to) automate generation of synthetic data from the model.

Currently, according to my search through the math library, all distributions which have an RNG also have a density.  Conversely, however, the following distributions which have a density are missing a random number generator (`_rng` form of distribution):

```
./stan/math/prim/mat/prob/bernoulli_logit_glm
./stan/math/prim/mat/prob/gaussian_dlm_obs
./stan/math/prim/mat/prob/lkj_cov
./stan/math/prim/mat/prob/matrix_normal_prec
./stan/math/prim/mat/prob/multi_gp_cholesky
./stan/math/prim/mat/prob/multi_gp
./stan/math/prim/mat/prob/multi_normal_prec
./stan/math/prim/mat/prob/neg_binomial_2_log_glm
./stan/math/prim/mat/prob/normal_id_glm
./stan/math/prim/mat/prob/poisson_log_glm
./stan/math/prim/scal/prob/binomial_logit
./stan/math/prim/scal/prob/normal_sufficient
./stan/math/prim/scal/prob/std_normal
./stan/math/prim/scal/prob/wiener
```

## Example
For example, a first implementation of

`./stan/math/prim/mat/prob/bernoulli_logit_glm_rng_hpp`
(which is still missing the templating to allow for arguments that are scalars)

could do something like

```
#ifndef STAN_MATH_PRIM_MAT_PROB_BERNOULLI_LOGIT_GLM_RNG_HPP
#define STAN_MATH_PRIM_MAT_PROB_BERNOULLI_LOGIT_GLM_RNG_HPP

#include <stan/math/prim/scal/prob/bernoulli_logit_rng.hpp>
#include <stan/math/prim/scal/err/check_consistent_sizes.hpp>
#include <stan/math/prim/scal/err/check_consistent_size.hpp>
#include <stan/math/prim/scal/err/check_bounded.hpp>
#include <stan/math/prim/scal/err/check_finite.hpp>

namespace stan {
namespace math {
/**
 * Returns a draw from the Generalized Linear Model (GLM)
 * with Bernoulli distribution and logit link function.
 *
 * This is a convenience wrapper around
 * <code>bernoulli_logit_rng(alpha + x * beta, rng)</code>.
 *
[...]

PR #1083, opened 2018-12-14
Feature/issue 1040 missing rng

peterwicksstringfield:
## Summary

Add RNGS for 4 of the distributions currently missing them.

Partially addresses #1040.

Add an overload to check_pos_semidefinite to let it take an Eigen::LDLT instead of just an Eigen::Matrix, to avoid recomputing the LDLT.

Extend the chi-squared goodness-of-fit helpers in the testing code to accept non-equiprobable bins.

## Tests

Add tests for the 4 new RNGS and the check_pos_semidefinite overload.

Run tests like this:
runTests test/unit/math/prim/mat/prob/multi_normal_prec_rng_test.cpp test/unit/math/prim/mat/prob/bernoulli_logit_glm_rng_test.cpp test/unit/math/prim/mat/prob/gaussian_dlm_obs_rng_test.cpp test/unit/math/prim/mat/prob/matrix_normal_prec_rng_test.cpp test/unit/math/prim/mat/err/check_pos_semidefinite_test.cpp

## Side Effects

None.

## Checklist

- [X] Math issue #1040 


stan/math/prim/mat/err/check_pos_semidefinite.hpp
```diff
@@ -42,13 +42,36 @@ inline void check_pos_semidefinite(
...
+ * @param name Variable name (for error messages)
+ * @param cholesky Eigen::LDLT to test, whose progenitor matrix must
+ * not have any NaN elements.
+ * @throw <code>std::domain_error</code> if the matrix is not positive
+ *   semi-definite.
+ */
+template <typename Derived>
+inline void check_pos_semidefinite(const char *function, const char *name,
+                                   const Eigen::LDLT<Derived> &cholesky) {
+  // From the Eigen::LDLT we cannot check for NaNs in the original
```
syclik: Could you move this documentation to the doxygen comment? To be concrete, I think that you can omit the fact that we don't check NaNs (I believe we've already checked that to get an `Eigen::LDLT` object), we can remove the symmetry comment also.

What should be included is when describing the variable `cholesky` it should mention that it's requiring that the size is greater than 0. (Is that true? Can we not construct an `Eigen::LDLT` with a 0x0 matrix?)

peterwicksstringfield: The Eigen docs don't specify. But

```
Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> y2(0, 0);
Eigen::LDLT<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> > y_ldlt2(y2);
```

is enough to cause a segfault on my machine.

We can't make an Eigen::LDLT out of a rectangular matrix, Eigen throws an exception. But Eigen::LDLT will happily consume a non-symmetric square matrix as if it were symmetric and only use half of it. We can even hide nans in the other half and not get errors. (There's a test in the PR demonstrating this.)

I guess just delete the whole comment. the LDLT object passed in needs to be valid and its fair to leave that assumption unstated.

stan/math/prim/mat/prob/bernoulli_logit_glm_rng.hpp
```diff
@@ -0,0 +1,82 @@
+#ifndef STAN_MATH_PRIM_MAT_PROB_BERNOULLI_LOGIT_GLM_RNG_HPP
+#define STAN_MATH_PRIM_MAT_PROB_BERNOULLI_LOGIT_GLM_RNG_HPP
+
+#include <Eigen/Dense>
```
syclik: Instead of this, we need to be including our own Eigen header which includes the traits in the right way.

```
#include <stan/math/prim/mat/fun/Eigen.hpp>
```

syclik:
@peterwicksstringfield, I'm going to do some cosmetic changes to the code. Thanks for submitting it.

PR #1086, opened 2018-12-17
Bugfix/issue 1075 map rect fail

wds15:
## Summary

Fixes an issue with `map_rect` whenever threading is used.

## Tests

There is a new test which triggers the issue:

`test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp`

If you run this test on commit `cb1f5b4` then it will fail as this has the current develop `map_rect_concurrent` with the failing behavior. The next commit then fixes the problem.

## Side Effects

Nope.

## Checklist

- [X] Math issue #1075 

### DIFF
--- a/stan/math/prim/mat/functor/map_rect_concurrent.hpp
+++ b/stan/math/prim/mat/functor/map_rect_concurrent.hpp
@@ -102,8 +102,7 @@ map_rect_concurrent(
 
 #ifdef STAN_THREADS
   if (num_threads > 1) {
-    const int num_big_threads
-        = (num_jobs - num_jobs_per_thread) % (num_threads - 1);
+    const int num_big_threads = num_jobs % num_threads;
     const int first_big_thread = num_threads - num_big_threads;
     for (int i = 1, job_start = num_jobs_per_thread, job_size = 0;
          i < num_threads; ++i, job_start += job_size) {
--- a/test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp
+++ b/test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp
@@ -1,0 +1,93 @@
+// the tests here check that map_rect_concurrent works correct as such we
+// enforce that STAN_MPI is NOT defined while STAN_THREADS is turned on
+
+#ifdef STAN_MPI
+#undef STAN_MPI
+#endif
+
+#ifndef STAN_THREADS
+#define STAN_THREADS
+#endif
+
+#include <stan/math/prim/mat.hpp>
+#include <gtest/gtest.h>
+
+#include <test/unit/math/prim/mat/functor/hard_work.hpp>
+
+#include <iostream>
+#include <vector>
+#include <string>
+#include <stdlib.h>
+
+// utility to set number of threads to use
+void set_n_threads(std::size_t num_threads) {
+  static char env_string[256];
+  std::string num_threads_str = std::to_string(num_threads);
+  snprintf(env_string, sizeof(env_string), "STAN_NUM_THREADS=%s",
+           num_threads_str.c_str());
+  putenv(env_string);
+}
+
+STAN_REGISTER_MAP_RECT(0, hard_work)
+STAN_REGISTER_MAP_RECT(1, hard_work)
+
+struct map_rect : public ::testing::Test {
+  Eigen::VectorXd shared_params_d;
+  std::vector<Eigen::VectorXd> job_params_d;
+  std::vector<std::vector<double> > x_r;
+  std::vector<std::vector<int> > x_i;
+  const int N = 7;
+
+  virtual void SetUp() {
+    shared_params_d.resize(2);
+    shared_params_d << 2, 0;
+
+    for (int n = 0; n != N; ++n) {
+      Eigen::VectorXd job_d(2);
+      job_d << 0, n * n;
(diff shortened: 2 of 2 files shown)


seantalts:
Sorry, the EC2 Jenkins node became unavailable because the spot request gave us an instance type with not enough disk space 😓 Should be going again soon. Thanks for the fix! Will code review now...

stan/math/prim/mat/functor/map_rect_concurrent.hpp
```diff
@@ -102,8 +102,7 @@ map_rect_concurrent(
 
 #ifdef STAN_THREADS
   if (num_threads > 1) {
-    const int num_big_threads
-        = (num_jobs - num_jobs_per_thread) % (num_threads - 1);
+    const int num_big_threads = num_jobs % num_threads;
```
seantalts: What are big threads?

test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp
```diff
@@ -0,0 +1,94 @@
+// the tests here check that map_rect_concurrent works correct as such we
+// enforce that STAN_MPI is NOT defined while STAN_THREADS is turned on
+
+#ifdef STAN_MPI
+#undef STAN_MPI
+#endif
+
+#ifndef STAN_THREADS
+#define STAN_THREADS
+#endif
```
seantalts: I'm a little unsure if we should have tests that change these compile flags like this versus changing them from the Jenkins file and running under various conditions? Like I think it might be better to just have these tests only execute if STAN_THREADS is defined, but maybe this is a question for @syclik.

stan/math/prim/mat/functor/map_rect_concurrent.hpp
wds15: In case the number of jobs and threads do not line up and we have more jobs than threads there will be a few threads with one extra job. That are the big threads.

wds15:
This function is already refactored in the other parallel stl pr. Things are much cleaner there...and the code runs >30% faster.

test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp
wds15: This test is intended to tickle the scheduling problems. The test does not use ad and as such it is only to get the scheduling code „active“ in order to test it. I think it is ok to always test it, but I see your point.

seantalts:
Sounds good re: refactoring! I will take a look at that one next if no one else has.

test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp
seantalts: It would be more like the GPU tests which are completely ifdef'd out if STAN_OPENCL isn't defined and only compile and run if it is defined. We have to go through all of these configurations anyway on Jenkins, seems like we might continue to just use that to trigger the proper tests...

I just realized - I think we do need to take these out and rely on external control of these make variables because otherwise if you are running all tests, whenever this one starts getting compiled it will change these flags for all subsequent tests, I think.

stan/math/prim/mat/functor/map_rect_concurrent.hpp
seantalts: Gotcha! Makes sense now.

test/unit/math/prim/mat/functor/map_rect_concurrent_threads_test.cpp
wds15: Are u sure about that change for others? I thought that each cpp test file is its own translational unit and becomes and independent exe file.

seantalts: Oh yeah, you're right, so it's just a style thing. I don't feel strongly but would think we would wanna do this test the way we've done most of the other tests that use compile flags that I'm aware of - #ifdef it out completely if STAN_THREADS isn't defined. But we can ask @syclik for a final judgement on that.

seantalts:
Would you mind updating this with the related stuff from the parallel STL PR and also having the tests cycle through different job size / num thread combinations?

seantalts:
> I can add a bit more cycling for varying the number of jobs (I think threads are already varied, but I can nest that in a loop for number of jobs).

They are, yep! Thanks, just figured we might as well be thorough given that we know the issues pop up with specific interactions there. And Daniel pointed out there was confusion between this and the other thread whether this had been tested with some specific job numbers that had failed the original fix, I think.

Which one of those two options regarding the parallel STL PR would you find easier? I think either would work (assuming the for_each stuff is left out of both versions and no new dependencies or compile flags are added, etc).

wds15:
My preference ist the first one, but that would bring in the for each code...without any extra dependency or any extra compile time flag though.

Writing the code with for each makes it a lot cleaner to me.

If you want to avoid that for now, then that’s fine...your call.

seantalts:
Oh, I see what you mean. Let's keep this PR tiny and hope to get out a new Stan version with fixed threading ASAP. Perhaps just add the extra loop over job sizes to this one since it seems the rest of the nice-to-haves in the other PR are linked up with for_each? Thanks for fixing!

wds15:
replaced by #1089

Issue #864, opened 2018-05-16
Add Matern Kernels to Gaussian Process Library


drezap:
#### Summary:
Add special cases Matern class of covariance functions to the gaussian process library.
These include:
Exponential (Matern 1/2) (`gp_exp_cov()`)
Matern 3/2 (`gp_matern32_cov()`)
Matern 5/2 (`gp_matern52_cov()`)
Exponential Quadratic (Matern lim v -> inf) (`gp_exp_quad_cov()`)

#### Description:
Each matern covariance function will have 4 parameterizations:

`gp_function_cov(x, sigma, length_scale)`
`gp_function_cov(x, sigma, length_scale, jitter)`
`gp_function_cov(x1, x2, sigma, length_scale)`
`gp_function_cov(x1, x2, sigma, length_scale, jitter)`

Where we x is the data vector, x1 and x2 are two different data vectors of type `real` (or possible ints), `sigma` is the marginal standard deviation, `length_scale` is the length scale, and `jitter` is a small real number that is added onto the diagonal for numerical stability during cholesky decomposition.

#### Inputs to support
The input datatypes that should be tested for `x`, `x1` and `x2`  are:
`std::vector<double>`, `std::vector<Eigen::Matrix<double, 1, -1>` (row vec w/ `doubles`), `std::vector<Eigen::Matrix<double, -1, 1>` (col vec w/ `double`), and varying x lengths.

The input datatypes that should be tested for `l` are:
`double`, `std::vector<double>` (the latter for ARD (automatic relevance determination)).

For `rev` versions, we also need to test all combinations of `var` and `double`, (and their related gradients).

I'm leaving in both Eigen's vectors and row vectors, because the same code generally supports both, so we (meaning mostly I) might as well throw some extra test cases in there.

For `sigma` we only need `double`.

#### Expected Output:

A data matrix that is a covariance function. If the input is only `x`, the output is symmetric real matrix that is positive semi-definite. Otherwise, output a matrix of reals.

#### Additional Information:
We want to support easy summation of kernels:

`gp_exp_quad(x, sigma, length_scale) + gp_exp_quad(x, sigma, length_scale, jitter)`

We like to use Rasmussen and WIlliams as a reference, however, in Chapter 4 they do not include the parameterization that has the marginal standard deviation, although this parameter is used later in the book in many places. We include this in our parameterization for these special cases of Matern kernels. 

#### Current Version:
v2.17.0

syclik:
[...]

PR #1096, opened 2018-12-29
add Matern32 prim kernel

drezap:
## Summary

Adds the matern32 kernel primitive

## Tests

Tests column vectors and error handling.

## Side Effects

No side effects.

## Checklist

- [x] Math issue #864 

- [x] Copyright holder: (Copyright 2018 Andre Zapico)


stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,278 @@
...
+  for (size_t n = 0; n < x_size; ++n)
+    check_not_nan("gp_matern32_cov", "x", x[n]);
+
+  check_positive_finite("gp_matern32_cov", "magnitude", sigma);
+  check_positive_finite("gp_matern32_cov", "length scale", length_scale);
+
+  for (size_t n = 0; n < x_size; ++n)
+    check_not_nan("gp_matern32_cov", "length scale", length_scale[n]);
+
+  check_size_match("gp_matern32_cov", "x dimension", x[0].size(),
```
syclik: This will fail in the same way as the other PR that was closed. If `x` is size 0, this will seg fault.

drezap:
coupled_ode_system unit tests fail, something to be worried about?

drezap:
@syclik

drezap:
Yo @syclik thanks for the reviews. Hope I’m not being too difficult. I appreciate you. As we all do.

Yeah, you’re right. I should be testing this better. so far, I’ve got: different datatypes for autodiff to work/templating, all of Stan’s different interfaces, obviously mathematical accuracy, trivial edge cases, invalid user input via the language... starting to accumulate these things thank you. I thought using the same test cases as before would be valid, but it’s not as good programmers make their code robust to bugs they don’t always test.

syclik:
@drezap, you're welcome! You're not... some of this stuff is hard to explain. Sorry, been sick for the last few days. I'm back on reviewing all this.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,279 @@
...
+ * where \f$d(x, x')\f$ is the Euclidean distance
+ *
+ * @tparam T_x1 type of scalars
+ * @tparam T_x2 type of scalars
+ * @tparam T_s type of element of sigma, marginal standard deviation
+ * @tparam T_l type of elements of length scale
+ *
+ * @param x1 std::vector of Eigen column vectors of scalars
+ * @param x2 std::vector of Eigen column vectors of scalars
+ *    This function assumes each column of x1 and x2 are the same size.
```
syclik: Please remove this doc. We don't assume that `x1` and `x2`'s columns have the same length.

drezap:
@syclik  yo - just wanna make sure I'm being clear. I'm going to explain this in a really naive way, but we're interested in different covariance functions because with different covariance functions, we can impose different properties on the process. They can be any number of things, smoothness, periodic properties, etc. Which is why the ultimate goal is to let users implement their own. Anyway - I'm implementing pretty much exactly as it is on the wikipedia page/GPML (with some exceptions to get it to work with C++ and stan's language). With that said, we have the first two for "in sample data"/training data, what ever you wanna call it. Second two with two input data for the out of sample data. in the multivariate gaussian mean function, this is f = K_{x*, x} (K_{x,x}^{-1}) y, for a posterior predictive with no noise (2.19 in GPML).

This really shouldn't need any maintenance moving forward, as the code is pretty f* simple, but if it does need maintenance, it's copy written to me, and I take this seriously, and if you ping me I'll come fix it

syclik:
Just a quick thing to make it easier to review: can you merge in develop? I feel like I'm looking at changes on GitHub that aren't relevant.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,285 @@
...
+    check_not_nan("gp_matern32_cov", "length scale", length_scale[n]);
+
+  check_size_match("gp_matern32_cov", "x1 dimension", x1[0].size(),
+                   "number of length scales", l_size);
+  check_size_match("gp_matern32_cov", "x2 dimension", x2[0].size(),
+                   "number of length scales", l_size);
+
+  T_s sigma_sq = square(sigma);
+  T_l root_3 = sqrt(3.0);
+  T_l neg_root_3 = -1.0 * sqrt(3.0);
```
syclik: For efficiency, can you declare `root_3` and `neg_root_3` as `double`? As written, this will promote these to type `T_l`.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
syclik: Did you see this comment? It still holds. Doc is incorrect here.

drezap:
@syclik yup sorry. pulled from develop branch and now you should only see the 3 files I've edited. Thanks

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,284 @@
...
+  size_t l_size = length_scale.size();
+  for (size_t n = 0; n < l_size; ++n)
+    check_not_nan("gp_matern32_cov", "length scale", length_scale[n]);
+
+  check_size_match("gp_matern32_cov", "x1 dimension", x1[0].size(),
+                   "number of length scales", l_size);
+  check_size_match("gp_matern32_cov", "x2 dimension", x2[0].size(),
+                   "number of length scales", l_size);
+
+  double sigma_sq = square(sigma);
```
syclik: This isn't correct any more. The tests shouldn't allow this to compile. The type of `sigma_sq` needs to be `T_s`. There are a handful of these problems in your latest merge. 

`sqrt(3.0)` should be `double` because we know that type exactly.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,284 @@
...
+ * to compute the posterior predictive density.
+ *
+ * @tparam T_x1 first type of scalars contained in vector x1
+ * @tparam T_x2 second type of scalars contained in vector x2
+ * @tparam T_s type of parameter sigma, marginal standard deviation
+ * @tparam T_l type of parameter length scale
+ *
+ * @param x1 std::vector of scalars that can be used in squared_distance
+ * @param x2 std::vector of scalars that can be used in squared_distance
+ *    This function assumes all elements of x1 and x2 are the same length.
```
syclik: Can we clarify this? I think we should assume all elements of x1 has the same size. Same with X2, but not them both. I'm thinking users from the Stan language can make that mistake. 

Can we add the checks from the other PR where we check that the size of the elements of x1 matches X2?

drezap: from the other PR, we check where the size matches x1 and x2 implicitly when we check to see that the number of of dimensions are same as 'P' (or D) for each x. I agree, and I'll go ahead and add checks for all, so the user doesn't make this mistake

drezap: Ok, so here I'm encountering issues, and I have reason to believe this is why the assumption was here in the first place. my compiler is not happy trying to get the `size()` of an element when it's a scalar (because T_x can either be a scalar type or vector type for the 1st and 3rd functions... so I have to do some tedious metaprogramming or something)

syclik: We have a template meta program for that! I don't know what it is off the top of my head, but it's there! Don't rewrite it unless we really don't have it.

syclik: We have a template meta program for that! I don't know what it is off the top of my head, but it's there! Don't rewrite it unless we really don't have it.

drezap:
@syclik So here's a summary of the changes in this commit: fixed all the templating and using `return_type` where appropriate. I now check _every element_ of x, or x1, x2 for the cross covariance calculation to see if it's the same size. I'm also using `stan::size_of` because this is much safer. I've also added relevant test cases.

For the ARD case, I'm only checking to see if the first element is the same size as number of length scales. If you also want me to parse through all elements of x to check the length I can do so, just let me know.

drezap:
I just realized that this size check is done implicitly in `squared_distance`, though. So it's not even that necessary to make sure they're the same size, but the errors could serve a pedagogical purpose.

syclik:
Thanks for adding these changes. It looks like some stuff got lost in translation. I'll have to review from a computer, but I'll try to provide a lot more specific guidance.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+  size_t x_size = size_of(x);
+  Eigen::Matrix<typename return_type<T_x, T_s, T_l>::type, Eigen::Dynamic,
+                Eigen::Dynamic>
+      cov(x_size, x_size);
+
+  if (x_size == 0)
+    return cov;
+
+  const char *function = "gp_matern32_cov";
+  for (size_t i = 0; i < x_size; ++i)
```
syclik: This logic doesn't seem right. It'll do the right thing, but it's really wasteful. `x_size * (x_size - 1) / 2` times, but you should only need to make `x_size` comparisons.

So, first off, it's perfectly acceptable to document this to say `@param x std::vector of scalars that can be used in square distance. It's assumed each element of x is of the same size`. Then we could just check the first element. If you don't want to do that, the logic here should be something like:
```
size_t x_element_size = size_of(x[0]);
for (size_t i = 1; i < x_size; ++i)
  check_size_match(function, "x row", size_of(x[i]), "x's other row", x_element_size);
```

Since we've already checked that `x_size > 0` (line 51), this is ok to do this way.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+  for (size_t n = 0; n < x_size; ++n)
+    check_not_nan(function, "x", x[n]);
+
+  check_positive_finite(function, "magnitude", sigma);
+  check_positive_finite(function, "length scale", length_scale);
+
+  T_s sigma_sq = square(sigma);
+  typename return_type<double, T_l>::type root_3_inv_l
+      = sqrt(3.0) / length_scale;
+  typename return_type<double, T_l>::type neg_root_3_inv_l
```
syclik: This type should be `T_l` also.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+      = -1.0 * sqrt(3.0) / length_scale;
+
+  for (size_t i = 0; i < x_size; ++i) {
+    cov(i, i) = sigma_sq;
+    for (size_t j = i + 1; j < x_size; ++j) {
+      typename return_type<T_x, T_s, T_l>::type distance
```
syclik: This type is incorrect. It should be `T_x`.

`return_type<T_x, T_s, T_l>::type` will be `stan::math::var` if any of the three are `stan::math::var`. You'll be getting overpromotion if `T_x` is `double`. Meaning if `T_x` is actually `double`, then the `sqrt(squared_distance(x[i], x[j)))` will also be `double`, but if `T_s` happens to be `stan::math::var`, we've now constructed an autodiff variable and placed it on the stack. So please change this to `T_x`.

syclik: [question]
do we not have a `distance` function?

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+  T_s sigma_sq = square(sigma);
+  double root_3 = sqrt(3.0);
+  double neg_root_3 = -1.0 * sqrt(3.0);
+
+  std::vector<Eigen::Matrix<typename return_type<T_x, T_l>::type, -1, 1>> x_new
+      = divide_columns(x, length_scale);
+
+  for (size_t i = 0; i < x_size; ++i) {
+    for (size_t j = i; j < x_size; ++j) {
+      typename return_type<T_x, T_s, T_l>::type distance
```
syclik: This type is incorrect. It should be `return_type<T_x, T_l>::type`.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+  size_t x2_size = size_of(x2);
+  Eigen::Matrix<typename return_type<T_x1, T_x2, T_s, T_l>::type,
+                Eigen::Dynamic, Eigen::Dynamic>
+      cov(x1_size, x2_size);
+
+  if (x1_size == 0 || x2_size == 0)
+    return cov;
+  const char *function = "gp_matern32_cov";
+
+  for (size_t i = 0; i < x1_size; ++i)
```
syclik: Similar comment to that above. This does the right thing, but is inefficient.

You can either document that you assume all of `x1`'s elements are the same size and the same for `x2`. But this has to be consistent with the other function. Or... you should do this with two loops, the first over `x1_size` and the second over `x2_size`.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+  for (size_t n = 0; n < x1_size; ++n)
+    check_not_nan(function, "x1", x1[n]);
+  for (size_t n = 0; n < x2_size; ++n)
+    check_not_nan(function, "x2", x2[n]);
+
+  check_positive_finite(function, "magnitude", sigma);
+  check_positive_finite(function, "length scale", length_scale);
+
+  T_s sigma_sq = square(sigma);
+  typename return_type<double, T_l>::type root_3_inv_l_sq
```
syclik: Change type to `T_l`.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+      = sqrt(3.0) / length_scale;
+  typename return_type<double, T_l>::type neg_root_3_inv_l_sq
```
syclik: Change type to `T_l`.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+      = -1.0 * sqrt(3.0) / length_scale;
+
+  for (size_t i = 0; i < x1_size; ++i) {
+    for (size_t j = 0; j < x2_size; ++j) {
+      typename return_type<T_x1, T_x2>::type distance
```
syclik: [comment]

This is the correct type. The result is `stan::math::var` if either `T_x1` or `T_x2` is `stan::math::var`, but `double` otherwise.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+    check_not_nan(function, "x2", x2[n]);
+
+  check_positive_finite(function, "magnitude", sigma);
+  check_positive_finite(function, "length scale", length_scale);
+
+  size_t l_size = length_scale.size();
+  for (size_t n = 0; n < l_size; ++n)
+    check_not_nan(function, "length scale", length_scale[n]);
+
+  check_size_match(function, "x1 dimension", size_of(x1[0]),
```
syclik: This function is inconsistent with the rest. The others make sure that all elements of `x1` are the same length and do not assume they are. Here, this only checks `x1[0]` and not `x1[1], ..., x1[x1_size-1]`.

stan/math/prim/mat/fun/gp_matern32_cov.hpp
```diff
@@ -0,0 +1,304 @@
...
+  double neg_root_3 = -1.0 * sqrt(3.0);
+
+  std::vector<Eigen::Matrix<typename return_type<T_x1, T_l>::type, -1, 1>>
+      x1_new = divide_columns(x1, length_scale);
+  std::vector<Eigen::Matrix<typename return_type<T_x2, T_l>::type, -1, 1>>
+      x2_new = divide_columns(x2, length_scale);
+
+  for (size_t i = 0; i < x1_size; ++i) {
+    for (size_t j = 0; j < x2_size; ++j) {
+      typename return_type<T_x1, T_x2, T_s, T_l>::type distance
```
syclik: Incorrect type. This should be `return_type<T_x1, T_x2, T_l>::type`.

syclik:
> I just realized that this size check is done implicitly in `squared_distance`, though. So it's not even that necessary to make sure they're the same size, but the errors could serve a pedagogical purpose.

It really does help to provide good error messages. Can you imagine if we just passed error messages through like that? Calling `y ~ normal(mu, sigma);` could end up with an error message that says the problem is in `divide` and it'll really be difficult to debug.

One thing you could do... wrap it in a try-catch block and throw an appropriate error if it fails. We haven't used that pattern in our code base, but I don't see why it couldn't be done that way.

drezap:
If people are confused I can see that being a good idea. I'm opting out of using the word "dimension" because it's ambiguous, and I'm sticking saying "number of elements in a row" (which means number covariates, columns, etc)

drezap:
Yo @syclik - can we keep this moving? I’ll still keep the flatten going, but I like to work on a 1-3 different things at once, and I’d like to get this in. There’s 5-6 more of these, but I think this is getting close to being ready for user- use

syclik:
@drezap: yup. Just FYI, the reviewer doesn't get notified when the code changes. So ping me whenever you get to that point.

Issue #1008, opened 2018-08-27
stan_sundials_printf_override.hpp is not that helpful


bgoodri:
## Description

The stan_sundials_printf_override.hpp file is intended to make sundials print output compatible with the interface, but it does not actually do that. First, it uses variadic macros, which are legal under C++11 but sundials seems not to compile with a C++ compiler anymore. You can compile sundials with a C compiler, but variadic macros are not part of the C standard used by `gcc-4.9` on Windows. There is a standard approach, which is to use `__VA_ARGS__`, but stan_sundials_printf_override.hpp just spits stuff onto stdout, which is what we are trying to avoid. To make this all work with R, I have to change stan_sundials_printf_override.hpp to
```
#ifndef STAN_MATH_SUNDIALS_PRINTF_OVERRIDE_HPP
#define STAN_MATH_SUNDIALS_PRINTF_OVERRIDE_HPP

#include <R_ext/Print.h>

#define STAN_SUNDIALS_PRINTF(...)   Rprintf(__VA_ARGS__)
#define STAN_SUNDIALS_FPRINTF(...)

#endif
```

## Example

Replace stan_sundials_printf_override.hpp in StanHeaders with stan_sundials_printf_override.hpp from Math and try to install the StanHeaders package on Windows with `gcc-4.9`. You get a compiler error.

## Expected Output

None

## Additional Information

None

## Current Math Version
v2.18.0

rok-cesnovar:
Has this issue been observed since #1098 was merged. If not, please close this issue.

mcol:
I believe this can be closed, as StanHeaders 2.19.0 includes exactly the same `stan_sundials_printf_override.hpp` file as was introduced by #1098, so no other workaround is necessary.

@bgoodri please reopen if my understanding is incorrect.

Issue #1097, opened 2019-01-05
Upgrade Sundials to version 4.0.1


wds15:
## Description
Sundials is a full major version ahead and we should upgrade.

## Example
No change for Stan.

## Expected Output
Nothing should change for Stan.

#### Current Version:
v2.18.1

PR #1098, opened 2019-01-05
upgrade sundials to 4.1.0

wds15:
## Summary

This upgrades Sundials to the new major version 4.1.0.

The changes to the source base of stan-math are:

- Change of initialization for CVODES solver. `CVodeCreate` does not have anymore a second argument for the solver which is now by default the newton solver.

- The former way of dealing with the `printf` & `fprintf` from Sundials was flawed and is replaced with a new scheme as suggested by @bgoodri , see the issue #1008 . This PR follows the recommendations given in the issue #1008, but there is no guarantee that #1008 is fixed due to troubles in reproducing the issue.

- To ease upgrading to future Sundials packages this PR includes a script `lib/upgrade-sundials.sh` which automates all steps needed to prepare the official Sundials package & add things to git.

## Tests

All existing tests (should) still run just fine. In particular the ODE & IDAS tests:

```
test/unit/math/rev/mat/functor/integrate_ode_adams_prim_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_adams_rev_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_bdf_rev_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_cvodes_grad_rev_test.cpp
test/unit/math/rev/mat/functor/integrate_ode_bdf_prim_test.cpp
test/unit/math/rev/mat/functor/idas_integrator_test.cpp
test/unit/math/rev/mat/functor/idas_system_test.cpp
```

## Side Effects

- The Jenkinsfile had to change. We are using `-Werror` to run `make test-headers`. There's a warning in CVODES that prevents us from getting past this -- `make test-headers` was meant to only test our code. I applied a fix, which is to add a new make target called `sundials`, build that in the Jenkins test with our normal compiler settings, then calling `make test-headers` to use `-Werror` to test the header files.

## Checklist

- [X] Math issue #1097 

- [X] Copyright holder: Sebastian Weber, Generable


wds15:
This is strange: I am getting unresolved symbol errors for sundials things in tests which do not involve at all any sundials stuff. If someone has an idea where this is coming from - that would be great.

wds15:
I think a number of PRs right now hang as the Windows test machine seems to be down. Can someone (@bgoodri , @seantalts ) maybe kick the respective machine? Thanks.

seantalts:
Yeah, unfortunately it's completely offline right now and located in NYC. @bgoodri said he should hopefully be able to physically kick it soon. Sorry about that.

bob-carpenter:
> On Jan 5, 2019, at 10:02 AM, wds15 <notifications@github.com> wrote:
> 
> This is strange: I am getting unresolved symbol errors for sundials things in tests which do not involve at all any sundials stuff. If someone has an idea where this is coming from - that would be great.

Sounds like a makefile issue, where it's trying to build a sundials translation unit.  This may just be from an include of sundials from the higher-level lib include.

wds15:
@bob-carpenter : I think to have found the issue. With Sundials 4.0 the interface to the non-linear solvers was changed. The compatibility wrappers which make 4.0 source compatible with pre-4.0 codes was written in a way such that it does not only declare functions, but defines them instantly in the header files. This causes the problems here. Turning to the new 4.0 interface solves the issue, I think.

@seantalts  No worries... I was hoping you have some remote access to the window machines.

@bgoodri Please kick those Windows boxes hard! It's Windows after all.

syclik:
@wds15, how can we confirm if this fixes #1008? If it doesn't, can we be clear that it doesn't?

wds15:
I am fairly sure it fixes it and was hoping that @bgoodri could confirm by looking and/or testing it. My windows machine I am sitting on is broken right now. The way to test it is to download StanHeaders 2.18.0 revert the changes @bgoodri did to the respective header file (`include/stan_sundials_printf_override.hpp`) to confirm that things break under Windows with the gcc 4.9. Then you can change that file to be like the one I have created in this PR. This should verify that things work.

Like I say I don't have a windows available which is not broken - maybe a VirtualBox at home will do, but if you have a Windows available then go ahead and test.

syclik:
@wds15: ok. We'll wait for @bgoodri to respond.

wds15:
Ok, i got my windows toolchain working again. However, I am not able to trigger the issue @bgoodri has identified. Thus before this PR gets stalled, I suggest that we take down the claim from this PR to fix the issue @bgoodri found. I think #1008 will be fixed with this, but I cannot verify. I am modifying the PR text and I hope you agree, @syclik, that we can proceed and merge.

bgoodri:
You can merge this PR. If necessary, I will just keep fixing it.

syclik:
@bgoodri, what do you have to do to fix it? Is there a general way to fix
this?

bgoodri:
I just change that file to the one on
https://github.com/stan-dev/math/issues/1008

wds15:
I think that @bgoodri won't need to change the file again as his instructions are taken care of wrt to the `__VA_ARGS__` macro use... as a last resort `rstan` can revert back to the file currently in use with `rstan` (which we cannot use in `stan-math` as it pulls in R specifics). Sounds like we are good now or anything else, @syclik ?

wds15:
Bump. Would be great if we can merge this or define what requirements are needed to get this in.

I haven't changed anything other than aligning with the new sundials 4.0.1 interface to their solvers. No change in behavior should be implied by this. Note that we are at sundials 3.1 and in the meantime there was 3.2 and now there is 4.0.1.

@syclik , @bob-carpenter , @seantalts ... thoughts?

syclik:
I spoke with @wds15.

I need to follow up with pystan developers to find out if this will build once merged.

syclik:
@ariddell @ahartikainen: how is Sundials packaged for PyStan now? If we bump the version, does this pose a problem for PyStan?

riddell-stan:
No problem.

syclik:
Thanks!

make/libraries
```diff
@@ -15,20 +15,24 @@ CPPLINT ?= $(MATH)lib/cpplint_4.45
...
 
 SUNDIALS_CVODES := $(patsubst %.c,%.o,\
-  $(addprefix $(SUNDIALS)/src/cvodes/, cvodes.c cvodes_io.c cvodea.c cvodea_io.c cvodes_direct.c cvodes_diag.c cvodes_spils.c cvodes_bandpre.c cvodes_bbdpre.c) \
+  $(wildcard $(SUNDIALS)/src/cvodes/*.c) \
   $(wildcard $(SUNDIALS)/src/sundials/*.c) \
-  $(wildcard $(SUNDIALS)/src/sunmat_band/*.c) \
-  $(wildcard $(SUNDIALS)/src/sunmat_dense/*.c) \
-  $(wildcard $(SUNDIALS)/src/sunlinsol_band/*.c) \
-  $(wildcard $(SUNDIALS)/src/sunlinsol_dense/*.c))
+  $(wildcard $(SUNDIALS)/src/sunmat_band/[^f]*.c) \
```
syclik: Question: What does the `[^f]` do?

I didn't expect these changes and I haven't been able to find a reference that explains what that is. Also, should line 18 also have it?

syclik:
@wds15, I'm looking at it. So far, so good. Should we just go to v4.0.2 since we're upgrading?

lib/prepare-sundials.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
syclik: @wds15, could you document this script at a high-level? What would be extremely helpful:

1. Why / when a maintainer would use this.
2. What the maintainer needs to set up before using this.
3. What variables the maintainer needs to set in the script.
4. What the expected output is and how the maintainer is supposed to know this worked correctly.

Finally, we need to put this in the wiki. Probably here: [Updating Libraries](https://github.com/stan-dev/math/wiki/Updating-Libraries)

lib/prepare-sundials.sh
```diff
@@ -0,0 +1,21 @@
...
+tar xvzf $SUNDIALS_TAR
+mv sundials-${SUNDIALS_VERSION} sundials_${SUNDIALS_VERSION}
+
+cd sundials_${SUNDIALS_VERSION}
+
+rm -rf INSTALL_GUIDE.pdf config/ doc/ examples/ test/ */cvode */ida */arkode */kinsol
+find . -name CMakeLists.txt -exec rm {} \;
+
+find src -name "*.c" -type f -exec sed -E -i _orig  's#[^sf]printf\(#STAN_SUNDIALS_PRINTF(#'g {} \;
```
syclik: I would prefer if we would actually commit to git the source directly from Sundials at this point prior to changing things. Even if it's just one commit followed by another, I think it'll help us back out changes if we mess up the process. What do you think?

make/libraries
wds15: The f* files are used by sundials to provide fortran compatible interfaces somehow. We don't need those - it would be a waste of compile time. Line 18 & 19 don't need this as there are no files starting with f.

syclik: Awesome. I think it warrants writing down. Especially since we're not in charge of Sundials and this convention isn't anything they guarantee.

lib/prepare-sundials.sh
syclik: What's `sundials-config`?

lib/prepare-sundials.sh
syclik: > Hmm... I would not do that as it complicates things. Moreover, git is super clever with these version changes and only tracks actual changes. Having those vanilla sundials -> adapted sundials commits in between bloats this.

I have the opposite reaction when I know this. 

I'm looking for easier review and maintenance. The overall result will be the same. 

If we have one commit that's the pruned Sundials and the next that's the changes to Sundials, I find using the GitHub interface it's easy to tell that the second commit actually does what we intend. And if we make a mistake, it's easy to back out that one commit. If the two actions aren't separated in git, then we can't tell and it's more work in verifying that this is correct. 

In fact, this would have helped with this current PR. Try thinking about the person reviewing. If I told you that I've updated a library and changed things, how would you verify it? My first thought was to download and do a diff, but I can't compare that to the final. If there's a better way to validate the PR, let me know. If not, let's add the additional git commit, which, as you say, git is clever and won't actually impact the user (but will positively impact the reviewer).

lib/prepare-sundials.sh
wds15: sundials-config contains all stan specific configuration items for Sundials. I introduced it in my last commit. It serves to separate out things.

wds15:
I really do not think we need two git commits here. The old state was not ideal as teasing out the differences between vanilla sundials and our changes were not fully transparent - you are right in that and this is now solved.

The PR as it is right now has achieved a very clear separation between what is vanilla sundials and what is our Stan specifics:

- sundials_4.0.1 contains now the vanilla sundials distribution and the only changes applied to it are all contained in the `prepare.sh` script. There is no manual intervention anymore needed. Thus it is transparent what is vanilla sundials and what not.

- sundials-config contains all Stan specifics for our needs.

I am just not good enough with my git fu to script what you are asking for - and I do think the goals you are looking for are fully achieved which is transparency and low maintenance burden by semi-automation.

I will add

wds15:
It would be great if we could merge this PR either by

- someone helping out with more git fu knowledge (@seantalts ?) to do what @syclik asks for
- or we merge what I have done

I have invested a significant amount of work into maintenance of the CVODES system and things are in far better shape with this PR than without it in my opinion - but if things should be done differently I can take down this PR.

ahartikainen:
What do you need to do?

For merging/rebasing I do

    git log --oneline
    # select hash to rebase (atleast one before the ones you edit
    git rebase -i hash385643d
    # then edit the text file and save and exit
    git push -f

wds15:
Not quite... I have proposed a rather automatic way of applying all changes which we need to the sundials source as described above (see my comments). This is done in a single go in a single script and hence all changes are contained in that script. I think (not sure) that @syclik looks for documenting changes to sundials using git history. Automating that is beyond me and I probably have a different view on this. The overall goal to have it transparent what is done to the sundials source is achieved with PR as it is.

However, if you, @ahartikainen, would like to settle with @syclik on the last details of how he wants it, then that would help. Thanks.

ahartikainen:
So you want to do the changes at compile time or as a hook in github (change the source)?

wds15:
Not quite... maybe you will be our guinea pig with the instruction I put up in this PR in the README.md here: 

https://github.com/stan-dev/math/tree/2dbd4466163322ae02c85d22d3bbfd3f5cb5a276/lib/sundials-config

and update the sundials from 4.0.1 to 4.0.2?

This PR intends to update to version 4 sundials and to make things easier to maintain in the future by documenting those steps to do the update.

All needed changes to the sundials distribution are done once by executing the prepare.sh script which is in the above folder I link.

ahartikainen:
I think the instructions are clear. (I will test this later when I'm on linux).

Why there is a comment about git for filechanges (`git add .`)? Is the idea that after update the files are uploaded back to github?

syclik:
Sorry, I've been sick. The git fu isn't bad. I'll add notes on it later.

But what's bothering me now is that we're scattering things in the math library. It's not really how I think we should just leave things and merge. Where did `stan_sundials_printf_override.hpp` come from? Why did it need to move? What's the motivation here?

wds15:
Oh, hope you are better now.

- `stan_sundials_printf_override.hpp` is around for a while. I think @yizhang-cae introduced it with the 3.1 sundials inclusion.
- the file needed to move because I wanted to keep the sundials 4.0.1 clean modulo the automatic changes from the script.

Everything is automated to a very high degree right now and things are transparent in terms of what we modify from the sundials source distribution.

lib/sundials-config/prepare.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
syclik: Please add documentation to this script. At the very least, let the reader know the doc is in a markdown, but I'd much rather there be documentation here also.

Here's how I think you can make this much more useful (I'll put in inline comments):
1. Start by `git rm` the existing sundials library.
2. `git commit` after line 13.
3. `git commit` after line 18.

lib/sundials-config/prepare.sh
```diff
@@ -0,0 +1,20 @@
...
+
+SUNDIALS_VERSION="4.0.1"
+
+SUNDIALS_TAR=sundials-${SUNDIALS_VERSION}.tar.gz
+
```
syclik: Here, you could do something like `git rm lib/sundials-*` to mark all of the files that we need to delete.

lib/sundials-config/prepare.sh
```diff
@@ -0,0 +1,20 @@
...
+tar xvzf $SUNDIALS_TAR
+mv sundials-${SUNDIALS_VERSION} sundials_${SUNDIALS_VERSION}
+
+cd sundials_${SUNDIALS_VERSION}
+
+rm -rf INSTALL_GUIDE.pdf config/ doc/ examples/ test/ */cvode */ida */arkode */kinsol
+find . -name CMakeLists.txt -exec rm {} \;
+
```
syclik: At this point, we should `git add sundials_${SUNDIALS_VERSION}`. This is the clean, direct addition of all the Sundials code.

Follow that by adding something like:
```
git commit -m "Upgrading Sundials to version ${SUNDIALS_VERSION}"
```

lib/sundials-config/prepare.sh
```diff
@@ -0,0 +1,20 @@
...
+find src -name "*.c" -type f -exec sed -E -i _orig  's#[^sf]printf\(#STAN_SUNDIALS_PRINTF(#'g {} \;
+find src -name "*.c" -type f -exec sed -E -i _orig  's#fprintf\(#STAN_SUNDIALS_FPRINTF(#'g {} \;
+
```
syclik: remove this whitespace -- these 3 lines go together. Adding the single line makes it seem like the next line is different.

lib/sundials-config/prepare.sh
```diff
@@ -0,0 +1,20 @@
...
+find src -name "*.c_orig" -exec rm {} \;
+
```
syclik: At this point, we want to do something like:
```
git add lib/sundials_${SUNDIALS_VERSION}
git commit -m "replacing printf and fprintf for rstan to be accepted by CRAN"
```

wds15:
To me, the `sundials-config` solution makes perfect sense to me.

Where did you plan to stick the config file `math/lib/sundials-config/include/sundials/sundials_config.h`. That file gets generated from a call to cmake and is based on the respective sundials_config.in file. It almost never changes, but is clearly not part of sundials vanilla.

If you have time to work on this - that would be great.

seantalts:
@syclik I really like that you have clearly defined goals for how to get this in, thank you for writing those up. As a third party I wonder if we could have a compromise - it does seem like there are a couple of files that we might want to put in the right places with a script, and ideally we might actually want multiple scripts here for importing these libraries (well, the alternative might actually be better - just using git subtree, leaving all of their build system et al intact, and porting our changes through versions... but for now, we might want a few scripts). What about putting these in a scripts directory somewhere, either in `lib` or up a level if that messes up the hierarchy? Or even just allowing for multiple loose files in the `lib` directory if we can be trusted to know those aren't libraries.

syclik:
@seantalts, thanks for weighing in. I realize that you're seeing this at this point in time without having seen the original PR. I feel like this is a compromise and perhaps what I should be doing is requesting @wds15 to follow the process more carefully. I'll give some more info below. @wds15... I should be communicating more of this about all PRs more frequently. Hopefully you can start seeing PRs from the reviewer's point of view.

The original issue was to upgrade Sundials. Great; no objections to the issue itself. The PR did that, but in addition, left behind a single helper script without any doc and in a place where there are no additional scripts. The original PR indicated that this was being included, but it's not really cool to just leave behind a script that's undocumented, no information other than inspection, no usage information, and no forethought into the location of the script. The script was put in without making the community maintenance of the Math library a priority (which needs to be front and center on a PR like upgrading libraries). Also, tacking on additional things for a reviewer to figure out that's not part of the original issue isn't really cool when it actually involves a decision.

After evaluating the original PR, I asked for what I thought was reasonable review (this is the compromise) where I requested something that wouldn’t spawn an actual discussion: please document the one script, leave it where it is, make a couple of changes to the script to make it easier for the reviewer of an upgrade to validate that the changes to the library are sound, and let’s move on.

The changes after review were much deeper than requested. Some of the things I asked for were done, but now the PR is spawning a new folder, there are multiple files within that folder (outside of any main branch), the script is still not documented, and there are now new side effects of having the makefiles change in different ways. (And... it seems like I'm finding out a new file that's checked in that's generated in a different way that I didn't know about?)

If we're going to go with this type of change, I'm right with @seantalts in thinking that "putting these in a scripts directory somewhere, either in lib or up a level if that messes up the hierarchy." But the key is that this is a different issue. We should discuss and decide that issue as a separate topic instead of trying to sneak it in a PR that's supposed to upgrade a library. (PRs should really be narrow in scope.)

@wds15: the thing about adding additional, undiscussed fixed / functionality / changes is that it puts more work on the reviewer. If I could look at the issues that you highlight, see what the community discussed, and can just verify and validate that your PR fixes those things, reviewing is pretty straightforward. When a reviewer is asked to weigh in on decisions that aren't discussed, well... honestly, PRs should probably be closed and kicked back to a design discussion at that point. I've been encouraged to do so, but I still feel bad just closing PRs for skipping this step. Would it benefit you if I did that? (See a PR, realize that there's a non-trivial discussion that should be had, close the PR and request a design [essentially lay out the options, which one you want to take, and why]?)

seantalts:
I don't think your requests were in any way unreasonable, by the way! Just trying to help find a good solution given the extra work @wds15 has done to help make the process of upgrading sundials easier in the future (even though the script doesn't have doc yet, it's probably overall still an improvement in maintainability because it helps the next person figure out how to upgrade sundials). 

I think you're right to ask that people highlight extra stuff that doesn't follow an existing pattern. Though it's not always obvious to the submitter what is and isn't going to be a controversial decision - not sure if we can help much with that as it's a sort of intuitive guess at what will surprise the reviewer. But we could certainly ask people to err a little more one way or the other (e.g. seems like it would be good if @wds15 adjusted his criteria for this stuff to be more conservative and surface more things as problems/solutions that might require additional discussion). 

I'm not sure if you're asking that this decision have a formal design review in the way that the large projects in [the design docs repo](http://github.com/stan-dev/design-docs/) have had, but certainly calling it out as an open question in the PR and maybe starting a discourse discussion would be good. Is that basically all you're asking for? That makes total sense to me.

wds15:
@syclik My view is that I am doing a service to the community in bumping to a  new major version. That is a lot of work and it was rough to do it for various reasomns. I ran into a lot of trouble and decided for the next one to be save from the trouble by semi-automating the upgrading process. Sure - that adds to the PR, but for the good, I think. This is an investment to the community and this upgrading in a semi-automatic fashion is something new here - so it is totally fine from my perspective that the reviewer helps in this.

Maybe I should have first started a discussion on discourse, but I decided in the spirit to accelerate turn-over time of PRs which I recall from @bob-carpenter is a goal for stan-math.

To move on here I see the options:

- revert this PR to an earlier stage, remove the script and the readme which we can file on the wiki.
- find a place where these auxiliary files can go

Wrt. to messing with git history I am still really not in favor for this as I don't quite see the use of it. It doesn't make the life of the reviewer any easier at all - there are just so many changes for printf commands such that it is a lot easier to document this change by having the script around. The sed command is just fine to me and a better way to deal with it. In addition I do not feel comfortable scripting git. I barely arrive at managing it. I don't object against this change if you have a different view - but if we go with this then I would really ask for a hand here.

The other thing, which is missing in what you laid out above are the config header files (or did I overlook that bit). These are usually never changed, but they are automatically generated during the sundials config step which is in most cases not needed to be rerun.

So I am happy to move forward whatever way.

syclik:
@wds15, how exactly do you generate `sundials_config.h`? It wasn't in any of the doc.

syclik:
or... @yizhang-cae, know off the top of your head?

syclik:
Let me know if this is right...
```
tar xvzf sundials-4.0.1.tar.gz 
cd sundials-4.0.1
mkdir build
cd build
cmake ..
```

and I need to grab `sundials-4.0.1/build/include/sundials/sundials_config.h` and `sundials-4.0.1/build/include/sundials/sundials_fconfig.h`

syclik:
@wds15, @seantalts, please review. I bumped the version to v4.0.2 as a demonstration of how it works.

There is now only one file that sits alone and it's a documented bash script: `lib/upgrade-sundials.sh`. This automatically takes care of everything for the installation including putting the config files in the right place, updating the `make/libraries` and README files, generating the header for `fprintf` and `printf`.

@wds15: please look at this commit. This allows us to quickly check to make sure that the replacement doesn't ruin everything: [upgrading to sundials v4.0.2; removing printf and fprintf for CRAN](https://github.com/stan-dev/math/pull/1098/commits/afdc740bc2885452ebe1ef105eeddb21aca04071)

And notice that each step is its own git commit. This makes it much, much, much easier for a reviewer to verify. If someone wanted to make sure that I used the correct version, they could check out ` [4367bb5](/stan-dev/math/pull/1098/commits/4367bb55beef9f489ddff8391fb04c79f123c214)`, then untar the downloaded file and move that into the folder and have git tell them if any files were different.

seantalts:
Do you mind if I wait for tests to pass? If it's better to take a look before that I can look at it this evening!

syclik:
Absolutely. There's no rush.

If you actually wanted to give it a test run, upgrade to 4.1.0.

wds15:
@syclik Thanks much for this! Let's cross fingers the fully automatic scripting will work for a while.

The current error with spfmg looks familiar to me. Since we do not use this functionality I did remove this file from our makefiles as I recall. Everything should just work if this file is left out. Do you want me to take care of this and @seantalts does the final review?

syclik:
@wds15: yes, if you don’t mind. I might have messed up what you did there. I’ll try to see if I can spot it, but if you know what’s wrong, that’s great. 

If you do remove it, we can revert the changes to Jenkinsfile so we’re not building it with different compiler settings.

syclik:
@wds15, did we avoid this in the past due to this change to make/libraries?

from develop:
```
  $(addprefix $(SUNDIALS)/src/cvodes/, cvodes.c cvodes_io.c cvodea.c cvodea_io.c cvodes_direct.c cvodes_diag.c cvodes_spils.c cvodes_bandpre.c cvodes_bbdpre.c)
```

from this PR:
```
 $(wildcard $(SUNDIALS)/src/cvodes/*.c)
```

Should we just revert this change?

wds15:
I think I did this one purpose, because some files were missing otherwise.

wds15:
Assuming the testing suite gives a green light this is good to go, I hope.

wds15:
Hmm... we now have an upstream CmdStan test failing with

```
mpicxx.openmpi -std=c++1y  -Wno-sign-compare    -Wno-delete-non-virtual-dtor -O3 -I src -I stan/src -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.3 -I stan/lib/stan_math/lib/boost_1.66.0 -I stan/lib/stan_math/lib/sundials_4.1.0/include -I stan/lib/stan_math/lib/gtest_1.8.1/include -I stan/lib/stan_math/lib/gtest_1.8.1     -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION   -DSTAN_MPI -DGTEST_USE_OWN_TR1_TUPLE       -Wl,-L,"/home/jenkins-slave/jenkins-slave-files/workspace/CmdStan_downstream_tests/stan/lib/stan_math/lib/boost_1.66.0/stage/lib" -Wl,-rpath,"/home/jenkins-slave/jenkins-slave-files/workspace/CmdStan_downstream_tests/stan/lib/stan_math/lib/boost_1.66.0/stage/lib"   -include src/test/test-models/print_uninitialized.hpp src/cmdstan/main.cpp        stan/lib/stan_math/lib/sundials_4.1.0/lib/libsundials_nvecserial.a stan/lib/stan_math/lib/sundials_4.1.0/lib/libsundials_cvodes.a stan/lib/stan_math/lib/sundials_4.1.0/lib/libsundials_idas.a stan/lib/stan_math/lib/boost_1.66.0/stage/lib/libboost_serialization.so stan/lib/stan_math/lib/boost_1.66.0/stage/lib/libboost_mpi.so stan/lib/stan_math/stan/math/prim/arr/functor/mpi_cluster_inst.o -o src/test/test-models/print_uninitialized
stan/lib/stan_math/lib/sundials_4.1.0/lib/libsundials_cvodes.a: file not recognized: File format not recognized
collect2.exe: error: ld returned 1 exit status
make: *** [make/program:37: src/test/test-models/printer.exe] Error 1
```

I am not sure why this is popping up now nor where this is coming from. I will need to look further into this.

wds15:
Ok... when restarting only the downstream job the pipeline succeeds. Thus, I am kicking Jenkins to do the full thing once more.

wds15:
Finally! Green... this is hopefully good to go in. @seantalts ... do you need anything to review this?

seantalts:
Which part(s) am I to review? Any tips for looking at just the stuff we changed vs. the vendored unchanged sundials library?

wds15:
I think @syclik was fine with this PR except the upgrade script. So maybe you try that out (grab the script and run it over a fresh branch off from develop)? @syclik did invest quite some work into this script, I have to say; all bits are automatic (unless sundials moves things around which makes our makefiles fail over).

The changes to stan-math sources are reviewed and the makefile changes were also waived, I think. So it's down to the script.

As a last sanity check you should execute the bdf+adams integrate ode tests.

syclik:
@seantalts, I can approve it. I think it's clear that both @wds15 and I have looked carefully at the script and merges and are confident that it's readable.

If you have time and want to look carefully, this is the file to check: lib/upgrade-sundials.sh. https://github.com/stan-dev/math/pull/1098/files#diff-1ca3f358ea4ffad65560c3bf9575d698

I just reverted the changes to the makefile and Jenkinsfile so they're not part of this PR. With @wds15's last changes, this shouldn't be a problem.

Issue #1036, opened 2018-09-14
quantile functions


bob-carpenter:
## Description

Add a function 

`real quantile(real[] x, real p)`

that returns the `p`-th quantile of `x`.  The sort functions can be used or the Boost accumulators to do it directly.  The value `p` should be tested that it's between 0 and 1 inclusive.

It could also be implemented for `vector` and `row_vector`.

This should come with a non-differentiability warning in the manual.

## Current Math Version
v2.18.0

sakrejda:
@bob-carpenter : I believe we should add a detail to this issue so the next person doesn't make the same mistake as the first implementer.  My suggestion is: 

> Because this is a step function w.r.t. all arguments, when this function is instantiated with `var`/`fvar<T>` types it should still only return a `real` (explicitly drop the autodiff type and use only its value). 

I don't think there's a policy about whether to do this (or anything) with step functions in the math lib but maybe @syclik could comment on that part too?

bob-carpenter:
Why did we want to cut the value here so there's no autodiff?  I know the whole function won't be continuous, but why not return the same type as the input?

For example, if I have `a = { b, c, d, e }` with `b < c < d < e` and I set `f = quantile(a, 0.75) * g`, should I not get the same result as `d * g`?  If we reduce the type and return a `double`, that won't be the case.   This will be tricky to doc one way or another.

I realize now I don't know how `quantile` is even supposed to be defined if say, we have 5 elements and ask for a 50% quantile or 3 elements and ask for a 25% quantile or something.  There must be general principles for this.  R at least must do something.  R seems to linearly interpolate:

```
> quantile(c(1,2,3,4), 0.5)
50% 
2.5 

> quantile(c(1,2,30,1000), 0.5)
50% 
 16 

> quantile(c(0, 1), 1/3)
33.33333% 
0.3333333 
```

syclik:
@sakrejda: what should I be commenting on? I think we should just respect the autodiff types if it makes sense?

I'm with @bob-carpenter here... why should we make this a non-autodiff return type?

sakrejda:
I said (above) autodiff doesn't make sense because this is always a step function with regards to all arguments so there's no sensible autodiff to be had (zero or NaN?), but I tagged you (@syclik) because I thought you could add some detail about what makes sense since this is marked as a "good first issue" but actually it's pretty complicated.  We could always implement a smoothed quantile function but I didn't think that's what this issue was about.

bgoodri:
[...]

PR #1105, opened 2019-01-22
Feature/issue 1036 add quantile method

Licht-T:
## Summary

This implements the method for evaluating the `q`-th quantile of the specified data.

## Tests

- `test/unit/math/prim/arr/fun/quantile_test.cpp`
- `test/unit/math/prim/mat/fun/quantile_test.cpp`

## Side Effects

No.

## Checklist

- [x] Math issue #1036, and also related to #1106

- [X] Copyright holder: Licht Takeuchi

### DIFF
--- a/stan/math/prim/arr.hpp
+++ b/stan/math/prim/arr.hpp
@@ -28,6 +28,7 @@
 #include <stan/math/prim/arr/fun/promote_elements.hpp>
 #include <stan/math/prim/arr/fun/promote_scalar.hpp>
 #include <stan/math/prim/arr/fun/promote_scalar_type.hpp>
+#include <stan/math/prim/arr/fun/quantile.hpp>
 #include <stan/math/prim/arr/fun/rep_array.hpp>
 #include <stan/math/prim/arr/fun/scaled_add.hpp>
 #include <stan/math/prim/arr/fun/sort_asc.hpp>
--- a/stan/math/prim/arr/fun/quantile.hpp
+++ b/stan/math/prim/arr/fun/quantile.hpp
@@ -1,0 +1,44 @@
+#ifndef STAN_MATH_PRIM_ARR_FUN_QUANTILE_HPP
+#define STAN_MATH_PRIM_ARR_FUN_QUANTILE_HPP
+
+#include <stan/math/prim/scal/err/check_greater_or_equal.hpp>
+#include <stan/math/prim/scal/err/check_less_or_equal.hpp>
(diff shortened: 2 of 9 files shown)


bob-carpenter:
We don't allow `scal` to import `arr` or `mat`---the includes have to go the other way.  So if there are specialized implementations for `arr` or `mat`, they should go in the respective directories.

It's always better to template definitions rather than cut-and-pasting.   We can block or unblock `int` with `disable_if` or specialize if it needs a different implementation.  

The implementation of quantiles here isn't particularly efficient as it does a complete sort whereas, for example, you only need to keep 5% of the entries sorted to do a 5% quantile.  Boost has some accumulators which implement the quantiles in a streaming way with minimal memory overhead: https://www.boost.org/doc/libs/1_55_0/doc/html/accumulators/user_s_guide.html

Just to be clear, though, I don't think this the added efficiency is necessary for this PR.

syclik:
@Licht-T, feel free to reopen this PR if you address the review. I'm closing it for now.

sakrejda:
@syclik you said: 

> What was originally specified in the issue was to accept real -- that's the Stan type real. That actually maps into double and var (and hopefully fvar<T> soon).

Since the quantile function itself will never be differentiable for either of its arguments, does it need to be implemented for `var`? Is there a policy about this documented somewhere? We could kick this conversation to discourse if there is one to be had but without more background I would support having functions that are not differentiable take only data arguments and throwing errors on compilation that could be reported via `stanc`.

bob-carpenter:
We don't have any functions other than the functional like ode_integrate that require data arguments.  

I'd be OK restricting other functions.  We didn't want to restrict things like floor because we thought there'd be uses for interpolation functions.  But we should probably just build interpolation functions ourselves for people.

The new parser may support this more easily---the existing one doesn't have a slot in the signatures for ordinary functions to record data only.

syclik:
@sakrejda: I see... I read the issue first.

There hasn't been an policy written explicitly anywhere, but I've always assumed that if some expression using the Math library works for `double`, it should also work for `stan::math::var` and vice versa. Or at least be compilable even if we throw an exception stating that it doesn't work for that instantiation. Does that make sense? I'd rather we write the templated version that works for `stan::math::var` and be done with it. If there's an interpolation that happens, we can autodiff through that too.

bob-carpenter:
The exception are the ODE and algebraic solver and map_rect, all of which specify data-only arguments.  They're not regular functions, though---they're handled separately in the expression grammar as of the current C++ parser.

Issue #1117, opened 2019-02-12
Rev for Cholesky on GPU


SteveBronder:
## Description

Adds the GPU version of reverse mode for the Cholesky

## Example

This should be pretty plug and play where users can now use `matrix_v` and var types and have the gradient done on the GPU.

## Expected Output

Big and cool models!

#### Current Version:
v2.18.1

PR #1118, opened 2019-02-12
Rev for Cholesky on GPU

SteveBronder:
## Summary

Last PR!

When users define `STAN_OPENCL` during compilation the reverse mode cholesky will take place on the GPU.

Kinda nice, using all the stuff we built before the code here is pretty similar to the code for the Eigen version of the derivative in `cholesky_block`

## Tests

If you have clinfo installed, you can use `clinfo -l` to see your available platforms and device index  values. With only one platform and device the below should work, otherwise place the appropriate index values for your device and platform below.

```bash
echo STAN_OPENCL=true>> make/local
echo OPENCL_PLATFORM_ID=0>> make/local
echo OPENCL_DEVICE_ID=0>> make/local
./runTests.py ./runTests.py test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp 
```

## Side Effects

Moves Cholesky derivative from CPU to GPU

## Checklist

- [x] Math issue #1117

* Rok Češnovar and Erik Štrumbelj (Faculty of Computer and Information Science, University of Ljubljana)
* Steve Bronder


test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
```diff
@@ -226,7 +230,7 @@ void test_gp_grad(int mat_size, double prec) {
   Eigen::VectorXd test_vals(3);
   test_vals[0] = 1;
   test_vals[1] = 1.5;
-  test_vals[2] = 1;
+  test_vals[2] = 10;
```
SteveBronder: test_vals[2] is the jitter that is added to the diagonal. We up that a bit so that large matrices are more stable

SteveBronder:
(fyi I want to make one less pass at this then I think we are good to review!)

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -11,6 +11,16 @@
...
 #include <stan/math/prim/mat/err/check_symmetric.hpp>
+#include <stan/math/gpu/cholesky_decompose.hpp>
+#include <stan/math/gpu/constants.hpp>
+#include <stan/math/gpu/diagonal_multiply.hpp>
+#include <stan/math/gpu/copy.hpp>
+#include <stan/math/gpu/lower_tri_inverse.hpp>
+#include <stan/math/gpu/matrix_gpu.hpp>
+#include <stan/math/gpu/multiply.hpp>
+#include <stan/math/gpu/err/check_square.hpp>
+#include <stan/math/gpu/err/check_symmetric.hpp>
```
SteveBronder: TODO(me): Remove check_ from gpu since we don't use those here

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+      Lbar.sub_block(Cbar, 0, 0, k, j, m_k_ind, k_j_ind);
+    }
+    copy(Lbar_, Lbar);
+    pos = 0;
+    for (size_type j = 0; j < M_; ++j)
+      for (size_type i = j; i < M_; ++i)
+        variRefA_[pos++]->adj_ += Lbar_.coeffRef(i, j);
+  }
+};
+#endif
```
SteveBronder: Sean as I'm looking over the different methods we have here, it looks like they all pretty much do the same constructor but then have different chain methods. Do you think it's worth/good to refactor these in this PR to just have a single class or keep them separate?

test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
```diff
@@ -432,3 +437,21 @@ TEST(AgradRevMatrix, check_varis_on_stack_large) {
...
+  test_simple_vec_mult(45, 1e-08);
+}
+
+TEST(AgradRevMatrix, check_varis_on_stack_large_opencl) {
+  stan::math::matrix_v m1 = stan::math::matrix_v::Random(1250, 1250);
+  stan::math::matrix_v m1_pos_def
+      = m1 * m1.transpose() + 500 * stan::math::matrix_v::Identity(1250, 1250);
+  test::check_varis_on_stack(stan::math::cholesky_decompose(m1_pos_def));
+}
+#endif
```
SteveBronder: Sean we don't have tests here for exceptions, my thought process here was that since those checks go off for gpu and non-gpu code they would be redundant to include for larger matrices

test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
```diff
@@ -432,3 +437,21 @@ TEST(AgradRevMatrix, check_varis_on_stack_large) {
...
+
+#ifdef STAN_OPENCL
+TEST(AgradRevMatrix, mat_cholesky_1st_deriv_large_gradients_opencl) {
+  stan::math::opencl_context.tuning_opts().cholesky_size_worth_transfer = 25;
+  test_gradient(51, 1e-08);
+  test_gp_grad(1300, 1e-08);
+  test_gp_grad(2000, 1e-08);
+  test_chol_mult(50, 1e-08);
+  test_simple_vec_mult(45, 1e-08);
+}
```
SteveBronder: `test_gradient`, `test_chol_mult`, and `test_simple_vec_mult` all evaluate the gradient `size` times. And I think they need to?

Is it cool to just lower the transfer to gpu size here so the GPU code goes off?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -265,16 +412,34 @@ inline Eigen::Matrix<var, -1, -1> cholesky_decompose(
...
+      cholesky_block* baseVari = new cholesky_block(A, L_A);
+      size_t pos = 0;
+      for (size_type j = 0; j < L.cols(); ++j) {
+        for (size_type i = j; i < L.cols(); ++i) {
+          L.coeffRef(i, j).vi_ = baseVari->variRefL_[pos++];
+        }
+        for (size_type k = 0; k < j; ++k)
+          L.coeffRef(k, j).vi_ = dummy;
+      }
```
SteveBronder: Related to refactoring, if these all called one class and we had the chain method use the different algorithms it would clean the code up there a lot. Again not sure if that should be done in this PR or a separate one

stan/math/rev/mat/fun/cholesky_decompose.hpp
SteveBronder: Sorry, for this comment I mean we use the cpu versions here, not that we don't test for these.

Though we could use the gpu version. I can play with that

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+   */
+  cholesky_gpu(const Eigen::Matrix<var, -1, -1>& A,
+               const Eigen::Matrix<double, -1, -1>& L_A)
+      : vari(0.0),
+        M_(A.rows()),
+        variRefA_(ChainableStack::instance().memalloc_.alloc_array<vari*>(
+            A.rows() * (A.rows() + 1) / 2)),
+        variRefL_(ChainableStack::instance().memalloc_.alloc_array<vari*>(
+            A.rows() * (A.rows() + 1) / 2)) {
+    size_t pos = 0;
```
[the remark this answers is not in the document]
SteveBronder: I prefer to try to tidy this up so yeah I'll do that

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+   *
+   * variRefL aren't on the chainable autodiff stack, only used for storage
+   * and computation. Note that varis for L are constructed externally in
+   * cholesky_decompose.
+   *
+   *
+   * @param A matrix
+   * @param L_A matrix, cholesky factor of A
+   */
+  cholesky_gpu(const Eigen::Matrix<var, -1, -1>& A,
```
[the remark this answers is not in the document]
SteveBronder: Yes totally. After this PR I'm going to also make another PR changing all the `matrix_gpu`s to `matrix_cl` and rename the folder

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+      : vari(0.0),
+        M_(A.rows()),
+        variRefA_(ChainableStack::instance().memalloc_.alloc_array<vari*>(
+            A.rows() * (A.rows() + 1) / 2)),
+        variRefL_(ChainableStack::instance().memalloc_.alloc_array<vari*>(
+            A.rows() * (A.rows() + 1) / 2)) {
+    size_t pos = 0;
+    for (size_type j = 0; j < M_; ++j) {
+      for (size_type i = j; i < M_; ++i) {
+        variRefA_[pos] = A.coeffRef(i, j).vi_;
```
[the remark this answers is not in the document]
SteveBronder: I feel like there's been a discussion about `coeffRef` somewhere, I'll try to look on discourse.

From the docs it looks like `coeffRef` skips an assertion? So we might use it because it's faster

https://eigen.tuxfamily.org/dox/classEigen_1_1DenseCoeffsBase_3_01Derived_00_01WriteAccessors_01_4.html#ae66b7d18b2a85f3139b703126974c900

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+      for (size_type i = j; i < M_; ++i) {
+        Lbar_.coeffRef(i, j) = variRefL_[pos]->adj_;
+        L_.coeffRef(i, j) = variRefL_[pos]->val_;
+        ++pos;
+      }
+    }
+
+    matrix_gpu L(L_);
+    matrix_gpu Lbar(Lbar_);
+    int M = M_;
```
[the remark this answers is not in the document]
SteveBronder: yikes!

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+    int block_size_ = std::max((M / 8), 8);
+    block_size_ = std::min(block_size_, 512);
```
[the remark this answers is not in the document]
SteveBronder: idk? @rok-cesnovar can we use the `cholesky_partition` here? Or should this be its own tuning parameter?

stan/math/rev/mat/fun/cholesky_decompose.hpp
SteveBronder: So my idea to do this simply would be something like

```cpp
class cholesky : public vari {
 public:
  int M_;
  vari** variRefA_;
  vari** variRefL_;
  cholesky(const Eigen::Matrix<var, -1, -1>& A,
                 const Eigen::Matrix<double, -1, -1>& L_A) {
  ...all share the same constructor...
  }
 void chain_scalar() {code..}
 void chain_block() {code..}
 void chain_cl() {code..}
 virtual void chain() {
  if (M_ < small_num) {
    chain_scalar();
  } else if (M_ > big_num) {
    chain_cl();
  } else {
    chain_block();
  }
 }
};
```

If you've tried it before I'll trust your judgement that there's stuff going on here that would make the above more complicated than I think it is

test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
SteveBronder: @rok-cesnovar didn't eigen also fail for large N here?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +233,133 @@ class cholesky_scalar : public vari {
...
+    // The following is a GPU implementation of
+    // the chain() function from the cholesky_block
+    // vari class implementation
+    for (int k = M; k > 0; k -= block_size_) {
+      int j = std::max(0, k - block_size_);
+      int k_j_ind = k - j;
+      int m_k_ind = M - k;
```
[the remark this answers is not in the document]
SteveBronder: These will change over each loop so they can't be const right?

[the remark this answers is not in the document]
seantalts: (Assuming you mean the loop on 307) They can be const! Each time the loop goes around it gets a new scope with new variables. For the existence of `k_j_ind` during a single loop iteration, it's constant and can be marked const. The next time the loop goes around, you define another `k_j_ind` that is in almost every sense unrelated to the first one, and also constant for its lifespan.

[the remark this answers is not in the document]
SteveBronder: Oh! scope, totally forgot. Indeed yes. Would it be better to move the creation of those out of the loops scope or make them const?

test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
rok-cesnovar: Yep, it seems that this way of producing the input matrix is not suitable for large matrices.

stan/math/rev/mat/fun/cholesky_decompose.hpp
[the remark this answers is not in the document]
rok-cesnovar: This is separate from the `cholesky_partition`. The size here is determined the same way as in the Eigen version of chain(),  only that the max size of block_size is a bit bigger. This should be a separate parameter.

rok-cesnovar:
Apologize for being an idiot with these merges/reverting. I synced our fork and forgot that develop doesnt have prim GPU changes atm. Fortunately we have to wait for #1120 to go through anyway.

Jenkinsfile
```diff
@@ -172,12 +172,19 @@ pipeline {
                     }
                     post { always { retry(3) { deleteDir() } } }
                 }
-                stage('Windows Headers & Unit') {
+                stage('Windows Headers') {
```
rok-cesnovar: What are this changes to the Jenkins files for?

seantalts: Looks like a bad merge with develop - we changed to "Windows Headers & Unit" on develop in the meantime

rok-cesnovar: Yeah, I messed up and also got in some gp_cov_exp stuff. My bad.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -241,9 +384,19 @@ inline Eigen::Matrix<var, -1, -1> cholesky_decompose(
   check_symmetric("cholesky_decompose", "A", A);
 
   Eigen::Matrix<double, -1, -1> L_A(value_of_rec(A));
+#ifdef STAN_OPENCL
+  if (L_A.rows() > opencl_context.tuning_opts().cholesky_size_worth_transfer) {
+    // NOTE: We don't do the pos def check here because
+    // The GPU cholesky only returns the lower left
+    L_A = cholesky_decompose(L_A);
```
rok-cesnovar: Just to clarify this, we do check if the input was pos def in the same way that Eigen does [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/err/check_pos_definite.hpp#L84), its just that we do that inside the cholesky_decompose call. 

check_pos_definite checks if the decomposition was successful (no numerical issues, basically no NaNs) and if the diagonal is non-zero which is the same as we do [here](https://github.com/stan-dev/math/blob/develop/stan/math/gpu/cholesky_decompose.hpp#L92).

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -241,9 +384,19 @@ inline Eigen::Matrix<var, -1, -1> cholesky_decompose(
   check_symmetric("cholesky_decompose", "A", A);
```
rok-cesnovar: More on the debate on checks. For check_square I think leaving this here is fine, since performance wise this does the same thing (access to the the rows and cols members).

check_symmetric would be more performant if we did it with OpenCL, but that means that we would call check_symmetric at line 390 and before 393 but on L_A not A, so an unnecessary value_of_rec call happens if its actually not symmetric.

Perforamance wise this is nothing to worry about so I am fine leaving this as is for the sake of cleaner code.

rok-cesnovar:
I am going to do similar tests as we did for Stancon with the same or some other model with cholesky_decompose. 

I will post the comparison of fits for different input sizes with OpenCL and without it so we can show that there is no practical difference between the two. Also the speedups only for the gradient is something we want to show here, I will do those for the Titan Xp & V100 vs an i7/i5.

stan/math/rev/mat/fun/cholesky_decompose.hpp
SteveBronder: Yeah, the call to `value_of_rec` is also v expensive. I'll look at this later to see if there is a more clever way to do it but I'm fine with the CPU version

SteveBronder: Also a good example of where Seans previous PR that saves flags for when these checks pass on the data would solve this without having to do any gpu stuff

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,7 +235,157 @@ class cholesky_scalar : public vari {
...
+
+  for (size_type j = 0; j < L.cols(); ++j) {
+    for (size_type i = j; i < L.cols(); ++i) {
+      L.coeffRef(i, j).vi_ = variRef[pos++];
+    }
+    for (size_type k = 0; k < j; ++k)
+      L.coeffRef(k, j).vi_ = dummy;
+  }
+  return L;
+}
```
SteveBronder: @seantalts I like this because it cleans up the code below a lot, but I'm not really sure where to put it.

 Good, bad, thoughts?

Also not sure if it should return `Matrix<var, -1, -1>` or `void` since it's manipulating a reference to L

syclik:
re: input checking. I just want to make sure everyone understands the goals.

- we don't expect users to provide valid inputs to the functions; this is a reasonable assumption since the user is accessing this from the Stan language and they eventually do pass in values that fail.
- if there is an input error, the goal is to notify the user what the error is as specifically and as soon in the chain as we can trap it. It's really not useful if the user calls `foo()` and the exception back says the throwing function is `baz()`.
- if there is an input error, we're not concerned with speed; the try / catch block goes into affect and essentially destroys performace. We have to be careful to clean up resources, but we can take our time in writing nice error messages at this point.
- our functions should never seg fault.

There's no reason we must check all input at the top. We could have some other mechanism for checking. If there's a way to validate after running and that can trap everything, that works too.

stan/math/rev/mat/fun/cholesky_decompose.hpp
SteveBronder: Thanks Rok! I'll update the comment

SteveBronder:
Thanks Dan! I think I caused some confusion on this so my bad.

wrt `check_square` and `check_symmetric` I was not saying, "Let's not do checks" I was saying, "Let's do the CPU version of the checks instead of the GPU version of the checks" because idt the transfer to the GPU is worth it just to do these checks on the GPU.

For `check_pos_definite` we do the same thing Eigen does. Rok clarifies that in this comment [here](https://github.com/stan-dev/math/pull/1118#discussion_r258646878). I need to fix my comment in the code because I think it's not truthful

stan/math/rev/mat/fun/cholesky_decompose.hpp
seantalts: In most of the communities I've been in, it's considered good style to either have a function be pure and return the result, or return `void` and have the inputs clearly marked for modification. So here I would return void. 

Might be good to put it in `internal` - I think it's fine in this file.

SteveBronder: :+1 agree on void I'll put this in `internal`

seantalts:
Is this ready for review?

SteveBronder:
Yep! Was about to write a message. Only two things that are unresolved atm are code cleaning that's not directly related to the GPU code.

- [ ] Moving that little assignment function above to `internal` 
- [ ] Function for assignment in the cholesky classes that gets repeated a few times (from comment [here](https://github.com/stan-dev/math/pull/1118#discussion_r256533834))

rok-cesnovar:
I seem to have underestimated the execution time of `check_symmetric` on the CPU. 

For larger sizes (rows >6000+) the `check_symmetric` on an high-range i7 can represent up to 1/3 of the execution time when using the TitanXp for the rest of cholesky. Fix incoming.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -238,12 +402,20 @@ class cholesky_scalar : public vari {
 inline Eigen::Matrix<var, -1, -1> cholesky_decompose(
     const Eigen::Matrix<var, -1, -1>& A) {
   check_square("cholesky_decompose", "A", A);
-  check_symmetric("cholesky_decompose", "A", A);
```
rok-cesnovar: I moved the check_symmetric to the CPU versions as in the OpenCL version, this is actually duplicated (this check happens in `prim`). I left check_square as is, as it does not affect performance.

SteveBronder: Great catch!

rok-cesnovar:
Just to give you some taste on where are the bottlenecks for the execution time and what is worth exploring in the (near) future if anything can be done at all.

This is a figure for the reverse mode specialization of cholesky decomposition (the function that starts [here](https://github.com/stan-dev/math/blob/0e2ef93c87ccfa7ce53dc9ff4a93858d6f82b5ef/stan/math/rev/mat/fun/cholesky_decompose.hpp#L402)).

execution times for parts of the reverse mode specialization of cholesky_decompose

For the cholesky itself (red on the figure), we are now 15x/33x faster than an i7 with Eigen when using the TitanXP/V100 at input size 10000x10000. But the actual speedup is in the range of 7-10 for the reverse mode specialization.

This is all without `chain()` where my initial perf analysis shows that we are more in the speedup range of 15-25. There is still a lot of space for optimization in the chain function (even some stuff that we used for the Stancon paper), but I would much rather get this through with the 15-25 speedup and then do an optimization.

bob-carpenter:
I completely agree that it'd be nice to have these features released before we take the next pass in optimizing them.

SteveBronder:
Agree with Bob, we have good performance now and there's def a lot of layups in terms of optimization we can do.

I have toy examples written out for both expression templates and async queues. Async queues are working well, but I'm having some trouble with assignment in the expression template. @bob-carpenter if I put up a PR with the expression templating example would you have time to take a look at it?

seantalts:
Ah okay, I can take a look at this tomorrow morning. Thanks! 

Also Steve did you mess around with the Eigen Map approach to value_of? We could try making `check_symmetric` take an `Eigen::DenseBase` and see if passing in an Eigen Map with a stride of 2 (pointing at a var matrix) cuts times down? That's probably a thing for another PR I guess...

SteveBronder:
> Also Steve did you mess around with the Eigen Map approach to value_of?

Not since we last spoke. Been working on some goofy side projects and haven't had time. I think it's another PR but yeah it seems reasonable (?)

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
...
+      D.sub_block(L, j, j, 0, 0, k_j_ind, k_j_ind);
+      B.sub_block(L, k, 0, 0, 0, m_k_ind, j);
+      C.sub_block(L, k, j, 0, 0, m_k_ind, k_j_ind);
+
+      Rbar.sub_block(Lbar, j, 0, 0, 0, k_j_ind, j);
+      Dbar.sub_block(Lbar, j, j, 0, 0, k_j_ind, k_j_ind);
+      Bbar.sub_block(Lbar, k, 0, 0, 0, m_k_ind, j);
+      Cbar.sub_block(Lbar, k, j, 0, 0, m_k_ind, k_j_ind);
+
+      Cbar = Cbar * lower_triangular_inverse(D);
```
seantalts: Do we want to use a solver here instead? 
image

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
...
+      const int j = std::max(0, k - block_size_);
+      const int k_j_ind = k - j;
+      const int m_k_ind = M_ - k;
+
+      matrix_gpu R(k_j_ind, j);
+      matrix_gpu D(k_j_ind, k_j_ind);
+      matrix_gpu B(m_k_ind, j);
+      matrix_gpu C(m_k_ind, k_j_ind);
+
+      matrix_gpu Rbar(k_j_ind, j);
```
seantalts: So the "bar"s here just refer to sensitivities, an alternative name for which (that we use) is "adjoints," often abbreviated `adj`.

image

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
     }
   }
 };
+#ifdef STAN_OPENCL
+class cholesky_opencl : public vari {
+ public:
+  int M_;
+  vari** variRefA_;
```
seantalts: We use the naming rules from Google's style guide: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules and https://google.github.io/styleguide/cppguide.html#Variable_Names

In particular, we don't use mixed case for variable names, preferring all lower case with underscores between words.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
...
+      Dbar.sub_block(Lbar, j, j, 0, 0, k_j_ind, k_j_ind);
+      Bbar.sub_block(Lbar, k, 0, 0, 0, m_k_ind, j);
+      Cbar.sub_block(Lbar, k, j, 0, 0, m_k_ind, k_j_ind);
+
+      Cbar = Cbar * lower_triangular_inverse(D);
+      Bbar = Bbar - Cbar * R;
+      Dbar = Dbar - transpose(Cbar) * C;
+
+      // the implementation of the symbolic_rev inline function
+      // for the GPU
```
seantalts: Any reason not to pull this out into an `inline` function?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
...
+      Dbar = transpose(D) * Dbar;
+      Dbar.triangular_transpose<TriangularMapGPU::LowerToUpper>();
+      D = transpose(lower_triangular_inverse(D));
+      Dbar = D * transpose(D * Dbar);
+      Dbar.triangular_transpose<TriangularMapGPU::LowerToUpper>();
+      // end of symbolic_rev inline function
+
+      Rbar = Rbar - transpose(Cbar) * B;
+      Rbar = Rbar - Dbar * R;
```
seantalts: seems like you could put this on the line above

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -265,16 +437,21 @@ inline Eigen::Matrix<var, -1, -1> cholesky_decompose(
...
+        > opencl_context.tuning_opts().cholesky_size_worth_transfer) {
+      cholesky_opencl* baseVari = new cholesky_opencl(A, L_A);
+      internal::set_lower_tri_coeff_ref(L, baseVari->variRefL_);
+    } else {
+      cholesky_block* baseVari = new cholesky_block(A, L_A);
+      internal::set_lower_tri_coeff_ref(L, baseVari->variRefL_);
     }
+#else
+    cholesky_block* baseVari = new cholesky_block(A, L_A);
+    internal::set_lower_tri_coeff_ref(L, baseVari->variRefL_);
```
seantalts: Seems like this line is the same in all 3 branches of the `ifdef` and `if` - can we put it after?

test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
```diff
@@ -432,3 +437,22 @@ TEST(AgradRevMatrix, check_varis_on_stack_large) {
   X = stan::math::multiply(X, stan::math::transpose(X));
   test::check_varis_on_stack(stan::math::cholesky_decompose(X));
 }
+
+#ifdef STAN_OPENCL
+TEST(AgradRevMatrix, mat_cholesky_1st_deriv_large_gradients_opencl) {
+  stan::math::opencl_context.tuning_opts().cholesky_size_worth_transfer = 25;
```
seantalts: Can we add a test below the threshold too?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -11,11 +11,46 @@
...
+namespace internal {
+/**
+ * Set the lower right triangular of a var matrix given a set of vari**
+ *
+ * @param L Matrix of vars
+ * @param variRef Values to be set in lower right triangular of L.
+ * @return None, L modified by reference.
+ */
+inline void set_lower_tri_coeff_ref(Eigen::Matrix<var, -1, -1>& L,
+                                    vari**& variRef) {
```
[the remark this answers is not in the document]
SteveBronder: Yeah this is probably a weird goof on my part

stan/math/rev/mat/fun/cholesky_decompose.hpp
seantalts: I forgot to go back and put some more context on this comment. We've talked about trying to use the names from the papers we're implementing as a way to easily go back and forth, but I think we didn't want to talk about that we also want to balance that with the Math library being internally consistent, such that once you understand one area of it you could more easily translate your knowledge into another area. We do this with little things like naming conventions and trying to use the same vocabulary everywhere. But we do want to create a balance - I think in this case a reasonable compromise might be something like `R_adj` here while mentioning the adj -> bar mapping in the doxygen comment containing the reference to the paper. I think that lets a reader go back and forth with the paper while being a little more fluent based on their previous knowledge of the math library. It's definitely a negotiable judgement call but just wanted to explain the two goals we want to balance here.

rok-cesnovar: We were more trying to make a 1-to-1 translation of the existing Cholesky_block code than trying to figure out our own names. Thats where both the bar an camel case variRefA come from. I do agree with you that we want to have things go by the naming convention throughout the code base. Can we also change the cholesky_block in this PR or do that separately?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
...
+      for (size_type i = j; i < M_; ++i) {
+        Lbar_.coeffRef(i, j) = variRefL_[pos]->adj_;
+        L_.coeffRef(i, j) = variRefL_[pos]->val_;
+        ++pos;
+      }
+    }
+
+    matrix_gpu L(L_);
+    matrix_gpu Lbar(Lbar_);
+    int block_size_
```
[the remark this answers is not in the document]
rok-cesnovar: Copy-pasting from `cholesky_block` where `block_size` is a class member. Thanks.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: I did a few tries with solvers but never managed to implement a faster solver for a GPU than our current `lower_tri_inverse` plus a matrix multiply. There is also an optimization coming here in one of the next PR because we can get some additional performance with knowing the second matrix in the multiplication is lower triangular.

If you want I can have another go at a solver, no problem.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: Nevermind my question, I have to change it for all vari classes, otherwise `baseVari->vari_ref_L_` wont compile.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: It wont work, I tried it but I get the compiler error `error: ‘baseVari’ was not declared in this scope`.
The problem is that the first one is a `cholesky_opencl` class while the other 2 are `cholesky_block`.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -223,6 +258,135 @@ class cholesky_scalar : public vari {
...
+   * Reverse mode differentiation algorithm using a GPU
+   *
+   * Reference:
+   *
+   * Iain Murray: Differentiation of the Cholesky decomposition, 2016.
+   *
+   */
+  virtual void chain() {
+    using Eigen::MatrixXd;
+    MatrixXd Lbar_(M_, M_);
```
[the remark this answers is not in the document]
rok-cesnovar: Thanks! I had to add an .eval().

test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp
rok-cesnovar: Added a `test_gradient` and `test_gp_grad` under the threshold.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: No reason at all. Fixed.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: Done, thanks.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: variRefA_ and variRefL_ were replaced. There are similar cases in some other files, should I make a "cleanup mixed case" issue?

rok-cesnovar:
@seantalts I believe I addressed all your comments.

Regarding precision: my plan was to compare fits with STAN_OPENCL and without for either this example ([example.txt](https://github.com/stan-dev/math/files/2911630/example.txt)) from the Stan manual or the model we used for [StanCon](https://github.com/SteveBronder/stancon2018/blob/master/_Simulations/_Models/gp.stan) ?
Any thoughts on this? I could run this with different values for the primitive cholesky_decompose parameters.

stan/math/rev/mat/fun/cholesky_decompose.hpp
SteveBronder: Yeah, Sean is there a trick for the above scoping issue idk about in C++? Agree it would be nice to not have this repeated multiple times

stan/math/rev/mat/fun/cholesky_decompose.hpp
seantalts: Ah that'd be great, thanks! In retrospect I should have mentioned this earlier probably, but I felt like we already had a lot to deal with that was more important. Now things are all really good and so we're getting into marginal returns 😂

stan/math/rev/mat/fun/cholesky_decompose.hpp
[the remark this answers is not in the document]
seantalts: Oops, good catch.

stan/math/rev/mat/fun/cholesky_decompose.hpp
seantalts: Oops. The only trick I can think of would be if this was a virtual method call on an interface declared above the if. Might be a potential way to refactor if we saw lots of use cases, but for now we can leave it.

seantalts:
Re: Precision. Hmm. I talked with Bob a little about this, and I don't think there's a clear way to measure precision as it kind of depends on the matrix sizes and contents thereof. There are many possible paths, and I think we should just pick one or two and be clear about the assumptions and what we tested (and code is usually pretty clear). I'll propose doing two things:
1. SBC on the model from the manual to see if we can spot any issues with fitting. Ben is adding SBC support to RStan and it's almost in, see https://github.com/stan-dev/rstan/pull/611. Let's run something like 100 replications of 10 posterior draws each and see what we can see at first.
2. @bgoodri's idea for a Cholesky decomposition unit test:
> https://en.wikipedia.org/wiki/Cholesky_decomposition#Updating_the_decomposition
where we start with a Cholesky factor of a big covariance matrix and then imagine rank-1 or rank-2 updates to the covariance matrix. That can be done more-or-less exactly by direct manipulation of the original Cholesky factor that we can compare to the naive way of multiplying the Cholesky factor by its own transpose, adding the updates, and taking the Cholesky factor of the result.

We can add a test for the relative error for this procedure on GPU and our CPU versions to the source code (see [this](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html ) for measuring relative error). Neither of these have to be done before this PR goes in, but we should do them before we release GPUs to the world, and then we can add a little section describing GPU precision to the manual section on GPU operations.

bob-carpenter:
My point was that whatever the precision, it probably shouldn't block adding GPUs.

I think testing with a known Cholesky factor is the way to go.  I'm not sure what you mean by rank 1 or rank 2 updates.  Is there a motivation for manipulating the Cholesky factor rather than just generating new ones?  The tricky numerical situations often arise with bad conditioning (high ratio of largest to smallest eigenvalue).  But I'm not sure how to generate a test matrix with a given condition.  We might be able to do it approximately in Stan with a prior on the eigenvalues.  Or is that what the low rank updates provide?

seantalts:
> My point was that whatever the precision, it probably shouldn't block adding GPUs.

But did I interpret you correctly as saying it would be good to somehow characterize the precision before adding them? Or can we wait on that as well?

seantalts:
@bgoodri do you know how to do what Bob is suggesting (basically generating good test matrices that are tricky for decomposition)?

SteveBronder:
>  I'm wondering if there's anything quick&easy we could do in the test to ensure the GPU version is being called... any ideas?

We could set `opencl_context.tuning_opts().cholesky_size_worth_transfer = 0;` to have that condition always go off right? Or is there somethign else you are worried about that would stop the test from happening

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -238,12 +404,20 @@ class cholesky_scalar : public vari {
 inline Eigen::Matrix<var, -1, -1> cholesky_decompose(
     const Eigen::Matrix<var, -1, -1>& A) {
   check_square("cholesky_decompose", "A", A);
-  check_symmetric("cholesky_decompose", "A", A);
-
   Eigen::Matrix<double, -1, -1> L_A(value_of_rec(A));
+#ifdef STAN_OPENCL
+  if (L_A.rows() > opencl_context.tuning_opts().cholesky_size_worth_transfer) {
```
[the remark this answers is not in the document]
rok-cesnovar: The reasons why we left it here is due to the comment above that states: 

```
Internally calls Eigen::LLT rather than using stan::math::cholesky_decompose in order to use an inplace decomposition.
```
But on second thought, if someone defines STAN_OPENCL then the CPU version would only be called for smaller matrices where it doesnt really matter if we do them inplace or not.

seantalts:
I'm thinking more to verify that the threshold is working correctly... like could we check the opencl context to see that it has had some matrices pass through it or something? any other little side effects like that?

SteveBronder:
I'll have to check, @rok-cesnovar do you know? I think since we don't track events or do profiling 
 there's nothing specific in the queue to tell us that info.

In the OpenCL Spec there's a lot of `getInfo()` functions so we may be able to get some stuff from there

I'm not totally understanding the reason to test that tho'? Like isn't that checking if the if statement goes off?

rok-cesnovar:
I cant think of a simple way of doing that for the primitve cholesky without using events that we are currently not using. For the rev cholesky we could possibly check the allocated memory in the arena, idk?

seantalts:
Yeah, basically checking that these tests are actually running on the GPU code we think it's running on. You could imagine someone coming along in the future and changing the templating, overloads, if statement, or tuning params slightly and losing that accidentally while the test will still tell you the cholesky is good to go... Hmm, there might just not be a good way to do it right now, so never mind about that. 

Also @rok-cesnovar I missed that that version was in-place, woops. Makes sense.

SteveBronder:
Ah, yeah Sean I agree that's a very good point! I'm going to iterate on the Async stuff this weekend and with that we will be able to use the events and profiling to check out what is happening and what went off.

SteveBronder:
Also pretty darn cool! I'm going to make an issue tonight with a checklist of things that we want to happen before we make a full announcement about this

SteveBronder:
Also gosh, huge thanks to @rok-cesnovar and @seantalts, pretty cool this is finally kicking!

seantalts:
Thank you for driving this forward relentlessly :) super excited at the progress and looking forward to the announcement :)

rok-cesnovar:
@SteveBronder @seantalts I also want to thank you both for the hard work. Excited :)

Issue #1058, opened 2018-10-29
Add GPU Cholesky Decomposition


rok-cesnovar:
## Description

Adds the GPU Cholesky decomposition for the `matrix_gpu` type, and adds the GPU functionalites to the  `/prim` and `/rev` implementations of the Cholesky Decomposition. This issue will be tackled with 2 pull requests:

- the kernel code for small blocks, the `cholesky_decompose(matrix_gpu&)` function and `/prim` implementation
- `/rev` implementation

## Example
Use of `cholesky_decompose(matrix_gpu&)`:
```
stan::math::matrix_gpu m0(m0_eigen);
using stan::math::cholesky_decompose;
stan::math::matrix_gpu foo = cholesky_decompose(m0);
```
Use of `cholesky_decompose(stan::math::matrix_d)` and `cholesky_decompose(stan::math::matrix_v)`:
```
using stan::math::cholesky_decompose;
stan::math::matrix_gpu foo = cholesky_decompose(m0_eigen);
```

## Expected Output

The expected output is faster execution of Cholesky Decomposition.

## Additional Information

/

## Current Math Version
v2.18.0

PR #1120, opened 2019-02-17
Feature/issue 1058 gpu choleksy primitive

SteveBronder:
## Summary

Below is copy/pasted from original PR. Includes bug fixes for

1. Needed to make the kernel code `static const char*`'s and inlined `stan::math::opencl_kernels::compile_kernel` so that they will work across multiple translation units
2. Breaks out the Cholesky primitive into two functions, one that works on doubles and one that works on everything else. The tests are passing, but the template is looking a bit odd so I'd appreciate a thorough look at that.

----------------------

Adds the gpu cholesky decomposition primitive and it's associated kernel. This also includes two small bug fixes for the inverse cholesky.

When users define `STAN_OPENCL` during compilation the cpu based cholesky will be replaced by the gpu version of the cholesky. 
```cpp
m = cholesky_decompose(x);
```
So users code will stay the same, but will then have access to the gpu routines. This was decided in favor of including a `cholesky_decompose_gpu` method since if users want fine grain decision making we can always introduce it later, but it will be harder to remove it.

The Cholesky's gpu algorithm is recursive, where The parameters `block`, `divider`, and `min_block` act as tuning parameters for the recursive step of the GPU based Cholesky decompostion. The matrix is subset by the `block` size, and if the `block` size is less than `min_block` then the Cholesky decomposition on the GPU is computed using that submatrix. If `block` is greater than `block_size` then `cholesky_decompose` is run again with `block` equal to `block/divider`. Once the Cholesky decomposition is computed, the full Cholesky is created by propagating the Cholesky forward as given in the reference report below.

https://github.com/SteveBronder/stancon2018/blob/master/report.pdf

## Tests

Assuming OpenCL is setup and the preferred device has an index of 0, tests can be run with

```bash
echo STAN_OPENCL=true>> make/local
echo OPENCL_PLATFORM_ID=0>> make/local
echo OPENCL_DEVICE_ID=0>> make/local
./runTests.py ./test/unit/math/gpu/cholesky_decompose_test.cpp 
```
`cholesky_decompose_cpu_vs_gpu`
- Tests whether cpu and gpu outputs are close to each other

`cholesky_decompose_small`
- Runs the `cholesky_decompose_test` tester on small input matrices

`cholesky_decompose_big`
- Runs the `cholesky_decompose_test` tester on input matrices of size 500, 1000, 2000

The `cholesky_decompose_test` tester checks whether the GPU implementation's max error for any cell in the cholesky decomposition relative to Eigen's LLT method is less than 1E-8.

## Side Effects

None

## Checklist

- [x] Math issue #1058 

- Rok Češnovar and Erik Štrumbelj (Faculty of Computer and Information Science, University of Ljubljana)
- Steve Bronder


SteveBronder:
Sean need to re-run all the tests one more time for sanity check then I'll request a review here

stan/math/prim/mat/fun/cholesky_decompose.hpp
```diff
@@ -1,10 +1,17 @@
 #ifndef STAN_MATH_PRIM_MAT_FUN_CHOLESKY_DECOMPOSE_HPP
 #define STAN_MATH_PRIM_MAT_FUN_CHOLESKY_DECOMPOSE_HPP
 
+#include <stan/math/gpu/opencl_context.hpp>
```
rok-cesnovar: Should this be moved inside the STAN_OPENCL ifdef?

SteveBronder:
```
In file included from ./stan/math/prim/mat/fun/cholesky_decompose.hpp:10:0,
                 from ./stan/math/prim/mat.hpp:74,
                 from ./test/unit/math/rev/mat/fun/util.hpp:5,
                 from test/unit/math/rev/scal/fun/logit_test.cpp:5:
./stan/math/gpu/cholesky_decompose.hpp: In function ‘stan::math::matrix_gpu stan::math::cholesky_decompose(stan::math::matrix_gpu&)’:
./stan/math/gpu/cholesky_decompose.hpp:67:73: error: cannot convert ‘stan::math::var’ to ‘const int’ in initialization
       = floor(A.rows() / opencl_context.tuning_opts().cholesky_partition);
```

That seems, kinda weird? Going to look into this today

stan/math/prim/mat/fun/cholesky_decompose.hpp
```diff
@@ -24,11 +24,15 @@ namespace math {
...
+ * <code>cholesky_decompose</code> functions. One that works on doubles
+ * (this one) and another that works on all other types.
  * @throw std::domain_error if m is not a symmetric matrix or
  *   if m is not positive definite (if m has more than 0 elements)
  */
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_decompose(
+template <typename T, typename std::enable_if_t<std::is_same<T, double>::value,
+                                                T>* = nullptr>
+inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_decompose(
```
seantalts: So... this is saying we want to activate this specialization only for doubles, right?

Why don't we just specialize on double? e.g.
```
inline Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> cholesky_decompose(
        const Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>& m) {
```

SteveBronder: I think that makes a lot more sense.

Though the templating here is gross, I wanted them to have similar templates for consistency. But yeah the above is a lot nicer and does the job

seantalts: The other one also becomes simpler, assuming my mental model doesn't have any holes in it:

```
template <typename T>
inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_decompose(
    const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& m) {
```

seantalts: I think the first one might need a `template <>` above it, oops.

This page has a lot more detail on how template specialization works: https://isocpp.org/wiki/faq/templates#template-specialization

SteveBronder: Oh interesting! My assumption was that the template

```cpp
template <typename T>
inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_decompose(
    const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& m) {
```

Would overwrite the doubles specific template. But your above makes that seem false. Ty! I'll read that more tonight and try out your approach.

rok-cesnovar: If you dont mind, I can make the changes and commit rn.

rok-cesnovar: I messed up the order in the first try, the double specialization needs to be written after the general template function.

rok-cesnovar:
To double check I ran `./runTests.py -j2 test/unit -f chol` locally and it all passes.

SteveBronder:
Rok can you test
```
/runTests.py -j2 test/unit/math/rev/scal/fun/logit_test.cpp
```

That was the one giving me an error before

tbh I'd rather take a minute to run the whole test suite so that fwd etc is checked out

rok-cesnovar:
This one is fine too, gonna run the whole test suite now, I agree.

rok-cesnovar:
PASSED on the entire test suite.

seantalts:
Looks good, thank you both. Merging & on to rev

rok-cesnovar:
This develop failure is on sundials stuff not on us right?

wds15:
Sundials stuff?

rok-cesnovar:
I got the FAILURE email from jenkins, on the link I saw this error:

C:\Rtools\mingw_64\bin\ar.exe: unable to rename 'lib/sundials_3.1.0/lib/libsundials_cvodes.a'; reason: Permission denied

I am on mobile right now so I havent gone into details.

rok-cesnovar:
Here is the link: http://d1m1s1b1.stat.columbia.edu:8080/blue/organizations/jenkins/Math%20Pipeline/detail/develop/218/pipeline

syclik:
@rok-cesnovar: yup, not yours. @wds15: not yours either. =).

@seantalts: this looks really familiar... out of space / permissions? What
fixed it last time?

seantalts:
It's not this PR's fault, no worries. I haven't figured it out yet but it
looks like Windows 10 is more adversarial than Windows 7 was about
arbitrarily locking files. There may be some virus scanner or something to
disable. I'm looking into it.

seantalts:
I set the virus protection to exclude the Jenkins working directory (C:\Jenkins2) - apparently if you turn it off completely, it will just turn itself back on later. Which is supposed to be a feature. I also logged out of users and closed Explorer windows, which someone on SO suggested might help. Fingers crossed.

syclik:
Next time I see you, I'm gonna buy you a beer.

Issue #1112, opened 2019-02-05
BesselK continuous function


stemangiola:
## Description
In order to build robust count models based for example on SICHEL distribution, a continuous version of the BesselK function is needed (nu being real instead of int).

There is a tread here https://discourse.mc-stan.org/t/besselk-with-the-order-v-as-real-parameter/7417/4 but I thought to lounch a feature request since is quite a delicate math function (that is not working natively in log scale in the R version and therefore unstable)

For **feature requests**:

The code is present in R 

https://stat.ethz.ch/R-manual/R-devel/library/base/html/Bessel.html

## Expected Output

count ~ sichel(mu, sigma, nu)

#### Current Version:
v2.18.1

martinmodrak:
A related issues in stan: https://github.com/stan-dev/stan/issues/1939

zottelef:
Could this be useful to build a Matérn covariance function  for the Gaussian Processes?

martinmodrak:
For anyone wondering here: the PR is closed, but I am making slow progress now and then and have code that works for many contexts. If you need this function, let me know, I might try to increase the priority of resolving it completely :-)

PR #1121, opened 2019-02-18
Implemented log(BesselK) for fractional orders

martinmodrak:
## Summary

Implemented logarithm of the modified Bessel function of the second kind (Bessel K) for fractional orders, supporting differentiation with respect to both variables.

Log(BesselK) is useful for computing the lpdf of some distributions, such as [Generalized inverse Gaussian](https://en.wikipedia.org/wiki/Generalized_inverse_Gaussian_distribution) or [SICHEL](
https://www.rdocumentation.org/packages/gamlss.dist/versions/5.1-1/topics/SICHEL)

The computation is based on https://github.com/stan-dev/stan/wiki/Stan-Development-Meeting-Agenda/0ca4e1be9f7fc800658bfbd97331e800a4f50011 (but modified to allow for stan::math::var arguments). Thanks to @bgoodri for providing it, I would not have been able to find this formula on my own.

The code snippet linked above is in turn based on Equation 26 of Rothwell: Computation of the 
logarithm of Bessel functions of complex argument and fractional order
https://scholar.google.com/scholar?cluster=2908870453394922596&hl=en&as_sdt=5,33&sciodt=0,33

Both derivatives are computed by auto-diffing over the 1d integrator. It should be possible to get an explicit formula for the derivative with respect to `v` similarly to the way it is computed in `modified_bessel_second_kind`.

The name of the main function (`log_modified_bessel_second_kind_frac`) is provisional. The function is now in `rev/arr/fun`. This is IMHO weird, but I include `rev/arr/functor/integrate_1d` so the linter complained when the function was in `rev/scal/`

In addition, I tried to improve the error messages from `integrate_1d` and its documentation, as following the current documentation led me astray. This could be moved to a separate pull request if desired.

## Tests

A grid of values of the function and its gradients was computed in Mathematica (code is part of the comments in the test file). The relative error between Mathematica and this implementation is <1e-7 for all values and gradients tested.

## Side Effects

The error messages of `integrate_1d` have been modified to include the error threshold.

## Checklist

- [x] Math issue #1112


bgoodri:
CRAN won't allow those `sprintf` things

bgoodri:
I think you should complex-step the derivative with respect to the order when it is a `var` but that is going to require the Boost 1.7.0 PR to get merged so that you can use complex integrands.

bgoodri:
The derivative with respect to the primary argument has many formulas.

bgoodri:
Also, these tests against Mathematica are not very strong in my opinion. It is much better to test with identities, such as at http://functions.wolfram.com/Bessel-TypeFunctions/BesselK/17/ , that hold for all input values. Usually you can just divide the left-hand side of an identity by the right-hand side and check that the ratio is near one and `1 + ` the derivative is near one.

bgoodri:
Also, it may be worth thinking about whether we need a specialized integration routine for `log_foo` functions that utilizes `log_sum_exp` logic.

bgoodri:
I think you can ultimately avoid using `stan::math::integrate_1d` and just use `boost::math::quadrature::tanh_sinh`, in which case you could move the functions back to the scal directories if they do not all get flattened first.

martinmodrak:
Thanks for the feedback - I assume that there is no major known reason to not include this function in the library and it is therefore worthwhile for me to spend time working on it. I have tried to further improve the code.

In particular, I have added analytical derivative wrt. z, but the derivative wrt. v is still from autodiff. There is a trick I use to combine those in the `<var,var>` case which I am not sure is great, but it works:

```
  if(is_var<T_v>::value) {
    operands.push_back(value);
    gradients.push_back(1);
  }
  operands.push_back(z);
  gradients.push_back(gradient_dz);

  return precomputed_gradients(value_of(value), operands, gradients);
```
Where `value` contains the result of autodiffed computation when `v` is `var` (and is `double` when `v` is `double`).

> CRAN won't allow those `sprintf` things

I updated to snprintf - is that still an issue? Would you suggest using std::ostringstream instead? Or do you believe the improved messages are not worth the hassle?

> I think you can ultimately avoid using `stan::math::integrate_1d` and just use `boost::math::quadrature::tanh_sinh`, in which case you could move the functions back to the scal directories if they do not all get flattened first.

That's what I have done. I have also tried getting analytical derivative of the inner integral wrt. `v`, to avoid using nested autodiff, but I could not get it to compute stably. I now have a custom nested autodiff code for this which is just simplified `gradient_of_f`.

However, I use std::vector to be able to use precomputed_gradients which means the linter complains that the function should not be in `scal`  :-(  

> It is much better to test with identities

Currently looking into this

bob-carpenter:
> On Feb 19, 2019, at 10:58 AM, Martin Modrák <notifications@github.com> wrote:
> 
> Thanks for the feedback - I assume that there is no major known reason to not include this function in the library and it is therefore worthwhile for me to spend time working on it. I have tried to further improve the code.

I have no idea about utility.

But all the analytic derivatives are great for anything we do implement.

...

> I updated to snprintf - is that still an issue? Would you suggest using std::ostringstream instead? Or do you believe the improved messages are not worth the hassle?

You can't write to the standard streams.  Writing to a std::stringstream is fine.  But the only way our functions should create text output is through exceptions.

...

> However, I use std::vector to be able to use precomputed_gradients which means the linter complains that the function should not be in scal :-(

There may be a bug in the linter.  std::vector is OK to use in the scal directory inside functions.  What we don't want is to have functions that apply to std vectors defined in scal.  Those should go in arr.

syclik:
> I have no idea about utility. 
> 
> But all the analytic derivatives are great for anything we do implement.

I'll cosign. If you think it'll help you, then feel free to continue to implement. If it's written well, we can include it in the Math library for your use. ("well" is a very loosely defined term, but essentially, if there's very little maintenance burden for the future, then it's fine.)

> There may be a bug in the linter. std::vector is OK to use in the scal directory inside functions. What we don't want is to have functions that apply to std vectors defined in scal. Those should go in arr.

@bob-carpenter, that's not correct. The difference in `scal` vs `arr` was that `scal` does not include any `std::vector`. In retrospect, that's probably not really a useful distinction and we're trying to flatten the structure, but for the time being, it would have to be in `arr` since it depends on `std::vector`.

@martinmodrak, will you put a comment here when it's ready for review?

@bgoodri, would you be able to review once it's ready?

bob-carpenter:
I recall the decision going the other way, but it's no big deal as long as @syclik can lay out an acceptable workaround until the `scal` / `arr` / `mat` distinction is collapsed.  

@syclik:  The question is where the definition of `real foo(real)` should go if there is a`std::vector` in the implementation?  Is it OK to put it in `arr`?

syclik:
> @syclik <https://github.com/syclik>: The question is where the definition
of real foo(real) should go if there is astd::vector in the implementation?
Is it OK to put it in arr?

Yes, it should go in `arr`.

bgoodri:
Can the `std::vector` be avoided by using `var(new precomp_v_vari(...))` or `var(new precomp_vf_vari(...))`?

martinmodrak:
So it is not *completely* done, but I would like to ask for feedback and possibly code review (but I understand if you don't have time to help me solve my problems :-) )

Here is a plot of errors wrt. recursive formulas http://functions.wolfram.com/Bessel-TypeFunctions/BesselK/17/01/01/ (ratio of the log of the two sides should be 1 and derivative of the ratio should be 0 for all `v` and `z`)


image

It is not the most readable plot ever, but the bigger points (is_error) show where the error > 1e-7, which fails the test. 

Currently I combine the  approach of Rothwell (used for `v < 75` and for `v < 0.5 || log(z) > 300*(v-0.5)`, outside this the integral diverges) with asymptotic formulae for large `v` and large `z`.  As is visible on the plot, it mostly works, but there are two problematic regions:

1) large `v` and roughly equally large `z`. I can slightly narrow down the problematic region by using formula 5 from http://mathworld.wolfram.com/ModifiedBesselFunctionoftheSecondKind.html but the improvement is very small and the core problem still stands. 

2) `z < 1e-3` and `v < 1` especially for computing the gradient wrt. `v`. 

For 2), the problem is likely as much imprecision in the function as imprecision in the test code - for `v` in `[0,1]`, the recursive formula can only be computed as a difference of exponentials:
`log(K(v,z)) = log_diff_exp(log(K(v-2,z)), LOG_2 + log(-v + 1) - log(z) + log(K(v - 1, z)))`
and I am not sure how much I trust `log_diff_exp` to be accurate in those cases. Do you have any idea? This interpretation is boosted by the fact that when comparing to results from Mathematica, the problems with small `z` and `v` only remain for gradient wrt. `v` and even there, the relative error is at most `1e-3`. 

Do you happen to have any suggestions on how to progress further?

Finally, as per @bgoodri 's suggestion, I've used `precomp_v_vari` to avoid depending on vector and have happily moved to `scal`

bgoodri:
This is getting better. Now that the Boost 1.69 PR has been merged, you can avoid the autodiff when `v` is a var by using the [complex step](https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/) method to calculate the derivative of `log(besselK(v, x))` with respect to `v` that has a tiny imaginary component. There is an [example](https://www.boost.org/doc/libs/1_69_0/libs/math/doc/html/math_toolkit/double_exponential/de_tanh_sinh.html) at the bottom of doing an integral with complex numbers.

bgoodri:
There is another asymptotic expansion at the bottom of
https://en.wikipedia.org/wiki/Bessel_function#Asymptotic_forms
for arguments that are small in absolute terms and relative to the order.

martinmodrak:
Just an informative update if anyone's following:

**Complex step**
So it turns out that complex-stepping the integral has almost exactly the same error as the integral-of-the-nested-autodiff approach. This is weird - maybe the complex-step magic (which I totally don't understand) is actually somehow related to the way complex-step works (I can see how action on imaginary component might be similar to action on autodiff value)? As the complex-step has shorter code, I will keep it for now.

**Asymptotics**
The asymptotic formula for small arguments relative to order unfortunately does not work for very small orders. (the log transform probably messes it up) It does however improve for a few cases where both Rothwell and other asymptotics failed so I've kept it, but I am on the search for a better one.

bob-carpenter:
Yes, there are lots of deep connections between complex numbers and autodiff to the point where you can apparently compute derivatives if you have complex number arithmetic.  There was a lot of discussion on the forums under "complex number support".  I couldn't find a simple reference of how it's done or what the exact connection is.

bgoodri:
I would expect both autodiff and complex step to be about the same in terms of accuracy but autodiff requires memory allocations. Also, you can probably move more stuff to prim now.

To get more accuracy, you probably have to use `long double` or `complex<long double>`. For relatively small orders, there might be some sort of approximation based on a pertubation of K_0(z). Also, in principle, it might be possible to use `log_sum_exp` logic with quadrature since we are taking the logarithm of a sum that approximates an integral of a positive function. But we would have to figure out where the integrand reaches is maximum.

stan/math/prim/scal/meta/complex_step.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <limits>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+template <class F> 
+// Complex step differentiation as described at: 
+// https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/
+double complex_step(F f, double x) {
```
bgoodri: Boost already has a function for this
https://www.boost.org/doc/libs/1_69_0/libs/math/doc/html/math_toolkit/diff.html

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,343 @@
...
+    return value;
+  }
+};
+
+template <typename T_v, typename T_z>
+typename boost::math::tools::promote_args<T_v, T_z>::type compute_lead_rothwell(
+    const T_v &v, const T_z &z) {
+  typedef typename boost::math::tools::promote_args<T_v, T_z>::type T_Ret;
+
+  using boost::math::lgamma;
```
bgoodri: We are using `std::lgamma` these days

bgoodri:
@bob-carpenter It is explained at https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/ which is what @martinmodrak is using in the case the order is a `var`.

bgoodri:
@martinmodrak For small order, try the first series at
https://www.wolframalpha.com/input/?i=series+Log%5BBesselK%5Bv,+z%5D%5D+as+v+-%3E+0

bgoodri:
... which would seem to require derivatives wrt the order as it goes to zero. According to 
http://functions.wolfram.com/Bessel-TypeFunctions/BesselK/20/01/01/0003/
the first derivative is zero for all values of the argument, but I'm not sure about the higher derivatives.

martinmodrak:
@bgoodri Thanks for the hints. My previous comment however contained an error as the problem is with small argument (z), not order (sorry, I edited the comment now, but it is my fault). The series at https://www.wolframalpha.com/input/?i=series+Log%5BBesselK%5Bv,+z%5D%5D+as+z+-%3E+0 is much less promising as it is a log of sum where some terms are negative.

I am also convinced an important part of the problem is the way I compute my checks. Testing against Mathematica for small z gives relative errors for gradient wrt. `v` of at most 1e-4 and all other errors <1e-8. Also the errors on the plots above start to show up for v <= 1 which is where the method of evaluation changes. 

This change is necessary as for `v` in `[0,1]` the recursive formulas (http://functions.wolfram.com/Bessel-TypeFunctions/BesselK/17/01/01/) can no longer be evaluated with `log_sum_exp`, and I need `log_diff_exp` instead. I however cannot find formulas that would let me avoid `log_diff_exp` for evaluation.

martinmodrak:
Actually, I've just figured out a different formula for the checks and the error stops being inflated - thanks for rubber ducking!

Anyway, here is the current plot of errors - now the maximal relative error is 3.6e-02 for v = 5 and z = 1.1 and the most problematic region is for `v` and `z` comparable but not small. 

image

bgoodri:
I think you gotta try `long double` and `complex<long double>` internally.

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+  } else {
+    return ComputationType::Asymp_z;
+  }
+}
+
+////////////////////////////////////////////////////////////////
+//                    UTILITY FUNCTIONS                       //
+////////////////////////////////////////////////////////////////
+
+void check_params(const double &v, const double &z) {
```
SteveBronder: I think you can use our internal stuff instead of this

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+  using std::pow;
+
+  check_params(value_of(v), value_of(z));
+
+  if (z == 0) {
+    return std::numeric_limits<double>::infinity();
+  }
+
+  T_v v_ = fabs(v);
+  switch (choose_computation_type(value_of(v_), value_of(z))) {
```
SteveBronder: I prefer if statements instead of switch because to the fallthrough gotchas

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+    default: {
+      stan::math::domain_error("log_modified_bessel_second_kind_frac",
+                               "Invalid computation type ", 0, "");
+      return asymptotic_large_v(v_, z);
+    }
+  }
+}
+
+template <typename T_v>
+var log_modified_bessel_second_kind_frac(const T_v &v, const var &z) {
```
SteveBronder: I think if you clean up the above you won't need the template specialization

test/unit/math/rev/scal/fun/log_modified_bessel_second_kind_frac_test.cpp
```diff
@@ -0,0 +1,504 @@
+#include <stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp>
+#include <stan/math/rev/scal/fun/modified_bessel_second_kind.hpp>
+#include <stan/math/rev/core/set_zero_all_adjoints.hpp>
+#include <gtest/gtest.h>
```
SteveBronder: Move gtest below stan includes

SteveBronder (changes requested):
Sorry this took so long to review and thanks for the contribution!

I only had time to look over the code tonight. Docs would make this a bit easier to review.

martinmodrak:
Thanks very much @SteveBronder for going through the code. Most of the suggestions are very sensible and once I resume working on this, I will implement them and will explain where I differ in opinion. 

Unfortunately however, the feature is far from complete as the function still has excessive errors for some parameters and does not pass tests (see the previous post: https://github.com/stan-dev/math/pull/1121#issuecomment-479034814 for some details). I am unsure about the timeline of resolving this as it seems to require some additional math insights (which I do not have and don't have more ideas where to look) and also because the project that drove this feature is now on the back burner as we encountered some other problems. 

Should have probably made the status of PR more prominent, sorry if the feature does not make it in the end and your effort ends up not improving Stan codebase.

SteveBronder:
No worries man! if this is still a WIP then I'm going to close the PR for now

bgoodri:
@martinmodrak Did you try equation (24) of Rothwell in the remaining problematic region? Rothwell says equation (24) is "difficult to compute when |z| is small", but I think it might be accurate enough when `z` is not small but close to `v`. For z = 11 and v = 50, I get function values of
`58.0217473497870557` from the log of equation(24) with long doubles
`58.02174734978707484` from WolframAlpha
and complex-step derivatives with respect to `v` of
`2.20960269175438162` from log of equation(24) with long doubles
`2.209602691754382164` from WolframAlpha

Here is my Rcpp version for the log of equation (24) in Rothwell:
```
// [[Rcpp::plugins(cpp14)]]
// [[Rcpp::depends(BH)]]
#include <Rcpp.h>
#include <boost/math/quadrature/exp_sinh.hpp>
#include <boost/math/special_functions/digamma.hpp>

// [[Rcpp::export]]
std::complex<double> log_besselK(double x, double nu, double h = 1e-9) {
  using std::exp;
  using std::log;
  using std::log1p;
  
  const std::complex<long double> vmhalf(nu - 0.5, h);

  boost::math::quadrature::exp_sinh<long double> integrator;
  auto log_Q = log(integrator.integrate([&](const long double t) {
    return exp(-t +  vmhalf * (log(t) + log1p(0.5 * t / x)));
  }, std::sqrt(std::numeric_limits<long double>::epsilon()));
  std::complex<long double> log_K(0.5 * log(0.5 * M_PI / x) 
                                  - x - std::lgamma(nu + 0.5) + log_Q.real(),
                                  - h * boost::math::digamma(nu + 0.5) 
                                  + log_Q.imag());
  return static_cast<std::complex<double> >(log_K);
}
```

martinmodrak:
@bgoodri Thanks for the hint. I actually tried the formula given at the question https://math.stackexchange.com/questions/1960778/approximating-the-log-of-the-modified-bessel-function-of-the-second-kind (the answer is mine), which helps in a bit of the parameter space but does not work toward the larger values. The formula 24 in Rothwell is actually a bit less stable than that: I haven't found parameter values for which the Rothwell 24 integral does not blow up while the "gamma integral" from the SE question does blow up. 

One thing that might help a lot would be a reimplementation of the exph_sinh integrator using the log_sum_exp logic, which Bob hinted is a possibility, allowing me to stay on the log scale. (for the problematic parameters the maximum log of the integrand can crawl over several hundred)

Here is the current state of affairs.
image

bgoodri:
Here is a version of equation (24) from Rothwell with a log-sum-exp thing, which I would guess works better when `v` is large.
```
// [[Rcpp::plugins(cpp14)]]
// [[Rcpp::depends(BH)]]
#include <Rcpp.h>
#include <boost/math/quadrature/tanh_sinh.hpp>
#include <boost/math/quadrature/exp_sinh.hpp>
#include <boost/math/special_functions/digamma.hpp>

std::complex<double> log_besselK(double x, double nu, double h = 1e-9) {
  using std::exp;
  using std::log;
  using std::log1p;
  
  const std::complex<long double> v(nu, h);
  const std::complex<long double> half(0.5);
  const std::complex<long double> x_(x);
  auto maximum = std::sqrt(v * v - v + x_ * x_ + half * half) + v - x_ - half;

  auto vmhalf = v - half;
  boost::math::quadrature::exp_sinh<long double> integrator;
  auto log_Q = maximum.real() > 0 
             ? maximum + log(integrator.integrate([&](const long double t) {
                 auto t_max = t + maximum;
                 return exp(-t_max + vmhalf * (log(t) + log1p(0.5 * t / x)));
               }, std::sqrt(std::numeric_limits<long double>::epsilon())))
             : log(integrator.integrate([&](const long double t) {
                 return exp(-t + vmhalf * (log(t) + log1p(0.5 * t / x)));
             }, std::sqrt(std::numeric_limits<long double>::epsilon())));
  std::complex<long double> log_K(0.5 * log(0.5 * M_PI / x)
                                  - x - std::lgamma(nu + 0.5) + log_Q.real(),
                                  - h * boost::math::digamma(nu + 0.5)
                                  + log_Q.imag());
  return static_cast<std::complex<double> >(log_K);
}

```

martinmodrak:
Thanks. However, this is almost exactly the way I implemented it at: https://github.com/martinmodrak/math/commit/8fe7fe528ae96539b1cde59f2f2929900488f29a
The reason is AFAIK that the maximum achieved by the integrand in Equation 24 ( `std::sqrt(v * v - v + x_ * x_ + half * half) + v - x_ - half;`) turns out to be almost always larger than the maximum achieved by the integrand from the SE ` (std::sqrt(v_ * v_ - 2*v_ + z*z + 1) + v_ - 1) / z`.

martinmodrak:
Thanks @bgoodri , moving the maximum out helped a lot. This is the current error plot (and I think I can still tweak it a bit, so maybe I am almost there)

image

bgoodri:
Great, although I am slightly confused as to how it would help a lot if it
was similar to what you already had. But I'll take it. Are you still using
doubles? Maybe switching to long double would get us home?

bgoodri:
Also, I have not tried this yet but

https://www.boost.org/doc/libs/1_71_0/libs/math/doc/html/math_toolkit/trapezoidal.html

which cites Table 17.1 of

https://epubs.siam.org/doi/pdf/10.1137/130932132

claims that the trapezoid rule is good for evaluating complex Bessel
functions.

bgoodri:
and https://arxiv.org/pdf/1209.1547.pdf

martinmodrak:
Thanks for the help, I am finally moving forward. Just to report current status: It turns out the trapezoid rule (on the log scale, via `log_sum_exp`) works great for `v ~= z`, but only for the actual value. Autodiffing through the trapezoid rule yields bad values for d/dv. 

image

But d/dv was computed OK with the previous algorithms, so my next plan is to combine the value from the trapezoid rule with derivatives from the asymptotic expressions. After that, the derivatives for very small v and z are still bad (error around 1e-3), which I still don't know how to tackle (the other integrals, including the trapezoid give even worse results than Rothwell there).

bgoodri:
I think the trapezoid rule with a log_sum_exp of equation 1.1 in
https://arxiv.org/pdf/1209.1547.pdf would work really well, except that the
maximum of the integrand apparently cannot be found analytically. So, I
guess we would have to use Boost's root-finding methods to find the maximum
and then subtract it.

bgoodri:
Here is a rough version of equation 1.1 with root finding to get the mode and then use it with log-sum-exp logic. Unfortunately, Boost only added Newton root-finding of complex functions in version 1.70 so we can't (easily) use the complex step method to get the sensitivity to `v`.

```
// [[Rcpp::plugins(cpp14)]]
// [[Rcpp::depends(BH)]]
#include <Rcpp.h>
#include <boost/math/quadrature/exp_sinh.hpp>
#include <boost/math/tools/roots.hpp>
#include <tuple> // for std::tuple and std::make_tuple.


// [[Rcpp::export]]
double log_besselK(double x, double nu) {
  using std::exp;
  using std::log;
  using std::log1p;
  using std::cosh;
  using std::sinh;
  using std::tanh;
  
  double ub = 1.0;
  
  while ( -x * sinh(ub) + nu * tanh(ub * nu) > 0 ) {
    ub *= 2.0;
  }
  double guess = 0.75 * ub;
  
  boost::uintmax_t maxit = 20;
  using boost::math::tools::newton_raphson_iterate;
  
  auto maximum = newton_raphson_iterate(
    [&](double t){
      auto sinh_t = sinh(t);
      auto v_t = nu * t;
      auto tanh_vt = tanh(v_t);
      auto cosh_vt = cosh(v_t);
      auto x_sinh_t = x * sinh_t;
      auto value = nu * tanh_vt - x_sinh_t;
      auto ratio = nu / cosh_vt;
      auto ratio2 = ratio * ratio;
      auto D1 = ratio2 - x * cosh(t);
      return std::make_tuple(value, D1);
    }, guess, ub > 1.0 ? 0.5 * ub : 0, ub, 8, maxit);
  
  double error_estimate;
  double L1;

  boost::math::quadrature::exp_sinh<double> integrator;
  double termination = sqrt(std::numeric_limits<double>::epsilon());
  
  auto log_K = maximum + log(integrator.integrate(
    [&](double t) {
      auto nut = nu * t;
      // log(cosh(x)) == x + log1p(exp(-2 * x)) - log(2)
      return exp(-maximum + nut + log1p(exp(-2.0 * nut)) 
                 - log(2.0) - x * cosh(t));
    }, termination, &error_estimate, &L1));

  std::cout << "error_estimate = " << error_estimate << std::endl 
            << "L1 = " << L1 << std::endl
            << "maximum = " << maximum << std::endl;
  return log_K;
}

```

bgoodri:
Sorry. I was doing all this Newton stuff to get the value of `t` that maximizes the integrand but then shifting by that argmax instead of shifting by the maximal log-value. This version should actually be good (for the value; we still have to deal with the autodiff) because when shifted properly, there is relatively little area under the curve:

```
// [[Rcpp::plugins(cpp14)]]
// [[Rcpp::depends(BH)]]
#include <Rcpp.h>
#include <boost/math/quadrature/exp_sinh.hpp>
#include <boost/math/tools/roots.hpp>
#include <tuple> // for std::tuple and std::make_tuple.

// [[Rcpp::export]]
double log_besselK(double x, double nu) {
  using std::exp;
  using std::log;
  using std::log1p;
  using std::cosh;
  using std::sinh;
  using std::tanh;
  
  double ub = 1.0;
  
  while ( -x * sinh(ub) + nu * tanh(ub * nu) > 0 ) {
    ub *= 2.0;
  }
  double guess = 0.75 * ub;
  
  boost::uintmax_t maxit = 20;
  using boost::math::tools::newton_raphson_iterate;
  
  auto arg_max = newton_raphson_iterate(
    [&](double t){
      auto sinh_t = sinh(t);
      auto v_t = nu * t;
      auto tanh_vt = tanh(v_t);
      auto cosh_vt = cosh(v_t);
      auto x_sinh_t = x * sinh_t;
      auto value = nu * tanh_vt - x_sinh_t;
      auto ratio = nu / cosh_vt;
      auto ratio2 = ratio * ratio;
      auto D1 = ratio2 - x * cosh(t);
      return std::make_tuple(value, D1);
    }, guess, ub > 1.0 ? 0.5 * ub : 0, ub, 8, maxit);
  
  const double log2 = log(2.0);
  // log(cosh(x)) == x + log1p(exp(-2 * x)) - log(2)
  auto pivot = nu * arg_max + log1p(exp(-2.0 * nu * arg_max)) - log2
             - x * cosh(arg_max);
  
  double error_estimate;
  double L1;

  boost::math::quadrature::exp_sinh<double> integrator;
  double termination = sqrt(std::numeric_limits<double>::epsilon());
  
  auto log_K = pivot + log(integrator.integrate(
    [&](double t) {
      auto nut = nu * t;
      return exp(-pivot + nut + log1p(exp(-2.0 * nut)) - log2 
                 - x * cosh(t));
    }, termination, &error_estimate, &L1));

  std::cout << "error_estimate = " << error_estimate << std::endl 
            << "L1 = " << L1 << std::endl
            << "arg_max = " << arg_max << std::endl
            << "pivot = " << pivot << std::endl;
  return log_K;
}

```

jacob-hjortlund:
Don't have much to add, just wanted to express my gratitude for you guys working on this problem, as I've been having quite a bit of trouble tackling the asymptotic behavior of the modified Bessel function with nowhere near the knowledge you guys have. Thanks and keep up the great work @martinmodrak  @bgoodri !

martinmodrak:
I am not sure finding the maximum is important for offsetting the `log_sum_exp` - everywhere I look, the maximum is at most 1, usually much lower. I however use the maximum to find the step size to integrate with. Luckily, for large `v*x > 17`, you get `log(cosh(v*x)) ~= abs(v*x) - log(2)`. With this substitution you can [analytically find](https://www.wolframalpha.com/input/?i=solve+for+x%3Ad%2Fdx+%7Cv+*+x%7C+%2B+log%282%29+-+z+*+cosh%28x%29+%3D%3D+0%3B+v+%3E0%3B+z+%3E+0%3B+x+%3E0) the arg max of the integrand as `asinh(v/z)`. 

For large `v`, the integrand decays very quickly past the maximum. 
The behaviour at low `v` and `z` seems to be more tricky, and I have yet to figure out if the method would be useful there. 

Here's the code I used to create the latest plot:

```
template <typename T>
T logcosh(const T& x) {
  //Taken from the limma R package, maybe I should check it works as good as I
  //expect it to?
  //Credited in limma as 	Gordon Smyth  8 April 2007
  T x_ = abs(x);
  if(x_ < 1e-4) {
    return 0.5 * x * x;
  } else if(x_ < 17) {
    return log(cosh(x));
  } else {
    return x_ - std::log(2);
  }
}

template <typename T_v>
T_v trapezoid_cosh(const T_v &v, const double &z) {
  const double& approximate_maximum = std::asinh(value_of(v) / z);
  const int max_steps = 500;
  const double& h = approximate_maximum / (0.5 * max_steps);
  std::vector<T_v> terms;
  terms.reserve(max_steps);
  for(int n = 0; n < max_steps; n++) {
    const double x = n * h;
    const T_v last_term = logcosh(v * x) - z * cosh(x);
    terms.push_back(last_term);
    //TODO create some sensible stopping criterion
  }
  return log_sum_exp(terms) + std::log(h);
}
```

Also, it seems like we are doubling some work here - I keep almost up-to-date code at https://github.com/martinmodrak/math/blob/feature/issue-1112-continuous-besselk/stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp for you to check out. Maybe I should open a new PR / reopen this one, so that my commits are shown in context? Also - if you have the time - it might be sensible for you to try out your implementations with the tests I've developed...

Finally, I've had mostly unsatisfying experiences with complex step - on the (two) occasions I used it, it produced almost exactly the same results as Stan's autodiff and was slower to compute. Do you have some specific reason to believe it would help here?

@jacob-hjortlund I don't think I _have_ knowledge. Before I started working on this, I didn't even know what Bessel functions are. I am learning it all as I go, and it hurts :-) Also I wouldn't get anywhere near where I am without the Just wanted to let you know this, before you assume my

martinmodrak:
Just tried to include your implementation using `exp_sinh` to integrate and it runs into problems for values like `v = nu = 37634, z = x = 4236` where the logarithm of the value swings between -4236 and +70535 so no amount of pivoting can help. However, using your implementation of `logcosh(x) = x + log1p_exp(-2 * x) - log(2)` makes the stupid trapezoid rule also resolve the derivatives wrt. `v`.  And some of the problems can be ameliorated by increasing resoluion. With:

```
template <typename T>
T logcosh(const T& x) {
  //By bgoodri
  return x + log1p_exp(-2 * x) - log(2);
}

template <typename T_v>
T_v trapezoid_cosh(const T_v &v, const double &z) {
  const double& approximate_maximum = std::asinh(value_of(v) / z);
  const int max_steps = 5000;
  const double& h = approximate_maximum / (0.1 * max_steps);
  std::vector<T_v> terms;
  terms.reserve(max_steps);
  for(int n = 0; n < max_steps; n++) {
    const double x = n * h;
    const T_v last_term = logcosh(v * x) - z * cosh(x);
    terms.push_back(last_term);
    //TODO create some sensible stopping criterion
  }
  return log_sum_exp(terms) + std::log(h);
}
```

I get:
image

Leaving the Rothwell for small `v` and `z` the biggest problem. However, with `max_steps = 5000` this is heavily inefficient and I need to find good ways to determine `h` and a stopping criterion.

bgoodri:
The trapezoid rule function in Boost

https://www.boost.org/doc/libs/1_71_0/libs/math/doc/html/math_toolkit/trapezoidal.html

has the adaptive stopping criterion but requires a finite upper bound to integrate to. I suppose we could use the largest floating point number, but maybe there is a way to use something smaller.

martinmodrak:
Thanks, I'll look into it. Note that I can't use the boost implementation directly, because I need to keep the `log_sum_exp` logic. But I might be able to reimplement it. Also sorry for bothering on the weekend - you are in no way obliged to respond at this time :-)

bgoodri:
No worries; it is a holiday here. You can utilize the log-sum-exp logic without a fixed number of max steps, as I did in 

https://github.com/stan-dev/math/blob/develop/stan/math/prim/scal/fun/log_modified_bessel_first_kind.hpp#L208

which in this case would be something like

```
template <typename T_v>
T_v trapezoid_cosh(const T_v &v, const double &z) {
  using stan::math::value_of;
  using stan::math::log_sum_exp;
  
  const double& approximate_maximum = std::asinh(value_of(v) / z);
  const double& h = approximate_maximum / (0.1 * 5000);
  T_v s = 0.0;
  T_v old_s = 1.0;
  int n = 0;
  while (s > old_s || old_s > s) {
    const double x = n * h;
    const double last_term = logcosh(v * x) - z * cosh(x);
    old_s = s;
    s = log_sum_exp(s, last_term);
    n++;
  }
  return s + std::log(h);
}
```
But it is not super-accurate yet. Now, we need a better procedure to choose `h`.

bgoodri:
OK, I have some ideas, but need to eat before I attempt to implement the rest of them. 

1. I think it would be best if we find the actual `arg_max` of the integrand with Newton-Raphson but your approximation is probably a better starting point than the starting point I was using.
2. The `while` loop will terminate quicker if we initialize `s = logcosh(v * arg_max) - z * cosh(arg_max), then set `n = 1`, and then evaluate the while loop with both `x = arg_max + n * h` and `x = arg_max - n * h`. Although then you have to return `s + std::log(0.5 * h)`.
3. The `h` should be set based on the curvature of the integrand at `arg_max` or maybe we can set `h` adaptively based on the curvature at the last point we evaluated at.
3.

bgoodri:
I am still having trouble when the order is `37634` and the argument is `4236`? WolframAlpha says it is "undefined"

https://www.wolframalpha.com/input/?i=Log%5BBesselK%5B37634%2C+4236.0%5D%5D

but arb says it is `70531.3573385249910109`, which is fairly consistent with the asymptotic expansion for large order and fixed argument at

https://dlmf.nist.gov/10.41.E2

which evaluates to `0.5 * log(0.5 * pi / v) - v * (1 + log(z) - log(2 * v))` which is about `70650.37`. With some more tricks I can get `70522.99`, which I suppose is ok in the relative error sense, but I think we're going to have to wait another week for Boost 1.72 to get released.

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+// Combining multiple approaches, documented at relevant sections
+// Due to symmetry around v, the code assumes all v's positive except for
+// top-level calls which pass the absolute value.
+
+namespace stan {
+namespace math {
+
+template <typename _Tp>
+inline int is_inf(const std::complex<_Tp> &c) {
```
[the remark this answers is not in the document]
martinmodrak: I need to only have this branch for infinite (not NaN), will add `is_scal_infinite`

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+    using std::pow;
+
+    auto v_mhalf = v - 0.5;
+    auto neg2v_m1 = -2.0 * v - 1.0;
+    auto beta = 16.0 / (2.0 * v + 1.0);
+
+    T_Ret value;
+    T_Ret uB = pow(u, beta);
+    T_Ret first
+        = beta * exp(-uB) * pow(2.0 * z + uB, v_mhalf) * boost::math::pow<7>(u);
```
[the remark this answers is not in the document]
martinmodrak: I assumed it is better when the exponent is known at compile time, but no strong opinion on this.

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+    T_Ret second = exp(-1.0 / u);
+    if (non_zero(second)) {
+      //    if (abs(second) > 0) {
+      second = second * pow(u, neg2v_m1);
+      if (is_inf(second)) {
+        second = exp(-1.0 / u + neg2v_m1 * log(u));
+      }
+      second = second * pow(2.0 * z * u + 1.0, v_mhalf);
+    }
+    value = first + second;
```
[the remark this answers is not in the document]
martinmodrak: Not sure. Those are just two terms of the expression and I don't think they have any particular meaning.

stan/math/rev/scal/fun/log_modified_bessel_second_kind_frac.hpp
```diff
@@ -0,0 +1,332 @@
...
+    typedef T_u value_type;
+    value_type tolerance
+        = std::sqrt(std::numeric_limits<value_type>::epsilon());
+
+    value_type error;
+    value_type L1;
+    size_t levels;
+
+    boost::math::quadrature::tanh_sinh<value_type> integrator;
+    T_Ret value = integrator.integrate(*this, 0.0, 1.0, tolerance, &error, &L1,
```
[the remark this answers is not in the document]
martinmodrak: Will check, but when I wrote this, I needed slightly finer control over what is happening... Maybe integrate_1d changed during the time.

bgoodri:
I think we can get a pretty good version of log(besselK) with the new Boost. I just need to pull some things together.

bbbales2:
@martinmodrak, @bgoodri just scrolling through old pulls and found this. Ping!

bgoodri:
@martinmodrak I recently read at https://functions.wolfram.com/Bessel-TypeFunctions/BesselK/17/02/01/ that

`besselK(v, z) = z / (2 * v) * [besselK(v + 1, z) - besselK(v - 1, z)]`

so we can avoid the numerically problematic case where `v < 1/2` by computing

`log(z) - log(2 * v) + log_diff_exp(log_besselK(v + 1, z), log_besselK(fabs(v - 1), z))`

although that might incur too much cancellation if `v` is too close to zero. That looks like twice as many calls to achieve more numerical stability, but we have to compute `log_besselK(v + 1, z)` anyway in order to take the derivative wrt `z`.

Do you want to re-open this PR or open a similar one? I think numerically stable logic would be like

1. If `v` is very close to zero, do an asymptotic approximation
2. Otherwise, if `v < 1/2`, utilize the above identity to shift the orders to both be greater than `1/2`
3. If `v` is very large, log an asymptotic approximation (possibly based on a Laplace approximation)
4. If `z` is very large, log an asymptotic approximation (possibly based on a Laplace approximation)
5. If `z` is much smaller than `sqrt(1 + v)`, log an asymptotic approximation
6. Otherwise, log Rothwell's integral using the `log_sum_exp` logic to pull out the peak, also splitting Rothwell's integral into two parts at the mode in case all the area is in the neighborhood of the mode

There might be another case that we have to deal with when `z` and `v` are similar but I never understood what drove the numerical inaccuracy in that situation.

bgoodri:
Upon further review, I think that an approach based on the logarithm of

https://functions.wolfram.com/Bessel-TypeFunctions/BesselK/07/01/01/0006/MainEq1.L.gif

may work well. All the overflow / underflow stuff is attributable to the x^v in the denominator, which is well-handled by logarithms. The integral is oscillatory, but Boost has a customized function for integrals like that

https://www.boost.org/doc/libs/1_76_0/libs/math/doc/html/math_toolkit/fourier_integrals.html

that is said to have rapid convergence when the poles of the `f` function are not too close to the real axis. The underlying Ooura and Mori (1999) paper has a fast-converging example in equation 4.13 that is similar to our case in that the poles are at plus or minus the imaginary unit.

In other words, the main logic of a Stan Math implementation could look like this simple Rcpp one:

```
// [[Rcpp::plugins(cpp17)]]
// [[Rcpp::depends(BH)]]
#include <Rcpp.h>
#define BOOST_MATH_INSTRUMENT_OOURA
#include <boost/math/quadrature/ooura_fourier_integrals.hpp>
#include <iostream>

// [[Rcpp::export]]
double log_besselK(const double x, const double nu) {
  
  using std::pow;
  using std::log;
  using std::lgamma;
  
  auto integrator = boost::math::quadrature::ooura_fourier_cos<double>();  
  
  auto f = [&nu](double t) {
    return pow(t * t + 1, -nu - 0.5);
  };

  auto [result, relative_error] = integrator.integrate(f, x);
  std::cout << "Integral = " << result << 
    ", relative error estimate " << relative_error << std::endl;
  
  return nu * (log(2.0) - log(x)) - 0.5 * log(M_PI) + lgamma(nu + 0.5) 
         + log(result);
}
```
It does not seem to be very accurate for large `x` and small `v`, but we could deal with that case separately.

bgoodri:
Leaving this here

https://arxiv.org/abs/2201.00090

which sounds like a good approach too

spinkney:
@bgoodri here's my attempt at coding up the algorithm in R

https://gist.github.com/spinkney/56d8b716248dae8545e5049f4dec33b1

jaburgoyne:
@bgoodri : Commenting here on the advice of the 'Contributing to Stan' workshop. I'm at StanCon and understand that you also would really love to have this function. Let's try to find each other. Perhaps the new-ish hypergeometric functions have opened some new ways forward?

bgoodri:
@jaburgoyne I am unfortunately not at StanCon this year, but there are plenty of people who could help you get started if you had a way to implement this function. Actually, the problem is not a lack of a plan, but that there are several approaches and none of them are universally better than all the others throughout the parameter space so we have to branch in a bunch of places. The same was true with Boost, Arb, etc. when they implemented special functions like the hypergeometric.

Issue #1126, opened 2019-02-25
Replace boost:: type traits with std:: versions


rok-cesnovar:
## Description

We should move the type traits from boost to the std library versions.

Type traits that could be replaced are:

- [x] [is_same](https://en.cppreference.com/w/cpp/types/is_same)
- [x] enable_if/disable_if could both be replaced by [std::enable_if
](https://en.cppreference.com/w/cpp/types/enable_if)
- [x] [is_unsigned](https://en.cppreference.com/w/cpp/types/is_unsigned)
- [x] [is_convertible](https://en.cppreference.com/w/cpp/types/is_convertible)
- [x] [is_arithmetic](https://en.cppreference.com/w/cpp/types/is_arithmetic)

## Example

```
template <typename T>
struct is_constant {
  /**
   * A boolean constant with equal to <code>true</code> if the
   * type parameter <code>T</code> is a mathematical constant.
   */
  enum { value = boost::is_convertible<T, double>::value };
};
```

should be changed to 

```
template <typename T>
struct is_constant {
  /**
   * A boolean constant with equal to <code>true</code> if the
   * type parameter <code>T</code> is a mathematical constant.
   */
  enum { value = std::is_convertible<T, double>::value };
};
```

## Expected Output

Replaced type traits.

#### Current Version:
v2.18.1

rok-cesnovar:
There are also some places where the entire `<boost/type_traits.hpp>` header is included, which should be investigated.

rok-cesnovar:
The first PR is open, if it goes through I will open the next PR for the `is_same` replacement, which is already prepared in [this](https://github.com/bstatcomp/math/tree/cleanup_boost_is_same) branch. enable_if replacement will be last.

PR #1129, opened 2019-02-27
Replace boost::enable_if_c, boost::enable_if and boost:disable_if with std::enable_if

rok-cesnovar:
## Summary

This PR is the final sequel of my boost cleanup trilogy. The changes are simple, there are 3 different cases: 

- replace `boost::enable_if_c` directly with  [std::enable_if](https://en.cppreference.com/w/cpp/types/enable_if). 

- replace `boost::enable_if` with `std::enable_if` and add an additional `::value` to the first parameter of the type trait

- replace `boost::disable_if` with `std::enable_if`, add an additional `::value` to the first parameter of the type trait and a negation

## Tests

There are no new tests added.

## Side Effects

/

## Checklist

- [x] Math issue #1126 

### DIFF
--- a/make/test-math-dependencies
+++ b/make/test-math-dependencies
@@ -1,3 +1,27 @@
+## 
+#  Function to search code for warnings.
+#
+#  After all the warnings are found, use
+#  print_math_warnings to print the warnings.
+#
+#  The output of the warnings look like cpplint warnings.
+#  It displays the filename, the line number, the message,
+#  and the type of error message.
+#
+#  Example:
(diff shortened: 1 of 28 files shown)


rok-cesnovar:
This is ready for review.

syclik:
@rok-cesnovar, awesome!

I pushed an update to `make/test-math-dependencies` so no one can revert the changes you've made! (Rules are only as good as can be enforced.)

If you want to verify, just try sneaking in a use of one of boost functions in and you'll find that running `make test-math-dependnencies` will fail.

Want to take a look at the change I've made just to make sure I've put in enough rules?

rok-cesnovar:
Looks good. I will merge once this passes.

Issue #50, opened 2015-07-06
error_handling: expose individual check functions in the language


syclik:
_From @syclik on October 26, 2014 6:44_

We want simple functions that can be exposed through the language. 

_Copied from original issue: stan-dev/stan#1104_

syclik:
Things that need to happen:

1. Create a new, parallel file to the `check_*` file called `is_*`.
2. Add a new, parallel file to the test file.
3. Add tests for the new function.
4. Write the new function.
5. Change the implementation of the check.

Later down the road, we'll need to reopen stan#1104 and expose all the functions. And add it to the doc.

seantalts:
Hey, Bob was suggesting I take a crack at this and I have a few questions about what we want to do here. First, my attempt to paraphrase the high level goal: We want to allow Stan users to use a standard set of error checking functions for use typically in their Stan code's `functions` block (i.e. User Defined Functions), and we think the functions listed above on this ticket are good candidates for end use. I'm a bit lost as to why we want to create copies of the checks with slightly different names (the `check_*` -> `is_*` step above)? Also, Bob mentioned that we wanted to do some clean up and vectorization here, and I'm wondering what the clean up especially should look like. @bob-carpenter might be the best person to answer at least parts of this.

bob-carpenter:
The behavior's different.  These will return true if the tests pass, but not throw errors.

```
bool is_positive(real x);
bool is_positive(int x);
```

These will throw errors (equivalent to the reject() statement in Stan) of the tests fail:

```
void validate_positive(real x);
void validate_positive(int x);
```

And they should be usable as statements.  I think that'll work as is.

We want both because the user may patch together the is_X tests and return a custom message, whereas the error messages in the validate_X forms will be predetermined.

seantalts:
Just to be clear, are we expecting things like `is_positive` to be vectorized, or just the `check_positive` version? Given that `bool is_positive(...)` couldn't return information about where the error occurred, I think the latter makes more sense but just want to run that by you guys while I'm working on it.

bob-carpenter:
I was originally thinking yes, but you explained why that's not going to make sense in practice, so no.  Maybe we don't even need to expose boolean versions, since `is_positive(x)` is just `x > 0`.  So maybe it only makes sense to implement the `check_X` functions.  Definitely starting with those would make sense and we can see if we ever need the other ones later and do them in another pull request.

oliverdechant:
[...]

PR #1130, opened 2019-02-28
Issue #50: boolean functions - 1

increasechief:
## Summary

Adds test functions with boolean return statements rather then calling `domain_error` like existing `check_*` functions.

https://github.com/stan-dev/math/issues/50

## Tests

Each function included has its own test file modeled off the existing tests used for the `check` functions.

## Side Effects

None applicable.

## Checklist

- [x] Math issue #(50)

- [x] Copyright holder: (Original source)

### DIFF
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -9,9 +9,6 @@ namespace math {
 /**
  * Return <code>true</code> if the variable sizes match.
  *
- * This function checks the runtime sizes for
- * variables that implement a <code>size()</code> method.
- *
  * @tparam T_y1 Type of the first variable, requires class method
  *    <code>.size()</code>
  * @tparam T_y2 Type of the second variable, requires class method
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -13,8 +13,10 @@ namespace math {
  * variables that implement a <code>size()</code>
  * method.
  *
- * @tparam T_y1 Type of the first variable
- * @tparam T_y2 Type of the second variable
+ * @tparam T_y1 Type of the first variable, requires function 
+ *    access to <code>.size()</code>
+ * @tparam T_y2 Type of the second variable, requires function 
+ *    access to <code>.size()</code>
  *
  * @param y1 First variable
  * @param y2 Second variable
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -26,6 +26,6 @@ inline bool is_matching_size(const T_y1& y1, const T_y2& y2) {
   return is_size_match(y1.size(), y2.size());
 }
 
-} // namespace math
-} // namespace stan
+}  // namespace math
+}  // namespace stan
 #endif
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -7,7 +7,7 @@ namespace stan {
 namespace math {
 
 /**
- * Check if two structures are the same size.
+ * @return <code>true</code> if the variable sizes match
  *
  * This function checks the runtime sizes for
  * variables that implement a <code>size()</code>
@@ -18,8 +18,6 @@ namespace math {
  *
  * @param y1 First variable
  * @param y2 Second variable
- *
- * @return <code>true</code> if the variable sizes match
  */
 template <typename T_y1, typename T_y2>
 inline bool is_matching_size(const T_y1& y1, const T_y2& y2) {
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -13,9 +13,9 @@ namespace math {
  * variables that implement a <code>size()</code>
  * method.
  *
- * @tparam T_y1 Type of the first variable, requires function 
+ * @tparam T_y1 Type of the first variable, requires function
  *    access to <code>.size()</code>
- * @tparam T_y2 Type of the second variable, requires function 
+ * @tparam T_y2 Type of the second variable, requires function
  *    access to <code>.size()</code>
  *
  * @param y1 First variable
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -7,19 +7,20 @@ namespace stan {
 namespace math {
 
 /**
- * @return <code>true</code> if the variable sizes match
+ * Return <code>true</code> if the variable sizes match.
  *
  * This function checks the runtime sizes for
- * variables that implement a <code>size()</code>
- * method.
+ * variables that implement a <code>size()</code> method.
  *
- * @tparam T_y1 Type of the first variable, requires function
- *    access to <code>.size()</code>
- * @tparam T_y2 Type of the second variable, requires function
- *    access to <code>.size()</code>
+ * @tparam T_y1 Type of the first variable, requires class method
+ *    <code>.size()</code>
+ * @tparam T_y2 Type of the second variable, requires class method
+ *    <code>.size()</code>
  *
  * @param y1 First variable
  * @param y2 Second variable
+ *
+ * @return <code>true</code> if the variable sizes match
  */
 template <typename T_y1, typename T_y2>
 inline bool is_matching_size(const T_y1& y1, const T_y2& y2) {
--- a/stan/math/prim/arr/err/is_matching_size.hpp
+++ b/stan/math/prim/arr/err/is_matching_size.hpp
@@ -1,0 +1,31 @@
+#ifndef STAN_MATH_PRIM_ARR_ERR_IS_MATCHING_SIZE_HPP
+#define STAN_MATH_PRIM_ARR_ERR_IS_MATCHING_SIZE_HPP
+
+#include <stan/math/prim/scal/err/is_size_match.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Check if two structures are the same size.
+ *
+ * This function checks the runtime sizes for
+ * variables that implement a <code>size()</code>
+ * method.
+ *
+ * @tparam T_y1 Type of the first variable
+ * @tparam T_y2 Type of the second variable
+ *
+ * @param y1 First variable
+ * @param y2 Second variable
+ *
+ * @return <code>true</code> if the variable sizes match
+ */
+template <typename T_y1, typename T_y2>
+inline bool is_matching_size(const T_y1& y1, const T_y2& y2) {
+  return is_size_match(y1.size(), y2.size());
+}
+
+} // namespace math
+} // namespace stan
+#endif
--- a/stan/math/prim/arr/err/is_nonzero_size.hpp
+++ b/stan/math/prim/arr/err/is_nonzero_size.hpp
@@ -5,12 +5,13 @@ namespace stan {
 namespace math {
 
 /**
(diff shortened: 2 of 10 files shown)


stan/math/prim/arr/err/is_matching_size.hpp
```diff
@@ -0,0 +1,31 @@
+#ifndef STAN_MATH_PRIM_ARR_ERR_IS_MATCHING_SIZE_HPP
+#define STAN_MATH_PRIM_ARR_ERR_IS_MATCHING_SIZE_HPP
+
+#include <stan/math/prim/scal/err/is_size_match.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Check if two structures are the same size.
```
syclik: Let's update the documentation here. For our guidelines, see https://github.com/stan-dev/stan/wiki/How-to-Write-Doxygen-Doc-Comments

I think what makes sense here is to change the first line to something like:
"Returns <code>true</code> if two variables are the same size."

stan/math/prim/arr/err/is_nonzero_size.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef STAN_MATH_PRIM_SCAL_ERR_IS_NONZERO_SIZE_HPP
+#define STAN_MATH_PRIM_SCAL_ERR_IS_NONZERO_SIZE_HPP
+
+namespace stan {
+namespace math {
+
+/**
+ * Check is the specified matrix/vector is on non-zero size
```
syclik: Similar comment as above.

"Returns <code>true</code> if the variable is non-zero size."

stan/math/prim/arr/err/is_matching_size.hpp
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This function checks the runtime sizes for
+ * variables that implement a <code>size()</code>
+ * method.
+ *
+ * @tparam T_y1 Type of the first variable
```
syclik: We should mention here `T_y1` type must have a `.size()` function (like the comment right above).

So... maybe something like:
"@tparam T_y1 Type of the first variable. This variable must have a <code>size()</code> method defined"

If you have a better way to describe that, it'll work too.

stan/math/prim/arr/err/is_nonzero_size.hpp
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @tparam T_y Type of container
```
syclik: Similar to the comment above: we should mention that this needs to follow a concept... that it needs a `.size()` function defined.

stan/math/prim/arr/err/is_ordered.hpp
```diff
@@ -0,0 +1,32 @@
...
+#define STAN_MATH_PRIM_ARR_ERR_IS_ORDERED_HPP
+
+#include <vector>
+#include <string>
+
+namespace stan {
+namespace math {
+
+/**
+ * Tests for ascedning order in a vector.
```
syclik: "Returns <code>true</code> if the vector is in ascending order."

stan/math/prim/scal/err/is_size_match.hpp
```diff
@@ -0,0 +1,30 @@
...
+#define STAN_MATH_PRIM_SCAL_ERR_IS_SIZE_MATCH_HPP
+
+#include <boost/type_traits/common_type.hpp>
+#include <stan/math/prim/scal/meta/likely.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Check if the provided sizes match
```
syclik: Similar comment as above.

syclik: I think this one should be documented a little differently because we're not actually passing in variables.

How's something like this?
```
Returns <code>true</code> if the provided sizes match.

@tparam T_size1 Type of the first provided size. Expecting this to be <code>int</code> or <code>size_t</code>.
@tparam T_size2 Type of the second provided size. Expecting this to be <code>int</code> or <code>size_t</code>.
```
And leaving the rest pretty much alone.

stan/math/prim/scal/err/is_size_match.hpp
```diff
@@ -0,0 +1,30 @@
...
+ * @tparam T_size1 Type of size 1
+ * @tparam T_size2 Type of size 2
+ *
+ * @param i Size 1
+ * @param j Size 2
+ *
+ * @return <code>true</code> if provided dimensions match
+ */
+template <typename T_size1, typename T_size2>
+inline bool is_size_match(T_size1 i, T_size2 j) {
```
syclik: Question... do you know if `check_size_match()` is actually used this way? I'm wondering why we don't just pass in the variables... there's probably a good reason for it.

test/unit/math/prim/arr/err/is_nonzero_size_test.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+  std::vector<double> a;
+  a.push_back(3.0);
+  a.push_back(3.0);
+  a.push_back(3.0);
+  a.push_back(3.0);
+
+  EXPECT_TRUE(stan::math::is_nonzero_size(a));
+
+  a.resize(2);
```
syclik: I think we can remove this test -- we've already got something with size 4. (just these two lines)

test/unit/math/prim/arr/err/is_nonzero_size_test.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+  std::vector<double> a;
+  a.push_back(nan);
+  a.push_back(nan);
+  a.push_back(nan);
+  a.push_back(nan);
+
+  EXPECT_TRUE(stan::math::is_nonzero_size(a));
+
+  a.resize(2);
```
syclik: Same here... we can remove these two lines.

test/unit/math/prim/scal/err/is_size_match_test.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+TEST(ErrorHandlingMatrix, checkSizeMatch) {
+  using stan::math::is_size_match;
+  int x;
+  size_t y;
+
+  x = 3;
+  y = 4;
+  EXPECT_FALSE(is_size_match(x, y));
+  EXPECT_FALSE(is_size_match(x, y));
```
syclik: We can remove lines 13-15... they're just duplicates.

test/unit/math/prim/scal/err/is_size_match_test.cpp
```diff
@@ -0,0 +1,23 @@
...
+  EXPECT_FALSE(is_size_match(x, y));
+  EXPECT_FALSE(is_size_match(x, y));
+
+  x = 2;
+  y = 2;
+  EXPECT_TRUE(is_size_match(x, y));
+  EXPECT_TRUE(is_size_match(x, y));
```
syclik: Same with lines 20-22.

test/unit/math/prim/scal/err/is_size_match_test.cpp
syclik: Maybe what we should be doing is:
```
is_size_match(x, y);
is_size_match(y, x);
is_size_match(x, x);
is_size_match(y, y);
```

That would be to flex the types. If that were the case, I'd maybe rename the variables so it's clear that we're testing types. Or you can use a string afterwards: so... something like
```
EXPECT_FALSE(is_size_match(x, y)) << "int, size_t";
```

oliverdechant:
>  do you want to have the matching `check_` functions call the `is_` functions?

That is still forthcoming.

oliverdechant:
Am I spamming with all these commits? I see three updates from build-bot.

test/unit/math/prim/scal/err/is_size_match_test.cpp
oliverdechant: Oh yeah.. it was late. I think I have what you are looking for.

stan/math/prim/scal/err/is_size_match.hpp
oliverdechant: No, I based `is_size_match()` around passing the replicated test cases from `check_size_match()`.

test/unit/math/prim/arr/err/is_nonzero_size_test.cpp
oliverdechant: Can you clarify? Remove lines 16 and 34?

stan/math/prim/arr/err/is_matching_size.hpp
```diff
@@ -0,0 +1,31 @@
+#ifndef STAN_MATH_PRIM_ARR_ERR_IS_MATCHING_SIZE_HPP
+#define STAN_MATH_PRIM_ARR_ERR_IS_MATCHING_SIZE_HPP
+
+#include <stan/math/prim/scal/err/is_size_match.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * @return <code>true</code> if the variable sizes match
```
syclik: Ah... this wasn't quite what I meant.

The first line should be a sentence about what it does. This line should go where it was... roughly line 23. Here, we should write:
```
* Returns <code>true</code> if the two variable sizes match.
```

stan/math/prim/arr/err/is_matching_size.hpp
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This function checks the runtime sizes for
+ * variables that implement a <code>size()</code>
+ * method.
+ *
+ * @tparam T_y1 Type of the first variable, requires function
+ *    access to <code>.size()</code>
```
syclik: maybe instead of "requires function access to" something like "requires class method"?

stan/math/prim/scal/err/is_size_match.hpp
syclik: thanks. I was able to find its uses. It's actually used when we check things like the number of columns of a variable matches the number of rows of another.

test/unit/math/prim/arr/err/is_nonzero_size_test.cpp
syclik: Sorry, remove lines 18 and 19. We've tested with a vector of size 4. We don't need to test also with a vector of size 2. It makes sense to test with a vector of size 0.

syclik:
and... it's ok to continue to add commits. Although, what may help is understanding when builds are triggered. They're triggered on pushes. You can commit a lot of times and if you only push once, it'll only try to build once.

stan/math/prim/scal/err/is_size_match.hpp
increasechief: Ah sorry I thought you were asking why the sizes are passed directly into `likely(i == static_cast<T_size1>(j))` rather than by some other method.

increasechief:
OK I think I will do this locally then push all the changes at once. It started out seeming easiest to just jump into a file and fix it on GitHub but by the end it was getting confusing. Anyway I hope GitHub didn't send you an email every time I committed something, I won't do that again.

stan/math/prim/scal/err/is_size_match.hpp
increasechief: I looked into this closer. Clearly this is optimized in ways that some other `check` functions may not be. Many are passing entire structures and then calculating simple `.size()` operations. 

Would it be desirable to change the use of all the `check` functions to pass the calculated values (i.e `check_fun(y.size())` rather than `check_fun(y)`) where possible?

increasechief: I'm a little behind on the rationale for this design pattern. Would the uses of `check_size_match` be changed to `check_matching_size`? Since `check_matching_size` simply passes the size values of the referenced structure into `check_size_match` anyway?

increasechief: Yes, the code is beyond me in those cases, but I now see what you are saying.

Issue #1122, opened 2019-02-18
Documenting, testing and cleanup of metaprograms


rok-cesnovar:
## Description

Some metaprograms in `meta` subfolders are missing documentation and/or tests. @syclik mentioned this in on Discourse https://discourse.mc-stan.org/t/help-with-the-math-library/7390/.

I will post the list of files that need commenting and those that have no tests in their respective `test/unit/math/` as a response to this issue.

Other minor cleanup stuff that I noticed while doing a thorough check:

- [x] [prim/scal/meta/seq_view.hpp](https://github.com/stan-dev/math/blob/develop/stan/math/prim/scal/meta/seq_view.hpp) is an empty file and should be removed. 
- [x] ~~[prim/scal/meta/scalar_seq_view.hpp](https://github.com/stan-dev/math/blob/develop/stan/math/prim/scal/meta/scalar_seq_view.hpp) should could be renamed (not sure on this) to seq_view~~
- [ ] `prim/arr/VectorBuilderHelper`, `prim/scal/VectorBuilder`, `prim/scal/VectorBuilderHelper`,  `prim/scal/StdVectorBuilder` should be  in snake_case not camelCase
- [x] replace the use of boost with the std lib equivalent where applicable
- [x] simplify `is_var_or_arithmetic`, `contains_nonconstant_struct` and others by using C++11 parameter packs (mentioned by @syclik in #1127)
- [ ] resolve the inconsistent namespaces for the metaprograms (or at least create some guidelines on when to use which namespace)
- [ ] create a matrix_return_type for cases like [this](https://github.com/stan-dev/math/pull/1185/files#diff-47fd23464dc444b604d42a25edfc1d8fR65)
- [x] ~~make a parameter packs version of `is_constant_struct`~~ consolidate all metaprograms that deal with type being constant to use `is_constant_all`

## Expected Output

Documented structs, functions, classes and cleaned up minor stuff related to `meta` subfolders.

#### Current Version:
v2.18.1

rok-cesnovar:
Files that are missing comments. * denotes that only some functions are missing comments or that the comments should be "doxygenized".
```
fwd/mat
	operands_and_partials
fwd/scal
	ad_promotable
	is_fvar
	operands_and_partials*
	partials_type
prim/arr
	contains_std_vector
	get
	index_type
	is_constant_struct
	is_vector
	length
	scalar_type
	VectorBuilderHelper
prim/mat
	empty_broadcast_array
	get
	is_constant_struct
[...]

PR #1132, opened 2019-03-01
Cleanup/generalize metaprograms with parameter packs

rok-cesnovar:
## Summary

This is an attempt to cleanup and generalize 4 metaprograms with parameter packs.

## Tests

There are 2 new tests added (there were none for `is_var_or_arithmetic`)

`test/unit/math/rev/scal/meta/is_var_or_arithmetic.cpp`
`test/unit/math/prim/scal/meta/is_var_or_arithmetic.cpp`
EDIT: 
There were a few tests added to `contains_fvar` tests.

## Side Effects

Hopefully none.

## Checklist

- [x] Math issue #1122 


stan/math/prim/scal/meta/contains_fvar.hpp
```diff
@@ -3,25 +3,16 @@
...
-    value = is_fvar<typename scalar_type<T1>::type>::value
-            || is_fvar<typename scalar_type<T2>::type>::value
-            || is_fvar<typename scalar_type<T3>::type>::value
-            || is_fvar<typename scalar_type<T4>::type>::value
-            || is_fvar<typename scalar_type<T5>::type>::value
-            || is_fvar<typename scalar_type<T6>::type>::value
-  };
-};
+template <typename... T>
+using contains_fvar = or_<is_fvar<typename scalar_type<T>::type>...>;
```
rok-cesnovar: Would a better name for `or_` this be `is_any`?

stan/math/prim/scal/meta/contains_nonconstant_struct.hpp
```diff
@@ -2,18 +2,12 @@
...
-          typename T4 = double, typename T5 = double, typename T6 = double>
-struct contains_nonconstant_struct {
-  enum {
-    value = !is_constant_struct<T1>::value || !is_constant_struct<T2>::value
-            || !is_constant_struct<T3>::value || !is_constant_struct<T4>::value
-            || !is_constant_struct<T5>::value || !is_constant_struct<T6>::value
-  };
-};
+template <typename... T>
+using contains_nonconstant_struct = or_not_<is_constant_struct<T>...>;
```
rok-cesnovar: Would a better name for this be `is_any_not`? I am really bad with naming, as witnessed here...

rok-cesnovar:
@seantalts the 
`C:\Rtools\mingw_64\bin\ar.exe: unable to rename 'lib/sundials_4.1.0/lib/libsundials_nvecserial.a'; reason: Permission denied` error is happening again.

seantalts:
Yikes. I restarted that phase and used icacls to try to set permissions
*even harder this time*. I have no other ideas left, except possibly making
the Jenkins user an administrator, which seems bad. For now let's just
restart the job phase when this happens and I'll try to hire someone who
knows these things better than I do.

stan/math/prim/scal/meta/contains_fvar.hpp
```diff
@@ -3,25 +3,16 @@
...
 #include <stan/math/prim/scal/meta/is_fvar.hpp>
 #include <stan/math/prim/scal/meta/scalar_type.hpp>
+#include <stan/math/prim/scal/meta/or.hpp>
 
 namespace stan {
 
 /**
- * Metaprogram to calculate the base scalar return type resulting
- * from promoting all the scalar types of the template parameters.
+ * Metaprogram to determine if any of the
```
rok-cesnovar: This comment seemed false to me (very similar to the /prim/scal/meta/return_type comment). I am not sure if this is descriptive enough.

stan/math/prim/scal/meta/and.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef STAN_MATH_PRIM_SCAL_META_AND_HPP
+#define STAN_MATH_PRIM_SCAL_META_AND_HPP
+
+#include <type_traits>
+
+namespace stan {
+namespace math {
```
rok-cesnovar: I put the helper metaprograms for conjuction and disjunction to the stan::math namespace, as the metaprograms are in all sorts of namespaces. See debate in #1122.

rok-cesnovar:
This is now ready for review. I added a few comments that will help anyone that will review this/use this.

stan/math/prim/scal/meta/and.hpp
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * A helper metaprogram to peform conjuction
+ * on type conditionals for zero or more provided types.
+ * See \link prim/scal/meta/is_var_or_arithmetic.hpp \endlink
+ * for an example.
+ */
+template <typename... T>
+struct and_ : std::true_type {};
```
bob-carpenter: We reserved the trailing underscore convention for member variables, so `and_` needs to be renamed.

stan/math/prim/scal/meta/and.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef STAN_MATH_PRIM_SCAL_META_AND_HPP
+#define STAN_MATH_PRIM_SCAL_META_AND_HPP
+
+#include <type_traits>
+
+namespace stan {
+namespace math {
+/**
+ * A helper metaprogram to peform conjuction
```
bob-carpenter: In general, you don't need to say things like "a helper metaprogram to".  Just say

```
Extends std::true_type when instantiated with zero or more template parameters,
all of which extend the std::true_type.
```

At least I think that's what it's doing.  And I don't think the definition of `and` should link to `or` for an example.

stan/math/prim/scal/meta/and.hpp
bob-carpenter: Are these just duplicating the C++17 types`conjunction` and `disjunction`?  E.g., see

https://en.cppreference.com/w/cpp/types/conjunction

There's a sizable type traits library now:

https://en.cppreference.com/w/cpp/header/type_traits

stan/math/prim/scal/meta/contains_fvar.hpp
bob-carpenter: Yes, this is good.  You just want to reformulate the doc so that you're not describing the function, but saying what it returns.  So this one should be something like:

> Defines a public `enum` named `value` which is defined to be true (1) 
if any of the template parameters includes a `var` as their base scalar and false (0) otherwise.
```

File: stan/math/prim/scal/meta/contains_fvar.hpp [hunk 7]
```diff
@@ -3,25 +3,16 @@
...
-    value = is_fvar<typename scalar_type<T1>::type>::value
-            || is_fvar<typename scalar_type<T2>::type>::value
-            || is_fvar<typename scalar_type<T3>::type>::value
-            || is_fvar<typename scalar_type<T4>::type>::value
-            || is_fvar<typename scalar_type<T5>::type>::value
-            || is_fvar<typename scalar_type<T6>::type>::value
-  };
-};
+template <typename... T>
+using contains_fvar = math::or_<is_fvar<typename scalar_type<T>::type>...>;
```
bob-carpenter: I don't understand any of this.  What is the `using` doing here?  And how is that function `is_fvar` getting applied to all the types?  Is there doc for how this works in the language somewhere?  I didn't know what to search for.

stan/math/prim/scal/meta/contains_fvar.hpp
bob-carpenter: `disjunction` is what the C++17 lib used.

stan/math/prim/scal/meta/contains_nonconstant_struct.hpp
```diff
@@ -2,18 +2,16 @@
...
-  enum {
-    value = !is_constant_struct<T1>::value || !is_constant_struct<T2>::value
-            || !is_constant_struct<T3>::value || !is_constant_struct<T4>::value
-            || !is_constant_struct<T5>::value || !is_constant_struct<T6>::value
-  };
-};
+/**
+ * Metaprogram to determine if any of the
+ * provided types have a base scalar
+ * type that can not be assigned to type double.
```
bob-carpenter: "can not" -> "cannot"

stan/math/prim/scal/meta/contains_nonconstant_struct.hpp
```diff
@@ -2,18 +2,16 @@
...
+ */
+template <typename... T>
+using contains_nonconstant_struct = math::or_not_<is_constant_struct<T>...>;
```
bob-carpenter: [optional]
Rather than using `or_not` (which again shouldn't have trailing underscore), how about using `is_nonconstant_struct` and just conjoining?

stan/math/prim/scal/meta/or.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * for an example.
+ */
+template <typename... Conds>
+struct or_ : std::false_type {};
+
+template <typename Cond, typename... Conds>
+struct or_<Cond, Conds...>
+    : std::conditional<Cond::value, std::true_type, or_<Conds...>>::type {};
+/**
+ * A helper metaprogram to peform disjunction
```
bob-carpenter: Again, document what it does rather than describing it abstractly.  It defines a null class where the work's in the inherited type.

stan/math/prim/scal/meta/contains_fvar.hpp
```diff
@@ -3,25 +3,16 @@
...
 
 }  // namespace stan
 #endif
```
bob-carpenter: [question]
Was there no need to change contains_fvar_test?

stan/math/prim/scal/meta/and.hpp
rok-cesnovar: Yes, this is duplicating the conjunction and disjunction for sub C++17 and will be replaced directly with them once Stan Math switches to C++17. As discussed in the [issue thread](https://github.com/stan-dev/math/issues/1122#issuecomment-468016606), we have 2 options to do something like this in C++14, that are described here: https://stackoverflow.com/questions/29603364/type-trait-to-check-that-all-types-in-a-parameter-pack-are-copy-constructible/29603857#29603857

rok-cesnovar: Yes, this is duplicating the conjunction and disjunction for sub C++17 and will be replaced directly with them once Stan Math switches to C++17. As discussed in the issue thread, we have 2 options to do something like this in C++14, that are described here: https://stackoverflow.com/questions/29603364/type-trait-to-check-that-all-types-in-a-parameter-pack-are-copy-constructible/29603857#29603857

stan/math/prim/scal/meta/contains_fvar.hpp
rok-cesnovar: The `using` is a [type alias declaration](https://en.cppreference.com/w/cpp/language/type_alias). This [stackoverflow anwser](https://stackoverflow.com/a/2795024) describes this way better than I could. This same idea could be used to simply most of the metaprograms ( I would gladly take this task if you guys feel its of any value to do this).

This gets applied to all types by doing what `fold_expressions` does in C++17 ([example](https://www.codingame.com/playgrounds/2205/7-features-of-c17-that-will-simplify-your-code/fold-expressions)). This will be greatly simplified once we get there.

stan/math/prim/scal/meta/contains_fvar.hpp
rok-cesnovar: Not really, the API remained the same. I added a few tests for more than 1 type.

stan/math/prim/scal/meta/or.hpp
rok-cesnovar: `or_not_` was removed.

rok-cesnovar:
I believe I made all the changes and comments requested. Ready for re-review.

stan/math/prim/scal/meta/contains_fvar.hpp
bob-carpenter: Thanks for the links.  C++ never fails to amaze me with its odd overloading of syntax.  I'll think of it like a templated typedef.

Issue #1133, opened 2019-03-02
Stuff to do before making GPU announcement


SteveBronder:
There's a list of a few smaller open issues we want to address before we make a formal announcement that we have GPU stuff now.

#### Blocking:

-------------

- [x] Change the folder name to `opencl` and the matrices from `matrix_gpu` to `matrix_cl`
  - The OpenCL stuff can also use CPUs, so I think it's appropriate we change this
- [x] Make sure the install instructions in the wiki are clear and correct
  - I think this is mostly good, but I want to spin up a docker container / aws instance and go through them to make sure the instructions make sense. We could probably even include a gist to an install script for aws.
- [ ] Make a commit to Cmdstan with the right compile flags
#### Non-blocking but still to-do

----------------------------

- [ ] Add higher level support for tuning the parameters in the GPU code during model warmup. (possibly allow users to set those)
- [ ] Run sampling tests on a few example models and tuning parameters for sanity checks
  - How to do this will probably be the biggest discussion in this issue, I'll review below how we are thinking about this.

#### Things to talk about on Thursday March 7th Meeting

---------------------------------------
 
- [x] Discuss how to push this up to cmdstan, Rstan and Pystan
- [ ] Add to Seans repo for performance tests so easier to test branches of the math library
- [ ] Reach out to Nvidia / Khronos Group about joint press release (?)

Is there anything else I'm missing here?

rok-cesnovar:
I will tackle the renaming stuff today.

bob-carpenter:
Thanks a lot for laying this out.  I'm really looking forward to GPU support!

> On Mar 1, 2019, at 8:04 PM, Steve Bronder <notifications@github.com> wrote:
> 
> There's a list of a few smaller open issues we want to address before we make a formal announcement that we have GPU stuff now.
> 
> Blocking:

...

> Non-blocking but still to-do

...

syclik:
I was trying to run doxygen from `develop` and it failed. It looks like it's coming from the GPU code:
```
$ make doxygen
mkdir -p doc/api
doxygen doxygen/doxygen.cfg
.../math/stan/math/opencl/kernels/sub_block.hpp:41: error: argument 'src_rows' from the argument list of stan::math::opencl_kernels::sub_block has muliple @param documentation sections (warning treated as error, aborting now)
make: *** [doxygen] Error 1
```

Could someone verify that this is happening? If it runs correctly for you, then maybe something's misconfigured on my end.

[...]

PR #1137, opened 2019-03-02
Change GPU to OpenCL

rok-cesnovar:
## Summary

There are a number of simple changes: 

- moved `stan/math/gpu` to `stan/math/opencl`
- moved `test/unit/math/gpu` to `test/unit/math/opencl`
- changed all references of `matrix_gpu` to `matrix_cl`
- changed the references `matrix on the GPU` to `OpenCL matrix`

## Tests

`/gpu` Tests were moved to `/opencl`

## Side Effects

Not sure on the Jenkins changes, hopefully I didnt break anything regarding the tests. Other changes are trivial.

## Checklist

- [x] Math issue #1133 


rok-cesnovar:
The GPU tests fail because the files were moved to `/opencl` or because I changed the Jenkins file.

SteveBronder:
Looks like some changed wasn't reflected on Jenkins. Tried to kick it off but we'll see. I'm running all the tests locally to check if something was missed

SteveBronder:
@seantalts do you know why Jenkins is still referencing `./runTests.py -j16 test/unit/math/gpu --make-only`?

It looks like it's changed in the Jenkinsfile

Jenkinsfile
```diff
@@ -159,16 +159,16 @@ pipeline {
                     }
                     post { always { retry(3) { deleteDir() } } }
                 }
-                stage('GPU Tests') {
-                    agent { label "gpu" }
+                stage('OpenCL Tests') {
+                    agent { label "opencl" }
```
seantalts: we probably want to keep this label as 'GPU' because it specifies that we want a node with a GPU

seantalts:
You can more easily kick off just a specific part of a Jenkins job by going to the job, making sure you're logged in, and then hitting the button to restart just that phase:
image

That way we don't get a bunch of garbage commits in our git history, too :)

seantalts:
I thiiiink this might be an artifact of a security policy that basically doesn't trust new Jenkinsfiles from forks by default... Can we put this onto a branch on `stan-dev` and try it? I think you should both be members of the stan-dev org and writers to math already, no?

SteveBronder:
Yeah that makes sense I'll make a new branch in stan-dev

Issue #1142, opened 2019-03-04
Add optimization for special cases of matrix multiply for OpenCL


rok-cesnovar:
## Description

For cases when we know that an input matrix has certain properties, we can optimize the OpenCL kernels significantly. 

In the near future we would like to add at least the following: 

- [x] `lower_tri` x `rect` 
- [x] `rect` x `lower_tri`
- [x] `upper_tri` x `rect`
- [x] `rect` x `upper_tri`

Where `lower_tri` and `upper_tri` are lower and upper triangular matrices and `rect` is a non-triangular matrix. These are all pretty straightforward to implement (some indexing wizardry but not much more) but reduce the multiplication execution times down to 55% of the regular OpenCL matrix multiply.

- [x]  [N, K] x [K, M] where K >>> N,M; NxM is "not large" and K is "large"

This cases do not offer as much speedup (more in the range of 10-20% depending on N,K &M).However these cases of matrix multiply come up quite often in Stan models which makes this worthwile. Large is in quotes as this are soft limits that will require a thorough performance analysis to set. 

With this issue we would also like to:

- [x] replace regular OpenCL multiply in the merged Cholesky OpenCL implementation with applicable optimizations 
- [x] investigate the possibility to check these conditions in a general matrix multiply and call the appropriate kernels if the conditions are met [ will maybe be possible with the new compiler ]
- [x] improve the `multiply_transpose` kernel with this optimizations
- [x] investigate if we can remove padding without huge impact to the overall performance

## Example
In order to use the special case of `lower_tri` x `rect` we will be able to use:
```
stan::math::matrix_cl m11(m1); // lower triangular matrix
stan::math::matrix_cl m22(m2);

auto m33 = multiply<TriangularViewCL::Lower, TriangularViewCL::Entire>(m11, m22);

```
## Expected Output

Faster OpenCL matrix multiplication for special cases that occur quite often in Stan/Stan Math.

#### Current Version:
v2.18.1

SteveBronder:
Esoteric, but it could be cool in the overloaded `*` operator to check the matrices for these conditions and call the appropriate kernels as well as exposing them normally.

Keeps feeding into my want for doing Seans PR where we check square and symmetry conditions on data once and then keep those check results. Then for the end user and us there would be no overhead in having `*` check these conditions and do the appropriate routine.

rok-cesnovar:
Great point, thanks! Will add to the issue to investigate this.

PR #1143, opened 2019-03-04
Triangular matrix multiplication with OpenCL

rok-cesnovar:
## Summary

This PR adds special matrix multiplication OpenCL kernel for cases when either of the matrices are lower/upper triangular. 

The following are the ratios of the execution time of the lower x square  and upper x square compared to the regular OpenCL matrix multiply.

```
N	lower	upper
200: 1.01714	0.948261
500: 0.861604	0.862615
1000: 0.758448	0.782959
1500: 0.709263	0.722426
2000: 0.706571	0.678224
3000: 0.638379	0.643627
4000: 0.602315	0.600731
5000: 0.585764	0.58994
7500: 0.569375	0.567946
```
EDIT: 
This PR also improves the performance of `multiply_transpose` with the same optimizations.

## Tests

The tests were added to the `test/unit/math/opencl/multiply_test.cpp`:

- lower x square (small & large)
- lower x rectangular (large)
- upper x square (small & large)
- upper x rectangular (large)
- square x lower (small & large)
- rectangular x lower (large)
- square x upper (small & large)
- rectangular x upper (large)

## Side Effects

/

## Checklist

- [x] Math issue #1142 


rok-cesnovar:
This is a work in progress PR. For now `lower` x `rect` and `upper` x `rect` are almost finished (need to check if we can get some more speedup at lower sizes). Will add `rect` x `lower` and `rect` x `upper` tomorrow.

rok-cesnovar:
The `tri_multiply` now supports all combinations of `lower/upper/rect` x `lower/upper/rect`. All versions are faster than the regular multiply, except for `rect` x `rect` which is a regular matrix multiply and `tri_multiply` should even be used here.

This are are percentages of execution time of a regular matrix multiply measured with [this gist](https://gist.github.com/rok-cesnovar/289caee15f5ce28a7e84c64432c3cbe8).
```
N	lower	upper
200: 1.01714	0.948261
500: 0.861604	0.862615
1000: 0.758448	0.782959
1500: 0.709263	0.722426
2000: 0.706571	0.678224
3000: 0.638379	0.643627
4000: 0.602315	0.600731
5000: 0.585764	0.58994
7500: 0.569375	0.567946
```
@SteveBronder do you think it would be a good idea to change the regular multiply function to have a signature like this

```
template <TriangularViewCL triangular_view_A = TriangularViewCL::Entire,
          TriangularViewCL triangular_view_B = TriangularViewCL::Entire>
inline auto multiply(const matrix_cl& A, const matrix_cl& B) {
```
And then have the default combination of template parameters call the regular multiply kernel while all the other combinations call this triangular multiply kernel? Or should this be a separate function like `tri_multiply`? Need to decide on this and do another sweep to double check then its ready for review.

rok-cesnovar:
This is ready for review.

@ anyone that will review this:

There are not many changes, the added kernel in [kernels/tri_multiply.hpp](https://github.com/stan-dev/math/blob/d033c484bda73cdef18031e657e8ce141625f22d/stan/math/opencl/kernels/tri_multiply.hpp) is based on [kernels/matrix_multiply.hpp](https://github.com/stan-dev/math/blob/develop/stan/math/opencl/kernels/matrix_multiply.hpp) with an added calculation of the start/end tile and a check to ignore elements above/under the diagonal for the diagonal tiles that can not be  completely skipped.

rok-cesnovar:
@seantalts It seems that the aws instance is down? Sorry for tagging, nothing urgent, just not sure if you were aware.

```
+ ./runTests.py -j40 test/prob

Cannot contact ec2-spot-c5d18x: hudson.remoting.RequestAbortedException: java.nio.channels.ClosedChannelException
```

seantalts:
At least now it's set up to automatically come back up 😅 you may sometimes find retrying is necessary, sorry about that. Hoping to hire a professional soon to deal with it.

rok-cesnovar:
The same optimization as for the lower/upper multiply was added to the multiply_tranpose kernel. I updated the PR and issue.

SteveBronder:
> @SteveBronder do you think it would be a good idea to change the regular multiply function to have a signature like this
```c++
template <TriangularViewCL triangular_view_A = TriangularViewCL::Entire,
          TriangularViewCL triangular_view_B = TriangularViewCL::Entire>
inline auto multiply(const matrix_cl& A, const matrix_cl& B) {
```
> And then have the default combination of template parameters call the regular multiply kernel while all the other combinations call this triangular multiply kernel? Or should this be a separate function like tri_multiply? Need to decide on this and do another sweep to double check then its ready for review.

Sorry for the delayed response. Yeah I think that's the template is the nicest way to do this.

seantalts:
I don't think we can expose templates to the Stan language, right? Because of that, we have typically opted to provide new functions for those use cases, like (`mdivide_left_tri_low`)[https://mc-stan.org/docs/2_18/functions-reference/linear-algebra-functions-and-solvers.html#matrix-division-operators-and-functions]

rok-cesnovar:
These are `matrix_cl` functions that are not exposed to the Stan language and are just called inside prim/rev/etc.

seantalts:
Isn't it exposed to the Stan language? It's in the `stan::math` namespace and adding an overload (and templated version) to an existing `multiply` function, no?

seantalts:
If it's intended to be only for opencl stuff, let's just decrease complexity and keep it in an `stan::math::opencl` namespace and only add things to the `stan::math` namespace that we explicitly want to add to Stan? I say this somewhat motivated by the recent and current failures on `develop` where adding overloads created totally non-obvious test failures in unrelated parts of the unit tests, so I'm quickly becoming a fan of trying to be very explicit and purposeful when adding overloads to existing Stan functions.

rok-cesnovar:
I am confused. The `cholesky_decompose` stuff isnt directly exposed to the Stan language, only through `/prim/cholesky_decompose` and `/rev/cholesky_decompose`. I am not sure how can we expose with matrix_cl arguments to the Stan language?

seantalts:
Which cholesky_decompose stuff specifically? Right now there are `cholesky_decompose` functions in `stan::math` that take in `matrix_cl` types that are exposed to the Stan language in the sense that a Stan model runs in the `stan::math` namespace, but you are also right that a Stan program doesn't currently generate `matrix_cl` types on its own. That could easily change in stanc without a change to the Math library.

Basically, if it's going in `stan::math`, I think we want to assume that it could one day be exposed to the Stan language and generally follow the conventions in `stan::math` that make code generation easy. We want to look forward to the time when the Stan compiler is emitting Math code that operates on `matrix_cl` types. We want to avoid configuration templates for anything like that and prefer the simpler mechanism of a different name as we don't code generate templates for specific functions, unless there are strong reasons to prefer templates.

I think it's okay to keep overloading stuff in `stan::math` but only for the use case that we expect stanc to eventually generate code to take advantage of that.

rok-cesnovar:
Ok, I get it now, sorry for being slow :) If we are thinking ahead and the use of `matrix_cl` with stanc than yeah, that makes total sense. Having a general template multiply in opencl:: and then exposing some useful cases with specific names to the Stan language is then a reasonable solution to this right? So we would have a multiply function that works on `matrix_cl` in `opencl::` and another one in `stan::math`. Those in stan::math would be just calls to the one in `opencl::`.

If so, I am going to modify this PR and notify when ready for review again.

seantalts:
Yeah, that makes sense to me. One more potential tidbit: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t65-use-tag-dispatch-to-provide-alternative-implementations-of-a-function

Not sure if this is relevant, but if we knew at compile time what was triangular and what wasn't, we could use tag dispatch to send to the correct algorithm.

rok-cesnovar:
For now I am only exposing the general multiply to the Stan language. The general templated `multiply` is in the `opencl` namespace. For now the triangular stuff is only going to be used inside the OpenCL implementations of `prim` and `rev` `cholesky` & `mdivide_left_tri`. We can add stuff like `lower_tri_rect_multiply`, `upper_tri_rect_multiply` later if we find them to be useful in the Stan language.

Anyway, this should be ready for review once it passes again.

seantalts:
Sounds good, thanks! Will get to it once `develop` is back in good shape.

SteveBronder:
Is multiplication the only thing in the opencl namespace with this PR? If we are going to make an opencl namespace I think it should be a separate PR and figure out what should be in there. Something like

- `opencl` 
  - everything in the Opencl folder. Change name of `opencl_context` to `context` so it's called via `opencl::context`

We would also get rid of the `opencl_kernel` namespace. The `add` kernel would now be called via `opencl::add(...)` and expose it in `stan::math` via `+`. 

Thoughts?

seantalts:
Why?

SteveBronder:
Which why? Like why opencl namespace in general? idk, isn't it kind of weird to only have multiply in the opencl namespace?

seantalts:
A journey begins with a single step and all that. I don't think the refactor needs to hold up this PR (we can make an issue). 

I also don't think we'd get rid of the kernels namespace (maybe nest it under opencl though?) - that `multiply` is different from this `multiply`.

SteveBronder:
Aight word, as long as we flag it as an issue and do it at some point in time I'm cool with it

rok-cesnovar:
I agree that we need an issue/PR to separate the functions that we dont envision being exposed to the `opencl` namespace. I will make an issue for this.

stan/math/opencl/kernels/multiply_transpose.hpp
```diff
@@ -63,23 +63,24 @@ static const char* multiply_transpose_kernel_code = STRINGIFY(
...
+                            * B_local[thread_block_col
+                                      + w * THREAD_BLOCK_SIZE_COL][block_ind];
+                }
               }
             }
           }
+          barrier(CLK_LOCAL_MEM_FENCE);
         }
-        barrier(CLK_LOCAL_MEM_FENCE);
       }
```
seantalts: Hah, this misplaced `}` seems like a pretty major bug. If you haven't yet, could you write a regression test that fails on the old version? Thanks :D

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+static const char* tri_multiply_kernel_code = STRINGIFY(
+    // \endcond
+    /**
+     * OpenCL Matrix multiplication of a triangular and rectangular matrix
+     *
+     * @param[in] A the lower/upper triangular matrix in matrix multiplication
+     * @param[in] B the right matrix in matrix multiplication
+     * @param[out] C the output matrix
+     * @param[in] M Number of rows for matrix A
+     * @param[in] N Number of rows for matrix B
```
seantalts: cols?

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+     * @param[in] K Number of cols for matrix A and number of rows for matrix B
+     * @param[in] lower_upper_A the triangularity of A (lower, upper or none)
```
seantalts: forgive me if we already went over this, but could we use a C enum here?

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+      const int thread_block_row = get_local_id(0);
+      const int thread_block_col = get_local_id(1);
+      // global thread index
+      const int i = THREAD_BLOCK_SIZE * get_group_id(0) + thread_block_row;
+      const int j = THREAD_BLOCK_SIZE * get_group_id(1) + thread_block_col;
+
+      // indexes that determine the last indexes that need to compute
+      // in order to remove the unnecesary multiplications in the special
+      // multiplication of A*A^T
+      const int jMin = THREAD_BLOCK_SIZE * get_group_id(1);
```
seantalts: style - j_min or jmin

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+      // This kernel is based on the kernel for the regular
+      // matrix multiply where the multiplitcation is split
+      // in tiles.
+      // In cases where a matrix is lower triangular
+      // its not necessary to multiply the elements
+      // over the diagonal, therefore those tiles
+      // in the matrix multiply can be skipped.
+      // With upper triangular matrices we dont need
+      // to multiply the elements under the diagonal,
+      // so thosee tiles can be skipped.
```
seantalts: those*

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+      // The following code determines the start and
+      // end tile based on triangularity of the input matrices
+      // If no matrices are triangular the starting tile
+      // is 0 and the end tile is num_tiles-1 which
+      // is the same as in \link kernels/multiply.hpp multiply()
```
seantalts: does that mean we can get rid of multiply.hpp's multiply() or outsource that impl to this one?

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+      const int end_tile_A = !is_A_lower * (num_tiles - 1)
```
seantalts: Could we just make these ternary expressions or if/else? multiplying by a boolean is a little too clever :P

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+                             + is_A_lower * (i / THREAD_BLOCK_SIZE);
+      const int end_tile_B = !is_B_upper * (num_tiles - 1)
+                             + is_B_upper * (j / THREAD_BLOCK_SIZE);
+      const int start_tile_A = is_A_upper * (i / THREAD_BLOCK_SIZE);
+      const int start_tile_B = is_B_lower * (j / THREAD_BLOCK_SIZE);
+      int start_tile = 0;
```
seantalts: 1) can be const I think
2) could write
```cpp
const start_tile = max(start_tile_A, start_tile_B);
```
or w/ ternary op
```cpp
const start_tile = (start_tile_A > start_tile_B) ? start_tile_A : start_tile_B;
```

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+        start_tile = start_tile_A;
+      } else {
+        start_tile = start_tile_B;
+      }
+      if (end_tile_A > end_tile_B) {
+        end_tile = end_tile_B;
+      } else {
+        end_tile = end_tile_A;
+      }
+      for (int tile_ind = start_tile; tile_ind <= end_tile; tile_ind++) {
```
seantalts: i think `idx` is a more common abbreviation for index

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+        const int tiled_i = THREAD_BLOCK_SIZE * tile_ind + thread_block_row;
+        const int tiled_j = THREAD_BLOCK_SIZE * tile_ind + thread_block_col;
+        // each thread copies WORK_PER_THREAD values to the local
+        // memory
+        for (int w = 0; w < WORK_PER_THREAD; w++) {
+          const int tiled_i = THREAD_BLOCK_SIZE * tile_ind + thread_block_row;
+          const int tiled_j = THREAD_BLOCK_SIZE * tile_ind + thread_block_col;
```
seantalts: should these be here? they are declared above

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+                   [thread_block_row]
+                = A[(tiled_j + w * THREAD_BLOCK_SIZE_COL) * M + i];
+          } else {
+            A_local[thread_block_col + w * THREAD_BLOCK_SIZE_COL]
+                   [thread_block_row]
+                = 0.0;
+          }
+          if ((is_B_lower && ((j + w * THREAD_BLOCK_SIZE_COL) <= tiled_i))
+              || (is_B_upper && ((j + w * THREAD_BLOCK_SIZE_COL) >= tiled_i))
+              || (!is_B_lower && !is_B_upper)) {
```
seantalts: Do we not have a triangular copy kernel lying around somewhere? I guess we can't easily call it, is that right? I wonder if we should be including more helper functions in that helpers.cl file...

stan/math/opencl/kernels/tri_multiply.hpp
```diff
@@ -0,0 +1,155 @@
...
+                   [thread_block_row]
+                = B[(j + w * THREAD_BLOCK_SIZE_COL) * K + tiled_i];
+          } else {
+            B_local[thread_block_col + w * THREAD_BLOCK_SIZE_COL]
+                   [thread_block_row]
+                = 0.0;
+          }
+        }
+        barrier(CLK_LOCAL_MEM_FENCE);
+        for (int block_ind = 0; block_ind < THREAD_BLOCK_SIZE; block_ind++) {
```
seantalts: idx

test/unit/math/opencl/multiply_test.cpp
```diff
@@ -152,9 +151,263 @@ TEST(AgradRevMatrix, multiply_big) {
...
+                                          stan::math::TriangularViewCL::Upper>(
+      m11, m22);
+
+  stan::math::copy(m3_cl_res, m33);
+
+  EXPECT_MATRIX_NEAR(m3, m3_cl_res, 1e-10);
+}
+
+TEST(MathMatrix, rect_upper_tri_multiply_big_rect) {
+  int size = 512;
```
seantalts: I think we can make a simple tweak to get a little more test coverage: would you mind changing all (except for one) of these test `size`s to some arbitrary numbers that aren't handy powers of 2 or even necessarily even numbers?

stan/math/opencl/kernels/multiply_transpose.hpp
rok-cesnovar: Luckily in this instance its just a performance issue, not a bug. The values `jMin` and `iMax` determine if the thread actually does anything or it could just skip the entire calculation. In the previous location all the threads were syncing all the time, even those that could skip everything. That slowed things down a bit, but it did not corrupt the resulting matrix.

Screenshot from 2019-03-13 17-35-15

`if (jMin <= iMax) {` eliminates threads that are responsible for the red parts of the resulting matrix. Now that I had to think of this again, this check could be moved to the outmost level as these threads actually dont need to work at all.

All threads responsible for the grey tiles need to at least copy the data to the shared memory (to "help" the threads in the tile that actually need to multiply). The `if ((j + w * THREAD_BLOCK_SIZE_COL) <= i) {` (line 74 & 88) check eliminates the threads above the diagonal in grey tiles from multiplication and writing to the result matrix.

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: Ah, have to fix this in the regular multiply also.

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: Yes we could. I am not sure if we discussed this or not either :) 

We would have to put it in the helpers where we would do something like this:
```
char* helpers =
    R"(
  enum triangular_view{
      TRIANGULAR_VIEW_LOWER = LOWER,
      TRIANGULAR_VIEW_UPPER = UPPER,
      TRIANGULAR_VIEW_ENTIRE = ENTIRE
  }; 

... rest of helpers
```
The LOWER, UPPER, ENTIRE are defines that are set at kernel compile times based on the values in our `TriangularViewCL` enum. This was done so that if someone goes and changes the values in TriangularViewCL, the whole thing would still work. If you think this is a fine solution, then I can add this easily.

Sadly the STRINGIFY() doesnt want to work with enums :/

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: I am going to do another perf test to see if we get any performance regression for the general matrix multiply if we just use this one. My instinct was that the below checks on the copy to `A_local` and `B_local` have a bad impact, hence the 2 kernels. I will double check.

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: Ah, force of habit as branching used to be slower on GPUs than this clever stuff :)

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: my bad :/

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: Again this was my instinct on performance that this check is faster than queuing another kernel. I will double check.

rok-cesnovar: Turns out my instinct were right, adding another copy does negatively effect performance. 

But based on your idea, I added the triangular template parameter to the `sub_block` and that way Its actually faster than the original version. So a big thanks for that!!

stan/math/opencl/kernels/multiply_transpose.hpp
```diff
@@ -44,15 +44,15 @@ static const char* multiply_transpose_kernel_code = STRINGIFY(
       for (int w = 0; w < WORK_PER_THREAD; w++) {
         acc[w] = 0.0;
       }
+      if (j_min <= i_max) {
+        const int numTiles = (N + THREAD_BLOCK_SIZE - 1) / THREAD_BLOCK_SIZE;
```
seantalts: sorry, another camelCase -> snake_case :<

rok-cesnovar:
So this has now changed significantly: 

- `tri_multiply` is now merged into `matrix_multiply`
- `sub_block` supports triangular src matrices

The only open comment is the one on the enums. If you agree with putting them in the helpers string as I describe in the comment, I can add that. Although I can also see that being a separate issue+PR in which we cleanup that for all the current kernels. I can tackle that after this gets merged.

stan/math/opencl/kernels/tri_multiply.hpp
seantalts: Gotcha. importing the actual int values from C++ makes sense. Should we just do this:
```cpp
std::ostringstream enums;
enums << "enum triangular_view {\n";
enums << "    LOWER=" << TriangularViewCL::Lower << ";\n";
enums << "    UPPER=" << TriangularViewCL::Upper << ";\n";
enums << "    ENTIRE=" << TriangularViewCL::Entire << ";\n};";
```
Or possibly the `const char*` version with sprintf, that might even be easier to read. We have another enum we could do this way too.

This is assuming that this will help you simplify the code using it somewhat, and make the type signature communicate a little more. This is just based on me guessing, though, so if you think or see that it's not making it easier we can drop it.

stan/math/opencl/kernels/tri_multiply.hpp
seantalts: Ah okay. That's fine, we can have both.

stan/math/opencl/kernels/tri_multiply.hpp
seantalts: hah, nice!

stan/math/opencl/kernels/tri_multiply.hpp
rok-cesnovar: Yeah, that looks way better than my idea. I believe this is useful but would rather make the changes to the entire `opencl_kernels` code base at once if you agree with that. I can take care of this tomorrow.

rok-cesnovar:
Thanks! Yes, I am going to create an issue for the `enums `and hopefully also tackle that today.

seantalts:
Should we merge this then?

rok-cesnovar:
I think we should. A simple PR that replaces the regular matrix multiply with these optimization in the prim and rev cholesky will also be up later today so we can take advantage of all this.

Issue #1064, opened 2018-11-10
Make a cache for the gpu


SteveBronder:
## Description
This is an issue for a PR to attach an OpenCL buffer to Eigen matrices so that we can keep data  on the GPU.

We now have two `matrix_gpu` constructors, one that works for `is_constant` types (1) and another that works for `vars` (2). When a developer passes data over to the GPU, constructor (1) will fill the Eigen matrix's OpenCL buffer with the data so that the next time that matrix is passed to the GPU we perform a device to device copy instead of a host to device copy. In the case of vars we don't store a buffer and still have to pass the var's `val_` from host to device.

The `copy` method also has this now, where Eigen matrices with primitive types are cached while `var` types are not.

## Example

```cpp
auto m = stan::math::matrix_d::Random(100, 100).eval();
// Caches the data
stan::math::matrix_gpu m1(m);
// Uses the cache
stan::math::matrix_gpu m2(m);
```

## Expected Output

Expected output is a side effect that creates an OpenCL buffer on first copy from host to device for Eigen matrices with primitive types.

## Additional Information

I'll go into performance tests in the associated PR, but the long and short is that data transfers are a lot faster but var transfers are still v v v slow.

## Current Math Version
v2.18.0

PR #1164, opened 2019-03-17
Packed data transfers to and from the OpenCL device

rok-cesnovar:
## Summary

This PR optimizes the data transfers of triangular matrices by transferring the packed std::vectors of length `N*(N+1)/2` instead of `N*N` flat matrices. It does this with 2 kernels that pack and unpack matrices form the std::vector to the flat matrix_cl buffer.

## Tests

There are 2 added tests for the `packed_copy`: `matrix_cl_pack_unpack_copy_upper` and `matrix_cl_pack_unpack_copy_lower`. Both test both the pack and unpack kernels. The current sizes are 42 and 51, I think there is no added value to test larger sizes as the kernel is done with the `global_range_kernel` and there is no edge cases that would involve block sizes and such.

## Side Effects

/

## Checklist

- [x] Math issue #1064 


rok-cesnovar:
Benchmarks with perf-math: 

```
------------------------------------------------------------------------
Benchmark                                 Time           CPU Iterations
------------------------------------------------------------------------
BM_non_packed/BM_non_packed_500     1548709 ns    1547219 ns        450
BM_non_packed/BM_non_packed_1000    4921541 ns    4920931 ns        144
BM_non_packed/BM_non_packed_1500   10558682 ns   10557476 ns         66
BM_non_packed/BM_non_packed_2000   17979217 ns   17976223 ns         39
BM_non_packed/BM_non_packed_2500   27412439 ns   27410331 ns         25
BM_non_packed/BM_non_packed_3000   39217019 ns   39210717 ns         18
BM_non_packed/BM_non_packed_3500   53163948 ns   53163562 ns         12
BM_non_packed/BM_non_packed_4000   69458951 ns   69459702 ns          9
BM_non_packed/BM_non_packed_5000  108139491 ns  108141607 ns          6
BM_non_packed/BM_non_packed_6000  156129094 ns  156130292 ns          4
BM_non_packed/BM_non_packed_7000  214435438 ns  214437223 ns          3
BM_non_packed/BM_non_packed_8000  285304509 ns  285282310 ns          2
BM_non_packed/BM_non_packed_9000  360151845 ns  360153652 ns          2
BM_non_packed/BM_non_packed_10000  450269802 ns  449562296 ns         2
BM_non_packed/BM_non_packed_12000  727367335 ns  727312646 ns         1

----------------------------------------------------------------
Benchmark                         Time           CPU Iterations
----------------------------------------------------------------
BM_packed/BM_packed_500     1749498 ns    1719261 ns        404
BM_packed/BM_packed_1000    4134980 ns    4088902 ns        182
BM_packed/BM_packed_1500    8495267 ns    8468982 ns         83
BM_packed/BM_packed_2000   15430075 ns   14244970 ns         50
BM_packed/BM_packed_2500   20633337 ns   20628705 ns         34
BM_packed/BM_packed_3000   28338156 ns   28337208 ns         24
BM_packed/BM_packed_3500   38031626 ns   38031421 ns         18
BM_packed/BM_packed_4000   48328480 ns   48325412 ns         15
BM_packed/BM_packed_5000   74635423 ns   74636415 ns         10
BM_packed/BM_packed_6000  102338552 ns  102338122 ns          7
BM_packed/BM_packed_7000  134571228 ns  134571074 ns          6
BM_packed/BM_packed_8000  178851482 ns  178818437 ns          4
BM_packed/BM_packed_9000   220327944 ns  220316289 ns         3
BM_packed/BM_packed_10000  266804774 ns  266653012 ns         3
BM_packed/BM_packed_12000  370485825 ns  370441945 ns         2
```

stan/math/opencl/copy.hpp
```diff
@@ -89,6 +91,69 @@ void copy(Eigen::Matrix<double, R, C>& dst, const matrix_cl& src) {
...
+/**
+ * Packs the flat triagnular matrix on the OpenCL device and
+ * copies it to the std::vector.
+ *
+ * @tparam triangular_view the triangularity of the source matrix
+ * @param dst the destination std::vector
+ * @param src the flat triangular source matrix on the OpenCL device
+ */
+template <TriangularViewCL triangular_view>
+inline void packed_copy(std::vector<double>& dst, const matrix_cl& src) {
```
rok-cesnovar: I used std::vector everywhere for the packed version. Should we use Eigen::VectorXd or something else?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -321,19 +322,22 @@ class cholesky_opencl : public vari {
    *
    */
   virtual void chain() {
-    auto L_adj_cpu = Eigen::MatrixXd::Zero(M_, M_).eval();
-    auto L_cpu = Eigen::MatrixXd::Zero(M_, M_).eval();
+    int packed_size = M_ * (M_ + 1) / 2;
+    std::vector<double> L_adj_cpu(packed_size);
+    std::vector<double> L_val_cpu(packed_size);
```
SteveBronder: Looking [here](https://stackoverflow.com/questions/8928547/vector-initialization-or-reserve) not a big difference, but may be better to declare these with `std::vector<double> L_val_cpu;` then use `L_val_cpu.reserve(packed_size);` to avoid some construction costs of the underlying type.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -321,19 +322,22 @@ class cholesky_opencl : public vari {
...
 
     size_t pos = 0;
     for (size_type j = 0; j < M_; ++j) {
       for (size_type i = j; i < M_; ++i) {
-        L_adj_cpu.coeffRef(i, j) = vari_ref_L_[pos]->adj_;
-        L_cpu.coeffRef(i, j) = vari_ref_L_[pos]->val_;
+        L_adj_cpu[pos] = vari_ref_L_[pos]->adj_;
+        L_val_cpu[pos] = vari_ref_L_[pos]->val_;
```
SteveBronder: Think we can remove the double for loop here and just iterate over the `packed_size`?

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -384,11 +388,13 @@ class cholesky_opencl : public vari {
       L_adj.sub_block(B_adj, 0, 0, k, 0, m_k_ind, j);
       L_adj.sub_block(C_adj, 0, 0, k, j, m_k_ind, k_j_ind);
     }
-    copy(L_adj_cpu, L_adj);
+    packed_copy<TriangularViewCL::Lower>(L_adj_cpu, L_adj);
     pos = 0;
     for (size_type j = 0; j < M_; ++j)
-      for (size_type i = j; i < M_; ++i)
-        vari_ref_A_[pos++]->adj_ += L_adj_cpu.coeffRef(i, j);
+      for (size_type i = j; i < M_; ++i) {
```
SteveBronder: Same thing here about iterating over packed_size

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: Yeah, definitely. Fixed.

stan/math/rev/mat/fun/cholesky_decompose.hpp
rok-cesnovar: That makes total sense. Fixed.

stan/math/opencl/copy.hpp
seantalts: I think `std::vector` is faster unless you need any of Eigen's features.

stan/math/rev/mat/fun/cholesky_decompose.hpp
seantalts: I think you can only use the reserve pattern with subsequent push_backs, see http://www.gotw.ca/gotw/074.htm

seantalts: specifically this part: 
> In this case, we used v.reserve(2) and therefore know that v.capacity() >= 2, and that's fine as far as it goes -- but we never actually added any elements to v, so v is still empty! At this point v just happens to have room for two or more elements. We can only safely use operator[]() (or at()) to modify elements that are actually present, which means that they count towards size(). At first you might wonder why operator[]() couldn't just be smart enough to add the element if it's not already there, but if operator[]() were allowed to work this way, you could create a vector with "holes" in it!

rok-cesnovar: Thanks Sean. I will switch to using push_back then rather than reverting back to avoid the construction costs.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -321,19 +322,20 @@ class cholesky_opencl : public vari {
    *
    */
   virtual void chain() {
-    auto L_adj_cpu = Eigen::MatrixXd::Zero(M_, M_).eval();
-    auto L_cpu = Eigen::MatrixXd::Zero(M_, M_).eval();
+    int packed_size = M_ * (M_ + 1) / 2;
+    std::vector<double> L_adj_cpu;
+    L_adj_cpu.reserve(packed_size);
+    std::vector<double> L_val_cpu;
+    L_val_cpu.reserve(packed_size);
```
seantalts: The power of benchmarking:
```
---------------------------------------------------------------
Benchmark                     Time             CPU   Iterations
---------------------------------------------------------------
push                       5100 ns         4505 ns       135015
reserve_push               2474 ns         2371 ns       297215
initialize_op_assign        900 ns          892 ns       652486
```
Using this benchmark:
https://github.com/seantalts/perf-math/blob/master/stdvectorfun.cpp

I just whipped that up and it's late here (I'm not in EST, I'm actually in Portugal!) but seems like we should just do the naive thing 😂 

tagging @SteveBronder and honestly if this result holds up it seems like we should publish unintuitive stuff like this somewhere other Stan C++ devs will read...

stan/math/opencl/copy.hpp
seantalts: Or unless you're converting from Eigen types on the way in - assuming we're not expecting to do that right?

rok-cesnovar: Not really, this is intended primarily to be used to copy the vectors of `varis` as in the reverse mode cholesky decompose case.

stan/math/opencl/kernels/unpack.hpp
```diff
@@ -0,0 +1,69 @@
...
+     * @param part parameter that defines the triangularity of the
+     * input matrix
+     *  LOWER - lower triangular
+     *  UPPER - upper triangular
+     * if the part parameter is not specified
+     * @note Code is a <code>const char*</code> held in
+     * <code>unpack_kernel_code.</code>
+     * This kernel uses the helper macros available in helpers.cl.
+     */
+    __kernel void unpack(__global double* B, __global double* A,
```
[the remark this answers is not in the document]
rok-cesnovar: I am not sure I follow. I could add an argument and then choose the operation based on that. The vector is on different sides of the assignment so just replacing A with B wont work.

stan/math/opencl/copy.hpp
```diff
@@ -89,6 +91,69 @@ void copy(Eigen::Matrix<double, R, C>& dst, const matrix_cl& src) {
...
+  if (src.size() > 0) {
+    int packed_size = src.rows() * (src.rows() + 1) / 2;
+    dst.reserve(packed_size);
```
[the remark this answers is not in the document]
rok-cesnovar: I will add a check and throw if its uninitialized (not the expected size, we dont cover that case here). Will also add a separate test to check that it throws in that case.

stan/math/opencl/kernels/unpack.hpp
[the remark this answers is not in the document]
seantalts: Ah yeah, forgot they are stored in a different order. never mind, sorry :)

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
rok-cesnovar: Switched to throw on size mismatch, I think its better that way than to mess with creating. The other option is to switch this to

```cpp
inline std::vector<double> packed_copy(const matrix_cl& src) {
```
```cpp
inline matrix_cl packed_copy(const std::vector<double>& src) {
```

test/unit/math/opencl/copy_test.cpp
```diff
@@ -93,4 +93,63 @@ TEST(MathMatrixGPU, barebone_buffer_copy) {
...
+  for (size_t i = 0; i < packed_mat.size(); i++) {
+    EXPECT_EQ(packed_mat[i], packed_mat_dst[i]);
+  }
+}
+
+TEST(MathMatrixGPU, matrix_cl_pack_unpack_copy_upper) {
+  int size = 51;
+  int packed_size = size * (size + 1) / 2;
+  std::vector<double> packed_mat(packed_size);
+  std::vector<double> packed_mat_dst(packed_size);
```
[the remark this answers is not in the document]
rok-cesnovar: Added tests to check behaviour on unintialized inputs.

rok-cesnovar:
@seantalts You are probably pretty busy at this point, but this is ready for re-review. Other than switching the packed_copy functions to copy on return, I believe it looks good.

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
seantalts: I think we might wanna do the latter: https://isocpp.org/blog/2013/02/no-really-moving-a-return-value-is-easy-stackoverflow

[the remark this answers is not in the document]
rok-cesnovar: Thanks! I changed the interface, way more clean now.

rok-cesnovar:
The "Full unit with GPU" fails with the following:

```
+ echo CXX=clang++-6.0 -Werror
Terminated
+ echo STAN_OPENCL=true
Terminated
+ echo OPENCL_PLATFORM_ID=0
Terminated
[Clang (LLVM based)] Skipping execution of recorder since overall result is 'FAILURE'
Something something emails
```
Does the machine need a reboot or something?

seantalts:
There's a new bug in Jenkins :( I haven't figured out what to do about it yet. Restarting the job seems to work.

rok-cesnovar:
Ah, I only tried restarting the "Always run part 1" and that dint do the job. Didnt think of that, sorry.

seantalts:
If you have time could you make it the two functions return by value? I think that's the cleanest and apparently easiest for the compiler to optimize.

rok-cesnovar:
Like this? https://github.com/stan-dev/math/pull/1164/commits/323ce62b311806671247ef5b189d3e4aedd145e9

Or is there a better solution?

seantalts:
Yes! GitHub/my browser is fucking with me and not updating the PR when you update it. I keep having to do refreshes. Sorry about that.

rok-cesnovar:
No worries, I sometimes get the (1) in the Github tab name, sometimes not it depends.

seantalts:
Can merge as soon as the tests pass. Fingers crossed.

Issue #1168, opened 2019-03-20
Replace OpenCL buffer rectangular copies in kernels with API calls


rok-cesnovar:
## Description

When doing copies of a 2D buffer, you can use the OpenCL provided `enqueueCopyBufferRect`.
This provides a ~50% reduction when copying smaller chunks, less of speedup at larger sizes (see google benchmark results below). Our implementation of the Cholesky decompose does quite a number of these smaller copies so this is worthwhile. This can be used in both `sub_block` as well as `copy`.

## Example

There will be no changes to the API, all under the hood. Listing benchmark results here:
Copies of rectangular shape with varying offset. All in the form A(0:size, 0:size) = B(offset:size, offset:size). The first listed number is the dimension of B, the other is the offset.

```
-------------------------------------------------------------------
Benchmark                            Time           CPU Iterations
-------------------------------------------------------------------
BM_sub_block/20/0                23279 ns      23280 ns      29916
BM_copyBufferRect/20/0           12451 ns      12451 ns      55036
BM_sub_block/100/0               15651 ns      15651 ns      44366
BM_copyBufferRect/100/0          13002 ns      13003 ns      54267
BM_sub_block/200/0               15726 ns      15726 ns      42756
BM_copyBufferRect/200/0          13544 ns      13544 ns      48921
BM_sub_block/400/0               16186 ns      16186 ns      42611
BM_copyBufferRect/400/0          15239 ns      15239 ns      43772
BM_sub_block/500/0               22308 ns      22309 ns      31231
BM_copyBufferRect/500/0          21124 ns      21125 ns      33018
BM_sub_block/20/10               22349 ns      22350 ns      31237
BM_copyBufferRect/20/10          11402 ns      11402 ns      60719
BM_sub_block/100/10              14669 ns      14669 ns      46859
BM_copyBufferRect/100/10         11690 ns      11690 ns      58437
BM_sub_block/200/10              14152 ns      14152 ns      48505
BM_copyBufferRect/200/10         12309 ns      12309 ns      53378
BM_sub_block/400/10              15809 ns      15809 ns      43942
BM_copyBufferRect/400/10         15234 ns      15234 ns      45835
BM_sub_block/500/10              22617 ns      22617 ns      31963
BM_copyBufferRect/500/10         20946 ns      20946 ns      33294
BM_sub_block/100/40              14892 ns      14892 ns      46546
BM_copyBufferRect/100/40         11568 ns      11568 ns      59435
[...]

PR #1169, opened 2019-03-20
Replace the use of OpenCL kernels with API calls for copies of rectangular buffers

rok-cesnovar:
## Summary

This replaces the use of OpenCL kernels with the API calls for rectangular buffers. For triangular copies the kernels are still faster than API+`zeros()` kernel. See #1168 for benchmark results. Will post the benchmarks for triangular copies in a reply below.

## Tests

/

## Side Effects

/

## Checklist

- [x] Math issue #1168 

### DIFF
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -213,27 +213,27 @@ class matrix_cl {
     try {
       if (triangular_view == TriangularViewCL::Entire) {
         cl::size_t<3> src_offset;
-        src_offset[0] = A_i*sizeof(double);
+        src_offset[0] = A_i * sizeof(double);
         src_offset[1] = A_j;
         src_offset[2] = 0;
         cl::size_t<3> dst_offset;
-        dst_offset[0] = this_i*sizeof(double);
+        dst_offset[0] = this_i * sizeof(double);
         dst_offset[1] = this_j;
         dst_offset[2] = 0;
         cl::size_t<3> size;
-        size[0] = nrows*sizeof(double);
+        size[0] = nrows * sizeof(double);
         size[1] = ncols;
         size[2] = 1;
-        cmdQueue.enqueueCopyBufferRect(A.buffer(), this->buffer(),
-              src_offset, dst_offset, size,
-              A.rows()*sizeof(double), A.rows()*A.cols()*sizeof(double),
-              sizeof(double)*this->rows(),
-              this->rows()*this->cols()*sizeof(double));
+        cmdQueue.enqueueCopyBufferRect(
+            A.buffer(), this->buffer(), src_offset, dst_offset, size,
+            A.rows() * sizeof(double), A.rows() * A.cols() * sizeof(double),
+            sizeof(double) * this->rows(),
+            this->rows() * this->cols() * sizeof(double));
       } else {
-              opencl_kernels::sub_block(cl::NDRange(nrows, ncols), A.buffer(),
-                                this->buffer(), A_i, A_j, this_i, this_j, nrows,
-                                ncols, A.rows(), A.cols(), this->rows(),
-                                this->cols(), triangular_view);
+        opencl_kernels::sub_block(cl::NDRange(nrows, ncols), A.buffer(),
+                                  this->buffer(), A_i, A_j, this_i, this_j,
+                                  nrows, ncols, A.rows(), A.cols(),
+                                  this->rows(), this->cols(), triangular_view);
       }
     } catch (const cl::Error& e) {
       check_opencl_error("copy_submatrix", e);
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -200,8 +200,8 @@ class matrix_cl {
    * @param ncols the number of columns in the submatrix
    */
   template <TriangularViewCL triangular_view = TriangularViewCL::Entire>
-  void sub_block(const matrix_cl& A, size_t A_i, size_t A_j, size_t this_i, size_t this_j,
-                 size_t nrows, size_t ncols) {
+  void sub_block(const matrix_cl& A, size_t A_i, size_t A_j, size_t this_i,
+                 size_t this_j, size_t nrows, size_t ncols) {
     if (nrows == 0 || ncols == 0) {
       return;
     }
@@ -212,9 +212,12 @@ class matrix_cl {
     cl::CommandQueue cmdQueue = opencl_context.queue();
     try {
       if (triangular_view == TriangularViewCL::Entire) {
-        cl::size_t<3> src_offset = opencl::to_size_t({A_i * sizeof(double), A_j, 0});
-        cl::size_t<3> dst_offset = opencl::to_size_t({this_i * sizeof(double), this_j, 0});
-        cl::size_t<3> size = opencl::to_size_t({nrows * sizeof(double), ncols, 1});
+        cl::size_t<3> src_offset
+            = opencl::to_size_t({A_i * sizeof(double), A_j, 0});
+        cl::size_t<3> dst_offset
+            = opencl::to_size_t({this_i * sizeof(double), this_j, 0});
+        cl::size_t<3> size
+            = opencl::to_size_t({nrows * sizeof(double), ncols, 1});
         cmdQueue.enqueueCopyBufferRect(
             A.buffer(), this->buffer(), src_offset, dst_offset, size,
             A.rows() * sizeof(double), A.rows() * A.cols() * sizeof(double),
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -200,8 +200,8 @@ class matrix_cl {
    * @param ncols the number of columns in the submatrix
    */
   template <TriangularViewCL triangular_view = TriangularViewCL::Entire>
-  void sub_block(const matrix_cl& A, int A_i, int A_j, int this_i, int this_j,
-                 int nrows, int ncols) {
+  void sub_block(const matrix_cl& A, size_t A_i, size_t A_j, size_t this_i, size_t this_j,
+                 size_t nrows, size_t ncols) {
     if (nrows == 0 || ncols == 0) {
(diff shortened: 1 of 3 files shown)


stan/math/opencl/matrix_cl.hpp
```diff
@@ -211,10 +211,30 @@ class matrix_cl {
     }
     cl::CommandQueue cmdQueue = opencl_context.queue();
     try {
-      opencl_kernels::sub_block(cl::NDRange(nrows, ncols), A.buffer(),
-                                this->buffer(), A_i, A_j, this_i, this_j, nrows,
-                                ncols, A.rows(), A.cols(), this->rows(),
-                                this->cols(), triangular_view);
+      if (triangular_view == TriangularViewCL::Entire) {
+        cl::size_t<3> src_offset;
```
rok-cesnovar: I could find a way to initialize this cl::size_t vectors in a shorter manner :/

rok-cesnovar:
The benchmark file: https://gist.github.com/rok-cesnovar/a9e7196bcd0730f29ca2591f54fd3a10
The benchmark times for the triangular copies to confirm its not faster: 
[triangular_copies.txt](https://github.com/stan-dev/math/files/2989957/triangular_copies.txt)

stan/math/opencl/matrix_cl.hpp
seantalts: Hah. this is a silly problem with their API it seems. I think the general purpose solution is something like:
```cpp

template <typename Arg, typename... Ts, typename std::enable_if<std::is_integral<Arg>::value>::type * = nullptr>
auto to_size_t(Arg a, Ts... all) {
  constexpr size_t len = sizeof...(Ts)+1;
  int values[len] = {a, all...};
  cl::size_t<len> s;
  for (size_t i = 0; i < len; i++)
    s[i] = values[i];
  return s;
}
```
(we do all the variadic weirdness to [make sure we're getting ints at compile time](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL50-CPP.+Do+not+define+a+C-style+variadic+function)). 

this one is a little hackier and if you wanted to pass more than three, you'd have to specify a template parameter. But it's much less confusing imo and is probably fine forever for us:
```cpp
template <int len=3>
cl::size_t<len> to_size_t(std::vector<int> values) {
  assert(values.size() ==len);
  cl::size_t<len> s;
  for (size_t i = 0; i < len; i++)
    s[i] = values[i];
  return s;
}
```

Called like this:
```cpp

int main() {
  cl::size_t<3> s = to_size_t({1, 2, 3});
  return s[0];
}
```

rok-cesnovar: Thanks Sean!! I am going to try and incorporate the second one today, I thinks this one is going to be fine forever, for sure :)

rok-cesnovar:
I added the `to_size_t` to the context file as it maybe come in handy in other places. If that is OK with you, please approve again and I will merge once it passes. Thanks again for the help!

rok-cesnovar:
@seantalts sooo, this one is weird. 

on a routine `./runTests.py a test/unit -f opencl` after I made some unrelated changes I got this message:

```
In file included from test/unit/math/opencl/opencl_context_test.cpp:2:0:
./stan/math/opencl/opencl_context.hpp: In function ‘cl::size_t<len> stan::math::opencl::to_size_t(std::vector<long unsigned int>)’:
./stan/math/opencl/opencl_context.hpp:43:3: error: there are no arguments to ‘assert’ that depend on a template parameter, so a declaration of ‘assert’ must be available [-fpermissive]
   assert(values.size() == len);
   ^~~~~~
./stan/math/opencl/opencl_context.hpp:43:3: note: (if you use ‘-fpermissive’, G++ will accept your code, but allowing the use of an undeclared name is deprecated)
```
This also happens on develop, although not on Jenkins apparently. So I thought it must be clang. clang then gave this error: 
```
In file included from test/unit/math/opencl/opencl_context_test.cpp:2:
./stan/math/opencl/opencl_context.hpp:43:3: error: use of undeclared identifier 'assert'
  assert(values.size() == len);
```
Google told me to add #include <cassert> and after I did that, no more errors.

So the extra weird is that this passed Jenkins, and this exact same code compiles with every other test 
 that includes the same file except for `opencl_context_test`.

Am I missing something? If not, what is the process here? Revert this PR or open a new one that fixes this? Anyways, I apologize...

bob-carpenter:
I had thought we'd decided *not* to include C-style asserts in any of our code. 

If that test is to catch user errors, it should throw rather than assert.  If it's there to catch programmer errors, it should be replaced with unit tests.

Hopefully, @syclik and @seantalts will weigh in on policy and correct me if I'm misremembering.

seantalts:
I haven't heard about the assert policy. This one is to help developers and I'm not sure how to replace it with unit tests.

seantalts:
> Am I missing something? If not, what is the process here? Revert this PR or open a new one that fixes this? Anyways, I apologize...

I think we should roll forward since the build isn't broken. Bob, what would you do here? We're just creating a simple helper function that creates a `cl::size_t`, which is kind of an annoying type to create. We talked about two solutions here: https://github.com/stan-dev/math/pull/1169#discussion_r267701100 one of which is fairly complicated templating that is always correct and doesn't have an assert, and the other is simpler but has an assert. Is there a way to get rid of the assert there? Or should we go with the full template solution? or some other thing?

bob-carpenter:
> I think we should roll forward since the build isn't broken. Bob, what would you do here? We're just creating a simple helper function that creates a cl::size_t, which is kind of an annoying type to create. We talked about two solutions here: #1169 (comment) one of which is fairly complicated templating that is always correct and doesn't have an assert, and the other is simpler but has an assert. Is there a way to get rid of the assert there? Or should we go with the full template solution? or some other thing?

The problem with asserts is that they can catch programmer error, but their only behavior at runtime is to crash the environment (and report which asssert failed).   Do you anticipate the assert being triggered at runtime?  If so, I'd suggest throwing an exception instead, and if not, I'd suggest removing it since it's not doing anything.

seantalts:
I don't anticipate it happening at runtime but I do anticipate it happening during development...

syclik:
I think Bob’s remembering right. I don’t think we ever came up with an
explicit policy, but we haven’t had a good example where we needed to use
an assert where a runtime exception would have functioned more
appropriately.

Currently, if code asserts, what’s the downstream effect? RStan would just
crash? Would R crash too? Or would it treat it like an error and move on?

seantalts:
Do we have an exception we use for Stan developer error? I guess we could use that? Or else just code it the dumb way for the specific case of 3 inputs:
```cpp
auto to_size_t(int a, int b, int c) {
  cl::size_t<3> s;
  s[0] = a;
  s[1] = b;
  s[2] = c;
  return s;
}

int main() {
  cl::size_t<3> s = to_size_t(1, 2, 3);
  return s[0];
}
```

syclik:
So far, we've been squeezing our exceptions into things that subclass `std::exception`: https://en.cppreference.com/w/cpp/error/exception

It's been ok, but there's no reason we can't define our own subclass. I haven't followed the content of where this came up from, so I don't have a good suggestion. Hopefully one of them makes sense. If not, create one that's sensible?

seantalts:
I'm going to put up a PR changing it to the correct-but-complicated templated version, above. If we have a policy on assert, maybe we can add a test for it as well similar to the math dependencies one?

seantalts:
Still weirded out about why it passed on Jenkins before:
> So the extra weird is that this passed Jenkins, and this exact same code compiles with every other test that includes the same file except for opencl_context_test.

Rok did you figure out why? The "Full Unit with GPU" tests run on every pull request now, so I don't think that's it...?

syclik:
Let me know if you want extra eyes on it. (Just tag me if you need it.)

rok-cesnovar:
@seantalts unfortunately no. 

this doesnt error out on this branch 

```
/usr/local/opt/llvm@6/bin/clang++ -Werror -std=c++1y -Wno-unknown-warning-option -Wno-tautological-compare -Wno-sign-compare    -I lib/opencl_1.2.8  -O3  -I . -I lib/eigen_3.3.3 -I lib/boost_1.69.0 -I lib/sundials_4.1.0/include -I lib/gtest_1.8.1/include -I lib/gtest_1.8.1 -I lib/gtest_1.8.1/include -I lib/gtest_1.8.1      -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION  -DSTAN_OPENCL -DOPENCL_DEVICE_ID=1 -DOPENCL_PLATFORM_ID=0 -DCL_USE_DEPRECATED_OPENCL_1_2_APIS  -DGTEST_USE_OWN_TR1_TUPLE -DGTEST_USE_OWN_TR1_TUPLE  -c -o test/unit/math/opencl/sub_block_test.o test/unit/math/opencl/sub_block_test.cpp
```

This failed on Sebastian's PR:

```
clang++-6.0 -Werror -std=c++1y  -Wno-sign-compare   -I lib/opencl_1.2.8  -O3  -I . -I lib/eigen_3.3.3 -I lib/boost_1.69.0 -I lib/sundials_4.1.0/include -I lib/gtest_1.8.1/include -I lib/gtest_1.8.1 -I lib/gtest_1.8.1/include -I lib/gtest_1.8.1      -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION  -DSTAN_OPENCL -DOPENCL_DEVICE_ID=0 -DOPENCL_PLATFORM_ID=0 -DCL_USE_DEPRECATED_OPENCL_1_2_APIS  -DGTEST_USE_OWN_TR1_TUPLE -DGTEST_USE_OWN_TR1_TUPLE  -c -o test/unit/math/opencl/sub_block_test.o test/unit/math/opencl/sub_block_test.cpp
```

The other difference that in this PR it was on the Always run part 1, on the other PR it was Always run part 2. Not sure if that makes any difference.

Issue #1134, opened 2019-03-02
map_rect with threading broken on windows


wds15:
## Description
Programs which use `map_rect` with threading on Windows with the gcc 4.9.3 do not work. Programs compile, but once execution reaches the threaded code parts, the program just terminates

## Example

Execute on windows with RTools (gcc 4..9.3) and threading configured the following:
```
STAN_NUM_THREADS=4 /runTests.py test/unit/math/rev/mat/functor/gradient_test.cpp
```

This will abort execution once the threaded tests are reached.

## Expected Output
It should just run.

#### Current Version:
v2.18.1

Issue #1102, opened 2019-01-15
Thread performance penalty


wds15:
## Description
Enabling threading in the `stan-math` library with the `STAN_THREADS` compiler macro leads to significant slow-downs. This is caused by the `thread_local` storage (TLS) implementation used in a Mayer singleton approach for the AD tape.

## Example

Running the performance cmdstan benchmark suite (for vanilla cmdstan 2.18.1) with 5 replications on the `stat_comp_benchmark` set of examples produces on a MacOS Mojave, `clang++ Apple LLVM version 10.0.0 (clang-1000.11.45.5)`, 2.9 GHz Intel Core i9 system these results without TLS (so `STAN_THREADS` not being defined):

```
~/work/performance-tests-cmdstan]$ cat performance_2_18_1.csv
stat_comp_benchmarks/benchmarks/garch/garch.stan,0.371638154984
stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan,7.15906739235
stat_comp_benchmarks/benchmarks/sir/sir.stan,119.077228165
stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan,0.0128675460815
stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan,2.5181746006
stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan,0.078618812561
stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan,0.181181192398
stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan,0.0303950309753
stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan,0.271839380264
stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan,18.0576345444
stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan,5.63970680237
stat_comp_benchmarks/benchmarks/arK/arK.stan,1.43102579117
stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan,2.22463345528
stat_comp_benchmarks/benchmarks/arma/arma.stan,0.51917424202
```

The same benchmark runs *with* `-DSTAN_THREADS` turned on with these metrics:

```
[...]

PR #1170, opened 2019-03-21
WIP Feature/faster ad tls v3B

wds15:
This PR is a continuation of the reverted one from this merged PR: https://github.com/stan-dev/math/pull/1135

(I am not sure why git is unhappy and is in a state of not wanting to merge... do I need to rebase?)

**I am still in the process to sort things, but I want to see the tests on Jenkins.**

This PR speeds up the access to the AD tape whenever `thread_local` storage (TLS) is needed for the AD tape. In addition programs with TLS now work on Windows with gcc 4.9.3.

## Summary

This PR has the main goal to reduce the performance penalty implied by turning on thread capability of stan-math. As it turns out, the proposed implementation works out of the box on Windows with the gcc 4.9.3 compiler which is not the case with the current threading implementation in develop. So we fix a bug which affects many users given the popularity of Windows. The Windows bug is tickled in the commit a7bf7782432d57c195cc2fe90e9eb3fe7dffbfe1 on Jenkins and it is fixed in the current form of the PR.

The performance benefits of this PR are documented in the [previous PR](https://github.com/stan-dev/math/pull/1103) which served as a first prototype. In summary, we loose ~20% on develop when turning on threading while with this PR we reduce this to less than 3% on average.

The key change of this PR is to change how the global AD tape instance is declared and accessed. Currently in develop the global AD tape is declared 

- for the non-threading case as a static global instance.
- for the threading case as a TLS which is declared as a static instance of a static function (Meyer singleton design).

**This PR changes this to a global pointer to the AD tape, which is a TLS for threading, and which is initialized to the constant `nullptr` expression at compile-time**. Initialization to a constant expression at compile-time (as opposed to a dynamic expression right now in develop) allows the compiler to use the TLS global pointer more efficiently. Furthermore, the `thread_local` keyword used to declare TLS can be swapped for the compiler-specific `__thread` keyword whenever the expression is constant on gnu extension compatible compilers. The `__thread` keyword has been around much longer than the C++11 `thread_local` keyword and only the `__thread` keyword guarantees to the compiler the the variable is (in fact must be) initialized with a constant expression which allows the use of fast TLS code.

**The runtime initialisation of the AD tape for the main process is automatically achieved through the instantiation of a global object** which manages the life-time of the AD stack (as performed in the new file `math/stan/math/rev/core/init_chainablestack.hpp`).

## Side-effects

There is no change in behavior for programs whenever `STAN_THREADS` is not defined. If `STAN_THREADS` is defined, then any C++ client library which uses the stan-math functions (`map_rect` in particular) will continue to work unchanged. However, C++ clients which use AD in child threads will stop working. The reason is that the AD tape requires initialization in child threads. The initialization of the AD tape can be done by constructing a `ChainableStack` instance which will trigger AD tape initialization for the thread. The first instantiation of `ChainableStack` in a given child thread will take ownership of the AD tape. Thus only the first created instance will free the AD tape resource upon destruction.

## Tests

`test/unit/math/rev/core/thread_stack_instance_test.cpp` is a new test and is written to test matters with and without threading. The test is setup to test different things depending on enabling threading or not.

- STAN_THREADS NOT defined (no threading). In this case we only have a global AD tape. All child-threads created by a given program will see the same AD tape. Thus initializing the AD tape on the main thread implies that it is initialized on the child-thread, etc.

- STAN_THREADS IS defined (threading case). Now all new threads created must see their own TLS of the AD tape pointer which is not initialized at the beginning and is independent of the main thread AD tape when performing AD operations.

In addition I have added to the `map_rect_concurrent` tests code which enforces that these tests are always run with 4 threads regardless of the `STAN_NUM_THREADS` definition (which is often not set and this things do not run threaded during testing).

## Checklist

- [X] Math issue #1102 & #1134 

## Remaining TODO

- [x] @seantalts will review the code at the implementation level.
- [ ] We will run these benchmarks after we've caught up to develop (and hopefully after we've made any additional code changes):
    - [ ] microbenchmark for develop vs this branch without any threading. This will be run on Window, Linux, Mac for the compilers we've said we'd support (one per OS).
    - [x] @wds15's non-stiff integrator example for develop vs this branch with threading. This will run on Linux and Mac for the compilers we've said we'd support. (Windows does not work currently on develop.)
- [ ] If all of the benchmarks show no more than 5% speed regression, then it's acceptable performance-wise. (We have every reason to believe that we'll get a large speed benefit for threading.)
- [x] @syclik will run the benchmarks for the non-stiff integrator example by Monday. If he's unable to run the benchmarks by Monday, we will use the microbenchmark results only.

## End-to-end performance results

@syclik ran @wds15's benchmark on Mac with clang++ 6 and Linux with g++ 5 using these scripts: [syclik/benchmark-stan](https://github.com/syclik/benchmark-stan/tree/math/1135-ad). These results were generated on the git hash 95844ac using the all pointer design (with an instance() method abstraction).

### Summary

This PR is definitely faster for threading (compared to develop with threading)!

* Linux, g++ 5:
  * 1 thread: 18% faster
  * 2 threads: 12% faster
  * 4 threads: 13% faster
* Mac, clang++ 6:
  * 1 thread: 27% faster
  * 2 threads: 26% faster
  * 4 threads: 29% faster

### Data (time in seconds per run)

| os | threads | branch	| n | mean | sd	| min |	max |
|----|---------|--------|---|------|----|-----|-----|
| linux | 1 | new | 10 | 204.9988 |  3.7411436 | 201.402 | 215.134 |
| linux | 1 | old | 10 | 250.7093 |  1.0396194 | 249.544 | 252.021 |
| linux | 2 | new | 10 | 228.5871 |  0.5270710 | 228.024 | 229.507 |
| linux | 2 | old | 10 | 259.4587 |  1.7354262 | 257.465 | 262.409 |
| linux | 4 | new | 10 | 156.1332 |  0.1720515 | 155.854 | 156.346 |
| linux | 4 | old | 10 | 180.4681 |  0.2238524 | 180.118 | 180.887 |
|   mac | 1 | new | 10 | 302.0826 | 13.1204998 | 289.895 | 334.361 |
|   mac | 1 | old | 10 | 412.2389 |  7.1865066 | 403.996 | 425.580 |
|   mac | 2 | new | 10 | 167.9633 |  3.5238670 | 161.854 | 173.285 |
|   mac | 2 | old | 10 | 225.8895 |  4.1751089 | 218.869 | 231.755 |
|   mac | 4 | new | 10 |  96.2758 |  1.6974981 |  93.061 |  98.700 |
|   mac | 4 | old | 10 | 134.9916 |  3.3910946 | 129.339 | 141.984 |

(I still wouldn't suggest using this for single-threaded mode. I haven't benchmarked as thoroughly since it's not what this is about, but on Linux, it's about break even, but on Mac, it's about 10% slower.)

### DIFF
--- a/stan/math/rev/arr/fun/sum.hpp
+++ b/stan/math/rev/arr/fun/sum.hpp
@@ -29,7 +29,7 @@ class sum_v_vari : public vari {
 
   explicit sum_v_vari(const std::vector<var>& v1)
       : vari(sum_of_val(v1)),
-        v_(reinterpret_cast<vari**>(ChainableStack::instance().memalloc_.alloc(
+        v_(reinterpret_cast<vari**>(ChainableStack::instance_->memalloc_.alloc(
             v1.size() * sizeof(vari*)))),
         length_(v1.size()) {
     for (size_t i = 0; i < length_; i++)
--- a/stan/math/rev/core/autodiffstackstorage.hpp
+++ b/stan/math/rev/core/autodiffstackstorage.hpp
@@ -9,12 +9,10 @@ namespace math {
 
 /**
  * Provides a thread_local singleton if needed. Read warnings below!
- * With STAN_THREADS defined, the singleton is a thread_local static pointer
- * for performance reasons. When STAN_THREADS is not set, we have the old
- * static AD stack in the instance_ field because we saw odd performance
- * issues on the Mac Pro[4]. The rest of this commentary is specifically
- * talking about the design choices in the STAN_THREADS=true case.
- * When a TLS is used then initialization with
+ * For performance reasons the singleton is a global static pointer
+ * which is stored as thread local (TLS) if STAN_THREADS is
+ * defined. The use of a pointer is motivated by performance reasons
+ * for the threading case. When a TLS is used then initialization with
  * a constant expression is required for fast access to the TLS. As
  * the AD storage struct is non-POD it must be initialized as a
  * dynamic expression such that compilers will wrap any access to the
@@ -61,7 +59,6 @@ namespace math {
  * [2] https://github.com/stan-dev/math/pull/826
  * [3]
  * http://discourse.mc-stan.org/t/potentially-dropping-support-for-older-versions-of-apples-version-of-clang/3780/
- * [4] https://github.com/stan-dev/math/pull/1135
  */
 template <typename ChainableT, typename ChainableAllocT>
 struct AutodiffStackSingleton {
@@ -70,12 +67,10 @@ struct AutodiffStackSingleton {
 
   AutodiffStackSingleton() : own_instance_(init()) {}
   ~AutodiffStackSingleton() {
-#ifdef STAN_THREADS
     if (own_instance_) {
       delete instance_;
       instance_ = nullptr;
     }
-#endif
   }
 
   struct AutodiffStackStorage {
@@ -96,21 +91,15 @@ struct AutodiffStackSingleton {
   AutodiffStackSingleton &operator=(const AutodiffStackSingleton_t &) = delete;
 
   static constexpr inline AutodiffStackStorage &instance() {
-    return
-#ifdef STAN_THREADS
-        *
-#endif
-        instance_;
+    return *instance_;
   }
 
  private:
   static bool init() {
-#ifdef STAN_THREADS
     if (!instance_) {
       instance_ = new AutodiffStackStorage();
       return true;
     }
-#endif
     return false;
   }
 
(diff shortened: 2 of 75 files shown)


seantalts:
Hm, I think the best way to get this PR started is to start from `develop`, `git revert 883d78193718e7d5105a2ba4711a2166e8080cc`, and then start from there (and you can revert other individual commits from the old PR if you like then as well, I think).

wds15:
Ok... closing this one... now we can look at v4 of this.

PR #1172, opened 2019-03-22
New version of `to_size_t` to get rid of an `assert`

seantalts:
We had a discussion here https://github.com/stan-dev/math/pull/1169 about potentially having a policy advocating for no `assert`s, which led to this PR.

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses: Columbia University


stan/math/opencl/opencl_context.hpp
```diff
@@ -35,14 +35,12 @@ namespace stan {
...
- * to cl::size_t<3>
+ * A helper function to convert an array to a cl::size_t<N>
  */
-template <size_t len = 3>
-cl::size_t<len> to_size_t(std::vector<size_t> values) {
-  assert(values.size() == len);
-  cl::size_t<len> s;
-  for (size_t i = 0; i < len; i++)
+template <int N>
+auto to_size_t(const size_t (&values)[N]) {
```
syclik: Are there any downstream tests that should fail from this not asserting?

Is there some reasonable limit N should be? That’s be nice to doc since this is a template type. (Unless c++14 now goes with that being a const expr instead of what it used to do.... still catching up on changes to the lang)

rok-cesnovar: The way cl::size_t si defined now, N is always going to be 3. This cl::size_t is really weirdly defined and was a topic of a lot of discussions on the Khronos group dev boards. You can see why.

test/unit/math/opencl/opencl_context_test.cpp
```diff
@@ -4,6 +4,13 @@
 #include <vector>
 #include <fstream>
 
+TEST(MathGpu, to_size_t_test) {
+  cl::size_t<3> a = stan::math::opencl::to_size_t<3>({1, 2, 3});
```
rok-cesnovar: @syclik would this do? Or were you thinking of something more complicated?

stan/math/opencl/opencl_context.hpp
syclik: Ah. I see. If that’s the intention, then one way that might make sense is to make the base template throw (and not have a meaningful implementation). And then have a template specialization for N = 3. That way you get the best of both worlds. 

There’s a potentially better solution, but I don’t know if it’s possible. For the base template, have it not be compilable. That’ll stop things at compile time. But once again, not sure if that’s easy to set up.

syclik: And if you don’t want that, just adding that info to the function doc would be useful to anyone that comes across this in the future.

test/unit/math/opencl/opencl_context_test.cpp
syclik: Yes, this will do! If, for some unseen reason, someone decides to mess with the implementation, this will prevent that from happening.

bob-carpenter:
Yes, and thanks!

rok-cesnovar:
There were missing inlines and template parameters I missed :/, please review again @syclik when you have the time.

SteveBronder:
Semi related to this PR, can we a check for asserts with cpplint?

syclik:
@SteveBronder: not sure. If it's possible, it'd be in the cpplint doc.

Issue #1179, opened 2019-03-24
Add Intel TBB library


wds15:
## Description
The Intel TBB (Threading Building Blocks) is used a lot in scientific computing to add optional parallelization to existing programs in a C++ framework. Key benefits for stan-math are:

- faster malloc functions which are beneficial for stan-math and in particular for threading
- work stealing queuing facilities which will optimally distribute work packets
- no more need to focus on programming with threads - the library rather abstracts those details away
- lot's of engineering went into the library to make best use of cache-locality and what not else is out there to take into account for threading.

Additional info is available at:

https://github.com/stan-dev/math/wiki/Dependency-Checklist-for-the-Intel-TBB

## Example

The first examples where we will use the library are:

- use it as a drop-in replacement for malloc to speedup programs
- code `map_rect` on the basis of `tbb::parallel_for` to get automatic sharding by the work stealing queuing.

## Expected Output

Things will run faster.

#### Current Version:
v2.19.0

rok-cesnovar:
Not sure on all the TBB details yet, so posting this here if anyone has any idea. Could we do "speculative execution" with TBB? The largest use case for this would be in the GPU code, but it could be used elsewhere. 

So if we have 2 ways of computing something we could spawn 2 threads and whichever thread finishes successfully first can stop the other thread and we can continue on? Or do we have to wait for both threads to finish before going forward.

This wont work with anything that changes the memory arena but could potentially be plausible for everything else?

wds15:
That's very aggressive parallelisation, but I would guess that you can do that... I would check the docs of the TBB for the term of a `task_group`.

wds15:
addressed with PR #1180

PR #1180, opened 2019-03-24
Feature/intel tbb lib

wds15:
## Summary

Adds the Intel TBB 2019 Update 8 to stan-math's makefile system.

The new dependency checklist has been [filled out on the wiki](https://github.com/stan-dev/math/wiki/Dependency-Checklist-for-the-Intel-TBB).

The Intel TBB is dynamically linked to programs. For robust linkage with the final programs, the `rpath` facility of linkers are used. This hard-codes the absolute path to the TBB library into the final binaries such that no additional user settings have to be made.

However, the Windows platform is special in two ways:

1. The `rpath` linking logic is not supported. Instead, the `PATH` variable of the user must include the directory of the binary TBB libraries.
2. On Windows the make variant `mingw32-make` must be used to build the Intel TBB since only this variant of make is supported by the Intel TBB sources.

The Intel TBB is introduced to the C++ stan-math sources in the PR https://github.com/stan-dev/math/pull/1376

## Tests

The Intel TBB has been added as dependency to the tests such that the test itself is if it builds ok.

## Side Effects

- On Windows the `mingw32-make` has to be used from the RTools instead of the `make` command. Otherwise the TBB makefiles do not work.

- On Windows the `PATH` variable must include the directory where TBB binaries are stored which is configured to be by default `stan-math/lib/tbb`.

## Checklist

- [X] Math issue #1179 

- [X] Copyright holder: Sebastian Weber (copyright for the Intel TBB is with Intel, of course)


SteveBronder:
Q, is this going to be bad for non-Intel systems?

bgoodri:
Is there a restriction on the bitness of the Stan program when linking with an Intel TBB that was build with 32 bit mingw? I guess we are only using 32 bit integers at the moment, but I had started to write a script to convert to 64 bit integers.

bgoodri:
@SteveBronder there are a few mentions of changes to support ARM on the Intel TBB change log
https://github.com/01org/tbb/blob/tbb_2019/CHANGES
and the RcppParallel package has been including (an older version of) this for a while so I am pretty sure it is okay on SPARC.

bgoodri:
Also, Intel's Parallel STL, which is intended to be an implementation of the C++17 standard, uses TBB and Intel gave it to the LLVM repo and LLVM works fine on non-Intel hardware AFAIK.

wds15:
@bgoodri I am not sure why 64 bit integers are a consideration from your perspective for the Intel TBB. I don't see how that interacts, but I don't know. I guess we are good; not sure though.

@SteveBronder No, this is not bad news for AMD, for example. Reasons:

- Intel TBB is a C++ library. Any compiler you use will make optimised assembler out of it optimal for whatever processor it should run on
- The LLVM project will use the TBB as basis for their C++17 implementation of the parallel STL bits
- The gcc project does the same

What is certainly correct is that there are some ifdefs in their code make things work super-optimal on intel compilers and moreover the unit tests for the library are executed on Intel hardware only.

I asked myself this questions and for the above reasons I came to the conclusion we are good.

wds15:
@seantalts Does Jenkins wipe out all previous results between commits when retesting? I am asking as I wonder if I need to do a manual clean once in a while. Thanks.

bgoodri:
Like if Stan moves to 64 bit integers and someone passes `x_i` to `map_rect`, will their computer explode? I suspect not because other projects (albeit not R) must have been using 64 bit integers with TBB sometime over the past decade.

seantalts:
> @seantalts Does Jenkins wipe out all previous results between commits when retesting? I am asking as I wonder if I need to do a manual clean once in a while. Thanks.

It very much should, but I have seen (rare) cases of a job exiting in a really unusual way (maybe the machine drops out of the network or the whole process segfaults) and then there can be stuff left to clean up. That said, I think it should get a new workspace even in that situation.

wds15:
I really do not see how 64bit integers are a problem for the TBB. So, yes, this will work to my expectation.

wds15:
Whow... this thing builds now on Windows (!) and under Mac just fine. On Linux this works as well, but with MPI in conjunction there are still some problems which I will look at later. In any case, the WIP status can be removed soon.

seantalts:
Have you noticed that the builds on this are taking ~7h instead of ~1h? Any idea why that would be?

wds15:
Argh... it was a nightmare getting the TBB to work on Windows and with the specific Stan compiler. Long compile times sound like another "fun" to look into. I will try to reproduce the problem on my system. I haven't noticed any compile time increases when using the TBB yet.

wds15:
@seantalts & @serban-nicusor-toptal is our Jenkins down? I cannot reach the web-site atm.

seantalts:
Yep, looks like master is down.

wds15:
Yeah!!! The Linux Unit with MPI stage just finished in 50min... that look promising now, right @seantalts ?

(I had to remove a `.NOTPARALLEL` statement in the makefile which I replaced with a respective dependency setup)

wds15:
Hmm... to get this in I would need to have a PR for cmdstan as well to set

`TBB_CXX_TYPE` 

during the MPI build of cmdstan. Right now I am facing

```
stan/lib/stan_math/make/libraries:84: *** "Need to set TBB_CXX_TYPE for non-standard compiler.".  Stop.
```

which is OK as I am triggering this myself. Should I open a PR on cmdstan as well?

(we actually should also make sure the stan tests include the TBB targets)

seantalts:
Yeah, I think that makes sense. And then you can trigger a Jenkins build that uses that PR number.

SteveBronder:
Creeping on new gcc 9.1 release notes

```
Improved support for C++17, including:
    The C++17 implementation is no longer experimental.
    Parallel algorithms and <execution> (requires TBB 2018 or newer).
    <memory_resource>.
    Using the types and functions in <filesystem> does not require linking with -lstdc++fs now.
```

Though idt llvm has it's pstl stuff merged yet

SteveBronder:
Looks like we can use TBB with OpenCL as well!

https://software.intel.com/en-us/blogs/2015/12/09/opencl-node-overview

https://www.threadingbuildingblocks.org/docs/help/reference/appendices/preview_features/flow_graph/opencl_node_cls.html

Once tbb is merged I'm going to start looking into how we can use tbb in the OpenCL stuff. Main thing for me is that they have a load balancer algo already implemented that I'm pretty sure will work for multiple OpenCL devices.

wds15:
I am not sure where we can have a discussion on including the tbb, but I wanted to have this recorded what I recently shared by mail with @syclik 

The tbb: I have brought it up a few times in stan meetings. In one you were part of it and that formed the dependency checklist which I filled out for the tbb on the wiki. Apart from that I asked Ben if the tbb will work with rstan and he said „to make it work“. I am not sure about pystan. The pr is still on wip status which I wanted to lift once the design rfc was done. The pr is very progressed. So everything works for cmdstan. The tests still fail, because the upstream stan and cmdstan makefiles need easy tweaks.

I am not sure what is mIssing wrt to the tbb and I thought we would continue the tbb discussion on the tbb pr, but I am not sure on that either. Suggestions are very welcome.

We need to decide if

- we can use the tbb like boost such that it is mandatory to build it as a dependency for every stan program
- we make only optional use of it (likely with STAN_THREADS)
- we cannot use it...hopefully this is not the case

The tbb pr lists the main side effect right now: on windows i need to use mingw32-make which is a more capable (more make unix like) flavor of make. This is due to using the tbb makefiles which only work on windows with this make ... which is part of rtools, so it should be fine.

As a really nice side effect we get on macos about 15% speedup by using the intel tbb malloc on the sir model (other models speedup as well).

Wherever its appropriate to continue the discussion...not sure. Some thoughts and  from pystan devs would also be nice, of course.

bgoodri:
I am just assuming rstan will work if I link models to the shared TBB
object made by the RcppParallel package.

wds15:
...which would mean that we have to use the tbb version of rcppparallel...right?

bgoodri:
Or get rstudio to update rccparallel. It is a few versions behind but I
don't think the api has changed.

wds15:
@riddell-stan & @ahartikainen what are your thoughts?

@bgoodri ok...i will see if Rstudio does an update. Right now they have tbb 2018u1 which is 2y old and that should work, but there seem to be some improvements by now. In any case, i will give this a test drive to check the plan works.

wds15:
Looks like for python there is tbb4py available...would that be an option to link against?

@bgoodri can we distribute the headers of a newer tbb with stanheaders, but link against what is in rcppparallel? Intel maintains binary compatibility as far as I understood (need to verify though).

bgoodri:
Yeah, you can put whatever you want into lib and StanHeaders will just
ignore it like it does with Boost, Eigen, etc.

ahartikainen:
Hmm, all of this is going on in C++, so I don't think python touches anything here.

For compilers, does this work with mingw-w64 5.3? (Python cannot use RTools on Windows)

Not a question for python, but I rarely use RTools on my systems:

make, what version should be used on Windows?

wds15:
@ahartikainen Sorry, I was maybe assuming wrong things about Python. Under R the plan is to make the `RcppParallel` package a dependency of `StanHeaders`. Then we can just link against the already build libraries from that package and thus we do not need to handle the build on our end.

As I understand - you will have to build the TBB as part of the PyStan package - is that right? If so, then the `tbb4py` package can provide a good template as how to handle the build; at least I would hope for that. If you could have a look at the `tbb4py` package to check that building the TBB for Python is straightforward, that would be great?

How do you currently handle the CVODES build? Do you use the makefiles from cmdstan or did you setup the build yourself?

The gcc mingw-w64 5.3 should be supported, yes. At least this is included in what Intel writes they support. For CmdStan I have managed to get things build on Windows using the `mingw32-make`; could you check if that's available with the mingw-w64 distribution? It definitely should be around (a `mingw64-make` should also do, I suppose).

ahartikainen:
> How do you currently handle the CVODES build? Do you use the makefiles from cmdstan or did you setup the build yourself?

We build it ourselves. Or at least that was the case when it was C++ only. Now that it contains C, there is some problems (at least with `clang`) so it is broken and we have had other things to do before fixing it.

https://github.com/stan-dev/pystan/pull/518

ahartikainen:
> At least this is included in what Intel writes they support. For CmdStan I have managed to get things build on Windows using the `mingw32-make`; could you check if that's available with the mingw-w64 distribution?

Ok, I checked and yes, it comes with the mingw-w64. 

    mingw32-make --version
    # GNU Make 4.1.90

I also have make from `conda install make -c conda-forge`

    make --version
    # GNU Make 4.2.1

wds15:
@ahartikainen That sounds very good. So you should be alright wrt to the TBB. Is it an option for you to use the makefiles from the TBB as I am doing it as part of cmdstan? Or is probably the easier approach to just mimic what `tbb4py` does?

I am not sure why a C compiler is an issue, but let me ask on the issue you quote.

wds15:
Yeah! Finally the Jenkins light turns green... and the performance tests show an increase in performance of 6% on average with some models speeding up by 18% even. Note that the convention from the stan performance bot is old / new (so larger is better) while I did report in the PR description the reverse of new / old.

In short: We can speedup stan single-core performance by just linking the tbbmalloc_proxy library.

syclik:
@wds15, can you update the issue / pr top-level comment with a summary? It looks like the dependency checklist is still incomplete.

syclik:
I'm going to mark up the wiki with bold so you can see what's missing.

syclik:
> We need to decide if
> 
> * we can use the tbb like boost such that it is mandatory to build it as a dependency for every stan program
> * we make only optional use of it (likely with STAN_THREADS)
> * we cannot use it...hopefully this is not the case

@wds15, can you explain these questions more clearly? For the first question:

- what does this mean? it needs to be built as a separate binary. Does it need to be linked at runtime or linked at compile time? Are you suggesting we always link it in, even if there are no threads in the program? What's the penalty for doing that (in runtime, compile time, maintenance)?  Is there any chance we get version issues with different versions of TBB?

- second question: optional use. You're not talking about having two implementations of threads, are you?

- what sorts of things do we need to think about so we're not using it?

wds15:
Thanks for the questions in bold on the wiki page that helps me to understand where things are missing.

The above three points where not meant as a question, but rather the set of decisions we can make here. Mandatory use / optional use / no use.

I will try to answer in order:

1. Mandatory use of TBB is my preferred choice here (and I understand that this is a bold thing requiring careful consideration). In this case the TBB must be build once and the easiest thing is to link against it when the final binary is created. Linking in the tbbmalloc_proxy library replaces the new/malloc/... cals with variants from the TBB. Doing so speeds up even the non-threaded programs as you see above. So, by all means, yes,  we should always link in these libraries. Note that the TBB is written for optional parallelism. Thus we can rewrite things with the TBB and it will work in non-threaded programs and in threaded ones.

2. I don't suggest to clutter threading related code, no. If we use the TBB optionally, then the TBB will be used whenever `STAN_THREADS` is on. So for threaded stan programs the TBB must be available and working. 

3. I have looked at a lot like: Compilers+OS supported (all good, we got clang+gcc / Linux+MacOS+Windows). As I understand it we can build it under cmdstan and PyStan while RStan does not need to build it as the TBB is distributed in binary form with the RcppParallel.

Currently, the RcppParallel package ships with TBB 2018 update 1 - which is rather old. However, the library stayed binary backwards compatible, but I would still need to test this at least once. I asked for an update, but have not yet gotten an answer. We could probably just use 2018 update 1, but using ~2y old software feels odd.

Where I would also like to have some "data" is the performance on AMD. While the TBB is at the end of the day just C++ software I would love to see the benchmarks on AMD. However, given that so many people seem to just use it, I guess everything is alright as I have said it myself (after googling a lot on this).

Does this help? Did I miss something?

You have been asking for data on the additional compile time burden. Building the TBB once is relatively quick and the linking is something I haven't really noticed. I eye-balled the run-times of the unit test suite and there the runtimes are in the right order when I checked (in fact huge runtimes of the unit tests did lead me to a makefile bug which is solved now).

syclik:
That helped a lot. Thank you.

Does this with placement new that we defined as part of our arena memory
management?

SteveBronder:
> Where I would also like to have some "data" is the performance on AMD. While the TBB is at the end of the day just C++ software I would love to see the benchmarks on AMD. However, given that so many people seem to just use it, I guess everything is alright as I have said it myself (after googling a lot on this).

I have a threadripper on my home comp. If you send me the stuff to run I can do that this week

wds15:
@syclik I don't quite understand your question...but the tbbmalloc_proxy replaces all calls to new, malloc & all the related commands. So these are rerouted to the versions of the TBB which provide scalable behavior. This scalable property is of particular importance in multi-threaded applications (allocation of the global heap requires synchronization otherwise), but it turns out that single-core stan also sees a speed bump.

@SteveBronder It would make me feel better to see that AMD performance is at least not degraded from this. So if you could use the performance benchmark suite with the stat comp benchmark set with 5 runs to compare, then that would be great. You would need to run with a clean cmdstan develop checkout once and then once with a cmdstan checkout of `feature/intel-tbb-lib` coupled with a stan-math checkout of the same branch. I would love to see the speedups, but I would also be happy if it does not run slower with the TBB linked in. Let me know (on the forum) if you need more detailed instructions on what to do. Thanks!

wds15:
@syclik I added a lot more detail to the checklist page. Please let me know if things are still unclear - maybe by more questions on that wiki page as needed.

SteveBronder:
Sure! I'm pretty sure I can just run this using Sean's stuff

```
# For 10 runs test cmdstan develop against feature/intel-tbb-lib with the compare using the that math commit
./compare-git-hashes.sh "stat_comp_benchmarks -j8 --runs 10" develop feature/intel-tbb-lib false dc604f157054ab187462b2cca67ca8765293c42b
```

SteveBronder:
Got to run some alt benchmarks today and, may not be for our particular style of problem, but the tbb allocator seems to be a bit problematic on my machine

https://github.com/daanx/mimalloc-bench/issues/1#issuecomment-504857370

SteveBronder:
Yeah these are pretty rough. gp regression in particular

```
SUCCESS: Gold golds/stat_comp_benchmarks_benchmarks_arK_arK.gold passed.
+ ./comparePerformance.py develop_performance.csv performance.csv
('stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan', 0.89)
('stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan', 0.95)
('stat_comp_benchmarks/benchmarks/arK/arK.stan', 0.97)
('stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan', 0.91)
('stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan', 0.87)
('stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan', 0.97)
('stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan', 0.83)
('performance.compilation', 1.1)
('stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan', 1.01)
('stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan', 1.0)
('stat_comp_benchmarks/benchmarks/sir/sir.stan', 0.97)
('stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan', 0.92)
('stat_comp_benchmarks/benchmarks/garch/garch.stan', 0.99)
('stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan', 0.93)
('stat_comp_benchmarks/benchmarks/arma/arma.stan', 0.94)
0.950215004526
```

wds15:
So on your AMD machine you see a 5% performance loss whenever you link in the tbbmalloc_proxy?

SteveBronder:
That's what it looks like

SteveBronder:
Can we include the tbb without their malloc?

wds15:
Yes - we can include the TBB without the tbbmalloc, of course.

wds15:
@syclik What do you suggest to make progress?

Points to consider:

- Do we want to make it optional if the Intel tbbmalloc is used? We have one report on a 5% performance degradation on AMD processors so far - the question is if that justifies to make the tbbmalloc optional. Looking at the 5% degradation it seems to me that the slow-down happens for models which are anyway fast. Maybe we need more data on this topic? If more data... what data?
- If the tbbmalloc is only optionally used - how should we configure that? Using a makefile variable? How should the defaults be set for that variable?

I would guess that the TBB malloc is still preferable on AMD whenever threads are being used. The big win about the tbbmalloc is that it comes with the TBB so that we would not need yet another dependency (e.g. this https://github.com/gperftools/gperftools looks nice).

syclik:
@wds15, thank you for the additional information.

I think most of it's answered. We should really clean up the wiki page so it's coherent. Here's what I think is left:

1. Determine whether it's ok with licensing. I don't know what the current status is. If you think the question is clear, can you, @wds15, state it? We need to generate a question to ask an expert.
2. "Can you put up a simple example of how threading is different under C++ threads and TBB?" That question was ignored. It'd be useful to have that in determining whether we want to move with it. Here's what you haven't shown: why is using Intel TBB easier than threads for these features? Especially in the context of maintenance. It's on you, as the proposer, to show that this is indeed easier. I'm willing to believe if you provide an example.
3. Additional dependencies: make. We haven't discussed this at all. Is it hard to install? Is it reasonable? Is this going to cause any problems? Where are there instructions to install?

You have convinced me on time. You have not convinced me on maintenance burden for Math developers. You have not convinced me on the downstream effects of having users install this easily.

To answer your questions: we should not link conditionally unless there's a very good reason. Good reasons:
1. it's slower if we link it in compile time
2. it's slower if we link it for run time
3. we can't use it for some OS / compiler combination

We're almost there. I think it'll be good once you explicitly write how the maintenance burden isn't going to be nightmarish. You've linked to a sell sheet, but you haven't linked to a comparison to the code. 

Also, that side effect is a big one. A really big one. So anyone on Windows will stop working unless they update their tool chain. That's not really a great response. What do you suggest we do? Please propose an option or options for next steps regarding this.

rok-cesnovar:
> 3\. Additional dependencies: make. We haven't discussed this at all. Is it hard to install? Is it reasonable? Is this going to cause any problems? Where are there instructions to install?

Just stating here that on the Stan Math meeting that happened on the 18th June we came to an agreement that since the `make` version needed on Windows is supplied with the recommended Rtools package, this is not an additional dependency and this is thus a non-issue.

SteveBronder:
Related to the meeting we just had, does TBB require compilation and linking or is it header only? mimalloc does require compilation and linking so if we want to keep being header only that may be an issue. Maybe the answer here is to have a flag during compilation to statically override the general allocator and give users options.

Either way, this weekend I can hack up a version of stan-math that uses mimalloc so we can get some benchmarks

rok-cesnovar:
TBB requires dynamic linking, static linking is doable but discouraged.

ahartikainen:
What difference would it make? Considering our use.

rok-cesnovar:
Is that a question regarding my comment on static linking vs dynamic linking? 

For practical purposes probably only that Windows users need to set the PATH variable when linking dynamically. Except if you count larger executables as a factor, I dont.

ahartikainen:
It was a question for dynamic vs static. :)

SteveBronder:
@wds15 do you have any parallel models that can go in the cmdstan performance benchmarks? I have a branch [here](https://github.com/bstatcomp/math/tree/memory/mi-malloc-override) that sets up the stuff to use mimalloc? Should that branch be based on the branch you used for the other parallel models you posted a while ago?

wds15:
@SteveBronder You can use the map_rect example here: https://discourse.mc-stan.org/t/map-rect-threading/8802/3 You can run this example with and without STAN_THREADS. Moreover, it is worthwhile to run it with MPI which gives you the best multi-core performance I have seen so far (because you are using then independent processes compiled without STAN_THREADS).

@ahartikainen Is there a problem with dynamic linking? To explain: The Intel TBB developers intend their library to by used with dynamic linking as they saw many problems with static linking. My understanding is that the idea of the TBB is to manage the entire resources of a machine in order to give best performance. For example, when you have one Stan model of which you run two chains, then the TBB seeks to create *one* threadpool for both chains running concurrently as I understand. So the TBB gives you a system-wide singleton instance of the threadpool if I am guessing correctly here. That only works with dynamic linking, not static linking. I could be wrong with this... but in any case - why should I use static linking if it is strongly discouraged? static linking can be made working, but given the warnings I would not bother with it if I can avoid it.

ahartikainen:
probably there is no problem. Just wanted to know the reasons. That one threadpool thing sounds reasonable reason to use dynamic linking.

SteveBronder:
Apologies I didn't have time to work on the mimalloc stuff this weekend. Though it would be nice to have have mpi/threading benchmarks on the cmdstan performance repo.

No need to block this tho

wds15:
There is no hurry with this... we haven't even settled the license questions... but we are moving there.

mimalloc would be interesting, yes... map_rect tests would make a lot of sense with the benchmarking suite. MPI maybe as well, but would be a hassle due to the needed setup, I guess.

SteveBronder:
> MPI maybe as well, but would be a hassle due to the needed setup, I guess.

I think we could just modify the `comparePerformance.py` to include the MPI/threading flags

wds15:
threading: yes. MPI: maybe... MPI requires more system setup. In both cases you would need to recompile things which we haven't done yet in the performance tests; its just one compile config at this moment.

rok-cesnovar:
@wds15 I finally came around upgrading to 2019_U8 that we agreed on last week.

See the upgrade script [here](https://github.com/stan-dev/math/pull/1180/commits/d2df17b8c0cb7d30302bca820e40615ebcda7fbc)

The upgrade is done in three separate commits:
- removed the old one https://github.com/stan-dev/math/pull/1180/commits/19f74b6a2dbfd934dda3d71c9860745a69cec87b
- replaced the version in README.md https://github.com/stan-dev/math/pull/1180/commits/f2c691452620ff5000583916bfa5e438ec1e7ce7
- added 2019_U8 https://github.com/stan-dev/math/pull/1180/commits/91cd9fc6fe8bf820dd91e9bf8aac0cf4939d6cc0

The only thing that is missing (at least to me) in the upgrade file is how to test if the upgrade is working. So a test that will use TBB explicitly needs to be added later.

rok-cesnovar:
The `stack_alloc_test` seems to be failing on Jenkins after the upgrade. Do I need to change/add something?

I also extended the upgrade script with pruning of examples/ folder. You did the same thing before right @wds15 ?

wds15:
Thanks for the update script. I did prune the TBB sources somewhat, yes.

I can take a look at the `stack_alloc_test`. I plan to get my hands on this early next week so that we are possibly still in time...let's see.

wds15:
@rok-cesnovar could you have a look, please? I think we are good with this as is. I just added that tbbmalloc and tbbmalloc_proxy is only build whenever STAN_THREADS is set. This is the default behaviour and the user can override that if he likes.

I executed the stack-test locally and everything was fine....let's cross fingers this disappears.

Anything else we wanted to do? I don't think so.

I will now branch off from this branch and work on the next steps (integrate TBB with Stan-math and as a second PR rewrite map_rect with parallel_for from the TBB).

Jenkinsfile
```diff
@@ -157,6 +157,7 @@ pipeline {
                         deleteDir()
                         unstash 'MathSetup'
                         sh "echo CXX=${MPICXX} >> make/local"
+                        sh "echo CXX_TYPE=gcc >> make/local"
```
rok-cesnovar: Why do we need to specifically define the `CXX_TYPE` here? Isnt this handled by the make/compiler_flags code?

README.md
```diff
@@ -8,11 +8,12 @@ The Stan Math Library is licensed under the [new BSD license](LICENSE.md).
```
rok-cesnovar: Can you add a quick line under the Licensing section stating that Stan Math includes an Apache-2 licensed dependency and that anyone using Stan Math is fine as long as they dont redistribute unary binaries licensed GPL-2 and "see more in the wiki".

make/compiler_flags
```diff
@@ -7,6 +7,19 @@ ifneq ($(OS),Windows_NT)
...
+  LIBRARY_SUFFIX := .dll
+endif
+
+ifeq ($(OS),Darwin)
+  LIBRARY_SUFFIX := .dylib
+endif
+
+ifeq ($(OS),Linux)
+  LIBRARY_SUFFIX := .so
+endif
```
rok-cesnovar: I would feel more comfortable with something in the lines of:
```
ifeq ($(OS),Windows_NT)
  LIBRARY_SUFFIX := .dll
else ifeq ($(OS),Darwin)
  LIBRARY_SUFFIX := .dylib
else ifeq ($(OS),Linux)
  LIBRARY_SUFFIX := .so
else
  $(error Can't detect OS properly. This will impede building the TBB dependency of Stan Math. Please file an issue.)
endif
```
I know this case will probably/hopefully never happen but in case that it does, it makes support easier.

Jenkinsfile
wds15: No. This is special for mpi. With mpi the compiler is called mpicxx or something like that. The our internal detection fails for that such that you need to tell Stan what is the wrapped compiler type.

README.md
wds15: Good idea.

make/compiler_flags
wds15: I think that our makefiles anyway fall over if we are not in any of these 3 cases. So why should we suspect for this check that this is not the case? 

We could add a general check for the os and fail if it’s not one of the expected...but that would rule out to use this on other platforms from which we just don’t know yet that we support them.

Maybe using ?= is better here, since the user can override that?

The tbb supports android, for example...

Jenkinsfile
rok-cesnovar: Ah, OK. Did not know that.

make/compiler_flags
rok-cesnovar: > I think that our makefiles anyway fall over if we are not in any of these 3 cases. So why should we suspect for this check that this is not the case?
> 
> We could add a general check for the os and fail if it’s not one of the expected...but that would rule out to use this on other platforms from which we just don’t know yet that we support them.

Come to think of it, we should not do it here. Otherwise it will interfere with [this check](https://github.com/stan-dev/cmdstan/pull/737/files) that will be added for Stanc3. This one would trigger before that and that might be confusing.

> Maybe using ?= is better here, since the user can override that?

Yes, lets do that.

rok-cesnovar:
I am getting the same error on Windows as the one on Jenkins. Is there anything I can try to help squash this one?

I have a clean Windows install, cloned math + checked out this branch.

If I do this it starts building but then fails elsewhere:
```
$(TBB)/lib/tbb.dll: $(TBB)/lib/tbb.def
tbb.dll: $(TBB)/lib/tbb.def
$(TBB)/lib/tbbmalloc.dll: $(TBB)/lib/tbbmalloc.def
tbbmalloc.dll: $(TBB)/lib/tbbmalloc.def
$(TBB)/lib/tbbmalloc_proxy.dll: $(TBB)/lib/tbbmalloc.def
tbbmalloc_proxy.dll: $(TBB)/lib/tbbmalloc.def
```

seantalts:
tagging @alashworth - he's currently rewriting the build system with cmake and conan to manage dependencies like these and I think it's easier if we don't customize the source code directly.

rok-cesnovar:
There is no source code customization here, just makefile stuff.

seantalts:
Sweet!

wds15:
Sounds like a plan. I suggest that after this PR we have for Stan-math one more which initializes the TBB threads accordingly and provides an init threadpool function.

However, for this PR two more things came to my mind before we can merge:

- The PR description needs and update to describe what we actually do now (changed slightly)

and this is optional for this one, I think, but must be done with the next PR:

- The README.md needs an update wrt to build instructions. The complication now is that the TBB is mandatory to link against. This means that people need to have the TBB around and in case they do that with out makefiles, then `mingw32-make` is needed on windows and on windows you have to set the PATH variable accordingly. Maybe we put this special TBB setup info onto our wiki?

The PR update is mandatory for this PR; the TBB doc can be part of the next PR with the init stuff - as right now this PR does not make the TBB mandatory to link against (yet) given that we do not use the TBB in the source base at the moment.

So the next PR should be "integrating TBB with math" which means it becomes mandatory and as such we need the extra doc. Any opinion where to put the TBB doc (README vs wiki)?

rok-cesnovar:
> The PR description needs and update to describe what we actually do now (changed slightly)

Do you mean that we only use tbballoc for STAN_THREADS? Please add that info yes.

> as right now this PR does not make the TBB mandatory to link against

That info should be added in the PR that will make it mandatory, not here. I would extend the README since its going to be essential for the user to know it. 
Stuff like MPI and OpenCL are optional and dont need to be added in the README and have separate wikis.

wds15:
I added a check for make on Windows. In case it is not mingw32-make, the user will now get a nice error message instead of garbage from the failed build. This percolates up to cmdstan. How do you like that?

rok-cesnovar:
I think that is a great idea! I am working on adding other helpful warnings to cmdstan so I very much approve this.

wds15:
I am writing the CmdStan manual right now.

wds15:
I think we have slight mis-alignment here. I thought we wanted to have the tbb *sources* in lib/tbb_version and only the build *binary* will land always in lib/tbb. At least his is how I wrote the makefiles, but the upgrade script is now written such that the sources of the tbb end up in lib/tbb. So we need to change either way, but right now it's not consistent which is why tests break at the moment.

rok-cesnovar:
Ah ok, I misunderstood, sorry. In that case, the only thing I need to change is that the upgrade-script doesnt alter the makefiles? Just double checking.

wds15:
The `make/libraries` file needs updating since there I need to know the location of the TBB sources.

Other than that the tbb source need to be in `lib/tbb_2019_U8` (right now they are in `lib/tbb`).

The makefiles are setup to build everything in `lib/tbb` using the sources from `lib/tbb_2019_U8`.

rok-cesnovar:
Well in that case I didnt have to change the update-script at all and I will just revert back to what it was.

wds15:
Ok. If you can revert that would be great. Sorry for the confusion.

I just filed an important update to the makefile logic with the TBB: The makefile check on windows for the mingw32-make was flawed. The issue was that the check was marked ".PHONY" which implied that the TBB library targets (which do depend on the makefile check) were never clean for make, because they did depend on something which is never clean. This way the make for the tbb libraries were always called even when the TBB was already build.

I changed this such that the make check now depends on a file which I simply create by touch after a successful check for mingw32-make on Windows. This new behavior now ensures that you will only need mingw32-make to build the TBB and moreover the makefiles for the TBB are never triggered again once the TBB is build.

So the cool thing now is that users only have to remember the mingw32-make thing when they actually build the TBB. They can continue using "make examples/bernoulli/bernoulli.exe" on windows (after the TBB has been build). It will just work, because the TBB makefiles are not touched any longer.

rok-cesnovar:
> Ok. If you can revert that would be great. Sorry for the confusion.

Done. No problem. I was the one confused :)

> This way the make for the tbb libraries were always called even when the TBB was already build.
> So the cool thing now is that users only have to remember the mingw32-make thing when they actually build the TBB.

Great! I think this simplifies it a lot. I will review the changes later.

So to sum up: 
- cmdstan: doc question needs resolving + me testing your changes on path work
- math: tbbmalloc policy for mac only needs to be added here or in the integrate

Will we need to merge the "integrate TBB" PR to this branch or can you switch the PR base branch?

make/libraries
```diff
@@ -123,22 +123,23 @@ ifeq (,$(TBB_CC))
...
 endif
 
-tbb-make-check:
+$(TBB_BIN)/tbb-make-check:
 	@if [[ $(OS) == Windows_NT ]]; then \
 		if ! [[ $(MAKE) =~ mingw32 ]]; then \
-			echo "Error: Please use mingw32-make on Windows. This is packaged with RTools, for example."; \
+			echo "ERROR: Please use mingw32-make on Windows to build the Intel TBB library."; \
+			echo "This is packaged with RTools, for example."; \
 			exit 1; \
```
ahartikainen: Could this just throw a warning? There are other make versions than mingw32-make which work with CmdStan

wds15: but not with the Intel TBB. At least I don't know any other make which can build the TBB. An error is much preferable, since it will stop the whole thing at that point so that the user actually sees the message.

Have you tried other make's which can build the TBB on Windows?

(BTW... if you make a touch on the tbb-make-check file you can circumvent the check)

wds15:
- cmdstan: doc should be good now
- math: the init branch is branched of this one. So I think I can just merge this lib branch into the other init one and then thats good. We should merge like this:

1. this one into stan-math (PR-1180)
2. the respective branch on stan (PR-2769)
3. the cmdstan PR (PR-710)

What I do not know is how we handle the updates of the submodules. So the stan PR-2769 needs to be bumped to point to the right stan-math commit. Same thing for the cmdstan PR-710 which needs to point to stan PR-2769.

Right now I am running the tests for PR-1180, but that run now uses PR-2769 and PR-710 in its upstream tests - so that should come out "green". I will start respective ones for stan and for cmdstan.

The update to the makefile wrt to `tbbmalloc` should be handled in the init PR.

make/libraries
wds15: ... and BTW... once the TBB is build, you can revert to the standard `make` - even on Windows. The `mingw32-make` is really only mandatory for the Intel TBB build.

rok-cesnovar:
> What I do not know is how we handle the updates of the submodules. So the stan PR-2769 needs to be bumped to point to the right stan-math commit. Same thing for the cmdstan PR-710 which needs to point to stan PR-2769.
> * this one into stan-math (PR-1180)
> * the respective branch on stan (PR-2769)
> * the cmdstan PR (PR-710)

But how will that work if we first merge the math one to develop? Wont that break develop then? Develop tests wont have that info?

Couldnt we avoid setting submodules by first merging the cmdstan PR? That wont break anything as the tbb targets will just be empty. Then merge the Stan one (the changes there are only for tests). 

And once that is in we merge this one and then the map rect one?

wds15:
Hmm... that order could work, yes.

rok-cesnovar:
Ok, lets move the discussion to the cmdstan one and go one by one.

seantalts:
the Daily Stan Performance build failed and it seems related to this PR - https://jenkins.mc-stan.org/blue/organizations/jenkins/Daily%20Stan%20Performance/detail/Daily%20Stan%20Performance/1703/changes/

Can someone take a look? If we can't fix it by the 12th we should probably revert before the feature freeze.

rok-cesnovar:
It seems that its trying to rune the command "Stan". And there are spaces in the path (Daily Stan Performance) which leaves me to believe that its probably some compile/link path not in quotes somewhere.

rok-cesnovar:
This looks like the culprit:
```
tbb_root="/Users/Shared/Jenkins/gelman-group-mac/workspace/Daily Stan Performance/lib/stan_math/lib/tbb_2019_U8" CXX="/usr/local/opt/llvm@6/bin/clang++" CC="/usr/local/opt/llvm@6/bin/clang" LDFLAGS="-Wl,-L,"/Users/Shared/Jenkins/gelman-group-mac/workspace/Daily Stan Performance/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/Shared/Jenkins/gelman-group-mac/workspace/Daily Stan Performance/lib/stan_math/lib/tbb"" /Library/Developer/CommandLineTools/usr/bin/make -C lib/stan_math/lib/tbb -r -f "/Users/Shared/Jenkins/gelman-group-mac/workspace/Daily Stan Performance/lib/stan_math/lib/tbb_2019_U8/build/Makefile.tbbmalloc" compiler=clang cfg=release stdver=c++1y malloc
```

Specifically I am looking at:

`LDFLAGS="-Wl,-L,"/Users/Shared/Jenkins/gelman-group-mac/workspace/Daily Stan Performance/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/Shared/Jenkins/gelman-group-mac/workspace/Daily Stan Performance/lib/stan_math/lib/tbb""`

Those nested quotes look ok? seems iffy.

How can we be sure this fixed it? Get what we believe to be the fix merged and then manually restart the daily performance test?

Hopefully we can replicate this locally otherwise this one is going to be slow to debug.

wds15:
I just filed a PR on cmdstan which initialises the TBB threadpool. I would assume that the performance tests gets triggered there as well. So if we use the build with parameters there, then we should have a chance to debug in finite time - since this allows us to use any math branch and we jump over the math tests and the Stan tests.

I think you are right on and I will locally set this up and see what happens. Spaces should really be disallowed... I had this problem in the past with boost builds.

I am tempted to disallow spaces...

seantalts:
Ah, that's good that that's all it was. 

These "Daily " ones don't get triggered, they just run once a night on develop. It seems like it's just the path thing so if you can fix paths with spaces that should do it. Alternatively you could move to disallow spaces but that seems like a larger change that hinders backwards compatibility somewhat (e.g. if you upgrade your install could stop working) so we might wanna raise it at a meeting or something to be sure.

wds15:
The problem with space is resolved now. The bits for that are part of PR-1376.

There are two Windows specific things which are introduced:

- PATH variable must be set to include the TBB binary libraries. We have made nice make-file targets for that.
- minge32-make must be used on Windows in order to build the TBB

Both of these above things are not yet spread into the last corner of our pipelines. However, the math tests all run just fine with it - on Windows and on any other platform.

So we should by all means be able to sort this out. It's by now - I think - a configuration matter. These configuration requirements are well documented and users having to do this are being helped.

seantalts:
Where else do the PATH variable and mingw32-make need to go? CmdStan?

rok-cesnovar:
We have that in cmdstan/stan and math. We are just having issues with Windows+Jenkins setup.

ahartikainen:
If this is a Windows thing, you really should use the system-level (short) path

See `windows_short_path` in https://github.com/stan-dev/cmdstanpy/blob/master/cmdstanpy/utils.py

seantalts:
I guess my question is - what are the remaining issues? Because I think most of the Jenkins stuff should be purely reliant on CmdStan, so if CmdStan is fixed for e.g. paths with spaces in them, I'm not sure what else would need to change.

wds15:
The space thing is fixed for me. So the branch of PR-1376 has makefiles which can deal with

`/some-path/cmdstan some space/`

with such a weird path you get a working TBB binary. There is more trouble which I believe has to do with PATH stuff.

I don't know about this short path thing yet.

Issue #1184, opened 2019-03-26
Speedup GLMs


t4c1:
## Description
GLMs can be optimized, especially for small number of parameters. Right now there are multiple instances of unnecessary copying of data and repeating same calculations. Also most loops can be replaced with Eigen expressions that produce vectorized (faster) binary code. There was a previous take on this problem ([pull request 1038](https://github.com/stan-dev/math/pull/1038)).

Further speedups using GPUs are likely possible.

## Example
GLMs are faster.

## Expected Output
GLMs are faster.

#### Current Version:
v2.19.0

bob-carpenter:
Thanks for submitting.  We're trying to coax our issues into being more like technical specs.  For this issue, there needs to be a description of how the GLMs are to be sped up.  I think this issue has already been discussed, so maybe there could be a pointer to that discussion.

t4c1:
I have updated the issue with more details.

bob-carpenter:
Thanks much for adding the details.

t4c1:
I have just completed GPU implementations! Speedups are quite significant:

![GPU speedup k=2000](https://user-images.githubusercontent.com/7499452/56028324-2c93e900-5d18-11e9-9b9d-99e6e42e2374.png)
![GPU speedup k=2](https://user-images.githubusercontent.com/7499452/56028339-2f8ed980-5d18-11e9-8ad9-f48d5951bd76.png)

bob-carpenter:
That's really great.  Is it ready for review?  One of the GPU devs is going to have to take a look.

I take it that K = 2000 and K = 2 are the number of predictors?  

I'm surprised there's any speedup for K = 2 --- is it because things other than the matrix-vector multiply are also being sped up?

Any idea why the bernoulli-logit is so much faster on GPU?  I'd have expected normal_id to speed up the most as it's the one where the matrix-vector multiply should take up most of the cycles.  

Thanks for reporting CPU/GPU types.  This stuff's so hard to evaluate with all the hardware options being so important.

t4c1:
This is waiting for GPU caching and CPU GLM optimizations PRs to be merged. Meanwhile I will probably reorganize the code a bit resulting in another PR that will have to be merged before this one. 

K is the number of attributes of each data point.

Yes, there is a separate kernel for each model that does all the computationally heavy operations.

[...]

PR #1185, opened 2019-03-26
Glm optimizations

t4c1:
## Summary

This optimizes GLMs, especially for small number of attributes.

This is based on code from pull request [1038](https://github.com/stan-dev/math/pull/1038), which was further optimized and extended so that it does not drop support for std::vector and RowVector input arguments.

## Tests

One new test for normal_id GLM comes from previous pull request [1038](https://github.com/stan-dev/math/pull/1038). Also a bug in a test for normal_id GLM has been fixed.

## Side Effects

Some new functions are introduced, which do conversions from various possible input types, avoiding unnecessary copying:
as_scalar: Passes trough scalars and throws for other inputs.
as_column_vector_or_scalar: Passes trough scalars and column vectors. Converts row vectors and std::vectors to column vectors.
as_array_or_scalar: Passes trough scalars and and converts matrices and std::vectors to arrays.

## Checklist

- [x] Math issue #1184 

- [x] Copyright holder: 
Tadej Ciglarič, University of Ljubljana,
Matthijs Vákár, Columbia University


t4c1:
I have made some benchmarks ([code](https://gist.github.com/t4c1/6b6e79ef3b77113f1570572d2bb6c16c)). Here are the results:
speedup k=2
speedup k=2000

bob-carpenter:
Thanks!  That looks very compelling, including the test code.  

What compiler and optimization setting did you use to run?

bob-carpenter:
Also, for plotting these improvements, if you want to see what's going on at small sizes and large sizes, it helps to plot the ratio of new performance to old performnce on the log scale.  The old performance can be set to zero, so you just see the improvements.  It's what I did for [the autodiff paper for Stan](https://arxiv.org/pdf/1509.07164.pdf) if you want an example (see figure 4 for a comparison of the way you do it to how I'm suggesting, where Stan is the baseline running at ratio 1 (log ratio 0)). 

log(new speed / old speed) assuming the new ones are always faster.  It won't give you the overall trend, but it'll give you the trend in speedup based on size.

t4c1:
I used -O3 on g++ 4.9.3 on Windows.

t4c1:
@syclik Since you reviewed the previous glm preformance pull request, maybe you could review this one too?

bob-carpenter:
@syclik.  Let me know if you'd like me to review this one.

syclik:
Please do. Looks like I'll have my hands full with the threading stuff.
Thanks!

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
```diff
@@ -81,63 +89,63 @@ typename return_type<T_x, T_alpha, T_beta>::type bernoulli_logit_glm_lpmf(
...
-      theta_derivative[n] = signs[n] * exp_m_ythetan / (exp_m_ythetan + 1);
+  Eigen::Array<T_partials_return, Dynamic, 1> exp_m_ytheta = exp(-ytheta);
+  logp += (ytheta > cutoff)
+              .select(-exp_m_ytheta,
+                      (ytheta < -cutoff).select(ytheta, -log1p(exp_m_ytheta)))
+              .sum();
+  if (!std::isfinite(logp)) {
+    check_finite(function, "Weight vector", beta);
+    check_finite(function, "Intercept", alpha);
+    for (size_t n = 0; n < N; ++n) {
```
bob-carpenter: There should be a check-finite that applies to containers that will return more meaningful messages.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
```diff
@@ -81,63 +89,63 @@ typename return_type<T_x, T_alpha, T_beta>::type bernoulli_logit_glm_lpmf(
...
+                      (ytheta < -cutoff)
+                          .select(as_array_or_scalar(signs),
+                                  as_array_or_scalar(signs) * exp_m_ytheta
+                                      / (exp_m_ytheta + 1)));
+    if (!is_constant_struct<T_beta>::value) {
+      ops_partials.edge3_.partials_ = x_val.transpose() * theta_derivative;
+    }
+    if (!is_constant_struct<T_x>::value) {
+      ops_partials.edge1_.partials_
+          = (beta_val_vec * theta_derivative.transpose()).transpose();
```
bob-carpenter: I think this is `(a * b')'`, which I think just works out to `b * a'`, i.e., `theta_derivative.transpose() * beta_val_vec`.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -63,23 +72,43 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
   typedef
       typename stan::partials_return_type<T_y, T_x, T_alpha, T_beta,
                                           T_precision>::type T_partials_return;
+  typedef typename std::conditional<
+      is_vector<T_precision>::value,
+      typename std::conditional<
```
bob-carpenter: Now that Im seeing more use of these, is there a way to pull out some kind of traits metaprograms that will accomplish this in a more readable way?  These nested versionare hard to read.  

The `stan::` qualifiers aren't necessary as we're in the `stan` namespace, but feel free to leave if you think it's clearer this way.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -63,23 +72,43 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
+      is_vector<T_y>::value || is_vector<T_precision>::value,
+      Eigen::Array<typename stan::partials_return_type<T_y, T_precision>::type,
+                   -1, 1>,
+      typename stan::partials_return_type<T_y, T_precision>::type>::type
+      T_sum_val;
 
   using Eigen::Array;
   using Eigen::Dynamic;
   using Eigen::Matrix;
-  using std::exp;
```
bob-carpenter: Thanks for cleaning stuff like this up.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
   for (size_t n = 0; n < N; ++n)
-    theta_dbl[n] += value_of(alpha_vec[n]);
-  Array<T_partials_return, Dynamic, 1> log_phi = phi_arr.log();
-  Array<T_partials_return, Dynamic, 1> logsumexp_eta_logphi
-      = theta_dbl.binaryExpr(log_phi, [](const T_partials_return& xx,
-                                         const T_partials_return& yy) {
-          return log_sum_exp(xx, yy);
-        });
-  Array<T_partials_return, Dynamic, 1> y_plus_phi = y_arr + phi_arr;
+    check_finite(function, "Matrix of independent variables", theta[n]);
```
bob-carpenter: This check_finite should get pulled up to be a test on `theta`.  Or is the problem that it's a matrix and `check_finite` isn't defined for matrices?  If not, would you mind creating an issue to extend these test functions to matrices?

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
+  T_precision_val log_phi = log(phi_arr);
+  Array<T_partials_return, Dynamic, 1> logsumexp_theta_logphi
+      = (theta > log_phi)
```
bob-carpenter: [optional]
This really shold be a vectorized version of `log_sum_exp`.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
+            .select(theta + Eigen::log1p(Eigen::exp(log_phi - theta)),
```
bob-carpenter: Will these `Eigen::log1p` and `Eigen::exp` functions delegate back to our base versions for `var` and `fvar`?  If not, they won't work with nested autodiff.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
-    logp -= (y_arr + Array<double, Dynamic, 1>::Ones(N, 1))
-                .unaryExpr(
-                    [](const T_partials_return& xx) { return lgamma(xx); })
-                .sum();
+    logp -= sum(lgamma(y_arr + 1));
   }
   if (include_summand<propto, T_precision>::value) {
-    for (size_t n = 0; n < N; ++n)
-      logp += multiply_log(phi_arr[n], phi_arr[n]) - lgamma(phi_arr[n]);
+    if (is_vector<T_precision>::value) {
```
bob-carpenter: Thanks---much easier to read.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
+              * (multiply_log(as_scalar(phi_val), as_scalar(phi_val))
+                 - lgamma(as_scalar(phi_val)));
+    }
   }
   if (include_summand<propto, T_x, T_alpha, T_beta, T_precision>::value)
-    logp -= (y_plus_phi * logsumexp_eta_logphi).sum();
+    logp -= (y_plus_phi * logsumexp_theta_logphi).sum();
   if (include_summand<propto, T_x, T_alpha, T_beta>::value)
-    logp += (y_arr * theta_dbl).sum();
+    logp += (y_arr * theta).sum();
```
bob-carpenter: I hesitate to comment on a lot of this, but we'd generally prefer to use `stan::math::sum` as it's more efficient for autdiff types (it consolidates expression graphs).  This is only going to matter for higher-order autodiff, which we aren't even using now.  But if we leave these in as `sum()`, we'll probably have to change them later for efficiency.  Anyway, no big deal leaving them in.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
-    }
-    if (!is_constant_struct<T_alpha>::value) {
-      if (is_vector<T_alpha>::value)
-        ops_partials.edge2_.partials_ = theta_derivative;
-      else
-        ops_partials.edge2_.partials_[0] = theta_derivative.sum();
+        && is_constant_struct<T_alpha>::value
+        && is_constant_struct<T_precision>::value)) {
+    Array<T_partials_return, Dynamic, 1> theta_exp = theta.exp();
+    if (!(is_constant_struct<T_x>::value && is_constant_struct<T_beta>::value
```
bob-carpenter: [optional]
It'd be nice if this was just

```
if (!is_constant_struct<T_x, T_beta, T_alpha>::value)
```

because it's hard to read the values we're testing out of all the boilerplate.  But I don't think `is_constant_struct` extends this way.  It'd be easy with parameter packs, but a distraction for this PR.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -95,107 +124,94 @@ neg_binomial_2_log_glm_lpmf(const T_y& y, const T_x& x, const T_alpha& alpha,
...
+          && is_constant_struct<T_alpha>::value)) {
+      Matrix<T_partials_return, Dynamic, 1> theta_derivative
+          = y_arr - theta_exp * y_plus_phi / (theta_exp + phi_arr);
+      if (!is_constant_struct<T_beta>::value) {
+        ops_partials.edge3_.partials_ = x_val.transpose() * theta_derivative;
+      }
+      if (!is_constant_struct<T_x>::value) {
+        ops_partials.edge1_.partials_
+            = (beta_val_vec * theta_derivative.transpose()).transpose();
```
bob-carpenter: same comment as before --- only one transpose should be necessary

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
```diff
@@ -86,56 +92,45 @@ normal_id_glm_lpdf(const T_y &y, const T_x &x, const T_alpha &alpha,
...
-  for (size_t n = 0; n < N; ++n) {
-    if (include_summand<propto, T_scale>::value)
-      logp -= log(value_of(sigma_vec[n]));
-    inv_sigma[n] = 1 / value_of(sigma_vec[n]);
-    y_minus_mu_over_sigma[n]
-        = (value_of(y_vec[n]) - mu_minus_alpha_dbl[n] - value_of(alpha_vec[n]))
-          * inv_sigma[n];
-    y_minus_mu_over_sigma_squared[n] = square(y_minus_mu_over_sigma[n]);
-  }
+  T_scale_val inv_sigma = 1 / as_array_or_scalar(sigma_val_vec);
```
bob-carpenter: [optional]
This is another palce where our `inv(v)` will be more efficient than `1 / v` for autodiff types.

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
```diff
@@ -145,21 +140,48 @@ normal_id_glm_lpdf(const T_y &y, const T_x &x, const T_alpha &alpha,
...
+            = (y_minus_mu_over_sigma_squared_sum - N) * as_scalar(inv_sigma);
       }
     }
+  } else {
+    y_minus_mu_over_sigma_squared_sum
+        = (y_minus_mu_over_sigma * y_minus_mu_over_sigma).sum();
+  }
+
+  if (!std::isfinite(
+          y_minus_mu_over_sigma_squared_sum)) {  // only do potentially
```
bob-carpenter: I like the approach.  The code's so clear that the comment isn't necessary.

stan/math/prim/mat/prob/poisson_log_glm_lpmf.hpp
```diff
@@ -77,62 +85,59 @@ typename return_type<T_x, T_alpha, T_beta>::type poisson_log_glm_lpmf(
...
-  scalar_seq_view<T_alpha> alpha_vec(alpha);
-  scalar_seq_view<T_y> y_seq_view(y);
-  for (size_t n = 0; n < N; ++n) {
-    // Compute the derivative wrt theta.
-    double theta_dbl_n = theta_dbl[n] + value_of(alpha_vec[n]);
-    double exp_theta_n = exp(theta_dbl_n);
-    theta_derivative[n] = y_seq_view[n] - exp_theta_n;
-    if (!is_vector<T_alpha>::value)
-      theta_derivative_sum += theta_derivative[n];
+  Matrix<T_partials_return, Dynamic, 1> theta = x_val * beta_val_vec;
```
bob-carpenter: This is another case where we'll wind up delegating to Eigen's `operator*()` rather than using `stan::math::multiply` for nested autodiff types.  But given that we don't have any yet, feel free to just ignore this issue for now.  This is all fine if we're only worrying about efficiency for reverse mode and we really haven't been worrying about forward mode efficiency anywhere at all.

stan/math/prim/mat/prob/poisson_log_glm_lpmf.hpp
```diff
@@ -77,62 +85,59 @@ typename return_type<T_x, T_alpha, T_beta>::type poisson_log_glm_lpmf(
...
+      = as_array_or_scalar(y_val_vec) - exp(theta.array());
+  double theta_derivative_sum = theta_derivative.sum();
+  if (!std::isfinite(theta_derivative_sum)) {
+    check_finite(function, "Weight vector", beta);
+    check_finite(function, "Intercept", alpha);
+    check_finite(function, "Matrix of independent variables", theta);
+  }
+  if (include_summand<propto>::value) {
+    if (is_vector<T_y>::value) {
+      logp -= sum(lgamma(as_array_or_scalar(y_val_vec) + 1.0));
```
bob-carpenter: [optional]
Don't need the `.0` after `1`---the compiler will sort out the literals.

stan/math/prim/scal/fun/as_array_or_scalar.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef STAN_MATH_PRIM_SCAL_FUN_AS_ARRAY_OR_SCALAR_HPP
+#define STAN_MATH_PRIM_SCAL_FUN_AS_ARRAY_OR_SCALAR_HPP
+
+namespace stan {
+namespace math {
+
+/**
+ * Converts a matrix type to an array. For scalar inputs this is identity
```
bob-carpenter: This should just document this specialization.  So it can just be `Returns specified input value.`

stan/math/prim/scal/fun/as_array_or_scalar.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * function.
+ *
+ * @tparam T Type of element.
+ * @param v Specified value.
+ * @return Same value.
+ */
+template <typename T>
+inline T as_array_or_scalar(const T& v) {
```
bob-carpenter: Return type should be `const T&` because the scalar type can be `var` or `fvar<var>` in nested autodiff, and we want to pass those by reference.

stan/math/prim/scal/fun/as_column_vector_or_scalar.hpp
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Converts input argument to a column vector or a scalar. For scalar inputs
+ * that is an identity function.
+ *
+ * @tparam T Type of scalar element.
+ * @param a Specified scalar.
+ * @return 1x1 matrix that contains the value of scalar.
+ */
+template <typename T>
+inline T as_column_vector_or_scalar(const T a) {
```
bob-carpenter: This should have argument type `const T&` and same return type `const T&` to deal with scalars that aren't primitives.

stan/math/prim/scal/fun/as_scalar.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef STAN_MATH_PRIM_SCAL_FUN_AS_SCALAR_HPP
+#define STAN_MATH_PRIM_SCAL_FUN_AS_SCALAR_HPP
+
+namespace stan {
+namespace math {
+
+/**
+ * Converts input to a scalar. For scalar arguments this is an identity
```
bob-carpenter: This isn't quite true.  For scalar argument that are integers, they get converted to double values.  If that's not the intent, there needs to be another specialization or this just needs to be replaced with

```
<template typename T> 
inline const T& as_scalar(const T& x) { return x; }
```

stan/math/prim/scal/fun/sum.hpp
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Calculates sum of values given as input. For scalar inputs this is identity
+ * function.
+ *
+ * @tparam T Type of element.
+ * @param v Specified value.
+ * @return Same value (the sum of one value).
+ */
+template <typename T>
+inline T sum(const T& v) {
```
bob-carpenter: Return type should be `const T&` for same reasons as before.

Doc should also doc this version of the function.  It's just an overload, not a specialization.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
```diff
@@ -81,63 +89,63 @@ typename return_type<T_x, T_alpha, T_beta>::type bernoulli_logit_glm_lpmf(
...
-  {
-    scalar_seq_view<T_beta> beta_vec(beta);
-    for (size_t m = 0; m < M; ++m) {
-      beta_dbl[m] = value_of(beta_vec[m]);
-    }
-  }
-  Eigen::Array<T_partials_return, Dynamic, 1> ytheta
-      = signs.array() * (value_of(x) * beta_dbl).array();
-  scalar_seq_view<T_alpha> alpha_vec(alpha);
+  const auto &x_val = value_of_rec(x);
```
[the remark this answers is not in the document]
t4c1: Actually it is picked up by autoformatter - I originally wrote it the way you suggest.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
t4c1: Can you suggest what the error message should be? I just left them as they were.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
t4c1: This is intentional as beta_val_vec could be a scalar that does not define .transpose(). If you prefer more readability here at expense of some code elsewhere I can make a transpose() function that will be a no-op for scalars.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
t4c1: check_finite is defined for matrices, but not for Eigen::Array that we have here. I could write `check_finite(function, "Matrix of independent variables", theta.matrix());`, but that would create a copy which is bad performance-wise. I can write an issue for that or just implement it here if you prefer.

stan/math/prim/mat/fun/as_array_or_scalar.hpp
```diff
@@ -0,0 +1,57 @@
...
+as_array_or_scalar(const Eigen::Map<const Eigen::Matrix<T, R, C>>& v) {
+  return v.array();
+}
+
+/**
+ * Converts a std::vector type to an array.
+ *
+ * @tparam T Type of scalar element.
+ * @param v Specified vector.
+ * @return Matrix converted to an array.
```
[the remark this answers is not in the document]
t4c1: While I am using this functions only on vectors, it is written in general way so that it can work on any matrix, not just vectors.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
t4c1: I think that requires a separate issue and pull.

stan/math/prim/mat/fun/as_array_or_scalar.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * Converts a matrix type to an array.
+ *
+ * @tparam T Type of scalar element.
+ * @tparam R Row type of input matrix.
+ * @tparam C Column type of input matrix.
+ * @param v Specified matrix.
+ * @return Matrix converted to an array.
+ */
+template <typename T, int R, int C>
+inline Eigen::ArrayWrapper<const Eigen::Matrix<T, R, C>> as_array_or_scalar(
```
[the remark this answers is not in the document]
bob-carpenter: `scalar_seq_view` also wraps `std::vector<T>` and `Eigen::Matrix<T, ...>`.  Perhaps we should've had that view be an `Eigen::Array`, but right now, it's our own custom view.

The directory location's not super critical in that it's easy to change. Although not strictly metaprograms, all of our template expressions have also been dumpted into `meta`.  @syclik might want to weigh in with an opinion.

stan/math/prim/mat/fun/as_array_or_scalar.hpp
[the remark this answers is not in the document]
bob-carpenter: Thanks for the clarification.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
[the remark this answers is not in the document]
bob-carpenter: It must be inconsistent as some of them are `const auto&` and some are `const auto &`.  Anyway, no big deal one way or the other.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
bob-carpenter: Sorry that wasn't clearer.  The message doesn't need to change, it just shouldn't be in a loop.  We shold have `check_` methods that work on containers so that they can indicate the index(es) at which the failure occurs.  The problem with having a loop here like this is that we don't indicate which element failed.

This also isn't a big deal if the container tests don't (yet) exist, but we should be creating issues for them so we can get rid of these loops.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
bob-carpenter: When the autodiff type is `fvar<var>`, then the scalar types will be `var`.  We don't actually use that anywhere in our interfaces, but we have experimental uses and hope to expose it soon.  But we can back-fix this stuff easily if it doesn't work.

bob-carpenter:
Can the Array be converted to a Map<Matrix<T,...>>?  Then we'd get the indexing.  

Anyway, this also isn't that big of a deal.

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
t4c1: I don't see a general way how to do it, as we have different types (Array/Matrix, sometimes even std::vector), and one version even depends on two different inputs. Except for nested one that is not actually needed, I will just leave that as it is for now.

t4c1:
Yes.

t4c1:
@bob-carpenter I think I have answerd your questions and fixed what needed fixing. This is ready for next review.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
bob-carpenter: I see.  It's OK as it is.  This is the kind of thing that deserves a comment, though, as it looks wrong at first glance without knowing that `beta_val_vec` can be a scalar.

bob-carpenter:
Thanks.  I took another look and feel like I have to double check with Rok on the metaprograms (he's refactoring them) and with Sean on efficiency.  I think the efficiency for forward mode is OK as is, and probably the metaprograms, too, but I just want to check there's not an existing way that's about to surface to do them.  Then we can merge.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
```diff
@@ -55,6 +62,10 @@ typename return_type<T_x, T_alpha, T_beta>::type bernoulli_logit_glm_lpmf(
   static const char *function = "bernoulli_logit_glm_lpmf";
   typedef typename stan::partials_return_type<T_y, T_x, T_alpha, T_beta>::type
       T_partials_return;
+  typedef typename std::conditional<
```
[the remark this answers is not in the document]
rok-cesnovar: Added this task to #1122

stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
rok-cesnovar: Also added to #1122

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
```diff
@@ -54,6 +62,10 @@ normal_id_glm_lpdf(const T_y &y, const T_x &x, const T_alpha &alpha,
   static const char *function = "normal_id_glm_lpdf";
   typedef typename stan::partials_return_type<T_y, T_x, T_alpha, T_beta,
                                               T_scale>::type T_partials_return;
+  typedef typename std::conditional<
+      is_vector<T_scale>::value,
+      Eigen::Array<typename partials_return_type<T_scale>::type, -1, 1>,
+      typename partials_return_type<T_scale>::type>::type T_scale_val;
```
rok-cesnovar: Will be cleaned up with the `matrix_return_type` PR.

rok-cesnovar:
I wrote down the 2 tasks that fall more under the metaprograms cleanup to #1122. They are both out of scope out this PR. Will get to them once #1186 is merged.

I would advocate on moving the `as_array_or_scalar` functions to the corresponding `meta` subfolders. We do not have clear guidelines (or maybe I missed them) on what goes in meta and what not. The meta subfolder is not just type trait stuff and I feel that these functions are closer to the views in `meta` than the functions in `fun`.

bob-carpenter:
Thanks, @rok-cesnovar!

So the bottom line is that the only thing that needs to be done now is move the views (we tend to drop this kind of stuff in the `meta` folders even if they're not strictly template metaprograms).

PR #1186, opened 2019-03-29
feature/issue 1122 Add missing metaprogram tests

rok-cesnovar:
## Summary

This PR adds tests for the following meta programs:

```
fwd/scal
	is_fvar
prim/arr
	index_type
prim/mat
	append_return_type
	length_mvt
prim/scal
	is_fvar
	is_var
	length_mvt
	max_size_mvt
	max_size
	partials_type
	size_of
rev/scal
	is_var
```

It also removes the empty `prim/scal/meta/seq_view.hpp` file.

This only leaves [scalar_type_pre](https://github.com/stan-dev/math/blob/develop/stan/math/prim/scal/meta/scalar_type_pre.hpp) without any tests among metaprograms. If anyone has an idea how to test that function that would close out the tests part of #1122.

## Tests

Tests for all above mentioned metaprograms.

## Side Effects

/

## Checklist

- [x] Math issue #1122 

### DIFF
--- a/test/unit/math/prim/scal/meta/is_fvar_test.cpp
+++ b/test/unit/math/prim/scal/meta/is_fvar_test.cpp
@@ -1,0 +1,8 @@
+#include <stan/math/prim/scal.hpp>
+#include <gtest/gtest.h>
+
+TEST(MetaTraits, is_fvar) {
+  using stan::is_fvar;
+  EXPECT_FALSE(is_fvar<int>::value);
+  EXPECT_FALSE(is_fvar<double>::value);
+}
--- a/test/unit/math/fwd/scal/meta/is_fvar_test.cpp
+++ b/test/unit/math/fwd/scal/meta/is_fvar_test.cpp
@@ -1,0 +1,11 @@
+#include <stan/math/fwd/scal.hpp>
+#include <gtest/gtest.h>
+
+TEST(MetaTraits, is_fvar) {
+  using stan::is_fvar;
+  using stan::math::fvar;
+  EXPECT_TRUE(is_fvar<fvar<int>>::value);
+  EXPECT_TRUE(is_fvar<fvar<double>>::value);
+  EXPECT_TRUE(is_fvar<fvar<fvar<double>>>::value);
+  EXPECT_TRUE(is_fvar<fvar<fvar<fvar<double>>>>::value);
+}
--- a/test/unit/math/prim/arr/meta/index_type_test.cpp
+++ b/test/unit/math/prim/arr/meta/index_type_test.cpp
@@ -1,4 +1,4 @@
-#include <stan/math/prim/scal.hpp>
+#include <stan/math/prim/arr.hpp>
 #include <stan/math/prim/arr/meta/index_type.hpp>
 #include <test/unit/math/prim/scal/fun/promote_type_test_util.hpp>
 #include <gtest/gtest.h>
--- a/test/unit/math/prim/arr/meta/size_of_test.cpp
+++ b/test/unit/math/prim/arr/meta/size_of_test.cpp
@@ -1,0 +1,17 @@
+#include <stan/math/prim/arr.hpp>
+#include <gtest/gtest.h>
+#include <vector>
+#include <Eigen/Dense>
+
+TEST(MetaTraits, size_of) {
+  using stan::size_of;
+
+  std::vector<double> x2(3);
+  EXPECT_EQ(3U, size_of(x2));
+
+  std::vector<std::vector<double>> x3(6);
+  EXPECT_EQ(6U, size_of(x3));
+
+  std::vector<Eigen::MatrixXd> x4(8);
+  EXPECT_EQ(8U, size_of(x4));
+}
--- a/test/unit/math/prim/arr/meta/size_of_test.cpp
+++ b/test/unit/math/prim/arr/meta/size_of_test.cpp
@@ -1,7 +1,7 @@
 #include <stan/math/prim/arr.hpp>
 #include <gtest/gtest.h>
-#include <vector>
 #include <Eigen/Dense>
+#include <vector>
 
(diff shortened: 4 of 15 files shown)


test/unit/math/prim/scal/meta/is_fvar_test.cpp
```diff
@@ -0,0 +1,8 @@
+#include <stan/math/prim/scal.hpp>
+#include <gtest/gtest.h>
+
+TEST(MetaTraits, is_fvar) {
+  using stan::is_fvar;
+  EXPECT_FALSE(is_fvar<int>::value);
+  EXPECT_FALSE(is_fvar<double>::value);
```
seantalts: Would you mind making sure this doesn't return true for `var` with a test?

syclik: The test should be in the rev directory. The reason is that the includes are different and there might be a specialization included in rev that’s not included in prim or fwd.

syclik: The test should be in the rev directory. The reason is that the includes are different and there might be a specialization included in rev that’s not included in prim or fwd.

seantalts: Agreed that if there is a test for `is_fvar<var>` that that should go in `rev`. Didn't mean to say we should add it to this file.

rok-cesnovar: Added: https://github.com/stan-dev/math/blob/2e9b0377725f1b854332d04d6e1da8ddd3927831/test/unit/math/rev/scal/meta/is_fvar_test.cpp

rok-cesnovar:
I am not sure if I understand correctly. Do you mean joining files like these together? 
https://github.com/stan-dev/math/blob/2e9b0377725f1b854332d04d6e1da8ddd3927831/test/unit/math/prim/arr/meta/size_of_test.cpp and
https://github.com/stan-dev/math/blob/2e9b0377725f1b854332d04d6e1da8ddd3927831/test/unit/math/prim/scal/meta/size_of_test.cpp

seantalts:
Basically. We essentially only use the Math library at the rev/mat.hpp level, so it's not the worst thing if most of the tests exist in rev/mat in my head, especially if you have like 4 lines of test per file and you can join all of the tests for a single function in a single place. Should really help with test compile times too, not just readability. But I'm not sure how @syclik feels about this idea so you can ignore me.

syclik:
It’s technically not the right thing to do.

From prim, we only include a prim header, so we don’t even know what var
is.

From rev, we know what double and var are, but not fvar.

From fvar, we don’t know what var is.

From mix, we know about both.

seantalts:
This is true, but if our supported API is just rev/mat.hpp, then it seems okay to me to only test that set of include orders (well, ignoring forward and mixed mode tests in this discussion), even if more exist. We'll be flattening the (scal/arr/mat) layer and collapsing that distinction anyway soon hopefully. the other distinction might stick around but many tests can be at the API level. Like I don't know what the point of making sure that prim/scal.hpp works without any rev or mat stuff if we don't support that... But I don't feel strongly about this, just trying to talk through my reasoning.

syclik:
Let’s talk through this on discourse, not here. (It’s sorta tangential to
the actual issue at hand and it might involve a larger decision.)

For this PR, the test you’re asking for should belong in rev. And it is
something we should test.

rok-cesnovar:
But the `rev/scal/meta/is_fvar_test.cpp` could also do the "prim" tests right?
This would be fine? or not?
```
#include <stan/math/rev/scal.hpp>
#include <gtest/gtest.h>

TEST(MetaTraits, is_fvar) {
  using stan::is_fvar;
  EXPECT_FALSE(is_fvar<int>::value);
  EXPECT_FALSE(is_fvar<double>::value);
  EXPECT_FALSE(is_fvar<stan::math::var>::value);
}
```
If this cuts down test compile time, this might be worthwhile, but speaking from my own experience of my Stan Math beginnings I was looking at tests a lot and if a function at `stan/math/x/y/z` had no test at `test/unit/math/x/y/z` I felt a bit lost. When talking to Davor he is basically saying the exact same thing. 

> For this PR, the test you’re asking for should belong in rev. And it is something we should test.

@syclik ths test was added.

seantalts:
Yeah, I guess it's probably a larger discussion to move all prim/rev scal/arr/mat tests to rev/mat.

syclik:
Thanks for adding it!

Yes, it could do the test in rev, but it actually is testing against a
different set of headers! Here, it’s simple, but in more complicated cases,
it will do something different depending on which include you bring in. I
can put together an example later on in the weekend.

This is all by design. It might be overly complicated if we only support
rev, but at the moment, we do have prim, fwd, and mix and in order to have
that compile in the right way we have implemented it this way.

seantalts:
> Yes, it could do the test in rev, but it actually is testing against a different set of headers! 

I get that, but we decided recently that we don't support using other sets of headers anymore: https://github.com/stan-dev/math/wiki/Supported-external-API so my question is what is the utility of testing other sets of headers besides the ones we use and support?

syclik:
I think it’s a stretch to say that was decided. It’s something we’ve been
discussing, but I can’t pinpoint a “who” or “when.” I think there are
enough people relying on fwd and mix for us to think seriously before
making that decision and make it a transparent decision.

rok-cesnovar:
Is there anything else that needs to be changed in this PR otherwise please re-approve.

seantalts:
Sorry, I was not saying we should delete forward mode with this PR! I was just saying that my reading of the externally supported API means we can chunk prim/rev + scal/arr/mat as a single API under test, but if you don't want to do that that's fine.

Moving the rest of the discussion to https://discourse.mc-stan.org/t/what-is-the-supported-external-api-of-the-math-library/7438/8

Issue #1191, opened 2019-04-01
Speedup GPU matrix-vector and vector-matrix multiplication


t4c1:
## Description
Current implementation of GPU matrix multiply is very slow (slower than CPU on my computer!) for cases vector\*matrix and matrix\*vector as it creates many threads and does work with only a few of them.

## Expected Output
matrix\*vector and vector\*matrix is faster on GPU.

#### Current Version:
v2.19.0

PR #1192, opened 2019-04-01
optimized matrix-vector and vector-matrix multiply

t4c1:
## Summary
Adds two new opencl kernels for operations matrix\*vector and vector\*matrix. They are used within `multiply(matrix_cl, matrix_cl)` when appropriate. Both are significantly faster than general kernel for matrix multiplication.

## Tests
New tests are added to opencl/multiply_test.cpp for cases matrix\*vector and vector\*matrix.

## Side Effects

None

## Checklist

- [ ] Math issue #1191

- [ ] Copyright holder: Tadej Ciglarič, University of Ljubljana

### DIFF
--- a/stan/math/opencl/kernel_cl.hpp
+++ b/stan/math/opencl/kernel_cl.hpp
@@ -82,7 +82,7 @@ class kernel_functor {
                  std::map<const char*, int> options) {
     auto base_opts = opencl_context.base_opts();
     for (auto& it : options) {
-      if (base_opts[it.first] != it.second) {
+      if (base_opts[it.first] > it.second) {
         base_opts[it.first] = it.second;
       }
     }
--- a/stan/math/opencl/kernel_cl.hpp
+++ b/stan/math/opencl/kernel_cl.hpp
@@ -82,7 +82,7 @@ class kernel_functor {
                  std::map<const char*, int> options) {
     auto base_opts = opencl_context.base_opts();
     for (auto& it : options) {
-      if (base_opts[it.first] > it.second) {
+      if (base_opts[it.first] != it.second) {
         base_opts[it.first] = it.second;
       }
     }
--- a/stan/math/opencl/kernel_cl.hpp
+++ b/stan/math/opencl/kernel_cl.hpp
@@ -82,7 +82,7 @@ class kernel_functor {
                  std::map<const char*, int> options) {
     auto base_opts = opencl_context.base_opts();
     for (auto& it : options) {
-      if (base_opts[it.first] != it.second) {
+      if (!base_opts[it.first] || base_opts[it.first] > it.second) {
         base_opts[it.first] = it.second;
       }
     }
--- a/stan/math/opencl/kernel_cl.hpp
+++ b/stan/math/opencl/kernel_cl.hpp
@@ -82,7 +82,7 @@ class kernel_functor {
                  std::map<const char*, int> options) {
     auto base_opts = opencl_context.base_opts();
     for (auto& it : options) {
-      if (!base_opts[it.first] || base_opts[it.first] > it.second) {
+      if (base_opts[it.first] != it.second) {
         base_opts[it.first] = it.second;
       }
     }
--- a/stan/math/opencl/kernels/matrix_multiply.hpp
+++ b/stan/math/opencl/kernels/matrix_multiply.hpp
@@ -114,6 +114,84 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
     matrix_multiply("matrix_multiply", matrix_multiply_kernel_code,
                     {{"THREAD_BLOCK_SIZE", 32}, {"WORK_PER_THREAD", 8}});
 
+// \cond
+const char* matrix_vector_multiply_kernel_code = STRINGIFY(
+// \endcond
+    /**
+     * Matrix-vector multiplication R=A*B on the OpenCL device
+     *
+     * @param[in] A matrix in matrix-vector multiplication
+     * @param[in] B the vector in matri-vector multiplication
+     * @param[out] R the output vector
+     * @param[in] M Number of rows for matrix A
+     * @param[in] N Number of cols for matrix A and number of rows for vector B
+     */
+    __kernel void matrix_vector_multiply(const __global double* A, const __global double* B, __global double* R, const int M, const int N) {
+      const int gid = get_global_id(0);
+
+      double acc = 0;
+      for (int i = 0, j=0; i < N; i ++, j += M) {
+        acc += A[j+gid] * B[i];
+      }
+      R[gid]=acc;
+    }
+// \cond
+);
+// \endcond
+
+/**
+ * See the docs for \link kernels/matrix_multiply.hpp matrix_vector_multiply() \endlink
+ */
+const global_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int>
+        matrix_vector_multiply("matrix_vector_multiply", matrix_vector_multiply_kernel_code);
+
+// \cond
+const char* vector_matrix_multiply_kernel_code = STRINGIFY(
+// \endcond
+    /**
+     * Vector-matrix multiplication R=A*B on the OpenCL device
+     *
+     * @param[in] M matrix in vector-matrix multiplication
+     * @param[in] V the vector in vector-matrix multiplication
+     * @param[out] R the output vector
+     * @param[in] N Number of cols for vector A and number of rows for matrix B
+     * @param[in] K Number of cols for matrix B
+     */
+        __kernel void vector_matrix_multiply(const __global double* A, const __global double* B, __global double* R, const int N, const int K) {
+          const int lid = get_local_id(0);
+          const int gid = get_global_id(0);
+          const int wgid = get_group_id(0);
+
+          double acc = 0;
+          for (int i = lid; i < N; i += 64) {
+            acc += A[i] * B[i + wgid * N];
+          }
+          __local double res_loc[64];
+          res_loc[lid] = acc;
+          barrier(CLK_LOCAL_MEM_FENCE);
+          if (lid < 16) {
+            res_loc[lid] += res_loc[lid + 16] + res_loc[lid + 32] + res_loc[lid + 48];
+          }
+          barrier(CLK_LOCAL_MEM_FENCE);
+          if (lid < 4) {
+            res_loc[lid] += res_loc[lid + 4] + res_loc[lid + 8] + res_loc[lid + 12];
+          }
+          barrier(CLK_LOCAL_MEM_FENCE);
+          if (lid == 0) {
+            acc = res_loc[lid] + res_loc[lid + 1] + res_loc[lid + 2] + res_loc[lid + 3];
+            R[wgid] = acc;
+          }
+        }
+// \cond
+);
+// \endcond
+
+/**
+ * See the docs for \link kernels/matrix_multiply.hpp vector_matrix_multiply() \endlink
+ */
+const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int>
+        vector_matrix_multiply("vector_matrix_multiply", vector_matrix_multiply_kernel_code);
+
 }  // namespace opencl_kernels
 }  // namespace math
 }  // namespace stan
--- a/stan/math/opencl/kernels/matrix_multiply.hpp
+++ b/stan/math/opencl/kernels/matrix_multiply.hpp
@@ -115,13 +115,13 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
                     {{"THREAD_BLOCK_SIZE", 32}, {"WORK_PER_THREAD", 8}});
 
 // \cond
-const char* matrix_vector_multiply_kernel_code = STRINGIFY(
+static const char* matrix_vector_multiply_kernel_code = STRINGIFY(
     // \endcond
     /**
      * Matrix-vector multiplication R=A*B on the OpenCL device
      *
      * @param[in] A matrix in matrix-vector multiplication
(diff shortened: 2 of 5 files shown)


t4c1:
Here are the speedup measurements.

matrix(N*K) * vector(K):

| N / K | 30 | 100 | 300 | 1000 | 3000 | 10000 | 
| --- | --- | --- | --- | --- | --- | --- | 
| 30 | 3.40805 | 4.95758 | 4.57356 | 4.80392 | 7.04756 | 4.77418 | 
| 100 | 3.87582 | 4.97181 | 3.13868 | 8.43912 | 6.18135 | 4.79222 | 
| 300 | 4.18469 | 3.82494 | 8.36494 | 6.92778 | 5.37196 | 6.54332 | 
| 1000 | 4.3522 | 10.5011 | 11.1242 | 12.1841 | 10.9288 | 13.8965 | 
| 3000 | 10.5296 | 13.3024 | 10.9414 | 12.8083 | 20.1686 | 22.2748 | 
| 10000 | 17.598 | 18.2582 | 16.8314 | 20.9551 | 22.5631 | 22.4784 | 

row_vector(K) * matrix(K*N)

N / K | 30 | 100 | 300 | 1000 | 3000 | 10000 | 
| --- | --- | --- | --- | --- | --- | --- | 
30 | 4.32768 | 4.3975 | 3.87805 | 5.66853 | 9.66949 | 28.5197 | 
100 | 3.33855 | 5.73728 | 7.16763 | 12.5879 | 12.1413 | 20.5031 | 
300 | 4.91376 | 3.97669 | 8.37135 | 13.0816 | 13.7156 | 18.8603 | 
1000 | 4.81629 | 10.7405 | 7.9881 | 11.4812 | 15.1866 | 29.6695 | 
3000 | 10.3128 | 11.8582 | 13.551 | 14.218 | 21.4626 | 24.1479 | 
10000 | 16.4137 | 14.4072 | 18.532 | 22.02 | 22.2829 | 23.0122 |

bob-carpenter:
I'd have thought row-vector * matrix would be faster as the matrices are stored in column-major order by default.  But it looks like the opposite.

t4c1:
On a GPU reading from memory is fastest if consecutive threads read consecutive locations. This is achieved in both kernels (That is why they are so diferent). Differences in speed are mostly due to different number of threads that can be run for different kernels and problem sizes.

stan/math/opencl/multiply.hpp
```diff
@@ -37,6 +37,28 @@ inline auto multiply(const matrix_cl& A, const matrix_cl& B) {
     temp.zeros();
     return temp;
   }
+  if (A.rows() == 1 && triangular_view_A == TriangularViewCL::Entire
```
rok-cesnovar: Why do we need to check for triangularity here? The general matrix multiply is not faster here even with triangular optimizations, right? Is this only to ensure correct results for cases when someone wants to do a triangular multiply with non-zeroed out matrices and a row_vector/vector?

stan/math/opencl/multiply.hpp
```diff
@@ -37,6 +37,28 @@ inline auto multiply(const matrix_cl& A, const matrix_cl& B) {
...
+      && triangular_view_B == TriangularViewCL::Entire) {
+    try {
+      opencl_kernels::row_vector_matrix_multiply(
+          cl::NDRange(temp.cols() * 64), cl::NDRange(64), A.buffer(),
```
rok-cesnovar: This 64 should be retrieved from the implementations parameters like we read the local sizes:
`int local = opencl_kernels::matrix_multiply.make_functor.get_opts().at("THREAD_BLOCK_SIZE");`

stan/math/opencl/multiply.hpp
```diff
@@ -37,6 +37,28 @@ inline auto multiply(const matrix_cl& A, const matrix_cl& B) {
...
+          B.buffer(), temp.buffer(), B.rows(), B.cols());
+    } catch (cl::Error& e) {
+      check_opencl_error("multiply", e);
```
rok-cesnovar: Change the multiply to row_vector x matrix or something more specific then just multiply.

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -114,6 +114,94 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
...
+     */
+    __kernel void row_vector_matrix_multiply(
+        const __global double* A, const __global double* B, __global double* R,
+        const int N, const int K) {
+      const int lid = get_local_id(0);
+      const int gid = get_global_id(0);
+      const int wgid = get_group_id(0);
+
+      double acc = 0;
+      for (int i = lid; i < N; i += 64) {
```
rok-cesnovar: 64 should be a constant like THREAD_BLOCK_SIZE for matrix multiply. Also add a constant that defines that we are then doing subsequent reduces of size 4.

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -114,6 +114,94 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
...
+     * @param[in] N Number of cols for matrix A and number of rows for vector B
+     */
+    __kernel void matrix_vector_multiply(
+        const __global double* A, const __global double* B, __global double* R,
+        const int M, const int N) {
+      const int gid = get_global_id(0);
+
+      double acc = 0;
+      for (int i = 0, j = 0; i < N; i++, j += M) {
+        acc += A[j + gid] * B[i];
```
rok-cesnovar: You are leaning on GPUs own caching mechanism for memory reads right? Would we benefit with local memory here at all? This one is optional as this version is already faster, so we can leave that one for future optimizations.

rok-cesnovar:
Can you share the benchmark code for these tests?

stan/math/opencl/multiply.hpp
t4c1: Yes. If needed the new kernels could be extended to correctly do triangular multiplication.

stan/math/opencl/kernels/matrix_multiply.hpp
t4c1: Do we really need a constant for reduce size? Where would it be used?

stan/math/opencl/kernels/matrix_multiply.hpp
t4c1: Right, but only for reading elements of B. I think we would likely only see benefit on older GPUs that do not cache reads from global memory.

t4c1:
Benchmark source is here: https://gist.github.com/t4c1/db9f1cde5e98e9ddfeaa7e83e698259b

To run it I modified opencl::multiply to accept additional parameter that controls whether old or new implementation is run.

stan/math/opencl/multiply.hpp
rok-cesnovar: That would be great. I am not sure if these triangular cases come up often, but it would be nice to have these speedups for them.

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -114,6 +114,96 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
...
+          const int wgid = get_group_id(0);
+
+          double acc = 0;
+          for (int i = lid; i < N; i += LOCAL_SIZE_) {
+            acc += A[i] * B[i + wgid * N];
+          }
+          __local double res_loc[LOCAL_SIZE_];
+          res_loc[lid] = acc;
+          barrier(CLK_LOCAL_MEM_FENCE);
+          for (int step = LOCAL_SIZE_ / REDUCTION_STEP_SIZE;
```
rok-cesnovar: Thanks for this change, it looks much more general now. Any performance degradation due to this that you noticed?

stan/math/opencl/kernel_cl.hpp
```diff
@@ -82,7 +82,7 @@ class kernel_functor {
                  std::map<const char*, int> options) {
     auto base_opts = opencl_context.base_opts();
     for (auto& it : options) {
-      if (base_opts[it.first] > it.second) {
+      if (base_opts[it.first] != it.second) {
```
rok-cesnovar: Can you give us some background on why this change?

rok-cesnovar: This is going to override the context calculated value for `base_opts` with the kernel_functor supplied value each time, so we need a better solution here (its why you have failed tests). 

The kernel parameters have to be defined in opencl_context in `base_opts_` and the default value should be overwritten in `opencl_context_base()` if the device info says the default value is not possible. This may seem like you are defining things in 2 places but the idea is:

- define the parameter in `base_opts` and supply it a default value (the default value defined here will only be used if not supplied with the kernel functor or not overwritten by `opencl_context_base()`, most used for enum like defines `UPPER_TO_LOWER`, etc. and if the developer forgets to add a preferred value.)
- if there is a possibility that some device property may restrict its value, overwrite this default value in `opencl_context_base()` (for devices with restrictions (older GPUs and less common architectures))
- supply a prefferred option in the kernel functor

My suggestion would be to change the `LOCAL_SIZE` parameter to something more general like `THREAD_BLOCK_SIZE_1D` and set its default value to a common size useful for any 1D kernel (256 or 512 are usually good magic numbers). This same define can then be used by other kernels with 1D thread grid.
Then have the `opencl_context_base()` overwrite the value it if the max thread block size of the GPU is smaller and
finally supply 64 in the kernel functor if you feel that 64 is the best for this kernel across most GPUs. 


This compile-time kernel parameter stuff is a bit tricky but we havent yet found a way to make this simpler. We need to offer the flexibility of defining your own values and also have a fallback based on GPU properties.

I know that the absence of a thorough doc on this doesnt help, sorry about that. It will be much easier (I hope) once we finish the paper.

stan/math/opencl/kernels/matrix_multiply.hpp
t4c1: No.

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -141,6 +141,125 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
...
+static const char* matrix_vector_multiply_kernel_code = STRINGIFY(
+    // \endcond
+    /**
+     * Matrix-vector multiplication R=A*B on the OpenCL device
+     *
+     * @param[in] A matrix in matrix-vector multiplication
+     * @param[in] B vector in matrix-vector multiplication
+     * @param[out] R the output vector
+     * @param[in] M Number of rows for matrix A
+     * @param[in] N Number of cols for matrix A and number of rows for vector B
```
rok-cesnovar: This should probably state `and number of cols for vector B`.

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -141,6 +141,125 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
...
+     * @param[in] lower_upper_A the triangularity of A (lower, upper or none)
+     * @param[in] lower_upper_B the triangularity of B (lower, upper or none)
+     */
+    __kernel void matrix_vector_multiply(
+        const __global double* A, const __global double* B, __global double* R,
+        const int M, const int N, unsigned int lower_upper_A,
+        unsigned int lower_upper_B) {
+      const int gid = get_global_id(0);
+
+      int start = 0;
```
rok-cesnovar: Mind changing the start to be `const` like this:
`const int start = lower_upper_A == UPPER ? gid + 1 : 0; `

stan/math/opencl/multiply.hpp
```diff
@@ -37,6 +37,30 @@ inline auto multiply(const matrix_cl& A, const matrix_cl& B) {
     temp.zeros();
     return temp;
   }
+  if (A.rows() == 1) {
+    int local_size
```
rok-cesnovar: `const local_size`

stan/math/opencl/opencl_context.hpp
```diff
@@ -134,6 +134,13 @@ class opencl_context_base {
         base_opts_["THREAD_BLOCK_SIZE"] = thread_block_size_sqrt;
         base_opts_["WORK_PER_THREAD"] = 1;
       }
+      if (max_thread_block_size_ < base_opts_["LOCAL_SIZE_"]) {
+        // must be a power of base_opts_["REDUCTION_STEP_SIZE"]
+        int p = std::log(max_thread_block_size_)
```
rok-cesnovar: const p

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -141,6 +141,125 @@ const local_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int, int,
...
+// \cond
+static const char* row_vector_matrix_multiply_kernel_code = STRINGIFY(
+    // \endcond
+    /**
+     * Row vector-matrix multiplication R=A*B on the OpenCL device
+     *
+     * @param[in] A row vector in row vector-matrix multiplication
+     * @param[in] B matrix in row vector-matrix multiplication
+     * @param[out] R the output vector
+     * @param[in] N Number of cols for vector A and number of rows for matrix B
```
rok-cesnovar: rows of the row_vector A

stan/math/opencl/kernels/matrix_multiply.hpp
t4c1: I think that is correct as it is.

rok-cesnovar: Yep, ignore this :)

rok-cesnovar:
Also cheers on your first soon to be merged PR to the Math library!!

Issue #1200, opened 2019-04-15
Add option of having device functions shared across opencl kernels


t4c1:
## Description
Right now the only option of sharing code between opencl kernels is to put it into global kernel helpers, whicha are included into every kernel. This should be reorganized so that it is possible to select which of the common device functions/macros are included into a kernel.

## Example
Matrix addition kernel uses matrix indexing helper macros. That should be specified when creating the kernel.
```
const global_range_kernel<cl::Buffer, cl::Buffer, cl::Buffer, int, int> add(
    "add", {indexing_helpers, add_kernel_code});
```

## Expected Output
No changes in any outputs.

#### Current Version:
v2.19.0

PR #1201, opened 2019-04-15
Add option of having device functions shared across opencl kernels

t4c1:
## Summary

Allows selecting which common functionality is available in each opencl kernel. Existing kernel helpers are also refactored, so that they are not in all kernels by default.

## Tests

No new tests. Existing tests suffice.

## Side Effects

There are no more global opencl kernel helpers. Each kernel must specify what common functionality it needs.

## Checklist

- [x] Math issue #1200


SteveBronder:
Can I get some context on this one? Why is having these the same across each kernel bad?

rok-cesnovar:
We were discussing with @t4c1 how to add helper functions like `digamma` that he is using for the GPU GLM functions [here](https://github.com/bstatcomp/math/blob/gpu_glm2/stan/math/opencl/kernels/neg_binomial_2_log_glm_lpmf.hpp#L19).

The way we have things setup now, the two options would be to

-  copy/paste the entire digamma function to each kernel where its needed
-  add it to the helpers file which would mean that this would be compiled with every kernel, which seems unnecessary

For the amount of helper macros, defines, etc. we have right now, this PR may seem like overkill, but once we have a few more of functions like digamma, this seems like a good solution, I think.

SteveBronder:
Word makes sense!

stan/math/opencl/kernel_cl.hpp
```diff
@@ -146,7 +158,17 @@ struct local_range_kernel {
    */
   local_range_kernel(const char* name, const char* source,
                      const std::map<const char*, int> options = {})
-      : make_functor(name, source, options) {}
+      : make_functor(name, {source}, options) {}
+  /**
+   * Creates kernels that need access to defining the global thread
+   * siez and the thread block size.
```
rok-cesnovar: size

stan/math/opencl/opencl_context.hpp
```diff
@@ -379,6 +379,7 @@ class opencl_context {
     return {opencl_context_base::getInstance().platform_};
   }
 };
+
```
rok-cesnovar: can you remove this or is it autoformatter trying to work his magic?

stan/math/opencl/kernels/helpers.hpp
```diff
@@ -11,7 +11,7 @@ namespace opencl_kernels {
 /*
  * Defines some helper macros for the kernels
```
rok-cesnovar: Please expand these comments with a little infor on what the helpers help at. Same for the thread block helpers. Something like "Helpers for readable matrix element access."

stan/math/opencl/kernel_cl.hpp
```diff
@@ -25,20 +25,22 @@ namespace opencl_kernels {
  * Compile an OpenCL kernel.
  *
  * @param name The name for the kernel
- * @param source A string literal containing the code for the kernel.
+ * @param sources A std::vector of strings containing the code for the kernel.
  * @param options The values of macros to be passed at compile time.
  * @note The macros defined in kernels/helpers.hpp are included in the kernel
```
rok-cesnovar: This note can probably be removed now?

SteveBronder:
I'll take a look today

stan/math/opencl/kernels/helpers.hpp
```diff
@@ -9,9 +9,9 @@ namespace math {
 namespace opencl_kernels {
 
 /*
- * Defines some helper macros for the kernels
+ * Defines helper macros for common matrix indexing operations
  */
-static const char* helpers =
+static const char* indexing_helpers =
```
seantalts: I like that you're leaving them all in the same file. It's nice to think about the use-case of someone new to the code trying to get familiar and write their own kernels or edit existing ones. 👍

stan/math/opencl/kernel_cl.hpp
```diff
@@ -25,20 +25,20 @@ namespace opencl_kernels {
...
  *
  * @param name The name for the kernel
- * @param source A string literal containing the code for the kernel.
+ * @param sources A std::vector of strings containing the code for the kernel.
  * @param options The values of macros to be passed at compile time.
- * @note The macros defined in kernels/helpers.hpp are included in the kernel
- *  compilation for ease of writing and reading kernels.
  */
-inline auto compile_kernel(const char* name, const char* source,
+inline auto compile_kernel(const char* name, std::vector<const char*> sources,
```
seantalts: Very elegant! Excellent :)

seantalts:
I'm a fan of the way this is done. I am not 100% sure it feels totally motivated yet but I'm willing to go with y'all's intuitions / knowledge of future plans to motivate it, and it's still very clean.

stan/math/opencl/kernel_cl.hpp
SteveBronder: Agree! I think Pretty sure we can make that
 `const std::vector<const char*>& sources`

SteveBronder: ^ I think that's true most places. @t4c1 while you are here would you mind cleaning up the signatures for stuff we can make const &?

t4c1: Of course. I'm also making `std::map<const char*, int> options` a reference.

Issue #1210, opened 2019-04-19
Problem with coupled_ode_system


syclik:
## Description

There is a memory problem with the ~`coupled_ode_system`~ `coupled_ode_system` test; it is indexing past the end of a `std::vector`. I've only verified that this happens on one instantiation of the many different instantiations and I believe this problem may exist in some more of these instantiations.

This problem doesn't show up all the time because we've used `[]` for indexing, which does not check for bounds. By changing the implementation (at least temporarily) to use `at()` for indexing, it's clear that this problem exists.

This memory issue was masked for a long time since it's indexing into memory the process owns. It's only triggered a failed test in #1171.

This is actually a ~pretty big~ bug in the test:
1. it may cause programs to seg fault without indicating why
2. the results from using the ODEs under these conditions are undefined. It's clearly not computing the correct thing.

## Example

Check out branch `bugfix/issue-1210-coupled_ode_system`. Run:
```
./runTests.py test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp
```

Current output:
```
$ ./runTests.py test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp
------------------------------------------------------------
make -j1 test/unit/math/rev/arr/functor/coupled_ode_system_test
make: `test/unit/math/rev/arr/functor/coupled_ode_system_test' is up to date.
------------------------------------------------------------
test/unit/math/rev/arr/functor/coupled_ode_system_test --gtest_output="xml:test/unit/math/rev/arr/functor/coupled_ode_system_test.xml"
Running main() from lib/gtest_1.8.1/src/gtest_main.cc
[==========] Running 18 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 18 tests from StanAgradRevOde
[ RUN      ] StanAgradRevOde.coupled_ode_system_dv
[       OK ] StanAgradRevOde.coupled_ode_system_dv (0 ms)
[ RUN      ] StanAgradRevOde.decouple_states_dv
[       OK ] StanAgradRevOde.decouple_states_dv (0 ms)
[ RUN      ] StanAgradRevOde.initial_state_dv
[       OK ] StanAgradRevOde.initial_state_dv (0 ms)
[ RUN      ] StanAgradRevOde.size_dv
[       OK ] StanAgradRevOde.size_dv (0 ms)
[ RUN      ] StanAgradRevOde.memory_recovery_dv
test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp:152: Failure
[...]

PR #1214, opened 2019-04-23
Bugfix/issue 1210 coupled ode system test

wds15:
## Summary

This PR fixes a problem found in the test of the `coupled_ode_system`. The test `test/unit/math/rev/arr/functor/coupled_ode_system_test.cpp` called the `operator()` of the `coupled_ode_system` with a state which corresponded in length to the base system `y`, but the coupled ODE system expects that state to be equal to the length of the coupled ODE system state, which is of length `size()` as defined by the `coupled_ode_system`.

The bug lead to accesses past the end of the input vector which showed up in failing unit tests on Windows under very special circumstances.

Changes made:

- Correct the test to use `coupled_ode_system` according to its spec
- Went over the tests themselves and got notation more aligned with our definitions (`y` for the base system, `z` for the coupled system)
- Improved the doc for the `operator()` of the `coupled_ode_system` to make pre- and post-conditions more clear.

## Tests

No new tests. This PR fixes a test to align the test with the spec.

## Side Effects

Nope.

## Checklist

- [X] Math issue #1210 


syclik:
Thanks for taking care of this so quickly. I'll review when I'm in the right mindset to read the code careful. Hopefully in the next day or two. (but definitely by this weekend)

syclik:
(I tried looking at it right now and I just couldn't get through the doc...)

If anyone else wants to review, please do. Some specific things to review (this is what I'd be checking for):
1. the doc follows the guidelines here: [How to Write Doxygen Doc Comments](https://github.com/stan-dev/stan/wiki/How-to-Write-Doxygen-Doc-Comments)
2. the doc is understandable for devs. It clearly describes what it does, how to construct it, how to use it properly, what assumptions it's making, what part of the object is exposed to other parts of the Math library.
3. the tests make sense. This is really a judgment call, but making sure it tests the parts of the class that are exposed, the tests touch some of the edge cases, the tests aren't violating assumptions (this is what caused the bug in the first place), that there are enough tests in place.

wds15:
@seantalts... maybe you wanna have a look and get a dive into the ODE system? A fresh set of eyes on this code would probably be good in any case.

This PR is hopefully fixing our Windows trouble when merging the faster TLS PR into develop. So once this is in, we can go ahead and merge the PR for the faster TLS which also brings the fix for Windows threading.

wds15:
Not sure where doc is wrong... will wait for your comments.

Wrt. to the tests: I have rewritten the tests to use a more consistent notation. Since the tests were not used according to the spec of `coupled_ode_system` I don't see how I can add a test. The only thing I could do is to check in the operator() of `coupled_ode_system` that the states passed into it have the right length (so check that `z` and `dz_dt` have length `size()`)... but I understood we agreed to not do this as any check in this code will be visible performance wise and as this is an internal interface within Stan-math we do not need a check for this.

However, if I do not add an additional check, then there is no way for me to test that `z` has the the right length. The length of the output is given by `dz_dt` given as output.

Can you suggest something?

To me this is a matter of

1. fixing the test code - done by correcting it and making the test code easier to follow (consistent convention of `y` for base and `z` for coupled system).
2. improving the doc - I have copied the note about the length requirement for `z` and `dz_dt` to all specialisations.

Should improving the doc take a bit longer - would it be an option to file another issue for that, but merge this now (unless you have a proposal for additional tests, but I don't see them right now). I am saying this as it gates the threading PR, but we can go either way.

syclik:
Doc comments:
- [coupled_ode_system.hpp#L23](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L23): as a reference, see [How to Write Doxygen Doc Comments](https://github.com/stan-dev/stan/wiki/How-to-Write-Doxygen-Doc-Comments) wiki page. This class doc doesn't actually discuss what the class does or why this class should be used. It should really state the purpose of the class (being an adaptor that exposes a functor for ODE integrators to use...). 
- [L28](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L28): this is where it's technically incorrect. The doc states that there are `N + N * M` states, but then it goes to get into detail stating there are `N + 2 * M` states. Something is off.
- [L98](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L98): this looks like a stray comment
- [L91-92](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L91): it would really help to describe the parameters here. We should expect `z` to be of size `___`, does `dz_dt` need to be the right size and what is the size when it's complete, what are the conditions on `t`? We essentially need to write out what needs to hold for this to work since we're not checking inside the function (preconditions). If there are any side effects, that should be listed. Here, since we're messing with the autodiff stack, one of the things we need to mention is that a concurrent autodiff operation can't happen while this function is operating or else things are undefined.
- [L98](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L98)
- [L162](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L162): [unrelated to this PR] does this look really inefficient to you? Shouldn't it just return a copy of `y0_dbl_`? 
- [L170](https://github.com/stan-dev/math/blob/a74faebfd9a6a171cd4727b0c3e0c0994e743a53/stan/math/rev/arr/functor/coupled_ode_system.hpp#L170): this doc needs to be updated. It's missing a return statement (which is the most important thing for this function). It doesn't quite describe enough for someone to understand what the method is doing. It could use some help with what the input parameter is.

I bet the same comment comments apply to the next specialization. Maybe it's not so clear what the API of the `coupled_ode_system` class actually is from the doc and that's causing some of the confusion. (It was enough for me not to understand easily that the test was sending incorrect values.)

Btw, I'm not looking for perfect doc.... but it has to be enough where we won't run into this problem again. Let me know if you agree or disagree on the points above. I might have misunderstood what you meant by your doc and it'd be better if we just reworded the parts that weren't clear.

wds15:
Sounds like a lot of technical debt laundry to take care of here... let me take a stab according to your findings.

wds15:
This PR and the Intel TBB one suffer from insufficient disk space during the distribution test phase. Can we do something about this other than restarting the testing and crossing fingers?

rok-cesnovar:
I think that all PRs that pass through "Always run part 1" suffer from the same issue. Paging @serban-nicusor-toptal

seantalts:
The intention for that kind of problem with Jenkins is to fix it, not to restart. @serban-nicusor-toptal is off this week so I'll take a look at it.

syclik:
@wds15: I'd really like to see one new test that actually makes sure we're using the values correctly. We ended up here because of (at least) two things: the doc wasn't clear enough for a reviewer to tell that the test was wrong and the test wasn't actually using the values in any meaningful way.

Isn't there some way to check the output? Maybe use decouple system with explicit values and make sure those are propagated as gradients? Or have the state be non-zero and validate that the same test is correct? (using 0 in this case isn't useful because of the multiplication property of zero.)

P.S. edits don't propagate as notifications.
P.P.S. I'll also help on doc later today.

seantalts:
@bgoodri I think this is because gelman-group-linux is running out of space (has "only" 44G free). Looks like /home/jenkins-slave is only taking up 7GB. Is there anything else you could free up?

I'm re-running the tests and told Jenkins not to use that machine for distribution tests anymore, which should cause it to spin up an ec2 node for it, which is probably better anyway, so hopefully it works this time.

bgoodri:
I'll look at it. We shouldn't be using 40G for the tests.

seantalts:
The distribution tests as written a few years ago use more than 100GB if I recall correctly.

wds15:
@syclik We do have that test already which you are asking for. So the test here:

`test/unit/math/rev/arr/functor/integrate_ode_rk45_grad_test.cpp`

does test explicitly the numerical correctness of the gradients of a toy system. This test will break if the coupled ode system would give us wrong numerical results... and that test also uses non-zero values in all states.

This is why I concluded that everything is just fine.

However, what I really do not see is how we can test the input sizes to the coupled_ode_system functor being right - but as I say we do test that the thing returns correct results via the grad tests.

syclik:
How did that test pass with the bug?!?! It shouldn't have.

syclik:
Wait... I get it. There was never a bug. Ok. We're good.

syclik:
(there was never a bug in the code, just in the test)

syclik:
I started working on this, but I'm falling asleep now. I'll finish up tomorrow.

bgoodri:
/var/log/jenkins/jenkins.log is taking up 667G . I could delete it or
compress it but the deeper question seems to be WTF?

seantalts:
Hahaha, that does seem bad. I'll take it from here I guess, maybe we can
figure out why it's huge. Thanks! I'll check there next time, woops.

syclik:
So... the one thing that does bother me is that there's no unit test that fails on bad behavior. To rely on what's essentially an integration test to catch it isn't good practice. It leads to problems like we're seeing. (I'm not trying to say it's necessary, I'm just voicing a concern here.)

I'll continue to update the doc and we'll merge after the doc is at a good enough spot.

wds15:
That’s right...we never test with a unit test the correctness of the sensitivities. This only done as a integration test. However, this additional unit test would not have caught the bug of the unit test at all. We should file another issue for this if you think that should be addressed in order to improve test coverage, but since this is tested we are safe anyways.

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -323,14 +322,16 @@ struct coupled_ode_system<F, var, double> {
...
-   *
+   * 
    * <p>Because the starting state is unknown, the coupled system
    * incorporates the initial conditions as parameters.  The
    * initial conditions for the coupled part of the system are set
    * to zero along with the rest of the initial state, because the
    * value of the initial state has been moved into the
    * parameters.
    *
+   * FIXME(syclik): this documentation is incorrect.
```
syclik: @wds15, can you update this doc? It's clearly wrong; it states that all the elements should be 0 but are 1 instead.

stan/math/rev/arr/functor/coupled_ode_system.hpp
```diff
@@ -534,6 +531,8 @@ struct coupled_ode_system<F, var, var> {
    * a parameter, and hence the return of this function is a
    * vector of zeros.
    *
+   * FIXME(syclik): this documentation is incorrect
```
syclik: Same here.

wds15:
Hmm... I just found that you, @syclik , added to the doc of the `coupled_ode_system`

`* This method uses nested autodiff and is not thread safe.`

... but why should nested AD not be thread safe here? I mean, having `map_rect` use this facility during ode integration done in parallel is totally fine. Or how did you mean this? For a given thread running the ODE integration does indeed disallow running anything in parallel in the context of the thread's AD tape, of course (so things depend on whether you have thread local AD tapes or not). It's very subtle, but giving a blanket statement like the one above is somewhat confusing (like I say, it implies to me that we cannot use this with `map_rect`).

Threading and nesting can be tricky, yes; but per se it's safe here, I think.

syclik:
By “not thread safe,” I mean the text book definition of it. Wikipedia describes it a little bit: https://en.m.wikipedia.org/wiki/Thread_safety

Caveat: nothing we’ve written is truly thread safe, but we could fix that by locking placement new (I think). That still wouldn’t fix these methods. 

For these methods, it’s really easy to see. If one of the functors was called, it adds a nested level to the autodiff stack, then pops it off. It actually does it repeatedly through the functor. Let’s say there was another function, for concreteness say `sin`. If you called that function while the functor was operating, it would put a var on the nested level and it would be removed. That makes it inherently not thread safe. 

It doesn’t mean that you can’t construct a threaded operation using this functor. That’s not what it means. It means that if multiple threads operate, that it would mess with the guarantees that is required. (`sin` needs to put a `var` on the autodiff stack.)

seantalts:
Given that nothing in the Math library is thread-safe by default (to be clear to anyone else reading this - we aren't talking about using `-DSTAN_THREADS=true` here - that would make these functions and the rest of the library thread-safe, though we only guarantee that used via `map_rect`), why annotate this function with that doc? I imagine you had some use case in mind, I just can't think of one.

syclik:
These methods explicitly break the ability to be thread safe. It is worth documenting. There's nothing wrong with non thread safe behavior as long as we mark it for other devs to know they shouldn't blindly use these when parallelizing.

The use case? `map_rect` isn't the only way to use threading (now that we have C++14, we can actually try to use threads sensibly). 

We can build an expression graph up by parallelizing computation if we're careful. As I mentioned before, we're nearly there: if we're careful in handing out memory locations, we'd actually be able to generate parts of the expression graph in parallel. Except where we use nested autodiff.

wds15:
To me we have in Stan-math right now just two cases:

1. no thread support whatsoever whenever STAN_THREADS is NOT set. Then we have a single global AD tape.
2. thread support turned on, so STAN_THREADS is set. In this case each thread will have it's thread-local AD tape.

In both cases everything works just fine and we do not support any intermediate between the two.

The case of *concurrent* access to a single AD tape is not supported in Stan-math. This does not work for many reasons. Just two reasons:

a) pushing new vars to the tape with the std::vector push_back is not thread safe
b) nested AD is even worse in that one thread causing a nested AD will make any other access to the AD tape from a different thread impossible.

To me we support only the case 1 and 2 above and as such the statement that things are not thread safe here is confusing. In that sense nothing is thread safe in Stan-math. The nested AD is admittedly even worse in this regard, but no different.

The two problems a+b I mention are all resolved in the parallel design document which is up for review.

seantalts:
I'm not trying to be difficult, I just didn't know there was any way anyone could be using the Math library in a thread-safe way without STAN_THREADS. I think the documentation should maybe explain a little more what it's talking about - reading what's in there now and you'd come away thinking you couldn't use ODE solvers with STAN_THREADS enabled.

> We can build an expression graph up by parallelizing computation if we're careful. As I mentioned before, we're nearly there: if we're careful in handing out memory locations, we'd actually be able to generate parts of the expression graph in parallel.

That sounds cool! This is news to me - is there work or discussion about this somewhere? I admittedly am not up to date on discourse unless people explicitly tag me :x Seems like someone attempting this would be making significant changes to the AD system (given Sebastian's post above this one) such that they're probably off in their own land of unsupportedness and would need to be thinking carefully about thread safety on a different basis than the official Math library. 

Also curious about how the parallel changes Sebastian proposes interact with that project. I'd expect it to make it strictly easier, but don't know the details of the carefully-handing-out-memory-locations project.

bob-carpenter:
> (@syclick): These methods explicitly break the ability to be thread safe. It is worth documenting. 

> (@syclick): we could fix that by locking placement new

Almost nothing is thread safe in Stan.  Placement new isn't the problem.  
Almsot every one of our exposed functions breaks thread safety by allocating on the autodiff stack without synchronization.  


> (@seantalts)  I think the documentation should maybe explain a little more what it's talking about - reading what's in there now and you'd come away thinking you couldn't use ODE solvers with STAN_THREADS enabled.

I think that's a good compromise.

bob-carpenter:
> On May 8, 2019, at 9:23 AM, Daniel Lee <notifications@github.com> wrote:
> 
> (@bob-carpenter: that was the wrong handle.)
> 
> First, this wasn't meant to be controversial. What I doc'ed was just technically true.
> 
> Almost nothi'ng is thread safe in Stan. Placement new isn't the problem.
> Almsot every one of our exposed functions breaks thread safety by allocating on the autodiff stack without synchronization.
> 
> Isn't placement new where things get allocated on the autodiff stack?

No. The memory gets allocated by our memalloc class.  The placement new is just used to build into already allocated memory and can be done concurrently.  

If we completely synchronized the operator new specialization for vari with locks, that would only solve the problem of allocating memory on the heap---it wouldn't help with the reverse pass, say.  Nor would it let us race ahead and compute downstream until we get the value from the upstream calculation.

bob-carpenter:
> On May 8, 2019, at 9:40 AM, Daniel Lee <notifications@github.com> wrote:
> 
...

> I thought we got past the "supported API" discussion.

I haven't in that I still don't understand what I can and can't change.  So I'll just have to try things and hope whoever's code reviewing understands the notion of API better than I do.


> From the Math library perspective, we've got to tell developers what is safe or not with respect to C++, not only the "supported API."

I don't know where to the draw the line here, either.  We could start adding notes about malloc being unsafe, etc., everywhere as hints for future developers, but it seems like we probably want a higher-level description of memory and threading instead.

> 
> By "nearly there," I literally mean if we lock placement new and anywhere we toss things onto the autodiff stack (which I think it only done with placement new), we'd be complete.

This isn't sufficient.  None of the chain() methods can use threading to access the expression graph.  Nothing internal to those constructors can multithread direct calls to malloc (like array_alloc, etc., as used in the matrix functions everywhere).  This does nothing to help with the reverse pass.

syclik:
Yup. Thanks for clarifying. Yes, that's what I was actually thinking about.
As long as placement new behaves well concurrently, which I didn't know if
it would for sure, conditioned on the allocation being done safely, we
should be ok. (with the stuff about the forward pass and not being able to
use downstream computations being true)

And that's where these methods that were doced here are just different.
Since it uses nested autodiff, it would pop off things that are placed on
the stack. And that's something worth documenting.

I'm thinking the clarifying doc should be on map rect and it should state
that it deals with non-thread-safe functions safely. Does that make sense?
Maybe it already states it.

PR #1215, opened 2019-04-23
[WIP] Dummy PR for Boost trim

rok-cesnovar:
## Summary

This is just a WIP PR in order to run Jenkins tests to see which files can safely be deleted from the Boost library. Not intended for review.

Initial commit reduces the size of Boost from 608,5 to 193,9MB

```
develop
66805 items, totalling 608,5 MB

rm -R doc
58550 items, totalling 524,2 MB

rm -R libs/*/test/
45252 items, totalling 442,4 MB

rm -R libs/*/example/
41527 items, totalling 426,3 MB

rm -R libs/*/doc/
22183 items, totalling 193,9 MB
```

If this goes through, we can explore further if needed.


rok-cesnovar:
Ok, it seems that going further than what it written in the above post would require modification to boost files so I am stopping here. Will open a new PR that will be prepared using an upgrade script.

Issue #1197, opened 2019-04-04
Use Eigen's plugin system to extract/manipulate values from autodiff variables


andrjohns:
## EDIT: Description
Following discussions below, Eigen's plugin system will be used to implement the following views:

```Matrix<var, R, C>```:

-    ```.adj()```
-    ```.val()```
-    ```.vi()```

```Matrix<vari*, R, C>```:

-    ```.adj()```
-    ```.val()```

```Matrix<fvar<T>, R, C>```:

-    ```.d()```
-    ```.val()```

## Description
Currently, when working with a matrix of autodiff variables (fvars or vars), the values and gradients first need to extracted to separate matrices before they can be manipulated. We could instead use [Eigen's plugin function ](https://eigen.tuxfamily.org/dox/TopicCustomizing_Plugins.html) to add methods for directly accessing the values and gradients.

In it's simplest form, this would mean replacing the following loops:
```
  Eigen::Matrix<T, R, C> m_deriv(m.rows(), m.cols());
  Eigen::Matrix<T, R, C> m_val(m.rows(), m.cols());

  for (size_type i = 0; i < m.rows(); i++) {
    for (size_type j = 0; j < m.cols(); j++) {
      m_deriv(i, j) = m(i, j).d_;
      m_val(i, j) = m(i, j).val_;
    }
  }
```
with:
```
  Eigen::Matrix<T, R, C> m_deriv = m.d_();
  Eigen::Matrix<T, R, C> m_val = m.val_();
```
But it would also allow for the integration within other eigen commands (e.g. ```m.val_().trace()```) which could cut down on some of the existing code.

## Example
I've got an initial attempt working for extracting the gradients from a matrix of ```fvar<double>```:

Create a file "Eigen/MatrixBaseAddons.h":
```
template<typename T>
struct d_Op {
  EIGEN_EMPTY_STRUCT_CTOR(d_Op)
  typedef T result_type;
  EIGEN_DEVICE_FUNC
  EIGEN_STRONG_INLINE const result_type& 
    operator()(const stan::math::fvar<T> &v) const { return v.d_; }
};

template<typename T>
struct val_Op {
  EIGEN_EMPTY_STRUCT_CTOR(val_Op)
  typedef T result_type;
  EIGEN_DEVICE_FUNC
  EIGEN_STRONG_INLINE const result_type& 
    operator()(const stan::math::fvar<T> &v) const { return v.val_; }
};


[...]

PR #1216, opened 2019-04-24
Issue 1197 - Use Eigen's plugin system to views for autodiff types

andrjohns:
## Summary
As discussed in more depth in #1197, this PR adds methods to Eigen for returning vals/adjoints/tangents/varis from matrices of autodiff types.

The following methods have been added:

```Matrix<var, R, C>```:

-    ```.adj()```
-    ```.val()```
-    ```.vi()```

```Matrix<vari*, R, C>```:

-    ```.adj()```
-    ```.val()```

```Matrix<fvar<T>, R, C>```:

-    ```.d()```
-    ```.val()```

## Tests

No new tests have been added, as the methods maintain existing functionality but with less code

## Side Effects
Requires that ```prim/mat/fun/Eigen.hpp``` be included before any other Eigen header, otherwise the new methods won't be available

## Checklist

- [x] Math issue #1197

### CODE REVIEW

File: stan/math/prim/mat/matrix_addons.h [hunk 1]
```diff
@@ -0,0 +1,206 @@
...
+  typedef typename std::remove_pointer<T>::type decay_type;
+    struct Fallback { int d_; };
+    struct DerivedType : decay_type, Fallback { };
+
+    template<typename C, C> struct ChT; 
+    template<typename C> static char (&f(ChT<int Fallback::*, &C::d_>*))[1]; 
+    template<typename C> static char (&f(...))[2]; 
+
+    static bool const value = sizeof(f<DerivedType>(0)) == 2;
+};
```
SteveBronder: Neat! Reviewing this is certainly above my paygrade, but would you mind doc'ing more about what's going on here? For instance the line

```cpp
template<typename C> static char (&f(ChT<int Fallback::*, &C::d_>*))[1]
```

That's kinda wild looking, not sure how to read that

bob-carpenter: > That's kinda wild looking, not sure how to read that

I agree.  We should assume the reader is good at C++, but this one could use some unpacking.  At the very least, template parameters should get documentation.

In general, we should favor includes to directly including declarations like this as they're easier to maintain.  In some cases, it can help to just include declarations directly, but I'm not sure this is one of them.  

Some typedefs might help.

I can break it down piecewise to be declaring the function `f` to apply to some complex pointer to `ChT` and return something that can be dereferenced and applied as an array to index 1 to produce a `char`.  

The bit I couldn't parse was `int Fallback::*`.  What is that?

SteveBronder: Googling, Fallback::* is a pointer the member function which is declared as an int so doing implicit conversion.

idt an int is always guaranteed to be big enough to hold any pointer (from [here](https://stackoverflow.com/a/25225033/2269255)). So maybe that should be something like

```cpp
template<typename C> static char 
(&f(ChT<uintptr_t reinterpret_cast<uintptr_t>(Fallback::*),
 &C::d_>*))[1]; 
```
Where `d_` is also a `uintptr_t` (?)

andrjohns: I can't take the credit for this clever piece of work. I took the approach from [this blog post](https://cpptalk.wordpress.com/2009/09/12/substitution-failure-is-not-an-error-2/).

I had to resort to something like this because you can't ```#include``` external headers in the ```matrix_addons``` file. The other alternative would be to include the appropriate header before declaring the add-on in ```prim/mat/Eigen```. But I don't think that would be a good approach since it would create a more complex requirement on the order of header includes

SteveBronder: Huh! Pretty cool blog, may make me a bit too clever if I read his stuff

> I had to resort to something like this because you can't #include external headers in the matrix_addons file. The other alternative would be to include the appropriate header before declaring the add-on in prim/mat/Eigen. But I don't think that would be a good approach since it would create a more complex requirement on the order of header includes

Even for standard library headers? Cuz it looks kind of like this is doing something like `enable_if_t<is_fvar<T>, blah blah>`

SteveBronder: ^By "Even for standard library headers" I meant, is including a standard library header before declaring the add on fall into the realm of "too complex". If we doc the above well then I'm fine with it, but a little bit of complexity in the header order is also a fair trade for reduced code complexity here

SteveBronder: If we wanted to catch people who do the order in the wrong way we could also have do something like define `STAN_MATH_EIGEN_ADDONS_CHECK` in Eigen.hpp, then check and throw something nice during complication if the header wasn't included in the right order

andrjohns: If we use other math headers to differentiate between ```fvar<T>``` and ```var``` then we'd have to include headers from both ```rev``` and ```fwd``` when ```prim/mat/fun/Eigen.hpp``` is included, which could make things a bit messy.

I've updated the doc on that class, but let me know if there's a better approach I could be using here

andrjohns:
@seantalts looks like jenkins is having some trouble with the distribution tests:
```
fatal error: error in backend: IO failure on output stream: No space left on device
<builtin>: recipe for target 'test/prob/exp_mod_normal/exp_mod_normal_ccdf_log_00004_generated_v_test.o' failed
```

stan/math/prim/mat/matrix_addons.h
andrjohns: Turns out this method has been superseded by a much [simpler approach using ```void_t```](https://en.cppreference.com/w/cpp/types/void_t). Makes the code a bit more readable!

syclik: @andrjohns just thinking out loud... if you're having to redefine traits that we already have, would it just be easier to not have this as a plugin and just create a view? I think you're running into the dependency of: the Eigen plugin needs to know about Math's autodiff types AND Math's traits are set up so that we have specialization for Eigen's types.

I think that problem could be avoided by writing a view where we already know about both Eigen and Math.

seantalts:
Oh, I replied to this in an email but it isn't showing up, weird. anyway, tagging @serban-nicusor-toptal re: Jenkins

syclik:
Yeah... an email I sent yesterday just bounced back to me. I wrote this immediately I saw the PR yesterday. 

> Thanks. There should be unit tests for all the new methods were added. We want to make sure we can instantiate these class methods properly and they function properly.

stan/math/prim/mat/matrix_addons.h
SteveBronder: @syclik so my interest in this PR is a bit selfish. I'm working on a PR right now that would make the OpenCL API look v similar to what Andrew has here.

It's a good bit down the road, but in the stan3 compiler Sean wants to be able to generate code that directly work with `matrix_cl`s (instead of passing back and forth inside of our normal functions). If we can line up the regular api and OpenCL api's I'm pretty sure we could do something like template the signatures to accept either Eigen matrices or matrix_cl types and then it would "just work"

andrjohns: > if you're having to redefine traits that we already have, would it just be easier to not have this as a plugin and just create a view?

I would agree if the added code was substantial, but it's ended up only being a few lines:
```
template<class, class = void>
struct Hasd_ : std::false_type
{ };
template<class T>
struct Hasd_<T, decltype((void)(T::d_))> : std::true_type
{ };
```
So I think we should leave it with the plugin, to be honest.

andrjohns:
You're right, I'll get those tests written and added in

stan/math/prim/mat/matrix_addons.h
syclik: Cool. (I couldn't tell how much duplicated effort there was.)

stan/math/rev/mat/fun/Eigen_NumTraits.hpp
```diff
@@ -79,6 +79,10 @@ struct NumTraits<stan::math::var> : GenericNumTraits<stan::math::var> {
 };
 
 namespace internal {
+template <>
+struct remove_all<stan::math::vari*> {
+  typedef stan::math::vari* type;
+};
```
SteveBronder: Not seeing this used anywhere?

stan/math/prim/mat/matrix_addons.h
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Structure to return a view to the values in a var, vari*, and fvar<T>.
+ * To identify the correct member to call for a given input, templates
+ * check a combination of whether the input is a pointer (i.e. vari*)
+ * and/or whether the input has member ".d_" (i.e. fvar).
+ *
+ * The operators are overloaded for both const and non-const inputs
+ */
+struct val_Op{
+  EIGEN_EMPTY_STRUCT_CTOR(val_Op);
```
SteveBronder: Some some Eigen specifics it may be nice to doc them with a link to the place in Eigen where they are doc'd

https://eigen.tuxfamily.org/dox/XprHelper_8h_source.html

stan/math/rev/mat/fun/Eigen_NumTraits.hpp
andrjohns: Forgot to add doc for this. Needed to add this specialisation to stop Eigen from removing pointers when working with ```vari*```. Otherwise the ```.vi()``` method would return ```Matrix<vari,R,C>``` rather than ```Matrix<vari*,R,C>```

andrjohns:
The order of includes is now causing trouble in the upstream tests. It looks like several Stan headers are including ```Eigen/Dense``` before any of the Math headers get included, which stops the added methods from being loaded and causes compile errors.

A simple workaround would be to replace ```#include <Eigen/Dense>``` with ```#include <stan/math/prim/mat/fun/Eigen.hpp>```

@bob-carpenter would this cause any issues on the Stan side? Or is there a better way?

bob-carpenter:
I'm pretty sure <stan/math/prim/mat/fun/Eigen.hpp> must be included before any other Eigen include for Stan.

andrjohns:
That doesn't seem to be the case for the header tests though. For example, the headers test for ```src/stan/mcmc/hmc/nuts/softabs_nuts.hpp``` fails, because ```src/stan/mcmc/sample.hpp``` includes ```Eigen/Dense```

syclik:
Bob's right. The way we have our traits set up, we need to include
stan/math/prim/mat/fun/Eigen.hpp first. This also simplifies the plugin
architecture: there's no need to worry that you called Eigen before
defining that preprocessor variable.

It seems like we need to update the includes of src/stan/mcmc/sample.hpp.
For the Stan library, for a lot of operations we don't need to include Math
traits defined for Eigen first, but if we do use the plugin, we will have
to.

andrjohns:
Right, if we only need to change the headers that are causing the header tests to fail, and not all of the headers, then only these two need to be changed:

- ```stan/mcmc/sample.hpp```
- ```stan/model/model_functional.hpp```

Then everything passes fine

syclik:
If you don’t mind, please create an issue to fix those includes. Those
should be fixed regardless. I think we should provide more guidance to
clients that also use the Eigen library.

With that in mind, this limits a client from defining a second set of Eigen
plugins?

andrjohns:
> With that in mind, this limits a client from defining a second set of Eigen plugins?

Good question! They couldn't define a second file containing plugin methods (otherwise the preprocessor token ```EIGEN_MATRIXBASE_PLUGIN``` would have multiple definitions), but they could add them to the existing ```matrix_addons.h``` file

seantalts:
You could always `#undef EIGEN_MATRIXBASE_PLUGIN` before redefining it to point to a custom plugins file which includes our matrix_addons.h file.

andrjohns:
> You could always #undef EIGEN_MATRIXBASE_PLUGIN before redefining it to point to a custom plugins file which includes our matrix_addons.h file.

As in:
```
#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/matrix_addons.h"
#undef EIGEN_MATRIXBASE_PLUGIN
#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/matrix_addons_2.h"
```

That also doesn't work, since the methods in the second file simply overwrite the methods in the first.

A good approach could be to create a folder, something like ```prim/mat/plugins```, which would contain the files with plugin methods and have a single header that includes them.

For example, creating a file: ```prim/mat/eigen_plugins.h```, which contains:
```
#include <stan/math/prim/mat/plugins/matrix_addons.h>
#include <stan/math/prim/mat/plugins/matrix_addons_2.h>
```

Then we can just use ```#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/eigen_plugins.h"```

seantalts:
To clarify, what I was saying was that "stan/math/prim/mat/matrix_addons_2.h" would include "stan/math/prim/mat/matrix_addons.h". I think Daniel is asking here about 3rd party clients who want to use the Math library without editing it. The trick might be seeing if they can figure out which headers to use such that they can basically

```cpp
#include <lots of stan stuff including setting EIGEN_MATRIXBASE_PLUGIN but not Eigen>
#undef EIGEN_MATRIXBASE_PLUGIN
#define EIGEN_MATRIXBASE_PLUGIN "my_addons.h" // which includes stan/math/prim/mat/plugins/matrix_addons.h
#include <the rest of Stan stuff including Eigen>
```

andrjohns:
That makes sense, sorry for the misunderstanding.

I think it might be a good idea to just get the methods in with this pull and open a separate issue for the implementation of other plugins to be honest

syclik:
That's right. It's ok that we limit the ability to customize Eigen when
using Math, but we should clearly doc it so clients know that they won't be
able to use plugins easily.

I don't think this has any negative impact on Stan, so it should be fine.

SteveBronder:
Is this still waiting on the headers issue?

rok-cesnovar:
No, @syclik reviewed the Stan repo today.

andrjohns:
Now that this is passing the upstream tests, it's ready for a final review

stan/math/rev/mat/fun/determinant.hpp
```diff
@@ -27,33 +27,18 @@ class determinant_vari : public vari {
...
                 sizeof(vari*) * A.rows() * A.cols()))) {
-    size_t pos = 0;
-    for (size_type j = 0; j < cols_; j++) {
-      for (size_type i = 0; i < rows_; i++) {
-        A_[pos] = A(i, j).val();
-        adjARef_[pos++] = A(i, j).vi_;
-      }
-    }
+    Eigen::Map<Eigen::MatrixXd>(A_, rows_, cols_) = A.val();
+    Eigen::Map<matrix_vi>(adjARef_, rows_, cols_) = A.vi();
```
SteveBronder: It feels like we'll be using these maps a lot more, may be a good alias candidate

stan/math/prim/mat/matrix_addons.h
```diff
@@ -0,0 +1,206 @@
...
+ * The operators are overloaded for both const and non-const inputs
+ *
+ * For definitions of EIGEN_EMPTY_STRUCT_CTOR, EIGEN_DEVICE_FUNC, and
+ * EIGEN_STRONG_INLINE; see: https://eigen.tuxfamily.org/dox/XprHelper_8h_source.html
+ */
+struct val_Op{
+  EIGEN_EMPTY_STRUCT_CTOR(val_Op);
+  template<typename T = Scalar>
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE
+    typename std::enable_if_t<std::is_pointer<T>::value, const double&>
```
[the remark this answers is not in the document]
SteveBronder: Actually how would you feel about doing the substitution check in the template instead of the return. It feels like that would look cleaner

syclik:
@andrjohns, it looks like you've mentioned the side-effect of requiring `prim/mat/fun/Eigen.hpp` to be included prior to Eigen being included. That makes sense. Are there any other side-effects that we should know about? Does this behave properly with all our template traits if included in this manner?

Part of me thinks that this is the complicated way of implementing this functionality. Writing views would simplify the order dependency of includes. That said, I'm not opposed to this if it's easily maintained and is actually easy to use.

SteveBronder:
Andrew when you get a chance could you try this on a the Cholesky rev code? I tried that tonight but was having a lot of difficulty. I think it's nice and pretty non-trivial to sort out. Will probs help us verify this works for more complicated examples. 

When I tried it tonight I was getting a lot of errors where it was trying to use the `const` accessor for types that were non-const. It was also throwing the error `THIS_COEFFICIENT_ACCESSOR_TAKING_ONE_ACCESS_IS_ONLY_FOR_EXPRESSIONS_ALLOWING_LINEAR_ACCESS`

for things like `L.val().triangularView<Upper>().solveInPlace(L.adj().transpose());`

andrjohns:
Yep, can do! I'm working through my dev backlog this week, so I'll do some testing on this at the same time.
________________________________

syclik:
@roualdes, please take a look.

syclik:
note to self: if I can come up with a view implementation quickly, do it and let @rok-cesnovar  try it out. This message times out July 4, 2019. If I don't do anything by then, there are no technical objections from me.

andrjohns:
> things like L.val().triangularView<Upper>().solveInPlace(L.adj().transpose());

@SteveBronder This one sent me down a bit of a rabbit hole! The code only works with ```const``` inputs, e.g.:
```
const matrix_v const_mat_in = matrix_v::Random(100,100);

MatrixXd tri_out = const_mat_in.val().triangularView<Eigen::Upper>().solve(const_mat_in.adj().transpose());
```

This was because non-const inputs call ```CWiseUnaryView``` and const inputs call ```CWiseUnaryOp```. After some digging, this turned out to be a bug in ```CWiseUnaryView``` which was causing a compilation error when multiplying two views returning doubles ([more background here if interested](https://stackoverflow.com/questions/56635122/using-coeffref-with-const-cwiseunaryview-failing-when-multiplying-two-cwiseuna/56661442?noredirect=1#comment99990291_56661442))

I'll keep an eye on the bug report for a fix

rok-cesnovar:
Is there an alternative for these cases except waiting for them to fix the bug.

SteveBronder:
lol thanks for checking it out! I'm not sure whether this is blocking or non-blocking. Sure it means we can't use this for everything, but Eigen folks seem to be checking it out and as long as we keep an eye on it I'm pretty sure there's a good few other places in our code where the above is not an issue.

I looked over their bugzilla and don't see it over there yet

rok-cesnovar:
If there is only a few cases where we cant use it, I definitely dont see this as being blocking.

andrjohns:
I'd view this as non-blocking, since the autodiff matrix functions generally pass by constant reference. It does block the usage of ```SolveInPlace()``` however

andrjohns:
>I looked over their bugzilla and don't see it over there yet

I'm waiting on the devs to register a bugzilla account for me (they don't allow self-registration) so that I can properly report it

andrjohns:
@SteveBronder (or anyone else) this is ready to have another look over, thanks!

roualdes:
Re https://github.com/stan-dev/stan/pull/2754: Since this PR necessitates `#define EIGEN_MATRIXBASE_PLUGIN` before any Eigen headers are included, should we clean up any stan-dev/math files that include Eigen headers without including `stan/math/prim/mat/fun/Eigen.hpp`?  For instance, see `stan/math/prim/mat/err/is_lower_triangular.hpp`.

Otherwise, I don't see any immediate nor potential problems between this and the flattening efforts.

bob-carpenter:
I think it'd be great to clean up the includes.  That could be done in a separate PR.  I'd be OK combining with this PR, but it's not my call.

SteveBronder:
Sorry for my delay I'll check this out during the week or this weekend.

I wonder if there's someway we can still let users use the Eigen plugin themselves but with some reasonable catch.

andrjohns:
All good Steve, happy for the help.

> I wonder if there's someway we can still let users use the Eigen plugin themselves but with some reasonable catch.

As in allow users to add other methods with ```#define EIGEN_MATRIXBASE_PLUGIN``` on top of these ones? I was thinking of having a 'meta-header' that includes each 'addon' file, for example:
```prim/mat/fun/Eigen.hpp```:
```
...

#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/addons_meta.hpp"

...
```

```prim/mat/addons_meta.hpp```:
```
#include <stan/math/prim/mat/matrix_addons.h>
```

**Comment by SteveBronder** [comment]
Word! I'm about halfway through the review, all looks good! Mostly need to read over the unit tests.

As per the Jenkins error maybe try merging with master? Some stuff for windows threading was merged recently-ish

File: stan/math/prim/mat/matrix_addons.h [hunk 6]
```diff
@@ -0,0 +1,174 @@
...
+ */
+template<class, class = void>
+struct is_fvar : std::false_type
+{ };
+template<class T>
+struct is_fvar<T, decltype((void)(T::d_))> : std::true_type
+{ };
+
+template<typename T>
+using rev_rtn_type = std::conditional_t<std::is_const<typename std::remove_reference<T>::type>::value,
```
SteveBronder:
```suggestion
using rev_rtn_type = std::conditional_t<std::is_const<std::remove_reference_t<T>>::value,
```

Maybe put a todo here so that when we move to C++17 we can use `std::is_const_v<>`

stan/math/prim/mat/matrix_addons.h
```diff
@@ -0,0 +1,174 @@
...
+                                         const double&,
+                                         double&>;
+
+template<typename T>
+using vi_rtn_type = std::conditional_t<std::is_const<typename std::remove_reference<T>::type>::value,
+                                          const decltype(T::vi_)&,
+                                          decltype(T::vi_)&>;
+
+template<typename T>
+using fwd_rtn_type = std::conditional_t<std::is_const<typename std::remove_reference<T>::type>::value,
```
SteveBronder: How do you feel about just using `_t` since it returns a type?

```suggestion
using forward_return_t = std::conditional_t<std::is_const<typename std::remove_reference<T>::type>::value,
```

SteveBronder: Also I'm a fan of longer names

stan/math/prim/mat/matrix_addons.h
```diff
@@ -0,0 +1,174 @@
...
+ * Reimplements is_fvar without requiring external math headers
+ *
+ * decltype((void)(T::d_)) is a pre C++17 replacement for std::void_t
+ */
+template<class, class = void>
+struct is_fvar : std::false_type
+{ };
+template<class T>
+struct is_fvar<T, decltype((void)(T::d_))> : std::true_type
+{ };
```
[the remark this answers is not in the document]
andrjohns: That stackoverflow post was a real bacon-saver when it came to figuring out this templating, would be more than a little lost without that website some times

SteveBronder:
Nice ty! As per the 
```
#include <stan/math/prim/mat/matrix_addons.h>
```

Would we just have a note telling people to include the matrix_addons.h file in their add-on? If so that seems like a nice solution. @seantalts @syclik does that make sense to yinz? If we are not concerned about that then I'm cool with approving this now

SteveBronder:
Also neat article came across me the other day related to this PR

https://www.fluentcpp.com/2017/06/02/write-template-metaprogramming-expressively/

andrjohns:
That's a great blog as a whole, thanks!

andrjohns:
@serban-nicusor-toptal Got that odd windows threading failure here as well

seantalts:
develop is also broken on it. The runTests.py `-f` filtering doesn't work currently on Windows, it seems.

SteveBronder:
😅 srry for the time bomb

Did the python version change or something? That's kind of surprising. If you can get me access to the windows machine I'll look into it

seantalts:
I looked into it further - it seems to work on the Windows machine we have,
but isn't working on the AMI (but that could be for a variety of reasons).
I'll take the EC2 Windows machines offline and let @serban-nicusor-toptal
look into it and test it before adding them back.

SteveBronder:
Serb and I chatted he is going to set me up with an instance tmrw so I can take a look

andrjohns:
@SteveBronder looks like the windows testing issues are resolved, can you have a final look over things?

syclik:
> Nice ty! As per the
> 
> ```
> #include <stan/math/prim/mat/matrix_addons.h>
> ```
> 
> Would we just have a note telling people to include the matrix_addons.h file in their add-on? If so that seems like a nice solution. @seantalts @syclik does that make sense to yinz? If we are not concerned about that then I'm cool with approving this now

Will that work? I guess it matters when the `#define EIGEN_MATRIXBASE_PLUGIN` is defined. 

1) If someone else defines `EIGEN_MATRIXBASE_PLUGIN` and Eigen's includes are called, then ours won't get included unless they add our include to their includes. So... what you suggest works.
2) if someone else defines `EIGEN_MATRIXBASE_PLUGIN` and our header is called, then theirs won't get included at all. And what you suggest doesn't work. (I think.)

syclik:
Btw, that has no impact on the review. I wasn't able to get a few hours to check to see if we could do it with views, so my potential technical evaluation (in a previous comment) has timed out.

@andrjohns: THANK YOU!
@SteveBronder: action is on you to approve.

SteveBronder:
Yep! @andrjohns can you on do what you are suggesting above with `prim/mat/addons_meta.hpp`. I think that's a nice solution! A gotcha but one that is pretty explicit and solvable as long as we doc it somewhere (and should probs include it in the release notes)

andrjohns:
Alrighty, I've made a new folder ```prim/mat/eigen_plugins``` for the plugin files, with the meta-header ```prim/mat/eigen_plugins.h``` used to include the new plugin methods.

@SteveBronder can you check the doc on the ```prim/mat/eigen_plugins.h``` header and let me know if it's clear enough?

@syclik this approach adds another folder to the ```prim``` hierarchy, is that alright with you? Also let me know if you want any of the names changed

rok-cesnovar:
This may be a stupid question - feel free to ignore me: 

I was just wondering if it is possible to also add `.val()` to `Matrix<double>`? It can be a solution that gets compiled away like our current `value_of_rec(Matrix<double>)` solution. 

I was wondering this for cases where we would want to do x.val() and x could be `Matrix<double>` or `Matrix<var>`. This might potentially simply some of the codebase if it is possible to do.

SteveBronder:
> I was just wondering if it is possible to also add .val() to Matrix<double>? It can be a solution that gets compiled away like our current value_of_rec(Matrix<double>) solution.

I don't see why it wouldn't be possible. But I'm leaving this to Andrew. Also approving this PR!

> I was wondering this for cases where we would want to do x.val() and x could be Matrix<double> or Matrix<var>. This might potentially simply some of the codebase if it is possible to do.

Is there somewhere in particular this stood out to you in the library?

rok-cesnovar:
This definitely wasnt a question for this PR to address, it just kind of fits in the discussion. I think it comes up in the GLMs at least. 

Lets get this merged first. Thanks @andrjohns for this!

andrjohns:
> if it is possible to also add .val() to Matrix? 

Yep that's very simple to add, just pushed it through

rok-cesnovar:
Thanks again!

bob-carpenter:
I'd rather not duplicate existing functions as methods.

Is the suffix on value_of_rec () because it goes from fvar<var> to var to double and the basic value_of() just goes from fvar<var> to var?

stan/math/prim/mat/fun/Eigen.hpp
```diff
@@ -1,7 +1,7 @@
 #ifndef STAN_MATH_PRIM_MAT_FUN_EIGEN_HPP
 #define STAN_MATH_PRIM_MAT_FUN_EIGEN_HPP
 
-#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/eigen_plugins.hpp"
+#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/eigen_plugins.h"
```
SteveBronder: Something I just thought of. Here should we add a check to see if `EIGEN_MATRIXBASE_PLUGIN` is already defined and if so throw a warning to the user? I think that would help our users who add their own extensions know what to do when they want their own plugins

andrjohns: That's a good idea, what do you think of:
```
#if defined EIGEN_MATRIXBASE_PLUGIN
#error Please use the eigen_plugins.h header to add your own extensions
#else
#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/eigen_plugins.h"
#endif
```

andrjohns:
@bob-carpenter the ```.val()``` method was always (essentially) a re-implementation of ```value_of()``` as a method, so what Rok suggested just made the functionality of both identical - ```.val()``` only worked with autodiff types, whereas ```value_of()``` works with both autodiff types and doubles

stan/math/prim/mat/fun/Eigen.hpp
SteveBronder: I'd rather not have someone add a file to the library. Here is what I'm thinking

Add this to our plugin at the bottom.

```cpp
#define EIGEN_STAN_MATRIXBASE_PLUGIN 
```

Then we have the check look like

```cpp
#ifdef EIGEN_MATRIXBASE_PLUGIN
  #ifndef EIGEN_STAN_MATRIXBASE_PLUGIN
  #error "Stan uses Eigen's EIGEN_MATRIXBASE_PLUGIN macro. To use your own plugin add 
    the eigen_plugin.h file to your plugin. 
  #endif
#else 
#define EIGEN_MATRIXBASE_PLUGIN "stan/math/prim/mat/eigen_plugins.h"
#endif
```

Then users can go about writing their own plugin and only need to include the `eigen_plugins.h` file in their plugin. 

We can also then move the stuff in `autodiff_variables.h` back to `eigen_plugins.h`

syclik:
@andrsjohn, thank you for asking. Let's not add a top-level directory if we can avoid it. We should add things there very, very sparingly. If you think it really needs to go there (which might be the right thing), can you write up what you're proposing, the motivation, and the solution as a stand-alone post / doc so we can review it?

stan/math/prim/mat/fun/Eigen.hpp
andrjohns: Oh I see what you mean, that would be a bit tidier. Good call

andrjohns:
@syclik as per Steve's comment [here](https://github.com/stan-dev/math/pull/1216#discussion_r301892223), it won't be necessary to add another directory

andrjohns:
Tests have passed, it's finally done! Can someone merge this?

SteveBronder:
Oh whoops! Sorry I was going to let you press the button!

SteveBronder:
imo there's something satisfying about pressing the button and I didn't want to take that from you

andrjohns:
Haha you're all good, just as satisfying to have this off the to-do list

PR #1224, opened 2019-05-04
[Jenkins] Move windows tests to later in the build

seantalts:
Windows is a big bottleneck right now. with this PR, they'll still always run on every PR but this way they won't run until after the unit tests have already passed in some other configuration. As is, unit tests could fail on Linux and Mac and we'd still queue up and take up an hour of time on Windows even though we tend not to learn something new with Windows tests. Seems like a good general rule to run uncorrelated tests at the same time and if those pass run the correlated tests to be more sure.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -172,15 +172,6 @@ pipeline {
                     }
                     post { always { retry(3) { deleteDir() } } }
                 }
-                stage('Windows Headers & Unit') {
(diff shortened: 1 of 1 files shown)


syclik:
That sounds awesome!

syclik:
(thanks @rok-cesnovar! I didn't mean to double-approve it, but I did.)

Issue #1232, opened 2019-05-12
Variadic is_nan function


andrjohns:
## Description
Several functions call ```is_nan``` multiple times using the ```||``` operator, such as in ```prim/scal/fun/grad_reg_inc_gamma```:
```
  if (is_nan(a) || is_nan(z) || is_nan(g) || is_nan(dig))
    return std::numeric_limits<TP>::quiet_NaN();
```
This could be replaced with the introduction of a variadic version of ```is_nan``` (like that used for ```size_zero```):
```
  if (is_nan(a, z, g, dig))
    return std::numeric_limits<TP>::quiet_NaN();
```

#### Current Version:
v2.19.1

syclik:
This would be great for clarity and usability! Thanks!

bob-carpenter:
I don't find this clear without changing the name to is_any_nan or is_nan_any to make it clear it's an OR not an AND putting the multiple arguments together.

andrjohns:
> I don't find this clear without changing the name to is_any_nan or is_nan_any to make it clear it's an OR not an AND putting the multiple arguments together.

I see what you mean. That then raises the question of whether we have:

-  Two versions of ```is_nan```: single inputs call ```is_nan(x)``` and multiple inputs call ```is_nan_any(x,y)```
-  One version and update the naming for all functions  :  single inputs call ```is_nan_any(x)``` and multiple inputs call ```is_nan_any(x,y)```

@wds15 The Eigen plugin PR is waiting on a PR in [stan](https://github.com/stan-dev/stan/pull/2754), so there's not much more I can do for that atm

betanalpha:
Opinion: I think that “is_any_nan” sounds much nice grammatically.

Won’t variadic implementations just dispatch single-argument calls to
the base case?  In which case we could have a single function 
is_any_nan that holds for any size, no?

andrjohns:
> Won’t variadic implementations just dispatch single-argument calls to the base case? In which case we could have a single function is_any_nan that holds for any size, no?

Yep, that's what I was suggesting as the second option above

betanalpha:
My bad —for some reason I mistakenly thought you were suggesting 
an overload for the multiple-argument and single-argument.  Poor 
reading comprehension on my part.

I definitely support a single variadic “is_any_nan” implementation
which has the smallest maintenance burden and most clear naming.

andrjohns:
No worries, equally poor clarity on my part!

[...]

PR #1233, opened 2019-05-12
Issue 1232 - variadic is_nan function

andrjohns:
## Summary
As initially described in #1232 - Several functions call ```is_nan``` multiple times using the ```||``` operator, such as in ```prim/scal/fun/grad_reg_inc_gamma```:
```
  if (is_nan(a) || is_nan(z) || is_nan(g) || is_nan(dig))
    return std::numeric_limits<TP>::quiet_NaN();
```
This PR adds a variadic extension to the ```is_nan``` function so that multiple inputs can be passed in a single call
```
  if (is_nan(a, z, g, dig))
    return std::numeric_limits<TP>::quiet_NaN();
```

## Tests
Added tests for calling ```is_nan``` with multiple inputs

## Side Effects
N/A

## Checklist

- [X] Math issue #1232 

- [X] Copyright holder: Andrew Johnson


stan/math/prim/scal/fun/is_nan.hpp
```diff
@@ -14,8 +15,24 @@ namespace math {
...
  * @return <code>true</code> if the value is NaN.
  */
-inline bool is_nan(double x) { return std::isnan(x); }
+template <typename T>
+inline bool is_nan(T x) {
+  return std::isnan(x);
+}
 
+/**
+ * Returns true if the input is NaN and false otherwise.
```
syclik: Can you make this doc more precise? It should indicate that it returns `true` if any of the input is `NaN` (as opposed to all).

wds15:
I think what @bob-carpenter said is important: The function should become `is_nan_any` or something similar.

andrjohns:
@syclik I've made the changes from the issue discussion and the tests have passed, this is ready for re-review

stan/math/prim/scal/fun/is_any_nan.hpp
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Delegates to <code>stan::math::is_nan</code> so that
+ * appropriate specialisations can be loaded for autodiff
+ * types.
+ *
+ * @param x Value to test.
+ * @return <code>true</code> if the value is NaN.
+ */
+template <typename T>
+inline bool is_any_nan(T x) {
```
syclik: We should have caught this on the original PR that introduced `is_nan`. The arguments are passed by value here. They should be passed by reference for efficiency reasons. So... in the function signature: `is_any_nan(T& x)` instead of `is_any_nan(T x)`.

stan/math/prim/scal/fun/is_any_nan.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * Returns <code>true</code> if any input is NaN and false otherwise.
+ *
+ * Delegates to <code>std::isnan</code>.
+ *
+ * @param x first argument
+ * @param xs parameter pack of remaining arguments to forward to function
+ * @return <code>true</code> if any value is NaN
+ */
+template <typename T, typename... Ts>
+inline bool is_any_nan(T x, Ts... xs) {
```
syclik: Same here. This should be by reference.

stan/math/prim/scal/fun/is_nan.hpp
```diff
@@ -14,7 +14,10 @@ namespace math {
  * @param x Value to test.
  * @return <code>true</code> if the value is NaN.
  */
-inline bool is_nan(double x) { return std::isnan(x); }
+template <typename T>
+inline bool is_nan(T x) {
```
syclik: Hmm... this looks odd. In `develop`, we have 3 definitions of `is_nan`: prim/scal, rev/scal, fwd/scal. They are implemented using function overloading.

I think you either want to go the same route or you want to make this function the template function (as you have it) and then change the implementations in rev and fwd to be template specializations.

syclik: Does that make sense? If it's not clear what I'm talking about, we can discuss.

stan/math/prim/scal/fun/log_falling_factorial.hpp
```diff
@@ -54,7 +54,7 @@ namespace math {
 template <typename T1, typename T2>
 inline typename return_type<T1, T2>::type log_falling_factorial(const T1 x,
                                                                 const T2 n) {
-  if (is_nan(x) || is_nan(n))
+  if (is_any_nan(x, n))
```
syclik: [comment]
The changes look awesome! It does look like there's an impact on how we will write functions here on out.

andrjohns:
@serban-nicusor-toptal would you be able to restart the "Always-run tests part 1" for this pull? Looks like the failure was just randomness, as discussed in #1238

serban-nicusor-toptal:
@andrjohns Good morning!
I've restarted it, it's wating for a gpu machine. Keep me updated!

andrjohns:
Thanks for that!

andrjohns:
@serban-nicusor-toptal looks like the build is running twice for some reason (there's a run 8 and a run 9), not sure what happened there

rok-cesnovar:
I got the email and restarted it. Should have posted here sorry.

andrjohns:
No worries, I'm not about to complain for having too much help!

andrjohns:
@syclik This is ready for another look-over. I've updated ```is_any_nan``` to pass by reference and changed the rev/fwd versions of ```is_nan``` to template specialisations.

I also added in fwd & rev tests for ```is_any_nan```, just to be safe

stan/math/fwd/scal/fun/is_nan.hpp
```diff
@@ -2,21 +2,23 @@
...
  * Delegates to <code>is_nan</code>.
  *
  * @param x Value to test.
- * @return <code>1</code> if the value is NaN and <code>0</code> otherwise.
+ * @return <code>true</code> if the value is NaN and <code>false</code>
+ * otherwise.
  */
 template <typename T>
-inline int is_nan(const fvar<T>& x) {
+inline typename std::enable_if_t<is_fvar<T>::value, bool> is_nan(const T& x) {
```
bob-carpenter: What was the motivation for switching to `enable_if_t` from the simple pattern matching?  I think this signature should just be:

```
template <typename T>
inline bool is_nan(const fvar<T>& x);
```

stan/math/prim/scal/fun/is_any_nan.hpp
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Delegates to <code>stan::math::is_nan</code>.
+ *
+ * @param x first argument
+ * @param xs parameter pack of remaining arguments to forward to function
+ * @return <code>true</code> if any value is NaN
+ */
+template <typename T, typename... Ts>
+inline bool is_any_nan(const T& x, const Ts&... xs) {
+  return (is_any_nan(x) || is_any_nan(std::forward<const Ts>(xs)...));
```
bob-carpenter: These are both OK to be constant references so they'll be generic.  You don't need those outer parens---that's an R thing, not a C++ thing and they're just a distraction here.

stan/math/prim/scal/fun/is_nan.hpp
```diff
@@ -14,7 +15,11 @@ namespace math {
  * @param x Value to test.
  * @return <code>true</code> if the value is NaN.
  */
-inline bool is_nan(double x) { return std::isnan(x); }
+template <typename T>
+inline typename std::enable_if_t<std::is_arithmetic<T>::value, bool> is_nan(
```
bob-carpenter: Same comment here as in the issue for changing all of these.  At most, this should be

```
template <typename T>
inline typename std::enable_if_t<std::is_arithmetic<T>::value, bool> is_nan(T x) {
  return std::isnan(x);
}
```

so primitives get passed by value.

stan/math/fwd/scal/fun/is_nan.hpp
andrjohns: This was one was based on @syclik's review:
> Hmm... this looks odd. In develop, we have 3 definitions of is_nan: prim/scal, rev/scal, fwd/scal. They are implemented using function overloading.

> I think you either want to go the same route or you want to make this function the template function (as you have it) and then change the implementations in rev and fwd to be template specializations.

andrjohns:
@syclik This one is just waiting on your approval before it's ready to merge, thanks!

bob-carpenter:
I've never been clear on where we should use template specializations and where we should just use overloads.  Name resolution works by finding the best matching (template) function overload, then finding the best matching specialization.  So I see how it works differently, just not when it would make a difference in a real situation.

syclik:
@bob-carpenter: I'm with you. I sort of view them as exchangeable for this sort of application (although someone might be able to tell me why one would be preferred). What I was trying to get at with my comments was that we don't need the `enable_if` type of construction for this use case. It's just easier to read and follow when it's less complex.

andrjohns:
@syclik @bob-carpenter 
Should I just leave ```is_nan``` for autodiff types unchanged from develop? Seems like there's no preference either way, so I don't think there's much need to change it

bob-carpenter:
Fine by me, but it's Daniel's call.

drezap:
@syclik it's your call ^^^^^ (just trying to keep things moving. this is a good change).

rok-cesnovar:
@andrjohns when you have the time, please merge the latest develop in. There are conflicts due to the meta flattening refactor.

andrjohns:
@syclik Did you have a preference on the specification of ```is_nan``` for autodiff types (i.e. leave unchanged from develop or ```enable_if``` templating)?

syclik:
My preference is for simpler rather than more complex. Here, I'd prefer develop rather than using `enable_if`.

andrjohns:
Great! Changes have been reverted

andrjohns:
@serban-nicusor-toptal If the windows threading errors aren't an issue anymore can you restart the testing for this one? Thanks

PR #1234, opened 2019-05-12
Feature/issue 1122 make a parameter pack is_constant trait and consolidate all *_constant_* traits

rok-cesnovar:
## Summary

This PR refactors all the trait that deal with types being constant or not. 

On develop we currently have the following traits: 
- `is_constant`
- `is_constant_struct`
- `is_nonconstant_struct`
- `contains_nonconstant_struct`

Each of them have their own set of tests.

This PR makes a parameter pack version of  the `is_constant` trait -> `is_constant_all`. This trait is true if all types are constant, and false otherwise.

With this trait we can consolidate all existing "constant" traits by: 
- replacing the use of `is_constant` with `is_constant_all` (with only one type parameter we could actually still use `is_constant`, but I think its better to just use one)
- replacing `is_constant_struct<A> && is_constrant_struct<B>...` with `is_constant_all<A, B, ...>`
- replacing `contains_nonconstant_struct<A, B, ... >` with `!is_constant_all<A, B, ...>`
- removing `is_nonconstant_struct`

## Tests

- renamed the `is_constant_struct` tests to `is_constant_all` 
- extended the tests with more type parameters
- removed the `contains_nonconstant_struct` (they were testing the same cases as we are now testing in the is_constant_all tests, only with !)
- removed the `is_nonconstant_struct` tests

## Side Effects

/

## Checklist

- [x] Math issue #1122


syclik:
@bob-carpenter, would you mind looking at this PR?

bob-carpenter:
I can review it.

stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp
```diff
@@ -117,8 +117,7 @@ typename return_type<T_x, T_alpha, T_beta>::type bernoulli_logit_glm_lpmf(
 
   // Compute the necessary derivatives.
   operands_and_partials<T_x, T_alpha, T_beta> ops_partials(x, alpha, beta);
-  if (!is_constant_struct<T_beta>::value || !is_constant_struct<T_x>::value
-      || !is_constant_struct<T_alpha>::value) {
+  if (!is_constant_struct<T_beta, T_x, T_alpha>::value) {
```
bob-carpenter: [note---no action]
All of the use here seems to be `!is_constant_struct`.  I can't think of anything shorter that makes sense---maybe `non_const`.

stan/math/prim/scal/meta/is_constant_struct.hpp
```diff
@@ -2,18 +2,27 @@
...
-struct is_constant_struct {
+struct is_constant_struct_helper {
   enum { value = is_constant<T>::value };
 };
+/**
+ * Metaprogram to determine if a type has a base scalar
+ * type that can be assigned to type double.
+ * @tparam T Types to test
+ */
+template <typename... T>
```
bob-carpenter: This is super cool.  I've never seen this kind of functional definition before.

stan/math/prim/scal/meta/is_constant_struct.hpp
```diff
@@ -2,18 +2,27 @@
...
+#include <stan/math/prim/scal/meta/conjunction.hpp>
 
 namespace stan {
-
 /**
- * Metaprogram to determine if a type has a base scalar
- * type that can be assigned to type double.
- * @tparam T Types to test
+ * Defines a public enum named value and sets it to true(1)
+ * if the provided type is constant, false(0) otherwise.
```
bob-carpenter: The doc should explain that "constant" means a primitive type here.  

Being assignable to `double`, as used below, isn't quite right as that could be enabled by an external `operator=` definition.

For this program, I'd suggest:

```
Metaprogram defining an enum `value` whose value is `true` if the
type parameter is a primitive type and `false` otherwise.
```

stan/math/prim/scal/meta/is_constant_struct.hpp
```diff
@@ -2,18 +2,27 @@
...
+ * @tparam type to test
  */
 template <typename T>
-struct is_constant_struct {
+struct is_constant_struct_helper {
   enum { value = is_constant<T>::value };
 };
+/**
+ * Metaprogram to determine if a type has a base scalar
+ * type that can be assigned to type double.
```
bob-carpenter: How about:

```
Metaprogram defining an enum `value` which is `true` if all of 
the type parameters are constant (i.e., primtive types) 
and `false` otherwise.
```

Or you can point back to the doc for the helper function for the notion of constant.  It's nice to have only one source of doc, but this is so short I'd just include it redundantly here.

test/unit/math/prim/arr/meta/is_constant_struct_test.cpp
```diff
@@ -5,7 +5,14 @@
 TEST(MetaTraits, isConstantStruct) {
   using stan::is_constant_struct;
   using std::vector;
-  EXPECT_TRUE(is_constant_struct<vector<double> >::value);
-  EXPECT_TRUE(is_constant_struct<vector<vector<double> > >::value);
-  EXPECT_TRUE(is_constant_struct<vector<vector<vector<double> > > >::value);
+  EXPECT_TRUE(is_constant_struct<vector<double>>::value);
```
bob-carpenter: [optional]
I sometimes define helpers, like

```
<template typename... Ts>
void expect_is_const() {
  EXPECT_TRUE(is_constant_struct<Ts>::value);
}
```

and then the tests simplify to

```
expect_const<vector<double>>();
expect_const<vector<vector<double>>>();
...
```

test/unit/math/prim/arr/meta/is_constant_struct_test.cpp
```diff
@@ -5,7 +5,14 @@
...
+  EXPECT_TRUE(is_constant_struct<vector<vector<double>>>::value);
+  EXPECT_TRUE(is_constant_struct<vector<vector<vector<double>>>>::value);
+  bool temp = is_constant_struct<vector<double>, vector<double>,
+                                 vector<double>>::value;
```
bob-carpenter: Shouldn't we also have a test for

I'd want to see tests for the two edge cases, one where there's no type parameters and one where there's no structure.

```
is_constant_struct<>
is_constant_struct<double>
```

The first should be `true` as that's the unit for conjunction (or all/universal quantification).  

I'm not sure about the second---is this called `struct` because the `<double>` version should fail?

test/unit/math/prim/mat/meta/is_constant_struct_test.cpp
```diff
@@ -28,4 +28,10 @@ TEST(MetaTraits, isConstantStruct) {
   EXPECT_TRUE(is_constant_struct<const_v1>::value);
   EXPECT_TRUE(is_constant_struct<const_v2>::value);
   EXPECT_TRUE(is_constant_struct<const_v3>::value);
+  bool temp = is_constant_struct<const_t1, const_t2>::value;
```
bob-carpenter: I'd rather not see temporaries used like this and redefined. I'd suggest one of
1. just defining the tests directly,
2. using a new variable per test, or
3. define a helper test function.

rok-cesnovar: Unfortunately `EXPECT_TRUE(is_constant_struct<const_t1, const_t2>::value);` doesnt compile so I had to resort to this "hack". I will try (3) if I can find an easy solution, or will go with 2.

test/unit/math/prim/arr/meta/is_constant_struct_test.cpp
rok-cesnovar: `is_constant_struct<double>` is true. This one was already tested in the scal/meta along with `is_constant_struct<float>` and such. There isnt really a need to have `is_constant` and `is_constant_struct` separate. Will try and merged the two in this PR.

`is_constant_struct<>` is true, will add to the scal tests.

rok-cesnovar:
@bob-carpenter based on your comments I started thinking that we could maybe clean all this `is_constant` related traits we currently have.
 
Currently on `develop` we have:
- `is_constant`
- `is_constant_struct` (the same as is_constant as you note in the comments)
- `contains_nonconstant_struct` (essentially a parameter packed version of !is_constant_struct<>)
- `is_nonconstant_struct` (used only in `contains_nonconstant_struct`)

I propose a few options that would clear this up. Please comment if you like any of them or should leave this as is:
A) 
- make the `is_constant_all<>` trait that would replace `is_constant` and `is_constant_struct`
- replace `contains_nonconstant_struct` with `!is_constant_all<>`
- `is_nonconstant_struct` is removed
B)
- make the `is_constant_all<>` trait that would replace `is_constant` and `is_constant_struct`
- make the `is_nonconstant_any` trait that would replace  `contains_nonconstant_struct` and all the `!is_constant_struct<>` (80%+ `is_constant_struct` are used with `!`).
- `is_nonconstant_struct` is removed
C)
same a B) but the `is_constant_any` uses the current name `contains_nonconstant_struct`. Also replace `!is_constant_struct<>` with it.

My preference would be A, with B the backup if you prefer `is_nonconstant_any<>` over `!is_constant_all<>`.

bob-carpenter:
I like the plan to consolidate all calls into is_constant_all with ! for negation.

rok-cesnovar:
I updated the top-level comment in the PR to reflect that we are now consolidating all these traits. This has now also become a significantly larger PR. However, most of the changes are from `is_constant_struct` to `is_constant_all` and `contains_nonconstant_struct` to `is_constant_all`.

Non-find-and-replacing changes are in the `prim/scal/meta/is_constant`, `prim/mat/meta/is`_constant and `prim/arr/meta/is_constant` where I fixed the comments issue you raised in the review.

bob-carpenter:
Please ping me when this is ready to review again. Thanks.

rok-cesnovar:
@bob-carpenter It is ready. I had a few issues with leftover tests, its fixed now.

Issue #1241, opened 2019-05-15
Implement softmax regression GLM


t4c1:
## Description
Right now someone that wants to do softmax regression has to write something like this or equivalent in stan language:

```
var softmax_regression(const Matrix<int, Dynamic, 1>& y,
                                  const Matrix<double, Dynamic, Dynamic>& x,
                                  const Matrix<var, 1, Dynamic>& alpha,
                                  const Matrix<var, Dynamic, Dynamic>& beta) {
  Eigen::Matrix<T_return, Dynamic, Dynamic> tmp = (x * beta).array().rowwise() + alpha.array();
  var lpmf = 0;
  const size_t N_instances = x.rows();
  for (int i = 0; i < N_instances; i++) {
    lpmf += categorical_logit_lpmf(y[i], tmp.row(i).transpose().eval());
  }
  return lpmf;
}
```

This is a bit cumbersome and quite slow. Like existing GLMs this one should use analitically derived gradients to be significantly faster. GPU can be used for further speedup.

## Example
This is how the new GLM shall be used:
```
Matrix<int, Dynamic, 1> y = ...
Matrix<double, Dynamic, Dynamic> x = ...
Matrix<var, 1, Dynamic> alpha = ...
Matrix<var, Dynamic, Dynamic> beta  = ...

var lpmf = categorical_logit_glm_lpmf(y, x, alpha, beta);
```

## Expected Output
Should be equivalent to the function in description section.

#### Current Version:
v2.19.1

bob-carpenter:
This looks like it's using `K` coefficient vectors for a `K`-way classification.  That's not identified other than through the prior.  So the softmax regression defined above is just the non-identified likelihood.

If `categorical_logit` were vectorized like our other distributions and allowed a matrix as the parameter, the function you're trying to write would be a one-liner:

```
categorical_logit_lpmf(y, x * beta);
```

I think it's still worth building this to be in line with everything else and for the performance gains.

PR #1242, opened 2019-05-15
Implement softmax regression (categorical_logit_glm_lpmf)

t4c1:
## Summary

Adds CPU implementation of softmax regression. Since the distribution that is used in this GLM is already implemented under the name `categorical_logit_lpmf` I named the GLM `categorical_logit_glm_lpmf`.

## Tests
New tests are in test/unit/math/rev/mat/prob/categorical_logit_glm_lpmf_test.cpp . 

## Side Effects
None.

## Checklist

- [x] Math issue #1241 


t4c1:
I measured the speedup compared the version that uses autodiff instead of analytic derivatives. Benchmark code is [here](https://gist.github.com/t4c1/e91d2befa8eca5fe7b683fbd8e2afbd5). Results are in the graphs below.

glm_speedup

bob-carpenter:
There's an issue with this GLM in terms of whether you use `K` or `K - 1` coefficient vectors.  

The traditional approach in stats is to use `K - 1` coefficient vectors because they're not identified otherwise.  With Bayesian models, you can identify them with the prior.  We've gone around and round on this over time, with the advantage of `K` coefficient vectors being symmetry and `K - 1` being identifiability.

Which one are you using here, and if `K`, how are you dealing with the non-identifiability?

 If you use `K - 1`, then the model's identified in the traditional sense (assuming the data's not separable).

t4c1:
Right now it uses `K` coefficient vectors and leaves it to the user so specify appropriate prior. If `K-1` vectors is preferable, I can easily change that.

bob-carpenter:
I almost think we should implement both, but it'd be super confusing having two.  I think if we only have one, it should probably be the `K`-based one.  I wonder what other people think.  We haven't done a lot of analysis on efficiency.  Lately, we've been moving more toward softer identifications with priors on groups of coefficients (like `sum(alpha) ~ normal(0, 0.001)`), but I don't know how to do that here.

estrumbelj:
Hey, I'd suggest just k. With k-1 we take away some freedom from the user (like Bob said, some might prefer "soft" identifiability constraint via prior) and with k-1 you have multiple possible interpretations what they mean. I'd go as far as to say that k-1 would be a bad idea.

bob-carpenter:
> with k-1 you have multiple possible interpretations what they mean

What the coefficients mean?  Don't they only have meaning in a relative sense because of the non-identifiability?   If you fix one of the coefficient vectors to zero, the other ones are all interpreted relative to that value.

estrumbelj:
> 
> 
> > with k-1 you have multiple possible interpretations what they mean
> 
> What the coefficients mean? Don't they only have meaning in a relative sense because of the non-identifiability? If you fix one of the coefficient vectors to zero, the other ones are all interpreted relative to that value.

What I wanted to say was that with k-1 you have to know which one the developer picked as the "-1". At least the 1st one and the k-th one are possible. That is, I have to read the documentation. :) And then I need to reorder my categories to put the one I want as the reference in the correct spot.

betanalpha:
Just as a counterpoint — I’ve been running away from non-generative
“soft” prior constraints as quickly as I can, especially for categorical 
models like this.  These constraints often induce nasty geometries.
Moreover assigning one category as an anchor makes the whole thing 
more interpretable and facilities prior specification of the latent parameters. 

In that case (K - 1) with a well-defined convention of where the anchor 
class falls would be my vote.

bob-carpenter:
We discussed this issue in the Stan meeting today (@t4c1 --- we can send you the link if you'd like to attend---it's 11 am NY time on Thursdays).  The general consensus was that it should be coded with `K - 1` categories.  There's no binding decision from our meetings, but @andrewgelman and @bgoodri were both there and preferred `K - 1`.   We haven't heard from a couple of our regression/modeling experts, @dpsimpson and @avehtari.  

Andrew suggested you might have a motivating example for using `K` and identifying with a prior instead of `K - 1`.  If so, please share.

t4c1:
I don't have such an example, but @estrumbelj might have one.

estrumbelj:
> 
> 
> We discussed this issue in the Stan meeting today (@t4c1 --- we can send you the link if you'd like to attend---it's 11 am NY time on Thursdays). The general consensus was that it should be coded with `K - 1` categories. There's no binding decision from our meetings, but @andrewgelman and @bgoodri were both there and preferred `K - 1`. We haven't heard from a couple of our regression/modeling experts, @dpsimpson and @avehtari.
> 
> Andrew suggested you might have a motivating example for using `K` and identifying with a prior instead of `K - 1`. If so, please share.

What were the arguments for "K - 1"? It just doesn't make any sense to me.

By picking "K - 1" we are literally saying that this is the only possible way of identifying the model AND that the model must be identified.

On the other hand, going with "K" does not prevent the user from setting one row of Beta to zero. It is exactly what users have to do now with the categorical() and categorical_logit() when doing multinomial regression. And with "K" I have the additional freedom of choosing which category I want to use as the reference category.

Even the Stan manual categorical_logit example starts with a prior and then mentions k-1 as an alternative :) : https://mc-stan.org/docs/2_19/stan-users-guide/multi-logit-section.html

estrumbelj:
This GLMs raised a few specific and general questions on how GLMs should be implemented.

**1. Should we implement it with K categories or K - 1 categories, enforcing that the K-th category as the reference category and identifying the model.**

There seems to be a general consensus that in almost all cases you'd want to use a reference category. But there also exist cases where you (I) wouldn't want to. The shortest example I can come up with is when I'd want to regularize the coefficients for the covariates in my categorical logit regression (I might still use a reference for the intercepts). If I only have K-1 available, I'm forced to choose a reference category, breaking symmetry and making my results dependent on the category I chose.

My view is that if there is a possibility that K will be used, that option should be available. An additional argument for K is that you are still able to set a reference category by setting the corresponding column in beta. And it shouldn't mislead the Stan user into doing something silly, because users already have to "manually" set a reference category when doing categorical regression with categorical_logit().

On the other hand, going with K - 1 is strictly less flexible.

**2. Separate intercept or not or both? **

categorical_logit_glm_lpmf(y | x, alpha, beta)    VS   categorical_logit_glm_lpmf(y | x, beta)

I think having them separate is very useful. If nothing else, it makes it easier to regularize the coefficients, but not the intercept. The four GLMs that are currently in Stan are implemented only with separate intercept.

@syclik, you'll probably have something to add here.

**3. Signature/types/vectorization **

Based on what we discussed, I suggest the following:

categorical_logit_glm_lpmf(y | x, alpha, beta) =def= categorical_logit(y | alpha + x * beta);

- y can be a scalar OR array of size n,
- x can be a vector of size J OR n x J matrix (only if y is an array)
- alpha is a vector of size K
- beta is a J x K matrix.

J >= 0, K > 0, n > 0

The only broadcasting that should happen is when x is a vector and y is an array?

Anyway, someone needs to make a decision on these before we can move forward.

bob-carpenter:
That's a good summary of our previous discussions (over years) on K vs. K - 1.

For 1, The main reason to use K - 1 is identifiability.  If you go with K, the joint density is only identified through the priors.

For 3, thanks for defining the suggested signatures.

I have one modification I'd like to propose to avoid dependencies in the arguments.  Specifically, I'd like to make the legal types for y and x independent.  If x is a matrix and y a scalar, y gets broadcast;  if x is a vector and y an array, then x gets broadcast.

Given that you don't suggest vectorizing alpha and beta, does that mean we won't be trying to cover discrete choice models with this GLM?  I think it's best not to include them, but it was on the table earlier, so I thought I'd check.

estrumbelj:
> For 1, The main reason to use K - 1 is identifiability. If you go with K, the joint density is only identified through the priors.

This sounds like we have to give up indentifiability via reference category. Just to make it clear - if we go with K, no one will lose their right to identify their categorical logit with a reference category :)

> For 3, thanks for defining the suggested signatures. I have one modification I'd like to propose to avoid dependencies in the arguments. Specifically, I'd like to make the legal types for y and x independent. If x is a matrix and y a scalar, y gets broadcast; if x is a vector and y an array, then x gets broadcast.

It just seemed to me that broadcasting the same value of y is not something that would come up. But thinking about it, I agree with what you wrote. It is more elegant even if it never comes up.

> Given that you don't suggest vectorizing alpha and beta, does that mean we won't be trying to cover discrete choice models with this GLM? I think it's best not to include them, but it was on the table earlier, so I thought I'd check.

No, no discrete choice. I think it is a nice idea, but I don't think it is worth the extra implementation effort. And it would probably make it too complicated and confuse a lot of Stan users.

But it might be worth thinking about a "discrete_choice_logit_glm_lpdf" once we can do ragged structures.

t4c1:
I talked with @estrumbelj about argument types. The conclusion was a bit different than he wrote above.

Supported types should be:
y: `std::vector` or scalar - in case of single instance or broadcasting (`Eigen::Vector` and `Eigen::RowVector` are not needed)
x: `Eigen::Matrix` or `Eigen::RowVector` - in case of single instance or broadcasting (Special case for single attribute - `Eigen::Vector` is not needed)
alpha: `Eigen::Vector` (`std::vector` and `Eigen::RowVector` are not needed)
beta: `Eigen::Matrix` - no special cases for single attribute or class

If y is a scalar or x a `Eigen::RowVector` the values are broadcast - all instances use the same values.

@bob-carpenter Is this OK?

This is a bit tdifferent than other GLMs. They do not support broadcasting. Also y and alpha can be any of `std::vector`, `Eigen::Vector` or `Eigen::RowVector`. So I guess they should also be updated to match this?

bob-carpenter:
> On May 24, 2019, at 4:48 PM, Erik Štrumbelj <notifications@github.com> wrote:
> 
> For 1, The main reason to use K - 1 is identifiability. If you go with K, the joint density is only identified through the priors.
> 
> This sounds like we have to give up indentifiability via reference category. Just to make it clear - if we go with K, no one will lose their right to identify their categorical logit with a reference category :)

Right.  The K version is strictly more flexible in that it lets you implement the K - 1 version (albeit less efficiently).

bob-carpenter:
> @t4c1: I talked with @estrumbelj about argument types. The conclusion was a bit different than he wrote above.

> Supported types should be:
> y: std::vector or scalar - in case of single instance or broadcasting (Eigen::Vector and Eigen::RowVector are not needed)

This is what I'd like to see.

The Eigen versions aren't appropriate because we don't support integer matrices or vectors in Stan.

> x: Eigen::Matrix or Eigen::RowVector - in case of single instance or broadcasting (Special case for single attribute - Eigen::Vector is not needed)

We don't have any functions that require row vectors, but it's not out of the question.  What we've done in other cases here is allow a mishmash of linear algebra types with implicit transposition.  

So that'd allow Eigen::Vector (I'm following your notation for Eigen types).  I'm agnostic about whether we want to allow that.  We should ask @seantalts or @syclik.

> alpha: Eigen::Vector (std::vector and Eigen::RowVector are not needed)

We do have functions that require vectors, so this would be OK.  

> beta: Eigen::Matrix - no special cases for single attribute or class

This is OK.   I'm totally OK not allowing collections for alpha and beta in this function.

> If y is a scalar or x a Eigen::RowVector the values are broadcast - all instances use the same values.

Yes.

> @bob-carpenter Is this OK?
> This is a bit different than other GLMs. They do not support broadcasting.

Hmm.  I thought they did, but never checked closely.  I think it's OK to support broadcasting, or it could be simplified and `y` could be required to be an array and `x` required to be a matrix.  That'd also be OK.

> Also y and alpha can be any of std::vector, Eigen::Vector or Eigen::RowVector. So I guess they should also be updated to match this?

No, y can only be int or std::vector<int> as we don't allow Eigen::Matrix<int, R, C> in Stan programs.

I would rather *not* have alpha be a std::vector as it's going to be something involved in linear algebra.  For example, our multinormal doesn't allow the variate or the mean to be arrays---they have to be vector types.

syclik:
> Anyway, someone needs to make a decision on these before we can move forward.

@estrumbelj: what's the decision that needs to be made? Sorry, can you be more explicit about what is still unclear? Maybe it's because I'm not sure what function signatures you're suggesting. Maybe write the full set of things you want to support (in Stan types)?

----

I think the heart of the matter comes down to this:

From @bob-carpenter:
> For 1, The main reason to use K - 1 is identifiability. If you go with K, the joint density is only identified through the priors.

From @estrumbelj:
> On the other hand, going with K - 1 is strictly less flexible.

I really don't see why we're not implementing the `K` version in the Math library if going with `K - 1` is strictly less flexible. When implementing the lpmf, we don't really have to think through whether it's identified or not. We just need to think about how to compute the lpmf and whether that's stable.

Of course, if this is unusable as a function in the Stan language as-is, then there's no point in implementing it that way unless it simplifies something else.

---

estrumbelj:
> @estrumbelj: what's the decision that needs to be made? Sorry, can you be more explicit about what is still unclear? Maybe it's because I'm not sure what function signatures you're suggesting. Maybe write the full set of things you want to support (in Stan types)?

We want to add the categorical_logit_glm_lpdf with K categories and separate intercept (as described above) to the Math library and Stan language. The decision is basically if you (and @seantalts) approve.

I'm a bit lost on the signatures. @t4c1, can you please summarize again the signatures/Stan types that we want to implement, taking into account the latest posts by you and @bob-carpenter.

syclik:
Thanks. I guess I wasn't the only one confused on the actual signatures. I think it makes sense to let the Stan language drive the actual function that's exposed, but it wouldn't be terrible in Math if we had both implementations.

t4c1:
>We don't have any functions that require row vectors, but it's not out of the question. What we've done in other cases here is allow a mishmash of linear algebra types with implicit transposition.

@bob-carpenter Implicit transposition here could be confusing. If x is a matrix with 1 row that means we have 1 instance. If x is `Eigen::RowVector` it has the same meaning and same number of rows/cols. However if x is a matrix with 1 column that means 1 attribute. Implicit transposition would mean that `Eigen::Vector` (that has a same shape) would have a different meaning. So I am against implicit transposition.

@estrumbelj I have written suggested types [here](https://github.com/stan-dev/math/pull/1242#issuecomment-496027834). Mapping between c++ and stan types is:
`std::vector`: array,
`Eigen::Matrix`: matrix,
`Eigen::Vector`: vector,
`Eigen::RowVector`: rowvector

bob-carpenter:
I'm OK with that revised set of types and not doing implicit transposition.  I'm still not entirely clear on what's being proposed without a full signature list.  

I know these two are in play:

  A)  real categorical_logit_glm_lpmf(int[] y | matrix x, vector alpha, matrix beta)

  B)  real categorical_logit_glm_lpmf(int y | row_vector x, vector alpha, matrix beta)

QUESTION 1:  Should we allow the single data item case (B)?  The other GLMs don't.

Conditioned on the answer being yes, the next question is whether to allow broadcasting:

  C) real categorical_logit_glm_lpmf(int y | matrix x, vector alpha, matrix beta)

  D) real categorical_logit_glm_lpmf(int[] y | row_vector x, vector alpha, matrix beta)

QUESTION 2:  Should we allow broadcasting (C) and (D)?

QUESTION 3:  Should the dimensionality of alpha and beta be K and J x K, or (K - 1) and J x (K - 1).  I don't think anyone is still objecting to using K---it's strictly more expressive at the expense of some work.

t4c1:
Other GLMs do allow scalar y. I am not sure about row vector x. That case is not in tests, but I think it would likely work.

I assumed broadcasting is agreed upon, so I have almost completed the implementation. Though I don't see much value in it.

Another point about K/K-1 dilemma - the difference in computational effort is only significant if K is small.

avehtari:
> Other GLMs do allow scalar y. I am not sure about row vector x. That case is not in tests, but I think it would likely work.

No they do not allow scalar y nor row vector x, y has to be an array for lmpfs and a vector for lpdfs, and x has to be matrix. I have just last week tested this and had to write the following ugly things:
```
    log_lik[i] = bernoulli_logit_glm_lpmf(rep_array(y[i], 1) | to_matrix(x[i]), beta0, beta);
```
```
    log_lik[i] = normal_id_glm_lpdf(to_vector(rep_array(y[i], 1)) | to_matrix(x[i]), beta0, beta, sigma);
```

I've considered making an issue about this, but haven't had time and energy....

bob-carpenter:
OK---sounds like we should go the other way and allow broadcasting in the existing ones.

For future reference, I'd recommend 

{a}   instead of   rep_array(a, 1),

[a]   instead of   rep_row_vector(a, 1),

[a]'  instead of   rep_vector(a, 1),

and

[[a]]  instead of  to_matrix(a).

t4c1:
> No they do not allow scalar y nor row vector x
I was talking about what is implemented in c++. Likely not everything is exposed to stan language.

drezap:
https://github.com/stan-dev/math/pull/1242#issuecomment-496660471

Yeah, this is a good idea. Is it easy to write some casting meta program to do this?

Also, casting issue aside, @t4c1, is this ready to be reviewed?

Are we happy with K/K-1 issue? I can't really speak to this.

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
```diff
@@ -0,0 +1,206 @@
...
+ * @param alpha intercept vector (in log odds)
+ * @param beta weight matrix
+ * @return log probability or log sum of probabilities
+ * @throw std::domain_error x, beta or alpha is infinite or y is not within
+ * bounds
+ * @throw std::invalid_argument if container sizes mismatch.
+ */
+template <bool propto, typename T_y, typename T_x_scalar, int T_x_rows,
+          typename T_alpha_scalar, typename T_beta_scalar>
+typename return_type<T_x_scalar, T_alpha_scalar, T_beta_scalar>::type
```
drezap: This might be a lot to ask, but is there anyway we can make the templating more readable like @SteveBronder did in the gpu_cov_exp_quad?

Is it possible to write a meta program that is a general return type for all of our pmf/pdfs? We could just make this a default moving forward, reading templates were a huge issue for me starting to learn this library.

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
```diff
@@ -0,0 +1,206 @@
...
+                  * inv_sum_exp_lin;
+      // TODO(Tadej) maybe we can replace previous block with the following line
+      // when we have newer Eigen  ops_partials.edge1_.partials_ = beta_val(y -
+      // 1, all) - (exp_lin.matrix() * beta.transpose()).colwise() *
+      // inv_sum_exp_lin;
+    }
+  }
+  if (!is_constant_struct<T_alpha_scalar>::value
+      || !is_constant_struct<T_beta_scalar>::value) {
+    Array<T_partials_return, T_x_rows, Dynamic> neg_softmax_lin
```
drezap: For Array (which I believe is from Eigen), do we want to include prim/Eigen.hpp?

Any benefits of using an Array over matrix? (Array I guess has more dimensions than 2...)

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
```diff
@@ -0,0 +1,206 @@
...
+      // 1) += x_val.colwise.sum().transpose();
+
+      ops_partials.edge3_.partials_ = std::move(beta_derivative);
+    }
+  }
+  return ops_partials.build(logp);
+}
+
+template <typename T_y, typename T_x_scalar, int T_x_rows,
+          typename T_alpha_scalar, typename T_beta_scalar>
```
drezap: again, a `return_type_t_` could make this a lot more slick.

test/unit/math/rev/mat/prob/categorical_logit_glm_lpmf_test.cpp
```diff
@@ -0,0 +1,569 @@
...
+    y[i] = Matrix<int, Dynamic, 1>::Random(1)[0] % N_classes + 1;
+  }
+  MatrixXd x_double = MatrixXd::Random(N_instances, N_attributes);
+  Matrix<var, Dynamic, Dynamic> x1 = x_double, x2 = x_double;
+  MatrixXd beta_double = MatrixXd::Random(N_attributes, N_classes);
+  Matrix<var, Dynamic, Dynamic> beta1 = beta_double, beta2 = beta_double;
+  RowVectorXd alpha_double = RowVectorXd::Random(N_classes);
+  Matrix<var, Dynamic, 1> alpha1 = alpha_double, alpha2 = alpha_double;
+
+  var res1 = categorical_logit_glm_simple_lpmf<false>(y, x1, alpha1, beta1);
```
drezap: What's simple? the literal `softmax(inv_logit(foo))`, and you're testing against your new implementation?

test/unit/math/rev/mat/prob/categorical_logit_glm_lpmf_test.cpp
```diff
@@ -0,0 +1,569 @@
...
+  EXPECT_THROW(categorical_logit_glm_lpmf(y, x, alpha_val2, beta),
+               std::domain_error);
+  EXPECT_THROW(categorical_logit_glm_lpmf(y, x, alpha, beta_size1),
+               std::invalid_argument);
+  EXPECT_THROW(categorical_logit_glm_lpmf(y, x, alpha, beta_size2),
+               std::invalid_argument);
+  EXPECT_THROW(categorical_logit_glm_lpmf(y, x, alpha, beta_val1),
+               std::domain_error);
+  EXPECT_THROW(categorical_logit_glm_lpmf(y, x, alpha, beta_val2),
+               std::domain_error);
```
drezap: cool

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
t4c1: I can write something similar so that we have 
`typename return_type_t<T_x_scalar, T_alpha_scalar, T_beta_scalar>` 
instead of 
`typename return_type<T_x_scalar, T_alpha_scalar, T_beta_scalar>::type`

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
t4c1: `Array` and `Matrix` types both have 2 dimensions. They only differ in what multiplication does to them. For `Matrix` it is matrix multiplication and for `Array` it is element-wise multiplication. 

Which header do you mean? There is no Eigen.hpp in stan/math/prim. `Array` comes from Eigen/Core.

test/unit/math/rev/mat/prob/categorical_logit_glm_lpmf_test.cpp
t4c1: The function `categorical_logit_glm_simple_lpmf()` is implemented at start of the file. It is a simple implemtation of the new GLM - it uses autodiff instead of analytically derived gradients.

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
drezap: In prim/mat.hpp there's an Eigen.hpp, which includes core.

Not at all necessary, as you'd including stuff you're not using anyway.

stan/math/prim/mat/prob/categorical_logit_glm_lpmf.hpp
drezap: I'd say let's pass because it's not at all helping with readability/reducing lines of code.

PR #1245, opened 2019-05-18
Feature/faster ad tls v6

rok-cesnovar:
## Summary

This PR reapplies the changes of the PR faster TLS v5 which was reverted from develop after performance tests issues. See #1244

In short, this PR: 
- reverts the revert from #1244, reapplying the fix for threading on Windows
- reverts 2 commits in the TLSv5 branch, effectively goes back to using pointer access to the instance

First, lets see what the perf tests show now. If this wont be fine, we should try ifdefing the autodiffstackstorage.hpp with a more clean version of something like [this](https://gist.github.com/rok-cesnovar/07af3630e15b06d28208bc857cdae199)

## Tests

/

## Side Effects

The side effects to TLSv5 are kind of why we are doing this.

## Performance Results

| OS | compiler | threading | number of threads | schools-4 - develop | schools-4 - tls | schools-4 % improvement | warfarin - develop | warfarin - tls | warfarin % improvement | 
|-----|---------|-----------|-------------------|---------|--------------|---|---|------|---|
| linux | g++ 5.x.x  stdlibc++ |  no | n/a |  10.1534s | 10.1198s | 0.3% | 256.7028s | 238.0342s | 7.8% |
| linux | g++ 5.x.x  stdlibc++ |  yes | 1 |  11.7442s | 10.5511s | 11.3% | 321.1362s | 246.0260s | 30.5% |
| linux | g++ 5.x.x  stdlibc++ |  yes | 2 |  11.4333s | 10.2273s | 11.8% | 291.3092s | 257.3849s | 13.2% |
| linux | g++ 5.x.x  stdlibc++ |  yes | 4 |  11.2953s | 10.0267s | 12.7% | 204.3419s | 175.2958s | 16.6% |
| mac | clang++ 6.x.x  libc++ |  no | n/a |  10.0808s | 10.3644s | -2.7% | 281.9994s | 276.4570s | 2.0% |
| mac | clang++ 6.x.x  libc++ |  yes | 1 |  15.6746s | 10.3569s | 51.3% | 414.8224s | 291.6971s | 42.2% |
| mac | clang++ 6.x.x  libc++ |  yes | 2 |  14.9762s | 10.0897s | 48.4%. | 210.5288s | 156.2712s | 34.7% |
| mac | clang++ 6.x.x  libc++ |  yes | 4 |  14.8488s | 9.9837s | 48.7% | 128.4796s | 94.3915s | 36.1% |
| windows | g++ 4.9.3 |  no | n/a |  9.1471s | 9,1897s | -0.4% | 318.881s | 318,86s | -0.006% |
| windows | g++ 4.9.3 |  yes | 1 |  / | 24.8642s | / | / | 452.852s | / |
| windows | g++ 4.9.3 |  yes | 2 |  / | 23.0454s | / | / | 224.1456s | / |
| windows | g++ 4.9.3 |  yes | 4 | /  | 23,7629s | / | / | 151.7429s | / |


rok-cesnovar:
@seantalts will Jenkins run the entire performance tests suite on this PR?

seantalts:
Not until it’s merged - we have PRs just running on the stat comp
benchmarks suite since it’s shorter and more accepted as meaningful.

We’ll have to run it ourselves, but the scripts should make that relatively
easy other than finding the idle machines.

rok-cesnovar:
> We’ll have to run it ourselves

Just to clarify. This means: 
```
git clone --recursive https://github.com/stan-dev/performance-tests-cmdstan
cd performance-tests-cmdstan
./compare-git-hashes.sh example-models/bugs_examples/vol2/schools/ develop d013e55 false false
```
Or something else?

wds15:
I did run things and got on a Linux RHEL 7.5 box with gcc 6.3.1 these timings:

```
running things a bit shorter:
./example-models/bugs_examples/vol2/schools/schools-4 method=sample num_warmup=250 num_samples=250 data file=example-models/bugs_examples/vol2/schools/schools.data.R random seed=1234

develop-good.log:               267.42 seconds (Total) ## this is before TLSv5 merge
develop-bad.log:               277.86 seconds (Total) ## this is just after TLSv5 merge
develop-v6.log:               262.33 seconds (Total) ## this is for this PR
```

So it looks good to me!

seantalts:
Rok did that work? The math hash should go in the 3rd slot there:
```
./compare-git-hashes.sh example-models/bugs_examples/vol2/schools/ develop develop false d013e55
```

it's `compare-git-hashes.sh <test_dir> <cmdstan hash/branch of first run> <cmdstan hash of second run> <stan hash of 2nd run> <math hash of 2nd run>` where `false` just means use whatever the cmdstan hash has listed for that submodule.

rok-cesnovar:
Havent tried yet, just copied pasted from your post on discourse (and mixed up the arguments). Thanks for the info on the arguments. Going to run this on Windows & Ubuntu system.

seantalts:
Honestly I'm still worried we reverted the wrong PR, or at least that there's another PR out there that is also causing a dip in performance. I'll try to run some git bisect on that soon... Were there any other candidates we were worried about? I know it was suspicious the ODE one seemed to be the first one to cause the failure...

rok-cesnovar:
You mentioned the ODE, I am not aware of any other. I have only done the git bisect with the schools-4 model as that was the model mentioned. Doing git bisect with more models at a time seemed like a hassle. It seems that the amra model is also returing weird performance results right? At least if I am judging from #1233

seantalts:
You are, though I'm not sure why that PR is showing that issue... it doesn't have anything to do with the TLS stuff right?

rok-cesnovar:
> it doesn't have anything to do with the TLS stuff right?

Nope. At least as far as I know.

ode33:
Speculation: https://www.phoronix.com/scan.php?page=news_item&px=MDS-Zombieload-Initial-Impact

It looks like these fixes would have been pushed: https://support.apple.com/en-ca/HT210107

seantalts:
None of our Macs are on Mojave :sweat_smile:

SteveBronder:
My work mac is Mojave, if you send me a link to what you want tested I can do it tmrw

wds15:
I am a bit confused as to how this PR can move forward.

- Who is going to review it? If I were to do that then all is fine (I explain below)
- Why do we still need Mojave? I have Mojave, but why is it needed for what?
- The PR as it is right now addresses the performance regression which was found for the Linux+gcc+schools model constellation. For this setup I have shown above that the performance regression goes away with the code in this PR.

BTW, if I recall right then clang compilers speed up with this changes set a little bit for the single-core case if I am not mistaken... and one more BTW... the performance reports are great to have, but would be better to also have them say what compiler on what OS with what CPU it ran on. I am not saying to expand the tests, just adding that doc helps, I think.

syclik:
@wds15, I can review it. I agree with all your points below. They were the
same points I made when you originally submitted the PR.

Between the two of us, we can be the "honest broker" and collect the
results accurately and request the right things to be measured. We'll need
to look at: hardware, compiler version, os version, git hash, and relative
computational results.

One other thing we can do is try to simplify the test case so we can
observe these results without needing to go all the way through CmdStan. It
still wouldn't surprise me if there was an interaction with linking that we
didn't know about until now.

Does that sound like it'd work? Wed first define what needs to be measured.

wds15:
@syclik Thanks for reviewing. So what do you think about my suggestion about what to measure? I am pasting the results I posted from above here again:

I did run things and got on a Linux RHEL 7.5 box with gcc 6.3.1 these timings:

running things a bit shorter:
./example-models/bugs_examples/vol2/schools/schools-4 method=sample num_warmup=250 num_samples=250 data file=example-models/bugs_examples/vol2/schools/schools.data.R random seed=1234

```
develop-good.log:               267.42 seconds (Total) ## this is before TLSv5 merge
develop-bad.log:               277.86 seconds (Total) ## this is just after TLSv5 merge
develop-v6.log:               262.33 seconds (Total) ## this is for this PR
```

So it looks good to me!

EDIT: The CPU there is Intel(R) Xeon(R) CPU E5-2670 v2 @ 2.50GHz

So what I did is to run the schools model on this Linux+gcc system (this is exactly the problematic case)

1. before the TLSv5 merge... things are good
2. after the TLSv5 merge... things are bad
3. on the TLSv6 PR ... things are good

Do you think this is sufficient? Do we need more? I mean relative to what we already know, we are good, I think.

syclik:
@wds15: can you be explicit about what you think we need to measure? I'm trying to follow along and what I haven't seen is:
- what processes to run to measure time
- what configurations

I think you're trying to tell me, but it's hidden in comments and I want to make sure I understand what you're proposing. At the very least, it should cover all 3 OSes and have some reasonable test that flexes the autodiff stack.

syclik:
Btw, I'm expecting the key points from our discussions to be summarized and included at the PR description.

wds15:
I have marked in bold the info you are looking for and added information on how time was measured in the comment above.

I have only reported this exercise on Linux+gcc, since this is the only setup where the performance regression was seen. On macOS we did not have a problem as I recall, but I can verify if you think we need that.

If you think it is valuable to also do this on macOS - I can do that with a gcc version which I can grab from MacPorts (I would try to get a g++ 6ish which is similar to the Linux one). For macOS+clang I have posted on discourse that there never was a problem. For Windows I am not sure if I manage to do these test (if at all I would have to use a virtual box which adds another complication).

So Linux+gcc caused all of this trouble and this is why I restricted my benchmarks above to that combination. Which other combinations are needed from your perspective?

1. macOS+clang => I wasn't able to see the issue, that's documented on discourse for the moment
2. macOS+gcc8 => I did run it there and did not see an issue
3. Windows is tough for me to test and I could only use a virtual box

So I will add the Linux+gcc bit to the PR description and what other combination should we document? I can add macOS bits, but for Windows someone else would need to provide them (this time we can actually do the Windows stuff since threading is not used in the example).

(I know that the compiler versions I quote are not the "vanilla" ones, but that is a restriction due to the systems I use; not much I can do about that)

syclik:
Thanks for the info. 

> I have only reported this exercise on Linux+gcc, since this is the only setup where the performance regression was seen. On macOS we did not have a problem as I recall, but I can verify if you think we need that.

I absolutely think we need to reverify every time we change anything. We're not guaranteed that new changes maintain old behavior.

It looks like you're only reporting one single run. Is that the one benchmark we're going to use to evaluate this? Is that sufficient?

I think we only need to compare to current develop now (we should pick a hash) vs the latest version of this branch with that develop merged in (the exact git hash).

Here's what I think our table should look like:
| OS | compiler | threading | number of threads | benchmark - develop | benchmark - tls | % improvement | 
|-----|---------|-----------|-------------------|---------|--------------|---|
| linux | g++ 6.3.1 |  no | n/a |  267.42s | 262.33s | 1.9% |
| linux | g++ 6.3.1 |  yes | 1 |  ... | ... | ... |
| linux | g++ 6.3.1 |  yes | 2 |  ... | ... | ... |
| linux | g++ 6.3.1 |  yes | 4 |  ... | ... | ... |
| linux | g++ 5.x.x  stdlibc++ |  no | n/a |  ... | ... | ... |
| linux | g++ 5.x.x  stdlibc++ |  yes | 1 |  ... | ... | ... |
| linux | g++ 5.x.x  stdlibc++ |  yes | 2 |  ... | ... | ... |
| linux | g++ 5.x.x  stdlibc++ |  yes | 4 |  ... | ... | ... |
| mac | clang++ 6.x.x  libc++ |  no | n/a |  ... | ... | ... |
| mac | clang++ 6.x.x  libc++ |  yes | 1 |  ... | ... | ... |
| mac | clang++ 6.x.x  libc++ |  yes | 2 |  ... | ... | ... |
| mac | clang++ 6.x.x  libc++ |  yes | 4 |  ... | ... | ... |
| windows | g++ 4.9.3 |  no | n/a |  ... | ... | ... |
| windows | g++ 4.9.3 |  yes | 1 |  ... | ... | ... |
| windows | g++ 4.9.3 |  yes | 2 |  ... | ... | ... |
| windows | g++ 4.9.3 |  yes | 3 |  ... | ... | ... |
| windows | g++ 4.9.3 |  yes | 4 |  ... | ... | ... |

If you think 1, 2, 4 threads is overkill or not enough, tell me what we should test. Are there any other configurations we should be testing? g++ on mac? g++-8? g++-7? I know you believe that these other things should just be ignored, but they shouldn't.

I think we should add a few more columns meaning a few more benchmarks, but I don't know what flexes the autodiff stack in the right way.

What we're looking for is a better performance for all configurations. We'll accept things that are within noise. Do you agree? If any of these configurations shows that develop is better, we should not merge.

wds15:
This PR got reverted because of a performance regression under the non-threading case. We have already seen performance improvements under threading and nothing has changed for the threading case. Why do we need to retest this? I mean, this change in this PR takes away abstraction and I have seen speed improvements for clang when doing so. Thus, the threading performance should not degrade. We are a project with limited resources and should act accordingly.

Looking at the table I want to run away!!! Really... this is huge amount of work and requires resources which are out of reason from my perspective. I certainly do not have the time for this - so if you suggest such a huge thing where we don't have automatism then please also suggest how we achieve this. It is your call on this, no challenge, but I would do much more focussed testing and leave some residual risk of a small performance regression... which we found a little bit later in this case using the big performance suite run on the entire example model suite. This somewhat leaner approach will also find the performance regressions, but with using our automatisms and no need for our sweat.

As I say - you are the reviewer here - so more one comments: Why would we need to test more than 1 thread? The program in question does not use threading.

And yes, I think a single run should do given that the end-to-end test is large enough and runs for 200s walltime. This thing anyway needs scripting so if we see unreasonable results and suspect run-to-run variation then replications can be done.

rok-cesnovar:
Sorry, I was off for a few days, catching up now.

I agree with @wds15 that only the non-threaded tests are needed, the model that shows the regression does not use threading anyways. And also the Windows benchmark - develop for threading cases is N/A.

I can run the non-threaded tests on Win+gcc/clang and Linux+gcc/clang. If someone provides the mac ones, that means we can cover the non-threading case which should be enough if @syclik agrees.

syclik:
Thanks. The table was what I was expecting. It sounds like both of you
think otherwise.

Here’s why I think it’s important: when we change this code (autodiff stack
internals), any small change may have an unintended performance effect
that’s unintuitive and can be verified, but not reasoned about. I think
that we could miss a performance regression by only testing a subset of
that.

Question for both of you: why do you think only testing non-threading will
provide us with enough evidence that these changes improve threading? If
there is a good reason, then I’m happy to limit what we test.

(My belief here is that results may change drastically with a little bit of
change to the code, so.... any change in the autodiff stack code should be
evaluated. Let me know why this is incorrect.)

syclik:
And so we’re on the same page: the purpose of this PR is to improve
performance under threading. To me that means: without performance
degradation for non threading and we’re seeing performance benefits for
threading across the supported compilers and wherever we don’t see it, we
can point users to a workaround or we have an understanding.

Since develop doesn’t have this PR in, I think we do have to test both
threaded and non threaded cases. Am I missing the point of the PR?

rok-cesnovar:
As far as I understand the threading cases were already covered before in v1-v5. I wasnt involved in those PRs much, but I believe those tests were done.

Maybe the largest point is that the schools8-4 model that was the only one with the performance regression does not use threading. Or do we want to check if there is a regression if someone enables threading but runs a non-threaded model? If that is the reasoning than yeah, we should check for the entire table. Apart from Windows threading, as there is no baseline on develop to test it with.

wds15:
The key thing why threading got faster is the use of a `thread_local` construct which avoids upon each access a TLS wrapper function to be called. This is achieved by using a compile time constant expression (the null-pointer). This aspect has not changed at all. And as we have very extensively tested this we can omit retesting if threading got faster or not.

Thus, the relative benefits for threading in this PR have not changed and don't need re-evaluation from my understanding.

syclik:
Ok, I understand your reasoning, but I don’t think results hold. In this
PR, the autodiff implementation changed, so those performance measurements
may be similar, but aren’t guaranteed to hold. That's why we missed the
performance regression for non-threaded use; I prefer not to make the same
mistake for the threaded case. (I have no belief that this PR would not
have similar performance, but I wouldn't just merge it assuming that it
holds.)

I think you make a very good point. We should be running a model with
map_rect in addition to this one. For Windows, we don't need to run
multiple times: just once without threading and once without threading.

Does what I'm thinking make sense? I might be overthinking this.

wds15:
Then if we need that many tests again... what do you suggest how we get these?

I have proposed a much leaner approach which may leave residual risk, but that residual risk would be caught by our performance tests anyway. That would be an efficient use of our resources, I think (most of that is automated).

Your requirement for this review sets a huge bar. Will you require this more often so that this needs scripting? Do we have scripts? What do you suggest on how we get there?

And yes... we should make a `map_rect` example with enabled threading part of our regular benchmarks. Right now, the threading performance is basically ignored in all of our automated assessments.

However, adding now an additional model to this table adds with a whopping factor of 2x in the worst case to it...

syclik:
@wds15, I understand your point, but I don't see this as a "re-evaluation." We haven't evaluated this implementation for threading + non-threading; to be more specific, when I say "implementation," I'm actually talking about this particular git hash. The other evaluations done were valid for those implementations and those aren't really that useful because each of those implementations had issues that were uncovered. Let me know if you disagree with that assessment.

(I agree that it's unlikely that the latest changes impact performance, but we can verify empirically and we should before we release it to users.)

wds15:
You make the call here - I do disagree here that this is reasonable to repeat. I have a different take on spending resources balanced with its use here. In your definition of a implementation being equal to a git hash... even spelling mistakes in the doc would lead to different implementations; this is too extreme more me, but you are the authority on this ground to decide.

It would still be good to get suggestions from your end what you propose to get the table filled? It looks like a huge amount of work - at least for me, but maybe you have a clever idea or some tools around which make that easy and efficient. I don't have that.

I just think different - that's all.

syclik:
> Then if we need that many tests again... what do you suggest how we get these?

I think the process is the same we've talked about over the different PRs and in the meeting:

1. We decide what we measure. We can't measure the universe. But we have to measure enough where we believe this won't negatively impact users.
2. We decide on the criteria. I think this is settled. We see no performance degradation for threading and non-threading. (The only way I know how to verify this is to run it.)
3. We evaluate it. If we fail to meet the criteria, we rethink it.

We've looped around and failed to meet the criteria multiple times in different ways.

> Your requirement for this review sets a huge bar. Will you require this more often so that this needs scripting? Do we have scripts? What do you suggest on how we get there?

Yes, this is a huge bar!!!! This isn't a new function that only affects people that use the function. This fundamentally changes the autodiff stack!!! That means that everyone using the Math library is affected. The impact of this necessitates that the effort. (It doesn't matter if this is you, me, or anyone else that works on the project; if you start messing with the autodiff stack, you have to measure performance very carefully because it affects everyone.)

We have to create scripts. For an earlier iteration, I put together some. If you don't have your evaluations scripted, then how did you run your evaluations?


> And yes... we should make a map_rect example with enabled threading part of our regular benchmarks. Right now, the threading performance is basically ignored in all of our automated assessments.

> However, adding now an additional model to this table adds with a whopping factor of 2x in the worst case to it...

Ok. I think if we keep each run to ~5 minutes, it's not unreasonable? I don't think it's unreasonable to dedicate 8 hours of performance testing on each of those platforms. Or even 180 hours. It is that important.

--

The plan: we decide what needs to be tested first. We have:
1. `./example-models/bugs_examples/vol2/schools/schools-4` with these settings: `method=sample num_warmup=250 num_samples=250 data file=example-models/bugs_examples/vol2/schools/schools.data.R random seed=1234`
2. we need a `map_rect` example.

I'd be much more comfortable if we got to the autodiff stack operations that happen through a separate microbenchmark, but I'm not requiring it. (If you recall, I saw the performance degradation in a microbenchmark and I asked us to hold off on merging until we figured it out, but was outvoted. I'd rather not have us make that mistake again.)

wds15:
Well... I totally agree with the need for careful evaluation of this... no question. It's just that we have done that already.

The need to redo this comes from your definition of an implementation - and here I simply don't agree for the reasons I laid out. This is nothing personal, really.

I haven't scripted my evaluations, no. This is why I don't even attempt to start this from my end. If you have scripts which can be used easily for me (running on our cluster) then I can fire off some of these tests.

seantalts:
Quick question - why are we not just IFDEFing out the changes to AD for the non-threaded case to remain exactly the same as on develop? Seems like if we did that we could just do a quick benchmark to see if it improves threading on Linux OR Mac and call it a day.

syclik:
> The need to redo this comes from your definition of an implementation - and here I simply don't agree for the reasons I laid out. This is nothing personal, really.

Yup. I get it. But... these changes aren't just doc changes. These are substantial changes to the code. (Even a small spelling error in an `#ifdef` is going to change everything; let alone shifting pointers, etc.) So let's make sure to evaluate everything on the latest git has so we're certain.

> I haven't scripted my evaluations, no.

I now understand why you think this is a big deal. We'll script it. It won't be extra work. It'll just be extra time.

> EDIT: Committing on hard thresholds is something I would not like to do, really (feels like 5% p-value decisions). 15% regression is large, sure... but right now this is only for a single model. All in all we have done a very good job in the past to ensure there is no performance regression (constrained on the performance benchmark set and our understanding of things).

The existence of any model that shows a significant regression is problematic. The reason is that we haven't run a very large suite of models. If it's happening for one, then it will probably happen for other models. If we could clearly under what conditions this happens, that's at least something. Buf if we just have certain models that will run 15% slower, then it's a problem. This PR is not supposed to add a performance regression anywhere. This PR is supposed to dominate performance. Right?

syclik:
> Quick question - why are we not just IFDEFing out the changes to AD for the non-threaded case to remain exactly the same as on develop? Seems like if we did that we could just do a quick benchmark to see if it improves threading on Linux OR Mac and call it a day.

1. maintenance
2. conceptual debt (thanks for linking to that in the design docs wiki)

But yes, this PR could be redesigned so it left the non-threaded autodiff stack alone, but isn't wasn't how this is implemented.

seantalts:
> maintenance

Last time I coded it up for one of the other PRs, it was something like 4 extra lines contained to one file 

> conceptual debt (thanks for linking to that in the design docs wiki)

The IFDEFing doesn't leak out of the single file, and that file maintains the encapsulation. The concept of a global AD tape returned by a call to `instance()` remains the same and does not leak through its abstraction.

seantalts:
That said, if you all have figured out a way to do the necessary shenanigans without any performance hit on non-threaded develop, then that might be worth it (depending on the hoops that were required for that one, of course). Anyway, sorry to do a drive-by here, I am just struggling to keep up with this conversation.

syclik:
@seantalts: those are my thoughts exactly. There was an earlier version that changed the interface, but now that the interface isn't changing, it's ok to actually have this as an adapter / interface (with appropriate doc).

But first, I'm hoping that we can measure these things on this PR and be done with it. If we changed the implementation, we'd still have to evaluate performance for threading again. So the extra work is just measuring it for the non-threaded case. So we're not saving much by changing the implementation.

wds15:
The PR right now takes away the abstraction of the `instance()` call... and I have observed speedups under clang on macOS for the no threading performance... which everyone will appreciate I suppose. If you don't have this method abstraction than going with a pointer / no pointer design is at best mood.

@syclik No I don't have scripts... I only use the performance benchmark python scripts. That's all... I don't know many of the features of that thing... this is why I am very scared.

but again...this PR deserves large effort in testing the performance - and I have done that - and we do agree in the necessity for that; we just define relevant implementation changes differently.

wds15:
Ah... and as `map_rect` example we can use the warfarin case from my StanCon contribution. We used that before.

syclik:
Perfect. Let's just run those two examples.

seantalts:
> and I have observed speedups under clang on macOS for the no threading performance...

Whoa, that's awesome! I definitely lost track of that in the conversation. 

If we all worked in the same office, we could ring a gong and break out some prosecco every time someone gets the benchmark suite to run faster with a merge to develop. I'm not sure what the remote open source equivalent is...

wds15:
Do we have an idea when the requested benchmarks will be done? It would be really great if this PR makes it into 2.20 - which is approaching in a month as I understood.

rok-cesnovar:
Was the final conclusion to run the schools v4 and wafarin model? Is this just waiting on me to run this on ubuntu and Windows? Sorry I dropped the ball here. I will have resources available this week with finalizing the GPU paper done and can run those.

syclik:
Thanks for checking in.

1. Yes, those two models.
2. No, it's not waiting on you. It's waiting on me.

No worries about timing. We've all been tied up.

Here's the current status on my end:
- I have scripts set up in this repo / branch:
https://github.com/syclik/benchmark-stan/tree/math/1135-ad?files=1
- there was an issue with non-threaded generation, so I need to go back
through the scripts
- I haven’t tested windows yet.

If you’re good with batch and want to run on windows, that’d be great.
Essentially the same seed, run 10 times, under no threading, 1, 2, 4
threads, no threading on both, threading on just the new. If not, I was
going going to try to tackle that this week.

syclik:
whoops, wrong branch. correct branch: math/1245-tls

rok-cesnovar:
Sure, I can try and get Windows up and running. Will get on it tomorrow.

syclik:
Thanks. I'll post here if I get it up first.

syclik:
@wds15, I updated the top-level comment. The improvements on mac and linux with the right toolchains are great! We just need to check on Windows and we're good.

wds15:
Whow... this looks very nice! Thanks for all the work on this to confirm.

For Windows: Using more than one thread does not work with develop; so maybe not all comparisons are possible there.

syclik:
Yes, that's right. We'll measure what we can. There's a nice speed improvement with the warfarin example with more threads. If you notice, Mac does better than Linux in that regard. I don't know how Windows will do.

wds15:
Yup - this change makes threading work a lot faster. If we add the scalable memory allocator from the TBB, then you end up getting basically the same speed bump like MPI:

https://discourse.mc-stan.org/t/map-rect-threading/8802/4

There you see the performance scaling of MPI, new TLS, new TLS+TBB scalable memory allocation.

(of course - the TBB is offtopic here - sorry, but we have are about to get serious on threading)

rok-cesnovar:
@syclik I got the run-no-threads.sh up and running on Windows. Will post the raw results here. 

The only issue was with `tee` in statements like `echo $(date) | tee -a progress.txt` that is just replaced with  `echo $(date) >> progress.txt`

syclik:
awesome!!! I was just going to ping you. I think I also got the no-threads running. But I wanted to test with threading too. What I don't know is how threading actually scales. (if it doesn't compile on the old stuff, that's fine. But we should still know how it behaves with the new stuff)

https://github.com/syclik/benchmark-stan/tree/math/1245-tls

syclik:
Btw, the only reason I did that with `progress.txt` was so I could start the process with nohup and come back to see the progress; it's not critical to how the script works.

syclik:
@wds15, just so we're on the same page: before this PR, on Windows, it compiles, but has runtime errors? Because that's what I'm seeing.

syclik:
@rok-cesnovar, can you push to that branch? I can run it if you've got a script for it. And we should run it for threading too to make sure it works there.

wds15:
It compiles under Windows, yes. It also runs under Windows - but only if STAN_NUM_THREADS=1. Whenever you set STAN_NUM_THREDS>=2 then the binary will start, but at the point when the program tries to access the AD tree the program terminates immediately.

rok-cesnovar:
@syclik I am about halfway through the threading tests. I have all the results for the non-threaded tests (It go late last night when those finished). 

Then I just need to grep the numbers with analyze.sh and we should be good to go. The non-threaded numbers indicate that we are in the clear. Will ping you when I have everything.

syclik:
@rok-cesnovar, thanks!

syclik:
@wds15, thank you for that information. I think this has uncovered a different issue. If I run the warfarin model a few times in a row, I see something different:
```
Rejecting initial value:
  Error evaluating the log probability at the initial value.
Exception: Exception: Exception: gamma_lpdf: Inverse scale parameter[2] is -8.04187e-005, but must be > 0!  (in '../benchmark-warfarin-new.stan' at line 48)
...
Unrecoverable error evaluating the log probability at the initial value.
Exception: Exception: Max number of iterations exceeded (500).  (in '../benchmark-warfarin-new.stan' at line 151)
```

I don't think it should terminate the first time it hits max iterations, right? (I don't even know where that's coming from off the top of my head)

wds15:
Looks to me as if the initial leads to a log-lik value which is NaN. Therefore, Stan does not reach the part of the program where autodiff is needed, since the log-lik is evaluated in double only mode before sampling starts.

It's a bit weird that the problem happens under Windows now. The behaviour is the same under `STAN_THREADS` and without that?

Here is one more seed which I generated a while ago for this model (maybe it just runs with that):

```
theta <-
c(4.41959708856204, 3.23997089513979, 2.12421793294719)
Eta <-
structure(c(-0.688248187383304, 0.744680093701252, -0.87312505551833,
-0.378427566950146, 0.703664990470604, -0.597577248428612, -0.317997718239967,
-0.510579753740771, 0.507684062889378, 0.110256075607081, -0.977028657016065,
-0.55919410668034, 0.517432664733083, -0.13755612976235, -1.21729681828442,
0.974770378905777, 0.903864698185265, 0.29870862350923, -0.216401920029647,
0.00876521732189567, -0.285117825818102, -0.0499555039319689, 0.169276053133337,
0.192171211042593, 0.345980466957853, -0.225143924257664, 0.175210232131478,
-0.916697488535281, 1.22011176305615, -0.630232835212122, 0.555121034389788,
-0.868947709611385, -0.451069739995428, -0.451109720549416, -0.0248418523603254,
-0.293184152403504, 0.0240323637355454, -0.539375773159626, 0.547556570313531,
-0.0160069609940164, -0.674651403389227, 0.504913201382093, 0.255341722622165,
-0.530527237581262, -0.37592245465429, -0.418677546519841, 0.0671013287388712,
0.619973031935115, -0.237134044722137, 0.399927328640661, 0.378288558610415,
0.649407271155856, 0.393999107298632, 0.249552772545313, -0.769310807474185,
-0.946798265121144, 0.241608938749476, 0.23830579814668, -0.611445656642559,
-1.00529071402882, -1.24922186068239, -0.481397035229574, -0.367325781115606,
0.0716738574618031, 0.0447101604372022, -0.408015741482741, -0.0544900550162833,
0.0424837214329659, 0.692393114416288, -0.289936328993551, -0.150583561100659,
-0.576380454419351, -0.557948532694114, -0.567403944520817, 0.350128263725001,
-0.278313223250863, 0.291253358591588, 0.404988157479207, 0.180376205883475,
-0.343655710749863, -0.214261567753269, -0.421017844331495, -0.159577540535779,
-0.197581269886916, 0.139487629172406, -0.157859033760607, 0.129493637619296,
-0.875976122232035, -0.0669231418917255, -0.685125188493358, 0.380993970211202,
0.440506764954287, -0.0792972660415051, 0.988422562068841, 0.0947271220011503,
-0.640506532684065),
.Dim = c(32, 3))
sigma_eta <-
c(0.26607685943578, 0.414870876335879, 0.323991053083554)
sigma_y <- 0.66199295742126
```

rok-cesnovar:
Here is the time.txt file from analyze.sh for windows.
[time.txt](https://github.com/stan-dev/math/files/3315703/time.txt)

this include no-threads for both develop and TLSv6 and threads=1,2,4 for TLSv6. 

Bottom line is that this is good to go. 
@syclik do you have the time to populate the table with the results? I am not sure if you took the mean or median of the 10 runs to put in the table. I can do it if you let me know. I will update the benchmark repo with the windows script once I clean things up.

syclik:
Awesome! I took the mean.

syclik:
And I can populate it. In a little bit. Unless you’re doing it already.

rok-cesnovar:
I can do it, no problem.

rok-cesnovar:
I populated the table in the original post of the PR. When not using threads the execution times are basically the same and when using multiple threads things look similar to unix systems. @syclik and @wds15  please review the numbers and then LGTM!

wds15:
It is interesting to see the execution times go considerably up when going from no threading on Windows to threading. I remember reading in the docs of the TBB that windows threads are terribly expensive (which is why a threadpool should speed this up as implemented in the TBB). Other than that it looks all fine to my eyes.

syclik:
Great! Thank you. Yeah, that's good to know. We should really suggest people not use threading with 1 thread. I'm really glad there is an improvement at 2 threads.

syclik:
What's the deal with Jenkins? Anyone know why it started a new job?

wds15:
No idea...had the same thought...

bob-carpenter:
I think mean's better for testing unless you expect high outliers due to something like other system demand.

rok-cesnovar:
@serban-nicusor-toptal hey, when you have the time please take a look at what is going on here. The last push to this PR's branch was 11 days ago and it seems that the tests keep restarting.

wds15:
I did manually restart the last one since Jenkins did hung up on Windows. I think Windows testing is right now somewhat fragile.

wds15:
Now I see what is possibly the problem: On the windows the command run is

`runTests.py -jnull test/unit -f thread --make-only`

So the `-jnull` argument is the problem possibly. I will look quickly into that...

wds15:
To me this looks like a config error of the Jenkins on the Windows instance we are getting. On that machine `env.PARALLEL` has the value `null` which is not quite right.

seantalts:
Yeah, that looks right - seems like a config problem with the new EC2 Windows instances Nic has been setting up. @serban-nicusor-toptal for now I turned off the 'windows' label on those nodes so they shouldn't start up and I'll re-run this this job.

serban-nicusor-toptal:
Investigated and it's an issue from the Windows AMI.
Will push an updated version and everything should go back to normal.

Issue #1251, opened 2019-05-23
Implement ordinal regression GLM (ordered_logistic_glm_lpmf)


t4c1:
## Description
Implement a GLM for ordinal regression. 

## Example
`ordered_logistic_glm_lpmf(y, x, beta, cuts)` should be equivalent to `ordered_logistic_lpmf(y, x * beta, cuts)`.

## Expected Output
The new GLM should use analytically derived gradients and GPU to be faster than the existing option how to implement ordinal regression.

#### Current Version:
v2.19.1

PR #1249, opened 2019-05-21
Bugfix - Incorrect gradients for Ordered Logistic Distribution

andrjohns:
## Summary
Discovered by @t4c1 in #1248. The ordered logistic distribution is returning incorrect gradients for the ordered cutpoints if the input array of integers has multiple 1's

This was due to the following line:
```
      if (!is_constant_struct<T_cut>::value)
        ops_partials.edge2_.partials_vec_[n](0) = d;
```

When ```T_cut``` is an ```Eigen::Vector```, ```partials_vec_[n]``` always refers to ```partials_vec_[0]```, so the gradient calculation was overwriting ```ops_partials.edge2_.partials_vec_[n](0)``` at each of the ```n``` iterations, rather than incrementing it.

## Tests
Added a new test for the gradients when the input array of integers are all 1's

## Checklist

- [x] Math issue #1248 

- [x] Copyright holder: Andrew Johnson

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1252, opened 2019-05-23
Implement ordinal regression GLM (ordered_logistic_glm_lpmf)

t4c1:
## Summary
This implements ordinal regression GLM (ordered_logistic_glm_lpmf) for CPU.

## Tests
New tests are in `rev/mat/prob/ordered_logistic_glm_lpmf_test.cpp`.

## Side Effects
None.

## Checklist

- [x] Math issue #1251 


stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,174 @@
...
+                             + m_log_1p_exp_cut1))
+             .sum();
+
+  operands_and_partials<Matrix<T_x_scalar, Dynamic, Dynamic>, T_beta, T_cuts>
+      ops_partials(x, beta, cuts);
+  if (!is_constant_struct<T_x_scalar>::value
+      || !is_constant_struct<T_beta>::value
+      || !is_constant_struct<T_cuts>::value) {
+    Array<double, Dynamic, 1> d1
+        = 1 / (1 + exp(cut2)) - 1 / (1 - exp(cuts_y1 - cuts_y2));
```
andrjohns: It could be a good idea to implement ```1 / (1 + exp(cut2))``` using stans ```inv_logit``` function, since this is constructed to avoid underflow, something like:
``` cut2.unaryExpr([](double x){ return inv_logit(-x); })```

For ```1 / (1 - exp(cuts_y1 - cuts_y2))```, you could use Eigen's coefficient-wise functions to take advantage of any SIMD extensions:
``` (1 - (cuts_y1 - cuts_y2).exp()).inverse()```

Comment (t4c1) [reply stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp]: I'm avoiding stan functions, since using them creates unnecessary temporaries. Also they don't seem to take advantage of SIMD. But I can use the same trick to avoid underflow.

As for your second suggestion I think it should compile to the same binary code as my code.

**Comment by t4c1** [comment]
I measured speedups of this compared to using `ordered_logistic_lpmf()`. Benchmark code is [here](https://gist.github.com/t4c1/bb442ab936def8b0bf24f86fcb43363c). Here are results: 

glm_speedup

File: stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp [hunk 2]
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @tparam T_y type of integer vector of dependent variables (labels)
+ * @tparam T_x_scalar type of elements in the matrix of independent variables
+ * (features)
+ * @tparam T_beta T_beta type of the weight vector
+ * @tparam T_cuts T_beta type of the cutpoints vector
+ * @param y integer vector parameter
+ * @param x design matrix
+ * @param beta weight vector
+ * @param cuts vector of cutpoints
```
bob-carpenter: I assume this only accepts a column vector, since that's what we take "vector" to mean. Otherwise, please indicate this can be a row vector or column vector.  We want the doc to be clear on types here since the templating isn't providing that information when you template the whole container.

I'm just adding single comments rather than reviewing as I do'nt want to overreview someone else who might be doing this.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+  check_consistent_size(function, "Vector of dependent variables", y,
+                        N_instances);
+  check_consistent_size(function, "Weight vector", beta, N_attributes);
+  check_bounded(function, "Vector of dependent variables", y, 1, N_classes);
+  check_ordered(function, "Cut-points", cuts);
+  check_finite(function, "Final cut-point", cuts[N_classes - 2]);
+  check_finite(function, "First cut-point", cuts[0]);
+
+  if (size_zero(y, x, beta, cuts))
```
bob-carpenter: This isn't your fault, but it's another function that landed without sufficient review---it should be named `size_zero_any` or something like that.  Anyway, guess that's a different PR.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,174 @@
...
+    check_finite(function, "Weight vector", beta);
+    check_finite(function, "Matrix of independent variables", x);
+  }
+
+  Array<double, Dynamic, 1> cut2 = location - cuts_y2;
+  Array<double, Dynamic, 1> cut1 = location - cuts_y1;
+
+  // Not immediately evaluating next two expressions benefits performance
+  auto m_log_1p_exp_cut1
+      = -cut1 * (cut1 > 0.0).cast<double>() - (-cut1.abs()).exp().log1p();
```
bob-carpenter: What is the type of `(cut1 > 0)` before the `.cast<double>()`?  I'm having trouble understanding what this cast might be doing.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,174 @@
...
+    Array<double, Dynamic, 1> d1
+        = 1 / (1 + exp(cut2)) - 1 / (1 - exp(cuts_y1 - cuts_y2));
+    Array<double, Dynamic, 1> d2
+        = 1 / (1 - exp(cuts_y2 - cuts_y1)) - 1 / (1 + exp(cut1));
+
+    if (!is_constant_struct<T_x_scalar>::value
+        || !is_constant_struct<T_beta>::value) {
+      Matrix<double, 1, Dynamic> location_derivative = d1 - d2;
+      if (!is_constant_struct<T_x_scalar>::value) {
+        ops_partials.edge1_.partials_
```
bob-carpenter: Just checking that these shouldn't be `+=`.  We've run into that bug in other uses of `operands_and_partials` recently.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
rok-cesnovar: Yeah, this one is on me. This got in before the _any discussion in the `is_nan` PR. Sorry about that. I am assigning myself to fix this one.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: It is `Eigen::Array` of `bool`s. Eigen only allows operations between arrays/matrices that contain same type so we need to cast that to `Eigen::Array` of `double`s

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: No, `=` is correct here. This is not even in a loop, so there is no danger of overwriting previously calculated value.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
bob-carpenter: I see---it's an Eigen operation---makes sense.  But isn't that particular operation going to be very expensive?  I don't have a good feel for what doing all this multiplication by zero and one for selection is going to cost.  So I think if `cut1 = { -2, 1, 0, -3, 4 }`, then `-cut1 * (cut1 > 0.0).cast<double>()` is just `{ 0, -1, 0, 0, -4 }`.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
bob-carpenter: OK, thanks for checking.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: Multiplication is not expensive. I don't think the cast would be either. Anyway I compared this with stan `log_1p_exp()` and it is faster as it is vectorized and contains no branches. I can also try comparing that with `.select()` to see if that would be better.

bob-carpenter:
Whether multiplication is expensive or not depends on the context.  It's slower than addition, but faster than exponentiation and much faster than a cache miss on memory.

Doesn't evaluating .(cut > 0) require branching.  I can imagine it's better at allow vectorization than a literal branch in the code.

If you know how to do it, it would be great to speed up Stan's vectorized functions.  As is, they're vectorized naively by just calling the function repeatedly.  Not in this PR, of course!

t4c1:
cut>0 is just a comparison, not a branch. I am not sure, but I think comparisons can be vectorized.

Explicitely vectorizing code requires use of intrinsics, which is tedious and not portable. Luckily Eigen does it, so we do not have to. I just assume all Eigen expression will produce vectorized binary code. While there might be exceptions, Eigen expessions seem to always results in equivalent or faster calculations than doing things element-wise.

bob-carpenter:
> On May 28, 2019, at 2:54 PM, t4c1 <notifications@github.com> wrote:
> 
> cut>0 is just a comparison, not a branch. I am not sure, but I think comparisons can be vectorized.

That would be cool if they could be vectorized.  It's not a branch in the if/then structure, but internally, it has to be doing a comparison and taking different behavior based on the result as that's the semantics.  But it may be encapsulated such that it doesn't look like a branch to the compiler.  That's why I was asking.  My compiler knowledge isn't that deep.

> Explicitely vectorizing code requires use of intrinsics, which is tedious and not portable. Luckily Eigen does it, so we do not have to. I just assume all Eigen expression will produce vectorized binary code. While there might be exceptions, Eigen expessions seem to always results in equivalent or faster calculations than doing things element-wise.

That seems to imply we should redo our vectorization to lean on whatever Eigen's using to do it.  I'm not saying we should write our own proper vectorization library, just that we might be able to lean on Eigen's.  You seem to be writing faster versions of our vectorized functions using Eigen.  I'd like to expose that to users as elementwise application of functions happens everywhere.

t4c1:
>taking different behavior
That is not true. Comparisons are basic CPU instructions, just like for example adition. Just the result is either 0 or 1. Comparisons without if/loop are not branches.

bob-carpenter:
I think I see what you're getting at.  Just because comparison has an intrinsically branching behavior, it doesn't entail generating branching instructions.  Is the relevant feature here whether you need to jump for one of the branches?

On a side note, do you know if the ternary conditional operator (a ? b : c) generates branches?  

There's a lot of bad folk wisdom around all this stuff and I've never dug into it deeply enough or found it made enough of a difference to show up in end-to-end profiling.  So thanks for bearing with.

t4c1:
Sorry, I don't understant, what do you mean with the first question. Ternary operator is a branch. 

On a side note I assume we want this GLM to have similar input types as [softmax regression](https://github.com/stan-dev/math/pull/1242#issuecomment-496027834).

bob-carpenter:
Let me elaborate with an example.  Let's say I have a naive implementation of absolute value using the ternary operator.

```
a > 0 : a ? -a
```

The question is whether this generates branching instructions or not.  How would that compare to the 

```
(a > 0) * a + (1 - (a > 0)) * -a
```

which shouldn't generate any branching.  I saw a Stack Overflow discussion about this recently which said the first version's faster on modern compilers, but now I can't find it.

t4c1:
FIrst one would generate a branch and second would not. However, both might be optimized by a compiler.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: It turns out that `.select()` (which is a branch) is faster.

t4c1:
This test failure was expected and will disappear once [PR1249](https://github.com/stan-dev/math/pull/1249) gets merged.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,222 @@
+#ifndef STAN_MATH_PRIM_MAT_PROB_ORDERED_LOGISTIC_GLM_LPMF_HPP
+#define STAN_MATH_PRIM_MAT_PROB_ORDERED_LOGISTIC_GLM_LPMF_HPP
+
+#include <stan/math/prim/scal/meta/is_constant_struct.hpp>
```
rok-cesnovar: Please replaces the meta includes with prim/meta.hpp, rev/meta.hpp etc. We should really test for these with test-math-dependencies or something like that.

bob-carpenter: The reason to include higher level is to get the order of includes right among the generic, `std::vector`, and `Eigen::Matrix` signatures. 

If we only include **`prim`**/`meta.hpp`, will the higher level `rev/meta.hpp` get included in the right order elsewhere if any `rev` gets included?

syclik: @bob-carpenter: the short answer is yes. If this file is in prim, it should include the prim meta file. When it gets used with rev, the client of the Math library should be including rev/mat.hpp, which includes rev/meta.hpp prior to including this file.

rok-cesnovar:
Do we have any volunteers to review this? Bob's and my intial comments were addressed so this is definitely ready for a proper review.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+ * @param cuts cutpoints vector
+ * @return log probability
+ * @throw std::domain_error If any class is not between 1 and
+ * the number of cutpoints plus 2 or if the cutpoint vector is not sorted in
+ * ascending order or any input is not finite
+ * @throw std::invalid_argument if container sizes mismatch.
+ */
+template <bool propto, typename T_y, typename T_x_scalar, int T_x_rows,
+          typename T_beta_scalar, typename T_cuts_scalar>
+typename stan::return_type<T_x_scalar, T_beta_scalar, T_cuts_scalar>::type
```
andrjohns: This can just be ```return_type_t<T_x_scalar, T_beta_scalar, T_cuts_scalar>``` now

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+  using Eigen::Array;
+  using Eigen::Dynamic;
+  using Eigen::Matrix;
+  using Eigen::VectorXd;
+  using std::exp;
+  using std::isfinite;
+
+  typedef typename partials_return_type<T_y, T_x_scalar, T_beta_scalar,
+                                        T_cuts_scalar>::type T_partials_return;
+  typedef typename std::conditional<T_x_rows == 1, double,
```
andrjohns: Can use std::conditional_t here

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+  if (is_vector<T_y>::value && T_x_rows != 1) {
+    check_consistent_size(function, "Vector of dependent variables", y,
+                          N_instances);
+  }
+  check_consistent_size(function, "Weight vector", beta, N_attributes);
+  check_bounded(function, "Vector of dependent variables", y, 1, N_classes);
+  check_ordered(function, "Cut-points", cuts);
+  check_finite(function, "Final cut-point", cuts[N_classes - 2]);
+  check_finite(function, "First cut-point", cuts[0]);
```
andrjohns: What's the benefit to only checking if the first and last cut-points are finite? Shouldn't all of them be checked?

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+  if (size_zero(y, cuts))
+    return 0;
+
+  if (!include_summand<propto, T_x_scalar, T_beta_scalar, T_cuts_scalar>::value)
+    return 0;
+
+  T_partials_return logp(0);
```
andrjohns: The general preference is to have the variable declaration as close as possible to the variable's use

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+    if (c != 1) {
+      cuts_y2[i] = cuts_val_vec[c - 2];
+    } else {
+      cuts_y2[i] = -INFINITY;
+    }
+  }
+
+  T_location location = x_val * beta_val_vec;
+  if (!isfinite(sum(location))) {
```
andrjohns: Not for this pull, but it could be a good idea in the future to have a ```mat``` overload for ```is_inf```, so that this intermediary ```sum``` isn't required

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+              + m_log_1p_exp_cut1)
+                 .sum();
+    }
+  }
+
+  operands_and_partials<Matrix<T_x_scalar, T_x_rows, Dynamic>,
+                        Eigen::Matrix<T_beta_scalar, Eigen::Dynamic, 1>,
+                        Eigen::Matrix<T_cuts_scalar, Eigen::Dynamic, 1>>
+      ops_partials(x, beta, cuts);
+  if (!is_constant_all<T_x_scalar, T_beta_scalar, T_cuts_scalar>::value) {
```
andrjohns: I don't think you want ```!is_constant_all<>``` here, since you'll want to calculate gradients in the situation where some (but not all) are autodiff types

I wonder if an ```is_constant_any<>``` could be a good idea for the future

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+    Array<double, Dynamic, 1> exp_m_cut1 = exp(-cut1);
+    Array<double, Dynamic, 1> exp_m_cut2 = exp(-cut2);
+    Array<double, Dynamic, 1> exp_cuts_diff = exp(cuts_y2 - cuts_y1);
+    Array<double, Dynamic, 1> d1
+        = (cut2 > 0).select(exp_m_cut2 / (1 + exp_m_cut2), 1 / (1 + exp(cut2)))
+          - exp_cuts_diff / (exp_cuts_diff - 1);
```
andrjohns: Couldn't you replace both of these ```cut2``` calls with ```inv_logit(-cut2)```?

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+              = beta_val_vec * location_derivative.sum();
+        } else {
+          ops_partials.edge1_.partials_
+              = (beta_val_vec * location_derivative).transpose();
+        }
+      }
+      if (!is_constant_all<T_beta_scalar>::value) {
+        if (T_x_rows == 1) {
+          ops_partials.edge2_.partials_
+              = (location_derivative * x_val.replicate(N_instances, 1))
```
andrjohns: I'm not sure which is faster, but you could also do:
```
location_derivative.array().rowwise() * x_val.array()
```

Whether that's more efficient than replicating, I'm not sure

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
```diff
@@ -0,0 +1,212 @@
...
+        }
+      }
+    }
+  }
+  return ops_partials.build(logp);
+}
+
+template <typename T_y, typename T_x_scalar, int T_x_rows,
+          typename T_beta_scalar, typename T_cuts_scalar>
+typename stan::return_type<T_x_scalar, T_beta_scalar, T_cuts_scalar>::type
```
andrjohns: ```return_type_t``` could also be used here

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: If any other cutpoint were infinite and both endpoints finite, `check_ordered` would fail.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: That if says "if not all are constant", which is equivalent to "is any non-constant". This is exactely as intended, as we need derivatives only for non-constant variables.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: I could. However, calling a stan function requires evaluating its arguments before call and it returns evaluated result. What I am doing here is constructing an expression which can be evaluated whole at once. That is more efficient. Btw we had exactley the same discussion at the top of this PR.

I think @SteveBronder is working on removing these inefficiencies from stan functions. Once he is done, this can be replaced with call to `inv_logit()` without sacrificing performance.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
t4c1: I think both would likely produce same binary code. I guess your version might be a bit clearer.

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
andrjohns: Huh, very clever, I like that

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
andrjohns: Right you are

stan/math/prim/mat/prob/ordered_logistic_glm_lpmf.hpp
andrjohns: That makes sense, thanks for clearing that up (again)

t4c1:
@andrjohns This is ready for next review.

Issue #1260, opened 2019-06-06
GLM tests improvements


t4c1:
## Description
We have three GLMs that accept integer y (`bernoulli_logit`, `poisson_log` and `negbin_2_log`). Right now these GLMs are tested by passing in y of type `Eigen::Matrix<int, Dynamic, 1>`. However if used from actual stan models y can only be of type `std::vector<int>` or `int`. The reason being that vectors in stan implicitely contain elements of type `double`, while multiple `int`s can only be passed around in an array. So the tests should be changed to use `std::vector<int>` as well.

## Expected Output
No changes in any outputs.

#### Current Version:
v2.19.1

PR #1261, opened 2019-06-06
GLM tests improvements

t4c1:
## Summary

Changes testing of GLMs that accept integer y from using `Eigen::Matrix<int, Dynamic, 1>` to `std::vector<int>`. See the issue #1260 for more info.

## Checklist

- [x] Math issue #1260 

### DIFF
--- a/test/unit/math/rev/mat/prob/bernoulli_logit_glm_lpmf_test.cpp
+++ b/test/unit/math/rev/mat/prob/bernoulli_logit_glm_lpmf_test.cpp
@@ -9,12 +9,12 @@
 using Eigen::Dynamic;
 using Eigen::Matrix;
 using stan::math::var;
+using std::vector;
 
 //  We check that the values of the new regression match those of one built
 //  from existing primitives.
 TEST(ProbDistributionsBernoulliLogitGLM, glm_matches_bernoulli_logit_doubles) {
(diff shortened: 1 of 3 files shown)


t4c1:
This is not an issue with stan language (Unless you think the fast that there can be no matrix/vector of integers in stan, which is not in scope of this PR). This simply improves some test cases of stan-math to be closer to how GLMs are used from stan language. 

There is no change in this PR, except for the tests of stan-math. So there is nothing to test from interfaces. I haven't done any tests in stan repo so I don't know how things are tested there.

bob-carpenter:
> Unless you think the fast that there can be no matrix/vector of integers in stan

Stan only uses `Eigen::Matrix<T, R, C>` where the scalar type `T` is `double` or an autodiff type---it's never an integer type.

> So there is nothing to test from interfaces.

In general, we want tests in the math library to live independently in the math library.  When functions get imported into the Stan language, then we test that they compile in the stan repo.

> It's kind of obnoxious to cast something into an Nx1 matrix. It's wasteful, and cumbersome.

We definitely don't want to have to be doing that if we an avoid it.

t4c1:
@drezap Since this is just a simple change in tests, could you review it? Are any changes needed or is this OK?

Issue #1198, opened 2019-04-04
Give matrix_cl matrix type attributes


SteveBronder:
## Description

On my to-do list. Right now we have code in the OpenCL stuff that looks like

```cpp
  matrix_cl L_21
      = opencl::multiply<TriangularViewCL::Entire, TriangularViewCL::Upper>(
          A_21, transpose(lower_triangular_inverse(L_11)));
```

This lets us do optimized multiplications for matrix_cl's with specific characteristics. But the above is kind of a weird API to me. I'd like to have something like

```cpp
matrix_cl L_21 = A_21 * transpose(lower_triangular_inverse(L_11));
```

Where the LHS and RHS of * are `matrix_cl`s with a `TriangularViewCL matrix_view` (name subject to change) attribute. In functions where we know they will return a lower or upper triangular the function can change this value. Similarly things like transpose can see they are transposing a lower triangular to upper and change the `matrix_view` appropriately.

With that we can have the multiply code look at the `matrix_view` for each matrix and dispatch to the appropriate optimized multiplication or other optimized function

Sean / Rok, any weird gotchas or things I should think about when prototyping this?

## Example

```cpp
matrix_cl A<TriangularViewCL::Upper>(size, size);
matrix_cl B<TriangularViewCL::Entire>(size, size);
// This will do the optimized triangular * rectangular multiply
matrix_cl C = A * B;
```

## Expected Output

Output should stay the same

#### Current Version:
v2.19.0

rok-cesnovar:
Love this! I was not happy with the API either but had no good ideas on how to simplify it. This will definitely find some additional optimizations that we havent found manually.

Transpose, lower traingular inverse, cholesky decomopose, zeros etc seem clear candidates for setting/modifying the view. For kernels like add/subtract we should keep the view if both inputs have the same view, otherwise set to entire. But I am probably just repeating what you already know.

seantalts:
Yeah, this seems really good at least prima facie! Does this also imply we need to change e.g. `transpose` et al to be methods?

SteveBronder:
> Does this also imply we need to change e.g. transpose et al to be methods?

idt so, should it be? I think if something takes in a matrix_cl it can change the matrix_view.

Rok I was going to take a crack at fixing up the cache stuff on top of the async this weekend. Do you have any time on your plate to work on this? I figure it would look something like

1. Add a template to matrix_cl constructor like
```cpp
<TriangularViewCL matrix_view = TriangularViewCL::Entire>
[...]

PR #1266, opened 2019-06-14
Added triangularity attribute to matrix_cl

t4c1:
## Summary
Added `triangular_view_` attribute to `matrix_cl`. All OpenCL functions now respect and propagate this attribute. Especially `add()`, `subtract()` and `scalar_mul()` kernels were significantly modified. 

## Tests
New tests were added for `add()`, `subtract()` and `.sub_block()`. Some small changes to others.

## Side Effects
`enum class TriangularViewCL` now has an additional possible value: `Diagonal`. `TriangularViewCL` now works as a bitfield: `Lower | Upper = Entire`, `Entire & Lower = Lower`, `Diagonal | Lower = Lower` ...

`constants.hpp` renamed to `triangular.hpp`

## Checklist

- [x] Math issue #1198 


stan/math/opencl/triangular.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef STAN_MATH_OPENCL_TRIANGULAR_HPP
+#define STAN_MATH_OPENCL_TRIANGULAR_HPP
+#ifdef STAN_OPENCL
+namespace stan {
+namespace math {
+enum class TriangularViewCL { Diagonal = 0, Lower = 1, Upper = 2, Entire = 3 };
```
SteveBronder: Should we change the name of this to `TriangularTypeCL` or something like that? We could probably remove the CL part

t4c1: We still store whole matrix and this attribute only determines which part most functions view. So I prefere view to type. I also like that it has something OpenCL-related in its name. If we move it to some cl-related namespace than im fine with dropping CL from the name.

drezap:
So I just want to understand why this is important, I want to develop some literacy. 

So openCL is the kernel that manages operations on the GPU. This pull request is to enable operations having to do with PSD or triangular matrices, on the GPU. Exploiting triangular matrices has large speed gains and memory gains. For a simple example, we've added GPU operations for addition for triangular matrices. It's obviously faster to sum the n(n-1)/2 operations than sum half of (or half +/1 the diagonal) a matrix of zeros. But these speed gains are even better on the GPU, could be useful for any matrix operations in the future. Just very generally applicable. 

Is this correct? I can read the code to see how it's functioning, but hard to see how to improve it besides style, so I'm not really sure I can review this.

t4c1:
> Is this correct?

No. First there are no memory gains. We still store whole matrix, even if we need only a triangular matrix. This is simpler and prrobably does not affect performance that much. Operations that are faster on triangular matrices were optimized in this way before this PR. 

What this PR changes is that it lets the matrix track which part of it is actually in use. So that does not need to be specified when calling a function that operates on a triangular part. For example addition of two upper triangular matrices produces an upper triangular matrix. Some functions needed to be changed to only operate on triangular part if it is specified on input matrix. However I expect no or minimal benefits to performance.

OIn OpenCL terminology a kernel is any function that is called from host (CPU) and executes on OpenCL device (GPU). I don't know what PSD is.

bob-carpenter:
> We still store whole matrix, even if we need only a triangular matrix. This is simpler and prrobably does not affect performance that much. Operations that are faster on triangular matrices were optimized in this way before this PR.

This is also how Eigen deals with triangular matrices.


> What this PR changes is that it lets the matrix track which part of it is actually in use. So that does not need to be specified when calling a function that operates on a triangular part. For example addition of two upper triangular matrices produces an upper triangular matrix. Some functions needed to be changed to only operate on triangular part if it is specified on input matrix. However I expect no or minimal benefits to performance.

Wouldn't it be a factor of 2?  Or does it defeat built-in CPU vectorization?

The bigger deal is in autodiff, where those extra operations go on the autodiff stack and then add unnecessary work to the reverse pass of autodiff.

t4c1:
> Wouldn't it be a factor of 2? Or does it defeat built-in CPU vectorization?

Im not sure what you mean with defeat built-in CPU vectorization. I think you probably mixed something up, as we are discussing OpenCL code.

 I did not measure speedup. It might be up to 2. These simple kernels are not very optimized anyway, as they are not expected to bottlenecks in many scenarios. In case they are it is usually better to combine addition with as much other operations as possible into a single kernel. This reduces the amount data that needs to be transfered to/from relatively slow global GPU memory.

bob-carpenter:
I should've said GPU vectorization.  The reason I wonder is that the memory in a triangular matrix isn't contiguous in Eigen's representation. It uses just the upper or lower triangular part of a dense matrix.

I have it on good authority that minimizing back and forth to non-GPU memory is critical.  The Tensorflow Probability devs are trying to put everything on the GPU so you never have to leave the GPU.

t4c1:
>  Or does it defeat built-in GPU vectorization?

No, our implementation still launches threads for elements that are out of selected triangular parts. These threads just write zeros to appropriate elements of result matrix.

>I have it on good authority that minimizing back and forth to non-GPU memory is critical.

That is true. But I don't understand how it is relevant for this PR.

SteveBronder:
For clarification on this PR, this is a style change (but one that sorely needs done).

Right now if we want to call the kernels for multiply that accounts for triangularity we have to call

```cpp
matrix_cl C = multiply<triangular_view_cl::UPPER,
                       triangular_view_cl::LOWER>(A, B);
```

But with this PR we will be able to call `matrix_cl C = A * B`

Until the Stan language is very definite on how to generally handle sparse types (something I think I'll be working on with Dan Simpson and Aki starting July) eod we still treat this as a dense matrix with some cleverness taking into account the triangularity

> I have it on good authority that minimizing back and forth to non-GPU memory is critical.  The Tensorflow Probability devs are trying to put everything on the GPU so you never have to leave the GPU.

Yes def, I've been racking over how to best keep things over there for a while. The best thing imo is that, when we move to Stan 3, Sean and I have talked about doing this within the compiler (i.e. building signatures that take in `matrix_cl` objects when doing rev. That and moving over to OpenCL 2.0 to get non-zero copy buffers for CPUs and APUs (Devices where the CPU and GPU are embedded in one another).

But that's a different PR, this one is to make the API much friendlier :)

bob-carpenter:
> I have it on good authority that minimizing back and forth to non-GPU memory is critical.
> 
> That is true. But I don't understand how it is relevant for this PR.

You're right---not relevant for this PR.

stan/math/prim/mat/fun/mdivide_left_tri.hpp
```diff
@@ -91,14 +91,10 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
 #ifdef STAN_OPENCL
   if (A.rows()
       >= opencl_context.tuning_opts().tri_inverse_size_worth_transfer) {
-    matrix_cl A_cl(A);
+    matrix_cl A_cl(A, TriView == Eigen::Lower ? TriangularViewCL::Lower
```
rok-cesnovar: This is much nicer to read than the separate calls, thanks. 
Do you think making a "conversion function" would be viable here, something like https://codereview.stackexchange.com/a/24199

On paper this seems nice for readability reasons. So the call would be matrix_cl A_cl(A, Eigen::Lower) and it would be converted to the TriangularViewCL enum.

Could we actually extract the view type inside the constructor from the Eigen matrix? For instance we could check if the matrix is Eigen::Lower or Eigen::Upper? Or could that cause any issues? We would have the option of specifying an override if needed. Do you think this would be of any value?

stan/math/opencl/err/check_triangular.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * triangular.
+ *
+ * @param function Function name (for error messages)
+ * @param name Variable name (for error messages)
+ * @param A <code>matrix_cl</code> to test
+ *
+ * @throw <code>std::invalid_argument</code> if the <code>matrix_cl</code>
+ *    is not triangular
+ */
+inline void check_triangular(const char* function, const char* name,
```
rok-cesnovar: Thanks for separating this out. This will be very useful in the future.

stan/math/opencl/kernels/matrix_multiply.hpp
```diff
@@ -79,13 +79,13 @@ static const char* matrix_multiply_kernel_code = STRINGIFY(
       // is 0 and the end tile is num_tiles-1 which
       // is then a general matrix multiply
       const int end_tile_A
-          = lower_upper_A == LOWER ? (i / THREAD_BLOCK_SIZE) : (num_tiles - 1);
+          = lower_upper_A & UPPER ? (num_tiles - 1) : (i / THREAD_BLOCK_SIZE);
```
rok-cesnovar: Is the performance gain of this non-negligible? If not, we should not use this, otherwise we need to somehow prevent developers from changing enum values.

test/unit/math/opencl/multiply_test.cpp
```diff
@@ -202,64 +202,64 @@ TEST(MathMatrix, matrix_vector_tri_small) {
   stan::math::matrix_cl m0_cl(4, 1);
 
   m0 = m * v.triangularView<Eigen::Lower>();
-  m0_cl = stan::math::opencl::multiply<stan::math::TriangularViewCL::Entire,
-                                       stan::math::TriangularViewCL::Lower>(
-      m_cl, v_cl);
+  m_cl.triangular_view(stan::math::TriangularViewCL::Entire);
```
rok-cesnovar: The opencl/multiply test file is somewhat huge and I think with this change we could make some helper functions and than just call them with (size, tri_view_a, tri_view_b) to simplify this code. This might be is out of scope for this PR. I will open an issue for this if you feel this should be solved separately.

stan/math/opencl/triangular.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+inline TriangularViewCL operator|(TriangularViewCL a, TriangularViewCL b) {
+  typedef typename std::underlying_type<TriangularViewCL>::type underlying;
+  return static_cast<TriangularViewCL>(static_cast<underlying>(a)
+                                       | static_cast<underlying>(b));
+}
+
+inline TriangularViewCL operator|=(TriangularViewCL& a, TriangularViewCL b) {
+  a = a | b;
```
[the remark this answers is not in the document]
t4c1: Performance benefit is negligible. These operations are used next to kernel launches that have comparatively huge overhead.

Bitwise operations simplify the code quite a lot. Alternatively we need separate functions for combining two views (what `operator|` does now), creating a view from the part used in two views (what `operator&` does now) and checking whether a certian view has lower/upper triangular part (now done with combination of `operator&` and `static_cast<bool>()`). Last one is also needed in OpenCL kernels. The only thing I don't like here is the need for a cast when checking whether a view contains lower/upper part.

I don't see the problem with coupling of enum values with operators on that enum. It is obvious that if someone randomly changes a part of code things stop working. 

Does anybody else have a strong opinion on this? Maybe @bob-carpenter?

stan/math/prim/mat/fun/mdivide_left_tri.hpp
t4c1: conversion function seems like a good idea here. 

> Could we actually extract the view type inside the constructor from the Eigen matrix?

Eigen::Matrix does not contain information on triangular view. Alternatively we could change the function signature to accept Eigen::TriangularView. In this case the call will look like `mdivide_left_tri(A.triangularView<Eigen::Upper>, B)`

stan/math/opencl/kernels/matrix_multiply.hpp
t4c1: Did not check it, but I guess it is negligible. How would you prevent changes? Just add a coment that the values are coulpled to bitwise operations?

test/unit/math/opencl/multiply_test.cpp
t4c1: A separate PR is probably preferable.

bob-carpenter:
Nope, no strong opinion from me on this.

stan/math/opencl/triangular.hpp
```diff
@@ -0,0 +1,61 @@
...
+                                       & static_cast<underlying>(b));
+}
+
+inline TriangularViewCL operator&=(TriangularViewCL& a, TriangularViewCL b) {
+  a = a & b;
+  return a;
+}
+
+inline TriangularViewCL transpose(TriangularViewCL a) {
+  switch (a) {
```
SteveBronder: I'd rather have if statements here

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -297,13 +297,10 @@ class cholesky_opencl : public vari {
    * @param L_adj matrix of adjoints of L
    */
   inline void symbolic_rev(matrix_cl& L, matrix_cl& L_adj) {
-    L_adj = opencl::multiply<TriangularViewCL::Upper, TriangularViewCL::Entire>(
-        transpose(L), L_adj);
+    L_adj = transpose(L) * L_adj;
```
SteveBronder: Excellent!!

stan/math/opencl/sub_block.hpp
```diff
@@ -57,7 +57,35 @@ inline void matrix_cl::sub_block(const matrix_cl& A, size_t A_i, size_t A_j,
   } else {
     opencl_kernels::sub_block(cl::NDRange(nrows, ncols), A, *this, A_i, A_j,
                               this_i, this_j, nrows, ncols, A.rows(), A.cols(),
-                              this->rows(), this->cols(), triangular_view);
+                              this->rows(), this->cols(), A.triangular_view());
+  }
+  // calculation of extreme sub- and super- diagonal written
+  int diag_in_copy = A_i - A_j;
+  int copy_low = static_cast<bool>(A.triangular_view_ & TriangularViewCL::Lower)
```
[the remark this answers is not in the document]
t4c1: > Why not use A.triangular_view() here?

This is a method of matrix_cl so it can access private attribute. If for some reason going through getter is preferable I can change that.

> Why is this being cast to a bool and then implicitly an int?

It is not. bool is used in ternary operator on the next line.

stan/math/opencl/sub_block.hpp
```diff
@@ -57,7 +57,35 @@ inline void matrix_cl::sub_block(const matrix_cl& A, size_t A_i, size_t A_j,
...
+            : diag_in_copy;
+  int start = this_j - this_i;
+
+  if (start + copy_low < 0) {
+    triangular_view_ |= TriangularViewCL::Lower;
+  } else if (this_i <= 1 && this_j == 0 && nrows + this_i >= rows_
+             && ncols >= std::min(rows_, cols_) - 1
+             && !static_cast<bool>(A.triangular_view_
+                                   & TriangularViewCL::Lower)) {
+    triangular_view_ &= TriangularViewCL::Upper;
```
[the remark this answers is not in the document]
t4c1: ok

stan/math/opencl/cholesky_decompose.hpp
```diff
@@ -87,6 +86,7 @@ inline void cholesky_decompose(matrix_cl& A) {
   // copy L_22 into A's lower left hand corner
   cholesky_decompose(L_22);
   A.sub_block(L_22, 0, 0, block, block, block_subset, block_subset);
+  A.triangular_view(TriangularViewCL::Lower);
```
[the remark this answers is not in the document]
t4c1: We never explicitely set upper triangular part of A to zero . So A.triangular_view() is still Entire. (And just setting flag to Lower is simpler than setting all values in upper triangular part to zero)

stan/math/opencl/copy.hpp
```diff
@@ -100,12 +101,12 @@ inline Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> from_matrix_cl(
...
  * copies it to the std::vector.
  *
- * @tparam triangular_view the triangularity of the source matrix
  * @param src the flat triangular source matrix on the OpenCL device
  * @return the packed std::vector
+ * @throw <code>std::invalid_argument</code> if the matrix is not triangular
  */
-template <TriangularViewCL triangular_view>
 inline std::vector<double> packed_copy(const matrix_cl& src) {
+  check_triangular("packed_copy", "src", src);
```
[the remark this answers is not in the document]
t4c1: This function seems to only work with triangular matrices. I don't understand what exactley are you proposing here.

stan/math/opencl/kernels/add.hpp
```diff
@@ -19,17 +19,31 @@ static const char *add_kernel_code = STRINGIFY(
...
-                      __global double *B, unsigned int rows,
-                      unsigned int cols) {
+                      __global double *B, unsigned int rows, unsigned int cols,
+                      int part_A, int part_B) {
       const int i = get_global_id(0);
       const int j = get_global_id(1);
       if (i < rows && j < cols) {
-        C(i, j) = A(i, j) + B(i, j);
+        double a;
+        if ((!(part_A & LOWER) && j < i) || (!(part_A & UPPER) && j > i)) {
```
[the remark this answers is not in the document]
t4c1: I doubt it. Majority of time this kernel spends is for loading argument values from global memory and saving back the result. I guess actual addition and if are negligible.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -169,9 +175,11 @@ class matrix_cl {
 
   const cl::Buffer& buffer() const { return oclBuffer_; }
 
-  matrix_cl() : rows_(0), cols_(0) {}
+  matrix_cl()
+      : rows_(0), cols_(0), triangular_view_(TriangularViewCL::Diagonal) {}
```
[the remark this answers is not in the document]
t4c1: In most cases any triangularity value would be fine. But in some edge cases with matrix multiplication this is needed to get correct triangularity. Imagine A x B x C, where A is (0x0), created by this constructor, B is lower triangular (0x5) and C is lower triangular (5x5).

stan/math/opencl/triangular.hpp
t4c1: why?

bob-carpenter: These are just suggestions---@SteveBronder is still in charge of the review.

1.  Shouldn't this be a metaprogram as it's only returning a type?  Something like:

```
template <typename T>
struct tri_transpose {
};

template <>
struct tri_tranpose<Lower> {
  typedef Upper type;
};
...
```

and usage would be `typename tri_tranpose<TrinagularViewCl::Lower>::type`.

2.  I think we should extend this to the other view type pairs: `Upper, Lower`, `StrictlyUpper, StrictlyLower`, and `UnitUpper, UnitLower`.

3.  These programs need doxygen doc.

t4c1: 1. No. These funtions are called at runtime, so they can not be reworked into templates. 

2. Adding each of those would add a bit to complexity of manipulations of triangular views. Right now I see no need for them. They can be added at later time if some new functionality requires them.

3. Ok, I will add doxy.

bob-carpenter: For (1), what's the use case?  Specifically, when is the value of `a` not known at compile time?  I don't see how you could take one of these enums and plug it back in.

For (2), if you don't want to complete the function naturally, I would suggest throwing an exception rather than returning a clearly wrong value by default.  

For (3), thanks.  I think the problem posed by (2) will be clearer when you try to write doc.

t4c1: 1. The way things are implemented now verywhere. The most obvious case where making it static would be very hard is `.sub_block()`. The view of the result depends on view of the inputs and arguments - to which part of a matrix it is copying.

2. Now we don't even have enum values for `StrictlyUpper`, `StrictlyLower`, `UnitUpper`, `UnitLower`.

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
SteveBronder: Sorry, what I'm proposing here is that we make it work for `Entire` instead of throwing. So just an if where if it's `Entire` then we take the vector as is without doing any real unpacking etc.

t4c1:
There seems to be a problem with windows threading tests on Jenkins.

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
t4c1: That is just adding functionality we do not need. Actually I belefe throwing is preferable. I can't imagine a use case where someone would need to use `packed_copy()` instead of `from_matrix_cl()` for `Entire` matrix. I guess in vast majority if not all the cases that `Entire` matrix would be passed in `packed_copy()` that would be a bug. Even if we put all data from matrix in the `vector` the user would just have problems later, since he is expecting to get packed triangular matrix as a result.

stan/math/opencl/triangular.hpp
SteveBronder: > why?

It's mostly that we use if statements everywhere else. Though they are more efficient I personally don't like using switches because of the fall-through gotcha.

Bob for context, doing this at compile time is possible for pretty much every function besides `sub_block`. I would bet there is some weird algebra we could do at compile time to figure out the triangularity when inserting one matrix into another, but I thought about it for about 10 minutes and it started hurting my brain.

So instead of doing a mix of static and dynamic we chose to go dynamic. I'm still not against doing as much at compile time as we can

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
SteveBronder: > That is just adding functionality we do not need. Actually I believe throwing is preferable. I can't imagine a use case where someone would need to use packed_copy() instead of from_matrix_cl() for Entire matrix. 

I wouldn't think about it as "is there a use case we can think of" but "should we let users do this". I think if a user wants to do a vector -> matrix_cl or matrix_cl -> vector with `Entire` I don't see a reason we can't allow that.

> I guess in vast majority if not all the cases that Entire matrix would be passed in packed_copy() that would be a bug. 

If we had a check for entire then I'm not seeing how it would be a bug

> Even if we put all data from matrix in the vector the user would just have problems later, since he is expecting to get packed triangular matrix as a result.

One of the major points of this PR is that users don't have to worry about the triangularity. Based on their operations we deduce that for them. They just have to tell us when moving over to OpenCL if the matrix is triangular

stan/math/opencl/matrix_cl.hpp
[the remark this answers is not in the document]
SteveBronder: Sorry I'm not following the above
```
= (0x0) * (0x5) * (5x0)
= (0x5) * (5x0)
= (0x0)
```

Why would diagonal be needed as the default here?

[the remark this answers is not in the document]
t4c1: Sorry my previous post is a bit ambiguous. NxM is not meant as hex value. It is dimmensions of a matrix: N rows, M columns.

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
t4c1: > a user wants to do a vector -> matrix_cl or matrix_cl -> vector with Entire

Can you think of a use case where this is better or simpler of more convenient to the user than matrix_cl -> MatrixXd? 

Even if there is such a use case we should probably make a separate function that copies from `matrix_cl` to vector (and loses data on the shape of the matrix). I think of `packed_copy()` as a low level optimization that is only used in cases where one knows he has a triangular `matrix_cl` to get only nonzero elements back from GPU faster. In all other cases `from_matrix_cl()` should be used.

And most likely further processing of the data from `packed_copy()` is based on the assumption that this is packed form of triangular matrix. Which makes no sense if it is actually an entire matrix. Such as in `cholesky_opencl.chain()`, which is afaik the only place this function is used.

If we want to have just one function that does everything, than we can implement packed copying functionality within `˙from_matrix_cl()` for the case of triangular matrix. 

> One of the major points of this PR is that users don't have to worry about the triangularity. 

That holds true only as long as data is in `matrix_cl`. We are not changing anything outside it. So triangularity must be specified where user is copying data to GPU and if user does anything triangularity-specific after copying the data back from GPU he must know which triangularity he gets. If his assumption is wrong it is better if he gets a nice error.

SteveBronder:
@t4c1 hope you don't mind, I thought merging to the refactor would be wicked lame so I did it for you (since the conflicts were my fault anyway)

stan/math/opencl/kernels/sub_block.hpp
```diff
@@ -49,13 +49,21 @@ static const char *sub_block_kernel_code = STRINGIFY(
...
       int j = get_global_id(1);
-      if ((i + src_offset_i) < src_rows && (j + src_offset_j) < src_cols
-          && (i + dst_offset_i) < dst_rows && (j + dst_offset_j) < dst_cols) {
-        if ((triangular_view == LOWER && i >= j)
-            || (triangular_view == UPPER && i <= j)
-            || triangular_view == ENTIRE) {
-          dst((dst_offset_i + i), (dst_offset_j + j))
-              = src((src_offset_i + i), (src_offset_j + j));
+
+      int src_idx_i = i + src_offset_i;
```
SteveBronder: const

stan/math/opencl/sub_block.hpp
```diff
@@ -59,7 +59,40 @@ inline void matrix_cl<T, enable_if_arithmetic<T>>::sub_block(
   } else {
     opencl_kernels::sub_block(cl::NDRange(nrows, ncols), A, *this, A_i, A_j,
                               this_i, this_j, nrows, ncols, A.rows(), A.cols(),
-                              this->rows(), this->cols(), triangular_view);
+                              this->rows(), this->cols(), A.triangular_view());
+  }
+  // calculation of extreme sub- and super- diagonal written
+  int diag_in_copy = A_i - A_j;
```
SteveBronder: const

stan/math/opencl/sub_block.hpp
```diff
@@ -59,7 +59,40 @@ inline void matrix_cl<T, enable_if_arithmetic<T>>::sub_block(
...
+  int copy_low
+      = containsNonzeroPart(A.triangular_view_, TriangularViewCL::Lower)
```
SteveBronder: I would prefer underscores over CamelCase, or does CamelCase have precedence somewhere else in the stan library for this sort of function

stan/math/opencl/triangular.hpp
SteveBronder: Still gotta do this, Maybe just `SpecialViewsCL` or does Eigen have these sorted out already and we can just steal those?

stan/math/opencl/triangular.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * Combines two triangular views. Result is nonzero, where any of the inputs is
+ * nonzero.
+ * @param a first view
+ * @param b second view
+ * @return combined view
+ */
+inline TriangularViewCL combine(TriangularViewCL a, TriangularViewCL b) {
+  typedef typename std::underlying_type<TriangularViewCL>::type underlying;
+  return static_cast<TriangularViewCL>(static_cast<underlying>(a)
+                                       | static_cast<underlying>(b));
```
SteveBronder: Is your intent here to look at the two types? I'm not seeing why we don't just do some `if`s? These are defined as `int`s so I'm not really getting what the `|` is gonna do here

stan/math/opencl/triangular.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * Inverts a triangular view. Parts that are zero in the input become nonzero in
+ * output and vice versa.
+ * @param a view to invert
+ * @return inverted view
+ */
+inline TriangularViewCL invert(TriangularViewCL a) {
+  typedef typename std::underlying_type<TriangularViewCL>::type underlying;
+  return static_cast<TriangularViewCL>(
+      static_cast<underlying>(TriangularViewCL::Entire)
+      & ~static_cast<underlying>(a));
```
SteveBronder: What is the `~` of `Diagonal`?

stan/math/prim/mat/fun/mdivide_left_tri.hpp
```diff
@@ -91,14 +91,11 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
 #ifdef STAN_OPENCL
   if (A.rows()
       >= opencl_context.tuning_opts().tri_inverse_size_worth_transfer) {
-    matrix_cl<double> A_cl(A);
+    matrix_cl<double> A_cl(A, TriView == Eigen::Lower
```
SteveBronder: We do this in a few places, can we just have a function for it?

stan/math/prim/mat/fun/mdivide_left_tri.hpp
```diff
@@ -120,20 +117,16 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
  * @return x = A^-1 .
  * @throws std::domain_error if A is not square
  */
-template <int TriView, int R1, int C1>
+template <Eigen::UpLoType TriView, int R1, int C1>
```
SteveBronder: Nice!

SteveBronder:
@t4c1 can you checkout this PR? I made some changes on a branch that I think would make this good to go

https://github.com/bstatcomp/math/pull/14

t4c1:
I don't like the name change `TriangularViewCL` to `PartialViewCL`. I associate the word "partial" with partial derivatives which have nothing to do with matrix views. That could be confusing for anyone using this.

Instead I propose following changes to names:
`PartialViewCL` -> `MatrixCLView` (Since we are changing this we could also use snake_case instead of CamelCase)
`matrix_cl->partial_view_` -> `matrix_cl->view_`
`matrix_cl->partial_view()` -> `matrix_cl->view()`

What dou you think @SteveBronder @rok-cesnovar?

t4c1:
@SteveBronder Why did you change functions for manipulation of `TriangularViewCL` to `operator+` and `operator*`? You did essentially the same as I did at first when I used `operator|` and `operator&` - and you did not like it. Mine operators were only needed to call the same operation on the underlying type of `enum class TriangularViewCL`. Yours, however, are implemented by calling a different operation on the underlying type. That is extremely confusing!

If you changed your mind about operators I think you should revert that to `|` and `&` - the same as underlying operations. Otherwise I think functions should be used instead of operators. Using functions also has advantage that we can do same operations in c++ and OpenCL by calling same functions.

rok-cesnovar:
I agree with `view()`.

SteveBronder:
>  That could be confusing for anyone using this. 

Yes! Agree with the name changes Did not think of those names but they make a lot more sense.

> Why did you change functions for manipulation of TriangularViewCL to operator+ and operator*? You did essentially the same as I did at first when I used operator| and operator& - and you did not like it. Mine operators were only needed to call the same operation on the underlying type of enum class TriangularViewCL. Yours, however, are implemented by calling a different operation on the underlying type. That is extremely confusing!

tbh I was pretty confused by the bitshift operators `&` and `|` when I was looking at the code. Like `Entire = Upper & Lower` didn't make a lot of sense to me abstractly (and what was going on was not documented or tested). But then playing with it I realized `|` gave the type of the addition of two triangular matrices and `&` gave the output of multiplication of two matrices. I thought `Entire = Upper + Lower` conveyed what was going on a more abstract level that made sense so I changed them.

> If you changed your mind about operators I think you should revert that to | and & - the same as underlying operations. Otherwise I think functions should be used instead of operators. Using functions also has advantage that we can do same operations in c++ and OpenCL by calling same functions.

I'm fine with making them functions. How about just `union` for `|` and `intersection` for `&`? I think it makes sense that a union of upper and lower is an entire matrix while an intersection of upper and lower is just the diagonal.

stan/math/opencl/triangular.hpp
t4c1: I will try to explain this the long way. Hopefully it will include whatever is confusing here. In following explanation im prefixing binary numbers with b.

TriangularViewCL is a bitfield. Only the lowest two bits the underlying integral type are used - all other are always zero. The lowest bit carries information on whether upper triangular part of a matrix is zero (value 1) or nonzero (value 0). So Diagonal(0 = b00) and Upper (2 = b10) have this bit set to 0. Lower(1 = b01) and Entire(3 = b11) have this bit set to 1.

In the same way the second lowest bit carries the information about lower triangular part. Again 0 stands for any nonzero values and 1 stands for all zero. So Diagonal(0 = b00) and Lower(1 = b01) have this set to 0. Upper (2 = b10) and Entire(3 = b11) have this bit set to 1.

`|` is bitwise or operation. Each bit of the result is 1 that bit of at least one inputs is 1. `&` is bitwise and operation. Each bit of the result is 1 if that bit of both inputs is 1. `~` is bitwise negation. Each bit of the result is 1 if that bit of the input is 0.

Some examples: 
Entire | Lower = b11 | b01 = b11 = Entire
Upper | Lower = b10 | b01 = b11 = Entire
Lower | Diagonal = b01 | b00 = b01 = Lower
~Lower = ~ b01 = b10 = Upper
~Diagonal = ~ b00 = b11 = Entire
Lower & Upper = b01 & b10 = b00 = Diagonal

The last one even reads as english: If a matrix is lower triangular and upper triangular it is diagonal.

Since I have seen it written a few times that we should assume that whoever is reading the code is proficient in C++ I prefere it this way compared to many ifs.

t4c1:
Union and intersection are fine.

stan/math/opencl/triangular.hpp
SteveBronder: Thanks! After playing around with the code I get that now. I didn't know that C++ would treat `int`s as bitfields and assumed it would try to deduce them to boolean which was my original confusion.

With the above it makes sense, if you know the order of the `enum`. But without that info then `Upper & Lower` is still hard to read. The best way to put it is to imagine you are coming into the codebase with no knowledge and see `Upper | Lower`. You would have to go find the `enum` in order to deduce the output. 

I think if we change the `+` and `*` operators to `union` and `intersection` that will both convey the actual implementation and be more readable without knowing the underlying implementation. Does that work for you?

t4c1: Yes.

t4c1:
Union is keyword in c++. Any other suggestions?

SteveBronder:
`conjunction` and `disjunction` are in the standard library but I think are fine to use

t4c1:
`conjunction` and `disjunction` are already in stan. I tried `any` and `both`, but `any` is a function in OpenCL. I think `either` and `both` might be fine.

SteveBronder:
> conjunction and disjunction are already in stan.

Did it fail to compile? I think it's fine to have a method with the same name but for different types. The conjunction and disjuction funcs in stan rn are there just because the standard only defines them in C++17 and we use C++14

t4c1:
Yep, failed to compile. Existing ones in stan are structs (template metaprograms), which can't have same name as a function.

SteveBronder:
Gaahh my mistake. `either` and `both` are okay. How about `conjoint` and `disjoint`. I doesn't feel like `either(Upper, Lower)` and `both(Upper, Lower)` convey what the operation does

t4c1:
I agree `either` and `both` are not perfect. But I stil think `conjoint` and `disjoint` are worse. For example if a matrix is bot lower- and upper- triangular it is diagonal. Reads like english.

SteveBronder:
That's fair. Let's fix the test error and then this should be good to go!

SteveBronder:
@rok-cesnovar do you want to approve this as well? Think both reviewers need to approve before we can mege

rok-cesnovar:
I think you could also dismiss my review, idk. But anyways yes, this is good to go. Thanks both!

PR #1273, opened 2019-06-22
Use Eigen's aligned_malloc and aligned_free in stack_alloc

SteveBronder:
## Summary

There's been a [`FIXME`](https://github.com/stan-dev/math/blob/develop/stan/math/memory/stack_alloc.hpp#L36) in the `stack_alloc` class for a while now saying

```
// FIXME: enforce alignment
```

So, I started looking into how to enforce alignment and it turns out, not easy! I started reading Eigen's [`Memory.h`](https://github.com/eigenteam/eigen-git-mirror/blob/master/Eigen/src/Core/util/Memory.h) file to see how they do it. Lots of gotchas here and there.

So, I figured why don't we just use their stuff? Below are the cmdstan performance tests that just replace `malloc` and `free` with Eigen's `aligned_malloc` and `aligned_free`.

```
('stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan', 1.0)
('stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan', 0.94)
('stat_comp_benchmarks/benchmarks/arK/arK.stan', 1.0)
('stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan', 0.98)
('stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan', 1.02)
('stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan', 0.98)
('stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan', 0.98)
('performance.compilation', 1.0)
('stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan', 0.99)
('stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan', 0.98)
('stat_comp_benchmarks/benchmarks/sir/sir.stan', 1.06)
('stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan', 0.97)
('stat_comp_benchmarks/benchmarks/garch/garch.stan', 0.97)
('stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan', 1.07)
('stat_comp_benchmarks/benchmarks/arma/arma.stan', 1.08)
1.00239340085
```

For a few models this doesn't seem to matter much, but for the ones that have contiguous access like arma and sir it seems like this is good.  

The one major issue is that `aligned_malloc` and `aligned_free` sit in `Eigen::internal` and it's certainly not good to rely on stuff they don't expose. Would we want to rip this out of Eigen for ourselves? Or is okay to use this internal part of Eigen?

If we went the route of ripping this stuff out I'd like to take it all the way and actually write a `aligned_allocator` like Eigen does that uses Stan's local allocator for the standard containers. i.e. for the code in `sum_v_vari` I think having our own allocator type would let us do something like

```cpp
class sum_v_vari : public vari {
 protected:
  std::vector<vari, stan_local_allocator<vari>> v_;
  size_t length_;

  inline static double sum_of_val(const std::vector<var>& v) {
    double result = 0;
    for (auto x : v)
      result += x.val();
    return result;
  }

 public:
  explicit sum_v_vari(double value, vari** v, size_t length)
      : vari(value), v_(v), length_(length) {}

  explicit sum_v_vari(const std::vector<var>& v1)
      : vari(sum_of_val(v1)),
        v_(v1), // I think something like this would 'just work'??
        length_(v1.size()) {
  }
```

Though I have to guess someone has thought of this before. Is there a reason we never went this route?

## Tests

No new tests

## Side Effects

We now use Eigen's aligned_malloc and aligned_free. 

## Checklist

Steve Bronder

### DIFF
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -88,7 +88,7 @@ CXXFLAGS_SUNDIALS ?= -pipe
 # LDLIBS_MPI
 
 # Eigen Flags
-CXXFLAGS_EIGEN ?= -DEIGEN_MAX_ALIGN_BYTES=64
+CXXFLAGS_EIGEN ?= -DEIGEN_MAX_ALIGN_BYTES=32
 
 ################################################################################
 # Update compiler flags with operating system specific modifications
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -87,6 +87,9 @@ CXXFLAGS_SUNDIALS ?= -pipe
 # LDFLAGS_MPI
 # LDLIBS_MPI
 
+# Eigen Flags
+CXXFLAGS_EIGEN ?= -DEIGEN_MAX_ALIGN_BYTES=64
+
 ################################################################################
 # Update compiler flags with operating system specific modifications
 ##
--- a/stan/math/memory/stack_alloc.hpp
+++ b/stan/math/memory/stack_alloc.hpp
@@ -38,14 +38,6 @@ const size_t DEFAULT_INITIAL_NBYTES = 1 << 16;  // 64KB
 // big fun to inline, but only called twice
 inline char* eight_byte_aligned_malloc(size_t size) {
   char* ptr = static_cast<char*>(Eigen::internal::aligned_malloc(size));
-  if (!ptr)
-    return ptr;  // malloc failed to alloc
-  if (!is_aligned(ptr, 8U)) {
-    std::stringstream s;
-    s << "invalid alignment to 8 bytes, ptr="
-      << reinterpret_cast<uintptr_t>(ptr) << std::endl;
-    throw std::runtime_error(s.str());
-  }
   return ptr;
 }
 }  // namespace internal
--- a/stan/math/memory/stack_alloc.hpp
+++ b/stan/math/memory/stack_alloc.hpp
@@ -5,6 +5,7 @@
 //            is best we can do to get safe pointer casts to uints.
 #include <stdint.h>
 #include <stan/math/prim/meta.hpp>
+#include <Eigen/src/Core/util/Memory.h>
 #include <cstdlib>
 #include <cstddef>
 #include <sstream>
@@ -36,7 +37,7 @@ const size_t DEFAULT_INITIAL_NBYTES = 1 << 16;  // 64KB
 // FIXME: enforce alignment
 // big fun to inline, but only called twice
 inline char* eight_byte_aligned_malloc(size_t size) {
-  char* ptr = static_cast<char*>(malloc(size));
+  char* ptr = static_cast<char*>(Eigen::internal::aligned_malloc(size));
   if (!ptr)
     return ptr;  // malloc failed to alloc
   if (!is_aligned(ptr, 8U)) {
--- a/stan/math/memory/stack_alloc.hpp
+++ b/stan/math/memory/stack_alloc.hpp
@@ -145,7 +145,7 @@ class stack_alloc {
     // free ALL blocks
     for (auto& block : blocks_)
       if (block)
-        Eigen::internal::aligned_free(block);(block);
+        Eigen::internal::aligned_free(block);
   }
 
   /**
@@ -231,7 +231,7 @@ class stack_alloc {
     // frees all BUT the first (index 0) block
     for (size_t i = 1; i < blocks_.size(); ++i)
       if (blocks_[i])
-        Eigen::internal::aligned_free(block);(blocks_[i]);
+        Eigen::internal::aligned_free(blocks_[i]);
     sizes_.resize(1);
     blocks_.resize(1);
     recover_all();
--- a/stan/math/memory/stack_alloc.hpp
+++ b/stan/math/memory/stack_alloc.hpp
@@ -145,7 +145,7 @@ class stack_alloc {
     // free ALL blocks
     for (auto& block : blocks_)
       if (block)
-        free(block);
+        Eigen::internal::aligned_free(block);(block);
   }
 
   /**
@@ -231,7 +231,7 @@ class stack_alloc {
     // frees all BUT the first (index 0) block
     for (size_t i = 1; i < blocks_.size(); ++i)
       if (blocks_[i])
-        free(blocks_[i]);
+        Eigen::internal::aligned_free(block);(blocks_[i]);
     sizes_.resize(1);
     blocks_.resize(1);
     recover_all();


wds15:
Should we include the TBB I would attempt to refactor that piece of code using the TBB's memory pools. If we do more threading, then we need a scalable memory allocator to avoid the battle on the global heap by each thread.

Have you seen some speedups already with an aligned allocator?

bob-carpenter:
I think the doc's stale.

`eight_byte_aligned_malloc` includes this:

```
 if (!is_aligned(ptr, 8U)) {
    std::stringstream s;
    s << "invalid alignment to 8 bytes, ptr="
      << reinterpret_cast<uintptr_t>(ptr) << std::endl;
    throw std::runtime_error(s.str());
  }
```

The function `is_aligned` is defined as

```
template <typename T>
bool is_aligned(T* ptr, unsigned int bytes_aligned) {
  return (reinterpret_cast<uintptr_t>(ptr) % bytes_aligned) == 0U;
}
```

I'm not 100% sure this is the right way to check alignment.  I can't even remember where I got the idea to do it this way.

SteveBronder:
Yeah Bob that makes sense. If we add Eigen's stuff though I'm pretty sure we will know ahead of time that we are aligned so we don't have to make that check each allocation.

I only made this PR because I found it interesting that these two models saw a pretty nice improvement. This doesn't seem to effect any of the other models. Is the 7-8% improvement significant here or is that in the realm of random noise?

```
('stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan', 1.07)
('stat_comp_benchmarks/benchmarks/arma/arma.stan', 1.08)
```

bob-carpenter:
I'd say 7% is meaningful.  But you're not going to get 7% speedup by increasing malloc speed, as it
only happens a handful of times total in fitting models.  Once memory's allocated, we reuse it.  So we'd need to get to the bottom of what you're measuring before we could fix anything.

Does that alloc affect standard Eigen allocs?  Does it also affect std::vector and other allocs?  And our memory allocation (also via malloc).

SteveBronder:
> Does that alloc affect standard Eigen allocs?  Does it also affect std::vector and other allocs?  And our memory allocation (also via malloc).

The inclusion here just effects malloc's made by `stack_alloc`. I think the important question here is what alignment does this thing choose and how does that interact with both the models that saw a speedup. My guess is that it's choosing a larger alignment than what happens by default malloc. I can check when I get home.

SteveBronder:
After looking around I think I understand that my standard malloc just aligns to the largest value that we pass (i.e. a struct of two doubles aligns to 8 bytes). I think the speedup came from the fact that Eigen enforces 16 byte alignment. I'm just curious to add a flag for 32 byte alignment and see what happens. idt that's something we can add permanently, but maybe the real implication of this is that we should have 16 byte alignment. I think that makes sense because that sees no real loss on any of the speed tests but some decent gains on the two tests above.

bob-carpenter:
Speeding up basic memory would be fantastic.  Our malloc comes in big chunks and within those chunks, we don't maintain 16, much less 32-byte alignment.   I wonder where the speedup's coming from.  Any ideas?

SteveBronder:
I'll keep investigating this a bit. At this point my guess is gamma rays flipping the right bits in my memory.

Looking at each model that had interesting results
[`('stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan', 0.94)`](https://github.com/stan-dev/stat_comp_benchmarks/blob/64b6597cd17de77e91dcfb1a9977758407017d20/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan)

Only two mildly interesting things in this
1. access of first and second `mu` and `z` a
2. Only parameter here is `z`

```
  delta_var1 = square(z[1] - mu[1]) - sigma1 * sigma1;
  delta_var2 = square(z[2] - mu[2]) - sigma2 * sigma2;
  delta_corr = (z[1] - mu[1]) * (z[2] - mu[2]) / (sigma1 * sigma2) - rho;
```

[`('stat_comp_benchmarks/benchmarks/arma/arma.stan', 1.08)`](https://github.com/stan-dev/stat_comp_benchmarks/blob/64b6597cd17de77e91dcfb1a9977758407017d20/benchmarks/arma/arma.stan)
[`('stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan', 1.07)`](https://github.com/stan-dev/stat_comp_benchmarks/blob/64b6597cd17de77e91dcfb1a9977758407017d20/benchmarks/gp_regr/gen_gp_data.stan)
[`('stat_comp_benchmarks/benchmarks/sir/sir.stan', 1.06)`](https://github.com/stan-dev/stat_comp_benchmarks/blob/64b6597cd17de77e91dcfb1a9977758407017d20/benchmarks/sir/sir.stan)

These ones all have loops in them over a lot of params, just armchair'ing here, but since we often use vari's on the stack that could give us better mem reads when using vari's but worse reads when we have doubles on the stack.

But gen_gp_data has no params. So idk, to figure out what's going on here I think I'd need to dig pretty deep and run the models a few more times to make sure the above results were not just random noise.

Open to convo about this but I'm going to close it until I have a better answer

bob-carpenter:
> On Jun 25, 2019, at 10:48 AM, Steve Bronder <notifications@github.com> wrote:
> 
> I'll keep investigating this a bit. At this point my guess is gamma rays flipping the right bits in my memory.
> 
> Looking at each model that had interesting results
> ('stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan', 0.94)
> 
> Only two mildly interesting things in this
> 
> 	• access of first and second mu and z a
> 	• Only parameter here is z
>   delta_var1 = square(z[1] - mu[1]) - sigma1 * sigma1;
>   delta_var2 = square(z[2] - mu[2]) - sigma2 * sigma2;
>   delta_corr = (z[1] - mu[1]) * (z[2] - mu[2]) / (sigma1 * sigma2) - rho;
> 
> ('stat_comp_benchmarks/benchmarks/arma/arma.stan', 1.08)
> ('stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan', 1.07)
> ('stat_comp_benchmarks/benchmarks/sir/sir.stan', 1.06)
> 
> These ones all have loops in them over a lot of params, just armchair'ing here, but since we often use vari's on the stack that could give us better mem reads when using vari's but worse reads when we have doubles on the stack.

The higher numbers are slower, right?  I have no idea why 16-byte aligned allocation would slow these down other than perhaps using too much memory and hurting caching behavior.

Double values won't go on the Stan-allocated stack, but they will go into malloc-allocated containers.

> But gen_gp_data has no params. So idk, to figure out what's going on here I think I'd need to dig pretty deep and run the models a few more times to make sure the above results were not just random noise.

Definitely.

> Open to convo about this but I'm going to close it until I have a better answer

OK.  Is there a place to see exactly what change you made to get these results?

SteveBronder:
>The higher numbers are slower, right?  I have no idea why 16-byte aligned allocation would slow these down other than perhaps using too much memory and hurting caching behavior.

Higher numbers are faster so `(1 - above#)` is the percent increase

Yep!

I just used [cmdstan perf tests](https://github.com/stan-dev/performance-tests-cmdstan)

With

```
./compare-git-hashes.sh "stat_comp_benchmarks -j8 --runs 10 develop develop false bd8be8dddce31648ad8b00bd678a7e39fab4c7be
```

Issue #1276, opened 2019-06-28
Update to OpenCL 2.0


SteveBronder:
## Description

I want to update us to the OpenCL 2.0 headers. We are currently on version 1.2. This cleans up the OpenCL headers by just making it one and brings a few bug updates.

The slide below shows some of the new features we can start working on by using OpenCL 2.0

![](https://images.anandtech.com/doci/7161/OCL1.png)

The implications for each of the slides above

1. We can start focusing on supporting CPUs and APUs. Right now we _can_ use these, but doing stuff on the CPU/APU still requires a copy which is not cool. With Shared Virtual Memory (SVM) we can give ownership of the data to OpenCL without actually making a copy.

2. We can call kernels in kernels. Rok and I have discussed this previously and think we could actually fully do the cholesky on the device side without even making host side function calls to the device.

3. The generic address space thing doesn't help us very much tbh

#### Current Version:
v2.19.1

bob-carpenter:
Thanks for the update and summary.

(1) in particular looks huge.  Will (2) require compiler support?  If so, we may want to start thinking about what'll be required there---probably some kind of dataflow analysis.

SteveBronder:
Yes v pumped for (1)! (2) does require an SDK that supports OpenCL 2.0. There are some gotchas with that. For instance Mac Mojave dropped support for OpenCL since Apple wrote their own gpu language called metal. Tensorflow GPU doesn't run on Mojave either since it depends on cuda which also cannot be compiled

But setting the right flags in the above PR I think we can keep supporting Mac pre-mojave. If not, then we'll probably want to have an actual discussion in the math meeting about whether it's okay for the OpenCL side to drop mac support. Though we are fine on Windows (lol) and Linux which is a pretty decent part of our user base.

Another alternative for mac support (I think) is to have users install the Intel OpenCL SDK that supports 2.0. @seantalts would I be able to install that on the mac to see if that can work? We would run using the CPU on the mac, but idt that's like the worst thing ever

seantalts:
I think we should investigate how much of a win this is compared with any additional support costs we'd incur. Could you fill out this checklist in this issue? https://github.com/stan-dev/math/wiki/Checklist-for-adding-a-dependency-to-the-Math-library

SteveBronder:
^yeah def that's on my todo list before we merge this

seantalts:
[...]

PR #1277, opened 2019-06-28
Update to OpenCL 2.0

SteveBronder:
## Summary

This updates the OpenCL headers that allow for up to version 2.2 support. We use the flags `-DCL_HPP_TARGET_OPENCL_VERSION=200` so that our code base uses OpenCL 2.0 by default.

There are a few code changes to update us from some deprecated functions.

1. `to_size_t` was removed and instead we use `cl::array<size_t, 3>`
2. We no longer need to pass a pair to the OpenCL program to compile kernels (ex [here](https://github.com/stan-dev/math/compare/develop...bstatcomp:update/opencl-2.0?expand=1#diff-f48e30891ae00ab80f64217ec2e213bcL83))
3. `cl::make_kernel` was deprecated so we now use `KernelFunctor` when constructing kernels.
4. The OpenCL header is now just a single file and so I removed the `CL` folder in the dependency library folder. Now we just do `#include <cl.hpp>`

## Tests

All basic opencl tests pass

## Side Effects

We are not using any of the OpenCL 2.0 kernel features and so this should have no side effects for the kernels.

## Checklist

- [x] Math issue #1276

Steve Bronder
The Khronos Group Inc.


rok-cesnovar:
Can we somehow check what OpenCL version is supported on the test Mac machine?

SteveBronder:
Ughhh I totally forgot about metal and Mojave. When I get home I think I can set `CL_HPP_MINIMUM_OPENCL_VERSION=120` to let this compile with 1.2 and 2.0

rok-cesnovar:
I tried to google this awhile a go, but had not luck finding an anwser. On pre-Mojave OSx with an AMD/NVIDIA GPU, do the GPU OpenCL runtimes that come with the drivers support OpenCL 1.2.x or 2.0?

SteveBronder:
I'm pretty sure it's only 1.2 (mby even 1.1) that are supported on the pre-installed drivers before Mojave

SteveBronder:
Gonna close this until I write the dependency wiki

Issue #1221, opened 2019-05-01
Use OpenCL in prim & rev mdivide_left_tri and multiply


rok-cesnovar:
## Description

`mdivide_left_tri` and `multiply` are the next common bottleneck we can speedup with OpenCL. The `prim` and `rev` implementations of `mdivide_left_tri` are implemented with lower triangular inverses and matrix multiplications, both of which are already implemented & merged in Stan Math. Same goes for `multiply˙, where only the multiply and transpose kernels are needed. 

~~I am planning to first open the PR for prim and then rev afterwards.~~
Update 29.6.2019:
- `prim/mdivide_left_tri` has since been merged
- `prim/multiply` PR is open
- `rev/multiply` PR will go up once the prim is merged
- `rev/mdivide_left_tri` will wait for #1266

Update 7.8.2019:
- `prim/mdivide_left_tri` was merged
- `prim/multiply` PR was merged
- `prim/mdivide_right_tri` was merged
- `rev/mdivide_left_tri` is open
- `rev/multiply` is open

## Example

If the input size is large enough, OpenCL should be used to speedup mdivide_left_tri. No changes to the API from both the Stan Math user and stanc. Same goes for multiply.

## Expected Output

Faster `mdivide_left_tri` and `multiply`

#### Current Version:
v2.19.1

PR #1278, opened 2019-06-29
Feature/1221 OpenCL primitive multiply

rok-cesnovar:
## Summary

This PR enables the use of the OpenCL multiplication in prim/multiply. It will speed up the `matrix` x `matrix` and `row vector` x `matrix` multiplications. The `row vector` x `vector` and scalar x matrix multiplications that are dealt with separately in prim/multiply will not use OpenCL. For those cases only the high-end GPU beats out an i7, and even then only with a speedup of 1.3x.

The code and tests are done. The only missing thing is setting the correct thresholds for when to switch over to the GPU/OpenCL device.

## Tests

There is only one new test that checks that the call to `stan::math::multiply(m1, m2)` returns the equivalent result when the multiplication is done on the CPU or the OpenCL device. The OpenCL multiplication is tested extensively in `test/unit/math/opencl/mulitiply` already.

I am re-running the benchmarks on the Radeon VII and the Nvidia M60. Will post the benchmarks that will be used to set the thresholds and provide some idea of the expected speedups.

## Side Effects

There are no potential side effect when not using OpenCL.

When `STAN_OPENCL` flag is set, we need to make sure that the input sizes are large enough that the OpenCL device is faster than the CPU. For now this is done using conservative thresholds for when to switch over to OpenCL.

## Checklist

- [x] Math issue #1221 


wds15:
@serban-nicusor-toptal same windows problem here - probably everywhere.

serban-nicusor-toptal:
Investigating, idk what may have caused this.

rok-cesnovar:
This PR is missing the performance measurements that I will have to redo as @t4c1 has some additional matrix multiplication optimizations that speedup the kernels up to 5x for some cases.

rok-cesnovar:
This PR is only missing the performance tests that will help us set the limits for the CPU/GPU switch. @t4c1 has made additional optimization for matrix multiplication that will speedup some cases and I will wait until that PR is merged. This is not going to make it to 2.20 anyways.

rok-cesnovar:
Tadej mentioned that his changes will not effect the CPU/GPU breakpoints so here are the speedups at small sizes. I played around with a bunch of different sized matrices, squared and rectangular and the input variable for the decision when to switch to the GPU should just be `NxMxK`. 

Rplot

At these small sizes there isnt a lot of differences between high and low-end GPUs. My suggestion for the rule is `NxMxK > 2000000`, just to be safe. If you agree @SteveBronder I am going to make the change tomorrow and then its ready for review.

These are google benchmark timings, made with [this](https://github.com/stan-dev/math/files/3415026/mat_mul_cpp.txt) code:
[mat_mul_breakpoint_i7_vii.txt](https://github.com/stan-dev/math/files/3414997/mat_mul_breakpoint.txt)
[mat_mul_breakpoint_r9.txt](https://github.com/stan-dev/math/files/3414998/mat_mul_breakpoint_r9.txt)

The VII speedup tops at around 65 for NxMxK = 8192^3. The R9 Fury speedup ceiling is around 35. I will do a more thorough performance test once the rev multiply is merged.

SteveBronder:
sounds good!

rok-cesnovar:
Finnaly, Jenkins caved in. @SteveBronder this is finally ready to review.

stan/math/prim/mat/fun/multiply.hpp
```diff
@@ -4,6 +4,9 @@
 #include <stan/math/prim/mat/fun/Eigen.hpp>
 #include <stan/math/prim/arr/err/check_matching_sizes.hpp>
 #include <stan/math/prim/mat/err/check_multiplicable.hpp>
+#ifdef STAN_OPENCL
+#include <stan/math/opencl/multiply.hpp>
```
SteveBronder: Do you think it would be better to have `prim` and `rev` just include from opencl.hpp? Kind of like how we are trying to zip up the rest of stan it may be good to just have one include for everything

stan/math/prim/mat/fun/multiply.hpp
```diff
@@ -54,7 +57,19 @@ inline Eigen::Matrix<double, R1, C2> multiply(
...
+  if (m1.rows() * m1.cols() * m2.cols()
+      > opencl_context.tuning_opts().multiply_dim_prod_worth_transfer) {
+    matrix_cl<double> m1_cl(m1);
+    matrix_cl<double> m2_cl(m2);
+    matrix_cl<double> m3_cl = m1_cl * m2_cl;
+    return from_matrix_cl(m3_cl);
+  } else {
+#endif
+    return m1 * m2;
+#ifdef STAN_OPENCL
```
SteveBronder: Let's just make this one ifdef.

On a sidenote, been thinking about how to get rid of these ifdefs. A thing we could do is 

1. In the `opencl_context.hpp` have an `#ifundef STAN_OPENCL`, then define empty classes / methods
2. Define something like `#define STAN_USE_OPENCL false`
3. For if statements like the one above have `... & STAN_USE_OPENCL` at the end so the compiler sees the above will always evaluate to false

Then we can get rid of all these `#ifdef`s. Does the above make sense?

test/unit/math/prim/mat/fun/multiply_test.cpp
```diff
@@ -133,3 +133,31 @@ TEST(AgradRevMatrix, multiply_vector_int) {
...
+
+#ifdef STAN_OPENCL
+#define EXPECT_MATRIX_NEAR(A, B, DELTA) \
+  for (int i = 0; i < A.size(); i++)    \
+    EXPECT_NEAR(A(i), B(i), DELTA);
+
+TEST(MathMatrix, multiply_opencl) {
+  int multiply_dim_prod_worth_transfer
+      = stan::math::opencl_context.tuning_opts()
+            .multiply_dim_prod_worth_transfer;
```
SteveBronder: Why is this here?

stan/math/prim/mat/fun/multiply.hpp
rok-cesnovar: Oh yes, absolutely, great catch. Everything outside of `/opencl` should only include that. Will also extend the `test-math-dependencies`.

stan/math/prim/mat/fun/multiply.hpp
rok-cesnovar: With 1 ifdef it looks like this? Is that how you meant it. 

```cpp
#ifdef STAN_OPENCL
  if (m1.rows() * m1.cols() * m2.cols()
      > opencl_context.tuning_opts().multiply_dim_prod_worth_transfer) {
    matrix_cl<double> m1_cl(m1);
    matrix_cl<double> m2_cl(m2);
    matrix_cl<double> m3_cl = m1_cl * m2_cl;
    return from_matrix_cl(m3_cl);
  }
#endif
  return m1 * m2;
```

Yes, we really need a solution for this. On the mdivide and multiply rev branches its only more disgusting :)

test/unit/math/prim/mat/fun/multiply_test.cpp
rok-cesnovar: goof. Used to be two parameter and I replaced them real quick.

rok-cesnovar: Ignore the above comment.

Its there so we set the tuning parameter back to the original value that it was before the test. So that the next test case (there isnt one currently, but if someone does add one) doesnt have to manually re-set it. 

simple case:
```cpp
  stan::math::matrix_d m3 = multiply(m1, m2);

  EXPECT_MATRIX_NEAR(m3_cl, m3, 1e-10);
  stan::math::opencl_context.tuning_opts().multiply_dim_prod_worth_transfer
      = multiply_dim_prod_worth_transfer;
  stan::math::opencl_context.tuning_opts().multiply_dim_prod_worth_transfer
      = 42;
}

TEST(MathMatrix, multiply_opencl1) {
  // this will output 42 which is bad
  std::cout << stan::math::opencl_context.tuning_opts().multiply_dim_prod_worth_transfer << std::endl;
}
```

stan/math/prim/mat/fun/multiply.hpp
rok-cesnovar: I like your solution. I can take a crack at that if you are busy or want to finish the matrix_cl rev code.

SteveBronder: 1 ifdef like this

```cpp
#ifdef STAN_OPENCL
  if (m1.rows() * m1.cols() * m2.cols()
      > opencl_context.tuning_opts().multiply_dim_prod_worth_transfer) {
    matrix_cl<double> m1_cl(m1);
    matrix_cl<double> m2_cl(m2);
    matrix_cl<double> m3_cl = m1_cl * m2_cl;
    return from_matrix_cl(m3_cl);
  } else {
    return m1 * m2;
  }
#else
return m1 * m2;
#endif
```

duplicates code but gets ride of the ifdef. imo just prefer less ifdefs

SteveBronder: > I like your solution. I can take a crack at that if you are busy or want to finish the matrix_cl rev code.

Yeah if you have time that would be rad!

rok-cesnovar: Will do!

rok-cesnovar:
@SteveBronder ready to re-review. I bailed on adding the check to `test-math-dependencies` since it would require adding more make tricks again (it isnt as simple as the meta check) and redoing the test-math-dependencies to python is on my to-do list for the near future.

rok-cesnovar:
Thanks Steve!

syclik:
Woo hoo!

Issue #1280, opened 2019-07-11
Add template for matrix_cl types


SteveBronder:
## Description

With #1197 we can now access Eigen matrices of var's and fvars via `.val()`, `.adj()`, and `.d()`. I'd like that for the OpenCL matrices as well. This PR starts lining us up for that by templating `matrix_cl` and adding a template specialization for `double`. The next PR will add a `var` and `vari` template specialization then another for `fvar`.

If we wanted to support floats we could change the template specialization to something like 
`template <typename T, typename = enable_if_t<is_arithmetic<T>::value>`.

Their will be three PRs, in the first PR nothing will change besides having to template matrix_cl with `<double>`. In the second PR we will make a template specialization for `var` and `vari` such that users can call the var and adjoint parts of the matrix with `A.val()` and `A.adj()`.

The template specializations will be such that

1. `double`: The current matrix_cl type we have now
2. `var`: Has two `matrix_cl<double>`'s inside of it for `val` and `adj`
3. `fvar` Has two `matrix_cl<T>` types in it for `val` and `d` which can be either `matrix_cl<fvar<T>>`, `matrix_cl<var>`, or `matrix_cl<double>`
## Example

In `cholesky_decompose` for reverse mode the code will change to

```cpp
...
      R.adj().sub_block(L.adj(), j, 0, 0, 0, k_j_ind, j);
      D.adj().sub_block(L.adj(), j, j, 0, 0, k_j_ind, k_j_ind);
      B.adj().sub_block(L.adj(), k, 0, 0, 0, m_k_ind, j);
      C.adj().sub_block(L.adj(), k, j, 0, 0, m_k_ind, k_j_ind);

      C.adj()
          = opencl::multiply<TriangularViewCL::Entire, TriangularViewCL::Lower>(
              C.adj(), tri_inverse<TriangularViewCL::Lower>(D.val()));
      B.adj() = B.adj() - C.adj() * R;
      D.adj() = D.adj() - transpose(C.adj()) * C;

// Was previously       symbolic_rev(D, D_adj);
      symbolic_rev(D);

      R.adj() = R.adj() - transpose(C.adj()) * B.val() - D.adj() * R.val();
      D.adj() = diagonal_multiply(D.adj(), 0.5);
      D.adj().zeros<TriangularViewCL::Upper>();

      L.adj().sub_block(R.adj(), 0, 0, j, 0, k_j_ind, j);
[...]

PR #1281, opened 2019-07-11
Add a double template to matrix_cl

SteveBronder:
## Summary

This pull requests 

1. Adds templates to matrix_cl and it's methods to allow for arithmetic types.
2. Adds a new `enable_if_arithmetic` meta alias to turn templates on for arithmetic types via 

```cpp
template <typename T, typename = enable_if_arithmetic<T>>
```

3. Adds a `CallArgs` variadic template to the `kernel_cl` `operator()`. This was necessary as the previous method we depended on had only the `(in/out/in_out)_buffer` types which had no knowledge of the template type of the `matrix_cl<T>` object.

For background, With #1197 we can now access Eigen matrices of var's and fvars via `.val()`, `.adj()`, and `.d()`. I'd like that for the OpenCL matrices as well. This PR starts lining us up for that by templating `matrix_cl` and adding template specialization for `double`. The next PR will add a `var` and `vari` template specialization then another for `fvar`.

## Tests

No additional tests

## Checklist

- [x] Math issue #1280


SteveBronder:
I need to read over this one more time then I'll ping someone for review

andrjohns:
More of a general comment on the constructor than anything specific to this PR, but would it be difficult to generalise ```Matrix<double,R,C>``` to ```MatrixBase<Derived>```? That would make it simpler to pass ```Map<>``` and ```Block<>``` objects to ```matrix_cl```

SteveBronder:
> More of a general comment on the constructor than anything specific to this PR, but would it be difficult to generalise Matrix<double,R,C> to MatrixBase<Derived>? That would make it simpler to pass Map<> and Block<> objects to matrix_cl

Do you mean for the `matrix_cl` constructor from Eigen matrices? Yeah I think we can do that

andrjohns:
> Do you mean for the matrix_cl constructor from Eigen matrices? Yeah I think we can do that

Great! It might even be better to target ```EigenBase``` over ```MatrixBase```, so that it can be used with ```Array<>``` objects as well. The only difficulty is that you'll need more templating magic to separate instantiations for ```prim```/```fwd```/```rev```, but I've written a rough proof-of-concept for the templating [here](https://godbolt.org/z/AeKRxA)

Sorry about derailing this PR, just stuck in my head

SteveBronder:
> Great! It might even be better to target EigenBase over MatrixBase, so that it can be used with Array<> objects as well. The only difficulty is that you'll need more templating magic to separate instantiations for prim/fwd/rev, but I've written a rough proof-of-concept for the templating here

So one weird one here, the code was failing when I switched this to `MatrixBase<Derived>` because a matrix base is not guaranteed to have storage (aka a `.data()` method)

Looking at the stack overflow answer below it looks like I can get around this by using a `Ref`

https://stackoverflow.com/a/25099257

I have a version working on my laptop that I'll push up tonight!

SteveBronder:
Oddly I'm trying to use the same trick with `matrix_cl` and getting a error that sort of makes sense but not sure how to work around.

```cpp
template <typename T,
          typename = std::enable_if_t<std::is_arithmetic<T>::value>>
class matrix_cl {
...
// forward declare sub_block
  template <TriangularViewCL triangular_view = TriangularViewCL::Entire>
  void sub_block(const matrix_cl<T>& A, size_t A_i, size_t A_j, size_t this_i,
                 size_t this_j, size_t nrows, size_t ncols);
```

With the method defined later as
```cpp
 template <typename T,
           typename = std::enable_if_t<std::is_arithmetic<T>::value>>
template <TriangularViewCL triangular_view>
inline void matrix_cl<T>::sub_block(const matrix_cl<T>& A, size_t A_i,
                                         size_t A_j, size_t this_i,
                                         size_t this_j, size_t nrows,
                                         size_t ncols) try {
// do stuff
}
```

Gives the error 

```
In file included from ./stan/math/opencl/multiply.hpp:9:0,
                 from ./stan/math/prim/mat/fun/mdivide_left_tri.hpp:11,
                 from ./stan/math/prim/mat/fun/mdivide_left_tri_low.hpp:6,
                 from ./stan/math/prim/mat/fun/chol2inv.hpp:7,
                 from ./stan/math/prim/mat.hpp:68,
                 from ./stan/math/mix/mat.hpp:9,
                 from test/unit/math/opencl/matrix_cl_test.cpp:2:
./stan/math/opencl/sub_block.hpp:33:54: error: invalid use of incomplete type ‘class stan::math::matrix_cl<T>’
                                          size_t ncols) try {
```

Which kind of makes sense, but this does compile if I use just `template <typename T>`. I'll post this up tonight and fool around with it a bit more

SteveBronder:
@t4c1 tried that with no dice :-/ 

Just pushed up the version that works without the enable_if. @t4c1 @andrjohns if either of yinz have time to take a look at this it would be appreciated! I'd like to get this to work for all arithmetic types so we can do `double`, `int`, `float` etc.

After this works, I'm ruminating on some schemes to do some pseudo templating on the OpenCL kernels so we don't have to maintain a bunch of different versions

t4c1:
I may have time to look at this tommorow. Anyway I already implemented something similar when testing GLMs with floats. Feel free to check it out [here](https://github.com/bstatcomp/math/blob/gpu_glm_float/stan/math/opencl/matrix_cl.hpp).

andrjohns:
Took a bit of digging, but I figured out the template 'magic' required here. I got this to work by declaring a dummy base struct that ```matrix_cl``` was derived from:
```
template <typename T>
class matrix_cl_impl {};

template <typename T>
class matrix_cl : matrix_cl_impl<std::enable_if_t<std::is_arithmetic<T>::value>> {
```

SteveBronder:
🙏Feelin Blessed🙏

I was staring at @t4c1 's code and some crtp also came to mind

stan/math/opencl/add.hpp
```diff
@@ -4,6 +4,8 @@
 #include <stan/math/opencl/matrix_cl.hpp>
 #include <stan/math/opencl/kernels/add.hpp>
 #include <stan/math/opencl/err/check_matching_dims.hpp>
+#include <stan/math/prim/meta.hpp>
+
 #include <CL/cl.hpp>
```
rok-cesnovar: Not reviewing yet and not really related to this PR but can we remove these `<CL/cl.h> includes? I feel like it would be enough to just include it in the context file right?

SteveBronder: We could, it can def go out of some files but it's sort of a case of 'include what you use' in other files.

SteveBronder:
@rok-cesnovar also this is ready for review!

stan/math/opencl/add.hpp
rok-cesnovar: Lets leave if for now. Will review the PR sometime later today.

stan/math/prim/scal/meta/return_type.hpp
```diff
@@ -42,5 +42,8 @@ struct return_type<T> {
       double, typename scalar_type<T>::type>::type type;
 };
 
+template <typename... Args>
+using return_type_t = typename stan::return_type<Args...>::type;
```
rok-cesnovar: :heart_eyes:

stan/math/prim/scal/meta/enable_if_arithmetic.hpp
```diff
@@ -0,0 +1,12 @@
+#ifndef STAN_MATH_PRIM_SCAL_META_ENABLE_IF_ARITHMETIC_HPP
+#define STAN_MATH_PRIM_SCAL_META_ENABLE_IF_ARITHMETIC_HPP
+
+#include <type_traits>
+
+namespace stan {
+
+template <typename T>
+using enable_if_arithmetic = std::enable_if_t<std::is_arithmetic<T>::value>;
```
rok-cesnovar: Great! Doing this metaprogram was the first thing that I thought of when I started looking at the changes in this PR yesterday.

andrjohns: +1 on this metaprogram. Would it workable to have a variadic version, like ```enable_if_all_arithmetic``` (or something less wordy)? That could help condense some of the templating from:
```
template <typename T1, typename T2, typename = enable_if_arithmetic<T1>,
                  typename = enable_if_arithmetic<T2>>
```
to
```
template <typename T1, typename T2, typename = enable_if_all_arithmetic<T1, T2>>
```

SteveBronder: Yep! We can use `conjuction` and `disjunction` for `enable_if_all_<blah>` and `enable_if_any_<blah>`

stan/math/opencl/err/check_symmetric.hpp
```diff
@@ -22,18 +22,18 @@ namespace math {
  *    the matrix is not symmetric.
  */
 inline void check_symmetric(const char* function, const char* name,
-                            const matrix_cl& y) {
+                            const matrix_cl<double>& y) {
```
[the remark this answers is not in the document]
SteveBronder: Whups! Good catch

stan/math/opencl/cholesky_decompose.hpp
```diff
@@ -40,7 +41,8 @@ namespace math {
  * @throw std::domain_error if m is not
  *  positive definite (if m has more than 0 elements)
  */
-inline void cholesky_decompose(matrix_cl& A) {
+template <typename T, typename = enable_if_arithmetic<T>>
+inline void cholesky_decompose(matrix_cl<T>& A) {
```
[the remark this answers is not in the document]
SteveBronder: That's makes a lot of sense!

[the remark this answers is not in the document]
SteveBronder: I added a floating_point enable_if as well.

SteveBronder:
I got most of these review comments done and will rubber duck the docs tmrw

stan/math/opencl/buffer_types.hpp
```diff
@@ -62,27 +62,22 @@ struct to_matrix_cl {
...
 template <>
 struct to_matrix_cl<out_buffer> {
-  typedef matrix_cl type;
+  typedef matrix_cl<double> type;
 };
 
 template <>
 struct to_matrix_cl<in_out_buffer> {
-  typedef matrix_cl type;
+  typedef matrix_cl<double> type;
```
SteveBronder: I think I can delete these now? Need to double check

stan/math/opencl/matrix_cl.hpp
```diff
@@ -21,33 +22,41 @@
  */
 namespace stan {
 namespace math {
+
+// Dummy class to instantiate matrix_cl to enable for specific types.
+template <typename T, typename = void>
+class matrix_cl {};
+
```
SteveBronder: @andrjohns So we can actually get away with this. Was going through the C++ templates book section on template lookup rules / partial specializations.

Essentially, the compiler only goes with this default template if there is nothing else to go on. Then the below `matrix_cl` just becomes a standard template specialization that's preferred for anything arithmetic.


 I'm kind of tempted to get rid of the default constructor and have it throw and error like "This template class should never be constructed"

rok-cesnovar:
@SteveBronder tag me when you think its ready for another pass.

SteveBronder:
@rok-cesnovar go for it!

rok-cesnovar:
Once the Windows machine caves in, feel free to merge. Will try to restart if it fails again.

rok-cesnovar:
It might be the AWS instance acting up again. cc @serban-nicusor-toptal

SteveBronder:
Serban set me up with an aws windows instance I'm going to take a crack at figuring out what's going on tonight

Issue #1285, opened 2019-07-23
Remove some for loops from the jacobian calcs


SteveBronder:
## Description

Looking over the code for [fwd](https://github.com/stan-dev/math/blob/develop/stan/math/fwd/mat/functor/jacobian.hpp) and [rev](https://github.com/stan-dev/math/blob/develop/stan/math/rev/mat/functor/jacobian.hpp) Jacobian we have a loop of for loops that I think we can pull out of the larger loop or get rid of 

## Example

For instance at the line below I think we can just contruct the Eigen var matrix with x instead of looping into it.

https://github.com/stan-dev/math/blob/develop/stan/math/rev/mat/functor/jacobian.hpp#L22

From 
```cpp
    Matrix<var, Dynamic, 1> x_var(x.size());
    for (int k = 0; k < x.size(); ++k)
      x_var(k) = x(k);
```
to
```cpp
    Matrix<var, Dynamic, 1> x_var(x);
```

#### Current Version:
v2.20.0

PR #1286, opened 2019-07-23
Removes extra loops in Jacobian calculations

SteveBronder:
## Summary

This removes a few extra loops found in the Jacobian calculations. There were a few spots where we could both use the new Eigen views for var matrices and pulls out the gradient calculation for the first iteration of fx_var. We had an if statement that checked if the outer loop was on the first iteration at every iteration. this pulls out the first iteration and starts the loop at 1 instead of 0.

That seems fine? Though I've never touched this part of the library before. 

## Tests

No additional tests

## Side Effects

Hopefully no more than the functions already have

## Checklist

- [x] Math issue #1285


stan/math/fwd/mat/functor/jacobian.hpp
```diff
@@ -14,19 +14,21 @@ void jacobian(const F& f, const Eigen::Matrix<T, Eigen::Dynamic, 1>& x,
...
+    x_fvar(k) = fvar<T>(x(k), 0);
+  }
+  x_fvar(0) = fvar<T>(x(0), 1);
+  Matrix<fvar<T>, Dynamic, 1> fx_fvar = f(x_fvar);
+  fx = fx_fvar.val();
+  J.col(0) = fx_fvar.d();
+  const fvar<T> switch_fvar(0, 1);  // flips the tangents on and off
+  for (int i = 1; i < x.size(); ++i) {
+    x_fvar(i - 1) -= switch_fvar;
+    x_fvar(i) += switch_fvar;
```
SteveBronder: This is the one weird hacky place I'm not sure about

Before we were doing this construction inside of the loop each time
```cpp
x_fvar(k) = fvar<T>(x(k), i == k);
```

But now I just set the tangent to 0 at the very beginning and then in the larger loop use this little fvar as a little switch that flips on and off the tangents

stan/math/rev/mat/functor/jacobian.hpp
```diff
@@ -18,20 +18,17 @@ void jacobian(const F& f, const Eigen::Matrix<double, Eigen::Dynamic, 1>& x,
...
     fx.resize(fx_var.size());
-    for (int i = 0; i < fx_var.size(); ++i)
-      fx(i) = fx_var(i).val();
     J.resize(fx_var.size(), x.size());
-    for (int i = 0; i < fx_var.size(); ++i) {
-      if (i > 0)
-        set_zero_all_adjoints_nested();
+    fx = fx_var.val();
+    grad(fx_var(0).vi_);
+    J.row(0) = x_var.adj();
```
SteveBronder: Also this is done by rows, does anyone know how much we need to change to have this done by columns. I'm assuming this Eigen matrix is column major order so we'll get better speed from doing this by column

stan/math/fwd/mat/functor/jacobian.hpp
bob-carpenter: That should be much better on memory the way you're doing it.

stan/math/rev/mat/functor/jacobian.hpp
bob-carpenter: That's just how Jacobians are defined---each row is a gradient.

For reverse mode, it might be better to pay a single transposition cost at the end.  That is, instead of building up an N x M matrix row major, build up an M x N matrix column major, then do a single transpose at the end.  Given that you only hit each element once, it might not make a difference on modest size Jacobians.

Once we have forward-mode thoroughly tested, it might be faster to calculate Jacobians using that, which would also be naturally column major.

Getting rid of the extra test and just going from 1 probably helps.

SteveBronder: Word! That makes sense. I added the transposition at the end and do things by col. Idt it would matter that much either besides large matrices tho'.

SteveBronder:
> (stat_comp_benchmarks/benchmarks/garch/garch.stan, 1.11)

That look satisfying but I'm really now sure how that would be affected by this pr.

@Bob-carpenter are you cool to review this?

Issue #646, opened 2017-10-14
log_sum_exp and log_diff_exp boundaries


bob-carpenter:
#### Summary:

The boundary conditions are off for negative infinite values.  Currently, we have

```
log_sum_exp(log(0), log(0)) = nan
log_diff_exp(log(0), log(0)) = nan
log_diff_exp(log(1), log(1)) = nan
```

and they should all return negative infinity.  For example,

```
log_sum_exp(log(0), log(0))
  = log_sum_exp(-inf, -inf)
  = log(exp(-inf) + exp(-inf))
  = log(0 + 0)
  = -inf
```

Now some care is going to have to be taken to manage the derivatives.  We may not be able to chain rule through the resulting negative infinities for an expression like `exp(log_sum_exp(log(0), log(0)))`.  The more doc explaining what's going on in the math lib (and Stan doc itself), the better.

Originally reported by Dalton Hance on Discourse:

http://discourse.mc-stan.org/t/log-0-and-stan-arithmetic/2171

#### Current Version:
v2.17.0

aggarwalvinayak:
I would like to work on this issue. Can I be assigned this bug?

bob-carpenter:
@aggarwalvinayak  Thanks.  The first thing to do is update the issue to propose a solution---it wasn't clear to me what to do with derivatives when writing the issue.  Even if they blow up, it would be good to fix this to get the right values.

aggarwalvinayak:
@bob-carpenter Sure i would look into fixing this to get the right values and will also see if i can fix the derivative part too... This is my first time i am contributing to an open source project. So could you help me get started with it. i have cloned it and found the files containing the `log_sum_exp` and `log_diff_exp` declaration. what should i now look into to begin with fixing this

bob-carpenter:
Our process is described here:

https://github.com/stan-dev/stan/wiki/Developer-process-overview

It's fairly standard for a medium-sized open-source project.

I'm not sure the derivatives can be fixed.  That's an analytic question as much as a computational one.

aggarwalvinayak:
I did look into the implementation of the function but i dont think it can be improved any further. as the problem is not of the compiler not treating log(0) as -inf instead log(0) returns a -inf value and our `log_sum_exp` evaluates it as expected but it is unfortunately printed as `NaN` for the required `-inf` value. But i am not completely sure about it.. Could something else be responsible for this?

sakrejda:
[...]

PR #1290, opened 2019-07-26
Bugfix/646 log_sum_exp and log_diff_exp boundaries

nhuurre:
## Summary

Fixes issue #646 . Check for negative infinite arguments in log_sum_exp and log_diff_exp so that the correct result can be returned instead of NaN.
I did what I could for the derivatives but generally they just blow up at the boundary.

## Tests

Unit tests have been added.

## Side Effects

None.

## Checklist

- [X] Math issue #646 

- [X] Copyright holder: Niko Huurre


stan/math/fwd/scal/fun/log_diff_exp.hpp
```diff
@@ -13,8 +13,11 @@ namespace math {
 template <typename T>
 inline fvar<T> log_diff_exp(const fvar<T>& x1, const fvar<T>& x2) {
   using std::exp;
```
andrjohns: This wasn't introduced by this PR, but I don't see the need for the ```using std::exp;``` statements here

stan/math/fwd/scal/fun/log_diff_exp.hpp
```diff
@@ -23,16 +26,25 @@ inline fvar<T> log_diff_exp(const fvar<T>& x1, const fvar<T>& x2) {
 template <typename T1, typename T2>
 inline fvar<T2> log_diff_exp(const T1& x1, const fvar<T2>& x2) {
   using std::exp;
-  if (x1 <= x2.val_)
+  if (x1 <= x2.val_) {
+    if (x1 < INFTY && x1 == x2.val_)
+      return fvar<T2>(NEGATIVE_INFTY, x2.d_ * NEGATIVE_INFTY);
```
andrjohns: What's the logic of multiplying by negative infinity here? Is it so that the derivative will be positive or negative infinity depending on the sign of ```x2.d_```?

stan/math/prim/mat/fun/log_sum_exp.hpp
```diff
@@ -25,6 +26,8 @@ namespace math {
 template <int R, int C>
 double log_sum_exp(const Eigen::Matrix<double, R, C>& x) {
   const double max = x.maxCoeff();
+  if (!boost::math::isfinite(max))
```
andrjohns: I think the preference is to use standard library functions rather than boost where possible

stan/math/fwd/scal/fun/log_diff_exp.hpp
nhuurre: Yes. If x1==x2 then the result is negative infinite but if x2 is even a tiny bit smaller the result is finite so the derivative must be infinite and opposite in sign to the derivative of x2.
Also, if x2.d_ == 0 then it's ambiguous and NaN is the only possible derivative. Take for example `log_diff_exp(0, -exp(k*x))`. When x is very negative this is approximately `log(1-(1-exp(k*x)) == k*x` so the derivative should be k at `x==NEGATIVE_INFTY`. But exp "forgets" k and the autodiff sees only `log_diff_exp(0, fvar(0, 0))`.

andrjohns: That makes sense, but if the purpose is to determine the sign of INFTY, wouldn't it be more efficient to use the [copysign function](https://en.cppreference.com/w/cpp/numeric/math/copysign) rather than a multiplication?

nhuurre: I guess so, I'm not too familiar with C++. But that still leaves the NaN case and if we really care about efficiency it might be best to just ignore the derivatives at infinity, I can't really imagine those ever being useful in practice.

bob-carpenter: This is coming up in other PRs and my understanding is that @syclik wants the boundaries to behave properly.

Multiplication is probably faster than anything involving a branch as long as it doesn't involve autodiff.

andrjohns: That all sounds good to me

andrjohns:
I've approved the changes, it just looks like a merge conflict needs to be resolved first

nhuurre:
Dismissed, huh? That makes it sound so negative! Though I was wondering what would happen if I pushed to a PR that has already been approved.

andrjohns:
Github does have some passive-aggressive moments!

Issue #1292, opened 2019-07-30
Optimize OpenCL matrix multiplication for cases where one dimension is small


t4c1:
## Description
OpenCL matrix multiplication is inefficient in the case where matrix dimensions are not a multiple of local size. This is especially apparent if one dimension is significantly smaller than local size. This is due to splitting matrices into tiles to reduce amount of memory transfers needed. However the multiplication kernel always processes whole tile, even if most of it is outside matrix.

Extreme case where one dimension is 1 (row vector - matrix and vector matrix multiplication) were previously optimized by making separate kernels for these special cases (issue #1191 and PR #1192).

## Example
Multiplication of a matrix with dimensions 2x10000 and one with dimensions 10000x10000.

#### Current Version:
v2.20.0

PR #1293, opened 2019-07-30
OpenCL matrix multiplication optimizations

t4c1:
## Summary

Optimized matrix multiplication, especially for the case where one matrix dimension is small.

Mostly I changed code patterns like:

```
for(...){
  if(foo)
    bar()
  else
    baz()
}
```
into
```
if(foo)
  for(...)
    bar()
else
  for(...)
    baz()
```

I also changed splitting criterion. The problem is now split depending on the number of work groups launched and comput units on the device, so as to have at least a few work groups per compute unit.

## Tests
No changes to tests. OpenCL multiplication is already well tested.

## Side Effects
Nothing major. Kernel code for multiplication is significantly longer (by ~100 lines of code). This is close to copy-paste of existing code. I don't have any idea how to deduplicate it without writing most of it in macros, which would be harder to read.

## Checklist

- [x] Math issue #1292 


t4c1:
Here are comparisons of speed of new and existing code for multiplication a matrix of size NxM with one of size MxK:

| M | N | K | shapes | speedup |
| :-- | :-- | :-- | :-- | :-- | 
| 1 | 15000 | 15000 | ENTIRE * ENTIRE | 0.59* |
| 15000 | 1 | 15000 | ENTIRE * ENTIRE | 1.63 |
| 15000 | 15000 | 1 | ENTIRE * ENTIRE | 0.88* |
| 15000 | 15000 | 1 | LOWER * ENTIRE | 0.98* |
| 2 | 15000 | 15000 | ENTIRE * ENTIRE | 4.16 |
| 15000 | 2 | 15000 | ENTIRE * ENTIRE | 1.60 |
| 15000 | 15000 | 2 | ENTIRE * ENTIRE | 4.97 |
| 15000 | 15000 | 2 | LOWER * ENTIRE | 4.17 |
| 32 | 15000 | 15000 | ENTIRE * ENTIRE | 1.05 |
| 15000 | 32 | 15000 | ENTIRE * ENTIRE | 1.00 |
| 15000 | 15000 | 32 | ENTIRE * ENTIRE | 1.05 |
| 32 | 15000 | 15000 | LOWER * ENTIRE | 5.01 |
| 15000 | 32 | 15000 | LOWER * ENTIRE | 0.90 |
| 15000 | 15000 | 32 | LOWER * ENTIRE | 1.09 |
| 2000 | 2000 | 2000 | ENTIRE * ENTIRE | 1.03 |
| 32 | 32 | 15000 | LOWER * ENTIRE | 0.89 |

* These are comparisons of the new general kernel with specialized kernels for cases row vector x matrix and matrix x vector. Since specialized kernels are faster than new general kernel I am not removing them.

In summary: new kernel is up to 5 times faster than existing one and at worst 10% slower.

rok-cesnovar:
@SteveBronder I already went through these changes a week ago and I think they are great, but would still like another set of eyes here. So please review these when you have the chance.

I think this almost-duplicating is better than any sort of weird trickery that would be harder to read. The multiplication OpenCL stuff is now quite complex and with a bunch of special cases. I am thinking we should add a paragraph on them [here](https://github.com/stan-dev/math/wiki/OpenCL-Kernels). I can do that if you agree that its necessary.

SteveBronder:
That mat mul code is starting to look a little smelly, could we break up the methods inside to call functions for the particular methods?

t4c1:
I looked into into breaking it up with functions. However a function here would have a large number of parameters. That also does not look very good.

SteveBronder:
I'll take a read over these

SteveBronder:
I'll mess around with this today. Looking at the code I think there are a few places we could write functions to make the code readable. OpenCL inlines the functions anyway so their shouldn't be any performance loss from breaking multiply up into smaller functions

SteveBronder:
Weird that `test/unit/math/prim/mat/fun/get_base1_lhs_test.cpp` is failing

t4c1:
That is really wierd. All tests already passed on the same code. It seems they were restarted as I checked all checkboxes in the checklist.

rok-cesnovar:
Yes, changes to the top comment trigger another set of tests. Not sure if that is by design though.

t4c1:
@SteveBronder did you find any way to make the code prettier for this PR?

SteveBronder:
Give me till Sunday if I still haven't then I'll be fine without prettying it up

t4c1:
@SteveBronder Any progress here?

SteveBronder:
Nah sorry I'll cash out my opinion here. I don't like how it looks but skin in the game and all that

rok-cesnovar:
@SteveBronder can you give a bit more info on what you dont like. Just that its basically duplicating a lot of code? I know you are pretty busy, I just dont want this to look like we want to push suff through that should not go through.

Are fine we the general idea of the PR, recognize that its faster but feel that the duplication in the kernel is a bit much?

SteveBronder:
It's mostly the deepness of the if statements, but we also need those for doing the optimizations. I think my actual wuff with this is that we are in C98 world which we can't really do much about

SteveBronder:
I'm fine with the idea of the PR

rok-cesnovar:
Ok, thank you for the clarification. Yeah, living in the C98 world is a pain :/ I will go through it again, since its been some time since I last looked at this, to see if I have any more comments. I feel that we cant avoid the branching and have optimizations at the same, that is just the way it goes in OpenCL kernels. But will take a look at the other aspects. Is that cool beans with you?

SteveBronder:
Works for me!

rok-cesnovar:
@SteveBronder Not sure where to post this, but since we were talking about C98 being a pain for kernels, this might interest you: https://www.youtube.com/watch?v=8pJS3n4MITM

Its a talk on SYCL, not sure if you are familiar with it. Its a single source GPU programming solution kind of like our Stan Math system here, if you ignore the fact that the kernel is actually a string, which honestly you can with the way we set it up. They do support writing kernels as lambdas in C++. Thought the downside of their approach is that its not as performant as straight up OpenCL and you need specific compiler to compile it. We are getting close to writing kernels in C++ with Tadej's generator though :)

Issue #1297, opened 2019-08-05
Map of OpenCL macros should use std::string instead of const char*


SteveBronder:
## Description

As described in [this](https://discourse.mc-stan.org/t/opencl-work-per-thread-0/10057) discourse thread @alashworth found a bug when we grab the macros for the OpenCL kernels. We were using a map as `std::map<const char*, int>`, but when we tried updating new options in the kernels this would do a pointer comparison instead of a comparison of the non-null elements of the const char array.

## Example

[here](https://github.com/stan-dev/math/blob/develop/stan/math/opencl/kernels/inv_lower_tri_multiply.hpp#L131)

```cpp
const kernel_cl<in_buffer, out_buffer, int, int> inv_lower_tri_multiply(
    "inv_lower_tri_multiply",
    {thread_block_helpers, inv_lower_tri_multiply_kernel_code},
    {{"THREAD_BLOCK_SIZE", 32}, {"WORK_PER_THREAD", 8}});
```

The `{{"THREAD_BLOCK_SIZE", 32}, {"WORK_PER_THREAD", 8}}` is a `std::map<const char*, int>`. When we add / update these elements to the kernel's macros instead of looking up `"WORK_PER_THREAD"` the maps `find` method looks for the `const char*`s address. It doesn't see that and so adds the new map value even though what we actually wanted to do was just update the `"WORK_PER_THREAD"` key.

## Expected Output

No duplicates in OpenCL options map

#### Current Version:
v2.20.0

PR #1298, opened 2019-08-05
make key of map for opencl kernel options into a string

SteveBronder:
## Summary

As described in [this](https://discourse.mc-stan.org/t/opencl-work-per-thread-0/10057) discourse thread @alashworth found a bug when we grab the macros for the OpenCL kernels. We were using a map as `std::map<const char*, int>`, but when we tried updating new options in the kernels this would do a pointer comparison instead of a comparison of the non-null elements of the const char array.

This PR fixes this by just using a `string` instead of a `const char*` for the key

## Tests

Idk if it's necessary but we could add a test for the `opencl_context` where we check all the values of the keys internal elements are unique. But we use a std::string now so idt we need to.

## Checklist

- [x] Math issue #1297

### DIFF
--- a/stan/math/opencl/kernel_cl.hpp
+++ b/stan/math/opencl/kernel_cl.hpp
@@ -220,14 +220,14 @@ inline const std::vector<cl::Event> select_events(
  */
 inline auto compile_kernel(const char* name,
                            const std::vector<const char*>& sources,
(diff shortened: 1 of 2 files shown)


rok-cesnovar:
Thanks for acting quickly on this. Since we are switching to std::string there is no sense of making a tests for duplicates.

PR #1303, opened 2019-08-07
Feature/issue 1221 Use OpenCL in rev/mdivide_left_tri 

rok-cesnovar:
## Summary

This PR adds the use of existing OpenCL kernels in `rev/mdivide_left_tri `. This code has been ready for quite some time, but kept falling behind in the pipeline due to PRs that changed the API. 

The input size break point for when to switch from the CPU to the GPU is the same as for the `mdivide_left_tri<double>`.  It could probably be lower, but since its already pretty low (N=100) I think there is no need to complicate here.

Besides adding the use of OpenCL functions this PR adds 2 constructors for `matrix_cl`, one for `Eigen::Map` and the other for arrays of `double`.

## Tests

There are a few added tests. They compare the values and adjoints for `mdivide_left_tri<Lower>` and `mdivide_left_tri<Upper>` for all combinations of input in `rev/mdivide_left_tri`: (`var`, `double`), (`double`, `var`) and (`var`, `var`).

All tests follow the same scheme just the view and types differ:

```cpp
// force OpenCL
opencl_context.tuning_opts().tri_inverse_size_worth_transfer = MDIVIDE_OPENCL_OVERRIDE;
Cv_cl = mdivide_left_tri<Eigen::Lower>(Av, Bv);
Cv_cl(0, 0).grad();
// force CPU
opencl_context.tuning_opts().tri_inverse_size_worth_transfer = MDIVIDE_CPU_OVERRIDE;
Cv = mdivide_left_tri<Eigen::Lower>(Av, Bv);
Cv(0, 0).grad();
EXPECT_MATRIX_NEAR(Cv, Cv_cl, 1.0E-12);
EXPECT_MATRIX_NEAR(Cv.adj(), Cv_cl.adj(), 1.0E-12);
stan::math::recover_memory();
```

## Side Effects

/

## Checklist

- [x] Math issue #1221 


stan/math/opencl/matrix_cl.hpp
```diff
@@ -303,6 +303,40 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
...
+   *
+   * @tparam T type of data in the \c Eigen \c Matrix
+   * @param A the \c Eigen \c Map of the Eigen Matrix
+   * @param partial_view which part of the matrix is used
+   *
+   * @throw <code>std::system_error</code> if the
+   * matrices do not have matching dimensions
+   */
+  template <int R, int C>
+  explicit matrix_cl(Eigen::Map<Eigen::Matrix<T, R, C>>& A,
```
SteveBronder: I think we can just do an Eigen::MatrixBase<Derived> constructor here instead of a map constructor

rok-cesnovar: I tried that, but had issues then accessing `.data()` without doing stuff like deep copies so I resorted to a separate constructor. Will try again.

rok-cesnovar: I got it working with Eigen::Ref, is that fine with you?

So the signature of the constructor is now

```cpp
explicit matrix_cl(const Eigen::Ref<const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>>& A,
                     matrix_cl_view partial_view = matrix_cl_view::Entire)
      : rows_(A.rows()), cols_(A.cols()), view_(partial_view) {
```

SteveBronder: Nice! Yep that's cool with me

rok-cesnovar:
This is ready for review.

stan/math/rev/mat/fun/mdivide_left_tri.hpp
```diff
@@ -57,25 +57,54 @@ class mdivide_left_tri_vv_vari : public vari {
...
-
+    matrix_d adjA;
+    matrix_d adjB;
+
+#ifdef STAN_OPENCL
+    if (M_ >= opencl_context.tuning_opts().tri_inverse_size_worth_transfer) {
+      matrix_d tmp_variRefC = Map<matrix_vi>(variRefC_, M_, N_).adj();
+      matrix_cl<double> A_cl(A_, M_, M_, from_eigen_uplo_type(TriView));
+      matrix_cl<double> C_cl(C_, M_, N_);
+      matrix_cl<double> variRefC_cl(tmp_variRefC);
```
SteveBronder: We can get rid of the `tmp_variRefC` above
```suggestion
      matrix_cl<double> variRefC_cl(Map<matrix_vi>(variRefC_, M_, N_).adj());
```

SteveBronder: We can get rid of the `tmp_variRefC` above
```suggestion
      matrix_cl<double> variRefC_cl(Map<matrix_vi>(variRefC_, M_, N_).adj());
```

stan/math/rev/mat/fun/mdivide_left_tri.hpp
```diff
@@ -122,27 +151,52 @@ class mdivide_left_tri_dv_vari : public vari {
...
-    Map<matrix_vi>(variRefB_, M_, N_).adj()
-        += Map<matrix_d>(A_, M_, M_)
-               .template triangularView<TriView>()
-               .transpose()
-               .solve(Map<matrix_vi>(variRefC_, M_, N_).adj());
+#ifdef STAN_OPENCL
+    if (M_ >= opencl_context.tuning_opts().tri_inverse_size_worth_transfer) {
+      matrix_d tmp = Map<matrix_vi>(variRefC_, M_, N_).adj();
+      matrix_cl<double> A_cl(A_, M_, M_, from_eigen_uplo_type(TriView));
+      matrix_cl<double> C_cl(tmp);
```
SteveBronder: Remove tmp

```suggestion
      matrix_cl<double> C_cl(Map<matrix_vi>(variRefC_, M_, N_).adj());
```

stan/math/rev/mat/fun/mdivide_left_tri.hpp
```diff
@@ -122,27 +151,52 @@ class mdivide_left_tri_dv_vari : public vari {
...
+      A_cl = transpose(tri_inverse(A_cl));
+      matrix_cl<double> res_cl = A_cl * C_cl;
+      matrix_d ttmp = from_matrix_cl(res_cl);
+      Map<matrix_vi>(variRefB_, M_, N_).adj() += ttmp;
```
SteveBronder: Remove tmp
```suggestion
      Map<matrix_vi>(variRefB_, M_, N_).adj() += from_matrix_cl(res_cl)
```

rok-cesnovar:
Thanks for finding those. I removed those temporary matrices.

I had to however, revert the generalization of matrix_cl from Eigen. It seems that its not actually working as it should. Not sure if its `.eval()` or something else acting up. The develop failing was due to this and it was not a false positive.

I can address the general constructor in a separate issue and PR, if you agree. It certainly should be done. I also need to look at why the tests on jenkins for this PR passed before, because they should not and they failed locally. After the revert its green light on `python runTests.py test/unit -f opencl`.

SteveBronder:
That's odd, yeah we can handle that in a seperate PR. Odd though because it worked on machine 

Lemme give this a quick look and if tests pass we can merge!

PR #1305, opened 2019-08-08
Feature/issue 1221 Use OpenCL in rev/multiply

rok-cesnovar:
## Summary

This PR adds the use of existing OpenCL kernels in rev/multiply. With this PR we will close issue #1221 

The input size break point for when to switch from the CPU to the GPU is the same as for the `prim/multiply`. Same as in `mdivide_left_tri`, this could be set lower as the break point in prim is set for cases of A*B where we need to transfer 2 input matrices and the result matrix. In the rev/multiply there are also cases where we perform `A*B` and `A*C` where we transfer all three input matrices and both results. There is therefore less data transfers per multiplication here, but I think we can reuse the same limit here for the time being as.

This PR adds the same 2 constructors as the `rev/mdivide_left_tri` #1303

## Tests

There are a few added tests. They compare the values and adjoints of OpenCL and CPU `multiply()` for all combinations of input in `rev/multiply`: (`var`, `double`), (`double`, `var`) and (`var`, `var`).

## Side Effects

/

## Checklist

- [x] Math issue #1221 

- [x] Copyright holder: Rok Češnovar, University of Ljubljana


rok-cesnovar:
Thanks Steve for the changes and the reviews for both PRs! I know that you are pretty busy, so its really appreciated. 

I dont have any new PRs scheduled until after Stancon so I wont bother you with any new reviews for a week or two :). We also now have everything that we promised to deliver in `develop` before Stancon, which I think is nice :)

I am going to make an issue for separate varis, we should move them in the `/opencl/rev` folder, so that the OpenCL code in /prim and /rev etc is minimal. I would move all current opencl varis at once so that we are consistent (not have cholesky opencl vari in /rev other in /opencl).

rok-cesnovar:
Develop is failing after this merge, which is odd. I am looking into this. 

@seantalts @serban-nicusor-toptal are there any differences between the run of "Full unit with GPU" in the PR and after merge?

rok-cesnovar:
The "Full unit with GPU" was run with "OPENCL_DEVICE_ID=0", while all other recent develop tests were run with "OPENCL_DEVICE_ID=1". Does anyone have any info on the 2 devices and how to we determine which one is selected?

PR #1317, opened 2019-08-18
Add enable_if specializations

SteveBronder:
## Summary

This PR is part of a larger effort to make Stan's templates more general. This PR contains `enable_if` and `disable_if` aliases that we will use to specialize functions for vars, fvars, Eigen matrices, vectors, etc.

(Also it changes `to_var` so that in cases of a var matrix/vec coming in as input it also goes out via reference to the input. This stops a copy from happening and I figured it would be small enough that I could include it in this PR.

It looks like there is a lot here but I'd ballpark about 90% of it is tests and duplicated code wrt creating the different enabling functions. I'll tag the parts that I think need a look over 

## Tests

The enable_if methods are aliases so we check that the aliases go off for templated structs that use each of:

```
enable_if_all // all conditions are true
enable_if_any // any conditions are true
disable_if // The condition is false
disable_if_all // All conditions are false
disable_if_any // Any conditions are false
```

There are also tests that check whether the new is_* methods produce the correct true / false types

## Side Effects

None

## Checklist

- [x] Math issue #1308


stan/math/prim/arr/meta/is_vector.hpp
```diff
@@ -1,23 +1,45 @@
...
-//   extend to array/ptr types
+namespace internal {
+// FIXME(Steve): this is a stupid name
+template <typename T>
+using inherit_vector = std::vector<typename std::decay_t<T>::value_type,
+                                   typename std::decay_t<T>::allocator_type>;
+template <typename T>
+using is_same_vector
+    = std::is_same<typename std::decay_t<T>, internal::inherit_vector<T>>;
+}  // namespace internal
```
SteveBronder: This stuff is a bit wacky, the intent here is just to be able to verify that one vector is the same as another

stan/math/prim/arr/meta/scalar_type.hpp
```diff
@@ -2,27 +2,20 @@
...
 
+/**
+ * Return the underlying type from a standard vector.
+ * @param T The standard vector to extract the underlying type from.
+ */
 template <typename T>
-struct scalar_type<const std::vector<T> > {
-  typedef typename scalar_type<T>::type type;
+struct scalar_type<T, std::enable_if_t<is_std_vector<T>::value>> {
+  typedef typename scalar_type<typename T::value_type>::type type;
```
SteveBronder: fyi this uses `std::decay<>` so we don't have to have the const and ref versions of this functions. I like this more because we do the same things for const/refs etc.

stan/math/prim/mat/fun/accumulator.hpp
```diff
@@ -63,7 +63,7 @@ class accumulator {
    * @tparam S Type of argument
    * @param x Value to add
    */
-  template <typename S, typename = enable_if_not_arithmetic<S>,
+  template <typename S, typename = disable_if_arithmetic<S>,
```
SteveBronder: The standard veto'd `disable_if`s being in the standard but boost has them and imo I prefer `disable_if` to `enable_if_not`

stan/math/prim/mat/meta/is_vector.hpp
```diff
@@ -2,27 +2,55 @@
...
+template <class T, bool = is_eigen<std::decay_t<T>>::value>
+struct is_eigen_col_vector_impl
+    : std::integral_constant<bool, T::ColsAtCompileTime == 1> {};
 
+// not eigen type
+template <class T>
+struct is_eigen_col_vector_impl<T, false> : std::false_type {};
+}  // namespace internal
+
+// Check whether type is an eigen col vector
```
SteveBronder: We use an internal templated implementation here because if the preprocessor tries to access `ColsAtCompileTime` from a `T` of `double` it will throw. So this will only look for the columns at compile time member if the type is eigen

stan/math/prim/scal/meta/enable_if_generics.hpp
```diff
@@ -0,0 +1,810 @@
...
+using disable_if_all = std::enable_if_t<!math::disjunction<Checks...>::value>;
+
+/**
+ * Template metaprogram for deducing if a method should be remove a
+ * a function from overload resolution.
+ *
+ * Returns a type void if any of the conditions are false.
+ */
+template <class... Checks>
+using disable_if_any = std::enable_if_t<!math::conjunction<Checks...>::value>;
```
SteveBronder: All the stuff below here are just using the alias's above for specializations so we can write stuff like `enable_if_eigen_var<T>` etc.

stan/math/prim/scal/meta/enable_if_generics.hpp
```diff
@@ -0,0 +1,810 @@
...
+using disable_if_all_arithmetic = disable_if_all<is_arithmetic_decay<Types>...>;
+
+template <typename... Types>
+using disable_if_any_arithmetic = disable_if_any<is_arithmetic_decay<Types>...>;
+
+// Check if a type contains or is arithmetic
+template <typename T>
+struct is_contains_arithmetic
+    : std::integral_constant<
+          bool, std::is_arithmetic<scalar_type_t<std::decay_t<T>>>::value> {};
```
SteveBronder: All the structs here follow this structure where std::integral_constant is essentially `std::false_type` or `std::true_type`

test/unit/math/prim/arr/meta/is_vector_test.cpp
```diff
@@ -6,8 +6,22 @@ TEST(MetaTraits, is_vector) {
   using stan::is_vector;
   using std::vector;
 
-  EXPECT_TRUE(is_vector<std::vector<double> >::value);
-  EXPECT_TRUE(is_vector<std::vector<int> >::value);
-  EXPECT_TRUE(is_vector<std::vector<const double> >::value);
-  EXPECT_TRUE(is_vector<std::vector<const int> >::value);
```
SteveBronder: std::vectors can't have const types inside them since C++11

https://stackoverflow.com/a/17313104/2269255

SteveBronder:
fyi if the tests pass I think the only thing to really look over here are

1. The top part of `stan/math/prim/scal/meta/enable_if_generics.hpp ` as well as the structs in there.
2. Skimming over the new tests

SteveBronder:
Gonna make a smaller PR

PR #1318, opened 2019-08-26
Issue 1010 - Replace hand-coded math with standard library c++11 functions

andrjohns:
## Summary
This pull closes out issue #1010 by replacing the math functions with their std:: equivalents.

All functions, except for ```ldexp``` had existing implementations for ```fwd``` and ```rev```, so I added those (as well as their respective tests).

Once Bob's ad test framework (#1262) is merged, I'll update those tests.

## Tests
Autodiff tests for ```ldexp```, all others unchanged as existing behaviour should not change

## Side Effects
N/A

## Checklist

- [x] Math issue #1010 


stan/math/prim/scal/fun/hypot.hpp
```diff
@@ -23,8 +21,8 @@ namespace math {
  */
 template <typename T1, typename T2>
 inline return_type_t<T1, T2> hypot(const T1& x, const T2& y) {
-  using std::sqrt;
-  return sqrt(square(x) + square(y));
+  using std::hypot;
+  return hypot(x, y);
```
SteveBronder: just fyi, `std::hypot` is supposed to be a good bit slower. Though also more precise. Not sure which we'd rather go for in this situation

https://stackoverflow.com/questions/32435796/when-to-use-stdhypotx-y-over-stdsqrtxx-yy

stan/math/rev/scal/fun/fma.hpp
```diff
@@ -121,7 +121,8 @@ inline var fma(const var& a, const var& b, const var& c) {
  * @param c Summand.
  * @return Product of the multiplicands plus the summand, ($a * $b) + $c.
  */
-inline var fma(const var& a, const var& b, double c) {
+template <typename Tc, typename = enable_if_arithmetic<Tc>>
+inline var fma(const var& a, const var& b, Tc c) {
```
SteveBronder: Think below perfect forwarding is fine? Though idt it matters for an integral type

```suggestion
inline var fma(const var& a, const var& b, Tc&& c) {
```

stan/math/prim/scal/fun/hypot.hpp
andrjohns: Huh, I had no idea. I'll revert it for this pull but I'll look into different constructions of ```hypot```, since I think we could do better.

I found this page very interesting: https://deepai.org/publication/an-improved-algorithm-for-hypot-a-b

But of course they didn't report timings.

andrjohns: > Is there a situation where you see enough usage of hypot to be worth an optimization effort?

Mainly because it would be introducing a performance regression (in terms of speed, not precision) if we replaced ```sqrt(square(a) + square(b))``` with ```hypot(a, b)```. I'm not sure whether the speed cost would outweigh the precision benefits or whether we'd just end up having to revert it down the line

bob-carpenter:
I don't think it's a big deal either way as I don't think most users even know hypot() is there. I can't see it ever being a big chunk of computation (unlike, say lgamma, which dominates a lot of density calculations like the Dirichlet and gamma).  I can't imagine it's part of our performance regression tests anywhere.

I also wouldn't worry too much about reversion down the line.  We may come up with some principled decision on precision vs. speed, at which point the implementation of hypot() could go either way.

SteveBronder:
Word! I'll mark my comment as resolved

andrjohns:
Works for me!

stan/math/fwd/scal/fun/ldexp.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <stan/math/fwd/core.hpp>
+#include <stan/math/prim/scal/fun/ldexp.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/**
+ * Returns the product of a (the significand) and
```
SteveBronder:
```suggestion
 * Returns the product of a (the significand) times
```

stan/math/fwd/scal/fun/ldexp.hpp
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @tparam T Scalar type of significand
+ * @param[in] a the significand
+ * @param[in] b an integer that is the exponent
+ * @return product of a times 2 to the power b
+ */
+template <typename T>
+inline fvar<T> ldexp(const fvar<T>& a, int b) {
+  return fvar<T>(ldexp(a.val_, b), ldexp(a.d_, b));
+}
```
SteveBronder: Should their be an overload for fvar<T> being the exponent?

stan/math/prim/scal/fun/exp2.hpp
```diff
@@ -19,8 +19,8 @@ namespace math {
  * @return exponent base 2 of argument.
  */
 inline double exp2(double y) {
-  using std::pow;
-  return pow(2.0, y);
+  using std::exp2;
+  return exp2(y);
 }
```
SteveBronder: Should these just have a template T with an enable_if for arithmetic?
```cpp
template <typename T, typename = enable_if_arithmetic<T>>
inline double exp2(T&& y) {
  using std::exp2;
  return exp2(std::forward<T>(y));
}
```

stan/math/prim/scal/fun/fdim.hpp
```diff
@@ -23,11 +20,8 @@ namespace math {
  */
 template <typename T1, typename T2>
 inline return_type_t<T1, T2> fdim(T1 x, T2 y) {
-  typedef return_type_t<T1, T2> return_t;
-  using std::numeric_limits;
-  if (is_any_nan(x, y))
-    return NOT_A_NUMBER;
-  return (x <= y) ? 0 : x - y;
+  using std::fdim;
+  return fdim(x, y);
```
SteveBronder: Since there is a fdim method in var/fvar I think we can do the same templating trick here as well

SteveBronder: I'm pretty sure we can use auto as well since this works on scalars. @bob-carpenter is return_type<T> something from pre-C++11 or will auto deduction not work for these for some reason?

stan/math/prim/scal/fun/logical_lte.hpp
```diff
@@ -16,7 +15,7 @@ namespace math {
  * @return <code>true</code> iff <code>x1 &lt;= x2</code>
  */
 template <typename T1, typename T2>
-inline int logical_lte(const T1 x1, const T2 x2) {
+inline bool logical_lte(const T1 x1, const T2 x2) {
```
SteveBronder: Re-reading meyers C++11 book, wouldn't mind these being

```cpp
inline auto logical_lte(const T1 x1, const T2 x2) {
  return static_cast<bool>(x1 <= x2;);
}
```

stan/math/rev/scal/fun/fma.hpp
```diff
@@ -141,7 +142,8 @@ inline var fma(const var& a, const var& b, double c) {
  * @param c Summand.
  * @return Product of the multiplicands plus the summand, ($a * $b) + $c.
  */
-inline var fma(const var& a, double b, const var& c) {
+template <typename Tb, typename = enable_if_arithmetic<Tb>>
+inline var fma(const var& a, Tb b, const var& c) {
   return var(new internal::fma_vdv_vari(a.vi_, b, c.vi_));
 }
```
SteveBronder: I think something like think something like this would work
```cpp
template <typename Ta, typename Tb, typename Tc, 
 typename = enable_if_arithmetic<Tb>, typename = enable_if_all_var<Ta, Tc>>
inline var fma(Ta&& a, Tb&& b, Tc&& c) {
  return var(new internal::fma_vdv_vari(a.vi_, b, c.vi_));
}
```

Though not necessary, I think we would have to change `fma_vdv_vari` as well

stan/math/rev/scal/fun/fma.hpp
```diff
@@ -219,7 +224,8 @@ inline var fma(double a, double b, const var& c) {
  * @param c Summand.
  * @return Product of the multiplicands plus the summand, ($a * $b) + $c.
  */
-inline var fma(double a, const var& b, const var& c) {
+template <typename Ta, typename = enable_if_arithmetic<Ta>>
+inline var fma(Ta a, const var& b, const var& c) {
   return var(new internal::fma_vdv_vari(b.vi_, a, c.vi_));  // a-b symmetry
 }
```
SteveBronder: Out of the scope of this PR but it would be nice to rewrite these with one `fma_vari` method that has specializations of chain methods for the different implimentations in one class. Then all these specializations could be removed

SteveBronder:
#1262 was merged! Is this waiting for anymore review?

rok-cesnovar:
The question is probably only if @andrjohns is going to redo the tests with the new AD testing framework from Bob. I think it would be less work to do it here than creating a new separate PR just for that. We should use that everywhere we can anyways.

andrjohns:
I agree about the testing framework, I'll re-work those over the weekend

stan/math/fwd/scal/fun/ldexp.hpp
andrjohns: That's a good question. The ```std::ldexp(a,b)``` implementation only accepts integers for the exponent ([spec here](https://en.cppreference.com/w/cpp/numeric/math/ldexp)), and if a double is passed it gets internally cast to an int.

Given that the exponent needs to be an int, I don't think an ```fvar``` overload would work very well. Alternatively, we could just use the ```a * exp2(b)``` implementation, since that doesn't require ```b``` to be an integer.

According to the spec, it should also be faster as well:
> On many implementations, std::ldexp is less efficient than multiplication or division by a power of two using arithmetic operators.

bob-carpenter: Right---we shouldn't allow `fvar<T>` or `var` in positions that require integers.  More generally, we shouldn't ever be casting `real` values to `int` through the Stan language.  That is, if there's a function `foo(int)` in the Stan language that requires its first argument to be an integer, we should not implement `foo(real)` and have it cast.

stan/math/prim/scal/fun/logical_lte.hpp
andrjohns: Question with this one, what's the benefit of ```static_cast<bool>``` and an ```auto``` return type over just having a ```bool``` return type? Won't both approaches just return a ```bool```?

SteveBronder: Yeah actually ignore this comment I think it's unnecessary

stan/math/rev/scal/fun/fma.hpp
andrjohns: RValue references with ```rev``` functions is behaving very weirdly. If I update the function with the suggestion above, the test hangs. The same thing happens with:
```
inline var fma(const var&& a, const var&& b, const var&& c) {
  return var(new internal::fma_vvv_vari(a.vi_, b.vi_, c.vi_));
}
```

Is this something you've come across before?

andrjohns:
> Any idea where the test is hanging?  If not, could you show a version that worked and what changed in the one that doesn't?

Looks like it had something to do with template ambiguity, since adding ```enable_if_all_arithmetic``` to the ```prim``` definition got rid of the problem

andrjohns:
@SteveBronder this is ready for another look. I updated the ```prim``` functions with forwarding, but from the discussion in #1344 it looks like forwarding might not always be needed. Let me know if I should change that back

andrjohns:
Also, I needed to add ```std::decay``` to ```enable_if_arithmetic``` otherwise the templates weren't picking it up

SteveBronder:
> @SteveBronder this is ready for another look. I updated the prim functions with forwarding, but from the discussion in #1344 it looks like forwarding might not always be needed. Let me know if I should change that back

Apologies for the delay here, yes after doing stuff on godbolt plain old data inputs do not have a benefit from forwarding etc. Dan Lemire just had a good blog post that showed using [plain old copies is better for integer types](https://lemire.me/blog/2019/09/05/passing-integers-by-reference-can-be-expensive/). Feels like his results would carry over to other primitive types. These functions will always use primitive types right? Would you mind changing them back? If you are busy this week I can do that since it's my goof that led to you making the change.

bob-carpenter:
Yes, we've had the policy in place that primitives should be passed by value wherever possible.  

On a related note, const isn't useful for primitive arguments, as they are passed by value, so the API behavior doesn't change with "const int" vs. "int".

bob-carpenter:
> On Sep 24, 2019, at 2:23 PM, Rok Češnovar <notifications@github.com> wrote:
> 
> Doesn't const on a pass by value prevent that you modify the arguments value inside the function?

Sort of.  But you can always cast away const inside the body.

The bigger argument to not require references is that you can pass rvalues.  If you have foo(double& x), you can't call foo(2 + 3).


> It has no effect in terms of modification in the caller function, but it does inside the function.

Right.

Lots of discussion of this on stackoverflow.  I like the first answer to the big one:

* https://stackoverflow.com/questions/46292490/is-it-better-to-remove-const-in-front-of-primitive-types-used-as-function-pa/46292715

which is sort of what this one says

* https://stackoverflow.com/questions/16033751/c-const-modifier-with-primitive-types

And also some discussion on isocpp (no idea what status that has):

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

Overall, it's a minor point.  


> Its bad practice if you do it. But people do.

Why is it bad practice to modify an argument variable if it's a primitive?  That seems overly prescriptive to me, like requiring every function to have a single return statement.

The alternative would be to copy it.  I don't know the cost there---it might get compiled away since modifying the argument variable doesn't do anything to the caller.  

What the stackoverflow comments recommend is having the API not be const but the definition be const.  I don't even know if that's legal in C++.

Issue #1319, opened 2019-08-26
Implement GLMs in OpenCL


t4c1:
## Description
OpenCL GLMs were waiting for caching of matrices on GPU, since GPU GLMs require caching to achieve speedups compared to execution on CPU.

There was an attempt to implement caching (PR #1065), but it was abandoned due to problems with caching matrices that can change. On Stancon we have decided to simplify it by caching only matrices that are constant.

On discourse https://discourse.mc-stan.org/t/gpu-glms-float-or-double/9484/40 there was a discussion whether we could use floats instead of doubles in GLMs. The conclusion was that there might exist cases where reduced precision of floats could cause problems.

There will be a series of pull requests:
- [x] caching
- [x] normal_id
- [x] bernoulli_logit
- [x] poisson_log
- [x] neg_binomial_log
- [x] categorical_logistic
- [x] ordered_logistic ~~(after CPU code is merged - PR #1252)~~

## Example / Expected Output
GLMs are faster when using GPU.

#### Current Version:
v2.20.0

rok-cesnovar:
All of these were merged so closing. Thanks!

PR #1320, opened 2019-08-26
implement GPU caching

t4c1:
## Summary

Implements option for caching of matrices on GPU. Compared to PR #1065 this is simpler as only constant matrices are cached.

## Tests

Added a test to `matrix_cl_test.cpp`.

## Side Effects
caching  - `matrix_cl<T>::constant()` should only be used on matrices that do not change.

## Checklist

- [x] Math issue #1319 

- [x] Copyright holder: Tadej Ciglarič (University of Ljubljana)

### DIFF
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -180,7 +180,9 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
   cl::Buffer& buffer() { return buffer_cl_; }
   matrix_cl() : rows_(0), cols_(0) {}
 
-  matrix_cl(cl::Buffer& A, const int R, const int C, matrix_cl_view partial_view = matrix_cl_view::Entire) : rows_(R), cols_(C), view_(partial_view) {
+  matrix_cl(cl::Buffer& A, const int R, const int C,
+            matrix_cl_view partial_view = matrix_cl_view::Entire)
+      : rows_(R), cols_(C), view_(partial_view) {
     buffer_cl_ = A;
   }
 
@@ -345,41 +347,46 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
     }
   }
 
-    /**
-   * Constructs a const matrix_cl that contains a copy of the Eigen matrix on the OpenCL device.
-   * If the matrix already has a cached copy on the device, the cache is used and no copying is done.
-   * Changing the resulting matrix_cl would change cache, so do not do it! If changes are needed a copy must be made.
+  /**
+   * Constructs a const matrix_cl that contains a copy of the Eigen matrix on
+   * the OpenCL device. If the matrix already has a cached copy on the device,
+   * the cache is used and no copying is done. Changing the resulting matrix_cl
+   * would change cache, so do not do it! If changes are needed a copy must be
+   * made.
    *
    * @tparam R row type of input matrix
    * @tparam C column type of input matrix
    * @param A the Eigen matrix
    * @param partial_view which part of the matrix is used
    */
-    template <int R, int C>
-    static matrix_cl<T> constant(const Eigen::Matrix<T, R, C>& A, matrix_cl_view partial_view = matrix_cl_view::Entire){
+  template <int R, int C>
+  static matrix_cl<T> constant(const Eigen::Matrix<T, R, C>& A,
+                               matrix_cl_view partial_view
+                               = matrix_cl_view::Entire) {
 #ifdef STAN_OPENCL_CACHE
-      if (A.opencl_buffer_() != NULL) {
-        return matrix_cl<T>(A.opencl_buffer_, A.rows(), A.cols(), partial_view);
-      }
-      else{
-        matrix_cl<T> res(A, partial_view);
-        A.opencl_buffer_ = res.buffer();
-        return res;
-      }
+    if (A.opencl_buffer_() != NULL) {
+      return matrix_cl<T>(A.opencl_buffer_, A.rows(), A.cols(), partial_view);
+    } else {
+      matrix_cl<T> res(A, partial_view);
+      A.opencl_buffer_ = res.buffer();
+      return res;
+    }
 #else
-      return matrix_cl<T>(A, partial_view);
+    return matrix_cl<T>(A, partial_view);
 #endif
-    }
+  }
 
-    /**
-      * Constructs a const matrix_cl that contains a single value on the OpenCL device.
-      *
-      * @param A the value
-      * @param partial_view which part of the matrix is used
-      */
-    static matrix_cl<T> constant(T A, matrix_cl_view partial_view = matrix_cl_view::Entire){
-      return matrix_cl<T>(A);
-    }
+  /**
+   * Constructs a const matrix_cl that contains a single value on the OpenCL
+   * device.
+   *
+   * @param A the value
+   * @param partial_view which part of the matrix is used
+   */
+  static matrix_cl<T> constant(T A, matrix_cl_view partial_view
+                                    = matrix_cl_view::Entire) {
+    return matrix_cl<T>(A);
+  }
 
   /**
    * Construct from \c array of doubles with given rows and columns
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -181,7 +181,8 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
   matrix_cl() : rows_(0), cols_(0) {}
 
   /**
-   * Construct a matrix_cl<T> from an existing cl::Buffer object. The matrix directly uses given buffer - no copying is done.
+   * Construct a matrix_cl<T> from an existing cl::Buffer object. The matrix
+   * directly uses given buffer - no copying is done.
    *
    * @param A the cl::Buffer object to construct the matrix from
    * @param R number of rows
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -180,6 +180,10 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
   cl::Buffer& buffer() { return buffer_cl_; }
   matrix_cl() : rows_(0), cols_(0) {}
 
+  matrix_cl(cl::Buffer& A, const int R, const int C, matrix_cl_view partial_view = matrix_cl_view::Entire) : rows_(R), cols_(C), view_(partial_view) {
+    buffer_cl_ = A;
+  }
+
   matrix_cl(const matrix_cl<T>& A)
       : rows_(A.rows()), cols_(A.cols()), view_(A.view()) {
     if (A.size() == 0)
@@ -340,6 +344,42 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
       check_opencl_error("matrix constructor", e);
     }
   }
+
+    /**
+   * Constructs a const matrix_cl that contains a copy of the Eigen matrix on the OpenCL device.
+   * If the matrix already has a cached copy on the device, the cache is used and no copying is done.
+   * Changing the resulting matrix_cl would change cache, so do not do it! If changes are needed a copy must be made.
+   *
+   * @tparam R row type of input matrix
+   * @tparam C column type of input matrix
+   * @param A the Eigen matrix
+   * @param partial_view which part of the matrix is used
+   */
+    template <int R, int C>
+    static matrix_cl<T> constant(const Eigen::Matrix<T, R, C>& A, matrix_cl_view partial_view = matrix_cl_view::Entire){
+#ifdef STAN_OPENCL_CACHE
+      if (A.opencl_buffer_() != NULL) {
+        return matrix_cl<T>(A.opencl_buffer_, A.rows(), A.cols(), partial_view);
+      }
+      else{
+        matrix_cl<T> res(A, partial_view);
+        A.opencl_buffer_ = res.buffer();
+        return res;
+      }
+#else
+      return matrix_cl<T>(A, partial_view);
+#endif
+    }
+
+    /**
+      * Constructs a const matrix_cl that contains a single value on the OpenCL device.
+      *
+      * @param A the value
+      * @param partial_view which part of the matrix is used
+      */
+    static matrix_cl<T> constant(T A, matrix_cl_view partial_view = matrix_cl_view::Entire){
+      return matrix_cl<T>(A);
+    }
 
   /**
    * Construct from \c array of doubles with given rows and columns
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -191,9 +191,7 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
    */
   matrix_cl(cl::Buffer& A, const int R, const int C,
             matrix_cl_view partial_view = matrix_cl_view::Entire)
-      : rows_(R), cols_(C), view_(partial_view) {
-    buffer_cl_ = A;
-  }
+      : buffer_cl_(A), rows_(R), cols_(C), view_(partial_view) {}
 
   matrix_cl(const matrix_cl<T>& A)
       : rows_(A.rows()), cols_(A.cols()), view_(A.view()) {
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -180,6 +180,14 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
   cl::Buffer& buffer() { return buffer_cl_; }
   matrix_cl() : rows_(0), cols_(0) {}
 
+  /**
+   * Construct a matrix_cl<T> from an existing cl::Buffer object. The matrix directly uses given buffer - no copying is done.
+   *
+   * @param A the cl::Buffer object to construct the matrix from
+   * @param R number of rows
+   * @param C number of columns
+   * @param partial_view view of the matrix
+   */
   matrix_cl(cl::Buffer& A, const int R, const int C,
             matrix_cl_view partial_view = matrix_cl_view::Entire)
       : rows_(R), cols_(C), view_(partial_view) {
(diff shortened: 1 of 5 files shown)


stan/math/opencl/matrix_cl.hpp
```diff
@@ -341,6 +347,47 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
...
+   * @tparam R row type of input matrix
+   * @tparam C column type of input matrix
+   * @param A the Eigen matrix
+   * @param partial_view which part of the matrix is used
+   */
+  template <int R, int C>
+  static matrix_cl<T> constant(const Eigen::Matrix<T, R, C>& A,
+                               matrix_cl_view partial_view
+                               = matrix_cl_view::Entire) {
+#ifdef STAN_OPENCL_CACHE
```
rok-cesnovar: Do we still need this `STAN_OPENCL_CACHE` flag if caching will only be used when explicitly stated in memory transfer bound GPU functions? If anything, we might use a `STAN_OPENCL_NOCACHE` optionally. But I dont think event that is needed.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -180,6 +180,12 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
   cl::Buffer& buffer() { return buffer_cl_; }
   matrix_cl() : rows_(0), cols_(0) {}
 
+  matrix_cl(cl::Buffer& A, const int R, const int C,
```
rok-cesnovar: Please add doc to specify how this constructor is different from others.

test/unit/math/opencl/matrix_cl_test.cpp
```diff
@@ -26,4 +27,23 @@ TEST(MathMatrixCL, matrix_cl_types_creation) {
...
+  const matrix_cl<double> m1_cl = matrix_cl<double>::constant(m);
+
+  cl_mem mem_handle = m.opencl_buffer_();
+  EXPECT_EQ(mem_handle, m1_cl.buffer()());
+
+  const matrix_cl<double> m2_cl = matrix_cl<double>::constant(m);
+
+  EXPECT_EQ(mem_handle, m.opencl_buffer_());
+  EXPECT_EQ(mem_handle, m1_cl.buffer()());
+  EXPECT_EQ(mem_handle, m2_cl.buffer()());
```
rok-cesnovar: These are great. I would add 2 more tests: 
1. If we are going to stick with the `CACHE` flag or use a `NOCACHE` flag there should be a separate test to check the behavior when its not set. Again, I dont think the flag is actually needed.

2. a test that checks that no caching happens when using a regular constructor from eigen.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -341,6 +347,47 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
...
+   * would change cache, so do not do it! If changes are needed a copy must be
+   * made.
+   *
+   * @tparam R row type of input matrix
+   * @tparam C column type of input matrix
+   * @param A the Eigen matrix
+   * @param partial_view which part of the matrix is used
+   */
+  template <int R, int C>
+  static matrix_cl<T> constant(const Eigen::Matrix<T, R, C>& A,
```
SteveBronder: Why is this static? And why don't we just have a template flag  for `bool cache = false` instead of a new member function?

t4c1: If it was a member function instead of static one, we would need a matrix_cl object to call it on. To construct a new object it needs to be static.

It is impossible to explicitely set a constructor template parameter (One can set class template parameters. Constructor template parameters can only be deduced.)

stan/math/opencl/matrix_cl.hpp
t4c1: Option to turn caching off with STAN_OPENCL_NOCACHE might come in handy in some cases.

t4c1:
This is ready for next review once tests pass.

rok-cesnovar:
I am fine with the changes. If @SteveBronder is fine with the `constant()` function answer then LGTM.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -180,6 +180,21 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
...
+   *
+   * @param A the cl::Buffer object to construct the matrix from
+   * @param R number of rows
+   * @param C number of columns
+   * @param partial_view view of the matrix
+   */
+  matrix_cl(cl::Buffer& A, const int R, const int C,
+            matrix_cl_view partial_view = matrix_cl_view::Entire)
+      : rows_(R), cols_(C), view_(partial_view) {
+    buffer_cl_ = A;
```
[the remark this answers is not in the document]
t4c1: Nope. I will fix it.

stan/math/opencl/matrix_cl.hpp
t4c1: Adding a template to class just because we need two ways of constructing it from same argument type seems like an overkill. Especially since those two template instantinations of matrix_cl are identical once constructed. That would mean two versions or an additional template on every function that uses matrix_cl. Having separate static function for constructing one version is much simpler.

> Counterpoint to that though is if we will ever want to know in the future if something was cached or not.

Can you think of example where this would be needed?

SteveBronder: Not at the moment so I'm cool with the way we are doing this. Though I do want to think about that for a bit. But I also like that your way is v explicit

wds15:
Cool!!!

seantalts:
Whoa, this is our first state in the Math library! Not to be that guy but I don't think anyone super senior on the project had a chance to double check it; @SteveBronder can you ping me or @bob-carpenter or @syclik next time you guys are doing something really game changing like adding state? I would like to continue overseeing this type of work.

My main question here is: if this matrix is constant, why is it not marked `const`?

SteveBronder:
Yes this is my bad I wanted to hit you up but know how busy you are with the compiler stuff. No excuse for not pinging you tho' If there's anything here you have Qs Sure Rok, Tadej, and myself can make any fixes / answer Qs

> My main question here is: if the matrix returned by matrix_cl::constant is constant, why is it not marked const?

Yeah def I think that just slipped on me as I was reviewing. I'll look through this again tmrw and make another PR with some fixes

seantalts:
Hmm, I'm also thinking a bit more and wondering if this has the right semantics. What if I have a function that use a matrix as constant but it's called from another function that mutates it? Example:
```cpp
int foo(MatrixXd A) {
  mA = matrix_cl<double>::constant(A);
  ...
}

void bar(MatrixXd A) {
  int a = foo(A);
  A(20) = 4.0;
  int b = foo(A);
}
```

That seems bad, no? It's been a long day, maybe I'm wrong. This situation can easily happen in generated Stan code, though perhaps it wouldn't be super common we still need to think about the edge cases...

seantalts:
Compiler's basically done now :D I think you guys have come a long way and are great at reviewing each other's code, I just feel like in general some of these more major inflection points should still be highlighted to some other core devs as they're of considerable interest.

rok-cesnovar:
:/  Yeah, this case is bad...
```cpp
stan::math::matrix_d foo(const Eigen::MatrixXd& A) {
  stan::math::matrix_cl<double> mA = stan::math::matrix_cl<double>::constant(A);
  stan::math::matrix_cl<double> B(A.rows(), A.cols());
  B.zeros();
  const stan::math::matrix_cl<double> C = mA+B;
  return stan::math::from_matrix_cl(C);
}

TEST(MathMatrixCL, test_cache) {
  stan::math::matrix_d A = stan::math::matrix_d::Random(2,2);
  
  std::cout << A(0, 0) << std::endl;
  stan::math::matrix_d B = foo(A);
  A(0,0) = 2;
  stan::math::matrix_d C = foo(A);
  std::cout << C(0, 0) << std::endl;
}
```
Currently in the math lib `::constant` is only used on `const` matrices but there is no prevention of anyone doing something weird with it. Apologize for missing it.

t4c1:
>My main question here is: if the matrix returned by matrix_cl::constant is constant, why is it not marked const?

Unfortunately returning a const value does not do much in c++. Some explanation is [here](https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value). Only way to deter a user from doing something stupid I could think of was mentioning that result matrix should not be changed in the doxygen comment.

> :/ Yeah, this case is bad...

Huh, that seems like a huge issue. I don't see a way how to fix that. Such a code could come from a stan model, so documenting that this should not be done does not seem like a solution.

seantalts:
Agreed - I think we gotta figure out something to do about the external
mutation situation before we start using this code.

On the less important topic of const correctness - I took a look at that
SO, thanks for the link. I think returning const only messes with rvalue
references when you actually want to mutate the thing, right? We want to
make it clear that the result cannot be mutated in any way from the
signature, even if it doesn't do a ton in the way of compiler restrictions.
So it naively seems to me like their advice doesn't necessarily apply to
this case, though maybe there are more updated recommendations on this case
from Scott Meyers or Herb or someone.

rok-cesnovar:
I think it would be best to remove the use of `::constant` from the normal_id glms for now. Its not used anywhere else. Its going to make them considerably less fast but I would still leave the rest of the GPU GLM code even if it wont be as fast. I would also continue with the process of merging GPU GLMs (without the constant offcourse). We can put caching back once (if) we figure out a solution.

SteveBronder:
Little confused, If we only accept const matrices and return a const matrix when would the assignment problem show up?

Could we do something like 

```cpp

if (std::is_const<Mat>::value) {
// do the cache
} else {
// Throw or make a copy
}
```

seantalts:
Try Rok's test out locally, he did a good job codifying and verifying. The problem is that declaring a const function argument tells you nothing about its const-ness outside of that function's scope.

rok-cesnovar:
Yep, as long as foo has the const argument like this `foo(const Eigen::MatrixXd& A)` `matrix_cl::constant(A)` is going to treat the input as const even if you call foo without const, no SFINAE wizardry can help us here, I tried that also. 

If there would be a way to invalidate the buffer to null whenever we write to the Eigen matrix would work here. But that isn't doable without hacking Eigen probably. And that still wouldn't prevent the case of someone changing data through A.data()[i]. :(

SteveBronder:
Ah, so yeah the below does work but Sean your correct

```cpp
#include <iostream>
#include <type_traits>
#include <stan/math/opencl/opencl.hpp>
template <typename Mat>
Eigen::MatrixXd foo(Mat&& A) {
  if (std::is_const<std::remove_reference_t<Mat>>::value) {
    std::cout << "You Shall Not Pass!" << "\n";
    // for example purposes they do in fact Pass
    return A;
  } else {
    stan::math::matrix_cl<double> mA = stan::math::matrix_cl<double>::constant(A);
    stan::math::matrix_cl<double> B(A.rows(), A.cols());
    B.zeros();
    const stan::math::matrix_cl<double> C = mA+B;
    return stan::math::from_matrix_cl(C);
  }
}

TEST(MathMatrixCL, test_cache) {
  Eigen::MatrixXd A = Eigen::MatrixXd::Random(2,2);
  const Eigen::MatrixXd AA(2, 2);

  std::cout << A(0, 0) << std::endl;
  // pass non-const
  Eigen::MatrixXd B = foo(A);
  // pass const
  Eigen::MatrixXd BB = foo(AA);
  A(0,0) = 2;
  Eigen::MatrixXd C = foo(A);
  std::cout << C(0, 0) << std::endl;
}
```

But yeah replacing foo with this function makes it think everything is const

```
template <typename Mat>
Eigen::MatrixXd bad_boye(const Mat& A) {
  return foo(A);
}
```

Fun fact I mentioned in another thread, the above is why const refs cause invalidation on the cache. Best but obviously bad in terms of time would be to have stan use T&&, though that's a big issue that is surrounded by a swarm of technical debt.

The only "safe" answer I can think of while still getting speedups from cache is to, when calling constant, start the memory transfer and call a kernel comparing the new matrix with the old. Have the return value of that hooked into a callback that throws if the matrices don't match. Then you can have the program keep doing whatever with the async setup. We'll still 'pay' for the transfer, but we will also be doing computation as that transfer is happening.

That would actually be really cool to have for all of our error checking stuff on the gpu

seantalts:
On Sat, Sep 14, 2019 at 11:14 Steve Bronder <notifications@github.com>
wrote:

> Ah, so yeah the below does work but Sean your correct
>
> #include <iostream>
> #include <type_traits>
> #include <stan/math/opencl/opencl.hpp>template <typename Mat>
> Eigen::MatrixXd foo(Mat&& A) {
>   if (std::is_const<std::remove_reference_t<Mat>>::value) {
>     std::cout << "You Shall Not Pass!" << "\n";
>     // for example purposes they do in fact Pass
>     return A;
>   } else {
>     stan::math::matrix_cl<double> mA = stan::math::matrix_cl<double>::constant(A);
>     stan::math::matrix_cl<double> B(A.rows(), A.cols());
>     B.zeros();
>     const stan::math::matrix_cl<double> C = mA+B;
>     return stan::math::from_matrix_cl(C);
>   }
> }
> TEST(MathMatrixCL, test_cache) {
>   Eigen::MatrixXd A = Eigen::MatrixXd::Random(2,2);
>   const Eigen::MatrixXd AA(2, 2);
>
>   std::cout << A(0, 0) << std::endl;
>   // pass non-const
>   Eigen::MatrixXd B = foo(A);
>   // pass const
>   Eigen::MatrixXd BB = foo(AA);
>   A(0,0) = 2;
>   Eigen::MatrixXd C = foo(A);
>   std::cout << C(0, 0) << std::endl;
> }
>
> But yeah replacing foo with this function makes it think everything is
> const
>
> template <typename Mat>
> Eigen::MatrixXd bad_boye(const Mat& A) {
>   return foo(A);
> }
>
> Fun fact I mentioned in another thread, the above is why const refs cause
> invalidation on the cache. Best but obviously bad in terms of time would be
> to have stan use T&&, though that's a big issue that is surrounded by a
> swarm of technical debt.
>

Can you say more about this or link to the other thread? I'm confused about
both of those statements, 😂

> The only "safe" answer I can think of while still getting speedups from
> cache is to, when calling constant, start the memory transfer and call a
> kernel comparing the new matrix with the old. Have the return value of that
> hooked into a callback that throws if the matrices don't match. Then you
> can have the program keep doing whatever with the async setup. We'll still
> 'pay' for the transfer, but we will also be doing computation as that
> transfer is happening.
>
> That would actually be really cool to have for all of our error checking
> stuff on the gpu
>

I think it's a lot better to start building the transfer logic into stanc3
instead; it should be relatively easy even without changing Stan programs
at all if we just record which functions are overloaded for matrix
arguments and pass a flag noting we want to turn on OpenCL where applicable.

That's a great idea, to see if we can find speedups by doing error checking
asynchronously on GPU while it's also doing the rest of the actual
calculation.

rok-cesnovar:
@seantalts @SteveBronder How would you feel if we just reverse how we treat the caching flag for now? 

So that by default we dont use any caching, but developers/users can turn it on by setting a flag at their own risk, knowing that x and y should not change between iterations? The speedup is an order of magnitude larger so it might be worth it. But I am also fine if you say no.

seantalts:
Hm, I don't think we want to write the doc for that and do all the support for it only to rip it out when we add support to the compiler... Maybe we can even get the compiler support in for the next release.

t4c1:
> I think returning const only messes with rvalue references when you actually want to mutate the thing, right?

Not exactely. Assume we assign the result of function call to a variable. If that result is const it needs to be copied to the variable. If not it can be moved instead. I have no idea if returning a const value interfears with compiler doing copy elision.

> I think it would be best to remove the use of ::constant from the normal_id glms for now.

Than  we can as well revert whole GPU implementation as well. Without `matrix_cl::constant()` it will be slower than CPU. Depending on the number of attributes other GLMs glms might be slower than CPU too.

> Can you say more about this or link to the other thread?

I think he meant this PR: https://github.com/stan-dev/math/pull/1344.

SteveBronder:
> Than we can as well revert whole GPU implementation as well. Without matrix_cl::constant() it will be slower than CPU. Depending on the number of attributes other GLMs glms might be slower than CPU too.

I think Sean is saying he can do the cache stuff inside the compiler for matrices we know won't change through the life of the program

rok-cesnovar:
> Hm, I don't think we want to write the doc for that and do all the support for it only to rip it out when we add support to the compiler... Maybe we can even get the compiler support in for the next release.

That is reasonable yes. Caching is something Stan Math users/developers doesn't really
need as its their responsibility when things move. Its only an issue in combination with Stan.

Can we help in any way with the compiler support? This is probably the no.1 priority for our team, the other one being the matrix_cl<rev>. Will this be a flag on the Eigen matrix that will mean this Matrix's contents are immutable? Would it help if we were present for the meeting you mentioned on discourse?

> Than we can as well revert whole GPU implementation as well. Without `matrix_cl::constant()` it will be slower than CPU. Depending on the number of attributes other GLMs glms might be slower than CPU too.

That is your call, you did most if not all the work there. I think cleaning up the code of caching and preparing for potential compiler support will be better than reverting now and then doing issues, PRs and reviews again.

t4c1:
I was just thinking it would be a little weird if someone tested GPU implementation and found it slower no matter the problem size. Let's revert just the cache.

seantalts:
> Not exactely. Assume we assign the result of function call to a variable. If that result is const it needs to be copied to the variable. If not it can be moved instead. I have no idea if returning a const value interfears with compiler doing copy elision.

[Apparently ](https://stackoverflow.com/questions/14429988/can-a-c-compiler-perform-rvo-for-a-const-return-value) it doesn't prevent RVO but a const return value from `foo` would force a copy into `a` with assignment `a = foo()`. Today I learned! 

Re: reverting - I think it could make sense to revert the cache, but I think we need to keep the GPU GLM *specializations* so that they can be targeted by the compiler. It would be okay to remove any code from the GLMs that, if STAN_OPENCL was set, would automatically transfer an Eigen matrix to a matrix_cl and attempt computation on the GPU. Does that distinction make sense? We need the underlying kernels and function overloads, but the automatic CPU->GPU switching code can be turned off for now.

seantalts:
I'm going to move the compiler part of this to a discourse thread! I was typing a novel here on a closed PR, lol.

seantalts:
stanc3+gpu thread: https://discourse.mc-stan.org/t/stanc3-math-lib-opencl-integration/10991

PR #1322, opened 2019-08-26
Clean up metaprogramming

SteveBronder:
## Summary

1. is_* structs prefer SFINAE over partial specialization. This means that things like `scalar_type` for Eigen matrices are now specialized and pull out the `T::Scalar` value.

2. is_* base structs inherit from `std::false_type` in `scal` and then their specializations inherit from either `std::true_type` or `std::interal_constant<bool, value>`.

3. Removes `value_type` because it was redundant to `scalar_type`. It was only used in two places in the math library.

4. Some of the tests were looking at `std::vector<const T>` and `std::vector<T&>` which are both illegal C++11

5. The is_* functions use `std::decay_t<T>` s.t. const and ref types of the check will evaluate to `true`. I don't know of anywhere in the library where we want to check something is a ref of a type specifically.

## Tests

This is a refactor so idt any new tests are necessary

## Side Effects

The meta methods are now a bit more well defined. `is_vector` will now not work for `std::vector<const double>`, though that is also illegal in the C11+ standards

## Checklist

- [x] Math issue #1308


test/unit/math/prim/arr/meta/contains_vector_test.cpp
```diff
@@ -8,6 +8,6 @@ TEST(MetaTraits, contains_vector) {
 
   EXPECT_TRUE(contains_vector<std::vector<double> >::value);
   EXPECT_TRUE(contains_vector<std::vector<int> >::value);
-  EXPECT_TRUE(contains_vector<std::vector<const double> >::value);
```
SteveBronder: Items in a vector must be assignable

https://stackoverflow.com/questions/17313062/vector-of-const-objects-giving-compile-error

test/unit/math/rev/scal/meta/is_var_or_arithmetic_test.cpp
```diff
@@ -1,18 +1,21 @@
...
-  EXPECT_FALSE(stan::is_var_or_arithmetic<stan::math::var&>::value);
+  EXPECT_TRUE(stan::is_var_or_arithmetic<stan::math::var&>::value);
+  EXPECT_TRUE(stan::is_var_or_arithmetic<std::vector<stan::math::var>>::value);
   bool temp
       = is_var_or_arithmetic<stan::math::var, std::vector<stan::math::var>,
                              stan::math::var, stan::math::var, stan::math::var,
                              stan::math::var>::value;
   EXPECT_TRUE(temp);
   temp = is_var_or_arithmetic<std::vector<stan::math::var>, stan::math::var,
-                              std::vector<stan::math::var&>>::value;
```
SteveBronder: elements of a vector must have a well formed destructor

https://stackoverflow.com/questions/922360/why-cant-i-make-a-vector-of-references

SteveBronder:
Closing this till I have time to come back to it

PR #1326, opened 2019-08-28
Refactor/clang tidy missing braces

SteveBronder:
## Summary
I ran clang-tidy over the repo and found a bunch of places we don't use braces around if, for, etc. It would be nice to have that for readability. This PR adds braces around a lot of the conditional statements etc. that were missing them in the math library!

## Tests

No new tests

## Checklist

- [x] Math issue #1308 

- [x] Copyright holder: Steve Bronder

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1327, opened 2019-08-28
Adds clang-tidy to makefile

SteveBronder:
## Summary

This adds `clang-tidy`to the makefile which can be called like 

```bash
make clang-tidy folder={math test folder} files={regular_expression}
```

```bash
make clang-tidy-fix folder=./test/unit/math/rev/mat/prob files=*
```

I've gone through a bunch of the test and exclude some stuff idt we like. You can view them all here

https://clang.llvm.org/extra/clang-tidy/checks/list.html

I'm not sure when we should use this the fix version. Whether it should be ad-hoc or if we do it before a new version

## Tests

No new tests

## Side Effects

## Checklist

- [x] Math issue #1308


make/compiler_flags
```diff
@@ -13,7 +13,7 @@ ifeq (default,$(origin CXX))
     CXX := clang++
   endif
   ifeq ($(OS),Linux)
-    CXX := g++
+    CXX := clang++
```
SteveBronder: Not sure how we should go about fixing this one, clang-tidy uses clang but I'd like to keep using g++ on linux

make/clang-tidy
```diff
@@ -0,0 +1,16 @@
+
+tidy_base_checks=*perf*,*modernize*,*misc*,*google*,*readability*,*optin*
```
rok-cesnovar: Can we add a link to a clang-tidy doc here? That would help someone trying to understand these flags. If I am getting this right this does all the checks in `*perf*,*modernize*,*misc*,*google*,*readability* and *optin*` except the ones that are excluded?

I was exploring these a few months ago, but did not get near as far as you did now :) This is really cool!

make/compiler_flags
rok-cesnovar: Can we set a variable if clang-tidy is used and then set clang++ here?

rok-cesnovar:
I think this is really great. A quick link to the clang-tidy doc would help. But 

If or when we decide to use this in CI for PRs like `cpplint` we should make a wiki describing what clang-tidy checks for (in short). Until this is an optional thing we will run and fix on separate PRs that is probably overkill.

make/compiler_flags
rok-cesnovar: This comment would also apply here https://github.com/stan-dev/math/pull/1326#discussion_r318978191

syclik: Yeah. That sounds right. 

We want to leave g++ as the default.

make/clang-tidy
```diff
@@ -0,0 +1,16 @@
...
+readability_excludes=,-readability-magic-numbers,-readability-convert-member-functions-to-static,-readability-inconsistent-declaration-parameter-name,-readability-function-size,-readability-named-parameter,-readability-container-size-empty,-readability-implicit-bool-conversion,-readability-else-after-return,-readability-container-size-empty
+misc_excludes=,-misc-non-private-member-variables-in-classes,-misc-unused-parameters,-misc-unused-using-decls,-misc-definitions-in-headers,-misc-unconventional-assign-operator
+google_excludes=,-google-runtime-references,-google-global-names-in-headers,-google-build-using-namespace,-google-runtime-int
+things_we_should_look_at=,-google-readability-casting,-google-explicit-constructor,-modernize-use-using,-modernize-pass-by-value,-readability-implicit-bool-conversion,-readability-const-return-type,-readability-avoid-const-params-in-decls,-modernize-return-braced-init-list,-modernize-use-using
+tidy_checks?= $(tidy_base_checks) $(modern_excludes) $(readability_excludes) $(misc_excludes) $(google_excludes) $(things_we_should_look_at)
+
+.PHONY: clang-tidy
+clang-tidy:
```
syclik: So... we might be able to override a variable here just for this target. Before these lines, try adding:

```
clang-tidy: CXX=clang++
```

If this works, great! If it doesn't, it's probably some interaction with phony targets and rules for them.

SteveBronder: Turns out I'm dumb and g++ works just fine haha

make/clang-tidy
SteveBronder: Yep added a link to the top of the file!

make/compiler_flags
SteveBronder: Fixed! Turns out clang-tidy runs just fine with g++

SteveBronder:
Alrighty just updated this! If yinz could take a look and all looks well we can merge this and I'll start running some of the automated fixes and put them up as PRs

Issue #1294, opened 2019-07-30
Rewrite the test-math-dependencies script in Python


rok-cesnovar:
## Description
The [test-math-dependencies](https://github.com/stan-dev/math/blob/develop/make/test-math-dependencies) Makefile script is difficult to maintain and add new features to it. Mainly due to the weird scripting style of Make. It should be rewritten in Python.

Initially it was not written in Python due to some installation issues (see: https://github.com/stan-dev/math/issues/1270). We should be able to rewrite and use it now.

Additionaly add the following checks:
- check that all `.cpp` files inside test/unit/math end on `_test.cpp`

## Expected Output

Exact same functionality of test-math-dependencies, but in Python.

#### Current Version:
v2.20.0

syclik:
This would be awesome! Yes, we're well past the python 2 vs 3 battle!

rok-cesnovar:
I have added a note in the top comment to also add a check for any `*.cpp` inside `test/unit/math` that dont end on `_test.cpp`. So that we avoid issues like these https://github.com/stan-dev/math/issues/1312

PR #1329, opened 2019-08-28
Feature/Issue 1294 Rewrite the test-math-dependencies script in Python

rok-cesnovar:
## Summary

This PR attempts a rewrite of [make/test-math-dependencies](https://github.com/stan-dev/math/blob/develop/make/test-math-dependencies) to Python. I dont usually write Python scripts that would go out to the public and tend to not write it in the most pythonic way. I did my best and the script does the job.

It currently runs the same checks as the existing make script with an additional one that checks for any `test/unit/*/*.cpp` file that does not end with a `_test.cpp` suffix. Such a file is most likely a new added file with tests that would not get picked up by `runTests.py`.

This now also makes this check script available on Windows. `make test-math-dependencies` does not work on Windows, at least not with Rtools 3.5.

It also properly ignores line comments (`//`) compared to the make script. If someone where to write `// something stan::math::var` in a `/prim` file the existing script would trigger a warning.

Example of an output on a warning: 
```
stan/math/prim/mat/fun/acos.hpp at line 20:
	[prim] File uses stan::math::var.

```
## Tests

Make an error that would trigger a warning in an existing parser and run `python runChecks.py`.

## Side Effects

This is considerably slower compared to the make script (6s vs 0.5s). But I think the maintainability  outweighs the longer execution time since this is not intended for users and is not something you run all the time. Alternatively we can re-do the script to use `find` an `grep` as external processes.

## Checklist

- [x] Math issue #1294 


rok-cesnovar:
@serban-nicusor-toptal @seantalts Is there an option of using Python 3.5 on the Jenkins instances that runs the `'Linting & Doc checks'` step. I am using some feature of `glob` that are only [supported](https://stackoverflow.com/questions/56103938/glob-got-an-unexpected-keyword-argument-recursive/56103966) on python 3.5+. No issue if we cant, I will just need to find another solution.

serban-nicusor-toptal:
Hey, I can create a separate fleet in AWS that only spins up for that step and call it by `linux-python35` or something around this.
If @seantalts is okay with this I can do it asap.

rok-cesnovar:
A separate fleet for such a small thing seems a bit excessive, idk. Will try to rethink this so that it does wouldnt need python 3.5. Thank @serban-nicusor-toptal for the quick response.

serban-nicusor-toptal:
Maybe we can use something like pyenv? https://github.com/pyenv/pyenv

And we switch on the fly

rok-cesnovar:
@serban-nicusor-toptal wont be needed, I found a solution that works on Python 2.7 and doesnt feel too hacky. Thank you for the effort!

@ahartikainen could I trouble you for a review of a simple Python script here? Just a general overview if something could be done more "pythonic" :)

runChecks.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/python
+
+"""
+Replacement for test-math-dependencies target in Makefile.
+
+"""
+
+from __future__ import print_function
+import os
+import os.path
```
ahartikainen: not needed (import `os` will import `os.path`)

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+winsfx = ".exe"
+testsfx = "_test.cpp"
+def files_in_folder(folder):
+    files = []
+    for f in glob.glob(folder):
+        if os.path.isdir(f):
+            files += files_in_folder(f+os.sep+"**")
+        else:
+            files.append(f)
+    return files
```
ahartikainen: Maybe `os.walk` could be better fit to get all the files (including subfolders).

```
def files_in_folder(folder):
    files = []
    for root, _, filenames in os.walk(folder):
        for filename in filenames:
            files.append(os.path.join(root, filename))
    return files
```

And for only the main folder

```
def files_in_folder(folder):
    files = []
    for root, _, filenames in os.walk(folder):
        for filename in filenames:
            files.append(os.path.join(root, filename))
        break
    return files
```

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    """Checks the files in the provided folder for matches
+    with any of the patterns. It returns an array of
+    messages and the provided type with
+    the line number. This check ignores comments.
+    """
+    errors = []
+    folder.replace("/", os.sep)
+    exclude_files = []
+    for excl in exclude_filters:
+        exclude_files += files_in_folder(excl)
```
ahartikainen: `+=` works in most cases until there is some edge case, so `.append` and `.extend` are preferred.

```
exclude_files.extend(files_in_folder(excl))
```

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    files = files_in_folder(folder+os.sep+"**")
```
ahartikainen: `files = files_in_folder(folder)`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    files = [x for x in files if x not in exclude_files]
+    for f in files:
```
ahartikainen: I would use `f` --> `filepath` (`f` is usually filehandle from the `open` function)

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+        if os.path.isfile(f):
+            line_num = 0
+            for line in open(f, "r"):
```
ahartikainen: This will remember to call `.close` even if there is an exception

```
with open(filepath, "r") as f:
    for line in f:
        ...
```

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+                line_num += 1
+                
+                for p in patterns_and_messages:
+                    # exclude line starting with "/*" or " * " and
+                    # if the matched patterns are behind "//"
+                    if not re.search("^ \* |^/\*", line) and not re.search(".*//.*"+p["pattern"], line) and re.search(p["pattern"], line):
```
ahartikainen: I have this monster in ArviZ to remove comments, but not sure if it would even work here

```
pattern_remove_comments = re.compile(
        r'//.*?$|/\*.*?\*/|\'(?:\\.|[^\\\'])*\'|"(?:\\.|[^\\"])*"', re.DOTALL | re.MULTILINE
    )
...
stan_code = re.sub(pattern_remove_comments, "", stan_code)
```

https://github.com/arviz-devs/arviz/blob/cdbe656a0c759c3f0d83e7285b8b7e8ac5992704/arviz/data/io_pystan.py#L511

ahartikainen: But your current code does not handle multiline comments.

So you could find the locations for the comments and then eliminate stuff based on those limits (start, end)

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+                            errors.append(f + " at line " + str(line_num) + ":\n\t" + "[" + type + "] " + p["message"])
+    return errors
+
+def check_non_test_files_in_test():
+    all_cpp = files_in_folder("test/unit/*/*")
```
ahartikainen: `all_cpp = files_in_folder("test/unit")`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    # if the file is a .cpp file that doesnt end with _test.cpp
+    errors = ["Error: A .cpp file without the suffix " + testsfx + " in test/unit/*/*:\n\t" + x for x in all_cpp if x[-4:] == ".cpp" and x[-len(testsfx):] != testsfx]
```
ahartikainen: `x[-4:]` works here

there is also `os.path.splitext`

    os.path.splitext(x)[1]

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Check for files inside stan/math/prim that contain stan/math/rev or stan/math/fwd
+    prim_checks = [
+        { "pattern" : '<stan/math/rev/', "message" : 'File includes a stan/math/rev header file.'},
+        { "pattern" : 'stan::math::var', "message" : 'File uses stan::math::var.'},
+        { "pattern" : '<stan/math/fwd/', "message" : 'File includes a stan/math/fwd header file.'},
+        { "pattern" : 'stan::math::fvar', "message" : 'File uses stan::math::fvar.'},
+        { "pattern" : '<stan/math/mix/', "message" : 'File includes a stan/math/mix header file.'}
+    ]
+    errors += grep_patterns('prim', 'stan/math/prim', prim_checks)
```
ahartikainen: `errors.extend(grep_patterns('prim', 'stan/math/prim', prim_checks))`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    
+    # Check for files inside stan/math/rev that contain stan/math/fwd or stan/math/mix
+    rev_checks = [
+        { "pattern" : '<stan/math/fwd/', "message" : 'File includes a stan/math/fwd header file.'},
+        { "pattern" : 'stan::math::fvar', "message" : 'File uses stan::math::fvar.'},
+        { "pattern" : '<stan/math/mix/', "message" : 'File includes a stan/math/mix header file.'}
+    ]
+    errors += grep_patterns('rev', 'stan/math/rev', rev_checks)
```
ahartikainen: `errors.extend(grep_patterns('rev', 'stan/math/rev', rev_checks))`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    # Check for files inside stan/math/*/scal that contain stan/math/*/arr or stan/math/*/mat
+    scal_checks = [
+        { "pattern" : '<stan/math/.*/arr/', "message" : 'File includes an array header file.'},
+        { "pattern" : '<vector>', "message" : 'File includes an std::vector header.'},
+        { "pattern" : 'std::vector', "message" : 'File uses std::vector.'}, 
+        { "pattern" : '<stan/math/.*/mat/', "message" : 'File includes a matrix header file.'}, 
+        { "pattern" : '<Eigen', "message" : 'File includes an Eigen header.'},
+        { "pattern" : 'Eigen::', "message" : 'File uses Eigen.'}
+    ]
+    errors += grep_patterns('scal', 'stan/math/*/scal', scal_checks)
```
ahartikainen: `errors.extend(grep_patterns('scal', 'stan/math/*/scal', scal_checks))`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Check for files inside stan/math/*/arr that contain stan/math/*/mat or Eigen
+    arr_checks = [
+        { "pattern" : '<stan/math/.*/mat/', "message" : 'File includes an matrix header file.'},
+        { "pattern" : '<Eigen', "message" : 'File includes an Eigen header.'},
+        { "pattern" : 'Eigen::', "message" : 'File uses Eigen.'}
+    ]    
+    errors += grep_patterns('arr', 'stan/math/*/arr', arr_checks)
```
ahartikainen: `errors.extend(grep_patterns('arr', 'stan/math/*/arr', arr_checks))`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Check to make sure we use C++14 constructs in stan/math
+    cpp14_checks = [
+        { "pattern" : 'boost::is_unsigned<', "message" : 'File uses boost::is_unsigned instead of std::is_unsigned.'},
+        { "pattern" : '<boost/type_traits/is_unsigned>',\
```
ahartikainen: `\` is not needed if the line continues with the same indentation.

```
[
    {"pattern" :  "patt",
     "message" : "patt2"},
```

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+        { "pattern" : 'boost::is_same<', "message" : 'File uses boost::is_same instead of std::is_same.'},
+        { "pattern" : '<boost/type_traits/is_same.hpp>',\
+            "message" : 'File includes <boost/type_traits/is_same.hpp> instead of <type_traits>.'},
+        { "pattern" : 'boost::enable_if_c<', "message" : 'File uses boost::enable_if_c instead of std::enable_if.'},
+        { "pattern" : 'boost::enable_if<', "message" : 'File uses boost::enable_if instead of std::enable_if.'},
+        { "pattern" : 'boost::disable_if<', "message" : 'File uses boost::disable_if instead of std::enable_if.'},
+        { "pattern" : '<boost/utility/enable_if.hpp>',\
+            "message" : 'Replace \<boost/utility/enable_if.hpp\> with \<type_traits\>.'}
+    ]    
+    errors += grep_patterns('C++14', 'stan/math', cpp14_checks)
```
ahartikainen: `errors.extend(grep_patterns('C++14', 'stan/math', cpp14_checks))`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+    
+    # Check for includes of stan/math/*/meta/*.hpp inside stan/math, excluding meta.hpp files and the /meta subfolder
+    meta_checks = [
+        { "pattern" : '<stan/math/.*/meta/.*hpp', "message" : 'File includes */meta/*.hpp header file. Should include meta.hpp'}
+    ]
+    meta_exclude = ['stan/math/*/*/meta', 'stan/math/*/meta.hpp']
+    errors += grep_patterns('meta', 'stan/math', meta_checks, meta_exclude)
```
ahartikainen: `errors.extend(grep_patterns('meta', 'stan/math', meta_checks, meta_exclude))`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+
+    errors += check_non_test_files_in_test()
```
ahartikainen: `errors += check_non_test_files_in_test()`

runChecks.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if(len(errors) > 0):
```
ahartikainen: No need for len here

`if errors:`

runChecks.py
ahartikainen: These are for `.stan` files, but I think the same comment system is used with C++?

runChecks.py
rok-cesnovar: I tried using `os.walk` but  it does not support search with "wildcards", meaning we cant list file that match `'stan/math/*/*/meta'`.

runChecks.py
rok-cesnovar: Did not know that. Thank you!

runChecks.py
rok-cesnovar: This one was done because I want to ignore the `test/unit` files where we have the `translation_unit` cpp files that are the only non `_test.cpp` files we have. It was either that or a list of exception files, this seems cleaner.

runChecks.py
rok-cesnovar: Thanks!

runChecks.py
rok-cesnovar: You probably meant use .extend here. Thank you.

runChecks.py
rok-cesnovar: I changed it to `"test/unit/math"`. Thanks

runChecks.py
rok-cesnovar: Yeah, the original script we are replacing here only worked with Doxygen blocked comments in the form of:
```
/**
 * text text
 */
```
I actually think that there are no block comments in the form
```
/*
 text text
 more text
*/
in the Stan Math codebase. But yes, we should also cover that, good call. Will redo this, but avoiding removing the comments to preserved the count. Thank you for the example.

**Comment by rok-cesnovar** [comment]
A huge thanks @ahartikainen!! This was quick and thorough. I owe you one or two now :)
I am fixing everything based on your comments except for the `os.walk` one because I think we need the wildcard way of specifying files.

**Comment by rok-cesnovar** [comment]
If Ari doesnt have any further comments, I think this is ready for @syclik to review. 

This is nothing urgent and the branch is not in danger of coming into conflicts often (once `lgamma` gets merged I will include that check).

**Comment by ahartikainen** [comment]
Yes, everything looks good.

If there is a problem with performance, we can fix that later.

**Comment by syclik** [comment]
@rok-cesnovar, this is awesome! Did you want to remove the make/test-math-dependencies file while we're at it?

I looked at the code and it looked right, but before I go and approve it, how did you test this? Should I go through and make sure it picks up on the right violations of dependencies?

File: stan/math/rev/mat/fun/gp_periodic_cov.hpp [hunk 19]
```diff
@@ -350,7 +350,7 @@ gp_periodic_cov(const std::vector<T_x> &x, const var &sigma, const var &l,
  *   x is nan or infinite
  */
 template <typename T_x>
-inline typename std::enable_if<
+inline typename boost::enable_if<
```
SteveBronder: ?

SteveBronder: lol oh nvm

rok-cesnovar: This is a deliberate error commit to trigger the error message :) See the post on the PR thread in 5 seconds :P

rok-cesnovar:
I tested this before by making deliberate errors and commiting them: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1329/7/pipeline/

I actually did not do a thorough test since incorporating Ari's comments so here is the result of this run https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1329/10/pipeline/

I am now going to revert those back.

In commit  https://github.com/stan-dev/math/pull/1329/commits/347f8c2f1811e30449281e37e852f201d8943882 I changed the error message for the` _test.cpp` test to make it similar to the other error messages. I also tested this on Windows just now.

If you want to test it yourself checkout this branch on commit 4a8a873 and run `python runChecks.py`

rok-cesnovar:
And one more thing, I would hold off on merging this until the lgamma PR is merged. That PR is going to add some kind of check and I will incorporate it to this one and then tag you again. Does that sound reasonable?

SteveBronder:
ping me after the lgamma check and I'll look this over!

rok-cesnovar:
@syclik I added the std::lgamma check so this is now ready for another look. 

cc: @SteveBronder you wanted to be pinged when this will happen

SteveBronder:
Yep I'll look it over today!

makefile
```diff
@@ -42,14 +42,6 @@ help:
...
 	@echo '  Example: make stan/math/constants.hpp-test'
 	@echo ''
-	@echo '  - test-math-dependencies : walks through all the header files and indicates'
-	@echo '      when the math dependencies are violated. Dependencies should follow:'
-	@echo '      * rev -> prim'
-	@echo '      * fwd -> prim'
-	@echo '      * mix -> {rev, fwd, prim}'
-	@echo '      * within {prim, rev, fwd, mix}: mat -> arr -> scal'
-	@echo '      * only include {prim, rev, fwd, mix}/meta.hpp from the meta subfolders'
-	@echo ''
```
SteveBronder: Instead of removing `make test-math-dependencies` can we have that call run the python script?

runChecks.py
```diff
@@ -0,0 +1,243 @@
...
+    files = []
+    for f in glob.glob(folder):
+        if os.path.isdir(f):
+            files.extend(files_in_folder(f + os.sep + "**"))
+        else:
+            files.append(f)
+    return files
+
+
+def grep_patterns(type, folder, patterns_and_messages, exclude_filters=[]):
```
SteveBronder: Can you doc the input parameters?

runChecks.py
```diff
@@ -0,0 +1,243 @@
...
+
+def check_non_test_files_in_test():
+    all_cpp = files_in_folder("test/unit/math/")
+    # if the file is a .cpp file that doesnt end with _test.cpp
+    errors = [
+        x
+        + ":\n\t A .cpp file without the "
+        + testsfx
+        + " suffix in test/unit/math/"
+
```
SteveBronder: Can you remove the newline here?

runChecks.py
```diff
@@ -0,0 +1,243 @@
...
+import os
+import sys
+import re
+import glob
+
+winsfx = ".exe"
+testsfx = "_test.cpp"
+
+
+def files_in_folder(folder):
```
SteveBronder: Docstring

makefile
rok-cesnovar: That is a great idea, I dont have to change what Jenkins runs then also. Will do.

rok-cesnovar:
@SteveBronder Appreciate the review and the suggestion to add the command to make. Added the doc and reformmated again with black.

PR #1333, opened 2019-09-02
Implement normal_id_glm_lpdf in OpenCL

t4c1:
## Summary

This implements `normal_id_glm_lpdf()` in OpenCL. A new `matrix_cl<T>` constructor was added that accepts a single scalar. `from_matrix_cl()` has two new template parameters that control compile-time row and column count of returned matrix.

## Tests
Existing `normal_id_glm_lpdf()` tests have been modified to always use GPU if compiling with OpenCL support.

## Side Effects
None.

## Checklist

- [ ] Math issue #1319

- [ ] Copyright holder: Tadej Ciglarič (University of Ljubljana)

### DIFF
--- a/stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
+++ b/stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
@@ -1,0 +1,136 @@
+#ifndef STAN_MATH_OPENCL_KERNELS_NORMAL_ID_GLM_LPDF_HPP
+#define STAN_MATH_OPENCL_KERNELS_NORMAL_ID_GLM_LPDF_HPP
+#ifdef STAN_OPENCL
+
+#include <stan/math/opencl/kernel_cl.hpp>
+
+namespace stan {
+namespace math {
+namespace opencl_kernels {
+
+// \cond
+static const char *normal_id_glm_kernel_code = STRINGIFY(
+// \endcond
+        /**
+         * GPU implementation of Generalized Linear Model (GLM)
+         * with Normal distribution and identity link function.
+         *
+         * Must be run with at least N threads and local size equal to LOCAL_SIZE_.
+         * @param[in] y_glob vector parameter
+         * @param[in] x design matrix
+         * @param[in] alpha intercept (in log odds)
+         * @param[in] beta weight vector
+         * @param[in] sigma_glob (Sequence of) scale parameters for the normal
+         * @param[out] mu_derivative_glob intermediate variable used in the model
+         * @param[out] mu_derivative_sum partially summed mu_derivative_glob (1 value per work group)
+         * @param[out] y_minus_mu_over_sigma_squared_sum intermediate variable used in the model
+         * @param[out] sigma_derivative derivative with respect to sigma
+         * @param[out] partially summed logarithm of sigma (1 value per work group)
+         * @param N number of cases
+         * @param M number of attributes
+         * @param is_alpha_vector 0 or 1 - whether alpha is a vector (alternatively it is a scalar)
+         * @param is_sigma_vector 0 or 1 - whether sigma is a vector (alternatively it is a scalar)
+         * @param need_mu_derivative interpreted as boolean - whether mu_derivative needs to be computed
+         * @param need_mu_derivative_sum interpreted as boolean - whether mu_derivative_sum needs to be computed
+         * @param need_sigma_derivative interpreted as boolean - whether sigma_derivative needs to be computed
+         * @param need_log_sigma_sum interpreted as boolean - whether log_sigma_sum needs to be computed
+         */
+        __kernel void normal_id_glm(const __global double* y_glob, const __global double* x, const __global double* alpha, const __global double* beta, const __global double* sigma_glob,
+                                          __global double* mu_derivative_glob, __global double* mu_derivative_sum, __global double* y_minus_mu_over_sigma_squared_sum,
+                                          __global double* sigma_derivative, __global double* log_sigma_sum,
+                                          const int N, const int M, const int is_alpha_vector, const int is_sigma_vector, const int need_mu_derivative, const int need_mu_derivative_sum,
+                                          const int need_sigma_derivative, const int need_log_sigma_sum) {
+          const int gid = get_global_id(0);
+          const int lid = get_local_id(0);
+          const int lsize = get_local_size(0);
+          const int wgid = get_group_id(0);
+
+          __local double res_loc[LOCAL_SIZE_];
+
+          double y_minus_mu_over_sigma_squared = 0;
+          double log_sigma = 0;
+          double mu_derivative = 0;
+          if(gid<N){
+            double y_minus_mu_over_sigma=0;
+            for (int i = 0, j = 0; i < M; i++, j += N) {
+              y_minus_mu_over_sigma += x[j + gid] * beta[i];
+            }
+            double sigma = sigma_glob[gid * is_sigma_vector];
+            double inv_sigma = 1 / sigma;
+            y_minus_mu_over_sigma = (y_glob[gid] - y_minus_mu_over_sigma - alpha[gid*is_alpha_vector]) * inv_sigma;
+            mu_derivative = inv_sigma * y_minus_mu_over_sigma;
+            if(need_mu_derivative){
+              mu_derivative_glob[gid]=mu_derivative;
+            }
+            y_minus_mu_over_sigma_squared = y_minus_mu_over_sigma * y_minus_mu_over_sigma;
+            if(need_sigma_derivative){
+              sigma_derivative[gid]=(y_minus_mu_over_sigma_squared - 1) * inv_sigma;
+            }
+            if(need_log_sigma_sum){
+              log_sigma=log(sigma);
+            }
+          }
+          res_loc[lid] = y_minus_mu_over_sigma_squared;
+          barrier(CLK_LOCAL_MEM_FENCE);
+          for (int step = lsize / REDUCTION_STEP_SIZE; step > 0; step /= REDUCTION_STEP_SIZE) {
+            if (lid < step) {
+              for (int i = 1; i < REDUCTION_STEP_SIZE; i++) {
+                res_loc[lid] += res_loc[lid + step * i];
+              }
+            }
+            barrier(CLK_LOCAL_MEM_FENCE);
+          }
+          if (lid == 0) {
+            y_minus_mu_over_sigma_squared_sum[wgid] = res_loc[0];
+          }
+
+          if(need_mu_derivative_sum){
+            barrier(CLK_LOCAL_MEM_FENCE);
+            res_loc[lid] = mu_derivative;
+            barrier(CLK_LOCAL_MEM_FENCE);
+            for (int step = lsize / REDUCTION_STEP_SIZE; step > 0; step /= REDUCTION_STEP_SIZE) {
+              if (lid < step) {
+                for (int i = 1; i < REDUCTION_STEP_SIZE; i++) {
+                  res_loc[lid] += res_loc[lid + step * i];
+                }
+              }
+              barrier(CLK_LOCAL_MEM_FENCE);
+            }
+            if (lid == 0) {
+              mu_derivative_sum[wgid] = res_loc[0];
+            }
+          }
+
+          if(need_log_sigma_sum){
+            barrier(CLK_LOCAL_MEM_FENCE);
+            res_loc[lid] = log_sigma;
+            barrier(CLK_LOCAL_MEM_FENCE);
+            for (int step = lsize / REDUCTION_STEP_SIZE; step > 0; step /= REDUCTION_STEP_SIZE) {
+              if (lid < step) {
+                for (int i = 1; i < REDUCTION_STEP_SIZE; i++) {
+                  res_loc[lid] += res_loc[lid + step * i];
+                }
+              }
+              barrier(CLK_LOCAL_MEM_FENCE);
+            }
+            if (lid == 0) {
+              log_sigma_sum[wgid] = res_loc[0];
+            }
+          }
+        }
+// \cond
+);
+// \endcond
+
+/**
+ * See the docs for \link kernels/bernoulli_logit_glm_lpmf.hpp bernoulli_logit_glm() \endlink
+ */
+const kernel_cl<in_buffer, in_buffer, in_buffer, in_buffer, in_buffer, out_buffer, out_buffer, out_buffer, out_buffer, out_buffer, int, int, int, int, int, int, int, int>
+        normal_id_glm("normal_id_glm", {normal_id_glm_kernel_code}, {{"REDUCTION_STEP_SIZE",4},{"LOCAL_SIZE_", 64}});
+
+}  // namespace opencl_kernels
+}  // namespace math
+}  // namespace stan
+
+#endif
+#endif //STAN_MATH_OPENCL_KERNELS_NORMAL_ID_GLM_LPDF_HPP
--- a/stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
+++ b/stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
@@ -9,6 +9,11 @@
 #include <stan/math/prim/scal/fun/size_zero.hpp>
 #include <stan/math/prim/scal/fun/sum.hpp>
 #include <stan/math/prim/arr/fun/value_of_rec.hpp>
+
+#include <stan/math/opencl/kernels/normal_id_glm_lpdf.hpp>
+#include <stan/math/opencl/matrix_cl.hpp>
+#include <stan/math/opencl/multiply.hpp>
+
 #include <cmath>
 
 namespace stan {
@@ -61,6 +66,7 @@ return_type_t<T_y, T_x, T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
   using Eigen::Array;
   using Eigen::Dynamic;
   using Eigen::Matrix;
+  using Eigen::VectorXd;
 
   const size_t N = x.rows();
   const size_t M = x.cols();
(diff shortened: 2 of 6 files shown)


stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
```diff
@@ -92,56 +98,149 @@ return_type_t<T_y, T_x, T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
...
-        = inv_sigma * y_minus_mu_over_sigma;
-    if (!is_constant_all<T_y>::value) {
-      ops_partials.edge1_.partials_ = -mu_derivative;
+  // calculate this depends on
+  // template parameters
+
+#ifdef STAN_OPENCL
+  bool use_opencl = N * M * opencl_context.tuning_opts().normal_id_glm_coeff1
+                        + N * opencl_context.tuning_opts().normal_id_glm_coeff2
+                    > 1;
```
SteveBronder: (unrelated to this PR) I'm starting to think we should get rid of the tuning params. Do we have anyone / anytime to make a tuner? If not then I'd be for a separate PR ripping it out

t4c1: @rok-cesnovar is already working on a autotuning.

rok-cesnovar:
Thanks! Really excited for this GPU GLMs.

This is the first GPU GLM PR so this one should be reviewed to some extent by more developers. GPU GLMs that will follow will then just have to be done by the principles we agree on here and will be easier to review.

A few notes before going in to make a deeper code review:
- This changes the `stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp` for cases even when `STAN_OPENCL` is not set. Not saying it should not but the reasons behind it should written down along with a quick explanation of the changes should be put in the top-level comment. Also dont include the `/opencl` headers if `STAN_OPENCL` is not set.

- The CPU tests for the GLMs should always run. The way its setup now on this PR its either GPU tests if the flag is set or CPU if its not. This would decrease our test coverage here as CPU GLMs on Mac would not get tested ( the GPU tests are done on a Mac machine ). This will duplicate test code but thats better than worse test coverage. We really need to make a OpenCL test system like Bob did with the ad_test solution. But that is another issue/PR.

- Could we move the relevant non-kernel GPU code in `/opencl/prim/prob` or something like that? And just make the regular function call the `/opencl` one if the conditions are met? I am working on a branch that will reorganize the OpenCL use in prim and rev to keep the OpenCL in /prim and /rev to a minimum. That will help us in the long run with the goal of reducing transfers by replacing `/prim` and `/rev` calls with `/opencl`

- Please explain the tunings opts and how you got to the values.

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+// \cond
+static const char* normal_id_glm_kernel_code = STRINGIFY(
+    // \endcond
+    /**
+     * GPU implementation of Generalized Linear Model (GLM)
+     * with Normal distribution and identity link function.
+     *
+     * Must be run with at least N threads and local size equal to LOCAL_SIZE_.
+     * @param[in] y_glob vector parameter
```
SteveBronder: glob?

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,162 @@
...
+     * @param need_log_sigma_sum interpreted as boolean - whether log_sigma_sum
+     * needs to be computed
+     */
+    __kernel void normal_id_glm(
+        const __global double* y_glob, const __global double* x,
+        const __global double* alpha, const __global double* beta,
+        const __global double* sigma_glob, __global double* mu_derivative_glob,
+        __global double* mu_derivative_sum,
+        __global double* y_minus_mu_over_sigma_squared_sum,
+        __global double* sigma_derivative, __global double* log_sigma_sum,
```
SteveBronder: I'd rather have out parameters first

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,162 @@
...
+        const int N, const int M, const int is_alpha_vector,
+        const int is_sigma_vector, const int need_mu_derivative,
+        const int need_mu_derivative_sum, const int need_sigma_derivative,
+        const int need_log_sigma_sum) {
```
SteveBronder: I'd rather have out parameters first

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: Short for global. I can change the variable name if you want.

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
SteveBronder: And for names 


y_minus_mu_over_sigma_squared_sum: 
- y_scaled
  - Doc: The (independent variable) scaled by it's mean and variance.

sigma_derivative
- sigma_d
  - Doc: Derivative with respect to sigma

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
t4c1: @rok-cesnovar is already working on autotuning.

rok-cesnovar: Yes, autotuning is a pet project of mine that should see the light of day someday along with my PhD. But that damn thing is not willing to converge :)

The way we use "tuning opts" right now is actually just as a way of selecting whether to execute on the CPU or the GPU. We have circled around this topic at Stancon  the general thought is that we dont really need to be so smart about selecting the CPU or GPU. 
So I agree with you that we should throw these out and if a user specifies `STAN_OPENCL` all OpenCL supported functions will get executed on the GPU. 

This helps us with: 
- support (there were already a few users that set `STAN_OPENCL` but saw no GPU activity and that was confusing to most)
- we dont need to search for the CPU/GPU breakpoint on OpenCL PRs and test with different hardware to find an arbitrary limit that we have no way of evaluating
- cleaner code when we include OpenCL in /prim and /rev

Its the same with MPI and threading. If the overhead is not beaten by parallelization for the input size you will not see a speedup and that is just the way it is. And those that use it (should) expect it.

The autotuning I am working on will adapt the `OpenCL` implementation parameters to tune it to the device we are executing on. MCMC applications are perfect for this with the 1000+ of iterations of the same code. I was also working on a system that would help with this autoselect but I have not found a nice solution for it and bailed on that after Stancon due to the stuff mentioned above.

rok-cesnovar: So I would get rid of the "tuning opts" that are just use for selecting when to switch from CPU to GPU.

t4c1:
> Could we move the relevant non-kernel GPU code in /opencl/prim/prob or something like that? 

I was trying to implement this in same way as other prim/rev functions that call OpenCL if the problem is large enough. It just turns out that GLMs are more complex due to high number of  template arguments so there is more extra code if it runs on OpenCL.

I see multiple options how to separate code that handles OpenCL stuff, but I don't really like any of them.

a) Just put kernel call in separate function. This will not do much to shorten OpenCL specific code and will have a huge number of arguments. 

b) Pass `matrix_cl`s, `need_ ...` variables and vector lenghts in. Return `operands_and_partials` and logp.  Still a huge number of arguments. 

c) Separate `normal_id_glm_lpdf_opencl()` function with same templates and signature as `normal_id_glm_lpdf()`. Here we have some code duplication between CPU and OpenCL implementations. 

d) Put everything that is within `#ifdef STAN_OPENCL` in separate function. That would probably make the most readable code. However in this case it is hard to define, what exactley the new function does and what it does not do.

t4c1:
> Please explain the tunings opts and how you got to the values.

Calculations in GLMs can be separated in two parts. The first is matrix-vector multiplication. The time for that is proportional to number of instances times number of attributes. The weight for this part is `normal_id_glm_coeff1`. The second part are any operations that are done on result in this multiplication. The time for those is proportional to number of instances. The weight for this part is `normal_id_glm_coeff2`. 

I wrote a program that measures GPU speedup for many problem sizes. It saves the dimensions of cases where the speedup was close to 1. Than I least-squares fitted the coefficients to these measurements. I ran measurements on Core i5 2500 and GTX 1070. Benchmark and the script for fitting (for all GLMs) are here: https://gist.github.com/t4c1/37958d2afcf6ec8d1282be4164f5c507

The graph below shows that data fits this model quite nicely (except for neg_binomial). Lines on graph are threshold defined by coefficients and croses are benchmarked dimensions that had speedup close to 1.

thresholds

t4c1:
> The CPU tests for the GLMs should always run.

I will tackle this after we decide how to split the OpenCL part in a separate function.

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: > I'd rather have out parameters first

Hmm, so far we have been inconsistent. I will change it here, but there are still other kernels that have inputs before outputs.

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
SteveBronder: tbh I think it's fine to have y here

t4c1: Right. I thing I copied this from some other GLM that had y in local variable too.

rok-cesnovar:
I dug in a little bit more and now I understand why the call to OpenCL is not going to be as straightforward. There really isnt a nice solution here...
How about: 

a) have an `ifndef STAN_OPENCL` in the prim/mat/prob` and a `ifdef STAN_OPENCL` in /opencl/prim/prob.

b) have them same function in the stan::math::opencl namespace that is then called from the `prim/mat/prob` if STAN_OPENCL is defined.This does duplicate code but the code is also duplicated the way it is now just in the same function.

t4c1:
Now very little code is duplicated. Maybe a few if sentence header that decide whether certain partial derivatives need to be computed. However even some of these are different in OpenCL and CPU code.

Doing your a) would also duplicate all checks, using, value_of, as_column_vector_or_scalar calls, logp and some other computation. In total around 100 lines of code.

Your b) could duplicate less code, depending on what way we do it. Options I see are written in my earlier post [here](https://github.com/stan-dev/math/pull/1333#issuecomment-527086290).

SteveBronder:
Is this ready for review? I can look it over tmrw

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: > And for names

I think having names consistent with CPU implementation is preferable. Anyway I think whole `derivative` is not too loong for variable names. `y_minus_mu_over_sigma_squared_sum` could be shortened, but should be kept consistent with CPU implementation. 

I tried to also keep docs consistent. Imo they could also be improved. However your suggestion for `y_minus_mu_over_sigma_squared_sum` is wrong. I tried to incorporate the idea of your suggestion anyway.

t4c1:
Now it is ready. I talked with @rok-cesnovar and we decided not to split off any part of the implementation in this PR. Maybe at a point when we do `opencl/rev`.

SteveBronder:
Cool! I'll give this a readthrough today

SteveBronder:
@serban-nicusor-toptal this PR is also not able to have clang-format changes pushed to it, @seantalts @syclik @bob-carpenter do either of you have the email username/pw for @stan-buildbot? It should have email inviting it to contribute to the bstatcomp math fork.

I'm gonna try to figure out why this is happening in general, the github permissions are on for owners of the main repo to make commits to the fork.

syclik:
I don't have the password to @stan-buildbot. It's been changed since I set up that account.

I was able to see all the notifications, though, by logging into the registered email. That doesn't help without a valid login on that account.

serban-nicusor-toptal:
Hey Steve, I don't have them either. Sorry

SteveBronder:
> I was able to see all the notifications, though, by logging into the registered email. That doesn't help without a valid login on that account.

If no one has the password could we change it so we know it? I think the only overhead there is adding the PW to the Jenkins instance

serban-nicusor-toptal:
I have reversed the encryption methods and decrypted the pass, will send you an email shortly.

SteveBronder:
@serban-nicusor-toptal much appreciated!!!

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
```diff
@@ -92,20 +100,101 @@ return_type_t<T_y, T_x, T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
...
+
+  double y_minus_mu_over_sigma_squared_sum;  // the most efficient way to
+  // calculate this depends on
+  // template parameters
 
-  Array<T_partials_return, Dynamic, 1> y_minus_mu_over_sigma
-      = x_val * beta_val_vec;
+#ifdef STAN_OPENCL
+  const int local_size
+      = opencl_kernels::normal_id_glm.make_functor.get_opts().at("LOCAL_SIZE_");
```
SteveBronder: Can you fix the kernel_cl class so we don't need to grab the local size out of the make_functor?

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
```diff
@@ -92,20 +100,101 @@ return_type_t<T_y, T_x, T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
...
+  matrix_cl<double> mu_derivative_cl(need_mu_derivative ? N : 0, 1);
+  const bool need_mu_derivative_sum
+      = !is_constant_all<T_alpha>::value && !is_vector<T_alpha>::value;
+  matrix_cl<double> mu_derivative_sum_cl(need_mu_derivative_sum ? wgs : 0, 1);
+  matrix_cl<double> y_minus_mu_over_sigma_squared_sum_cl(wgs, 1);
+  const bool need_sigma_derivative
+      = !is_constant_all<T_scale>::value && is_vector<T_scale>::value;
+  matrix_cl<double> sigma_derivative_cl(need_sigma_derivative ? N : 0, 1);
+  const bool need_log_sigma_sum
+      = include_summand<propto, T_scale>::value && is_vector<T_scale>::value;
```
SteveBronder: Can you pull all of these out of this ifdef and reuse them in the #else side?

SteveBronder: Also may be good to write a `is_not_constant<>` type trait

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * GPU implementation of Generalized Linear Model (GLM)
+     * with Normal distribution and identity link function.
+     *
+     * Must be run with at least N threads and local size equal to LOCAL_SIZE_.
+     * @param[in] y vector parameter
+     * @param[in] x design matrix
+     * @param[in] alpha intercept (in log odds)
+     * @param[in] beta weight vector
+     * @param[in] sigma_glob (Sequence of) scale parameters for the normal
```
SteveBronder: more meaningful name than glob

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+     * @param[out] mu_derivative_glob intermediate variable used in the model
+     * @param[out] mu_derivative_sum partially summed mu_derivative_glob (1
+     * value per work group)
+     * @param[out] y_minus_mu_over_sigma_squared_sum Partial sum of squares of
```
SteveBronder: `y_normalized_sq_sum`

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+        const __global double* sigma_glob, const int N, const int M,
+        const int is_alpha_vector, const int is_sigma_vector,
+        const int need_mu_derivative, const int need_mu_derivative_sum,
+        const int need_sigma_derivative, const int need_log_sigma_sum) {
+      const int gid = get_global_id(0);
+      const int lid = get_local_id(0);
+      const int lsize = get_local_size(0);
+      const int wgid = get_group_id(0);
+
+      __local double res_loc[LOCAL_SIZE_];
```
SteveBronder: What does res_loc stand for?

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+    __kernel void normal_id_glm(
+        __global double* mu_derivative_glob, __global double* mu_derivative_sum,
+        __global double* y_minus_mu_over_sigma_squared_sum,
+        __global double* sigma_derivative, __global double* log_sigma_sum,
+        const __global double* y, const __global double* x,
+        const __global double* alpha, const __global double* beta,
+        const __global double* sigma_glob, const int N, const int M,
+        const int is_alpha_vector, const int is_sigma_vector,
+        const int need_mu_derivative, const int need_mu_derivative_sum,
+        const int need_sigma_derivative, const int need_log_sigma_sum) {
```
SteveBronder: This is pretty hard to review with so many paths in one function. We should think of a nicer way to do this

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+          barrier(CLK_LOCAL_MEM_FENCE);
+        }
+        if (lid == 0) {
+          mu_derivative_sum[wgid] = res_loc[0];
+        }
+      }
+
+      if (need_log_sigma_sum) {
+        barrier(CLK_LOCAL_MEM_FENCE);
+        res_loc[lid] = log_sigma;
```
SteveBronder: log_sigma is only used here why is it calculated so far above?

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
SteveBronder: I thnk it would be better to change this to `local_storage`

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+     * @param need_mu_derivative_sum interpreted as boolean - whether
+     * mu_derivative_sum needs to be computed
+     * @param need_sigma_derivative interpreted as boolean - whether
+     * sigma_derivative needs to be computed
+     * @param need_log_sigma_sum interpreted as boolean - whether log_sigma_sum
+     * needs to be computed
+     */
+    __kernel void normal_id_glm(
+        __global double* mu_derivative_glob, __global double* mu_derivative_sum,
+        __global double* y_minus_mu_over_sigma_squared_sum,
```
SteveBronder: Can we change this here and in the glm.hpp to be `y_scaled_sq_sum`

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+        const __global double* y, const __global double* x,
+        const __global double* alpha, const __global double* beta,
+        const __global double* sigma_glob, const int N, const int M,
+        const int is_alpha_vector, const int is_sigma_vector,
+        const int need_mu_derivative, const int need_mu_derivative_sum,
+        const int need_sigma_derivative, const int need_log_sigma_sum) {
+      const int gid = get_global_id(0);
+      const int lid = get_local_id(0);
+      const int lsize = get_local_size(0);
+      const int wgid = get_group_id(0);
```
SteveBronder: wg_id

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
```diff
@@ -0,0 +1,161 @@
...
+     * needs to be computed
+     * @param need_mu_derivative_sum interpreted as boolean - whether
+     * mu_derivative_sum needs to be computed
+     * @param need_sigma_derivative interpreted as boolean - whether
+     * sigma_derivative needs to be computed
+     * @param need_log_sigma_sum interpreted as boolean - whether log_sigma_sum
+     * needs to be computed
+     */
+    __kernel void normal_id_glm(
+        __global double* mu_derivative_glob, __global double* mu_derivative_sum,
```
SteveBronder: More meaningful name than glob

SteveBronder: Above is for all `globs`

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: If we refactor this kernel into multiple kernels we suffer performance penalty for loading data multiple times. If we refactor it into multiple functions within the same kernel we still have to pass all these arguments to the kernel. 

Maybe we could organize the arguments into some structs, but I don't see a nice way to do it.

SteveBronder: Yeah I recognize that, though I think the main performance we would lose is some function invocations and a few places where we can reuse local memory. Would it be significant for you to break this up into smaller kernels? If the performance penalty is ballpark less than 5% than I would feel a lot more comfortable with future maintanence

t4c1: I'm pretty sure I can construct a case where penalty would be close to 2 for splitting this into two kernels (low number of attributes, large number of cases), so I would not want to go that way.

I hope we will eventually be able to use kernel generator to replace these complicated GLM kernels, but we are not there yet.

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
t4c1: You mean in CPU code? The problem is they are not exactley the same as conditions for branches on CPU side. I sure can reuse them for those branches in GPU version. I will see what I can do for CPU side.

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
rok-cesnovar: @SteveBronder The thing is these kernels are memory transfer bound and we are dealing with execution times at the level of sub miliseconds even for large inputs. 

For stuff like cholesky, multiply, mdivide etc, it doesnt really make a difference if you extract something out because execution times are a few orders of magnitude larger there. Here its actually a huge difference. Its also why Tadej is working on the kernel generator.

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
t4c1: Actually I could only reuse `mu_derivative_sum_cl` on GPU side and none on CPU. Other `need_`... variables are only used for kernel arguments and not reused later.

t4c1: > Also may be good to write a is_not_constant<> type trait

I don't think we need a separate type trait for every thing we use - only of things that are tedious to write or complicated to understand. A simple `!` is neither IMHO. Anyway, the new trait would have to be named `is_not_constant_any`, which is longer than `!` ;).

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
SteveBronder: > but there are still other kernels that have inputs before outputs.

Let's make a separate PR to correct those later. It's nice to be consistent with in/out order

> I think having names consistent with CPU implementation is preferable. Anyway I think whole derivative is not too loong for variable names. y_minus_mu_over_sigma_squared_sum could be shortened, but should be kept consistent with CPU implementation.

Yeah maybe derivative is fine. I think it's fine to change the names in both here and the CPU implimentation

> I tried to also keep docs consistent. Imo they could also be improved. However your suggestion for y_minus_mu_over_sigma_squared_sum is wrong. I tried to incorporate the idea of your suggestion anyway.

In R calling `scale` is doing `(y - mean(y)) / sd(y)`. Is it not common parlance to call that the scaled value?

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
SteveBronder: lol alright fair

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: Also a bad suggestion - `y_minus_mu_over_sigma_squared_sum = sum((y - mu) / (sigma * sigma))`. Maybe just `y_normalized_sum`?

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
SteveBronder: Alright I'll trust yinz judgement here let's just make sure we doc this all well

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: It stands for local result. But I agree with your suggestion.

stan/math/prim/mat/prob/normal_id_glm_lpdf.hpp
rok-cesnovar: This was a topic of discussion already, we opted with going with ! : https://github.com/stan-dev/math/pull/1234#issuecomment-504744484

stan/math/opencl/kernels/normal_id_glm_lpdf.hpp
t4c1: It is calculated within `if (gid < N) {`. I did not want to duplicate this if.

PR #1335, opened 2019-09-03
clang-tidy auto fixes for braces, using, nullptr and braced return

SteveBronder:
## Summary

Running clang-tidy with the checks below. This refactor is only aesthetic and does not change how any of the code is executed.

[readability-braces-around-statements](https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html):
- Adds braces around conditionals and ifs

[modernize-use-using](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-using.html):
everywhere clang-tidy can replace a `typedef` with a `using` it will aka
`typedef double A;` becomes `using A = double;`

[modernize-return-braced-init-list](https://clang.llvm.org/extra/clang-tidy/checks/modernize-return-braced-init-list.html):
> Replaces explicit calls to the constructor in a return with a braced initializer list. This way the return type is not needlessly duplicated in the function definition and the return statement.

```cpp
Foo bar() {
  Baz baz;
  return Foo(baz);
}

// transforms to:

Foo bar() {
  Baz baz;
  return {baz};
}

```

[modernize-use-nullptr](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nullptr.html):
using `nullptr` instead of NULL

## Tests

refactor is only aesthetic so no new tests

## Checklist

- [x] Math issue #1308


SteveBronder:
@serban-nicusor-toptal 

Distribution tests have been running for about 6 hours now with the error below 

> + ./runTests.py -j25 test/prob 

> Cannot contact i-08bc3a088b9f6c8ef: hudson.remoting.ChannelClosedException: Channel "unknown": Remote call on i-08bc3a088b9f6c8ef failed. The channel is closing down or has closed down

Is there some way I can manually kickoff/restart the machine when these sorts of things happen?

bob-carpenter:
> On Sep 3, 2019, at 12:17 PM, Steve Bronder <notifications@github.com> wrote:
> 
> Summary
> 
> Running clang-tidy with the following checks adds these new additions
> 
> readability-braces-around-statements:
> 
> 	• Adds braces around conditionals and ifs

-1.  I hate the extra braces and vertical line.  But I'll go along if this
is the new standard.

> modernize-use-using:
> everywhere clang-tidy thinks it can safely swap out typedef for using it will

That seems really unsafe given scoping.  Can you explain more what it does and
what a potential change is going to look like?  Will it work properly with
argument-dependent lookup?


> modernize-return-braced-init-list:
> 
> Replaces explicit calls to the constructor in a return with a braced initializer list. This way the return type is not needlessly duplicated in the function definition and the return statement.
> 
> Foo bar() {
>   Baz baz;
>   return Foo(baz);
> }
> 
> 
> // transforms to:
> 
> Foo bar() {
>   Baz baz;
>   return {baz};
> }
> 

This looks very strange, but I could get used to it.  I find all
this auto stuff a tradeoff between easy understandability and
conciseness.


> modernize-use-nullptr:
> using nullptr instead of NULL

Good.  We shouldn't be using NULL anywhere.


> Tests
> 
> refactor so no new tests

We've been trying to add missing tests when we refactor.  But nobody's going
to be able to read through hundreds of munged files, which is why we need
to know the general replacements being done here and how safe they are.

serban-nicusor-toptal:
Hey @SteveBronder seems like the EC2 instance has been shut down, maybe an outbid on the spot instances.  Should rarely happen tho, will check if the prices have increased.  

For how to restart a job, see all the pull requests jobs [here](https://jenkins.mc-stan.org/job/Math%20Pipeline/view/change-requests/)
Pick yours and just do `Build with parameters` [here](https://jenkins.mc-stan.org/job/Math%20Pipeline/job/PR-1335/)

rok-cesnovar:
> That seems really unsafe given scoping. Can you explain more what it does and what a potential change is going to look like? Will it work properly with argument-dependent lookup?

In my meta refactor PRs I replaced some of the typedefs with using for type aliases. I think its considered "modern C++". Is it really an issue for argument-dependent lookup?

Example from https://www.internalpointers.com/post/differences-between-using-and-typedef-modern-c
`typedef std::map<std::string, std::vector<std::string>> Map;`
is equivalent to 
`using Map   = std::map<std::string, std::vector<std::string>>;`

The advantage of `using` is that it the syntax for cases like the following: 
```
template <typename T>
using str_map<T> = std::map<std::string, T>;
```
You can do the same with typedef but the syntax is less clean: 
```
template <typename T>
struct str_map {
    typename std::map<std::string, T> type;
};
```

A lot of the meta template metaprogramming in Stan Math could get cleaned up by this if we want to do it. I was planning on doing this but held off waiting for the meta flattening. But it seems that Steve will be able to do the work for me with a script :)

bob-carpenter:
Never mind.  This PR is doing the opposite of what I thought it was doing.  It's replacing `typedef`s with `using` statements not vice-versa.  E.g.,

```
typedef stan::math::fvar<T> ReturnType;
```

is being replaced with

```
using ReturnType = stan::math::fvar<T>;
```

but

```
  using std::exp;
```

is left unchanged.

SteveBronder:
> -1.  I hate the extra braces and vertical line.  But I'll go along if this
is the new standard.

There's more than one place where we do something like

```cpp
if (blah)
  for (blahblah) {
// doing stuff
}
```

imo I usually have to read that sort of thing twice.

More vertical space and indents can be a bit annoying but I do like that scope is given a bit nicer

> Never mind. This PR is doing the opposite of what I thought it was doing. It's replacing typedefs with using statements not vice-versa. E.g.,

Yes sorry for the poor wording, I updated it in the docs

> (the braced init lists) This looks very strange, but I could get used to it.  I find all
this auto stuff a tradeoff between easy understandability and
conciseness.

Yeah, it's more "dry" but I'm sort of indifferent one way or the other on this one

> We've been trying to add missing tests when we refactor.  But nobody's going
to be able to read through hundreds of munged files, which is why we need
to know the general replacements being done here and how safe they are.

This refactor shouldn't change how any code is executed since they changes are only aesthetic. Is there any more info you need on this one that is not provided in the summary at the top? I edited it just now to include links to the clang-tidy check descriptions

bob-carpenter:
Can that misalignment of the last closing brace happen in clang-format?  If so, that's too bad.

In any case, I agree thta we should have braces in cases like this conditional, even though they're not technically necessary:

```
if (a > b) {
  for (int i = 0; i < I; ++i) {
  ...
  }
}
```

What will this do to a one-liner like:

```
if (a > b) c = 3;
```

Will it just add braces

```
if (a > b) { c = 3; }
```

or will it turn it into a three-liner:

```
if (a > b) {
  c = 3;
}
```

Anyway, I don't care about this enough to have a long dragged-out discussion of it if we can make whatever you want to do happen automatically.  I'll start a new project if I want to get my way on everything. :-)

SteveBronder:
> Anyway, I don't care about this enough to have a long dragged-out discussion of it if we can make whatever you want to do happen automatically. I'll start a new project if I want to get my way on everything. :-)

Stan' 4: Electric Boogaloo

This script takes like 5 minutes to run so I'm gonna open up a new branch with separate commits for each. So if we are like "Actually I hate this" it's pretty easy to revert out

Issue #1336, opened 2019-09-04
Update OpenCL Headers


SteveBronder:
## Description

The Header file for OpenCL is about 2 years old and I'd like to update to the newer one with bug fixes. The PR associated with this issue *does* use the OpenCL 2.1 headers, but with compiler flags that tell it to give us the OpenCL 1.2 API. This also shrinks our OpenCL headers down from about a dozen files to just one

You can see the commit log here of the header

https://github.com/KhronosGroup/OpenCL-CLHPP/commits/master

#### Current Version:
v2.20.0

PR #1337, opened 2019-09-04
Update OpenCL Headers

SteveBronder:
## Summary

This updates the header files for OpenCL. The version of the OpenCL standard we use stays the same (1.2) while the headers include several deprecations and bug fixes from previous versions. 

The changes for this are

1. Old OpenCL headers were deleted and replaced with the new one from [here](https://github.com/KhronosGroup/OpenCL-CLHPP)

2. Since we only have one header file now the `#include <Cl/cl.hpp>` includes are replaced with `#include <cl.hpp>`

3. The flags `-DCL_HPP_TARGET_OPENCL_VERSION=120 -DCL_HPP_MINIMUM_OPENCL_VERSION=120` tell the OpenCL header to only include the 1.2 API.

4. The kernel constructor now takes in an `std::vector<std::string>` instead of `std::vector<const char*>`. This is how the new API is defined. We were getting away with the `const char*` one because it was a static constant pointer.

5. `to_size_t` function was removed. OpenCL now uses `std::arrays` to replace their own `cl::size_t` arrays.

## Tests

Maintenance PR so I assume our current test suite is fine

## Side Effects

Nope!

## Checklist

- [x] Math issue #1336


make/compiler_flags
```diff
@@ -151,7 +151,9 @@ ifdef STAN_OPENCL
   OPENCL_DEVICE_ID ?= 0
   OPENCL_PLATFORM_ID ?= 0
 
-  CPPFLAGS_OPENCL ?= -DSTAN_OPENCL -DOPENCL_DEVICE_ID=$(OPENCL_DEVICE_ID) -DOPENCL_PLATFORM_ID=$(OPENCL_PLATFORM_ID) -DCL_USE_DEPRECATED_OPENCL_1_2_APIS -D__CL_ENABLE_EXCEPTIONS -Wno-ignored-attributes
+  CPPFLAGS_OPENCL ?= -DSTAN_OPENCL -DOPENCL_DEVICE_ID=$(OPENCL_DEVICE_ID) -DOPENCL_PLATFORM_ID=$(OPENCL_PLATFORM_ID)
+  CPPFLAGS_OPENCL += -DCL_HPP_TARGET_OPENCL_VERSION=120 -DCL_HPP_MINIMUM_OPENCL_VERSION=120
+  CPPFLAGS_OPENCL += -DCL_HPP_ENABLE_EXCEPTIONS -Wno-ignored-attributes
```
rok-cesnovar: Thanks for splitting these up.

rok-cesnovar:
Github was having some hiccups I guess, stating that the automatic clang-format bot got a 403 error pushing to the bstatcomp repo. So I went to format locally, it seems we were both thinking the same thing :). I will review this since its not a lot of changes and then go to the meta refactor.

stan/math/opencl/opencl_context.hpp
```diff
@@ -32,35 +32,7 @@
...
- * @return the cl::size_t<3> converted from the input array
- */
-template <>
-inline cl::size_t<3> to_size_t(const size_t (&values)[3]) {
-  cl::size_t<3> s;
-  for (size_t i = 0; i < 3; i++)
-    s[i] = values[i];
-  return s;
-}
-}  // namespace opencl
```
rok-cesnovar: :raised_hands: :raised_hands: :raised_hands:

I hated that we had to do this before. Finally its gone.

SteveBronder: lol same. Part of updating these was because I wanted to zip that out

SteveBronder:
Great minds think alike ;)

Also on a side note I figured out some of the zero copy buffer stuff. See [here](https://github.com/bstatcomp/math/blob/testing-zero/stan/math/opencl/matrix_cl.hpp#L357) for using host pointer and [here](https://github.com/bstatcomp/math/blob/testing-zero/stan/math/opencl/copy.hpp#L152) for getting the data from the device (in the case of CPU it's a no-op I think?). Just doing it for the packed vectors in the Cholesky code made the rev test run about 15% faster. Which, like is a lot since that's such a small part of the function. Previous benchmarking showed me that like 60% of our time is spent transferring data so that speedup does make sense to me.

But the bad news is that we can't use this a lot of the time. OpenCL needs a non constant void pointer as the host pointer, but all of our functions bring in things via const ref. I have some schemes for that but for now the zero copy stuff can go back on the burner

rok-cesnovar:
> Also on a side note I figured out some of the zero copy buffer stuff.

Thank you for this investigation. Seems really promising. But agree, back to the burner for now.

Issue #1115, opened 2019-02-10
Solve algebraic equations with Newton solver


charlesm93:
Accidently created this as a Stan issue (stan#2735), but it seems it should (for the moment) be math issue. Reposting the issue is. There is however valuable conversation in the linked issue.

#### Summary:
Provide a new function to solve systems of nonlinear algebraic equations using Newton's method. This complements our current solver, which uses Powell's dogleg method. The Newton solver has been shown to be much faster in several examples.

The name of the function is `algebra_solver_newton()`. Just like `algebra_solver()`, it returns the root of a function specified by the user. See the details for the current algebraic solver. It can have the same signature. The more delicate task is exploiting parallelism. Here we need to figure out which additional tuning parameters we will make accessible to the user.

#### Description:
Newton's solver is fast when applied to convex optimization problems (i.e. finding the root of a convex function's derivative). One key application is finding the mode of a Laplace approximation -- here we want to solve a convex optimization problem with thousands of parameters. With 100 parameters, I got a 10-fold speed up. Other examples have been reported. A discussion can be found on discourse: https://discourse.mc-stan.org/t/newton-solver-in-stan/6684.

There is already a Newton solver, used for ADVI, but it is not exposed to the Stan language. Furthermore, it is not designed to handle large systems. KINSOL, from the CVODES suite, is a promising candidate, which allows for parallelization. See https://computation.llnl.gov/projects/sundials/toms_sundials.pdf. The parallelization happens at a vector operation level. CVODES provides a parallel vector routine. Since we rely on the implicit function theorem for the Jacobian, we do not need to propagate sensitivities through the solver and can use CVODES parallel vector routines. 

Beyond that, I'm not clear on how to incorporate this parallelization inside Stan. Can we do something akin to `map()` and MPI?

Other features of interest: the user can specify a step size, or get an adaptive step size. The latter doesn't always help, since it is costly to find the optimal step size. But, according to KINSOL's documentation, it makes a difference when we are "far" from the solution. This is likely to occur in high dimensions, where, I suspect, we'll be subject to a curse of dimensionality.

Note we would rename the current `algebra_solver()` `algebra_solver_powell()`. We need to run tests to better understand which solver works best when. Two important examples to consider: steady states in pharmacometrics, and constructing priors (something @betanalpha told me about). 

#### Reproducible Steps:
See the issues posted on the forum.

#### Expected Output:
A faster algebraic solver applicable to large systems.

#### Additional Information:
NA
#### Current Version:
[...]

PR #1339, opened 2019-09-05
Feature/issue 1115 newton solver

charlesm93:
## Summary

Provide a new function to solve systems of nonlinear algebraic equations using Newton-Krylov methods. This solver is faster, and for certain problems much faster, than our current dogleg solver. The solver is based on the KINSOL routine from the SUNDIALS suite. It uses reverse-mode AD to compute quasi Newton steps (i.e. the Jacobian is re-evaluated only when needed), and a line search scheme to optimize the Newton step size.

The function is `algebra_solver_newton()`. Its signature is identical to that of `algebra_solver`, except for one of the tuning parameters. The user can specify (i) the scaled step size, which stops the algorithm if too little progress is made, (ii) the function tolerance, used to measure if the solution is a root, and (iii) the max number of steps, which stops the algorithm after a certain number of iterations. The function is built on top of `kinsol_solve()`, which gives the user more control over tuning parameters (but will not be exposed to the Stan language).

Throws the same exceptions as `algebra_solver()`. In addition, an exception gets thrown when KINSOL returns a negative flag, via the `check_flag` function. For more details about the KINSOL solver, see their [manual](https://computing.llnl.gov/sites/default/files/public/kin_guide.pdf).

## Tests

The unit tests for the algebraic solver are generalized and now run on both the current solver and the Newton solver. See `test/unit/math/rev/mat/functor/algebra_solver_test.cpp` and `test/unit/math/rev/mat/functor/util_algebra_solver.hpp`.

## Side Effects

The KINSOL module has been added to the SUNDIALS library.

## Checklist

- [x] Math issue #1115 

- [x] Copyright holder: Charles Margossian, Columbia University

- [] the basic tests are passing (did the tests that run quickly on my machine)

    - [x] unit test: `./runTests.py test/unit/math/rev/mat/functor/algebra_solver_test.cpp`
    - [] unit tests pass (to run, use: `./runTests.py test/unit`)
    - [] header checks pass, (`make test-headers`)
    - [] docs build, (`make doxygen`)
    - [x] code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)


charlesm93:
Note: this looks like a bigger pull request than it is, because of the KINSOL files that are added to the SUNDIAL lib.

yizhang-yiz:
@charlesm93 
Could you run `make cpplint` to make sure linting is passed? Thanks.

test/unit/math/rev/mat/functor/util_algebra_solver.hpp
```diff
@@ -1,12 +1,35 @@
...
 #include <sstream>
 #include <vector>
 #include <limits>
 #include <string>
 
-/* define algebraic functions which get solved */
+/* wrapper function that either calls the dogleg or the newton solver. */
+template <typename F, typename T1, typename T2>
+Eigen::Matrix<T2, Eigen::Dynamic, 1> general_algebra_solver(
+    bool is_newton, const F& f, const Eigen::Matrix<T1, Eigen::Dynamic, 1>& x,
```
yizhang-yiz: To me it's not a good idea to run different solvers in a single unit test, as this is against the point of "unit test"(I know that I violate this rule sometimes, shame to me).  It's less intuitive to figure out which solver fails if `is_newton` loop reports failure in one of your tests. So it's better to separate than lump: `TEST(xxx, dogleg)` and `TEST(xxx, kinsol)`. To avoid repeating yourself you can take advantage of [test fixture](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests).

Another advantage of separate testing + test fixture is performance comparison, since that's one of the reasons why we add Newton.

stan/math/rev/mat/functor/kinsol_solve.hpp
```diff
@@ -0,0 +1,126 @@
...
+ * @throw <code>boost::math::evaluation_error</code> if Kinsol returns a
+ *        negative flag after attempting to solve the equation.
+ */
+template <typename F1, typename F2 = kinsol_J_f>
+Eigen::VectorXd kinsol_solve(
+    const F1& f, const Eigen::VectorXd& x, const Eigen::VectorXd& y,
+    const std::vector<double>& dat, const std::vector<int>& dat_int,
+    std::ostream* msgs = nullptr, double scaling_step_tol = 1e-3,
+    double function_tolerance = 1e-6,
+    long int max_num_steps = 1e+3,  // NOLINT(runtime/int)
```
yizhang-yiz: First question, where do you set `max_num_steps`? I looked for `KINSetNumMaxIters` but couldn't find it. Without setting it the later `check_flag_kinsol` would be reporting wrong results.

Second question, why set max iter to 1000 instead of using default 200? A grief I have on ODE solvers is that many control parameters are out of thin air without documented justification.

stan/math/rev/mat/functor/kinsol_solve.hpp
```diff
@@ -0,0 +1,126 @@
...
+               "KINSetJacFn");
+
+  N_Vector nv_x = N_VNew_Serial(N);
+  realtype* nv_x_data = N_VGetArrayPointer_Serial(nv_x);
+  for (int i = 0; i < N; i++)
+    nv_x_data[i] = x(i);
+
+  check_flag_kinsol(
+      KINSol(kinsol_memory, nv_x, global_line_search, scaling, scaling),
+      max_num_steps);
```
yizhang-yiz: see earlier comment of `max_num_steps`.

stan/math/rev/mat/functor/algebra_solver_newton.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * @throw <code>std::invalid_argument</code> if scaled_step_size is strictly
+ * negative.
+ * @throw <code>std::invalid_argument</code> if function_tolerance is strictly
+ * negative.
+ * @throw <code>std::invalid_argument</code> if max_num_steps is not positive.
+ * @throw <code>boost::math::evaluation_error</code> (which is a subclass of
+ * <code>std::runtime_error</code>) if solver exceeds max_num_steps.
+ */
+template <typename F, typename T>
+Eigen::VectorXd algebra_solver_newton(
```
yizhang-yiz: I suggest we stick to method postfix, to have `algebra_solver_powell`, `algebra_solver_newton`. Because one immediate enhancement after this PR is to add `algebra_solver_picard` for fixed point iteration, which may help steady state solver in Torsten significantly.

stan/math/rev/mat/functor/kinsol_solve.hpp
```diff
@@ -0,0 +1,126 @@
...
+ * @param[in] global_line_search does the solver use a global line search?
+ *            If equal to KIN_NONE, no, if KIN_LINESEARCH, yes.
+ * @return x_solution Vector of solutions to the system of equations.
+ * @throw <code>std::invalid_argument</code> if Kinsol returns a negative
+ *        flag when setting up the solver.
+ * @throw <code>boost::math::evaluation_error</code> if Kinsol returns a
+ *        negative flag after attempting to solve the equation.
+ */
+template <typename F1, typename F2 = kinsol_J_f>
+Eigen::VectorXd kinsol_solve(
```
yizhang-yiz: there are simply too many control parameters that may affect the solver behavior, especially consider immediately after this I'd like to add Picard solver in KINSOL(see earlier comment on solver name). So it's a good idea to add a `struct kinsol_control` dedicated for solver control & options. This way in the future the flow in `kinsol_solve` could remain the same even if we decide to add/change controls.

stan/math/rev/mat/functor/algebra_solver.hpp
```diff
@@ -131,152 +129,133 @@ Eigen::VectorXd algebra_solver(
...
+  template <typename F, typename T1, typename T2>
+  Eigen::Matrix<T2, Eigen::Dynamic, 1> algebra_solver(
+      const F& f, const Eigen::Matrix<T1, Eigen::Dynamic, 1>& x,
+      const Eigen::Matrix<T2, Eigen::Dynamic, 1>& y,
+      const std::vector<double>& dat, const std::vector<int>& dat_int,
+      std::ostream* msgs = nullptr, double relative_tolerance = 1e-10,
+      double function_tolerance = 1e-6,
+      long int max_num_steps = 1e+3) {  // NOLINT(runtime/int)
+    Eigen::VectorXd theta_dbl
+        = algebra_solver(f, x, value_of(y), dat, dat_int, 0, relative_tolerance,
```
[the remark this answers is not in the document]
charlesm93: Done. All tests are done on `algebra_solver_powell()` now, with a basic test for `algebra_solver()`.

stan/math/prim/mat/err/check_flag.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * Throws an exception when a Sundial function fails
+ * (i.e. returns a negative flag)
+ */
+inline void check_flag(int flag, const char* func_name) {
+  if (flag < 0) {
+    std::ostringstream ss;
+    ss << func_name << " failed with error flag " << flag << ".";
+    throw std::runtime_error(ss.str());
+  }
+}
```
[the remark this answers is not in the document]
charlesm93: Both the ODE integrator and algebraic solver now use `check_flag_sundials()`.

stan/math/rev/mat/functor/algebra_solver.hpp
```diff
@@ -131,34 +129,11 @@ Eigen::VectorXd algebra_solver(
...
-    check_finite("algebra_solver", "continuous data", i);
-  }
-  for (int x : dat_int) {
-    check_finite("algebra_solver", "integer data", x);
-  }
-
-  if (relative_tolerance < 0) {
+  algebra_solver_check(x, y, dat, dat_int, function_tolerance, max_num_steps);
+  if (relative_tolerance < 0)
     invalid_argument("algebra_solver", "relative_tolerance,",
```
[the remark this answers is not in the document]
charlesm93: Good catch. Implemented check_nonnegative and check_positive where relevant.

stan/math/rev/mat/functor/kinsol_solve.hpp
charlesm93: Good catch. The default of 1000 is conservative, but in practice, I've never dealt with systems that take more 20 iterations to solve. I'll set the default to 200.

stan/math/rev/mat/functor/kinsol_data.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Calculate the Jacobian of the system function with respect to x
+   * using the method specified by J_f_.
+   * By default, J_f is constructed as a method that computes the
+   * Jacobian with reverse-mode autodiff.
+   */
+  inline int jacobian_states(const double x[], SUNMatrix J) const {
+    return J_f_(f_, x_, y_, dat_, dat_int_, msgs_, x, J);
+  }
```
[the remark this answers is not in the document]
charlesm93: Ok, I see. What you have in mind is:
```
return
      static_cast<const system_data*>(user_data)->
      J_f_(f_, x_, y_, dat_, dat_int_, msgs_, NV_DATA_S(x), J);
```
which currently generates some compilation errors
```
error: invalid use of member 'f_' in static member function
```
and so forth for `x_`, `y_`, etc. Now further down the line, Kinsol requires `kinsol_jacobian` to return a `static_int`. I'll admit static variables are not my expertise, but my inclination is not to pass `f_`, `x_`, etc as static variables. Maybe you have a more informed opinion.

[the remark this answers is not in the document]
yizhang-yiz: From the previous comment, after `user_data` cast to `explicit_system`, you can use it to access object members, accessible through `->`. So instead of defining `jacobian_states`, you can just do
`return explicit_system -> J_f_(explicit_system -> f_, explicit_system -> x_, explicit_system -> y_, ...)` to get Jacobian.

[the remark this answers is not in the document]
charlesm93: Except `explicit_system` is not defined (after we remove the extra layer constructing it). So you would have
```
    return
      static_cast<const system_data*>(user_data)->
         J_f_(static_cast<const system_data*>(user_data)->f_,
              static_cast<const system_data*>(user_data)->x_,
              static_cast<const system_data*>(user_data)->y_,
              static_cast<const system_data*>(user_data)->dat_,
              static_cast<const system_data*>(user_data)->dat_int_,
              static_cast<const system_data*>(user_data)->msgs_,
              NV_DATA_S(x), J);
```
which works.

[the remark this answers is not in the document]
yizhang-yiz: `jacobian_states` is called in `kinsol_jacobian`, where `explicit_system` is created? I don't understand what you mean by `remove the extra layer constructing it`. The extra layer in my comment is function `jacobian_states`. What you can do is
```c++
  static int kinsol_jacobian(N_Vector x, N_Vector f, SUNMatrix J,
                             void* user_data, N_Vector tmp1, N_Vector tmp2) {
    const system_data* alg;
        = static_cast<const system_data*>(user_data);
    return alg->J_f_(alg -> f_, alg -> x_, alg -> y_, ..., NV_DATA_S(x), J);
```
Same for `f_system`.

stan/math/rev/mat/functor/kinsol_data.hpp
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Calculates the root function, using the user-supplied functor
+   * for a given value of x.
+   */
+  inline void f_system(const double x[], double f[]) const {
+    const std::vector<double> x_vec(x, x + N_);
+    const std::vector<double>& f_vec
+        = to_array_1d(f_(to_vector(x_vec), y_, dat_, dat_int_, msgs_));
+    std::move(f_vec.begin(), f_vec.end(), f);
+  }
```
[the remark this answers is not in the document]
charlesm93: See the new implementation of `kinsol_f_system`. Unfortunately, you still need to make one copy of `x`, because you cannot pass an `Eigen::Map` in lieu of an `Eigen::VectorXd` for a custom function. See this [page](https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html):

> All Eigen functions are written to accept Map objects just like other Eigen types. However, when writing your own functions taking Eigen types, this does not happen automatically: a Map type is not identical to its Dense equivalent.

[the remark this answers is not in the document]
andrjohns: You can workaround this using the ```Ref<>``` class, so rather than ```foo(VectorXd)``` you use ```foo(Ref<VectorXd>)```, then ```foo``` will accept ```Map<VectorXd>```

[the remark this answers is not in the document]
yizhang-yiz: @andrjohns I don't think user-defined stan function takes `Eigen::Ref` type.

@charlesm93 better with constructor instead of assignment:
```c++
    Eigen::VectorXd x_eigen(Eigen::VectorXd::Map(NV_DATA_S(x), explicit_system->N_));
    Eigen::Map<Eigen::VectorXd>(N_VGetArrayPointer(f), explicit_system->N_)
      = explicit_system->f_(x_eigen, explicit_system->y_,...                       
```

stan/math/rev/mat/functor/kinsol_solve.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  // construct Linear solver
+  check_flag(KINSetLinearSolver(kinsol_memory, kinsol_data.LS_, kinsol_data.J_),
+             "KINSetLinearSolver");
+
+  if (custom_jacobian)
+    check_flag(KINSetJacFn(kinsol_memory, &system_data::kinsol_jacobian),
+               "KINSetJacFn");
+
+  N_Vector nv_x = N_VNew_Serial(N);
```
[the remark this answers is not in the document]
charlesm93: `nv_x` is removed all together, per your next comment.

stan/math/rev/mat/functor/kinsol_solve.hpp
charlesm93: Your idea is to pass an argument `kinsol_control` to the solver, instead of individual arguments for each tuning option, right? While this simplifies the signature of `kinsol_solve`, it burdens us with a new structure, that will need to be edited as we add new controls. I don't see the benefit of doing this; except perhaps that you can give the structure different constructors and thereby gain some flexibility on which arguments are optional.

yizhang-yiz: As we add Picard solver and possibly sparse solver, this control param thing will likely get out of control. What I'm proposing is to preserve `kinsol_solve` 's solution flow, as it maps to the `skeleton` in the kinsol user guide, and have a separate class to provide algorithm controls, in line with https://refactoring.guru/smells/long-parameter-list

charlesm93: I'm still not quite convinced. It'd be helpful to see how we would take advantage of this when we add the Picard and sparse solver. For now, here's what I'm imagining:
```
str kinsol_control{
  double scaling_step;
  double function_tolerance;
  long int max_num_steps;
  . . .

  set_control(void* kinsol_memory) {
    check_flag_sundials(KINSetNumMaxIters(kinsol_memory, max_num_steps),
                        "KINSetNumMaxIters");
    . . .
}
```
Then we pass a `kinsol_control` object to `kinsol_solve()`.

yizhang-yiz: it's a blocker for this review, so feel free to move on.

charlesm93: I wouldn't say it's a blocker. I do want to better understand what you have in mind (my presumption being that it's a good idea).

yizhang-yiz: Sorry, I meant to say it's not a show-stopper for this PR. So I'm fine we let it play out in the (near) future when I work on Picard solver.

charlesm93: ok. I'm all for step-by-step PRs.

test/unit/math/rev/mat/functor/util_algebra_solver.hpp
charlesm93: > To me it's not a good idea to run different solvers in a single unit test, as this is against the point of "unit test"

Right. My goal was to avoid adding too much code, but I agree we should prioritize on writing "proper" unit tests. Test Fixture helps, even though there is still a lot of redundant code. See the refactored code.

stan/math/rev/mat/functor/algebra_system.hpp
```diff
@@ -151,6 +159,27 @@ struct hybrj_functor_solver : nlo_functor<double> {
...
+template <typename T>
+void algebra_solver_check(const Eigen::Matrix<T, Eigen::Dynamic, 1>& x,
+                          const Eigen::VectorXd y,
+                          const std::vector<double>& dat,
+                          const std::vector<int>& dat_int,
+                          double function_tolerance,
+                          long int max_num_steps) {  // NOLINT(runtime/int)
+  check_nonzero_size("algebra_solver", "initial guess", x);
+  for (int i = 0; i < x.size(); i++)
+    check_finite("algebra_solver", "initial guess", x(i));
```
yizhang-yiz: `check_finite` has matrix version, no need to loop.

test/unit/math/rev/mat/functor/algebra_solver_test.cpp
```diff
@@ -312,3 +339,205 @@ TEST(MathMatrix, system_functor_constructor) {
...
+    theta(k).grad(y_vec, g);
+
+    for (int i = 0; i < n_y; i++)
+      EXPECT_EQ(J(k, i), g[i]);
+  }
+}
+
+TEST_F(algebra_solver_simple_eq_test, newton_tuned) {
+  using stan::math::var;
+  bool is_newton = false;
```
yizhang-yiz: Why `is_newton` false in a newton test?

test/unit/math/rev/mat/functor/algebra_solver_test.cpp
```diff
@@ -312,3 +339,205 @@ TEST(MathMatrix, system_functor_constructor) {
...
+  // using y_scale.
+  // theta = degenerate_test(y_scale, x_guess_2, is_newton);
+  // EXPECT_FLOAT_EQ(5, theta(0));
+  // EXPECT_FLOAT_EQ(5, theta(1));
+  //
+  // theta = degenerate_test(y_scale, x_guess_3, is_newton);
+  // EXPECT_FLOAT_EQ(100, theta(0));
+  // EXPECT_FLOAT_EQ(100, theta(1));
+  //
+  // NOTE: causes Newton solver to send a flag -11 error.
```
yizhang-yiz: Could you investigate this? It doesn't look like something we should ignore.

yizhang-yiz: My guess is it's related to Jacobian calculation with `x_guess_3`. Maybe try removing `KINSetJacFn` to use default DQ method and see if the error persists, and start from there.

charlesm93: The error persists when using DQ. I also tried remove the linesearch step, and forcing the algorithm to evaluate the Jacobian at each step.

charlesm93: See comment here: https://discourse.mc-stan.org/t/newton-solver-in-stan/6684/68.

If the initial guess is (80, 80) the solver converges to one of the roots.

test/unit/math/rev/mat/functor/algebra_solver_test.cpp
```diff
@@ -312,3 +339,205 @@ TEST(MathMatrix, system_functor_constructor) {
...
+      EXPECT_NEAR(J2(k, l), g[l], tolerance);
+  }
+}
+
+TEST_F(degenerate_eq_test, newton_dbl) {
+  bool is_newton = true;
+
+  // This first initial guess produces the
+  // solution x = {8, 8}
+  Eigen::VectorXd theta = degenerate_test(y_dbl, x_guess_1, is_newton);
```
yizhang-yiz: There are unnecessarily amount complexity for a single test. This three layers of
`degenerate_test` -> `general_algebra_solver` -> `algebra_solver_newton` can simply be
```c++
algebra_solver_newton(degenerate_eq_functor(), x_guess_1, y_dbl, 
                      std::vector<double>(), std::vector<int>());
```
The three-func-call version makes unit test less readable and much error-prone. For example, `degenerate_test` has `x` and `y` in opposite order with `algebra_solver()`, increasing chance of bugs. Please just use `algebra_solver_powell` and `algebra_solver_newton` unless it's absolutely necessary for additional layers.

charlesm93: Agreed. Easy fix.

Usually, the `*_test()` functions do a few things, hence their utility, but `degenerate_test()` only adds an extra layer. I think this scheme made more sense in some earlier design of the unit test.

charlesm93: Fixed.

test/unit/math/rev/mat/functor/algebra_solver_test.cpp
charlesm93: Solved -- see discussion (https://discourse.mc-stan.org/t/newton-solver-in-stan/6684/86).

charlesm93:
> Memory safe issue: need to move KINCreate and KINFree to kinsol_data.

Done.

PR #1341, opened 2019-09-05
Refactor Type Traits

SteveBronder:
Edit: Moving to main repo since something is goofing with Jenkinsbot

## Summary

This refactors the metaprogramming structs and aliases to be a bit more consistent in their definitions and use more of the C++14 standard library methods. A few of the highlights are

1. Instead of using partial specializations that required explicit specializations for const and & types, we use SFINAE and `enable_if_t<std::decay_t<T>>` to turn on and off the specializations for things like `is_vector`

2. Edit: `scalar_type` removes const but leaves refs and pointers which is what the original implementation did

3. `scalar_seq_view` now uses SFINAE to make sure types use the correct specialization.

4. Deletes ad_promotable in fwd and rev. Instead we use the `std::is_convertible` check that the types are implicitly convertible.

5. is_fvar depended on knowing fvar, but the `.hpp` of fvar depends on the meta folder which defined fvar! So I made `fvar.hpp` not depend on the meta programming.

6. new structs `is_std_vector` and `is_eigen` to check if something is a standard vector or derived from EigenBase

7. `is_vector` now checks if the eigen vector has rows or columns at compile time equal to 1.

8. Moved a number of metaprogramming from arr and mat into scale when it could be instantiated there with specializations for `vector` and `eigen` left in `arr` and `mat`

9. Cleaned up some of the docs

10. The new struct `bool_constant`, which is just `std::integral_constant<bool, bool B>`, is inherited by most of the structs that are used to check for types at compile time.

11. Some of the tests checked types for `std::vector<const double>` or `std::vector<double&>` which is illegal under C++11 (types in std::vectors have to be movable and erasable so these were changed to look for pointers.

12. structs containing a type now use `using` instead of `typedef`

~13. constexpr helper variable templates so we can call `is_vector_v<T>` instead of `is_vector<T>::value`~ (THX WINDOWS)
## Tests

The only thing I still need to do is update the tests for `is_std_vector` and `is_eigen`

## Side Effects

Should be none

## Checklist

- [x] Math issue #1308 

- [x] (YAS) the code is written in idiomatic C++ and changes are documented in the doxygen


rok-cesnovar:
Will review this over the weekend.

rok-cesnovar:
Ah, meant to write this to the clang tidy cleanup. This is the meta PR I reviewed + the added decay to is_var_or_arithmetic? If so, I will approve.

SteveBronder:
> This is the meta PR I reviewed that was moved + the added decay to is_var_or_arithmetic? If so, I will approve.

yep!

PR #1347, opened 2019-09-12
[WIP] Use auto and cleanup templated function calls in scalar probability distributions

SteveBronder:
## Summary

This is a cleanup PR that 

1. replaces `return_type_t` in the return part of the scalar probability distributions with `auto` and adds inline where it previously was not

2. Changes the type trait `T_partials_return` in the PR to `T_partials` and uses another alias `T_return` for the catching return statements. There were a few places we were doing `return 0.0` which was fine without auto since it would just cast to the return type. But with auto we need to return the same type from `return`

3. A lot of the functions here would be something like 

```cpp
return normal_lpdf<propto, T_y, T_loc, T_scale>(...)
```

But we can just let the compiler deduce those types with

```cpp
return normal_lpdf<propto>(...)
```

4. Makes `seq_scalar_view` const

5. There's a few places where we can use RVO and return the same object. There's a lot of places where we can't use RVO without paying the cost of calling `build` on `operands_and_partials`. But the places we don't use it tend to just be failure paths and idt we care about the performance of those paths. 

6. There was some weirdness in the return types for some of the distributions. Previously returning 0.0 for a function with`return_type_t<>` as the return type would do a construct of the return type out of the value we returned (for example if return type is `var` we would get back `var(0.0)`) With auto we have to be consistent with the return of the function so this adds `T_return(error_case_value)` to the if statements

## Tests

Refactor so I wouldn't expect any new tests though can add some if requested

## Side Effects

Hopefully none!

## Checklist

- [x] Math issue #1308


SteveBronder:
:pray: Blessed by the distribution tests :pray: 

Once everything is passing I'm going to break these up into smaller PRs that also do some cheap math optimizations. They were a bit easier to see once some tech debt was paid

PR #1351, opened 2019-09-14
Feature/matrix power function

### LINKED ISSUE #47: matrix power operator
_From @bob-carpenter on October 9, 2014 17:13_

Extend the power operator (`^`) to signature `(matrix,int):matrix`, with definitions
- `A^0 = I`
- `A^n = A^(n-1) * A`  for `n > 0`

Throw an exception if `A` is not square or if `n` is not positive.

This should be easy to build.  Extra credit if there's something clever that can be done for the gradients of each `A^n[i,j]` with respect to each `A[i',j']`.  Just storing the gradients is going to be order `K^4` for a `K` by `K` matrix.

_Copied from original issue: stan-dev/stan#1073_

peterwicksstringfield:
## Summary

Implement M^n = M * M * ... * M for square matrices M and non-negative integer n and try to be efficient about it. Closes #47.

Adds functions:
matrix matrix_power(matrix, int)
matrix operator^(matrix, int)
(Aliased.)

For scalars, perform exponentiation by squaring. M^9 = (((M^2)^2)^2)*M.

For var, perform exponentiation naively, recording intermediates. M2 = M^2, M3 = M2 * M, M4 = M3 * M, etc. Use these to propagate gradients in chain. Alternative: perform exponentiation by squaring on the forward pass, generating M2 = M^2, M4 = M^4, M8 = M^8 etc., then fill in the gaps M9 = M8 * M, M10 = M9 * M, etc, inside chain. Same peak memory usage, but the peak would only be achieved inside chain, while the rest of the model was computing it would be O[log[n, 2] K^2] usage.

There's maybe a point where the K^4 memory usage of autodiffing through the exponentiation by squaring beats the n K^3 usage of this scheme?

Note the lack of any forward-mode specializations.

Anyway, here's a starting point.

## Tests

python runTests.py test/unit/math/prim/mat/fun/matrix_power_test.cpp test/unit/math/mix/mat/fun/matrix_power_test.cpp

## Side Effects

Overloads operator^ for matrices.

## Checklist

- [X] Math issue #47


andrjohns:
I'm not sure what the policy is on Eigen's unsupported functions, but there's a [matrix power function](https://eigen.tuxfamily.org/dox/unsupported/classEigen_1_1MatrixPower.html) in their unsupported module.

Alternatively, with larger values of ```n``` it could be more efficient to calculate the power [using the eigenvalues & eigenvectors of the matrix](https://proofwiki.org/wiki/Eigenvalue_of_Matrix_Powers)

SteveBronder:
We have Unsupported imports from

```
./stan/math/prim/mat/fun/autocorrelation.hpp:#include <unsupported/Eigen/FFT>
./stan/math/rev/mat/functor/algebra_solver.hpp:#include <unsupported/Eigen/NonLinearOptimization>
```

So maybe it's fine? @syclik

stan/math/prim/mat/fun/matrix_power.hpp
```diff
@@ -0,0 +1,57 @@
...
+  check_square("matrix_power", "M", M);
+  if (n < 0)
+    invalid_argument("matrix_power", "n", n, "is ", ", but must be >= 0!");
+  if (M.rows() == 0)
+    invalid_argument("matrix_power", "M.rows()", M.rows(), "is ",
+                     ", but must be > 0!");
+  check_finite("matrix_power", "M", M);
+  Eigen::Matrix<T, R, C> result
+      = Eigen::Matrix<T, R, C>::Identity(M.rows(), M.cols());
+  int nn = n;
```
bob-carpenter: I'm not going to review this, just leave a couple comments.

This can be a for loop, which keeps scope cleaner:

```
for (int nn = n;  nn > 0; nn /= 2) {
 ...
}
```

Also, you can start with a single matrix rather than the identity to avoid the identity multiplication (and return identityin special case that power is 0).

stan/math/rev/mat/fun/matrix_power.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+  explicit matrix_product_vari_n0(const Eigen::Matrix<var, R, C>& M)
+      : vari(0.0),
+        Mn_(reinterpret_cast<double*>(
+            ChainableStack::instance_->memalloc_.alloc(sizeof(double) * M.rows()
+                                                       * M.cols()))),
+        adjMnRef_(
+            reinterpret_cast<vari**>(ChainableStack::instance_->memalloc_.alloc(
+                sizeof(vari*) * M.rows() * M.cols()))) {
+    Eigen::Map<matrix_d> Mnd(Mn_, M.rows(), M.cols());
```
bob-carpenter: This should just be a single assignment---you don't need to allocate and then assign.  Also, there's no reason to map the identity, so this can just be:

```
matrix_d Mnd = Matrix<double, R, C>::Identity(M.rows(), M.cols());
```

stan/math/rev/mat/fun/matrix_power.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+template <int R, int C>
+class matrix_product_vari_n0 : public vari {
+ public:
+  double* Mn_;
+  vari** adjMnRef_;
+
+  explicit matrix_product_vari_n0(const Eigen::Matrix<var, R, C>& M)
+      : vari(0.0),
+        Mn_(reinterpret_cast<double*>(
```
bob-carpenter: There's an alloc-array method that'll avoid the casts and sizeof.

stan/math/rev/mat/fun/matrix_power.hpp
```diff
@@ -0,0 +1,182 @@
...
+      for (int j = 0; j < cols_; j++) {
+        double dijdkl = 0.0;
+        Eigen::Matrix<double, R, C> S
+            = Eigen::Matrix<double, R, C>::Zero(rows_, cols_);
+        for (int l = 0; l < cols_; l++) {
+          S(i, l) += Mds[n - 2](j, l);
+        }
+        for (int nn = 1; nn < n - 1; nn++) {
+          for (int k = 0; k < rows_; k++) {
+            for (int l = 0; l < cols_; l++) {
```
bob-carpenter: Matrices are column major, so the loops should go the other way to preserve memory locality.

peterwicksstringfield:
Eigens unsupported matrix power function passes the tests in this PR for doubles. We could use it if we wanted. Doesn't work for fvars. There's an implicit conversion in Eigen's code from fvar to bool that fails to compile. We could maybe work around that.

If we wanted to exploit diagonalization ourselves, Eigen's (supported) HessenbergDecomposition might be useful for non-diagonalizable matrices.

Thank you bob-carpenter for looking over the code.

stan/math/rev/mat/fun/matrix_power.hpp
peterwicksstringfield: Loops swapped.

stan/math/rev/mat/fun/matrix_power.hpp
peterwicksstringfield: So there is. Code looks a lot better now. Thank you.

stan/math/prim/mat/fun/matrix_power.hpp
peterwicksstringfield: Loop cleaned up.

stan/math/rev/mat/fun/matrix_power.hpp
peterwicksstringfield: Fixed.

bob-carpenter:
We could also update Eigen's code.  If they're converting to boolean, they'll be a non-arithmetic workaround.  They've been pretty good about patching or taking PRs.  Does the power function pass the tests for reverse mode?

The test framework that got merged recently should be useful for testing this at various autodiff levels.

peterwicksstringfield:
This works, using Eigen's pow for the primative version:
```
// double
template <int R, int C>
inline Eigen::Matrix<double, R, C> matrix_power(const Eigen::Matrix<double, R, C> &M,
                                                const int n) {
  return M.pow(n);
}
```

So does this, using Eigen's pow to compute M^2, ..., M^(n-1), and then using those to compute gradients for chaining.
```
    // double
    Eigen::MatrixPower<Eigen::Matrix<double, R, C> > MPow(Mds[0]);
    for (int i = 1; i < n - 1; i++) {
      Mds[i] = MPow(i + 1);
      // Mds[i] = Mds[i - 1] * Mds[0];
    }
```

This doesn't:

```
// var
inline Eigen::Matrix<var, R, C> matrix_power(const Eigen::Matrix<var, R, C>& M,
                                             const int n) {
  return M.pow(n);
}
```

```
lib/eigen_3.3.3/unsupported/Eigen/src/MatrixFunctions/MatrixPower.h:465:7: error: could not convert ‘p’ from ‘Eigen::MatrixPower<Eigen::Matrix<stan::math::var, -1, -1> >::RealScalar {aka stan::math::var}’ to ‘bool’
       if (p) computeFracPower(res, p);
       ^~
In file included from ./stan/math/rev/mat/vectorize/apply_scalar_unary.hpp:4:0,
                 from ./stan/math/rev/mat.hpp:11,
                 from test/unit/math/mix/mat/fun/matrix_power_test.cpp:1:
./stan/math/prim/mat/vectorize/apply_scalar_unary.hpp
```

"if (p)" here is a substitute for "if (p != 0)".

The part of Eigen's code that handles fractional powers has more stuff like that, where Eigen tries to convert things into ints/bools in ways our autodiff types don't support. Are we interested in trying to fix this so we can autodiff through Eigen's matrix power? Like this:
```
// T
template <typename T, int R, int C>
inline Eigen::Matrix<T, R, C> matrix_power(const Eigen::Matrix<T, R, C> &M,
                                           const int n) {
  return M.pow(n);
}
```

Note, for the integer powers we are plugging in, Eigen uses exponentiation by squaring (just like the primitive version in this PR), and all that fractional power stuff is dead code. Doesn't stop it from causing compilation errors though.

There are some expect_ad tests in this PR, that's the new framework right?

bob-carpenter:
> "if (p)" here is a substitute for "if (p != 0)".

Ouch.  I think Eigen might accept PRs to fix that kind of thing as they want their templated code to work with more general types, including their own autodiff types (also unsupported).

> Are we interested in trying to fix this so we can autodiff through Eigen's matrix power?

I would think the best thing to do would be to include our own own matrix derivatives within the pow function rather than just autodiffing thorugh Eigen, at least for reverse mode.  For forward mode, their code might be useful---is it that much different in what it does than the simple approach?

peterwicksstringfield:
For integer powers the Eigen code does the same thing we are doing.

```
template<typename MatrixType>
template<typename ResultType>
void MatrixPower<MatrixType>::computeIntPower(ResultType& res, RealScalar p)
{
  using std::abs;
  using std::fmod;
  RealScalar pp = abs(p);

  if (p<0) 
    m_tmp = m_A.inverse();
  else     
    m_tmp = m_A;

  while (true) {
    if (fmod(pp, 2) >= 1)
      res = m_tmp * res;
    pp /= 2;
    if (pp < 1)
      break;
    m_tmp *= m_tmp;
  }
}
```

They also handle non integer powers, M^0.75 = exp(0.75 * log(M)), which is where all the complexity comes from.

I agree that the "if (p)" thing is weird. The documentation for Eigen's pow does say "M has to be a matrix of float, double, long double, complex<float>, complex<double>, or complex<long double>." so presumably the non-generality is known and accepted.

peterwicksstringfield:
I don't want this to get lost so I'm gonna mention #1453 here (and subscribe to it). When that bug is closed the tests should pass on this PR, and I'll just merge develop in again to restart them.

Issue #1354, opened 2019-09-15
Implement rep_vector, rep_matrix and rep_row_vector for matrix_cl


rok-cesnovar:
## Description

With the coming Stan compiler support for `matrix_cl` overloaded functions it makes sense to implement `matrix_cl` overloads for more trivial functions that have been put on hold until now.

I am starting with the rep_vector, rep_row_vector and rep_matrix functions that can significantly reduce data transfers if their results are consumed by other matrix_cl overloaded functions.

## Example

```cpp
matrix_cl<double> a(1,1);
matrix_cl<double> a1(5,1);
matrix_cl<double> a2(1, 6);
...
matrix_cl<double> b = rep_matrix(a, 10, 15);
matrix_cl<double> b1 = rep_matrix(a1, 10);
matrix_cl<double> b2 = rep_matrix(a2, 15);
matrix_cl<double> b_rv = rep_row_vector(a, 15);
matrix_cl<double> b_v = rep_vector(a, 15);
```

The only difference compared to the existing prim implementation is that we cant replicate the following edge case of a vector/row_vector of size 1 with `matrix_cl`. If rep_matrix is called with 1 size parameter (for example `rep_matrix(a, 5)`) the prim implementation deduces from the template parameters whether we want to replicate rows or columns: 
```cpp
  stan::math::vector_d a(1);
  a << 6.0;
  stan::math::matrix_d b = stan::math::rep_matrix(a, 5);
```
matrix_cl does not have templates for size and we are not able to deduce this. If a vector or row_vector is of size 1 we will throw a runtime error when used with `matrix_cl`. In this case the user should use `rep_matrix(a, 5, 1)` or supply a vector of size > 1.

## Expected Output

The mentioned function will have overloads for matrix_cl arguments.

#### Current Version:
v2.20.0

PR #1355, opened 2019-09-15
Feature/issue 1354 Implement matrix_cl overloads for rep_vector, rep_row_vector and rep_matrix

rok-cesnovar:
## Summary

Implement `matrix_cl` overloads for `rep_vector`, `rep_row_vector` and `rep_matrix`. 
This PR additionally adds 2 check functions that were needed to implement this: 
- `check_mat_size_one` checks if the input matrix is 1x1
- `check_mat_not_size_one` checks if the input matrix is not 1x1

This PR also replaces a bug in the zeros kernel. Currently the zeros kernel was working with lower and upper view but in a "strict" fashion, always excluding the diagonal. The zeros() kernel was modified so that it now includes the diagonal. A separate `zeros_strict_tri` was added to replace the behavior of the existing zeros kernel. 
 
## Tests

Each overload has its own test: 
`/opencl/prim/rep_matrix_test.hpp`
`/opencl/prim/rep_vector_test.hpp`
`/opencl/prim/rep_row_vector_test.hpp`

## Side Effects

There is one side effect in terms of replicating the existing `Eigen::MatrixXd` prim implementation for the edge case of replicating a vector or row_vector of size 1 with 1 argument for dimension:
```
  stan::math::vector_d a(1);
  a << 6.0;
  stan::math::matrix_d b = stan::math::rep_matrix(a, 5);
```
The equivalent of this for matrix_cl will throw at runtime. For details see the issue.

## Checklist

- [x] Math issue #1354 

- [x] Copyright holder: Rok Češnovar, Tadej Ciglarič


stan/math/opencl/err/check_vector.hpp
```diff
@@ -31,9 +31,6 @@ inline void check_vector(const char* function, const char* name,
   if (x.cols() == 1) {
     return;
   }
-  if (x.rows() == 1 || x.cols() == 1) {
```
rok-cesnovar: This is just a cleanup of an unnecessary check, otherwise unrelated to the PR.

stan/math/opencl/kernels/constants.hpp
```diff
@@ -0,0 +1,100 @@
...
+      const int j = get_global_id(1);
+      if (i < rows && j < cols) {
+        if ((contains_nonzero(view_A, LOWER) && j <= i)
+            || (contains_nonzero(view_A, UPPER) && j >= i)) {
+          A(i, j) = val;
+        }
+      }
+    }
+    // \cond
+);
```
SteveBronder: Should the name of this be `fill`?

stan/math/opencl/matrix_cl.hpp
```diff
@@ -47,6 +47,8 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
   // Forward declare the methods that work in place on the matrix
   template <matrix_cl_view matrix_view = matrix_cl_view::Entire>
   void zeros();
+  template <matrix_cl_view matrix_view = matrix_cl_view::Entire>
```
SteveBronder: Is `zeros()` supposed to be deleted on line 49?

SteveBronder: Also why the need for a strict zeros function?

stan/math/opencl/kernels/constants.hpp
rok-cesnovar: Yes, that is a much better name :) I am really bad at naming, this is how Eigen names it I think so that will do. Thanks!

stan/math/opencl/err/check_mat_size_one.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * @throw <code>std::domain_error</code> if the <code>matrix_cl</code>
+ *    size is not 1
+ */
+template <typename T>
+inline void check_mat_not_size_one(const char* function, const char* name,
+                                   const matrix_cl<T>& A) {
+  if (A.size() == 1) {
+    invalid_argument(function, name, "should not have exactly one element.",
+                     "");
+  }
```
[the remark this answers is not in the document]
rok-cesnovar: Yes.

stan/math/opencl/kernels/constants.hpp
```diff
@@ -0,0 +1,100 @@
...
+      const int j = get_global_id(1);
+      if (i < rows && j < cols) {
+        if ((view_A == LOWER && j < i) || (view_A == UPPER && j > i)
+            || view_A == ENTIRE) {
+          A(i, j) = val;
+        }
+      }
+    }
+    // \cond
+);
```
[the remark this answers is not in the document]
rok-cesnovar: `fill_tri_strict` would be my choice. `fill` already has the triangular argument. I will explain the change in the next comment.

stan/math/opencl/matrix_cl.hpp
rok-cesnovar: The thing is the current zeros kernel was working with lower and upper view already but in a "strict" fashion, always excluding the diagonal (Entire worked normally). 

In all places we were using zeros with Lower or Upper this was the expected behavior and the zeros kernel tests were also expecting the strict triangular behavior. So this was the intended behavior, I just found it odd that this was done this way without explicitly stating it. This one was on me.

I contemplated adding StrictlyUpper and StrictlyLower views, but that would mean we need to add the "strictly" behavior to all kernels that use the view. I dont think that is worth it.

So zeros<Lower>() on a matrix of ones results in

0 1 1
0 0 1
0 0 0 

and zeros_strict_tri<Lower() results in

1 1 1
0 1 1
0 0 1 

Hopefully I am making sense, I am half-way through my morning coffee :)

stan/math/opencl/kernels/constants.hpp
t4c1: I think this won't work if `view_A==DIAGONAL`. You need to add `|| i==j` to the condition of the second if. That case can also be added to tests.

rok-cesnovar: Good catch. I forgot the diagonal case.

stan/math/opencl/matrix_cl.hpp
SteveBronder: Word!

SteveBronder:
fyi will look this over tmrw

stan/math/opencl/matrix_cl.hpp
SteveBronder: You deleted the code for this didnt you? 

```
template <matrix_cl_view matrix_view = matrix_cl_view::Entire> void zeros
```

How bad would it be to add the checks for strictly upper and strictly lower? And just have one zeros function? It feels like that would he nicer at the C++ level relative to having two methods

rok-cesnovar: No, the regular zeros remains, I did not remove it, there is also tests for it. 

To have a single zeros method we need to add `StrictlyUpper` and `StrictlyLower` to the view enum. That part is easy, extending the kernel would also be trivial.

Then its a question if we want to refactor all the other kernels to also support that behavior everywhere. Supporting those 2 enums only in zeros would be weired than this imo. I think a refactor to support `strictlyuppper`/`strictlylower` should be a separate issue/PR because that touches almost every kernel..

rok-cesnovar: The only other thing we could do is add a matrix_cl_strict_view enum and add another argument to the kernel. Would you prefer that? I am fine with that.

SteveBronder: Okay cool cool, if it's too much of a bother then I'm fine with a different method

rok-cesnovar: Adding another argument to denote if the view is strict or not is fine with me. Would you prefer that?

rok-cesnovar:
Thanks @SteveBronder. I will try to fix this up before the pair programming session with Sean today. Otherwise tomorrow the latest.

stan/math/opencl/prim/rep_matrix.hpp
```diff
@@ -0,0 +1,89 @@
...
+inline matrix_cl<double> rep_matrix(const matrix_cl<T>& x, int m) {
+  check_nonnegative("rep_matrix (OpenCL)", "rows/columns", m);
+  const int N = x.rows() == 1 ? m : x.rows();
+  const int M = x.cols() == 1 ? m : x.cols();
+  matrix_cl<double> A(N, M);
+  if (A.size() == 0) {
+    return A;
+  }
+  check_mat_not_size_one("rep_matrix (OpenCL)", "x", x);
+  check_vector("rep_matrix (OpenCL)", "x", x);
```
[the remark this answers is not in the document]
SteveBronder: Word that makes sense

bob-carpenter:
> Cool beans! ping me and I'll approve

Can we make "cool beans" the Stan tagline?  I had thought @bbbales2 coined the expression, but apparently it's been around since the 1960s and was popularized on the American sitcom <i>Full House</i> in the 1980s.

betanalpha:
You got it, dude.

I’m just here for the Full House puns.

betanalpha:
Have mercy.  (I think — I’m running out of memories of the show  :-p)

stan/math/opencl/kernels/fill.hpp
```diff
@@ -0,0 +1,99 @@
...
+     * @param val value to replicate in the matrix
+     * @param rows Number of rows for matrix A
+     * @param cols Number of columns for matrix A
+     * @param view_A triangular part of matrix A to use
+     *
+     * @note Code is a <code>const char*</code> held in
+     * <code>fill_kernel_code.</code>
+     * This kernel uses the helper macros available in helpers.cl.
+     */
+    __kernel void fill(__global double* A, double val, unsigned int rows,
```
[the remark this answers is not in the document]
rok-cesnovar: Moved to __constant.

stan/math/opencl/prim/rep_matrix.hpp
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @throw <code>domain_error</code> if the
+ * requested dimensions are negative
+ *
+ */
+template <typename T, typename = enable_if_all_arithmetic<T>>
+inline matrix_cl<double> rep_matrix(const matrix_cl<T>& x, int m) {
+  check_nonnegative("rep_matrix (OpenCL)", "rows/columns", m);
+  const int N = x.rows() == 1 ? m : x.rows();
+  const int M = x.cols() == 1 ? m : x.cols();
```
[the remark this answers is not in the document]
rok-cesnovar: In case of NxM the `check_vector` will throw.

stan/math/opencl/kernels/fill.hpp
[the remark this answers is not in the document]
rok-cesnovar: Scratch that, it didnt allow me to do it.

rok-cesnovar:
@SteveBronder addressed all your comments. Ready for re-review.

SteveBronder:
Looks like there was a goof here

```
[ RUN      ] MathMatrixCL.rep_rv_exception_pass
test/unit/math/opencl/prim/rep_row_vector_test.cpp:22: Failure
Expected: stan::math::rep_row_vector(a, 0) doesn't throw an exception.
  Actual: it throws.
test/unit/math/opencl/prim/rep_row_vector_test.cpp:23: Failure
Expected: stan::math::rep_row_vector(c, 0) doesn't throw an exception.
  Actual: it throws.
[  FAILED  ] MathMatrixCL.rep_rv_exception_pass (0 ms)
```

rok-cesnovar:
Yes, I addressed your comments to move the checks above the returns fro zero size which changes the exception behavior. Its fixed now.

Issue #1346, opened 2019-09-12
Asynchronous OpenCL calls have potential for undefined behaviour


t4c1:
## Description
There are some innocent-looking code patterns that can actually result in reading memory from destroyed objects. 

<s>1) Using any temporary `matrix_cl` variable in a function that calls a kernel. Synce kernel call is asynchrounous that temporary could be destroyed before the kernel is executed. This is only problem with arguments of type `matrix_cl` - `setKernelArguments()` makes a copy of scalars, but not of `cl::Buffer`s.</s>

2) Constructing a `matrix_cl` from any temporary object. Since we are using asynchronous copying that that temporary could be destroyed before copying happens.

## Example

<s>1)
```
N=100;
matrix_cl<double> a(N,N), b(N,N), c(N,N);
matrix_cl<double> d = a + b + c;
```
</s>

2)
```
MatrixXd a = ...
MatrixXd b = ...

matrix_cl<double> c(a + b);
```

## Expected Output
Well-defined behaviour.

#### Current Version:
v2.20.0

t4c1:
I just found out that documentation for `clReleaseMemoryObject` says that a memory object is not deleted until all comands using it have completed, even if its reference count is 0. So 1 is not actually an issue.

SteveBronder:
Ty! For (2) could we just make a copy of the object when it's a temp?

t4c1:
I guess we could allocate the object on the heap and construct it by moving from the temporary. Than bind a callback to the event that will delete the object. That requires two versions of every existing `matrix_cl` constructor, but I think it is a good option.

SteveBronder:
I think we could also use more generic templates and then for temporaries have an `if(is_rvalue_reference<T>::value) // do copy of temp`

t4c1:
You are right. Two versions of constructors are not needed. However the copies still need to be dynamically allocated.

SteveBronder:
Which devices and platforms did the tests fail on? Looking at your clinfo it looks like your drivers support out of order while the Intel(R) HD Graphics 4000 device does not. If I had to guess I bet we are only looking at if the driver supports it and not the device. I ran the test you posted on my laptop (which does not support async) and can confirm I'm getting it here as well. I'll work on a fix tmrw

t4c1:
I tested on NVidia GPUs on both computers.

rok-cesnovar:
Fixed via #1356

PR #1356, opened 2019-09-15
Bugfix for making matrix_cls from temporaries

SteveBronder:
## Summary

As described in #1346 when we would create a matrix_cl from a temporary eigen matrix `matrix_cl<double> A(x * y);` async would could cause the transfer to happen after the memory for the temporary variable was deleted on the host leading to UB.

This is bug was only for eigen expressions so the matrix_cl constructor for eigen matrices now does a check to see if the input is an eigen matrix or an eigen expression. If it's an expression then it is evaluated and a local copy is made to transfer to OpenCL. Eigen matrices will not have to cost of making a copy via `.eval()`. I was also worried this could arise from us not catching drivers and devices that do not support async OpenCL and so I added a check in the OpenCL context that makes sure all of the transfer functions are set to `blocking` when the device/platform does not support OpenCL called `in_order`. As async in OpenCL terms is called out-of-order a value of `CL_TRUE` for `in_order` means no async and a value of `false` means async is used

## Tests

a new test was added to the OpenCL unit tests showing the failing behavior and can be run with 

```bash
./runTests.py ./test/unit/math/opencl/copy_test
```

`is_eigen_matrix` is a new function to check whether a type has a `::Matrix()` constructor. The test can be run with

```bash
./runTests.py ./test/unit/math/prim/mat/meta/is_eigen_test
```

## Side Effects

None

## Checklist

- [x] Math issue #1346


t4c1:
Also, while I demo-ed the issue only for an expression of `Eigen::Matrix` I think it can happen when copyingdata of any type into a `matrix_cl`

SteveBronder:
> As for bugfix I think you misunderstood the nature of the problem. While I did not run the test on your code, I don't think it fixes the bug.

I'd suggest running the code to verify that. And if the test case we have does not show the nature of the bug then we still need a test case that shows the effect

> You put evaluated temporary in a stack variable, which can still go out of scope before actual copy happens.

If that's your concern my first suggestion would be to check the code for one or two open source icds to see how they handle the inputs. I would bet they make a copy to store with the transfer. Cuz otherwise idk how async could ever work

t4c1:
Now I ran the test for this bug on your code. It still fails. 

> check the code for one or two open source icds to see how they handle the inputs.

I think they must make a copy (to pinned memory needed for the transfer to GPU). However they don't need to make that copy immediately if working async. They assume the pointer they got is valid until the copy event is waited on.

SteveBronder:
If you have a fix that would be great! This fixed the error from the test case when I ran it on my machine so I'm having a hard time coming up with the fix. Could we just set blocking to true?

rok-cesnovar:
I would say we go with blocking for now for the enqueue write buffer. Async is still useful for running the kernels after that. 

This really might be something different runtimes handle differently and some obviously rely on the pointer argument to be valid for the entire duration of the application, without making their copy or what not. Seems weird and yuck but given that we are experiencing the "works on my machine" issues I would bet on that. Thanks to both for debugging and trying to fix this weird issue.

t4c1:
> Could we just set blocking to true* when the input to the constructor is an rvalue?

Let's do this for now. I'd like to use async for rvalues too, but right now I have some problems with templating. I think it will get easier once templating stuff @SteveBronder is working on gets merged. So I intend to reenable async later.

stan/math/opencl/copy.hpp
```diff
@@ -52,8 +52,9 @@ inline matrix_cl<T> to_matrix_cl(const Eigen::Matrix<T, R, C>& src) {
      */
     cl::Event copy_event;
     const cl::CommandQueue& queue = opencl_context.queue();
-    queue.enqueueWriteBuffer(dst.buffer(), CL_FALSE, 0, sizeof(T) * dst.size(),
-                             src.data(), &dst.write_events(), &copy_event);
```
rok-cesnovar: We need to either replace `CL_FALSE` with `CL_TRUE` or setup a separate function in `opencl_context.write_buffer_blocking()`. I dont have a preference.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -296,6 +297,32 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
     }
   }
 
+  template <typename Mat,
+            std::enable_if_t<is_eigen_matrix<std::decay_t<Mat>>::value>...>
+  void setup_buffer(Mat&& A) {
```
rok-cesnovar: We probably dont need this one then.

stan/math/prim/mat/meta/is_eigen.hpp
```diff
@@ -36,5 +36,28 @@ struct is_eigen<
...
+};
+
+template <class T>
+struct is_eigen_matrix_impl<T, void_t<typename T::SparseMatrix()>>
+    : std::true_type {};
+
+}  // namespace internal
+
+template <typename T>
+struct is_eigen_matrix : internal::is_eigen_matrix_impl<std::decay_t<T>> {};
```
rok-cesnovar: We probably dont need this type trait right now, but I would definitely leave it. Its also got some sparse stuff so it will probably be used soon-ish.

test/unit/math/opencl/copy_test.cpp
```diff
@@ -8,6 +8,35 @@
...
+  b(i) = no_opt;
+
+  MatrixXd correct = a + b;
+  MatrixXd result = from_matrix_cl(c_cl);
+  for (int i = 0; i < N * N; i++) {
+    EXPECT_EQ(result(i), correct(i));
+  }
+  no_opt = w.array().sum();
+}
+
```
rok-cesnovar: Remove the test.

rok-cesnovar: I am an idiot, offcourse this should stay in... Its the entire reason we got to this PR.

rok-cesnovar:
@SteveBronder do you mind if I finish what is left to fix here?

SteveBronder:
yes def! Sorry got a bit hectic lately

stan/math/opencl/opencl_context.hpp
```diff
@@ -101,9 +101,12 @@ class opencl_context_base {
       if (device_properties & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) {
         command_queue_ = cl::CommandQueue(
             context_, device_, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, nullptr);
+        block_reads_ = CL_FALSE;
       } else {
         command_queue_ = cl::CommandQueue(context_, device_, 0, nullptr);
+        block_reads_ = CL_TRUE;
       }
+      block_writes_ = CL_TRUE;
```
SteveBronder: Why is this always true?

rok-cesnovar: I just wanted to see if the tests would pass, if always set to true. Then the only missing thing would be to only set it true for rvalues.

stan/math/prim/mat/meta/is_eigen.hpp
rok-cesnovar: @SteveBronder  The tests for this fail with clang++ but pass with g++:

```
test/unit/math/prim/mat/meta/is_eigen_test.cpp:41: Failure
Value of: (is_eigen_matrix<Eigen::Matrix<double, -1, -1>>::value)
  Actual: false
Expected: true
test/unit/math/prim/mat/meta/is_eigen_test.cpp:42: Failure
Value of: (is_eigen_matrix<Eigen::SparseMatrix<double>>::value)
  Actual: false
Expected: true
test/unit/math/prim/mat/meta/is_eigen_test.cpp:45: Failure
Value of: (is_eigen_matrix<const Eigen::Matrix<double, -1, -1>>::value)
  Actual: false
Expected: true
test/unit/math/prim/mat/meta/is_eigen_test.cpp:46: Failure
Value of: (is_eigen_matrix<Eigen::SparseMatrix<double>&>::value)
  Actual: false
Expected: true
```
Any idea why? Should we split this to a separate PR? Your call.

SteveBronder: Aaaah yes I saw that problem on clang as well. Let me look tomorrow I bet there's an easy fix

rok-cesnovar:
The `is_eigen_matrix` issue is fixed now right? If so, I will go over this again later today.

rok-cesnovar:
Oh and we need to switch back to blocking writes only for rvalues.

SteveBronder:
Yep! Just swapped that out, This fix assumes the rvalue is just an eigen expression. If someone made an Eigen matrix within the creation of the matrix_cl. Though I think you can only do that with something like `Matrix::Random(n,m)`

rok-cesnovar:
I am not sure I understand completely. You also fixed the temporaries bug already? But I dont see a blocking write anywhere?

SteveBronder:
Waagh no I did not, but the tests pass for me? @t4c1 can you check whether this passes on the machine this was failing on before?

If it doesn't then I have a fix to catch the rvalues, but if it's not necessary then it'd be nice to keep the constructor simpler

t4c1:
This still does not pass? I see no blocking writes, so I am a bit confused. Which change could fix this bug?

stan/math/opencl/copy.hpp
```diff
@@ -52,8 +52,9 @@ inline matrix_cl<T> to_matrix_cl(const Eigen::Matrix<T, R, C>& src) {
...
+    queue.enqueueWriteBuffer(dst.buffer(), opencl_context.in_order(), 0,
```
t4c1: While I did not test this method in that bug demo It needs same kind of changes as matrix_cl constructors. Addeing a simiar test for this could also be benefitial.

stan/math/opencl/copy.hpp
```diff
@@ -160,7 +162,7 @@ inline matrix_cl<T> packed_copy(const std::vector<T>& src, int rows) {
     matrix_cl<T> packed(packed_size, 1);
     cl::Event packed_event;
     const cl::CommandQueue queue = opencl_context.queue();
-    queue.enqueueWriteBuffer(packed.buffer(), CL_FALSE, 0,
+    queue.enqueueWriteBuffer(packed.buffer(), opencl_context.in_order(), 0,
```
t4c1: I don't expect someone constructing a std::vector temporary and passing it directely into a function. But maybe we should also change this one.

stan/math/opencl/copy.hpp
```diff
@@ -251,8 +253,8 @@ inline matrix_cl<T> to_matrix_cl(const T& src) {
   try {
     cl::Event copy_event;
     const cl::CommandQueue queue = opencl_context.queue();
-    queue.enqueueWriteBuffer(dst.buffer(), CL_FALSE, 0, sizeof(T), &src,
-                             &dst.write_events(), &copy_event);
+    queue.enqueueWriteBuffer(dst.buffer(), opencl_context.in_order(), 0,
```
t4c1: This one should also be changed.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -261,7 +262,7 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
        */
       cl::Event write_event;
       queue.enqueueWriteBuffer(
-          buffer_cl_, CL_FALSE, sizeof(double) * offset_size,
+          buffer_cl_, opencl_context.in_order(), sizeof(double) * offset_size,
```
t4c1: This constructor should also be changed.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -298,6 +299,28 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
...
+    this->add_write_event(transfer_event);
+  }
+
+  template <typename Mat,
+            std::enable_if_t<!is_eigen_matrix<Mat>::value
+                             || std::is_rvalue_reference<Mat>::value>...>
+  inline void write_buffer_(Mat&& A) {
+    cl::Event transfer_event;
+    cl::CommandQueue& queue = opencl_context.queue();
+    queue.enqueueWriteBuffer(this->buffer_cl_, opencl_context.in_order(), 0,
```
t4c1:
```suggestion
    queue.enqueueWriteBuffer(this->buffer_cl_, CL_TRUE, 0,
```
This change fixes the test I provided. But there are still other constructors and functions in `copy.hpp` that are affected by this bug.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -347,7 +365,7 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
     try {
       buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(T));
       cl::Event transfer_event;
-      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, 0, sizeof(T), &A, NULL,
+      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, 0, sizeof(T), &A, nullptr,
```
t4c1: This constructor should also be changed.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -376,8 +394,9 @@ class matrix_cl<T, enable_if_arithmetic<T>> {
     try {
       buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(T) * A.size());
       cl::Event transfer_event;
-      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, 0, sizeof(T) * A.size(),
-                               A.data(), nullptr, &transfer_event);
+      queue.enqueueWriteBuffer(buffer_cl_, opencl_context.in_order(), 0,
+                               sizeof(T) * A.size(), A.data(), nullptr,
+                               &transfer_event);
```
t4c1: This constructor should also be changed.

stan/math/opencl/opencl_context.hpp
```diff
@@ -144,7 +146,7 @@ class opencl_context_base {
   std::string device_name_;          // The name of OpenCL device
   size_t max_thread_block_size_;  // The maximum size of a block of workers on
                                   // the device
-
+  bool block_transfer;            // Whether to use out of order execution.
```
t4c1: I don't see how this is connected to any blocks. I think a better name for this variable would be `in_order_` or `use_asnyc_transfers_` or something like that.

stan/math/opencl/copy.hpp
rok-cesnovar: I tried this with the following test: 

```
  MatrixXd a = MatrixXd::Random(N, N);
  MatrixXd b = MatrixXd::Random(N, N);
  matrix_cl<double> c_cl = to_matrix_cl(a + b);  // the problem
```
and that produces the following error: 
```
test/unit/math/opencl/copy_test.cpp:50:28: error: no matching function for call to 'to_matrix_cl'
  matrix_cl<double> c_cl = to_matrix_cl(a + b);  // the problem
                           ^~~~~~~~~~~~
./stan/math/opencl/copy.hpp:248:21: note: candidate template ignored: requirement
      'is_arithmetic<CwiseBinaryOp<scalar_sum_op<double, double>, const Matrix<double, -1, -1, 0, -1,
      -1>, const Matrix<double, -1, -1, 0, -1, -1> > >::value' was not satisfied [with T =
      Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, const Eigen::Matrix<double,
      -1, -1, 0, -1, -1>, const Eigen::Matrix<double, -1, -1, 0, -1, -1> >]
inline matrix_cl<T> to_matrix_cl(const T& src) {
                    ^
./stan/math/opencl/copy.hpp:40:21: note: candidate template ignored: could not match 'Matrix' against
      'CwiseBinaryOp'
inline matrix_cl<T> to_matrix_cl(const Eigen::Matrix<T, R, C>& src) {
                    ^
1 error generated.
```
So the bug cant happen in this case. We will need to support this in the future but that is not part of this bugfix PR.

t4c1: I think you can probably work around this error by doing : ` matrix_cl<double> c_cl = to_matrix_cl((a + b).eval());`.

rok-cesnovar: Yep, that fails.

stan/math/opencl/copy.hpp
```diff
@@ -147,23 +168,24 @@ inline std::vector<T> packed_copy(const matrix_cl<T>& src) {
...
   }
   try {
-    matrix_cl<T> packed(packed_size, 1);
+    matrix_cl<Vec_scalar> packed(packed_size, 1);
     cl::Event packed_event;
     const cl::CommandQueue queue = opencl_context.queue();
-    queue.enqueueWriteBuffer(packed.buffer(), CL_FALSE, 0,
-                             sizeof(T) * packed_size, src.data(), nullptr,
-                             &packed_event);
+    queue.enqueueWriteBuffer(packed.buffer(), opencl_context.in_order(), 0,
```
t4c1:
```suggestion
    queue.enqueueWriteBuffer(packed.buffer(), std::is_rvalue_reference<Vec&&> || opencl_context.in_order(), 0,
```
Everything is prepared, but the actual bug is not fixed in this function.

stan/math/opencl/copy.hpp
```diff
@@ -242,17 +264,18 @@ inline T from_matrix_cl_error_code(const matrix_cl<T>& src) {
...
+  matrix_cl<std::decay_t<T>> dst(1, 1);
   check_size_match("to_matrix_cl ((OpenCL) -> (OpenCL))", "src.rows()",
                    dst.rows(), "dst.rows()", 1);
   check_size_match("to_matrix_cl ((OpenCL) -> (OpenCL))", "src.cols()",
                    dst.cols(), "dst.cols()", 1);
   try {
     cl::Event copy_event;
     const cl::CommandQueue queue = opencl_context.queue();
-    queue.enqueueWriteBuffer(dst.buffer(), CL_FALSE, 0, sizeof(T), &src,
+    queue.enqueueWriteBuffer(dst.buffer(), std::is_rvalue_reference<T&&>::value,
```
t4c1:
```suggestion
    queue.enqueueWriteBuffer(dst.buffer(), opencl_context.in_order() || std::is_rvalue_reference<T&&>::value,
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -260,9 +259,9 @@ class matrix_cl<T, require_arithmetic_t<T>> {
        * is finished transfering
        */
       cl::Event write_event;
-      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, sizeof(T) * offset_size,
-                               sizeof(T) * rows_, A[i].data(), nullptr,
-                               &write_event);
+      queue.enqueueWriteBuffer(buffer_cl_, opencl_context.in_order(),
```
t4c1:
```suggestion
      queue.enqueueWriteBuffer(buffer_cl_, opencl_context.in_order() || std::is_rvalue_reference<Vec&&>,
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -260,9 +259,9 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+                               sizeof(double) * offset_size, sizeof(T) * rows_,
```
t4c1:
```suggestion
                               sizeof(T) * offset_size, sizeof(T) * rows_,
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -318,12 +317,13 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
     cl::Context& ctx = opencl_context.context();
-    cl::CommandQueue& queue = opencl_context.queue();
     try {
       buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(T) * A.size());
       cl::Event transfer_event;
-      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, 0, sizeof(T) * A.size(),
-                               A.data(), nullptr, &transfer_event);
+      cl::CommandQueue& queue = opencl_context.queue();
+      queue.enqueueWriteBuffer(
+          this->buffer_cl_, std::is_rvalue_reference<Mat&&>::value, 0,
```
t4c1:
```suggestion
          this->buffer_cl_, opencl_context.in_order() || std::is_rvalue_reference<Mat&&>::value, 0,
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -346,14 +346,72 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+   */
+  explicit matrix_cl(const T&& A,
+                     matrix_cl_view partial_view = matrix_cl_view::Diagonal)
+      : rows_(1), cols_(1), view_(partial_view) {
+    cl::Context& ctx = opencl_context.context();
+    cl::CommandQueue& queue = opencl_context.queue();
+    try {
+      buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(std::decay_t<T>));
+      cl::Event transfer_event;
+      queue.enqueueWriteBuffer(buffer_cl_, std::is_rvalue_reference<T&&>::value,
```
t4c1:
```suggestion
      queue.enqueueWriteBuffer(buffer_cl_, opencl_context.in_order() || std::is_rvalue_reference<T&&>::value,
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -346,14 +346,72 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+  /**
+   * Constructor for the matrix_cl that
+   * creates a copy of a rvalue reference
+   * of a scalar on the OpenCL device.
+   * Regardless of `partial_view`, whole matrix is stored.
+   *
+   * @param A the scalar
+   * @param partial_view which part of the matrix is used
+   */
+  explicit matrix_cl(const T&& A,
```
t4c1:
```suggestion
template<typename Scal, typename = require_same<T,std::remove_reference_t<Scal>>>
  explicit matrix_cl(const Scal&& A,
```
I don't think we need two constructors for scalars. Just this one should be enough if you make a small change, so you can delete the one above.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -346,14 +346,72 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+    if (size() == 0) {
+      return;
+    }
+    cl::Context& ctx = opencl_context.context();
+    cl::CommandQueue& queue = opencl_context.queue();
+    try {
+      buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(T) * A.size());
+      cl::Event transfer_event;
+      queue.enqueueWriteBuffer(
+          buffer_cl_, std::is_rvalue_reference<Vec&&>::value, 0,
```
t4c1:
```suggestion
          buffer_cl_,  opencl_context.in_order() || std::is_rvalue_reference<Vec&&>::value, 0,
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -346,14 +346,72 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+   * @param R Number of rows the matrix should have.
+   * @param C Number of columns the matrix should have.
+   * @param partial_view which part of the matrix is used
+   * @throw <code>std::system_error</code> if the
+   * matrices do not have matching dimensions
+   */
+  template <typename Vec, require_std_vector_t<Vec>...,
+            require_same_vt<Vec, T>...>
+  explicit matrix_cl(Vec&& A,
+                     matrix_cl_view partial_view = matrix_cl_view::Entire)
```
t4c1: No need for separate implementation if we just need to set default values for number of rows and cols
```suggestion
                     matrix_cl_view partial_view = matrix_cl_view::Entire) : matrix_cl(A, A.size(), 1) {}
```

stan/math/opencl/matrix_cl.hpp
```diff
@@ -377,8 +435,9 @@ class matrix_cl<T, require_arithmetic_t<T>> {
     try {
       buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(T) * A.size());
       cl::Event transfer_event;
-      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, 0, sizeof(T) * A.size(),
-                               A.data(), nullptr, &transfer_event);
+      queue.enqueueWriteBuffer(
+          buffer_cl_, std::is_rvalue_reference<Vec&&>::value, 0,
```
t4c1:
```suggestion
          buffer_cl_,  opencl_context.in_order() || std::is_rvalue_reference<Vec&&>::value, 0,
```

test/unit/math/opencl/copy_test.cpp
```diff
@@ -8,6 +8,151 @@
...
+  using stan::math::from_matrix_cl;
+  using stan::math::matrix_cl;
+  double a = 1.0;
+  double b = 2.0;
+  double c = 3.0;
+  double d = 4.0;
+  matrix_cl<double> c_cl
+      = stan::math::to_matrix_cl(a + b + c + d);  // the problem
+  // attempt to scramble the memory that was used by the temporary
+  double f = a - b - c - d;
```
t4c1: Did this test reproduce the bug for you? I think compiler is likely to only use registers for scalars, so we have little or no chance to actually overwrite them.

test/unit/math/opencl/copy_test.cpp
```diff
@@ -8,6 +8,151 @@
...
+  no_opt = f + b + c + d;
+}
+
+TEST(MathMatrixGPU, std_vector_copy_destroyed_constructor) {
+  using Eigen::MatrixXd;
+  using stan::math::from_matrix_cl;
+  using stan::math::matrix_cl;
+
+  matrix_cl<double> c_cl(std::vector<double>({1, 2, 3, 4, 5, 6}), 2,
+                         3);  // the problem
```
t4c1: I think for the sake of clarity you should not shape this vector as 2x3 matrix on GPU. Since you made a constructor with default shape you can probably use it.

Does this reproduce the bug? I think it is unlikely to if you do not try to overwrite the values of the temporary object.

stan/math/opencl/copy.hpp
```diff
@@ -25,36 +25,56 @@
...
+inline matrix_cl<Vec_scalar> to_matrix_cl(Vec&& src) {
+  matrix_cl<Vec_scalar> dst(src.size(), 1);
+  if (src.size() == 0) {
+    return dst;
+  }
+  try {
+    cl::Event transfer_event;
+    cl::CommandQueue& queue = opencl_context.queue();
+    queue.enqueueWriteBuffer(
+        dst.buffer(), std::is_rvalue_reference<Vec_scalar&&>::value, 0,
```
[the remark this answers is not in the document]
rok-cesnovar: I dont understand this one. The first argument is the cl buffer not the vector?

stan/math/opencl/matrix_cl.hpp
rok-cesnovar: Yep, didnt think of this one. Thanks!

stan/math/opencl/matrix_cl.hpp
rok-cesnovar: Thanks.

test/unit/math/opencl/copy_test.cpp
rok-cesnovar: Yes, this reproduces the bug on my notebook. I didnt try on other machines because I didnt have access to any other today. It behaves like its undefined, 1E-316 and such.

test/unit/math/opencl/copy_test.cpp
rok-cesnovar: Will use the default constructor, Yes, it reproduces the bug, same as above.

stan/math/opencl/copy.hpp
```diff
@@ -25,36 +25,56 @@
...
-     */
-    cl::Event copy_event;
-    const cl::CommandQueue& queue = opencl_context.queue();
-    queue.enqueueWriteBuffer(dst.buffer(), CL_FALSE, 0, sizeof(T) * dst.size(),
-                             src.data(), &dst.write_events(), &copy_event);
-    dst.add_write_event(copy_event);
+    cl::Event transfer_event;
+    cl::CommandQueue& queue = opencl_context.queue();
+    queue.enqueueWriteBuffer(
+        dst.buffer(), std::is_rvalue_reference<Mat_scalar&&>::value, 0,
```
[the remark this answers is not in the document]
rok-cesnovar: Added .eval and added the test.

stan/math/opencl/copy.hpp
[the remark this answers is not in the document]
t4c1: My mistake. Ignore this.

stan/math/opencl/copy.hpp
```diff
@@ -214,7 +221,7 @@ inline matrix_cl<T> copy_cl(const matrix_cl<T>& src) {
   }
   try {
     /**
-     * Copies the contents of the src buffer to the dst buffer
+     * Copies the contents of the src buffer to the dst buffer<
```
t4c1: Revert this typo.

test/unit/math/opencl/copy_test.cpp
```diff
@@ -8,6 +8,35 @@
...
+TEST(MathMatrixGPU, copy_destroyed_constructor) {
+  using Eigen::Dynamic;
+  using Eigen::Matrix;
+  using Eigen::MatrixXd;
+  using stan::math::from_matrix_cl;
+  using stan::math::matrix_cl;
+  int N = 100;
+  MatrixXd a = MatrixXd::Random(N, N);
+  MatrixXd b = MatrixXd::Random(N, N);
+  matrix_cl<double> c_cl(a + b);  // the problem, this case with .eval()
```
t4c1: This has no eval anymore - fix the comment

stan/math/opencl/matrix_cl.hpp
```diff
@@ -391,15 +391,15 @@ class matrix_cl<T, require_arithmetic_t<T>> {
     if (size() == 0) {
       return;
     }
+    std::cout << "T" << std::endl;
```
t4c1: whoopsie?

rok-cesnovar: haha, debug :D

stan/math/opencl/matrix_cl.hpp
```diff
@@ -332,28 +332,47 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
     cl::CommandQueue& queue = opencl_context.queue();
     try {
-      buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(T));
+      buffer_cl_ = cl::Buffer(ctx, CL_MEM_READ_WRITE, sizeof(std::decay_t<T>));
       cl::Event transfer_event;
-      queue.enqueueWriteBuffer(buffer_cl_, CL_FALSE, 0, sizeof(T), &A, nullptr,
-                               &transfer_event);
+      queue.enqueueWriteBuffer(
+          buffer_cl_,
+          opencl_context.in_order() || std::is_rvalue_reference<T&&>::value, 0,
```
SteveBronder: Do we need the T&& here? If T is an rvalue it should have && already

t4c1: Yes. T is either value or lvalue reference, but never rvalue reference.

rok-cesnovar: We either use this or `std::is_rvalue_reference<decltype(A)>`, which I had before but then Tadej suggest T&& which I agree is nicer.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -332,28 +332,47 @@ class matrix_cl<T, require_arithmetic_t<T>> {
 
   /**
    * Constructor for the matrix_cl that
-   * creates a copy of a scalar on the OpenCL device.
+   * creates a copy of a rvalue reference
+   * of a scalar on the OpenCL device.
```
t4c1: This change in the documentation is wrong. Please revert. `Scal&&` here is not necesarilly a rvalue reference. It is so-called forwarding reference, which can be deduced into either rvalue reference or lvalue reference.

stan/math/opencl/matrix_cl.hpp
t4c1: This was not completely fixed. You skipped `opencl_context.in_order() || ` part.

test/unit/math/opencl/copy_test.cpp
```diff
@@ -8,6 +8,209 @@
 #include <algorithm>
 #include <vector>
 
+TEST(MathMatrixGPU, copy_destroyed_constructor) {
```
t4c1: [Optional] Now that the bug is fixed we are not copying destroyed objects anymore. I would suggest changing the names of these tests from "copy_destroyed..." to "copy_rvalue..."

stan/math/opencl/matrix_cl.hpp
rok-cesnovar: Yep, this is leftover from the way it was setup before with 2 separate constructors. Will fix in a minute. Tnx

rok-cesnovar: I just removed the "of a rvalue reference" part of the text. Is that ok with you?

test/unit/math/opencl/copy_test.cpp
rok-cesnovar: Good call. Went and changed all `_destroyed_` to `_rvalue_` in copy_test.cpp. If that is fine with you, then these comments were all addressed.

PR #1376, opened 2019-10-01
integrate Intel TBB

wds15:
## Summary

This PR integrate the Intel TBB with stan-math. Specifically this PR ensures the TBB's threadpool initialization and it switches `map_rect` C++11 threads implementation over to use the `tbb::parallel_for` (which has scheduling and a more efficient threadpool).

In order to keep the Intel TBB optional in the next release, the TBB code is only used whenever `STAN_THREADS` is set. This makes the TBB only mandatory whenever threads are being used. The Intel TBB will become mandatory in the near future regardless of the `STAN_THREADS` use (which has licensing implications for GPL-2 projects using Stan-math).

The initialisation of the TBB happens in two places:

- General initialisation of the maximal concurrency level in the TBB threadpool. This is handled by the new `init_threadpool_tbb` function defined in `stan/math/prim/core/init_threadpool_tbb.hpp`. The number of threads used is defined in the `STAN_NUM_THREADS` environment variable.

- For all TBB worker threads, the AD tape is initialised through an observer which is registered with the Intel TBB. Every time a worker thread enters the threadpool, the observer is called and ensures proper initialisation of the AD tape resource automatically. See `stan/math/rev/core/init_chainablestack_tbb.hpp`.

Note: This PR is a branch from the `feature/intel-tbb-lib` such that it includes these changes in addition to the actual changes introduced in addition. As reference branch the `feature/intel-tbb-lib` has been selected to make it easier to follow the actual changes introduced relative to the base branch.

## Performance Tests

All reported times are in seconds.

| threads | Ubuntu i7 develop | Ubuntu i7 TBB | Ubuntu i7 speedup | Ubuntu AMD develop | Ubuntu AMD TBB | Ubuntu AMD speedup |
|:-------:|:-----------------:|:-------------:|:-----------------:|:------------------:|:--------------:|:------------------:|
|    1    |        166        |      156      |       0.060       |       166.768      |     148.131    |       -0.125       |
|    2    |         91        |       84      |       0.077      |       90.459       |     155.651    |        0.418       |
|    4    |         80        |       45      |        0.438       |       49.331       |     93.949     |        0.475       |
|    8    |         76        |       42      |        0.448       |       28.109       |     80.378     |        0.650       |
|    12   |                   |               |                   |       22.028       |     72.776     |        0.697       |
|    14   |                   |               |                   |       22.421       |     72.742     |        0.697       |
|    16   |                   |               |                   |       23.166       |     66.685     |        0.653       |
|    18   |                   |               |                   |       23.056       |     69.001     |        0.666       |
|    24   |                   |               |                   |       23.525       |     80.779     |        0.709       |
|    32   |                   |               |                   |       23.047       |     87.403     |        0.736       |

| threads | Mac i9 develop | Mac i9 TBB | Mac i9 speedup |
|:-------:|:--------------:|:----------:|:--------------:|
|    1    |       253      |     184    |      0.272     |
|    2    |       145      |     103    |      0.290     |
| 4       | 91             | 64         | 0.297          |
| 6       | 81             | 55         | 0.321          |

| threads | Win i7 develop | Win i7 TBB | Win i7 speedup |
|:-------:|:--------------:|:----------:|:--------------:|
|    1    |       452      |     464    |     -0.027     |
|    2    |       222      |     230    |     -0.036     |
| 4       | 148            | 132        | 0.108          |
| 6       | 146            | 116        | 0.205          |

## Tests

- `test/unit/math/prim/core/init_threadpool_tbb_test.cpp` tests initialisation of the threadpool

- `test/unit/math/prim/core/init_threadpool_tbb_late_test.cpp` tests initialisation of the threadpool in case a c++ user has already initialized the scheduler (in which case nothing happens)

- `test/unit/math/rev/mat/functor/gradient_test.cpp` tests if the threaded gradient code runs fine when using the Intel TBB as backend. This tests if the effects of AD tape initialisation have taken place.

## Side Effects

The initialisation of the TBB threadpool is only optional. That is, should client code not call the `init_threadpool_tbb` function, then the first execution of any TBB code will trigger default initialisation of the TBB threadpool which is the default behaviour of the TBB. In case the client code initialises first the TBB through explicit instantiation of the `tbb::task_scheduler_init` interface, then a subsequent call to the `init_threadpool_tbb` has no effect - this is again the default behaviour of the TBB (only the very first call is honoured).

**For interfaces** it is recommended to make a call to `stan::math::init_threadpool_tbb()` prior to using Stan-math. This will ensure the proper initialisation of the threadpool and ensure that no more `STAN_NUM_THREADS` threads are running in the threadpool.

## Todo

- agreement on design
- a few more tests for `init_threadpool_tbb`
- agree on the fate of `stan::math::get_num_threads` (moved to `stan/math/prim/scal/fun/get_num_threads.hpp`), see comments within

## Checklist


stan/math/prim/scal.hpp
```diff
@@ -78,6 +79,7 @@
 #include <stan/math/prim/scal/fun/identity_free.hpp>
 #include <stan/math/prim/scal/fun/if_else.hpp>
 #include <stan/math/prim/scal/fun/inc_beta.hpp>
+#include <stan/math/prim/scal/fun/init_threadpool_tbb.hpp>
```
SteveBronder: Is there a reason these sit in prim/scal/fun? Maybe we should have a `backend` folder where we put both TBB and OpenCL stuff

stan/math/prim/scal/fun/get_num_threads.hpp
```diff
@@ -0,0 +1,76 @@
...
+ * @throws std::runtime_error if the value of STAN_NUM_THREADS env. variable
+ * is invalid
+ */
+// TODO(SW):
+// 1. get rid of num_jobs argument which is not needed to find
+// out about number of threads we need to use.
+// 2. Consider to switch from std::thread::hardware_concurrency() to
+// tbb::task_scheduler_init::default_num_threads() once TBB is
+// mandatory.
+// 3. pull out of internal?
```
SteveBronder: Would you need to use get_num_threads outside of this file?

stan/math/prim/scal/fun/get_num_threads.hpp
```diff
@@ -0,0 +1,76 @@
...
+inline int get_num_threads(int num_jobs) {
+  int num_threads = 1;
+#ifdef STAN_THREADS
+  const char* env_stan_num_threads = std::getenv("STAN_NUM_THREADS");
```
SteveBronder: Sidenote: We should start thinking about how to add the number of threads and the OpenCL devices so they are flags at runtime

stan/math/prim/scal.hpp
wds15: I am open to suggestions. Another option could be `stan/math/core` which is motivated by `rev/core` - but I am really open to where to stick this; I thought `prim/scal/fun` reflects the right dependency which is why I put it there.

stan/math/prim/scal/fun/get_num_threads.hpp
wds15: In the future I can imagine that client code will want to find out about `STAN_NUM_THREADS` and at that point it would be good to have this in the `stan::math` namespace... but maybe that is not needed and we can keep it where it is.

stan/math/prim/scal/fun/get_num_threads.hpp
wds15: Maybe... not sure. There are pros and cons for both ways (runtime environment vs interface configuration). I am continuing here the system we started for `map_rect`.

stan/math/prim/scal.hpp
rok-cesnovar: I think core is fine. I imagine there wont be much "TBB core" code outside of this which would justify making another folder. Other TBB code is going to live either inside of Stan code or specific stan math functions right?

rok-cesnovar: Either way this is a review-time detail.

stan/math/prim/scal/fun/get_num_threads.hpp
```diff
@@ -0,0 +1,76 @@
...
+ *   available cores on the machine
+ * - STAN_NUM_THREADS < -1, STAN_NUM_THREADS = 0 or STAN_NUM_THREADS is
+ *   not numeric => throws an exception
+ *
+ * Should num_threads exceed the number of jobs, then num_threads will
+ * be set equal to the number of jobs.
+ *
+ * @param num_jobs number of jobs
+ * @return number of threads to use
+ * @throws std::runtime_error if the value of STAN_NUM_THREADS env. variable
```
rok-cesnovar: This is more of a final review than what you need right now, but this throws an std::invalid_argument.

stan/math/prim/scal/fun/get_num_threads.hpp
```diff
@@ -0,0 +1,76 @@
...
+// 3. pull out of internal?
+inline int get_num_threads(int num_jobs) {
+  int num_threads = 1;
+#ifdef STAN_THREADS
+  const char* env_stan_num_threads = std::getenv("STAN_NUM_THREADS");
+  if (env_stan_num_threads != nullptr) {
+    try {
+      const int env_num_threads
+          = boost::lexical_cast<int>(env_stan_num_threads);
+      if (env_num_threads > 0)
```
SteveBronder: personally, especially with multiple if elses, I like having brackets to show leaving and entering a scope

stan/math/prim/scal.hpp
wds15: The `map_rect` which uses the TBB will, of course, stay where it is. So not much other TBB code (if any) will go into `stan/math/core`. It just needs to be included always.

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * active if there is already another instance created before calling
+ * this function.
+ *
+ * @param stack_size sets the stack size of each thread; the default 0
+ * let's the TBB choose the stack size
+ * @return active status of static tbb::task_scheduler_init
+ * @throws std::runtime_error if the value of STAN_NUM_THREADS env. variable
+ * is invalid
+ */
+inline bool init_threadpool_tbb(tbb::stack_size_type stack_size = 0) {
```
rok-cesnovar: What happens in someone calls this more than once? Should this throw, return silently without changing anything? Is there a TBB API call you can make to check whether the threadpool was initialized?

stan/math/prim/scal/fun/get_num_threads.hpp
```diff
@@ -0,0 +1,76 @@
...
+        num_threads = env_num_threads;
+      else if (env_num_threads == -1)
+        num_threads = std::thread::hardware_concurrency();
```
SteveBronder: From [the docs](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency)

> Number of concurrent threads supported. If the value is not well defined or not computable, returns ​0​

Should we have a check here and if the returned value is zero return 1?

Also does this return the number of logical or physical cores? I think previously we saw that hyperthreading etc was never that useful

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace math {
+
+/**
+ * Initialize the Intel TBB threadpool through the
+ * tbb::task_scheduler_init object. In case an instance of the
+ * tbb::task_scheduler_object is instantiated, then this has no effect,
+ * which is the default behavior of the Intel TBB library.
+ *
+ * The maximal number of threads is read from the environment variable
+ * STAN_NUM_THREADS using internal::get_num_threads. See conventions
```
SteveBronder: Well that answers my question lol. Let's put these in the same file, or do we use it elsewhere?

stan/math/prim/scal/fun/get_num_threads.hpp
SteveBronder: (moving all my comments up to here so that it's easier for others to read the code from the github page)

SteveBronder: line 50: personally, especially with multiple if elses, I like having brackets to show leaving and entering a scope

SteveBronder: From [the docs for hardware_concurrency](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency)

> Number of concurrent threads supported. If the value is not well defined or not computable, returns ​0​

Should we have a check here and if the returned value is zero return 1?

Also does this return the number of logical or physical cores? I think previously we saw that hyperthreading etc was not that useful

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
SteveBronder: Just so I'm clear, `stack_size` set the amount of memory each thread has at initialization right? Is there a reason to say 0 as the default and not something like 64 just to give them a bit?

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * is invalid
+ */
+inline bool init_threadpool_tbb(tbb::stack_size_type stack_size = 0) {
+#ifdef STAN_THREADS
+  int tbb_max_threads = internal::get_num_threads(
+      tbb::task_scheduler_init::default_num_threads());
+
+  static tbb::task_scheduler_init tbb_scheduler(tbb_max_threads, stack_size);
+
+  return tbb_scheduler.is_active();
```
SteveBronder: From [docs](https://www.threadingbuildingblocks.org/docs/help/index.htm)

> By default, Intel TBB 2.2 automatically creates a task scheduler the first time that a thread uses task scheduling services and destroys it when the last such thread exits. 

Does this function have a side effect somewhere that actually turns on some global scheduler?

stan/math/rev/core/init_chainablestack_tbb.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <iostream>
+#include <unordered_map>
+#include <thread>
+#include <utility>
+#include <tuple>
+
+namespace stan {
+namespace math {
+
+static std::mutex cout_mutex;
```
SteveBronder: For stuff like this I find it helpful to have a `// FIXME(DELETE)` somewhere just so I remember to delete this stuff

stan/math/rev/core/init_chainablestack_tbb.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+// initialize AD tape for TBB threads
+struct ad_tape_observer : public tbb::task_scheduler_observer {
```
SteveBronder: General docs would be nice. With this stuff especially I, Rok, pretty much any other reviewer will be reading a lot of these methods and schemes for the first time. So a little doc here is nice to lay down intent

stan/math/rev/core/init_chainablestack_tbb.hpp
```diff
@@ -0,0 +1,79 @@
...
+  ad_tape_observer() : tbb::task_scheduler_observer(), ad_tape_() {
+    std::cout << "Observer constructing..." << std::endl;
+    observe(true);  // activate the observer
+    std::cout << "Observer created" << std::endl;
+  }
+
+  ~ad_tape_observer() { std::cout << "Observer destructed" << std::endl; }
+
+  void on_scheduler_entry(bool worker) {
+    std::lock_guard<std::mutex> cout_lock(cout_mutex);
```
SteveBronder: For clarification, the lock here is for cout and not for anything else here?

stan/math/rev/core/init_chainablestack_tbb.hpp
```diff
@@ -0,0 +1,79 @@
...
+    delete elem->second;
+    ad_tape_.erase(elem);
+    if (worker)
+      std::cout << "a worker thread is finished to do some work." << std::endl;
+    else
+      std::cout << "the main thread is finished to do some work." << std::endl;
+  }
+
+ private:
+  std::unordered_map<std::thread::id, ChainableStack*> ad_tape_;
```
SteveBronder: 1. You can use `using` to reuse this type in the code for a little nicer readability

2. Why not a vector pair? I've only really found unordered_map to actually be faster when you have a pretty large N

test/unit/math/rev/mat/functor/gradient_test.cpp
```diff
@@ -1,5 +1,9 @@
 #include <stan/math/rev/mat.hpp>
 #include <gtest/gtest.h>
+
+#include <tbb/parallel_for.h>
+#include <tbb/blocked_range.h>
```
SteveBronder: Should be both be in an ifdef for now?

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
rok-cesnovar: Is that what is_active does? Should we return immediately if is_active is true?

rok-cesnovar:
From a design perspective I agree with how this is done. I dont think you can make a more clean solution here. There are minor review-time issues with braces, docs and such but I think Steve has already flagged most of those.

Initializing the threadpool is straightforward now with another layer and the details of get_num_threads hidden away. The open question there is what to do if someone calls the init threadpool more than once.

The chainable stack is also pretty straightforward, with the only question, I think, being whether unordered map is the way to go or not.

Can we write any sort of tests for the get_num_threads function. That is going to be tricky probably.

rok-cesnovar:
How much additional work is changing map_rect going to be? I am starting to think this is so close we need to merge the Cmdstan and Stan Math TBB PRs, the former does still need some work. Otherwise reviewing and preparing all these branches is going to be a huge pain for everyone.

For the sake of complying with our agreement made at the Stan Math meeting, I am stating here that take care of reverting all PRs if we, for some reason, fail to meet the demands we set .

stan/math/prim/scal/fun/get_num_threads.hpp
wds15: The `get_num_jobs` function is right now used inside `map_rect`, but its use will go away there once we switch over `map_rect` to use the TBB.

When we switch over `map_rect` to the TBB, we can then include this function inside the `init_threadpool_tbb.hpp` file, sure. Then we should also remove the `num_jobs` argument which is really not needed. The `get_num_threads` should only parse the `STAN_NUM_THREADS` and that's it.

The value `-1` means "use all cores". I would change this to then return the TBB default for that. As I recall that is without hyper-threading as I saw that last time.

stan/math/prim/scal/fun/get_num_threads.hpp
wds15: oh, you are right... will fix.

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
wds15: Comment 1: First note that the `tbb::task_scheduler_init` is **static**. So you ever create a single instance of this thing. Any further call is basically ignored. This is in line with the TBB way of working in this case. Repeated initialisation is not possible. You do it once and that's it - no more reconfiguration after that. If you want to have differently sized working areas then you can use so-called `task_arenas`.

Comment 2: Why should we return immediately? In case the *static* init object is active, then this one was used (and the parsing has worked) so that we will just return again `true` if it was true before.

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
wds15: like I say... we put it into one file once `map_rect` is TBBified.

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
wds15: 0 let's the TBB choose that initial size. We should leave the default here as is, I think.

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
wds15: The side effect is initialisation of the threadpool with the scheduler.

stan/math/rev/core/init_chainablestack_tbb.hpp
wds15: Sure... I was in a hurry yesterday night... can add some URL refs to TBB doc, for example.

stan/math/rev/core/init_chainablestack_tbb.hpp
wds15: the lock will go away once the debugging prints are out.

stan/math/rev/core/init_chainablestack_tbb.hpp
wds15: unordered_map is just a convenient way to have an index set which is indexed with `std::thread_id` keys. Speed is totally irrelevant here (this is called just once per created thread for the entire life-time of the threading in the pool). So I want to use `std::thread_id` as a key.

test/unit/math/rev/mat/functor/gradient_test.cpp
wds15: yes, they should; will do.

wds15:
There are three cases which will lead to no effect for a call to `init_threadpool_tbb`:

- Initializing with `tbb::task_scheduler_init` the TBB threadpool *before* calling `init_threadpool_tbb`. In that case the return will be `false` for the active state.
- Calling `init_threadpool_tbb` repeatedly will lead to initialising the threadpool on the first time. Then the active state returned will always be `true`. Any call after the first will not change anything.
- If you use any TBB code which leads to `tasks` being dispatched (so any access to scheduling facilities), then the TBB code will on the spot initialise its threadpool using defaults (use all cores). In that case calling `init_threadpool_tbb` will have no effect and the active state is `false`.

This is the default behaviour of the TBB. I don't see why we should deviate from that. I can add some URLs to the TBB doc if that helps?

`get_num_threads` is already well tested in `test/unit//math/prim/mat/functor/num_threads_test.cpp`... those need to be adapted when we take away the `num_jobs` argument.

For the observer: I want to use `std::thread_id` as a key and as such I have to use an `unordered_map`...I don't see another way doing that otherwise conveniently.

I can certainly add the `map_rect` code here. That is straightforward. I just thought it would be better to do it in a follow-up PR, no? Either way is fine with me.

I will address these things later tonight. Let me know if my responses don't make sense / you think differently / I misunderstood.

With this PR I would also like to update the wiki page which talks about threading.

SteveBronder:
Also it looks like the design here differs from the design doc's ScopedChainableStack?

https://github.com/wds15/design-docs/blob/parallel_autodiff/designs/0003-parallel_autodiff.md#step-2-independent-ad

wds15:
What do you think differs? I am relatively sure that things are consistent with one another (maybe not the same... the parallel design RFC is months old).

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
SteveBronder: Would be good to add that as a @note in the doxygen

stan/math/prim/scal/fun/init_threadpool_tbb.hpp
SteveBronder: Gah totally missed that in the docs my bad

rok-cesnovar:
> There are three cases which will lead to no effect for a call to `init_threadpool_tbb`:
> This is the default behaviour of the TBB. I don't see why we should deviate from that.

I just wanted to double check what would happen. A line in the doc stating something like "calling this after the threadpool has already been initialized will have no effect" would suffice in my opinion.

> `get_num_threads` is already well tested in `test/unit//math/prim/mat/functor/num_threads_test.cpp`... those need to be adapted when we take away the `num_jobs` argument.

:+1: 

> I can certainly add the `map_rect` code here. That is straightforward. I just thought it would be better to do it in a follow-up PR, no? Either way is fine with me.

I would do it here. If you say the map_rect changes are small, I think adding them here makes sense. The only non-trivial thing here is the observer code.
I would then run the warfarin testbench we did for TLSv* (I can run it on Ubuntu and Windows) to double check everything is as expected. Would that work for you?

wds15:
I think I got all your review comments now.

The only thing left to do (coming to my mind right now) are a few more tests for the init_threadpool_tbb function and a wiki page update.

Let me know if the doc is now fine to your eyes or more is needed.

bob-carpenter:
> +  const char* env_stan_num_threads = std::getenv("STAN_NUM_THREADS");
> +  if (env_stan_num_threads != nullptr) {
> +    try {
> +      const int env_num_threads
> +          = boost::lexical_cast<int>(env_stan_num_threads);
> +      if (env_num_threads > 0)
> 
> personally, especially with multiple if elses, I like having brackets to show leaving and entering a scope

Rather than:

if (error condition) {
  handle error, return, or throw
}
non-error condition

the following reduces nesting of the main code one level:

if (non-error) {
  non-error code
} else {
  error code 
}

bob-carpenter:
> ... The open question there is what to do if someone calls the init threadpool more than once.

Can you just use the lazy singleton pattern in C++?

const Foo& lazyFoo(...) {
  static const Foo x(...);
  return x;
}

Static constant locals are guaranteed to be initialized thread safely exactly once.  If you need to wrap up something like a C-style call inside, you can put it in the constructor of Foo.

wds15:
@bob-carpenter this is exactly the pattern I am using. It's just that the TBB default API is also involved here.

test/unit/math/prim/scal/fun/init_threadpool_tbb_test.cpp
```diff
@@ -0,0 +1,26 @@
+#include <stan/math/prim/core.hpp>
+
+#include <gtest/gtest.h>
+#include <test/unit/util.hpp>
+#include <test/unit/math/prim/mat/functor/utils_threads.hpp>
+
+#include <tbb/task_scheduler_init.h>
+#include <tbb/task_arena.h>
```
rok-cesnovar: These two should be inside `#ifdef STAN_THREADS`

wds15: thx for catching that.

rok-cesnovar:
We have 9 days until the feature freeze and this is more or less done which means TBB will get in the 2.21 release barring any unexpected events.

Can you you do the request changes on the cmdstan TBB PR https://github.com/stan-dev/cmdstan/pull/710 

I will review it in the afternoon/evening if you have time to fix it today. Once that is merged we can restart the Stan Math TBB PR tests and also get that merged.

Then we can have a final final pass of this PR + do performance test with all the branches in place. How do you feel about that?

wds15:
That sounds great. Let me look at CmdStan quickly today.

We still need documentation on the wiki and in cmdstan manual. In particular clear instructions for windwows (mingw32-make & PATH variable).

But all of that sounds very much doable. The threading tests just passed so that all should be ready code wise (apart from the last glitches).

I will also create a ticket for cleanup of the codebase from the extra ifdefs which we can lift once the TBB is mandatory. 

How I feel? Well, I believe it when it's in - but I have been working towards this at least 6 months (probably more) - so that prospect is really awesome! Thanks for the quick turn-around. 2.21 will be a BIG release from my perspective just for this feature.

rok-cesnovar:
Great! 

I added the requirements to the CmdStan PR so things dont get lost in the discussion.

bob-carpenter:
If it works that way, then no need to code around it.  

Usually the logic for an unexpected input is different and returns before going into the main loop.  But it's irrelevant if the TBB is already managing all this.

wds15:
argh... looks like there is some windows hiccup...will follow-up with that later today; let's cross fingers I can debug it locally in my virtual machine.

rok-cesnovar:
I see that you have probably already solved the issue at about the same time. I see that I get a similar error on my Windows machine if PATH isnt set. 

`runTests.py -j16 test/unit -f thread` passes on Windows on this branch with the path variable set.

wds15:
Yes - looks like PATH must be set adequately, which makes a lot of sense.

wds15:
Ok... adding PATH correctly on Windows fixes the issues. Right now I have the path to the tbb library hard-coded into the Jenkinsfile which is not too nice... can we make updating this part of the update script? Better ideas?

(when I saw this I was worried... these windows hiccups can kill you... glad its ironed out)

wds15:
Hmm... there is a problem with updating the runTests.py in math via the update command... this change needs to be done to the upstream repositories as well. So whenever stan or cmdstan runs any tests on windows then they must set the PATH variable on Windows.

One solution is to place the tbb builds into

`stan-math-root/lib/tbb_bin/`

So we exclude the version from the path. Doing so will allow us to refer to this directory without a version of TBB and thus the PATH will point to the same location regardless of the version of TBB. Does that make sense to you? Other ideas?

rok-cesnovar:
Yeah, people having to upgrade path after every release on Windows is another reason to go this route. /lib/tbb doesnt seem that bad. We just need to add a file inside with the version number so its easy to spot. Let me sleep on it, to see if I have any other ideas.

rok-cesnovar:
Lets go this route of removing version from the path. Any specific reason you want to name the folder `lib/tbb_bin` and not just `/lib/tbb`?

I would an empty file `/lib/tbb/tbb_2019_U8.ver` in the folder, which would make it easy to spot the version number and also simplify the update script. 

If you agree with what I propose here, I can modify the update script.

wds15:
Sounds good to me as you suggest - and yes, go ahead to modify the update script; I can handle the makefile pieces if you want.

wds15:
Given the clarification from @syclik [here](https://discourse.mc-stan.org/t/tbb-and-2-21/11179/14) we should probably remove all the extra if-defs and make the TBB mandatory. Do you agree, @rok-cesnovar ?

rok-cesnovar:
> I can handle the makefile pieces if you want.

Please do yes.

wds15:
argh... now I made the makefile edits on this branch and not on the other one...

Any suggestions? We could merge this PR into the lib PR and then merge that thing into develop. Is that an option?

rok-cesnovar:
Go ahead and merge this branch to the other PR. Please update the text there. 
These files are going to be listed at the beginning of end of changes on the PR, so I have no issue with that.

What is left to be done: 
- cmdstan manual doc
- windows path setting or some other way of solving that problem. I tried thex setx command yesterday and that seemed to work.
- perf test (I am aiming to do the Linux/Windows ones today)

wds15:
Let me try to cherry-pick commit 19300fd to the other makefile branch. That should work without too much headache and let's us proceed as we intended to.

I can run performance tests on macOS - can you share the files or point me to them on the forum to make sure I use the same and also let me know what settings to use (iterations/warmup).

I would hope that this is all ready by the weekend. So cool.

rok-cesnovar:
I propose we use what @syclik has prepared for the tls performance tests https://github.com/syclik/benchmark-stan/tree/math/1245-tls

We only need to run the warfarin case here (with 1,2 & 4 threads for example). I am working on updating the submodule references to the right spots and to modifiyng the scripts so they only run what we need. These were simple tests to run, and everyone ran the same settings. Does that sound good to you?

wds15:
Ok... will you fork that repository and update the references? Then I could start from there.

rok-cesnovar:
I have push permissions to that repository (just tested). So the branch will be `math/tbb`. Will tag you once that is ready.

rok-cesnovar:
Its ready and its working: https://github.com/syclik/benchmark-stan/tree/math/tbb . Clone (this one takes a while) and just run `run-thread.sh`. It works "out-of-the-box" on Linux, it should also on Mac.

One question does arise from these performance test. This is now always running with the default TBB threadpool because the init_threadpool function is not called, which means STAN_NUM_THREADS is not used.

Do we have a simple solution of calling it in cmdstan (would need an additional PR after all of this is merged)? Any other ways of doing this?

wds15:
Will run it later today.

This is how I intended things to work - so yes, we have to make in CmdStan a call to the init_threadpool_tbb function. Right now the default behaviour of the TBB kicks in, which is to use all cores you have, right?

I suggest including the TBB init in a second CmdStan PR which I branch of the existing one. I can prepare that quickly.

rok-cesnovar:
And to the REALLY good news. This is preliminary results from 2 runs each on Ubuntu but:
- fastest time for develop with STAN_NUM_THREADS=8 was **80s** (tried with 4, 6 & 8)
- with TBB **52s** (uses 8 threads because of the default threadpool)

I was a bit surprised this had such an effect, so I ran the TBB one a few more times (49s to 55s).

Great stuff @wds15 !!

wds15:
Thanks... I know... it's huge! You are getting MPI performance basically (without the pain of having to setup MPI).

wds15:
Here you go:

https://github.com/stan-dev/cmdstan/tree/feature/intel-tbb-init

This branch honors the `STAN_NUM_THREADS` variable correctly.

rok-cesnovar:
Updated the performance script branch to use this, running the full tests now.

wds15:
I am running this right now... whow 10 iterations with 1 core take forever. Now, this also shows that we get with 1 core significant speed differences:

- old: ~250s
- new: ~180s

Again - this is with one core only. So the difference here is the use of the tbbmalloc library. It really hurts to not turn this on by default. With absolute certainty this is on Intel CPUs an absolute win to do so. Thus, we really have to document it boldly that Intel users should always turn on the tbbmalloc_proxy library. It's going to take a while until all these 1 core runs are done.

rok-cesnovar:
> whow 10 iterations with 1 core take forever

Yeah, I know :/ 

My performance test are close to done. 

On Windows it isnt such a huge win, but still on-par or a bit better than what we had before. 

But threading on Windows has a huge overhead anyways. On the exact same system, 8 threads on Windows needs almost 3times as much time as 8 threads on Ubuntu (with TBB and without).

rok-cesnovar:
@wds15 on Windows I am seeing this message "TBB Warning: Exact exception propagation is requested by application but the linked library is built without support for it" when rejection happens.

```
TBB Warning: Exact exception propagation is requested by application but the linked library is built without support for it
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: Exception: Exception: gamma_lpdf: Shape parameter[1] is nan, but must be > 0!  (in '../benchmark-warfarin-new.stan' at line 48)
  (in '../benchmark-warfarin-new.stan' at line 160)
  (in '../benchmark-warfarin-new.stan' at line 225)
 [origin: unknown original type]
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.
```

Can we do something about this?

wds15:
Maybe... could be a Windows / mingw relict. see:
https://software.intel.com/en-us/blogs/2014/08/28/exact-exception-propagation-in-the-intel-threading-building-blocks-library-and-c11

So I will specifically turn c++11 on in the makefiles for tbb.

wds15:
wrt to windows threads being slow: As I understood there are multiple Windows threading implementations around for mingw. The one in RTools 3.5 is probably terribly old and slow. It would be interesting to test RTools 4.0 which brings a much newer toolchain (we don't need to do this now; it's a thing to try out at some point).

rok-cesnovar:
Ok so here are my performance test results. For both Windows I used a Intel i7-4790 CPU that has 4 physical cores (8 logical).

**TLDR: 35-45% faster execution on Ubuntu, 10-20% faster execution on Windows. Unexpected results on the effect of tbbmalloc, but we are good either way.**

On Windows I decided to experiment a bit, with the general threading performance being poor, not just for TBB. So I removed `tbbmalloc`. What I found is that without `tbbmalloc` things ran faster. Both cases are faster than the existing one for all cases we care about, so we are fine either way. 

I was then also curios what happens on Ubuntu. And to my surprise it had the same effect. Will run some test with less iterations on my laptops i5 tomorrow if I have time.

Ubuntu:

| num of threads | develop | TBB  | TBB no tbbmalloc |
|----------------|---------|------|------------------|
| 1              | 166s    | 179s | 156s           |
| 2              | 91s     | 94s | 84s |
| 4              | 80s     | 51s | 45s    |
| 8              | 76s     | 48s | 42s |

Windows:

| num of threads | develop | TBB  | TBB no tbbmalloc |
|----------------|---------|------|------------------|
| 1              | 452s    | 468s | 464s             |
| 2              | 222s    | 233s | 230s             |
| 4              | 148s    | 133s | 132s             |
| 8              | 146s    | 131s | 116s             |

wds15:
One question: When you say you disable tbbmalloc - do you take out tbbmalloc **and** tbbmalloc_proxy?

Can you decrease the number of repeats to 5 and bump the stan-math git ref for the performance tests? I will then run on my mac with and without tbbmalloc. Maybe this is even Mac specific which would mean to not turn it on by default probably.

When writing the last tests for the threadpool init I noticed that `is_active` does not behave as I expected. I change the behavior of the init function slightly:

- returning now a reference to the init scheduler object so that people can call terminate on it if they wish to. We could also return void - but the active status of the scheduler is pointless, since the schedule will be activated upon construction of the init object (either that happened before or it happens with the static instance in the function)
- instead of asking tbb for default_num_threads I am turning this back to std::thread::hardware_cocnurrency() just as before. These two seem to give the same result, but we should not change it, I suppose.

I will add now missing docs.

rok-cesnovar:
> do you take out tbbmalloc **and** tbbmalloc_proxy

Yes.

> Can you decrease the number of repeats to 5 and bump the stan-math git ref for the performance tests?

Done. Pull the branch from benchmark-stan.

> I change the behavior of the init function slightly

OK. I will take a look later in the afternoon.

> I will add now missing docs.

And I will move the TBB source to /tbb and update the update script. Probably in the evening. I havent forgotten.

SteveBronder:
If yinz need an AMD cpu to test this on lmk!

rok-cesnovar:
If you have free compute cycles it would not hurt :) The more we know, the better. You have an AMD on the mac machine or linux?

Steps to reproduce: 

- clone https://github.com/syclik/benchmark-stan/ , switch to math/tbb branch
- open the run-threads.sh file and change FOLDER on line 38 accordingly
- optionally add numbers to the threads array on line 44 if you have more than 8 cores

You can probably put N=3 (line 43), these tests are long enough, there isnt much variation.

If you also want to see what happens without `tbbmalloc` add the following line on line 35:
echo "TBB_LIBRARIES = tbb" >> local

Thank you!

wds15:
So here are my results with the `tbbmalloc` on my laptop where I have up to 6 physical cores:

| threads| new| old|
|-------:|---:|---:|
|       1| 184| 253|
|       2| 103| 145|
|       4|  64|  91|
|       6|  55|  81|

The time-unit is seconds here. So we see a 50% speed increase here!

I will run the pipeline once more without the `tbbmalloc`. Maybe we only want `tbbmalloc` on MacOS (which always runs on Intel).

I have now also updated the Wiki in stan-math. It will need a final update once we settle on `tbbmalloc`. The benchmark results should go into the PR description and then we are good, I think - or did I miss anything?

Note that I also cherry-picked the commit with the makefile changes such that these are now in the other PR which handles the makefile and the lib. So please do the update-tbb.sh changes there.

SteveBronder:
Mamma mia!

image

Pretty nice! speedup in table below is `1 - (new / old)`

| threads|     new|     old|    speedup|
|-------:|-------:|-------:|----------:|
|       1| 166.768| 148.131| -0.1258143|
|       2|  90.459| 155.651|  0.4188344|
|       4|  49.331|  93.949|  0.4749172|
|       8|  28.109|  80.378|  0.6502899|
|      12|  22.028|  72.776|  0.6973178|
|      14|  22.421|  72.742|  0.6917737|
|      16|  23.166|  66.685|  0.6526055|
|      18|  23.056|  69.001|  0.6658599|
|      24|  23.525|  80.779|  0.7087733|
|      32|  23.047|  87.403|  0.7363134|

User timings here as well. not sure how to interpret when threads > number of cores (16). I'm gonna try this with tbbmalloc off and see if things are different

image


[Data Here](https://gist.github.com/SteveBronder/cec78b0f5da15bb04fdd1739cd2cd25b)

rok-cesnovar:
@SteveBronder Thank you!! Fantastic. Sebastian has really done a great job here. 2.21 is going to be a huge update.

@wds15 

I updated the update script, moved tbb files to /tbb and add a .ver version  file in /tbb so the version is more obvious. 

The status of the TBB PRs:

- TBB lib PR in Stan Math is done. I went through the changes again and approved. My preference would be to keep that PR freezed as of now and do any modifications to how we handle tbbmalloc here. Otherwise its hard to keep track of various changes. **TLDR: TBB lib PR is DONE**

- TBB map rect PR in Stan Math (this PR) is ready except for tbbmalloc. Based on the results of our tests I would always link tbbmalloc on Mac  (STAN_THREADS or not) and never on other OSes. If you agree please update the makefiles here (maybe wait for Steve to update with what happens on his system). Also merge tbb-lib branch with the recent changes. **TLDR: merge with changes to lib + tbbmaloc q and its done** I will also edit your top comment with the performance results. Sebastian should then post that to Discourse.

- Cmdstan PR: manual and whatever we want to do with the PATH on Windows. I would be fine if we went with manual update from user, but the consensus on the meeting was that we need to find a solution that a user can do `make install-tbb` that either sets the path or whatever works. The setx command looked promising but didnt have time to explore it more. **TLDR: Manual and windows Path and its done**

wds15:
Well - it's the TBB doing the work!

My benchmarks are ongoing without the tbbmalloc right now - and it looks to me as if that has a major impact on MacOS. We will very likely endup turning `tbbmalloc` on for MacOS only; let's see.

We should use the Jenkins feature to run PRs with parameters. This way we can point Jenkins to use specific branches upstream.

I updated the cmdstan manual and also went over our threading wiki page. Both need minor updates depending on `tbbmalloc` and the final `PATH` solution for windows.

@rok-cesnovar can you take on the `PATH` thing? In case we endup with an additional make target to call we should print what the user needs to do on Windows when CmdStan is build - but let's discuss there.

rok-cesnovar:
> We should use the Jenkins feature to run PRs with parameters. This way we can point Jenkins to use specific branches upstream.

I think we can just merge Cmdstan PR when its ready, then TBB lib and finally this one. Its going to require some manual restart, but I can handle that.

> @rok-cesnovar can you take on the `PATH` thing?

On it. Will also review the manual changes.

SteveBronder:
I added `echo "TBB_LIBRARIES = tbb" >> local` to the sh script and got p similar results, looks like tbbmalloc is fine?

| threads|     new|     old|    speedup|
|-------:|-------:|-------:|----------:|
|       1| 170.347| 148.191| -0.1495098|
|       2|  89.947| 155.819|  0.4227469|
|       4|  49.146|  94.204|  0.4783024|
|       8|  27.955|  80.422|  0.6523961|
|      12|  22.039|  72.865|  0.6975365|
|      14|  22.535|  72.527|  0.6892881|
|      16|  23.037|  66.884|  0.6555678|
|      18|  23.224|  68.980|  0.6633227|
|      24|  23.435|  80.950|  0.7105003|
|      32|  22.933|  87.927|  0.7391814|

rok-cesnovar:
It might only be my workstation. I will do a quick check on my notebook tmrw, its only an i5, but I also have a dual boot there so we should get some more sense of this on Ubuntu/Win.

wds15:
Here are my results *without* `tbbmalloc`:

| threads| new| old|
|-------:|---:|---:|
|       1| 247| 245|
|       2| 138| 142|
|       4|  84|  88|
|       6|  75|  82|

You see that on MacOS using the `tbbmalloc` is really important!

Since we do not see benefits on other platforms, I suggest that we use the `tbbmalloc` on MacOS while we don't do that on the other platforms. Does that make sense to everyone?

rok-cesnovar:
Yes, based on these results I would agree.

rok-cesnovar:
I changed the base branch of this PR to develop. If tests pass I will have another look tomorrow morning.

rok-cesnovar:
One side question. Will this also fix the static init order problem for Pystan?

wds15:
the static init order problem is still there and I want to find the time this week to get a fix in (though we have never seen this problem in cmdstan and I haven't yet heard an alert from RStan).

The issues in PyStan seem to be related with the threads used by Python... things are still a bit unclear. However, if now the TBB manages the threads, then there is hope that this solves things for PyStan.

wds15:
Is it normal that the job is sitting there for quite a while, waiting for the integration tests to start? Does someone know?

rok-cesnovar:
Actually not sure if that will help as the test on this PR is associated with the specific Stan upstream test. @serban-nicusor-toptal Is this job just waiting for free resources?

rok-cesnovar:
@wds15 I restarted just the upstream parts, see here: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1376/24/pipeline

Lets see if that will work, otherwise we need a more thorough look.

wds15:
I will do one more update to the Stan-math wiki on threading and simplify the text on the TBB wrt to the tbbmalloc if you are fine with that. So basically we don't need to mention the tbbmalloc there, I think - it's probably just confusing for users. Or do you think the meaning of `TBB_LIBRARIES` should be documented on the wiki page?

Ok. Once this is in I can post a heads-up on discourse that this is coming; and we can include the performance tests.

Thanks for all the reviewing @rok-cesnovar !

rok-cesnovar:
I would do something like:

We now link the TBB on all operating systems by setting: 
TBB_LIBRARIES=tbb

On MacOS we also link tbbmalloc and tbbamalloc proxy as that has proven to provide additional speedups on macOS only (link to the top post in this PR).

I would not go much further than that.

wds15:
@serban-nicusor-toptal would it be possible to have at look at this PR? The Jenkins instance does not seem to run all the downstream tests as the job sits there and is waiting forever. Many thanks.

rok-cesnovar:
I am not 100% on this, but we might have a Windows issue: https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan/detail/downstream_tests/909/pipeline

See also (develop was ok, then weird stuff): https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan/activity

wds15:
Ok.. will debug later today. In case we have missed somewhere to edit PATH during Jenkins, then that will be a problem.

rok-cesnovar:
See https://github.com/stan-dev/cmdstan/pull/742

I believe this solves this issue.

rok-cesnovar:
I restarted just the upstream now, not sure if that uses Cmdstan reference from when the run started or fresh develop. If it fails I will just restart the entire run again.

wds15:
Looks like there is no windows machine up at the moment.

rok-cesnovar:
There is, its just busy doing other stuff: https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/develop/388/pipeline

It doesnt help us right now, but hopefully we will finally finish setting up a Windows box at Univ. of Ljubljana this week (its ready, just needs a static IP assigned by our IT dept.) and donate the free compute cycles to Stan CI. That should help in the future.

serban-nicusor-toptal:
Jenkins couldn't connect to the win instances, giving it a short reboot.

rok-cesnovar:
@serban-nicusor-toptal while I have you here, Jenkins does need a static IP, am I right? If I sent you teamviewer info to set Jenkins up on the machine that wouldnt help right? Also, I know Teamviewer is a pain...

serban-nicusor-toptal:
Hey, the EC2 instance where Jenkins is hosted has a static IP.  I've hard rebooted it and the Jenkins service failed to start for a strange reason, that's why the 5 min downtime. It's back up now.

wds15:
Do we need to start over with the full pipeline or just wait?

serban-nicusor-toptal:
It's resuming the upstream tests

rok-cesnovar:
The tests for this PR ended with a weird error (https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan/detail/downstream_tests/912/pipeline). 

Still cant figure out whether we fixed the issue with the Cmdstan PR or not. Restarting the entire set of test and crossing fingers we either go green or at least know what is up in the morning.

wds15:
This branch has the problem that we did not merge develop into it. Thus, this branch has still a few makefile flaws which we are reintroducing with merging this (but all is ironed out on develop). I am right now verifying that we can deal with spaces in the paths.

rok-cesnovar:
@wds15 have you made any progress? Need help figuring the spaces in path. Is the plan for this PR to merge in develop once we figure out the daily stan perfromance thing?

wds15:
I think I have a solution ready. Hold on for a moment.

wds15:
Let's cross fingers this avoids the issue on the Jenkins systems. We should see an earlier success/failure from here:

https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan/detail/PR-743/3/pipeline

So at least I was able to get things working when being in a directory

`/path/cmdstan space/`

The makefiles first barfed at me and now they can process this.

rok-cesnovar:
No luck on the Windows front for this PR. Could we just somehow use mingw32-make to build the TBB and then use the regular make for everything else. Would that help? This is really frustrating.

wds15:
My hunch is that the tests do change the directories. This then invalidates the PATH variable which requires that you are in the top-level directory of CmdStan. So maybe a Jenkins Guru can help us here ( @serban-nicusor-toptal ).

What we have in the CmdStan Jenkinsfile is this:

```
def runWinTests(String prefix = "") {
    withEnv(['PATH+TBB=./stan/lib/stan_math/lib/tbb']) {
       """ mingw32-make -j${env.PARALLEL} build
         ${prefix}runCmdStanTests.py -j${env.PARALLEL} src/test/interface
       """
    }
}
```

What I would like to have is something like this:

```
def runWinTests(String prefix = "") {
    withEnv(['PATH+TBB=ABSOLUTE-PATH-TO-CMDSTAN/stan/lib/stan_math/lib/tbb']) {
       """ mingw32-make -j${env.PARALLEL} build
         ${prefix}runCmdStanTests.py -j${env.PARALLEL} src/test/interface
       """
    }
}
```

How can this be done?

wds15:
@rok-cesnovar Actually, we can use `mingw32-make` for the Intel TBB and then use the normal `make`. That does work - I tried it.

serban-nicusor-toptal:
@wds15 To get the current working directory you can use the following variable: `"${env.WORKSPACE}"`   
See: https://jenkins.mc-stan.org/env-vars.html/

rok-cesnovar:
@serban-nicusor-toptal would it be possible for you to send me info for the instance this is running on? Or a separate one just like this. I ran this locally on the exact same branches and its all fine. We need to eliminate whether its an issue on the system or somthing Jenkins specific. And go from there.

serban-nicusor-toptal:
https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan%20Performance%20Tests/detail/Custom/57/pipeline/56/
I've ran that custom job and it fails on windows, not sure if tied into this.
@rok-cesnovar will send you server info in a bit.

wds15:
@serban-nicusor-toptal could you run this test again, but use this PR-1376 from Stan-math as math checkout please?

serban-nicusor-toptal:
Started a new build with `math_pr` set to `PR-1376`
See: https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan%20Performance%20Tests/detail/Custom/58/pipeline/52/

wds15:
Ok... the error we get is this:

`ERROR: Please use mingw32-make on Windows to build the Intel TBB library.`

So on windows we have to use `mingw32-make` which we have usually on the Windows machines available. What do we need to reconfigure here?

serban-nicusor-toptal:
The machine should have all the dependencies I've found in readmes and what I got from Sean.
I've spin up a machine for @rok-cesnovar and he'll look into it on the machine,

serban-nicusor-toptal:
@wds15 Yes, you can use `withEnv` or `environment`.
See: https://stackoverflow.com/questions/44378221/export-command-in-jenkins-pipeline/44380495#44380495
I think your code is correct but it should use `"` instead of `'` in here: `withEnv(["PATH+TBB=${env.WORKSPACE}/stan/lib/stan_math/lib/tbb"])`

SteveBronder:
Why does this keep going off?

rok-cesnovar:
We are running a cmdstan PR with a submodule reference to this PR. And that also sends the cmdstan perf results here.

rok-cesnovar:
I am restarting this again. It got stuck in the Stan upstream part with 

```
Cannot contact EC2 (Windows) - Build, Docker, WSL (i-0f525173fa0eaaf67): hudson.remoting.ChannelClosedException: Channel "unknown": Remote call on EC2 (Windows) - Build, Docker, WSL (i-0f525173fa0eaaf67) failed. The channel is closing down or has closed down

Could not connect to EC2 (Windows) - Build, Docker, WSL (i-0f525173fa0eaaf67) to send interrupt signal to process
```
and there was no other way than simply restarting...

wds15:
Again the same error? Is there a Jenkins / Windows problem or a problem on this branch?

rok-cesnovar:
Jenkins/Windows. I restarted the Always run part 2 now.

wds15:
Will this ever go through? I hope this time we go through the Windows stuff...

wds15:
ok let me merge develop into this one since the allow spaces went in.

serban-nicusor-toptal:
Windows seems to be fixed now, shouldn't happen again!

@wds15 It failed at the testing stage, restarted it here: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1376/34/pipeline/

rok-cesnovar:
Yeah, but it still points to PR-744 that was merged and doesnt exist anymore. So it will def fail again. It seems that Windows Jenkins is stable now so hopefully this is all over in the morning CET time.

wds15:
thanks for restarting! I need to merge now develop into this... but looking at stan downstream tests it looks to me as if we need there possibly as well the tbb.dll copy thing?

PR #1401, opened 2019-10-13
Bugfix/tbb cleanup

wds15:
## Summary

This cleans up two smaller things for the TBB:

1. the README.md is (hopefully) now a bit clearer as to why we have some additional constraints due to an Apache 2.0 license.
2. the Intel TBB makefiles have been altered to avoid use of "-g". This reduces the created library files considerably. In particular, the tbb.dll on windows decreases in size from 5.3MB down to 1.3MB. Since we copy this file quite a lot, this is nicer to users disk space.

## Tests

Existing pipeline.

## Side Effects

Less disk space used.

I added a `STAN_CHANGES` file in the tbb directory to comply with the Apache 2.0 requirement to document changes done to the source. Maybe this means that we need to update the update-script for the TBB.

## Checklist

- [x] Math issue #1402 


rok-cesnovar:
Great! Will have a look later today.

> 2\. Since we copy this file quite a lot,

Just so that its clear to anyone reading this. We copy the file a lot on Jenkins for tests because we cant configure the path correctly there and have to resort to this way of solving the problem. The end user does not need to copy any dll's.

wds15:
The latest PR on cmdstan for this reduces copying down to just 4 times of the tbb.dll. Still, some users may choose to not reconfigure their system under Windows... for them a 1.3MB copy vs a 5.4MB copy is a difference.

lib/tbb_2019_U8/STAN_CHANGES
```diff
@@ -0,0 +1,3 @@
+This file documents changes done for the stan-math project
```
rok-cesnovar: I would document all these changes in the update script if you dont mind. I would revert the change in all the *.inc files and remove this file and just run the following (make sure you are in the math folder):

```
find . -regex "./tbb_.*inc" -type f -exec sed -i 's/ -g / /g' {} \;
git add ./tbb_*.inc
```
Then commit these changes separately. This will touch all the .inc files instead of just .gcc and .clang but I think its better to have this done automatically.

Then add these two lines in the 4th step (maybe add a comment explaining what we are doing): 

```
# 4. Prune TBB
cd tbb_${tbb_version}

git rm -rf examples/*
find . -regex "./lib/tbb_.*inc" -type f -exec sed -i 's/ -g / /g' {} \;
git add ./tbb_*.inc
git commit -m "upgrading to TBB version ${tbb_version}; pruning files"

```

rok-cesnovar: I meant in the lib folder, not math. sorry

rok-cesnovar: Oh, I now read your comment that we need to have a list of changes due to the Apache 2.0 license. I doubt that includes the makefile but yeah, I would rather add it just to be sure. 

This means that we should also echo this file in the update script, yuck. If that turns out to be ugly just make the update-script print out: "Create a a STAN_CHANGES file with the list of changes done!".

wds15: I don't think the idea to keep the changes in the script is maintainable. Moreover, I do not want to do a blanket sed on the entire files. For the debug build target the -g should stay in there. Maybe we can cherry-pick the commits, get that diff and apply it to the new tbb directory?

wds15: So maybe something like

```
cd $(TBB)
git diff -p -a a38babf^..a38babf | patch -p3
git add -u
git commit -m "Re-applying a38babf change-set"
```

Can that work?

rok-cesnovar: It probably would, but I think I may be overcomplicating here, idk. 

You are right that a global sed would be bad. Unfortunately we cant make a test that would fail so we would remeber this. That would be the easiest solution.

Maybe just add a note at the end of script to do this? Or store the contents of the STAN_CHANGES and cat them at the end? I really dont expect much more changes to the TBB. :crossed_fingers:

wds15: Well.. the TBB 2019 Update 9 was released 6 days ago and the above scheme would not work with this upgrade. For the moment being I am simply documenting which change-set should be ported. If this is forgotten to be done in the future, then users will have used up more space... which is not nice, but this won't bring any program down.

BTW... should I upgrade to Update 9 ? Looking at the release notes I don't see an immediate need.

rok-cesnovar: If there isnt a specific feature/bugfix we are interested in or there are some performance benefits from upgrading I would be conservative. 

We could also agree on upgrading once before each stan release (or every 2 stan releases which is effectively twice a year) even if there are no direct upgrades. We can discuss that at some later point.

Issue #1306, opened 2019-08-13
static init order problem


wds15:
## Description

Currently, the AD tape initialisation is prone to the static init order catastrophe. This surfaced during the development of PyStan and leads to a segfault.

## Example

The problem is that we require in the `init()` method of `AutodiffStackSingleton` that the pointer to ad instance has been initialised, since we check there if the pointer is equal to `nullptr`. Thus, whenever the pointer to the (optionally thread local) AD tape is not yet initialised, then we run into the static init order problem.

As a solution we need to avoid the double use of the `instance_` pointer. Right now it flags "is not initialised" and it points to the instance we use in the thread when initialised. 

## Expected Output

Our library should not be prone to the static init order problem.

A workaround for C++ users right now is to compile Stan-math with `STAN_MATH_REV_CORE_INIT_CHAINABLESTACK_HPP` being defined. In that case no initialisation is kicked off by our includes. Instead, the C++ user must initialise himself the AD tape at a point in the program when all linked symbols are initialised.

#### Current Version:
v2.20.0

riddell-stan:
Thanks for writing this up @wds15. I really appreciate your work on this.

PyStan is particularly affected by this problem because future version of PyStan will *require* STAN_THREADS to be defined. PyStan 3 (via httpstan) is the only interface that supports parallel sampling (e.g., 4 chains) exclusively via threading. RStan, for example, does sampling for different chains in different processes.

Two notes:
- In debugging this problem, I ran into the static init problem even when STAN_MATH_REV_CORE_INIT_CHAINABLESTACK_HPP was defined.
- The workaround PyStan 3 / httpstan is currently using is not really a workaround at all. Since `clang` does not encounter the static init problem, we drop support for gcc on Linux. Since installing clang is not too difficult on linux, this seems ok. More troubling is that neither gcc nor clang work on osx, both segfault. Oddly, gcc works fine on Windows.

wds15:
I am sorry that you are hit so hard by this problem, but it hasn't surfaced any where else yet.

I do not understand at all how you can be affected by static init order issues whenever you make that define. In that case it's impossible to have that problem...to my understanding of matters, but I can easily miss details which are Python specific (or something else)?

I am happy to debug myself on the matter, but I would need from you some instructions on what I need to setup and on what branch to look at in order to make progress here. I am on mac, but could run on a Linux as well. Mac is preferred, but if Linux is easier then that.

riddell-stan:
[...]

PR #1403, opened 2019-10-15
Issue static init order

wds15:
## Summary

The current initialization of the AD tape is prone to the static init order catastrophe. The problem is that the `ChainableStack::instance_` pointer is checked for being `false` in the `init()` method of `ChainableStack`. The problem is that we cannot expect that the initialization to the `nullptr` value has happened at the time when the execution reaches line 122 of the old autodiffstackstorage.hpp. This is not guaranteed by the C++ standard. So the main problem is that the `instance_` pointer serves two purposes at the same time. For one it flags if the instance is initialized and for two it points to the instance of the AD tape. However, we can just use two variables for this. As it is guranteed that a locally declared static variable is available when execution goes over it, we can us a `is_initialized` flag inside the static init method to indicate if this is the very first initialization of the instance.

## Tests

The test

`test/unit/math/rev/core/thread_stack_instance_test.cpp`

has been expanded to ensure that the two different ways of flagging "instance is not initialized" are tested. There is no test written which shows the static init order resulting in a segfault as this has never been observed in our testing pipeline with a reproducible example.

## Side Effects

None

## Checklist

- [X] Math issue #1306 


rok-cesnovar:
I restarted the tests here manually. This would be great to get in 2.21 as its a problem for some interfaces. Want me to look at it or have someone else in mind?

wds15:
Feel free to look at it.

This patch hasn't solved the issue for PyStan; I do think it avoids the static init order catastrophe though/hope. The PyStan issues seem to be different, but with the recent unit tests we did there the PyStan issues appear to be python specific (as I understand it right now, I can easily be wrong).

riddell-stan:
Any and all work on this issue is welcome. PyStan does not work with STAN_THREADS on as of 2.20.

@wds15 I thought tbb was going to change all this and avoid the need for a downstream user like PyStan instantiating ChainableStack in each thread. Or is this not correct?

wds15:
That’s correct.

stan/math/rev/core/autodiffstackstorage.hpp
```diff
@@ -119,7 +119,14 @@ struct AutodiffStackSingleton {
...
  private:
   static bool init() {
+    static STAN_THREADS_DEF bool is_initialized = false;
+    if (!is_initialized) {
+      is_initialized = true;
+      instance_ = new AutodiffStackStorage();
+      return true;
+    }
     if (!instance_) {
+      is_initialized = true;
```
rok-cesnovar: Isnt this second if redundant? Is there a case where is_initialized would be true and instance would not exist?

wds15: Yes, there is: Suppose a thread creates its AD tape instance => is_initialized is true and instance is non-null. Next, the thread destroys for whatever reason the AD tape, but the thread continues to run. Now we have is_initialized=true and instance_=nullptr. At a later stage, the same thread creates again a new tape. At this point we need the second if statement to detect that instance_ is equal to nullptr. The crux with the is_initalized variable is that it can only be modified within the init function... and the good thing is that we know that the variable does exist and can be modified.

rok-cesnovar: Ah, that makes total sense. Thanks for the explainer.

wds15:
yack... again a test failure! I checked... and it looks like a Jenkins hiccup... I restarted the upstream cmdstan failing run. Let's see.

wds15:
we merge finally?

rok-cesnovar:
Finally :) sorry this took so long.

wds15:
No worries - we are not even aware of any issues from this so far.

I had the hope that this solves PyStan problems we saw recently...but I doubt that actually. The PyStan trouble are very likely the same as the linking problems we had under Rstan. That's my hunch.

(but the code definitely improves with this to my understanding; so it's good its in)

riddell-stan:
(moving discussion from https://github.com/stan-dev/math/issues/1306 here since the descriptions are different)

@wds15 can you clarify what changes for RStan/PyStan with this change (2.22, right?) vs. 2.21? You say it's cosmetic? That is, RStan/PyStan have to take the same care with linking before and after this change?

wds15:
Yes...the linking needs to be done right like in RStan now...it’s about what is a translational unit.

PR #1404, opened 2019-10-15
Add var specialization for matrix_cl

SteveBronder:
## Summary

Adds a specialization for `matrix_cl` for `var` types.

`matrix_cl<var>` is actually two `matrix_cl<doubles>` (`adj_` and `val_` are accessible via `val()` and `adj()`). This PR has two main intentions

1. With Rok's recent PR to the new Stan compiler, OpenCL functions can be chained together instead of being passed back and forth from the CPU->GPU between function calls. atm this is only setup for data (`matrix_cl<double>`, but with a `matrix_cl<var>` specialization we can start thinking about how to move parts of the expression graph for reverse mode to fully execute over on the CPU

2. I think using a `matrix_cl<var>` is a bit prettier :-) You can see this in the OpenCL specialization of `cholesky_decompose`'s `cholesky_opencl`.

This might be better to discuss somewhere else, but I'm actually not totally sure what the body of the following signature would look like

```cpp
inline matrix_cl<var> cholesky_decompose(const matrix_cl<var>& A) {
```

Should the `matrix_cl<var>` hold a pointer to a vari so the above returns?

```cpp
matrix_cl<var>(new cholesky_all_opencl(A, LA))
```

The rest of the PR is pretty standard, the only places that's are a bit odd are

1. When we pass back data from a `matrix_cl<var>` via `from_matrix_cl(A);` we only return back an Eigen matrix comprising the adjoints. This feels like the right approach since the `val` part of a `vari` is constant, though I'm open to discussion on this one.

2. I didn't add a `triangular_transpose()` specialization for `matrix_cl<var>` because I'm not sure if we would want that. But it's trivial to add if that seems like a good idea.

## Tests

Tests exist in `opencl/rev` for `matrix_cl<var>` and it's specializations of`copy`, `sub_block`, `zeros_strict_tri()` and `zeros()`. `cholesky_decompose` also uses the `matrix_cl<rev>` so it's tests also cover this PR.

## Side Effects

This is mostly reformatting the API structure of doing reverse mode on OpenCL devices.

## Checklist

- [x] Math issue #1280


SteveBronder:
I need to, review @t4c1's PR, rubber duck this one more time then I'll ping someone for a review

stan/math/opencl/rev/triangular_transpose.hpp
```diff
@@ -0,0 +1,34 @@
+#ifndef STAN_MATH_OPENCL_REV_TRIANGULAR_TRANSPOSE_HPP
+#define STAN_MATH_OPENCL_REV_TRIANGULAR_TRANSPOSE_HPP
+#ifdef STAN_OPENCL
+
+#include <stan/math/opencl/matrix_cl_view.hpp>
+#include <stan/math/opencl/triangular_transpose.hpp>
+#include <stan/math/opencl/rev/matrix_cl.hpp>
+
+namespace stan {
+namespace math {
```
rok-cesnovar: Just delete this file. We can add later if triangular_transpose is needed.

stan/math/rev/mat/fun/cholesky_decompose.hpp
```diff
@@ -339,42 +331,32 @@ class cholesky_opencl : public vari {
...
 
-      R_adj.sub_block(L_adj, j, 0, 0, 0, k_j_ind, j);
-      D_adj.sub_block(L_adj, j, j, 0, 0, k_j_ind, k_j_ind);
-      B_adj.sub_block(L_adj, k, 0, 0, 0, m_k_ind, j);
-      C_adj.sub_block(L_adj, k, j, 0, 0, m_k_ind, k_j_ind);
-
-      C_adj = C_adj * tri_inverse(D);
-      B_adj = B_adj - C_adj * R;
-      D_adj = D_adj - transpose(C_adj) * C;
+      C.adj() = C.adj() * tri_inverse(D.val());
```
rok-cesnovar: Love this!

test/unit/math/opencl/rev/copy_test.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/** TODO(Steve): Do we want this?
+TEST(MathMatrixCL, matrix_cl_matrix_copy_arithmetic) {
+  using stan::math::var;
+  double test_val = 5;
+  // Use this for successful copy
+  stan::math::matrix_cl<var> d22_cl(1, 1);
+  EXPECT_NO_THROW(d22_cl = stan::math::to_matrix_cl(test_val));
+  EXPECT_NO_THROW(test_val = stan::math::from_matrix_cl_error_code(d22_cl));
```
rok-cesnovar: Remove this from the test.

SteveBronder:
> The only larger comment is location of the files. I feel none of these should go in /opencl/rev. My thinking is that that subfolder should be reserved for implementations that would get exposed to the Stan language. For instance:

My only thing here is that, I think that putting them in the top level folder (or inside of the other files) would start mixing prim/rev. So if we put this in the larger file we kind of have to remember that var stuff is not included when we are working with prim stuff. Placing them in rev makes it clear that these are to be used in rev and not prim. Does that make sense?

stan/math/opencl/rev/copy.hpp
```diff
@@ -0,0 +1,150 @@
...
+  matrix_cl<T> dst(src.rows(), src.cols());
+  if (src.size() == 0) {
+    return dst;
+  }
+  dst.val() = to_matrix_cl(src.val().eval());
+  dst.adj() = to_matrix_cl(src.adj().eval());
+  return dst;
+} catch (const cl::Error& e) {
+  check_opencl_error("copy var Eigen->(OpenCL)", e);
+  matrix_cl<T> dst(src.rows(), src.cols());
```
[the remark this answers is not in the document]
SteveBronder: Ah yes sorry, so this will never actually return since `check_opencl_error()` throws a terminating error. The empty var creation is just so that we don't get a compiler warning about different return types

stan/math/opencl/rev/triangular_transpose.hpp
SteveBronder: I actually just added it so that that API is the same. Pure lazyness to not add it on my part

rok-cesnovar:
> My only thing here is that, I think that putting them in the top level folder (or inside of the other files) would start mixing prim/rev.

I see your point yes. Sorry for the late reply. I am starting to buy in your `prim/opencl` and `rev/opencl` idea.

SteveBronder:
@serban-nicusor-toptal I think you are already aware of Jenkins issues but seeing some weird stuff on this PR. (I tried copy/pasting the threading error that's popping up but it came out all cooky)

https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1404/20/pipeline/146

rok-cesnovar:
Also thanks @SteveBronder for tackling this one!

PR #1409, opened 2019-10-19
Fix integrate_1d to use msgs ostream pointer like all other functions

### LINKED ISSUE #933: ostream argument in integrate_1d
#### Summary:
@bob-carpenter 

Should the ostream argument that gets passed through integrate_1d to the thing being integrated be a pointer instead of a reference?

Currently we have it as a reference, but I just did a stanc on an ode model and get the signature:
```
...      const std::vector<int>& x_i, std::ostream* pstream__) {
```

which seems to indicate it should be a pointer.

If it should be a pointer, leave the issue open and I'll fix it. If not, close it and we're good.

#### Current Version:
develop (0b0e0a4788774157fdefe77952428394faa15260)

seantalts:
## Summary

Minor refactor of integrate_1d to use std::ostream* instead of std::ostream& like all other functions. Likely an oversight from its initial implementation.

## Checklist

- [x] Math issue: Fixes #933 

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses: Columbia University

### DIFF
--- a/stan/math/prim/arr/functor/integrate_1d.hpp
+++ b/stan/math/prim/arr/functor/integrate_1d.hpp
@@ -195,7 +195,7 @@ template <typename F>
 inline double integrate_1d(
     const F& f, const double a, const double b,
     const std::vector<double>& theta, const std::vector<double>& x_r,
-    const std::vector<int>& x_i, std::ostream& msgs,
+    const std::vector<int>& x_i, std::ostream* msgs,
     const double relative_tolerance
     = std::sqrt(std::numeric_limits<double>::epsilon())) {
   static const char* function = "integrate_1d";
@@ -209,7 +209,7 @@ inline double integrate_1d(
   } else {
     return integrate(
         std::bind<double>(f, std::placeholders::_1, std::placeholders::_2,
-                          theta, x_r, x_i, &msgs),
+                          theta, x_r, x_i, msgs),
         a, b, relative_tolerance);
   }
 }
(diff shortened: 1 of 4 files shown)


seantalts:
Anyone ever seen cc1plus run out of memory on the Windows Stan unit tests? Seems pretty unrelated to the code in this PR... @serban-nicusor-toptal ?

SteveBronder:
I think I've seen that before, it's sort of sporadic. I just kicked it off again so we will see

bob-carpenter:
I think we should go the other way and refactor toward passing references.  I don't think our I/O code is robust enough to handle null pointers---there would have to be conditionals everywhere.

seantalts:
I'm confused - every other function in the Math library uses stream pointers right now, so it seems like the code must be handling it, right?

Agreed that we should refactor to references some day, but I think it's much more important be in a consistent state. This remedies that in the shortest time. Please file an issue for refactoring the Math library to references.

seantalts:
It looks like the Windows EC2 nodes are the culprit - @serban-nicusor-toptal I'm going to turn them off for now in Jenkins until they get sorted.

rok-cesnovar:
This happens from time to time. Sometime it runs out of space and then can find the sundials .a file. Restarts usually solve that.

seantalts:
`java.io.IOException: Unexpected termination of the channel` is from running out of space?

rok-cesnovar:
Could be, either that or a mismatch of Java versions which is somethin Nic mentioned some time ago. Not sure what is the status of that. But yeah, when i see that i just restart the job or part of the job and that does the trick. It doesnt happen that often, so far it wasnt that annoying. But I dont get annoyed quickly :)

seantalts:
Ah, now I need to update the old compiler...

rok-cesnovar:
Is there a reason to no just switch back to stanc3 on cmdstan develop?

Issue #1284, opened 2019-07-14
Numerical precision of normal_lcdf


alecksphillips:
## Description
The normal_lcdf(x | mu, sigma) function suffers from precision problems at x >> mu/sigma when compared to the equivalent pnorm(x, mu, sigma, log.p=T) in base R.

## Example

R code:

```
library(rstan)

stan_model_code <- "
functions {
  real test_normal_lcdf(real x) {
    return normal_lcdf(x | 0, 1);
  }
}
parameters{}
model{}"

expose_stan_functions(stanc(model_code = stan_model_code))

x <- seq(7.5,7.51,0.00001)
y_stan <- sapply(x, test_normal_lcdf)
y_R <- pnorm(x,log.p=T)

df <- data.frame(
  x = c(x,x),
  y = c(y_R, y_stan),
  method = factor(c(rep_len("R", length(x)), rep_len("Stan", length(x))))
)

plot(df$x, df$y, col = df$method, xlab = "x", ylab = "normal_lcdf(x)")
legend(x = 7.502, y = -3e-14, legend = levels(df$method), col = c(1:2), pch = 16)
```

## Expected Output:
See plot comparing Stan and R's output here: https://discourse.mc-stan.org/t/numerical-precision-of-normal-lcdf/9685

#### Current Version:
v2.19.1

bob-carpenter:
Thanks for reporting.

This is a known problem with pretty much every one of our `lcdf` and `lccdf` functions, which are implemented as `log(foo_cdf(...))` and `log1m(foo_cdf(...))`.   What's needed are iterative algorithms for the cdfs that work on the log scale with stable derivatives.  Often what happens when you get out into the tails, `log(foo_cdf(...))` remains finite, but one or more of its derivatives, which have the term `1 / foo_cdf(...)`, explode.

avehtari:
I don't think derivatives are the problem here, but computing log(1+erf(z)), using erf from boost. 

R stats package pnorm.c https://github.com/SurajGupta/r-source/blob/a28e609e72ed7c47f6ddfbb86c85279a0750f0b7/src/nmath/pnorm.c has numerically more accurate case for computing lcdf in tails with trivial derivatives. both pnorm.c and Boost erf are using the same rational Chebyshev approximation, but pnorm.c handles the cdf and lcdf far in tails with more care (see lines 199-204 and 219-260). Boost has cdf for normal but not lcdf, so it seems that the solution would be the follow the approach used in pnorm.c

[...]

PR #1411, opened 2019-10-20
Bugfix/issue #1284 numerical precision of normal lcdf

PhilClemson:
## Summary

Wrote a new piecewise numerical approximation of the normal_lcdf function and its derivatives. This is based on algorithms derived from W. J. Cody, Math. Comp. 23(107):631-638 (1969) and Abramowitx and Stegun (1964), in addition to original Taylor expansions that have been derived to bridge the gap where the numerical approximations are unstable.

## Tests

Two new valid_values tests have been added to test/prob/normal_cdf_log_test.hpp to test the function at values far from mu/sigma, which fail for the old implementation and pass in the new implementation.

I also had to alter other tests which were based on the limits of the old function (expected -infinity in regions where actual approximations of the the function are now defined). The new limits are based on the maximum values for a double. In addition, some cases tested against a seemingly arbitrary absolute error (e.g. <10^-4), rather than a relative error. This makes sense when the function converges asymptotically or when the absolute value is small. However, in the case of the derivative of the log function this does not converge but increases fairly linearly as y -> -infinity. As such it is expected that the absolute error will be much larger than 10^-4 for these values, so the tests now check for the relative error before considering the absolute error test.

## Side Effects

None that I am aware of.

## Checklist

- [x] Math issue #1284

- [x] Copyright holder: University of Liverpool


bob-carpenter:
Thanks for submitting.  The unit tests for distributions are failing and Jenkins is pointing to a .zip file.  @serban-nicusor-toptal do you know how to find that file?

serban-nicusor-toptal:
@bob-carpenter Hey, the job ran on-demand so the machine is already gone. I will try to restart this job now

PhilClemson:
> 
> 
> Would it be better to factor this into a separate `log_Phi()` function? There are several other functions that could benefit from this.
> 
>     * `normal_lccdf`
> 
>     * `skew_normal_lpdf`
> 
>     * `exp_mod_normal_lpdf`
> 
>     * `exp_mod_normal_lcdf`
> 
>     * `exp_mod_normal_lccdf`
> 
>     * `ordered_probit_lpdf`
>       I guess only `normal_lccdf` is in scope of this PR, others should have their own issues.

Yes normal_lccdf could benefit from the same approximations (assumed it was better to do one change at a time). I believe the other functions would require independent numerical approximations though.

bob-carpenter:
Any step in the right direction is appreciated.  A log_Phi function would definitely be helpful.

As far as I know, all of our log cdfs are just implemented as logs of the cdf. The derivative of log F(x) is F'(x) / F(x), so if the CDF F(x) and its derivative F'(x) are not stable, it's hard to make the log cdf stable for derivatives.

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,132 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
+    const T_partials_return x2 = square(scaled_diff);
+
+    // check whether scaled_diff^10 term will overflow
+    const bool overflow_check
+        = 10.0 * log(-scaled_diff)
+          < log(std::numeric_limits<T_partials_return>::max());
+
+    // use erfc() instead of erf() in order to retain precision
+    // since for x>0 erfc()->0
+    if (scaled_diff > 0.0) {
```
bbbales2: It looks like this function is piecewise defined over four intervals:

```
(-inf, o), (o, -20), (-20, 0), (0, inf)
```

Where ```o = pow(std::numeric_limits<T_partials_return>::max(), 0.1)```

The overflow_check logic was a bit confusing. Can this be rearranged to just be 4 flat if-else statements? Instead of three if-elses with an overflow check condition in the third one?

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,132 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
-                      / one_p_erf;
+      // compute partial derivatives
+      // based on analytic form given by:
+      // dln(CDF)/dx = exp(-x^2)/(sqrt(pi)*(1/2+erf(x)/2)
+      T_partials_return dncdf_log_dbl = 0.0;
+      T_partials_return t = 0.0, t2 = 0.0, t4 = 0.0;
+
+      // calculate using piecewise funciton
+      // (due to instability / inaccuracy in the various approximations)
+      if (scaled_diff < 0.1) {
```
bbbales2: In the same way as above would it be possible to flatten these conditionals on the piece-wise defined derivatives? It looks like they're all defined in terms of scaled_diff.

Also would it be possible to write them in order of what interval they represent using similar notation? Like it's difficult to parse sometimes scaled_diff > 1.5 sometimes scaled_diff < 2.2 or whatever. If it's necessary to write it that way it's fine just checking :D.

test/prob/normal/normal_cdf_log_test.hpp
```diff
@@ -74,8 +87,61 @@ class AgradCdfLogNormal : public AgradCdfLogTest {
...
+        typename stan::return_type<T_y, T_loc, T_scale>::type x2
+            = square(scaled_diff);
+        typename stan::return_type<T_y, T_loc, T_scale>::type pow_x = 0.0;
+        for (int j = 0; j < 6; j++) {
+          pow_x = pow(scaled_diff, -2 * j);
+          temp_p = temp_p - p[j] * pow_x;
+          temp_q = temp_q - q[j] * pow_x;
+        }
+        cdf_log += log(0.5) + log(1.0 / SQRT_PI + (temp_p / temp_q) / x2)
+                   - log(-scaled_diff) - x2;
```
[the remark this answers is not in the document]
SteveBronder: **Actually you can use LOG_HALF for log(0.5)

stan/math/prim/scal/prob/normal_lcdf.hpp
PhilClemson: Yes that makes sense. We can simply replace

`} else {`
`      if (overflow_check) {`

with 

`} else if (10.0 * log(-scaled_diff)`
          `< log(std::numeric_limits<T_partials_return>::max())) {`

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,132 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
+    const bool overflow_check
+        = 10.0 * log(-scaled_diff)
+          < log(std::numeric_limits<T_partials_return>::max());
+
+    // use erfc() instead of erf() in order to retain precision
+    // since for x>0 erfc()->0
+    if (scaled_diff > 0.0) {
+      // CDF(x) = 1/2 + 1/2erf(x) = 1 - 1/2erfc(x)
+      cdf_log += log1p(-0.5 * erfc(scaled_diff));
+      if (isnan(cdf_log)) {
```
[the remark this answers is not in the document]
PhilClemson: Ah, sorry wasn't aware!

stan/math/prim/scal/prob/normal_lcdf.hpp
PhilClemson: Yep that makes a lot more sense. I think the way to do this would be

`if (scaled_diff > 2.2) {`
`...`
`} else if (scaled_diff > 1.5) {`
`...`
`} else if (scaled_diff > 0.8) {`
`...`
`} else if (scaled_diff > 0.1) {`
`...`
`} else if (scaled_diff > -4.0) {`
`...`
`} else {`
`...`
`}`

although there will be some code duplication in the 4th and 5th statements (but I'm guessing this is ok if it improves code readability?).

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,132 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
+      T_partials_return t = 0.0, t2 = 0.0, t4 = 0.0;
+
+      // calculate using piecewise funciton
+      // (due to instability / inaccuracy in the various approximations)
+      if (scaled_diff < 0.1) {
+        // approximation derived from Abramowitx and Stegun (1964) 7.1.26
+        // use fact that erf(x)=-erf(-x)
+        // technically only true for -inf<x<0 but seems to be accurate
+        // for -inf<x<0.1
+        t = 1.0 / (1.0 - 0.3275911 * scaled_diff);
```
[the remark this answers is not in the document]
PhilClemson: Nice spot, yes they should be moved to within the if braces.

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,132 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
+        if (overflow_check) {
+          dncdf_log_dbl
+              = 2.0
+                / (SQRT_PI
+                   * (0.254829592 * t - 0.284496736 * t2 + 1.421413741 * t2 * t
+                      - 1.453152027 * t4 + 1.061405429 * t4 * t));
+          if (scaled_diff < -4.0) {
+            // need to add correction term (from cubic fit of residuals)
+            dncdf_log_dbl += 0.00024073 * x2 * scaled_diff + 0.020656 * x2
+                             + 0.10349 * scaled_diff + 0.97431;
```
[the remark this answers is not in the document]
PhilClemson: Yep this is a valid comment. I derived these constants in Matlab which automatically rounded them to 5 sf.

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,132 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
+          t4 = pow(t, 4);
+          dncdf_log_dbl = 0.1585747034 + 0.3898677543 * t + 0.3515963775 * t2
+                          - 0.09748053605 * t2 * t - 0.04347986191 * t4
+                          + 0.02182506378 * t4 * t + 0.01074751427 * pow(t, 6);
+        } else {
+          // use Taylor expansion centred around x=1.85
+          t = scaled_diff - 1.85;
+          t2 = square(t);
+          t4 = pow(t, 4);
+          dncdf_log_dbl = 0.01849212058 - 0.06876280470 * t + 0.1099906382 * t2
```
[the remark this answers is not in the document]
PhilClemson: Yes we can rename it. This is a residual from the original implementation where I had the derivative defined in a separated function called "dncdf_log".

test/prob/normal/normal_cdf_log_test.hpp
```diff
@@ -74,8 +87,61 @@ class AgradCdfLogNormal : public AgradCdfLogTest {
...
+      // CDF(x) = 1/2 - 1/2erf(-x) = 1/2erfc(-x)
+      cdf_log += log(erfc(-scaled_diff)) + LOG_HALF;
+    } else {
+      // check whether scaled_diff^10 term will overflow
+      if (10.0 * log(-scaled_diff) < log(std::numeric_limits<double>::max())) {
+        // entering territory where erfc(-x)~0
+        // need to use direct numerical approximation of cdf_log instead
+        // the following based on W. J. Cody, Math. Comp. 23(107):631-638 (1969)
+        // CDF(x) = 1/2erfc(-x)
+        typename stan::return_type<T_y, T_loc, T_scale>::type temp_p = 0.0;
```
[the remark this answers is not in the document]
PhilClemson: Okay, didn't know that! :)

test/prob/normal/normal_cdf_log_test.hpp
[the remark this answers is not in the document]
PhilClemson: You're right, I should be using LOG_HALF here. And I agree, INV_SQRT_PI would be good to add to constant.hpp.

PhilClemson:
> 
> 
> Thanks for working through all this Phil! I had a couple comments about how things are arranged in the code.
> 
> I think my big question is how do these piecewise definitions affect derivatives? Like two functions join together at zero. Do the gradients of the functions on the right and the left match? And do they match the value of the analytical gradient given to ops partials?
> 
> Also what about higher order derivatives? Do we have 2nd derivatives through all our piecewise things? 3rd order I think is the highest we'd ever likely go -- any idea on those? Feel free to ask questions back. I'd have to think a bit how the third order stuff would work.
> 
> I understand the last implementation was fragile and would blow up a lot, so it's fine that this will have limitations too (we just want this to blow up less). I just want to be clear on the limitations before I do the deep dive on the tests.

The log cdf function is pretty tricky as the only useful symmetry you can rely on is erf(x)=-erf(-x). In general I made sure the error at the joins was no more than ~10^-6.

I checked the respective functions in R and scipy and both relied on similar piecewise approximations. However, neither of these had implemented a numerical approximation of the derivatives. The higher-order derivatives could also be defined but I imagine the approximations would look even more horrible. Do we have an idea of where they might be applied?

bbbales2:
> Am I ok to do a new commit with these changes?

Yeah, definitely. How this works is after we post reviews, you patch things until you're happy with it and then poke the reviewer again (me in this case) and I'll come back and have a look.

> I checked the respective functions in R and scipy and both relied on similar piecewise approximations.

Yeah I figure there's probably a lot of nastiness under the hood in a lot of these complex functions. I suppose there's probably a lot of these things that are just hidden from us in standard libraries and stuff.

> The higher-order derivatives could also be defined but I imagine the approximations would look even more horrible.

Since this is a one argument function, let's pull in Bob's autodiff tester and see what it tells us for the derivatives. Instructions here: https://github.com/stan-dev/math/wiki/Automatic-Differentiation-Testing-Framework, examples here: https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/scal/fun/bessel_first_kind_test.cpp

It should be pretty easy to add tests with this. Do you mind just playing around and seeing what you find? Like run the tests in the tails, at the joining points in values, at the joining points in derivatives, etc. I think if we find problems out in the super-far tails it's fine. We just want to convince ourselves things worked at least where they worked before (and hopefully farther out too).

Ask if it's not clear how to use that tester. It's some fancy C++.

bbbales2:
Bob's autodiff test thing basically assumes we're set up good prim tests to us (evaluating precision and checking edge cases). Once we have those it tries to automate all the gradients. Ofc. in this situation the finite differences and stuff might just break down before the new numerics but we can at least try :D.

bob-carpenter:
> ... The higher-order derivatives could also be defined but I imagine the approximations would look even more horrible. Do we have an idea of where they might be applied?

The gradients in reverse-mode get used in all of our current algorithms (HMC, L-BFGS, and ADVI).  We want to use higher-order autodiff to calculate Hessians for higher-order algorithms and for Laplace approximations and as nested components of our ODE and other solvers that need Jacobians at various points which will be more efficient in forward mode.

test/prob/normal/normal_cdf_log_test.hpp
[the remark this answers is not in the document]
SteveBronder: No worries! I'm working on a PR right now that should make these sorts of things easier to find on the math docs site

PhilClemson:
> It should be pretty easy to add tests with this. Do you mind just playing around and seeing what you find? Like run the tests in the tails, at the joining points in values, at the joining points in derivatives, etc. I think if we find problems out in the super-far tails it's fine. We just want to convince ourselves things worked at least where they worked before (and hopefully farther out too).

Sure I'll definitely have a go at this. Completely understand the need to check these things.

> The gradients in reverse-mode get used in all of our current algorithms (HMC, L-BFGS, and ADVI).  We want to use higher-order autodiff to calculate Hessians for higher-order algorithms and for Laplace approximations and as nested components of our ODE and other solvers that need Jacobians at various points which will be more efficient in forward mode.

Ah okay, just wanted to be sure there was a useful application before we delve down the rabbit hole!

PhilClemson:
I've added in all of the changes in the latest commit. After playing around with the autodiff tester I found some regions where it was failing some of the tests, so I added some new approximations (Taylor expansions and fits of the residuals) to improve the accuracy.

The only remaining issue is happening right on the boundaries in the piecewise function. It passes all but one of the autodiff tests, here's an example of the output:

`[----------] 1 test from mathMixScalFun`
`[ RUN      ] mathMixScalFun.lcdf_derivatives`
`./test/unit/math/expect_near_rel.hpp:50: Failure`
`The difference between 0 and relative_diff is 2, which exceeds tol, where`
`0 evaluates to 0,`
`relative_diff evaluates to -2, and`
`tol evaluates to 0.01.`
`expect_near_rel_finite(-1.8418968633584807, 0.16197045242952307, tolerance = 0.01); relative diff = -2`
`    in: expect_near_rel; require items x1(i) = x2(i): grad hessian grad_H_fd[i] == grad_H_ad[i]`
`[  FAILED  ] mathMixScalFun.lcdf_derivatives (1 ms)`
`[----------] 1 test from mathMixScalFun (1 ms total)`

I get `relative diff = -2` in each case, but I'm not sure what it means?

bbbales2:
Thanks for doing this! Did you add the mix tests to the branch? I only see test code for rev and prim.

Let me know when you have all the the tests up and I'll go about checking things and see what I can make of it. @bob-carpenter without looking at the code, does that output above (https://github.com/stan-dev/math/pull/1411#issuecomment-553863352) indicate any specific sort of failure to you?

PhilClemson:
> 
> Thanks for doing this! Did you add the mix tests to the branch? I only see test code for rev and prim.

No sorry, I'd added the tests to the prim test file. I've now included the autodiff tests in a new mix test file.

bbbales2:
```
./runTests.py test/unit/math/mix/scal/prob/normal_cdf_log_test
```

Runs for me (and passes). What's an example of one of the tests that was failing?

PhilClemson:
I left out the tests it was failing on to avoid confusion. It will fail when using the unscaled value of one of the borders in the function for the derivative. For example:

`stan::test::expect_ad(f(0.0, 1.0), -0.1 * stan::math::SQRT_2);`

bob-carpenter:
I doubt it was passing those tests before.

Usually that means the finite diff is failing.   When the difference between numbers is big, the relative difference approaches 2. 
   
   rel_diff(a, b) = abs(a - b) / (0.5 * (abs(a) + abs(b)))

Suppose without loss of generality that a >> b >> 0.  
Then it's easy to see the numerator goes to a and the denominator
to 0.5 * a and the rel diff to 2.

Given that this is happening at third-order derivatives, I'd
suggest just ignoring it and setting tolerances this way:

stan::test::ad_tolerances tols;
tols.grad_hessian_grad_hessian_ = 1e1;
...
expect_ad(tols, ...);

bbbales2:
Cool. Thanks Bob. @PhilClemson I'll dig through this tomorrow and get back to you.

test/unit/math/mix/scal/prob/normal_cdf_log_test.cpp
```diff
@@ -0,0 +1,18 @@
...
+  stan::test::expect_ad(f(0.0, 1.0), -50.0);
+  stan::test::expect_ad(f(0.0, 1.0), -20.0 * stan::math::SQRT_2);
+  stan::test::expect_ad(f(0.0, 1.0), -5.5);
+  stan::test::expect_ad(f(0.0, 1.0), 0.0);
+  stan::test::expect_ad(f(0.0, 1.0), 0.15);
+  stan::test::expect_ad(f(0.0, 1.0), 1.14);
+  stan::test::expect_ad(f(0.0, 1.0), 3.00);
+  stan::test::expect_ad(f(0.0, 1.0), 10.00);
+  stan::test::expect_ad(f(-1.0, 2.0), 1.50);
+  stan::test::expect_ad(f(2.0, 1.0), 0.50);
```
bbbales2: Can you add the failing test with Bob's tolerance stuff in here:

```
stan::test::ad_tolerances tols;
tols.grad_hessian_grad_hessian_ = 1e1;

stan::test::expect_ad(tols, f(0.0, 1.0), 0.1 * stan::math::SQRT_2);
```

There are a fair number of things that fail with third order autodiff and that's okay. Maybe we come back and fix them sometime, but having these tests around I think will be good for bookkeeping (if we need to find these functions we can grep for them).

test/prob/test_fixture_cdf_log.hpp
```diff
@@ -354,11 +354,15 @@ class AgradCdfLogTestFixture : public ::testing::Test {
         << "Number of finite diff gradients and calculated gradients must "
            "match -- error in test fixture";
     for (size_t i = 0; i < finite_dif.size(); i++) {
-      EXPECT_NEAR(finite_dif[i], gradients[i], 1e-4)
-          << "Comparison of finite diff to calculated gradient failed for i="
-          << i << ": " << parameters << std::endl
-          << "  finite diffs: " << finite_dif << std::endl
-          << "  grads:        " << gradients;
+      // use relative error check in first case as logs of functions do not
```
bbbales2: I understand the reasoning for doing relative error checks away from zero and absolute error checks near zero. I'm not familiar with this log asymptotic thing though. Is it related?

Also I get our tolerances are bad here, but 10% off for the relative error seems high? Was that just an arbitrary choice or does it need to stay like that?

I'm not super concerned about these tests, but this changes touches other code yada yada so gotta be careful. I trust Bob's autodiff stuff and I'm pretty sure it's covering these tests (didn't realize this finite difference stuff was being checked already!). I'll ask Bob about this separately in this thread.

test/unit/math/rev/scal/prob/normal_cdf_log_test.cpp
```diff
@@ -243,33 +243,44 @@ void test_value_and_derivatives(double expected_val, double y_dbl,
...
   EXPECT_FALSE(is_nan(gradients[1]));
   EXPECT_FALSE(is_nan(gradients[2]));
+  // use relative error check in first case as logs of functions do not
+  // necessarily converge to asymptotic values
   if (!is_nan(gradients[0])) {
-    if (!is_nan(finite_diffs[0]))
-      EXPECT_NEAR(finite_diffs[0], gradients[0], 1e-2);
-    else
+    if (!is_nan(finite_diffs[0])) {
+      if (abs((finite_diffs[0] - gradients[0]) / finite_diffs[0]) > 0.1) {
```
bbbales2: I guess these are the same sorts of tests as above. Just dropping a comment here so I remember them.

test/prob/test_fixture_cdf_log.hpp
bbbales2: I'm familiar with how relative differences near zero fail and we switch to absolute errors, but I don't follow the log of functions thing here. Is it similar? Is this something special with gradients?

Also 10% seems like a really coarse error? I know things in this test framework are pretty rough but 0.1 seems really large.

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -53,34 +55,157 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
...
+        t4 = pow(t, 4);
+        dncdf_log = 0.6245634904 - 0.9521866949 * t + 0.3986215682 * t2
+                    + 0.04700850676 * t2 * t - 0.03478651979 * t4
+                    - 0.01772675404 * t4 * t + 0.0006577254811 * pow(t, 6);
+      } else if (10.0 * log(fabs(scaled_diff))
+                 < log(std::numeric_limits<T_partials_return>::max())) {
+        // approximation derived from Abramowitx and Stegun (1964) 7.1.26
+        // use fact that erf(x)=-erf(-x)
+        // technically only true for -inf<x<0 but seems to be accurate
+        // for -inf<x<0.1
```
bbbales2: If it's technically only true for -inf<x<0 then why does it work for for -inf<x<0.1? Certainly tests pass in that region. Is it just that Abramowitz and Stegun only gave the expansion for -inf<x<0? Maybe reword

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -44,7 +46,7 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
   scalar_seq_view<T_scale> sigma_vec(sigma);
   size_t N = max_size(y, mu, sigma);
 
-  const double SQRT_TWO_OVER_PI = std::sqrt(2.0 / pi());
+  const T_partials_return SQRT_TWO_OVER_PI = std::sqrt(2.0 / pi());
```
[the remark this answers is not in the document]
PhilClemson: Yes, you're right

test/unit/math/mix/scal/prob/normal_cdf_log_test.cpp
PhilClemson: No problem. I've added this

test/prob/test_fixture_cdf_log.hpp
PhilClemson: The asymptotic thing is just from the fact that as a function asymptotically converges towards 0, the log of that function goes to -inf. When you've got a function that converges asymptotically then the absolue error is (hopefully) going to shrink the closer it gets to that value. For the other functions it's fine to use a single absolue error check over the whole function, but for log(x) functions you have to switch to a relative error check as x->0.

10% was a completely arbitrary choice. I've changed it ot 1% now (just checked and it's still passing all the tests).

stan/math/prim/scal/prob/normal_lcdf.hpp
PhilClemson: Yes that's right, I've reworded it

test/unit/math/rev/scal/prob/normal_cdf_log_test.cpp
PhilClemson: I've changed this to a 1% relative error check as well

test/prob/test_fixture_cdf_log.hpp
PhilClemson: > Oh yeah is there a particular thing you need a CDF to be super robust for? Like is this for truncated distributions? Or for a link function? Might be worth throwing a comment in there saying something like these expansions look complicated but they're here because of X.

Yes the reason we came across this issue is because we were investigating the case of rare base-rate logistic regression problems in stan. We're proposing an alternative link function for dealing with these problems, which is able to more closely fit the analytic posteriors.

> @PhilClemson question -- @nhuurre pointed out this code might translate to other functions. This isn't something that needs done in this pull req, but do any of these expansions generalize? Clearly it's a lot of work to get them right and so it'd be nice if it could be boxed up at all.

Other than normal lccdf (which could be implemented as a simple y-axis reflection of normal lcdf) I'm not sure what else would be generalizable for use in other functions. The gradient approximations are definitely very specific and that's where most of the work has gone in.

stan/math/prim/scal/prob/normal_lcdf.hpp
```diff
@@ -162,8 +162,8 @@ inline return_type_t<T_y, T_loc, T_scale> normal_lcdf(const T_y& y,
                  < log(std::numeric_limits<T_partials_return>::max())) {
         // approximation derived from Abramowitx and Stegun (1964) 7.1.26
         // use fact that erf(x)=-erf(-x)
-        // technically only true for -inf<x<0 but seems to be accurate
-        // for -inf<x<0.1
+        // Abramowitx and Stegun define this for -inf<x<0 but seems to be
```
mcol: Should be Abramowitz. :)

bob-carpenter:
> why are there separate implementations of every CDF in the tests (cdf_log vs. cdf_log_function)

No idea.  Maybe to have a baseline against which to test.

> does your autodiff testing stuff replace the distribution autodiff tests

No.

test/prob/test_fixture_cdf_log.hpp
nhuurre: `ordered_probit_lpmf` uses `log(Phi(...))` directly so it can borrow this without any problem. 
It's more complicated for `skew_normal` and `exp_mod_normal`. Naive substitution where they use `log(erfc(...))` would result in catastrophic cancellation in the derivatives (in a subset of the region where they currently fail anyway...)

bbbales2: @nhuurre are there problems where order_probit suffers from numeric breakdowns like happened here? If so I guess we make a separate issue out of this.

test/prob/test_fixture_cdf_log.hpp
```diff
@@ -354,11 +354,15 @@ class AgradCdfLogTestFixture : public ::testing::Test {
...
+      // necessarily converge to asymptotic values
+      if (abs((finite_dif[i] - gradients[i]) / finite_dif[i]) > 0.01) {
```
bbbales2: Just delete this extra code and the test that triggers it (the y = -50 example). I don't wanna worry about changing this test framework.

Weird to say delete a test, but lemme just justify that by saying:

I had a closer look today and I agree this totally looks like a shortcoming in the distribution test framework, not the code here.

Also, you have the test in test/unit/math/mix/scal/prob/normal_cdf_log_test.cpp that covers this exact case with Bob's autodiff stuff.

PhilClemson: Okay I've reverted back to the original file. I guess the relative error checks in the rev test should at least keep a record of the issue in the negative tail. Let me know if you need anything else!

bbbales2:
In we go! Thanks Phil!

bob-carpenter:
> does your autodiff testing stuff replace the distribution autodiff tests 

no.

> why are there separate implementations of every CDF in the tests (cdf_log vs. cdf_log_function)

I think one is used as a reference implementation to test the other, but I'm not 100% sure.  It prettymuch has to be that, though, if it's being used and defined only in the tests.

Issue #1420, opened 2019-10-25
Optimize matrix_cl constructors


t4c1:
## Description
When constructing a matrix_cl from `Eigen::Map` it creates a copy of the data by calling `.eval()`. Instead of this the data in Map should be directly accessed. 

When constructing a `matrix_cl` from a rvalue (temporary object), it uses blocking calls to OpenCL to avoid accessing these objects after they are destroyed. Instead these objects could be moved (or evaluated in case of Eigen expressions) on heap and deleted after the copy to OpenCL device is complete.

## Example
```
int rows, cols;
double* data;
Eigen::Map<Eigen::MatrixXd> map(data, rows, cols);
matrix_cl(map);
```

```
Eigen::MatrixXd a, b;
matrix_cl(a + b);
```

## Expected Output
Constructing `matrix_cl` from a Map is faster. Constructing `matrix_cl` from a temporary does not block.

#### Current Version:
v3.0.0

PR #1421, opened 2019-10-25
Optimize matrix_cl constructors

t4c1:
## Summary

Added separate constructor to `matrix_cl` that accepts Eigen Map. 

When constructing `matrix_cl` from rvalue it is moved to heap and deleted after the copy is done.

To achieve that some trait metaprograms were added to `is_eigen.hpp`.

## Tests

None. These are just optimizations that do not change interface.

## Side Effects

None.

## Checklist

- [x] Math issue #1420 


stan/math/prim/mat/meta/is_eigen.hpp
```diff
@@ -43,11 +44,37 @@ template <typename T, int R, int C>
...
+struct is_eigen_array : internal::is_eigen_array_impl<std::decay_t<T>> {};
+
+template <typename T>
+using is_eigen_matrix_or_array
+    = math::disjunction<is_eigen_matrix<T>, is_eigen_array<T>>;
+
+namespace internal {
+template <typename T>
+struct is_eigen_contiguous_map_impl : std::false_type {};
+template <typename T, int Opts>
```
SteveBronder: Could this just be variadic?

SteveBronder: We should probs do that for all of these that can take in multiple values

stan/math/opencl/matrix_cl.hpp
```diff
@@ -294,12 +294,48 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+      return;
+    }
+    if (is_eigen_matrix_or_array<Mat>::value
+        && std::is_lvalue_reference<Mat>::value) {
+      initialize_buffer(A.eval().data());
+    } else {
+      auto* A_heap = new Mat_type(std::move(A.eval()));
+      try {
+        cl::Event e = initialize_buffer(A_heap->data());
+        e.setCallback(CL_COMPLETE, &delete_it<Mat_type>, A_heap);
```
SteveBronder: I'm not following here, doesn't the initialization already use the heap? Why do we need new and delete here (i.e. explicit mem management over RAII)? We very much try not to do explicit new and deletes in the code so unless there's a big performance gain here (you can use Sean's perf-math to test that) I'd like to just use basic RAII

https://github.com/seantalts/perf-math

syclik: We also have to be very, very careful about managing resources if we have news floating around. Any possible early exit condition now is a potential leak. 

I was tracing through some of this code; I’d like to make sure it’s tested in exceptional cases so we don’t have leaks.

SteveBronder: Also though, I think we can use this callback scheme in the error checks so we don't have to stop there!

stan/math/prim/mat/meta/is_eigen.hpp
t4c1: I think I don't understand the question. Can you give an example?

stan/math/opencl/matrix_cl.hpp
t4c1: I don't think check* functions should add callbacks just because we have one use case that needs them.

I did not use RAII here because lifetime of the allocated memory is not bound to any object or srtack frame. So using RAII here is not possible. Here I am not allocating memory for the data, just metadata (Matrix object). Actual content of the matrix remains where it is - just the matrix is moved to heap so that it is not destroxed when we return.

I agree we need to be very careful about directley managing memory. I guess you, as reviewer must make sure the code is understandable and correct. In this case I set the callback to delete memory immediately after allocation. If any thing goes wrong and exception is throwed it is deleted in catch body.

I don't think we can test for memory leaks. So understanding the code and making sure it is correct is even more important.

syclik: I agree that we need to understand how to make this correct, but it’s not the role of the reviewer to figure all of it out. Since this is a change in the paradigm, the submitter really needs to help explain what is happening, why, and what the side effects are. (That’s part of the submission process and documentation of the PR. I’m happy to dig in as the reviewer, but as the submitter, you’ve got to justify this as a maintainable change, which I think it is.)

Just to avoid a situation like other PRs where I’ve tried to help submitters with their PR, I’m inclined to close this PR because it’s actually missing doc around the side effects of this PR. Since this introduces something where the client has to manage memory, it would make sense that the side effect is listed and you’ve outlined under what conditions this has to be cleaned up by the user. 

If you can’t test for leaks, maybe there’s a different design pattern that will allow for testing? Do you want to have a design review? Given the proposed change, we have to comb through all error conditions that could pop up and make sure it’s handled properly. Have you done that? If so, can you highlight how you’ve accomplished that?

(Sorry for so many questions.... these are all things you can help any reviewer with.)

syclik: This logic is actually pretty confusing. If any of those steps fail, it deletes memory? Why shouldn’t it be on the one call where it transfers the ownership of the memory? If it fails on the second call, why would this function still own the memory?

syclik: This logic is actually pretty confusing. If any of those steps fail, it deletes memory? Why shouldn’t it be on the one call where it transfers the ownership of the memory? If it fails on the second call, why would this function still own the memory?

syclik: I'm resolving this conversation. I had to dig further, but I think the logic here isn't quite clear / right. I'll make a new comment.

stan/math/opencl/matrix_cl.hpp
```diff
@@ -294,12 +294,48 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+    if (size() == 0) {
+      return;
+    }
+    if (is_eigen_matrix_or_array<Mat>::value
+        && std::is_lvalue_reference<Mat>::value) {
+      initialize_buffer(A.eval().data());
+    } else {
+      auto* A_heap = new Mat_type(std::move(A.eval()));
+      try {
+        cl::Event e = initialize_buffer(A_heap->data());
```
syclik: So first off, this is a side-effect. `initialize_buffer` now returns a `cl::Event` and does not return `void`. Please document this in the PR and help explain the design.

This is definitely leak-able. You've handled L312 appropriately: the `initialize_buffer()` call can throw. There's a catch which deletes the memory.

L313 can definitely leak. If the [sdk documentation](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clSetEventCallback.html) for `clSetEventCallback()` is correct, then we must check the error code returned to determine if there has been an error. If any of the errors described in that doc happens, then `e` is still constructed how it's constructed, but the callback is not registered and this memory will never be freed. I see that you've repeated this pattern multiple times, so it'll have to be fixed in each place.

t4c1: > So first off, this is a side-effect.

Can you explain what do you mean with that?

> initialize_buffer now returns a cl::Event and does not return void. Please document this in the PR and help explain the design.

`initialize_buffer` return value is documented in doxygen. Or did you mean I should more deeply explain here? We now in some cases (the one I am changing) need the event for the copy outside of the `initialize_buffer` function. So the return. Any other use can still ignore the return as if ti was void.

I am not calling `clSetEventCallback()` directely. We use c++ bindings which support throwing exceptions on error conditions if `CL_HPP_ENABLE_EXCEPTIONS` macro is defined. And we do define it. So `e.setCallback()` throws if `clSetEventCallback()` would return error value. So I think there are no bugs.

syclik: > Can you explain what do you mean with that?

Sorry about being imprecise with terminology. I was actually mixing two meanings of the term "side effect;" the first is what's used commonly in [computer science (link to wikipedia)](https://en.wikipedia.org/wiki/Side_effect_(computer_science)). This change really modifies states outside of `matrix_cl`, so it is a side effect and it should really be documented at the PR level (since this is the PR that introduces it) and also in the class itself. The existence of side effects aren't necessarily bad or good, but documenting them is something we must do. It's just a precondition for having any PR merged.

The second use of side effect was that this changes the API. It's interesting because it used to be listed in the same wikipedia article as "In computer science, a function or expression is said to have a side effect if it modifies some state or has an observable interaction with calling functions or the outside world," but was changed recently. It's still [quoted on StackOverflow](https://softwareengineering.stackexchange.com/a/325520).

> `initialize_buffer` return value is documented in doxygen. Or did you mean I should more deeply explain here? We now in some cases (the one I am changing) need the event for the copy outside of the `initialize_buffer` function. So the return. Any other use can still ignore the return as if ti was void.

Yes, now that you're introducing branching logic, this has to be documented for `initialize_buffer` -- for more concreteness, please leave enough documentation for a different person to be able to understand what cases there are, under what conditions they are responsible for dealing with things, and under what conditions they are not supposed to do things. 

> I am not calling clSetEventCallback() directely. We use c++ bindings which support throwing exceptions on error conditions if CL_HPP_ENABLE_EXCEPTIONS macro is defined. And we do define it. So e.setCallback() throws if clSetEventCallback() would return error value. So I think there are no bugs.

Thanks for the explanation! First, I'll start with questions regarding the implementation and leaking conditions:
1. Is the callback guaranteed to be called on normal exit conditions? If so, mind pointing out the documentation that explicitly writes this out?
2. Is the callback guaranteed to be called on abnormal exit conditions? If not, what happens? Can you point out the documentation that states this?
3. Is the callback guaranteed to be called exactly once? More specifically, are there any conditions under which this could be called two or more times? In particular, is any of this done in parallel where there could be a race condition where the callback could be triggered twice due to asynchronous processes trying to figure out ending conditions?
4. Is it possible for the event to continue to be used after the callback?
5. When is the event supposed to be cleaned up? (Since it's outside of this function, just a quick description would help.)

Since you're changing this, I think we should talk about design; this PR skipped that phase and went directly into implementation. @SteveBronder brought up a good point about RAII and I think it's appropriate. With this sort of memory-management, it would make sense to encapsulate this logic. It could even be done using some sort of factory function. (There's no reason `matrix_cl` needs to allocate memory and conditionally delete it... we could have a factory function that creates `cl::Event` that either allocates memory and adds the callback or throws and leaves the world in a recoverable place.)

SteveBronder: Dan I think before we go into the above the base thing I am confused about is what it's solving. Is this to handle the async data movement? If that's the case, can you write a test to show what behavior this fixes?

t4c1: So are side effect in the PR template meant as a changes that are not required to fix an issue or any side effects of functions a PR touches?

I allways assumed it was the first case. If it is the second one I think PR template should be changed to "Side effects in functions touched by this PR" or something similar. 

Also starting any async kernel launch or data transfer is essentially a side effect of a function. I think it was never mentioned as a side effect of a PR that introduced it. So I guess the first case is the right one?

> Yes, now that you're introducing branching logic, this has to be documented for initialize_buffer

That is not correct. I am adding no new branches to `initialize_buffer` in this PR.

>Thanks for the explanation! First, I'll start with questions regarding the implementation and leaking conditions:
Is the callback guaranteed to be called on normal exit conditions? If so, mind pointing out the documentation that explicitly writes this out?
Is the callback guaranteed to be called on abnormal exit conditions? If not, what happens? Can you point out the documentation that states this?

Questions 1 and 2 are irrelevant for this PR - all allocated memory is freed anyway on program exit regardless of whether it is normal or abnormal.

>Is the callback guaranteed to be called exactly once? More specifically, are there any conditions under which this could be called two or more times? In particular, is any of this done in parallel where there could be a race condition where the callback could be triggered twice due to asynchronous processes trying to figure out ending conditions?

Yes it will always be called exactley once. OpenCL API is thread-safe.

> Is it possible for the event to continue to be used after the callback?

Yes. Callback does not affect the event in any way.

> When is the event supposed to be cleaned up?

`cl::Event` object is within the function `initialize_buffer()`. So it is destroyed when the function exits. The event this object contains handle to is a part of OpenCL implementation. User of OpenCL is not required to manage memory of OpenCL implementation's internal objects.

t4c1: >I am confused about is what it's solving. Is this to handle the async data movement?

Yes. With your async PR #1162 we started using asynchronous versions of all OpenCL calls. Than in PR #1356 the bug with constructing `matrix_cl` from rvalues was fixed in simplest possible way. Which meant that when `matrix_cl` is constructed from rvalue blocking call is used. Here I am reverting that change from PR #1356 meaning that async calls can also be used in these cases.

> If that's the case, can you write a test to show what behavior this fixes?

No. Even if we use async calls OpenCL implementation is not actually required to copy the data asynchronously. Even if it was, I think we could not reliably test that function returns before copying is complete. We also don't have any tests that check whether other OpenCL calls are actually done asynchronously.

SteveBronder: > > I am confused about is what it's solving. Is this to handle the async data movement?

> Yes. 

Can you change the name of this PR then? It sounds like this is a bugfix and not an optimization. It may be good to explain some of the idea around the bugfix as well in the top level comment

> With your async PR #1162 we started using asynchronous versions of all OpenCL calls. Than in PR #1356 the bug with constructing matrix_cl from rvalues was fixed in simplest possible way. Which meant that when matrix_cl is constructed from rvalue blocking call is used. Here I am reverting that change from PR #1356 meaning that async calls can also be used in these cases.

>> If that's the case, can you write a test to show what behavior this fixes?

> No. Even if we use async calls OpenCL implementation is not actually required to copy the data asynchronously. Even if it was, I think we could not reliably test that function returns before copying is complete. We also don't have any tests that check whether other OpenCL calls are actually done asynchronously.

I'm very wary about fixing bugs we can't replicate on at least a particular setup. If we need a particular environment for a specific bug we may be able to make a dockerfile to replicate that environment. The platform and device both have to have async support, we can test whether we have that and if not then we can probably make an ec2 setup that has it. i.e.  doing the test on a new intel cpu with their platform since they support async.

Is there a reason something like below wouldn't work

```cpp
TEST(yadayada, blah) {
stan::math::matrix_cl;
for (i in 1:ALot) {
// do a bunch of opencl moving around and ops to build up the async stack
}

matrix_cl foo(Eigen::BigIdentityMatrix(BigN) + 1);
int check_val = func_to_flag_if_foo_not_right(foo);
EXPECT_TRUE({check_val == 1})

```


>> Is the callback guaranteed to be called on normal exit conditions? If so, mind pointing out the documentation that explicitly writes this out?
Is the callback guaranteed to be called on abnormal exit conditions? If not, what happens? Can you point out the documentation that states this?

> Questions 1 and 2 are irrelevant for this PR - all allocated memory is freed anyway on program exit regardless of whether it is normal or abnormal.

They are relevant. Page 179 of the [OpenCL Spec](https://www.khronos.org/registry/OpenCL/specs/opencl-1.2.pdf) gives the end conditions of an event. The setup you have here

```cpp
        e.setCallback(CL_COMPLETE, &delete_it<U_val>, obj_heap);
      } catch (...) {
        delete obj_heap;
        throw;
      }
```

Will delete `obj_heap` when the callback is complete or throws when inside of this functions scope, but it won't delete `obj_heap` if the event returns an error code when the program has exited this scope. Though I'm pretty sure since we set OpenCL up so that it terminates the program when there is a hardware error we would just clean up the whole program. That's a side effect though directly from the code in this PR. I've thought about turning those failures into domain errors instead of terminal conditions. For example when we start having multiple chains in a stan program it very well may happen that the chains send too much stuff over sometimes and we run out of gpu ram. But that's not guaranteed to fail on every iteration so we may just want to throw a `domain_error` for that and let the program keep running. 

For Q 2 this paragraph seems to imply that abnormal exit conditions in parts of the program before a call is finished is implementation defined

> If the execution of a command is terminated, the command-queue associated with this
terminated command, and the associated context (and all other command-queues in this context) may no longer be available. The behavior of OpenCL API calls that use this context (and command-queues associated with this context) are now considered to be implementation defined. The user registered callback function specified when context is created can be used to report appropriate error information.

There's also some stuff on 186 about how expensive ops can also lead to undefined behavior. I actually have no idea if deletion of a large object is considered expensive or not in the OpenCL Spec's idea of 'expensive'.

> So are side effect in the PR template meant as a changes that are not required to fix an issue or any side effects of functions a PR touches?
> I always assumed it was the first case. If it is the second one I think PR template should be changed to "Side effects in functions touched by this PR" or something similar.

It would be nice to clear up the PR templates 'side effects' section. I never know what to put there. I certainly hope my PRs don't have unintended side effects! Should it be 'consequences' for other parts of the stan math library or something like that?

I'm not sure if "side effect" is the right word for that part of the PR template, at least it has a lot of definitions. Though this is probs a conversation for somewhere else.

t4c1: > It sounds like this is a bugfix and not an optimization.

I am not sure if we call something potentially working slower a bug. Whatever, it is a matter of definition, so I don't mind changing title if you are sure it should be done.

> Is there a reason something like below wouldn't work

I think it will work on some setups, but definitely not all of them.

> They are relevant.

Oh, we are talking about kernel execution status? I understood exit conditions as what is happening at program exit. The doc [here](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clSetEventCallback.html) states: "The callback function registered for a command_exec_callback_type value of CL_COMPLETE will be called when the command has completed successfully or is abnormally terminated. " So the answer to both questions is yes, it is guaranteed.

> For Q 2 this paragraph seems to imply that abnormal exit conditions in parts of the program before a call is finished is implementation defined

> the behavior of OpenCL API calls that use this context (and command-queues associated with this context) are now considered to be implementation defined. The user registered callback function specified when context is created can be used to report appropriate error information.

It only says API calls that reference context are undefined. What we use is exactley what last sentence mentiones - a callback that was registered when the context was still valid to get some info on error.

> There's also some stuff on 186 about how expensive ops can also lead to undefined behavior. I actually have no idea if deletion of a large object is considered expensive or not in the OpenCL Spec's idea of 'expensive'.

That part seems a bit weird. I dont understand how a function that takes more than certain amount of time could break anything. Anyway I found [another OpenCL function](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clSetMemObjectDestructorCallback.html) with the same warning in doc that also states that its typical use is freeing the memory in the callback. So I guess this should be fine.

syclik: @t4c1: thank you for the responses! Especially providing the relevant parts of the API spec as links. For future reference, it would help if you, as the submitter, provided those references for 3rd party libraries. It’s hard to expect a reviewer to know any of the libraries well enough to know the edge cases. Since you’ve linked them in the discussion, can you update the doxygen documentation with this info? It should have been there with the PR as part of what is necessary to keep this maintainable. Thank you for pointing out the documentation. I now understand why you coded it this way and believe that the callback should be called once and only once and at the end. 

Re: branching. I think you’re misunderstanding the scope. This PR clearly introduces branching. Starting with L306:
```
	    if (is_eigen_matrix_or_array<Mat>::value
        && std::is_lvalue_reference<Mat>::value) {
      initialize_buffer(A.eval().data());
    } else {
      auto* A_heap = new Mat_type(std::move(A.eval()));
      try {
        cl::Event e = initialize_buffer(A_heap->data());
...
```

Prior to this PR, there was no if/else.

Re: expensive ops. This is where we have to code defensively. We should be coding defensively all the time, but APIs change. If we’re relying on certain behavior in order for our functions to work, we should have tests that check that condition (not necessarily as part of the code, but it could be a unit test that fails).

Question: is there any way that an event could fail to trigger the callback, but have this continue?

Also, why isn’t the new function `initialize_buffer_optionally_from_heap` used? It seems like this is the behavior I was describing. There’s no need for this constructor to call `new` and `delete` on its own. 

Also, have you considered using the memory arena that’s part of the autodiff stack? If it’s there, it won’t leak because the memory locations aren’t allocated when called and it is the burden of the client to recover that memory every time. 

@SteveBronder: thank you! The review comments you’re making are spot on, especially asking about the high-level context of the PR. 

Re: PR template and “side effect.” It was meant as everything. It was really meant for the submitter to write what they thought were side effects, either in code, changes to the api, changes to functions, changes to design. It’s really hard to be rigid about this because any time we do anything interesting, it’ll fall outside what we write. And if we have narrowly scoped things, most of the time it won’t be used. But.... if what’s written is not useful, then we should update it. Ideally, any new side effects in functions would be documented in the doxygen doc and then highlighted in the PR so the reviewer is told what side effects there are. That makes it easy for the reviewer. But that’s not the only way it could be done. Anyway, if you can make that clearer, please do.

t4c1: > It’s hard to expect a reviewer to know any of the libraries well enough to know the edge cases.

Well its only natural that there are questions and possibly some misunderstandings when looking at some code for the first time. One of the reasons for reviewing is that those are addressed. BTW next time you need some documentation for OpenCL you can google "OpenCL [function name]" and the first result should be the doc.

> Yes, now that you're introducing branching logic, this has to be documented for initialize_buffer
> I think you’re misunderstanding the scope. 
That seems to be the case. Since you mention `initialize_buffer()` in the same sentence I naturally assumed you are talking about branching in this function. 

> Re: expensive ops. This is where we have to code defensively. We should be coding defensively all the time, but APIs change. If we’re relying on certain behavior in order for our functions to work, we should have tests that check that condition (not necessarily as part of the code, but it could be a unit test that fails).

If something crashes or deadlocks any test will detect it. If undefined behavior is something else we may or may not detect it. But as I explained in previous post I am pretty sure we are not actually triggering any undefined behavior.

> Question: is there any way that an event could fail to trigger the callback, but have this continue?

I think the only way for this to happen is that program exits before copying is completed.

> Also, why isn’t the new function `initialize_buffer_optionally_from_heap` used?

It does not do the same thing! `initialize_buffer_optionally_from_heap()` copies an object to heap. Which would be fine for matrices, but not eigen expressions. These need to be evaluated before they can be copied to GPU. In other words the difference is in `.eval()` (which is a no-op for matrices).

> Also, have you considered using the memory arena that’s part of the autodiff stack?

No. It in an interestiong idea, but I have some concerns:
1. Do wa have a way to guarantee that memory is not released before copying to GPU is complete?
2. As far as I understand it one of advantages of using arena is memory locality for autodiff variables that are often used together. Adding matrices that are only used once in copying would decrease that locality.

I am moving the discussion about side effects to discourse: https://discourse.mc-stan.org/t/what-are-side-effects-in-pr-template/11776

syclik:
@t4c1, want to video chat sometime today or tomorrow? @SteveBronder, you free too? I'd like to discuss:

1. design. Have you talk through why it should be done this way, if you've evaluated the other ways to accomplish this, if there's some way to be a little more safe about this.
2. required things to do on the PR regardless of design. I think we've identified that this hasn't met the requirements of doxygen documentation and the PR description should be updated. I haven't even had a chance to see if there are tests that match, so we might have to do that.

Thanks for starting a new thread. I'll respond there for discussion about "side effects."

syclik:
Also... mind fixing the code conflict? We can't move on this PR until that's taken care of too.

t4c1:
That conflict is due to a refactor PR #1407 being reverted because of a bug. After that is fixed and refactor gets back into develop this conflict should go away.

We can do a chat tommorow. I have time between 8h and 15h CET.

syclik:
Ah. I don't think I'm going to be able to make those times tomorrow. (I'm in eastern time zone... if I interpreted the times correctly, that's 2am-9am). Usually, I'd be able to make 14h CET, but tomorrow I have an early meeting.

Friday?

stan/math/opencl/matrix_cl.hpp
syclik: So... I think we're converging on the questions of the lifetime of the `matrix_cl` object, the `cl::Event` object, and the lifetime of the memory in the autodiff stack. That's great!

It wasn't really described in this PR... how long does the `cl::Event` object live? Will `matrix_cl` be in scope longer than the `cl::Event` object that's created? Under what conditions will the `cl::Event` object still be active after `matrix_cl` falls out of scope? Will it ever continue past the computation of a gradient?

We do have arena memory already allocated that's not used for autodiff. It could possibly used for this application. We do that in other parts of the Math library for different things, so it's not a new pattern. The caller is responsible for calling `recover_memory()` after running the reverse sweep of autodiff (calling the `grad()` function or `.chain()` method). I believe that should be enough time. Since it's on the caller, there's no guarantee that the caller will not call `recover_memory()` while building up the stack, but the Math library treats that as undefined behavior. I think from the outside, a client would call `recover_memory()` on an exception, so that might be the condition we have to think through... if we kicked of something with the `cl::Event` still running, but then had an exception and the memory was recovered, what would happen? 


>  BTW next time you need some documentation for OpenCL you can google "OpenCL [function name]" and the first result should be the doc.

FYI, I did do that and I found that particular doc. It really would have helped anyone reviewing if that was included in the PR as support for the design you've used. That said, it wasn't enough information. It would have just been a really good starting point.

SteveBronder:
I could do Friday 14-15 CET

syclik:
Me too.

t4c1:
Friday 14h CET it is!

stan/math/opencl/matrix_cl.hpp
t4c1: > It wasn't really described in this PR... how long does the cl::Event object live? 

`cl::Event` has no different lifetime than any other object - untill the end of the scope it is declared in. In this case that is until the constructor returns. How underlying implementation handles callback so that they can be executed after the event stops existing is not defined.

> Will matrix_cl be in scope longer than the cl::Event object that's created?

It can be and I think it usually would be.

> Under what conditions will the cl::Event object still be active after matrix_cl falls out of scope?

Please be more precise. What active means here? I assume you think in what cases callback is not yet called when matrix goes out of scope. That would be if copying the data is not complete, regardles of how far the execution of the program moved forward.

> but then had an exception and the memory was recovered, what would happen?

It is not defined in the doc so i guess it is undefined behavior. The matrix could contain garbage data or the program would simply segfault.

Thanks for the explanation about arena memory. Unfortunately it raises another concern. That would mean we are essentially leaking data untill the end of the reverse sweep. We could quickly run out of memory if this was used in an iterative algorithm.

syclik: > Thanks for the explanation about arena memory. Unfortunately it raises another concern. That would mean we are essentially leaking data untill the end of the reverse sweep. We could quickly run out of memory if this was used in an iterative algorithm.

Great point! Yes, I didn't quite understand the use case behind this. So... this could be called multiple times within one reverse sweep? Is it possible for there to be more than one `cl::Event` at a time? If so, is there some limit?

> `cl::Event` has no different lifetime than any other object - untill the end of the scope it is declared in. In this case that is until the constructor returns. How underlying implementation handles callback so that they can be executed after the event stops existing is not defined.

This doesn't make sense to me. This is L311-314:
```
      try {
        cl::Event e = initialize_buffer(A_heap->data());
        e.setCallback(CL_COMPLETE, &delete_it<Mat_type>, A_heap);
      } catch (...) {
```

I think you're saying that the `cl::Event` object itself has a lifetime that lasts only within that try block. If that's the case, then why is `setCallback()` a method on `cl::Event`? This code leaves something behind in some global state and it clearly has some lifetime beyond this try block.

- If the callback is set for the event, `e`, itself, then how does `e` accept the notification for the callback and call the different functions that it has queued up?
- If the callback is set for a global event queue (not for this instance of `e`), then how does the global event queue know how to only call these callbacks for this event (which has fallen out of scope)? If it doesn't distinguish across events and only handles `CL_COMPLETE`, then this would be a problem if more than one of the things that are left behind in the global state and the first one triggered these callbacks... that would clear the memory for all of these things at once.

I think either way, it's pretty confusing. You're right... the object `e` drops out of scope. But something is left behind. That's what really needs to be documented clearly. How / why / when is its lifetime over / what assumptions are we making / etc.

> Please be more precise. What active means here? 

Sorry about that. I was just talking about lifetime. To be more clear, this page on cppreference describes lifetime as I know it: https://en.cppreference.com/w/cpp/language/lifetime.

Here, there are two things: 
1. "For any object of class or aggregate types if it, or any of its subobjects, is initialized by anything other than the trivial default constructor, lifetime begins when initialization ends.
For any object of class types whose destructor is not trivial, lifetime ends when the execution of the destructor begins." -- that's for `cl::Event  e`. 
2. "For all other objects (class objects initialized by a trivial default constructor, non-class objects, arrays of those, etc.), lifetime begins when the properly-aligned storage for the object is allocated and ends when the storage is deallocated or reused by another object." That's for the memory that's been allocated.

If I understand correctly, the `matrix_cl` class passes ownership of the memory allocated on the heap to the newly created `cl::Event e`. There's a callback that's registered with `e`, but `e` falls out of scope immediately, so... who owns the memory at this point?

t4c1: > So... this could be called multiple times within one reverse sweep? Is it possible for there to be more than one cl::Event at a time? If so, is there some limit?

Yes it can. There is no limit to number of the number of times this is called in one sweep and no limit on amount of `cl::Event`s in existance.

> I think you're saying that the cl::Event object itself has a lifetime that lasts only within that try block. If that's the case, then why is setCallback() a method on cl::Event? This code leaves something behind in some global state and it clearly has some lifetime beyond this try block.

That `cl::Event` object only lives until the und of try. OpenCL standard does not define how these callbacks are implemented or what state is left behind in the global state so I can not answer how that and the following quiestions. Also it is entirebly possible that it is not done in same way for multiple OpenCL implementations.

> what assumptions are we making

Only that OpenCL functions do what the standard says they should. Nothing about how they are implemented.

> who owns the memory at this point?

Memory ownership AFAIK does not need to be precisely defined. We could say nobody. Or we could say the `matrix_cl` that scheduled the callback. Or we could say it is OpenCL implementation that is responsible for calling the callback.

syclik:
Awesome. Talk to you tomorrow! I'm sending conference info via a Discourse pm.

SteveBronder:
Ack sorry, I don't understand the concept of time zones and thought this was at 9 AM EST.

Was going to ask why we don't just use a shared pointer here instead of the callback. Since then we can keep track of reference counts and it automatically deletes when there are zero references. Little bulky but very clean

syclik:
I just got off a video call with @t4c1. This is neat! Some things that happened:

1. @t4c1 walked me through the design. This one makes sense. He was able to walk me through the common patterns of managing memory and explained how they actually don't quite work out. Specifically, he walked me through the logic of how the common pattern of having a member variable storing the location of the allocated memory isn't ideal. There are two constraints driving the existing design:

   1. this is asynchronous code. In order to handle things with different callbacks or hold onto the reference and delete in the destructor, we'd have to be threadsafe and wait for certain things to happen, which are often out of our control.
   2. early recovery of allocated memory. Using the callback allows the memory to be released back into the system when the function is called. Under most normal conditions, this will happen within the lifetime of the `matrix_cl` object, so it makes sense not to have it when destructed.
    
    That leads to the current design making sense, even though on first glance it looks like a leak. The next point addresses that.

2. We discussed some of the required documentation. Things like `initialize_buffer()` assume ownership of memory under certain conditions and it's not clear what the requirements are for the function. @t4c1 is going to update some of the documentation to make things clearer.

3. @t4c1 found a spot in the constructor that needed to be fixed; he has the details and will fix it.

Overall, the state of the code is nearly there. What is required prior to merging is an update on the documentation so it's clear why this deviates from a standard pattern. That should also include the required documentation on functions mentioning their requirements on memory persistence so it's clear to the next reader what's happening.

Thanks, @t4c1!

t4c1:
@SteveBronder If we used shared pointer (or any other smart pointer) an store it in the `matrix_cl` object we could not release the memory until the `matrix_cl` object goes out of scope. This way we can release the memory as soon it is not needed anymore (which is often much sooner than the matrix goes out of scope).

SteveBronder:
idt you even need to store it though since the shared_ptr is not deleted until it's reference count goes to zero no? You could just make one, pass it into `initialize_buffer` then once the last reference goes away all is well

That sounds right? I'll mess with this later to see if it works

t4c1:
Reference count for the shared ponter is the same thing as the number of shared pointer objects that hold the same pointer. So if we create shared pointer in constructor and don't store it as a member variable that shared ponter will go out of scope when the constructor returns and the memory will be deleted. This is not what we want.

syclik:
@t4c1, this has conflicts that has to be resolved. (Sorry -- I didn't realize you added the doc; adding commits alone does not notify reviewers.)

stan/math/opencl/matrix_cl.hpp
```diff
@@ -220,9 +220,11 @@ class matrix_cl<T, require_arithmetic_t<T>> {
   /**
    * Constructor for the matrix_cl that creates a copy of a std::vector of Eigen
    * matrices on the OpenCL device. Each matrix is flattened into one column
-   * of the resulting matrix_cl.
+   * of the resulting matrix_cl. If a lvalue is passed to this constructor the
+   * caller must make sure that the vector does not go out of scope before
```
syclik: @t4c1, first off, thank you for adding information that there is a requirement on the caller to handle memory.

The documentation isn't quite there yet. It's not clear from reading this when "copying is complete." If I'm trying to understand the requirements on memory management, it would help to be more precise about what this means.

t4c1: Hopefully it is clear now.

t4c1:
@syclik This is now ready for review.

syclik:
@SteveBronder, I "dismissed" your review on the PR. @t4c1 went through and addressed all the comments and I checked it over. There was one outstanding question that was in your court, but since it went unanswered, I think it's fine.

SteveBronder:
Sorry, I stopped responding because I'm uncertain whether this is good or not. If your both comfortable with it then I'm fine with abstaining

SteveBronder:
I've also been pretty busy at work and a lot of the stuff in this PR was unclear and I didn't have time to meaningfully look at it

syclik:
It’s all good. 

The important things are:
1. It’s documented we’ll enough for someone to pick up in the future. 
2. It’s passing tests.
3. It had been reviewed carefully. 

As with most implementations, we’re not looking for the perfect solution. Things change and hopefully we know enough to fix it if it needs to change in the future.

PR #1429, opened 2019-11-03
Adds require_* methods for generic containers, index types, and string convertible

SteveBronder:
## Summary

This adds the following metaprogramming traits

`is_container` - Checks if type is either Eigen or std::vector. I think these are the only two containers we use. We can also expand this definition so that any contiguous memory std container is accepted here (which I think is what we would want)

`is_string_convertible` - We don't use this in math, but I do need it in the rest of Stan. Since this is where we keep our metaprogramming I figured this would be the best place for it. As the name says it checks whether a type is implicitly convertible to a string

This also includes the `require_*` methods for each of the above

`void_t` is backported from C++17 along with `is_detected`, these are used to change the definition of `is_vector_like` to be `true` for any class with a defined `operator[](int)` method (aka anything with array like access)

I also was seeing some weird issues with the new testing framework, making `is_nan` have a more generic template seemed to fix the issue

## Tests

```cpp
./runTests.py test/unit/math/prim/mat/meta/is_container_test.cpp test/unit/math/prim/scal/meta/is_container_test.cpp  test/unit/math/prim/scal/meta/is_string_convertible_test.cpp
```

## Side Effects

Yes, `is_vector_like` previously would work for any eigen type (including generic matrices). The intent was for `is_vector_like` to be `true` for classes with array like access and so I've fixed it up to use some void_t tomfoolery to be `true` for any method with a `operator[](int)`

## Checklist

- [x] Math issue #1308


SteveBronder:
@andrjohns mind if I tag you to review on this one?

stan/math/prim/scal/meta/is_index.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef STAN_MATH_PRIM_SCAL_META_IS_INDEX_HPP
+#define STAN_MATH_PRIM_SCAL_META_IS_INDEX_HPP
+
+#include <stan/math/prim/scal/meta/bool_constant.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Deduces whether type is non-floating point arithmetic
```
andrjohns: This doc doesn't really convey the purpose of the type trait

What's the ideal use-case for this? Is it different from ```is_integer```?

stan/math/prim/scal/meta/require_generics.hpp
```diff
@@ -174,11 +177,37 @@ template <typename T, typename... Types>
 using require_any_not_convertible_t = require_all_not_t<
     std::is_convertible<std::decay_t<T>, std::decay_t<Types>>...>;
 
+// Checks if type is something we would use for string_convertible
+template <typename T>
+using require_string_convertible_t
+    = require_t<is_string_convertible<std::decay_t<T>>>;
```
andrjohns: Decaying the type before passing it to ```is_string_convertible``` could cause a false negative: https://godbolt.org/z/_8H6rK

stan/math/prim/scal/meta/is_index.hpp
SteveBronder: > What's the ideal use-case for this? Is it different from is_integer?

Lol I'm dumb, `is_integer is exactly what I need so I can delete `is_index`

Side note:

stan/math/prim/scal/meta/require_generics.hpp
SteveBronder: Oh my! Thanks for the catch. I use decay a lot in the requires library. I may need to look over if this effects other things

stan/math/prim/scal/meta/is_container.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace stan {
+
+/**
+ * Deduces whether type is convertible to string
+ * @tparam T type to check
+ */
+template <typename Container>
+using is_container = bool_constant<
+    math::disjunction<is_eigen<Container>, is_std_vector<Container>>::value>;
```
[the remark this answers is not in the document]
SteveBronder: `is_vector_like` would fail for Eigen matrices right? I have to look at that. If so then maybe we should change the name of `is_vector_like` to `is_container`? That seems to be more of what it actually looks for (or at least guarantees is true)

stan/math/prim/scal/meta/is_container.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <stan/math/prim/scal/meta/bool_constant.hpp>
+#include <stan/math/prim/scal/meta/disjunction.hpp>
+#include <stan/math/prim/scal/meta/is_eigen.hpp>
+#include <stan/math/prim/scal/meta/is_vector.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Deduces whether type is convertible to string
```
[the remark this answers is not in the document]
SteveBronder: Whups! Will fix

SteveBronder:
Every comment I make starts a new review? Are you able to see the comments I just made or do I have to click all the little review buttons I'm seeing above?

stan/math/prim/scal/meta/is_container.hpp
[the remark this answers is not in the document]
andrjohns: It actually returns true for almost any Eigen type, the ```is_eigen``` type trait tests whether the type is derived from ```EigenBase```, ```DenseBase```, and ```MatrixBase``` - which would cover just about everything, I think

andrjohns:
Classic helpful github! Yep, can see all those comments

stan/math/prim/scal/meta/require_generics.hpp
SteveBronder: We both read too quickly haha, gotta use the `_t` on that `std::decay`!

https://godbolt.org/z/ElJ1YM

stan/math/prim/scal/meta/is_container.hpp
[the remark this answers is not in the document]
SteveBronder: Ended up backporting `void_t` and `is_detected` from C++17 so that `is_vector_like` checks for `operator[int]` now (which I think is a better definition. Still need to write tests for those tho' so I'll ping you for review

SteveBronder:
@bob-carpenter would you have time to take a look at this?

I'm seeing

```bash
./test/unit/math/expect_near_rel.hpp:39: Failure
The difference between x1 and x2 is 1, which exceeds tol, where
x1 evaluates to 0,
x2 evaluates to 1, and
tol evaluates to 0.0001.
expect_near_rel_finite(0, 1, absolute tolerance = 0.0001); absolute diff = 1    in: expect_near_rel; require items x1(i) = x2(i): gradient() grad
```

I'm not too familiar with the new AD testing framework. This PR doesn't touch much of anything near the rev and fwd subtract so not sure why it would be goofing

mcol:
Also my PRs cause the same error in `expect_near_rel`, I hope you find a solution as I  have  no idea! :)

SteveBronder:
@andrjohns apologies for the delay, finally had time to fix this up. Also updated the top level comment to reflect the changes

stan/math/prim/scal/meta/require_generics.hpp
```diff
@@ -174,11 +176,37 @@ template <typename T, typename... Types>
...
+
+template <typename T>
+using require_not_string_convertible_t
```
syclik: Can you add tests in test/unit/math/prim/mat/meta/require_generics_test.cpp?

stan/math/rev/scal/fun/is_nan.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <stan/math/rev/meta.hpp>
 #include <stan/math/rev/core.hpp>
+#include <stan/math/prim/meta.hpp>
```
syclik: This shouldn't be necessary -- it's pulled in through `stan/math/rev/meta.hpp`. If it doesn't build with this line removed, there's a problem with our includes that we need to fix.

stan/math/prim/scal/meta/require_generics.hpp
```diff
@@ -174,11 +176,37 @@ template <typename T, typename... Types>
 using require_any_not_convertible_t = require_all_not_t<
     std::is_convertible<std::decay_t<T>, std::decay_t<Types>>...>;
 
+// Checks if type is something we would use for string_convertible
```
syclik: Could you add the doxygen doc for this?

SteveBronder:
I got it! I'll fix this up tmrw

bob-carpenter:
I can review this if @andrjohns doesn't have time.

SteveBronder:
much appreciated!

stan/math/fwd/scal/fun/is_nan.hpp
```diff
@@ -16,9 +17,9 @@ namespace math {
  * @param x Value to test.
  * @return <code>1</code> if the value is NaN and <code>0</code> otherwise.
  */
-template <typename T>
-inline int is_nan(const fvar<T>& x) {
-  return is_nan(x.val());
+template <typename T, require_fvar_t<T>...>
+inline int is_nan(T&& x) {
```
andrjohns: Should this be a ```bool``` return?

stan/math/prim/scal/meta/is_detected.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+template <typename, template <typename> class, typename = void>
+struct is_detected : std::false_type {};
+
+/**
+ * Checks whether a valid type is detected. Most commonly used to detect
+ *  attributes of objects.
+ * @tparam T The type to be checked.
+ * @tparam Op a template template type which attempts to define a member
+ *  of an object.
```
andrjohns: Is this a typo ('template template')? Or is it actually a template template type?

stan/math/prim/scal/meta/is_vector_like.hpp
```diff
@@ -2,12 +2,21 @@
...
 
 namespace stan {
 
+namespace internal {
+/**
+ * @brief Used to detect if object has operator[](int) defined
+ */
+template <typename T>
+using operator_bracket_t = decltype(std::declval<T>()[int{}]);
+}  // namespace internal
```
andrjohns: Will this also need to account for ```operator[](size_t)```?

stan/math/prim/scal/meta/is_vector_like.hpp
```diff
@@ -22,8 +31,7 @@ namespace stan {
  */
 template <typename T>
 struct is_vector_like
-    : bool_constant<stan::is_vector<T>::value || std::is_pointer<T>::value
-                    || is_eigen<T>::value || std::is_array<T>::value> {};
+    : bool_constant<is_detected<T, internal::operator_bracket_t>::value> {};
```
andrjohns: No changes, just want to say that this method of checking ```vector_like```-ness is neat. Very nice!

stan/math/prim/scal/meta/void_t.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef STAN_MATH_PRIM_SCAL_META_VOID_T_HPP
+#define STAN_MATH_PRIM_SCAL_META_VOID_T_HPP
+
+#include <stan/math/prim/scal/meta/bool_constant.hpp>
+#include <stan/math/prim/scal/meta/is_eigen.hpp>
+#include <stan/math/prim/scal/meta/is_vector.hpp>
```
andrjohns: Are all of these includes needed?

stan/math/rev/scal/fun/is_nan.hpp
```diff
@@ -17,7 +17,10 @@ namespace math {
  *
  * @return <code>1</code> if the value is NaN and <code>0</code> otherwise.
  */
-inline bool is_nan(const var& v) { return is_nan(v.val()); }
+template <typename T, require_var_t<T>...>
+inline bool is_nan(T&& v) {
+  return is_nan(std::forward<decltype(v.val())>(v.val()));
```
andrjohns: Can this be simplified to ``` std::forward<double>``` since we know the type of the ```var``` value?

test/unit/math/prim/meta/is_container_test.cpp
```diff
@@ -0,0 +1,15 @@
...
+TEST(MetaTraits, is_scalar_container_check) {
+  using stan::is_container;
+  EXPECT_FALSE(is_container<int>::value);
+  EXPECT_FALSE(is_container<double>::value);
+}
+
+TEST(MetaTraits, is_mat_container_check) {
+  using stan::is_container;
+  EXPECT_TRUE(is_container<Eigen::MatrixXd>::value);
+  EXPECT_TRUE(is_container<std::vector<double>>::value);
```
andrjohns: Can you also add tests for nested containers? i.e. ```EXPECT_TRUE(is_container<std::vector<Eigen::MatrixXd>>::value)```

I don't think there would be an issue, but just to be safe

stan/math/prim/scal/meta/is_container.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <stan/math/prim/scal/meta/disjunction.hpp>
+#include <stan/math/prim/scal/meta/is_eigen.hpp>
+#include <stan/math/prim/scal/meta/is_vector.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Deduces whether type is eigen matrix or standard vector.
+ * @tparam T type to check
```
andrjohns: Need to update this tparam from ```T``` to ```Container```

andrjohns:
> 
> 
> I can review this if @andrjohns doesn't have time.

@bob-carpenter Ah sorry if you were already in the middle of reviewing this, I didn't see the reply

stan/math/prim/scal/meta/is_vector_like.hpp
SteveBronder: Ty!

stan/math/prim/scal/meta/void_t.hpp
SteveBronder: Nope!

stan/math/fwd/scal/fun/is_nan.hpp
SteveBronder: Yep!

stan/math/prim/scal/meta/is_detected.hpp
SteveBronder: It's actually a template template type. So this wouldn't work for non-templated detectors. This template param is for the detector for the op so it let's us write

```cpp
template <typename T>
using operator_bracket_t = decltype(std::declval<T>()[int{}]);
```

```cpp
template <typename T>
struct is_vector_like
    : bool_constant<is_detected<T, internal::operator_bracket_t>::value> {};
```

Where `operator_bracket_t` is not fully a fully constructed type

I'm pretty sure it needs to be like this so that the type construction is delayed until `void_t<Op<T>>` in 

```cpp
template <typename T, template <typename...> class Op>
struct is_detected<T, Op, void_t<Op<T>>> : std::true_type {};
```

stan/math/prim/scal/meta/is_vector_like.hpp
SteveBronder: Good catch! So what happens here is because `size_t` is implicitly convertible to int the above works fine and dandy. I added a test for this in `is_vector_list_test.cpp` that checks that tests pass for `size_t` and `unsigned int` `operator[]`

SteveBronder:
Ty for the review! I think I got everything, once tests pass I'll ping you

SteveBronder:
@andrjohns feel free to review!

stan/math/prim/meta/is_container.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef STAN_MATH_PRIM_SCAL_META_IS_CONTAINER_HPP
+#define STAN_MATH_PRIM_SCAL_META_IS_CONTAINER_HPP
```
andrjohns: Now that the meta flatten has happened, these header guards should just be be PRIM_META

stan/math/rev/scal/fun/is_nan.hpp
SteveBronder: Yes!

stan/math/prim/meta/is_container.hpp
SteveBronder: Word!

SteveBronder:
Apologies for the delay, think this is all fixed up tho'!

andrjohns:
Some weird opencl failures in the tests:
```
[ RUN      ] MathMatrixOpenCL.cholesky_decompose_big_tuning_opts

test/unit/math/opencl/cholesky_decompose_test.cpp:60: Failure

Expected: (max_error) < (1e-8), actual: 3909.75 vs 1e-08

```
Not sure where that's coming from

rok-cesnovar:
Yeah, those happen rarely but they do. I have looked into it but still havent found the cause of the issue.

andrjohns:
Lgtm, merge away!

PR #1454, opened 2019-11-15
Make matrix_exp_pade work on size 0 inputs

### LINKED ISSUE #1435: matrix_exp_pade should not raise assertion errors on size 0 input
## Description

This test should pass in `test/unit/math/mix/mat/fun/matrix_exp_pade_test.cpp`:

```
  Eigen::MatrixXd m00(0, 0);
  stan::test::expect_ad(f, m00);
```

We need to catch the size 0 x 0 case and throw an exception before the assertion from Eigen kicks in.

#### Current Version:
v3.0.0

mcol:
## Summary

Fixes #1435 so that on inputs of size 0 it returns an empty matrix.

## Tests

Added test mentioned in the issue.

## Side Effects

None.

## Checklist

- [X] Math issue #1435

### DIFF
--- a/stan/math/prim/mat/fun/MatrixExponential.h
+++ b/stan/math/prim/mat/fun/MatrixExponential.h
@@ -55,7 +55,7 @@ namespace Eigen {
   };
 
 
-  /** \brief Compute the (5,5)-Pad&eacute; approximant to the exponential.
+  /** \brief Compute the (3,3)-Pad&eacute; approximant to the exponential.
    *
    *  After exit, \f$ (V+U)(V-U)^{-1} \f$ is the Pad&eacute;
    *  approximant of \f$ \exp(A) \f$ around \f$ A = 0 \f$.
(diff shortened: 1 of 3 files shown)


wds15:
I thought that size zero inputs lead to size zero outputs for all our functions. No?

mcol:
I have followed what @bob-carpenter mentioned in the issue, but I could make it return an empty matrix instead.

bob-carpenter:
The decision should be based on whether the 0 x 0 result has the right properties to be a result.

mcol:
I'm not an expert, but as the exponential of a matrix is defined as the infinite Taylor series `exp(M) = I + M + M^2/2! + M^3/3! + ...`, then I think that returning an empty matrix for a size 0 input should be correct. This is also what happens in R:
```
> a <- matrix(0, 0, 0)
> a
<0 x 0 matrix>
> expm::expm(a)
<0 x 0 matrix>
```

However, it doesn't work in Python (https://github.com/scipy/scipy/issues/11082).

bob-carpenter:
We should just do the same thing for all the matrix exponentials.    The basic definition is

exp(X) = SUM_k X^k / k!

So it makes sense to have a 0 x 0 matrix return a 0 x 0 matrix.

mcol:
Thanks, I've amended the description of this PR and made it return an empty matrix. I've filed #1456 to cover the remaining matrix exponentials.

PR #1457, opened 2019-11-21
Make the matrix exponential functions work on 0 size inputs

mcol:
## Summary

This fixes #1456, so that for size 0 inputs the matrix exponential functions return an empty matrix.

## Tests

There are already tests for size 0 inputs regarding these functions, I've added a couple more and modified one that was expecting a throw. It's worth checking that I haven't missed any.

## Side Effects

None.

## Checklist

- [X] Math issue #1456

- [X] Copyright holder: Marco Colombo

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [X] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [X] the code is written in idiomatic C++ and changes are documented in the doxygen

- [X] the new changes are tested

Issue #1456, opened 2019-11-19
make matrix exponentials work on size 0 inputs


mcol:
## Description
Currently matrix exponentials use `check_nonzero_size` to throw on 0x0 inputs. The discussion in #1454 led to the decision to make all matrix exponentials return a 0x0 matrix on 0x0 inputs. This requires changing `matrix_exp` and `matrix_exp_multiply`.

rok-cesnovar:
Reasons for reopening are given [here](https://github.com/stan-dev/math/pull/1457#issuecomment-557544562)

PR #1459, opened 2019-11-22
follow-up fixes for matrix exponentials on size 0 inputs

mcol:
## Summary

The checks for 0 size for `matrix_exp_multiply` and `scale_matrix_exp_multiply` introduced in #1457 should happen before `check_multipliable`, as the latter throws on size 0 inputs. This was not caught by the existing tests. I've now added tests for this, and I have patches that move those 0 size checks around so that those functions really return an empty matrix on size 0 inputs.

## Tests

Added tests that fail with #1457 but succed with this PR.

## Side Effects

None

## Checklist

- [X] Math issue #1456

### DIFF
--- a/stan/math/prim/mat/fun/matrix_exp_multiply.hpp
+++ b/stan/math/prim/mat/fun/matrix_exp_multiply.hpp
@@ -18,10 +18,11 @@ namespace math {
 template <int Cb>
 inline Eigen::Matrix<double, -1, Cb> matrix_exp_multiply(
     const Eigen::MatrixXd& A, const Eigen::Matrix<double, -1, Cb>& B) {
-  check_multiplicable("scale_matrix_exp_multiply", "A", A, "B", B);
-  check_square("scale_matrix_exp_multiply", "input matrix", A);
-  if (A.size() == 0)
+  if (A.size() == 0 && B.size() == 0)
     return {};
+
+  check_multiplicable("matrix_exp_multiply", "A", A, "B", B);
+  check_square("matrix_exp_multiply", "input matrix", A);
 
   return matrix_exp_action_handler().action(A, B);
 }
--- a/stan/math/prim/mat/fun/scale_matrix_exp_multiply.hpp
+++ b/stan/math/prim/mat/fun/scale_matrix_exp_multiply.hpp
@@ -20,10 +20,11 @@ template <int Cb>
 inline Eigen::Matrix<double, -1, Cb> scale_matrix_exp_multiply(
     const double& t, const Eigen::MatrixXd& A,
     const Eigen::Matrix<double, -1, Cb>& B) {
+  if (A.size() == 0 && B.size() == 0)
+    return {};
+
   check_multiplicable("scale_matrix_exp_multiply", "A", A, "B", B);
   check_square("scale_matrix_exp_multiply", "input matrix", A);
-  if (A.size() == 0)
-    return {};
 
   return matrix_exp_action_handler().action(A, B, t);
 }
--- a/stan/math/rev/mat/fun/matrix_exp_multiply.hpp
+++ b/stan/math/rev/mat/fun/matrix_exp_multiply.hpp
@@ -25,10 +25,11 @@ template <typename Ta, typename Tb, int Cb>
 inline Eigen::Matrix<typename stan::return_type<Ta, Tb>::type, -1, Cb>
 matrix_exp_multiply(const Eigen::Matrix<Ta, -1, -1>& A,
                     const Eigen::Matrix<Tb, -1, Cb>& B) {
+  if (A.size() == 0 && B.size() == 0)
+    return {};
+
   check_multiplicable("matrix_exp_multiply", "A", A, "B", B);
   check_square("matrix_exp_multiply", "input matrix", A);
-  if (A.size() == 0)
-    return {};
 
   return multiply(matrix_exp(A), B);
 }
--- a/stan/math/rev/mat/fun/scale_matrix_exp_multiply.hpp
+++ b/stan/math/rev/mat/fun/scale_matrix_exp_multiply.hpp
@@ -28,10 +28,11 @@ template <typename Ta, typename Tb, int Cb>
 inline Eigen::Matrix<typename stan::return_type<Ta, Tb>::type, -1, Cb>
 scale_matrix_exp_multiply(const double& t, const Eigen::Matrix<Ta, -1, -1>& A,
                           const Eigen::Matrix<Tb, -1, Cb>& B) {
+  if (A.size() == 0 && B.size() == 0)
+    return {};
+
   check_multiplicable("scale_matrix_exp_multiply", "A", A, "B", B);
   check_square("scale_matrix_exp_multiply", "input matrix", A);
-  if (A.size() == 0)
-    return {};
 
   return multiply(matrix_exp(multiply(A, t)), B);
 }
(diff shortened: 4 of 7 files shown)


stan/math/prim/mat/fun/matrix_exp_multiply.hpp
```diff
@@ -18,11 +18,12 @@ namespace math {
 template <int Cb>
 inline Eigen::Matrix<double, -1, Cb> matrix_exp_multiply(
     const Eigen::MatrixXd& A, const Eigen::Matrix<double, -1, Cb>& B) {
-  check_multiplicable("scale_matrix_exp_multiply", "A", A, "B", B);
-  check_square("scale_matrix_exp_multiply", "input matrix", A);
-  if (A.size() == 0)
+  if (A.size() == 0 && B.size() == 0)
```
rok-cesnovar: Should this be 
```cpp
if (A.size() == 0 || B.size() == 0)
```

stan/math/prim/mat/fun/scale_matrix_exp_multiply.hpp
```diff
@@ -20,10 +20,11 @@ template <int Cb>
 inline Eigen::Matrix<double, -1, Cb> scale_matrix_exp_multiply(
     const double& t, const Eigen::MatrixXd& A,
     const Eigen::Matrix<double, -1, Cb>& B) {
+  if (A.size() == 0 && B.size() == 0)
```
rok-cesnovar: Same question here.

stan/math/rev/mat/fun/matrix_exp_multiply.hpp
```diff
@@ -25,10 +25,11 @@ template <typename Ta, typename Tb, int Cb>
 inline Eigen::Matrix<typename stan::return_type<Ta, Tb>::type, -1, Cb>
 matrix_exp_multiply(const Eigen::Matrix<Ta, -1, -1>& A,
                     const Eigen::Matrix<Tb, -1, Cb>& B) {
+  if (A.size() == 0 && B.size() == 0)
```
rok-cesnovar: And here.

stan/math/rev/mat/fun/scale_matrix_exp_multiply.hpp
```diff
@@ -28,10 +28,11 @@ template <typename Ta, typename Tb, int Cb>
 inline Eigen::Matrix<typename stan::return_type<Ta, Tb>::type, -1, Cb>
 scale_matrix_exp_multiply(const double& t, const Eigen::Matrix<Ta, -1, -1>& A,
                           const Eigen::Matrix<Tb, -1, Cb>& B) {
+  if (A.size() == 0 && B.size() == 0)
```
rok-cesnovar: And also here.

test/unit/math/prim/mat/fun/matrix_exp_multiply_test.cpp
```diff
@@ -29,6 +29,11 @@ inline void test_matrix_exp_multiply() {
 }
 
 TEST(MathMatrix, matrix_exp_multiply) {
+  // the helper above doesn't handle 0 size inputs
+  Eigen::MatrixXd A(0, 0);
+  Eigen::MatrixXd B(0, 0);
+  EXPECT_EQ(stan::math::matrix_exp_multiply(A, B), A);
```
rok-cesnovar: Can you add a test for A(0,1) and B(1,2)?

test/unit/math/prim/mat/fun/scale_matrix_exp_multiply_test.cpp
```diff
@@ -30,6 +30,12 @@ inline void test_scale_matrix_exp_multiply() {
 }
 
 TEST(MathMatrix, scale_matrix_exp_multiply) {
+  // the helper above doesn't handle 0 size inputs
+  const double t = 1.0;
+  Eigen::MatrixXd A(0, 0);
+  Eigen::MatrixXd B(0, 0);
```
rok-cesnovar: Can you add a test for A(0,1) and B(1,2)?

test/unit/math/rev/mat/fun/scale_matrix_exp_multiply_test.cpp
```diff
@@ -182,6 +182,14 @@ TEST(MathMatrix, scale_matrix_exp_multiply_vv) {
   test_scale_matrix_exp_multiply_vv(8, 2);
 }
 
+TEST(MathMatrix, matrix_exp_multiply_0x0) {
+  using stan::math::var;
+  const double t = 1.0;
+  Eigen::Matrix<var, -1, -1> A(0, 0);
+  Eigen::Matrix<var, -1, -1> B(0, 0);
```
rok-cesnovar: Can you add a test for A(0,1) and B(1,2)

stan/math/prim/mat/fun/matrix_exp_multiply.hpp
mcol: If `A.size() == 0` but `B.size() != 0` then `exp(A)` and `B` are not multiplicable, and we should throw instead of returning an empty matrix. Same thing if If `A.size() != 0` and `B.size() == 0`. The only case where the multiplication can be done if A or B are 0 size is when the other is 0 size as well.

stan/math/prim/mat/fun/scale_matrix_exp_multiply.hpp
mcol: This is the same as above, as the scaling by `t` is ininfluential.

test/unit/math/prim/mat/fun/matrix_exp_multiply_test.cpp
mcol: In this case we should throw because A is not square. I'm not sure this is what you wanted to see, but I'll add this test anyway.

test/unit/math/prim/mat/fun/scale_matrix_exp_multiply_test.cpp
mcol: Added.

test/unit/math/rev/mat/fun/scale_matrix_exp_multiply_test.cpp
mcol: Added.

stan/math/prim/mat/fun/matrix_exp_multiply.hpp
rok-cesnovar: Ah yeah, forgot about that A must be square. How about the case where A(5,5) and B(5,0)? Not saying this is wrong, just a question.

rok-cesnovar: Or does the matrix_exp_action_handler handle that case as expected?

mcol: This case would cause `check_multiplicable` to throw because B has 0 columns. This is perhaps overly strict, but I'm not sure if there is any value in allowing it, as it would have failed even if we had A * B rather than exp(A) * B.

rok-cesnovar: Ah ok, I actually was not aware that check_multiplicable checks that rows of A > 0 and cols of B > 0. So this one is on me. 

I don't think that this behavior is the best as its not clear from the name that it does this, but that is a separate issue, definitely out of scope for this PR.

Sorry for the confusion here. This is good to go. Thanks again!

PR #1464, opened 2019-11-25
Throw when computing inverse of a size 0 matrix

### LINKED ISSUE #1463: inverse should have consistent errors on size 0 inputs
## Description
There's a test that's currently commented out in test/unit/math/mix/mat/fun/inverse_test.cpp because it fails, due to inconsistent behaviour of `inverse` on size 0 inputs between prim and rev.

That test could be replaced with the following, which is more general:

```
Eigen::MatrixXd t2(0, 0);
stan::test::expect_ad(f, t2);
```

It will work after making prim throw on size 0 inputs.

mcol:
## Summary

Fixes #1463.

## Tests

Added test mentioned in the issue and another one too.

## Side Effects

None.

## Checklist

- [X] Math issue #1463

### DIFF
--- a/stan/math/prim/mat/fun/inverse.hpp
+++ b/stan/math/prim/mat/fun/inverse.hpp
@@ -2,6 +2,7 @@
 #define STAN_MATH_PRIM_MAT_FUN_INVERSE_HPP
 
 #include <stan/math/prim/mat/fun/Eigen.hpp>
+#include <stan/math/prim/mat/err/check_nonempty.hpp>
 #include <stan/math/prim/mat/err/check_square.hpp>
 
 namespace stan {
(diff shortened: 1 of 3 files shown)


mcol:
While we are here we could perhaps remove this other comment:
```
  // fails with assertion from Eigen
  // stan::test::expect_ad(f, t);
```

If I understand correctly, expect_ad doesn't work with `t` being an `Eigen::Matrix<stan::math::var, -1, -1>`, so this will never be valid code.

bob-carpenter:
expect_ad(f, x) should work when x is a matrix of any shape, including 0 x 0.  If it doesn't, it's a bug.

mcol:
The problem there is that it's Matrix of `stan::math::var`: that comment is wrong, as the test doesn't fail, it doesn't even compile.

bob-carpenter:
I misunderstood---I missed the var.  You're right the framework doesn't work with var inputs!

Issue #1470, opened 2019-11-29
Generalize matrix function signatures


t4c1:
## Description
Matrix functions can be generalized to work with Eigen expressions not just matrices. By propagating matrix expressions around we can avoid creating temporary matrices, improving performance. This could also make adding support for sparse matrices easier.

There was some discoussion about this [on discourse](https://discourse.mc-stan.org/t/refactoring-math-with-more-general-signatures/10157/6). Also an abandoned PR #1309. Contrary to this PR I intend to split the this into smaller chunks:
- [x] element-wise functions (ones that use `apply_scalar_unary`). Prim implementation of these can in many cases use Eigen implementations that can be faster than std.
- [x] operator-like functions (add, subtract, multiply, divide, elt_multiply, elt_divide)
- [x] size and view functions (col, row, sub_col, sub_row, head, tail, diagonal, transpose, cols, rows, num_elements, dims)
- [x] reminder of */fun
- [x] */prob
- [ ] allow functions to return expressions

## Example
Instead of :
```
Eigen::MatrixXd f(const Eigen::MatrixXd& x)
```
We can do:
```
template<typename T>
auto f(const T& x)
```

## Expected Output
More general and possibly faster code.

#### Current Version:
v3.0.0

bob-carpenter:
+1 for this.  We also need to do this to allow support for sparse matrices.  @SteveBronder has been working on that along with perfect forwarding so we can reduce copying and exploit move semantics wherever possible.

andrjohns:
We'd need to start coding a bit more defensively after implementing this, otherwise there's the risk of accidentally evaluating the expression multiple times. Take ```log_sum_exp``` for example:
```
template <typename T>
auto log_sum_exp(T& x) {
  if (x.size() == 0) {
    return -std::numeric_limits<double>::infinity();
  }

  const double max = x.maxCoeff();
  if (!std::isfinite(max)) {
    return max;
  }
  return max + std::log((x.array() - max).exp().sum());
}
```
If ```log_sum_exp``` is passed an expression, something like ```log_sum_exp(m1.array().exp().matrix() * m2)```, then that expression will get evaluated three times: ```x.size()```, ```x.maxCoeff()```, and the final return statement. So we'd need to start evaluating the expression at the beginning of the function, and probably also updating the documentation so that other/new developers know why

t4c1:
[...]

PR #1471, opened 2019-11-29
Generalize element-wise functions

t4c1:
## Summary
This PR generalizes signatures of element-wise matrix functions (ones that use `apply_scalar_unary`) so that these accept and return expressions. For `Eigen::Matrix`, `Eigen::Array` and expressions of arithmetic types these functions now use Eigen implementation if it exists.

## Tests
These functions already have tests. They use the new ad-testing framefork. I had to modify the framework a bit so it evaluates expressions it gets from the tested functions. 

## Side Effects
These functions now also work with `Eigen::Array`s and array expressions.

## Checklist

- [x] Math issue #1470 


stan/math/prim/mat/fun/acos.hpp
```diff
@@ -27,11 +29,23 @@ struct acos_fun {
...
+ * @tparam Derived derived type of x
+ * @param x Matrix or matrix expression
+ * @return Arc cosine of each variable in the container, in radians.
+ */
+template <typename Derived,
+          typename = require_eigen_vt<std::is_arithmetic, Derived>>
+inline auto acos(const Eigen::MatrixBase<Derived>& x) {
+  return x.derived().array().acos().matrix();
+}
+
```
andrjohns: This implementation will call Eigen's ```acos``` for Eigen objects (```Matrix```,```Array```,etc.), but not ```std::vectors``` of Eigen objects - as those will be handled by ```apply_scalar_unary```, so there's some possible performance gain there

andrjohns:
Would it be worth testing the performance gain of ```Eigen::Map```-ing ```std::vectors``` so that the Eigen-specific implementations can be used?

t4c1:
I don't think that needs testing. I am quite sure `Map`-ing `std::vector`s would be beneficial. You are working on that anyway, aren't you? I tried not to steal work from you :)

SteveBronder:
I think the error on upstream is actually a 'bug' we need to fix upstream. Last time I tried this I got similar errors so switched to making the upstream stan repo much more generic (still working on that but waiting on some things in #1429

```
./test/test-models/good-standalone-functions/integrate.hpp:40:16: error: no viable conversion from returned value of type 'typename promote_scalar_type<double, MatrixWrapper<const CwiseUnaryOp<scalar_exp_op<double>, const ArrayWrapper<const CwiseUnaryOp<(lambda at lib/stan_math/stan/math/prim/mat/vectorize/apply_scalar_unary.hpp:52:9), const Matrix<double, -1, 1, 0, -1, 1> > > > > >::type' (aka 'double') to function return type 'Eigen::Matrix<typename boost::math::tools::promote_args<double>::type, Eigen::Dynamic, 1>' (aka 'Matrix<double, Eigen::Dynamic, 1>')

        return stan::math::promote_scalar<fun_return_scalar_t__>(stan::math::exp(minus(square(x))));
```

t4c1:
I think I might be able to fix this with some changes to `promote_scalar_type`.

SteveBronder:
Nice! With +600 changes it may be good to make this a first PR and make a new branch / PR off this one to continue

t4c1:
Yeah, I don't intend to add anymore changes to this PR, except what is needed for tests to pass.

andrjohns:
Sorry to dump this in so late in the cycle, but it looks like Eigen's ```exp``` implementation has some issues with underflow and overflow, compared to the standard library: https://gitlab.com/libeigen/eigen/issues/1693

So this could be a decision tradeoff between speed in the majority of cases and loss of precision in the minority of cases

t4c1:
That is bad. Although that bug report is for floats. I will test it with doubles. If doubles are also bugged I think we should not use Eigen's implementation.

t4c1:
Good news. Double implementation works for all values except denorms. I think that should be good enough for us.

t4c1:
I did some changes to Jenkinsfile to speed up testing. They need to be reverted before merge.

bob-carpenter:
That's what I suspected.  It's what Intel's compiler does for speed, too.  When we evaluated the Intel compiler, we had to increase the tolerance of our tests.

stan/math/prim/scal/meta/scalar_seq_view.hpp
```diff
@@ -33,7 +38,7 @@ class scalar_seq_view<
   int size() const { return c_.size(); }
 
  private:
-  C c_;
+  math::plain_type_keep_constness_and_ref_t<C> c_;
```
SteveBronder: I think we just want to keep the const and not the ref and that should fix the current thing on jenkins, guessing tho just from the stackoverflow post below 

https://stackoverflow.com/questions/38226443/const-reference-to-member-of-temporary-object

SteveBronder: Also we should keep meta type traits stuff in the stan namespace and not math, that's where the rest of the stuff is

rok-cesnovar: We are actually [not terribly consistent](https://github.com/stan-dev/math/issues/1122#issuecomment-468953210) with the namespaces, tho the majority is in stan::.

t4c1: > and not the ref 
We definitely want ref to avoid unnecessary copys whan a lvalue is passed to constructor. I am not extending lifetime of an object with a const ref (which is what that SO question is about). I just want to keep ref to the object that is passed in without making any deep copys. And I am running out of ideas how to achieve that. 

Afaik `scalar_seq_view` is only constructed from lvalues. I only used perfect forwarding in attempt to work around the warning this triggers in clang. Btw this is valid code gcc is happy with. However with -Wall that warning fails compilation with clang.

About that namespace. I agree with consistancy point. However such details can wait until code works.

SteveBronder: Word!

For the namespace thing we are def super dodgy on it. When I see meta stuff in math and not stan I try to fix it up but we def lack consistency

t4c1:
This is now ready for a review.

bob-carpenter:
I can review this one this week, but if someone else wants to, please go ahead.

stan/math/prim/mat/meta/as_column_vector_or_scalar.hpp
```diff
@@ -15,9 +16,8 @@ namespace math {
  * @param a Specified vector.
  * @return Same vector.
  */
-template <typename T>
-inline const Eigen::Matrix<T, Eigen::Dynamic, 1>& as_column_vector_or_scalar(
-    const Eigen::Matrix<T, Eigen::Dynamic, 1>& a) {
+template <typename T, typename = require_t<is_eigen_col_vector<T>>>
```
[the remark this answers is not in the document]
t4c1: This is the version for column vector. Below are also versions for row vector and std vector.

stan/math/prim/mat/fun/log1p.hpp
```diff
@@ -34,11 +35,25 @@ struct log1p_fun {
...
+// * Version of log1p() that accepts Eigen Matrix ar matrix expressions.
+// * @tparam Derived derived type of x
+// * @param x Matrix or matrix expression
+// * @return Elementwise log1p of members of container.
+// */
+// template <typename Derived, typename = require_eigen_vt<std::is_arithmetic,
+// Derived>>  inline auto log1p(const Eigen::MatrixBase<Derived>& x){
+//  return x.derived().array().log1p().matrix();
+//}
+
```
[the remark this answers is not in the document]
t4c1: I forgot to delete this. The problem is the scalar version of `log1p()` throws for invalid values. So we can not directely use eigen implementation that does not throw. Since eigen implementation is no faster that `applyScalarUnary` I decided not to complicate and leave this as is.

stan/math/prim/mat/meta/as_column_vector_or_scalar.hpp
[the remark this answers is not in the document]
SteveBronder: Word, I'm fine with require_t here, unless we do this all the time idt we need a specialized require for it

rok-cesnovar:
@t4c1 This PR had a lot of changes to meta that are affected by the meta flattening so I went ahead and merged it in.

t4c1:
This is now ready for a review.

SteveBronder:
@t4c1 did you see my thing about the performance tests above?

Once that and the merge conflicts are resolved I can have a look at this

t4c1:
I thout I fixed that, but I did not check. Now it should work.

SteveBronder:
> (stat_comp_benchmarks/benchmarks/arK/arK.stan, 1.33)

mamma mia!

rok-cesnovar:
Is that faster or slower. I keep forgetting :)

SteveBronder:
lol I always forget as well I'm pretty sure it's `develop/PR` so > 1 = good

serban-nicusor-toptal:
Hey, currently testing a PR in Jenkins which requires me to disable some nodes. I'll restart this job in a few minutes, sorry for the trouble!

stan/math/prim/mat/fun/exp.hpp
```diff
@@ -36,11 +38,23 @@ struct exp_fun {
...
  */
-template <typename T>
-inline typename apply_scalar_unary<exp_fun, T>::return_t exp(const T& x) {
+template <typename T, typename = require_not_eigen_vt<std::is_arithmetic, T>>
+inline auto exp(const T& x) {
   return apply_scalar_unary<exp_fun, T>::apply(x);
 }
 
+/**
+ * Version of exp() that accepts Eigen Matrix ar matrix expressions.
```
SteveBronder:
```suggestion
 * Version of exp() that accepts Eigen Matrix or matrix expressions.
```

stan/math/prim/meta/as_column_vector_or_scalar.hpp
```diff
@@ -16,7 +16,7 @@ namespace math {
  * @param a Specified scalar.
  * @return 1x1 matrix that contains the value of scalar.
  */
-template <typename T>
+template <typename T, typename = require_stan_scalar_t<T>>
```
SteveBronder: [comment]

Recently I've been trying to use more 'nice' names for template types. Like for this the template type could be Scalar, functions that take in matrices can have a template type of `Mat`. Not needed for this PR but for future reference I've found it to make my code a bit nicer to read

stan/math/prim/meta/get.hpp
```diff
@@ -44,7 +44,7 @@ inline T get(const std::vector<T>& x, size_t n) {
  * @return n-th element of the \c Eigen \c Matrix or expression
  */
 template <typename T, typename = require_eigen_t<T>>
-inline auto get(const T& m, size_t n) {
+inline typename T::Scalar get(const T& m, size_t n) {
```
SteveBronder: Pretty sure this would work fine
```suggestion
inline scalar_type_t<T> get(const T& m, size_t n) {
```

stan/math/prim/meta/plain_type.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <stan/math/prim/meta/plain_type.hpp>
+#include <stan/math/prim/meta/require_generics.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Determines plain (non expression) type associated with \c T. For non \c Eigen
+ * types it is the input type.
```
SteveBronder:
```suggestion
 * types it is the decayed input type.
```

stan/math/prim/meta/plain_type.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+template <typename T>
+using plain_type_t = typename plain_type<T>::type;
+
+/**
+ * Determines return type of calling .eval() on eigen expression. This is the
+ * same as \c plain_type, except if input is already a plain type. In such case
+ * eval return type is a const reference to plain type.
+ * @tparam T type to determine eval return type of
+ */
```
SteveBronder: I think I'd phrase this as

```
/**
 * Determines type for a call to .eval() on an eigen expression. if decaying \c T gives the same type as
 * \c plain_type_t<T> then the member \c type will be \c const \c T&. 
 * Otherwise \c type will be the same as the \c type of plain_type_t<T>
 * @tparam T type to determine eval return type of
 */
```

stan/math/prim/meta/plain_type.hpp
```diff
@@ -0,0 +1,51 @@
...
+template <typename T>
+struct eval_return_type {
+  using T1 = plain_type_t<T>;
+  using type = std::conditional_t<std::is_same<std::decay_t<T>, T1>::value,
+                                  const T1&, T1>;
+};
```
SteveBronder: idt I'm totally understanding the intent of this function i.e. why specifically is `type` a `const T1&` when the decayed type is the same as the `plain_type`

stan/math/prim/meta/promote_scalar_type.hpp
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * The promoted type.
+   */
+  using type = typename std::conditional<
+      std::is_same<typename Eigen::internal::traits<std::decay_t<S>>::XprKind,
+                   Eigen::MatrixXpr>::value,
+      Eigen::Matrix<typename promote_scalar_type<T, typename S::Scalar>::type,
+                    S::RowsAtCompileTime, S::ColsAtCompileTime>,
+      Eigen::Array<typename promote_scalar_type<T, typename S::Scalar>::type,
+                   S::RowsAtCompileTime, S::ColsAtCompileTime>>::type;
```
SteveBronder: `scalar_type_t<S>`

SteveBronder: Or is this defined before scalar_type?

SteveBronder: Trying to make sure I'm reading this correctly.

"If `S` is an Eigen expression type is an Eigen matrix whose `Scalar` is the promoted `Scalar` type of T and S. Otherwise type is an Eigen array whose `Scalar` type is the promoted type `Scalar` type of `T` and `S`"

Can you clarify what this is for? In isolation I'm not getting the logic

stan/math/prim/meta/scalar_seq_view.hpp
```diff
@@ -33,7 +38,7 @@ class scalar_seq_view<
   int size() const { return c_.size(); }
 
  private:
-  C c_;
+  eval_return_type_t<C> c_;
```
SteveBronder: ooohh okay I think I'm getting it. Looking at this though I think I'd prefer is `eval_return_type_t` just returned the eval'd type and at the definition using `const eval_return_type_t<C>&`. I think that would make `eval_return_type_t` more generic

stan/math/prim/scal/fun/value_of_rec.hpp
```diff
@@ -20,7 +20,7 @@ namespace math {
  * @param x Scalar to convert to double.
  * @return Value of scalar cast to a double.
  */
-template <typename T>
+template <typename T, typename = require_stan_scalar_t<T>>
```
SteveBronder: Just to double check, `require_stan_scalar_t` works for var, fvar, double, int, etc. Is that what you want here?

test/unit/math/opencl/device_functions/log1m_exp_test.cpp
```diff
@@ -27,7 +27,7 @@ TEST(MathMatrixCL, log1m_exp) {
   Eigen::VectorXd res = stan::math::from_matrix_cl<-1, 1>(res_cl);
 
   stan::test::expect_near_rel("log1m_exp (OpenCL)", res,
-                              stan::math::log1m_exp(a));
+                              stan::math::log1m_exp(a).eval());
```
SteveBronder: Guessing this is getting picked up by the more generic version of `expect_near_rel` which then requires the .eval()

```cpp
void expect_near_rel(const std::string& msg, const T1& x1, const T2& x2,
                     double tol = 1e-8) {
```

I think we could just slap a `require_not_eigen_t<T1, T2>` there? Then it should go to the right specialization

bob-carpenter:
> 
> Recently I've been trying to use more 'nice' names for template types. Like for this the template type could be Scalar, functions that take in matrices can have a template type of Mat. Not needed for this PR but for future reference I've found it to make my code a bit nicer to reade

What's nice for one dev is naughty for another :-)

I like shorter template parameter names like T for simple scalar types.  It's what the standard library doc does, too, and pretty much all of Boost.  

Eigen's prone to use longer template parameter names like Derived, but even they use T for the scalar type in a matrix.

SteveBronder:
Should probably move this somewhere else, but one more comment from me

> What's nice for one dev is naughty for another :-)

Fair!

> I like shorter template parameter names like T for simple scalar types.  

That makes sense to me for simple simple simple types like arithmetic or integral. But it feels like "simple" starts carrying a lot of weight when we start talking about even a _simple_ notch up in larger or more complicated types

I think it's more "self-documenting" to have expressive name. Like

```cpp
template <typename T1, typename T2> // boilerplate :-(
// ...
template <typename Arith, typename Var> // expressive :-D
```

> It's what the standard library doc does, too, and pretty much all of Boost.

Standard is gung-ho [about expressive template names](https://en.cppreference.com/w/cpp/language/constraints)!

stan/math/prim/meta/get.hpp
t4c1: I prefere using member type instead of using metaprogram to keep the code simpler. But we can also go your way for consistency.

stan/math/prim/meta/as_column_vector_or_scalar.hpp
t4c1: While I agree, I think this should go in a separate PR as this one is big enough.

stan/math/prim/meta/plain_type.hpp
t4c1: I find your version a bit more confusing. If you want I can add in code fragments from your version, but I would prefere not to use it in whole.

stan/math/prim/meta/plain_type.hpp
t4c1: If `std::is_same<std::decay_t<T>` is `true` than T is a plain type - either `Eigen::Matrix` or `Eigen::Array`. These two types have overloaded `.eval()` that just returns a const ref to `this`. As `eval_return_type` is intended to be used to hold return value from calling eval it can be a const ref in these cases. Always using `plain_type` whithout ref would work, but would make a copy in these cases.

stan/math/prim/meta/promote_scalar_type.hpp
t4c1: This could use `scalar_type_t<S>`, but it is unnecessary, since we know we have an Eigen type.

Btw this file is not new. I just moved it from /fun to /meta unchanged, as it was in wrong place, which meant I could not use it in other metaprograms.

t4c1: For example if we have some container that could be either eigen type, sdt::vector ... and we want the same container type with a different scalar we can use this. I needed this in changes to `empty_broadcast_array`.

Again I made no changes to this, just moved it to /meta.

stan/math/prim/meta/as_column_vector_or_scalar.hpp
SteveBronder: ^yes agree

stan/math/prim/meta/scalar_seq_view.hpp
t4c1: > I think I'd prefer is eval_return_type_t just returned the eval'd type and at the definition using const eval_return_type_t<C>&.

I think I don't understand this. Are you suggesting `eval_return_type_t<T>` to always be a const ref, even if `T` is not a plain type? That would not work for expressions - we would have a dangling ref in such case.

stan/math/prim/scal/fun/value_of_rec.hpp
t4c1: This works fine as we have overloads for double, var and fvar. I only need this overload for int and anything else that matches that "etc". Since I am not sure what that could be I just went with the more general requirement. Do we have any more specific `require_...` I should use here?

test/unit/math/opencl/device_functions/log1m_exp_test.cpp
t4c1: First we still need this to match plain Eigen types so your suggestion would not work. Second compiler can not make a conversion of argument and deduce its template at the same time, so this would not work even if we constructed a correct `require_...`

stan/math/prim/mat/fun/exp.hpp
t4c1: done

stan/math/prim/meta/get.hpp
t4c1: done

stan/math/prim/meta/plain_type.hpp
t4c1: done

stan/math/prim/meta/plain_type.hpp
t4c1: done

bob-carpenter:
Stan only produces double, int, var, and fvar types.   There's a bigger range of built-in types and some of our functions, particularly constructors, are adapted to take all of them without ambiguity.  I'm guessing that the constructors for var and fvar could be simplified with some traits metaprogramming.

stan/math/prim/meta/promote_scalar_type.hpp
SteveBronder: > Btw this file is not new.

Ah word word no need to worry about stuff in this. Optional fix but we can tidy it up another day

stan/math/prim/meta/plain_type.hpp
SteveBronder: > These two types have overloaded .eval() that just returns a const ref to this. As eval_return_type is intended to be used to hold return value from calling eval it can be a const ref in these cases. Always using plain_type whithout ref would work, but would make a copy in these cases.

Ah okay cool cool I didn't know that about Eigen with `.eval()`. Should we put that in the docs of plaintype somewhere? (less it's already there and I missed it)

stan/math/prim/meta/scalar_seq_view.hpp
SteveBronder: No I was mistaken about what eval_return_type_t is doing. What do you think about calling it `eigen_return_type_t<T>`? It feels like it's made for eigen specifically

SteveBronder: > What do you think about calling it eigen_return_type_t<T>? It feels like it's made for eigen specifically

Or actually just override `return_type_t` for eigen matrices and arrays?

stan/math/prim/scal/fun/value_of_rec.hpp
SteveBronder: Nope I think this is fine

rok-cesnovar:
I went ahead and merged recent develop in to fix the conflicts that arised from my flattening PRs. I will hold off flattening /prim until this PR is merged to avoid additional conflicts as this PRs seems close to being merged.

stan/math/prim/mat/fun/asin.hpp
```diff
@@ -29,11 +31,23 @@ struct asin_fun {
...
  */
-template <typename T>
-inline typename apply_scalar_unary<asin_fun, T>::return_t asin(const T& x) {
+template <typename T, typename = require_not_eigen_vt<std::is_arithmetic, T>>
+inline auto asin(const T& x) {
   return apply_scalar_unary<asin_fun, T>::apply(x);
 }
 
+/**
+ * Version of asin() that accepts Eigen Matrix ar matrix expressions.
```
mcol: The typo "ar matrix expressions" is repeated in a few files.

stan/math/prim/mat/fun/value_of_rec.hpp
```diff
@@ -40,15 +31,12 @@ inline Eigen::Matrix<double, R, C> value_of_rec(
  *
  * <p>This inline pass-through no-op should be compiled away.
  *
- * @tparam R number of rows in the matrix, can be Eigen::Dynamic
- * @tparam C number of columns in the matrix, can be Eigen::Dynamic
- *
```
mcol: Add doc for `@tparam T`.

stan/math/prim/meta/scalar_seq_view.hpp
t4c1: > What do you think about calling it eigen_return_type_t? It feels like it's made for eigen specifically

It is. However it is made more specifically to determine return type of the `.eval()` member function so I put that in the name.

> Or actually just override return_type_t for eigen matrices and arrays?

Despite the similar name this has different intentions than `return_type`. `return_type` promotes scalars to doubles, while this determines what type is returned by calling `.eval()`. I think maybe we could find a better name for `return _type`.

stan/math/prim/mat/fun/value_of_rec.hpp
t4c1: done

stan/math/prim/mat/fun/asin.hpp
t4c1: done

bob-carpenter:
For vectorized functions, shouldn't the return type be the same as the input type?

std::vector<T> log(const std::vector<T>&);

Eigen::Matrix<T, R, C> log(const Eigen::Matrix<T, R, C>&);

> What do you think about calling it eigen_return_type_t? It feels like it's made for eigen specifically
> 
> It is. However it is made more specifically to determine return type of the .eval() member function so I put that in the name.

t4c1:
`eval_return_type` has nothing to do with return type of any functions in stan math. It just determines what calling `.eval()` on an Eigen expression would return.

test/unit/math/opencl/device_functions/log1m_exp_test.cpp
SteveBronder: > First we still need this to match plain Eigen types so your suggestion would not work.

You can just do the inverse

Changing signature of `expect_near_rel` below passes these tests without `.eval()`. It's better for test writers if they don't have to remember another gotcha in the future so I think it's important to catch handle this stuff in the call

```cpp

// line 54
template <typename EigMat1, typename EigMat2, require_all_eigen_t<EigMat1, EigMat2>...>
void expect_near_rel_finite(const std::string& msg, EigMat1&& x1, EigMat2&& x2) {
  EXPECT_EQ(x1.rows(), x2.rows());
  EXPECT_EQ(x1.cols(), x2.cols());
  auto x1_eval = x1.eval();
  auto x2_eval = x2.eval();
  for (int i = 0; i < x1.size(); ++i)
    expect_near_rel_finite(msg, x1_eval(i), x2_eval(i));
}

void expect_near_rel_finite(const std::string& msg, const std::vector<T1>& x1,
                            const std::vector<T2>& x2) {
  EXPECT_EQ(x1.size(), x2.size());
  for (size_t i = 0; i < x1.size(); ++i)
    expect_near_rel_finite(x1[i], x2[i]);
}

// line 118
template <typename EigMat1, typename EigMat2, require_all_eigen_t<EigMat1, EigMat2>...>
void expect_near_rel(const std::string& msg, EigMat1&& x1, EigMat2&& x2, double tol = 1e-8) {
  EXPECT_EQ(x1.rows(), x2.rows()) << "expect_near_rel (Eigen::Matrix)"
                                  << " rows must be same size."
                                  << " x1.rows() = " << x1.rows()
                                  << "; x2.rows() = " << x2.rows() << std::endl
                                  << msg << std::endl;
  EXPECT_EQ(x1.cols(), x2.cols())
      << "expect_near_rel:"
      << "cols must be same size."
      << "x1.cols() = " << x1.cols() << "x2.cols() = " << x2.cols() << ")"
      << std::endl
      << msg << std::endl;
  std::string msg2 = "expect_near_rel; require items x1(i) = x2(i): " + msg;
  auto x1_eval = x1.eval();
  auto x2_eval = x2.eval();
  for (int i = 0; i < x1.size(); ++i)
    expect_near_rel(msg2, x1_eval(i), x2_eval(i), tol);
}
```

> Second compiler can not make a conversion of argument and deduce its template at the same time, so this would not work even if we constructed a correct require_.

idt they are simultaneous? template resolution is something like

1. pick all viable signatures candidates (including ones that require implicit conversion)
2. SFINAE the bad ones
3:N(?). Something complex to choose the most similar candidate (and if more than 1 than fail)

There's a chapter on this in the templates book I only skimmed

stan/math/prim/meta/scalar_seq_view.hpp
SteveBronder: Let's make an issue to change the name of return_type and we can call this resolved

stan/math/prim/meta/plain_type.hpp
SteveBronder: I think the docs satisfy this

test/unit/math/opencl/device_functions/log1m_exp_test.cpp
t4c1: Of course. Using explicit eval() works. Done.

andrjohns:
Can also add a to-do to replace the ```inv_logit``` implementation once Eigen 3.4 is released, since they've introduced an [elementwise ```logistic()``` function](https://gitlab.com/libeigen/eigen/commit/a5660744801116ad2f9ab4e9e389f194ba307a35) with AVX support (not sure if it's AVX for doubles or just floats though)

SteveBronder:
@t4c1 fyi I fixed up the rest of expect_near_rel, lemme take one more peek at this and then I think it's good to go!

test/unit/math/expect_near_rel.hpp
```diff
@@ -30,7 +30,7 @@ namespace internal {
  * @param x2 second argument
  * @param tol relative tolerance
  */
-template <typename T1, typename T2>
+template <typename T1, typename T2, require_all_not_eigen_t<T1, T2>...>
```
t4c1: I am refering to this in the next comment.

test/unit/math/expect_near_rel.hpp
```diff
@@ -87,7 +87,7 @@ void expect_near_rel_finite(const std::string& msg, const std::vector<T1>& x1,
  * @param x2 second argument to test
  * @param tol relative tolerance
  */
-template <typename T1, typename T2>
+template <typename T1, typename T2, require_all_stan_scalar_t<T1, T2>...>
```
t4c1: Both of these two functions accept scalars. Why do they have different requires?

test/unit/math/expect_near_rel.hpp
```diff
@@ -139,9 +139,9 @@ void expect_near_rel(const std::string& msg, EigMat1&& x1, EigMat2&& x2,
     expect_near_rel(msg2, x1_eval(i), x2_eval(i), tol);
 }
 
-template <typename T1, typename T2>
-void expect_near_rel(const std::string& msg, const std::vector<T1>& x1,
-                     const std::vector<T2>& x2, double tol = 1e-8) {
+template <typename T1, typename T2, require_all_std_vector_t<T1, T2>...>
+void expect_near_rel(const std::string& msg, const T1& x1,
+                     const T2& x2, double tol = 1e-8) {
```
t4c1: Why do we need this change? Both versions of code ere exactley equivalent as far as I see.

test/unit/math/opencl/kernel_generator/reference_kernels/binary_operation_div_mul_reuse_expression.cl
```diff
@@ -0,0 +1,13 @@
...
+double var1 = 0; if (!((!contains_nonzero(var1_view, LOWER) && j < i) || (!contains_nonzero(var1_view, UPPER) && j > i))) {var1 = var1_global[i + var1_rows * j];}
+double var2 = 0; if (!((!contains_nonzero(var2_view, LOWER) && j < i) || (!contains_nonzero(var2_view, UPPER) && j > i))) {var2 = var2_global[i + var2_rows * j];}
+double var3 = var1 + var2;
+double var4 = var3 * var3;
+double var5 = 0; if (!((!contains_nonzero(var5_view, LOWER) && j < i) || (!contains_nonzero(var5_view, UPPER) && j > i))) {var5 = var5_global[i + var5_rows * j];}
+double var6 = var4 / var5;
+double var7 = var3 / var6;
+double var8 = var7 * var6;
+var9_global[i + var9_rows * j] = var8;
+}
```
t4c1: Please don't push in random stuff from other PRs.

t4c1:
To speed things up I'll just fix those myself.

test/unit/math/expect_near_rel.hpp
SteveBronder: The other one also accepts vectors

test/unit/math/opencl/kernel_generator/reference_kernels/binary_operation_div_mul_reuse_expression.cl
SteveBronder: This was generated because I have the macro on. Shouldn't it have been there already?

test/unit/math/expect_near_rel.hpp
SteveBronder: Oh woah I totally missed the vector version below, idk how that happened. That's why those were different but yeah that could just be scalar

SteveBronder:
@serban-nicusor-toptal it looks like upstream in cmdstan it can't download the latest stanc3?

> curl: (6) Could not resolve host: github.com

serban-nicusor-toptal:
Looking now @SteveBronder

SteveBronder:
Thanks for the quick response! I was just kicking off the cmdstan part of the pipeline so maybe that was it

test/unit/math/opencl/kernel_generator/reference_kernels/binary_operation_div_mul_reuse_expression.cl
t4c1: Nope. I think this was a suggestion of yours, which I changed somewhat before putting it in.

SteveBronder: Oh weird maybe I need to run `make clean`

PR #1476, opened 2019-11-30
Flatten meta tests

rok-cesnovar:
## Summary

This flattens the `prim/*/meta` tests to `prim/meta`, `fwd/*/meta `to` fwd/meta` and `mix/*/meta` to `mix/meta`. 

I just copied all the tests and fixed the headers. The only test I had to modify was prim/meta/is_std_vector_test.cpp which had the following tests from scal:

```cpp
  EXPECT_FALSE((is_std_vector<std::vector<double>>::value));
  EXPECT_FALSE((is_std_vector<const std::vector<double>>::value));
  EXPECT_FALSE((is_std_vector<const std::vector<double>&>::value));
```
In a flattened math these kind of weird stuff ( I understand this is expected behavior for scal, but it feels weird) wont need to exist anymore so I think this is fine?

There are ~a lot~ some duplicated tests because of this flatten. But I am thinking that cleaning that should be a separate PR to make reviewing easier? 

For this PR we can test with a script that there were no lost tests in the process (list all test names on develop and this PR and do a match). I am really glad we did the unique test names thing. It makes this easier. The next PR where we remove duplicate will then need a proper review.

Any thoughts @syclik ?

## Tests

All of this is tests.

## Side Effects

Hopefully not.

## Checklist

- [x] Math issue #937 


rok-cesnovar:
You can list all tests to a file by adding the following on line 63 in runChecks.py
```python
    f = open("test_names.txt","w")
    for a in tests:
        f.write(str(a) + "\n")
    f.close()
```
and then just `python runChecks.py` on develop and this branch and change the filename in between.

You can check the diff here: https://www.diffchecker.com/X8udhCxD

The only difference is removal of `MathMetaPrimArr, is_std_vector` and `MathMetaPrimScal, is_std_vector` in favor of `MathMetaPrim, is_std_vector` which is due to what I mention above in the top-level post.

If I get green light on this we can then either: 
- flatten all meta tests in this PR
- merge this PR and cleanup the prim/meta tests first

rok-cesnovar:
fwd, mix and rev have very little meta tests so I changed this to flatten all meta tests, not just prim.

syclik:
@rok-cesnovar this PR is good to go. 

To answer your question:
> If I get green light on this we can then either:
> 
> * flatten all meta tests in this PR
> * merge this PR and cleanup the prim/meta tests first

I'd rather take the second option. I already looked at a diff of 128 files. I don't think we need to just pile on more diffs in single PRs. It's easy enough to break it apart and look at smaller diffs.

These reviews are straightforward, but they require a little bit of attention. I tried to do it on my phone yesterday and failed horribly, which is why it took another day.

And... thank you for putting in enough detail where you guided me through the couple of cases where things weren't straighforward. That helps immensely. If I stumbled on it myself, I would have asked and we would have had to answer those questions... instead it was just clear that you had already thought about it, you had a plan for it, there was clear justification, and it wasn't a mistake. Thank you.

rok-cesnovar:
Thank you for the review! The bulk of the work in the flattening PRs is going to fall on the reviewer. So its only fair to help out.

 I will request changes in the only open PR that deals with meta tests and then we can merge.

And no worries on taking your time. No rush here, the focus is on getting it right.

PR #1469, opened 2019-11-29
Increase tolerances for add_vec_mat

rok-cesnovar:
# Summary

We are currently experiencing issues on `develop`. @bob-carpenter [suggested](https://github.com/stan-dev/math/issues/1453) to increase the tolerances for the particular test that fails randomly. So this PR does just that. I am starting with 5E-3 and if that will still cause issues we can increase that to 1E-2.

## Tests

/

## Side Effects

/

## Checklist

- [x] Math issue #1453

- [x] Copyright holder: Rok Češnovar

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

Issue #1453, opened 2019-11-11
MathMixMatFun.add and MathMixMatFun.subtract tests seem to fail randomly?


SteveBronder:
## Description

After #1424 we've had a couple PRs which look fine fail the MPI unit tests (but pass the GPU branch?)

Ex:

#1445 #1446 #1429 

Not sure what's going on?

If we can find a quick patch for #1424 or understand why that's happening then cool and good otherwise let's do the revert and revert-revert PR

#### Current Version:
v3.0.0

SteveBronder:
Also the tests pass locally for me. Not sure if they are setting something off in the MPI that means we need a flag somewhere

bob-carpenter:
I don't understand what's being proposed here.  Did you want to revert #1424 for some reason?  I couldn't tell from #1445 what's going wrong.

SteveBronder:
I'm looking into what's going on right now. Not sure if it's some weird interaction between something and MPI since the tests on the GPU node pass fine

SteveBronder:
Alright idk y but breaking them up into individual tests fixed it for one of the PRs. Gonna close this. @rok-cesnovar if you have time / think the matrix_cl<var> PR can get in tmrw we can just leave the fix in there otherwise I can slam out a quick PR that just has those fixed up

bob-carpenter:
There are still a couple residual MPI-based tests that I didn't remove that do some setup.  One is:

test/unit/math/rev/mat/fun/multiply_test.cpp

Do the other tests need similar config?  I don't know how tests/MPI are calling everything from the outside.

SteveBronder:
I'm not sure either, something something curing the symptom and not the wound but we can just deal with those if they pop up as issues

rok-cesnovar:
Develop has just now failed with a similar error (see https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/develop/404/pipeline/142)

I will look into this tonight.

bob-carpenter:
The failure is this:

```
[ RUN      ] MathMixMatFun.add_vec_mat
./test/unit/math/expect_near_rel.hpp:39: Failure
The difference between x1 and x2 is 0.0010334642265417964, which exceeds tol, where
x1 evaluates to -0.0010334642265417964,
x2 evaluates to 0, and
tol evaluates to 0.001.
[...]

PR #1483, opened 2019-12-04
Increase tolerances in substract_vec_mat test

rok-cesnovar:
## Summary

We are still experiencing some issues on develop after https://github.com/stan-dev/math/pull/1469 was merged. This increases the tolerances for another test. If new test failures come up then I would say we should increase the tolerance globally. I will also investigate what causes the "randomness" of this tests that does not use any randomness whatsoever. Might be limited to Windows.

## Tests

This increases the tolerances in the tests.

## Side Effects

/

## Checklist

- [x] Math issue #1453 


bob-carpenter:
I'm guessing it's a different compiler or settings or different hardware.

rok-cesnovar:
This is ready for review. I will examine the strange windows behavior in tests in a separate issue/PR. This should be merged to prevent Jenkins issues for other developers.

Issue #1484, opened 2019-12-05
doxygen is broken on Mac OS X after latest upgrade


bob-carpenter:
## Description

Neither @mitzimorris nor I can get the new doxygen updates working on our Macs.  We've spent a couple hours on it and the formatting changes that were made in []() don't seem to be worth this effort.

One fix would be to revert the PR that made the changes as they're minor formatting things.  I personally don't think they're worth the trouble I've already had with this.

## Example

```
~/cmdstan/stan/lib/stan_math(feature/0123-complex-var)$ make doxygen
mkdir -p doc/api
doxygen doxygen/doxygen.cfg
warning: the dot tool could not be found at /usr/bin/dot
```

## More info

The problem originated in  https://github.com/stan-dev/math/pull/1397, which is why I assigned it to @SteveBronder.

Here's my doxygen setup:

```
~/cmdstan/stan/lib/stan_math(feature/0123-complex-var)$ which doxygen
/usr/local/bin/doxygen
~/cmdstan/stan/lib/stan_math(feature/0123-complex-var)$ doxygen --version
1.8.16
```

And here's `dot`:

```
~/cmdstan/stan/lib/stan_math(feature/0123-complex-var)$ which dot
/usr/local/bin/dot
```

The relevant config in `doxygen/doxygen.cfg` is:

```
# The default value is: YES.
HAVE_DOT               = NO
...
DOT_PATH               = /usr/bin/dot
```

Now if I change the `DOT_PATH` to the path to my version of `dot`, I get this:

```
~/cmdstan/stan/lib/stan_math(feature/0123-complex-var)$ make doxygen
mkdir -p doc/api
doxygen doxygen/doxygen.cfg

```

returning instantly and only the boilerplate doc is generated.

#### Current Version:
v3.0.0

SteveBronder:
Can you try commenting out the `DOT_PATH`? i.e.

# DOT_PATH = blah

That should remove the dot path thing

From there is there no index.html in doc/api/html ? When I build it on my work mac and open the index.hpp I can do the search and all of the other stuff? Are you on develop or on the complex branch?

SteveBronder:
Also what version of doxygen do you have? I do need to sit down and see what's going on with `
```
stan/math/opencl/copy.hpp:218: warning: argument 'src' from the argument list of stan::math::copy_cl has multiple @param documentation sections
```

mitzimorris:
[...]

PR #1487, opened 2019-12-05
Bugfix/1484 doxygen cfg

mitzimorris:
Thanks for submitting a pull request! Please remove this text when submitting.

## Summary

Remove unneeded doxygen config file directives for EXCLUDE paths and patterns because:

- the source directory `stan` doesn't contain any such paths and patterns
- on Mac OSX, these patterns seem to match everything (?) - when present, the stan source files aren't processed at all and the result is a documentation page which only information found in top-level markdown file and files in the doxygen directory.

## Tests

this only affects the API docs

## Side Effects

?

## Checklist

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:  Columbia University


mitzimorris:
more comments here: https://github.com/stan-dev/math/issues/1484

doxygen/doxygen.cfg
```diff
@@ -2353,7 +2350,7 @@ INTERACTIVE_SVG        = YES
 # found. If left blank, it is assumed the dot tool can be found in the path.
 # This tag requires that the tag HAVE_DOT is set to YES.
 
-DOT_PATH               = /usr/bin/dot
+DOT_PATH               =
```
SteveBronder: Here it may be good to comment out the dot path so it knows it's not an option we want to set 

```suggestion
#DOT_PATH               =
```

mitzimorris:
commented out #DOT_PATH and it builds; pushed change.


note that doxygen is still giving this warning:


```

/Users/mitzi/github/stan-dev/cmdstan/stan/lib/stan_math/stan/math/opencl/copy.hpp:184:
warning: argument 'src' from the argument list of stan::math::copy_cl has
multiple @param documentation sections

```


and everything shows up twice in the docs:


[image: Untitled.png]

mitzimorris:
pattern-matching like a monkey - this makes error go away - but is it
correct?

```

/** \ingroup opencl
 * Copies the source matrix to the
 * destination matrix. Both matrices
 * are stored on the OpenCL device.
 *
 * @tparam T An arithmetic type to pass the value from the OpenCL matrix to.
 * @param src the source matrix
 * @return matrix_cl with copies of values in the source matrix
 * @throw <code>std::invalid_argument</code> if the
 * matrices do not have matching dimensions
 */
template <typename T, typename = require_arithmetic_t<T>>
inline matrix_cl<T> copy_cl(const matrix_cl<T>&& src) {
  return matrix_cl<T>(std::forward<T>(src));
}
```
[image: Untitled.png]

SteveBronder:
Replacing copy_cl's docs in both `opencl/copy.hpp` and `opencl/rev/copy.hpp` with the above

```cpp
/** \ingroup opencl
 * Copies the source matrix to the
 * destination matrix. Both matrices
 * are stored on the OpenCL device.
 *
 * @tparam T An arithmetic type to pass the value from the OpenCL matrix to.
 * @param src the source matrix
 * @return matrix_cl with copies of values in the source matrix
 * @throw <code>std::invalid_argument</code> if the
 * matrices do not have matching dimensions
 */
```

did the trick for me!

mitzimorris:
OK - do you want that doc fix to go in with this PR?

SteveBronder:
Yeah I think its appropriate if you dont mind. Then no warnings on doxygen

mitzimorris:
ready for re-review and re-approval.

PR #1491, opened 2019-12-07
Interchange loops for divide and multiply

peterwicksstringfield:
## Summary

Partially addresses #1490. If I understand correctly, Eigen matrices are column major and these loops can be trivially interchanged. The compiler probably does it anyway.

## Tests

Multiply tests:
test/unit/math/prim/mat/fun/multiply_test.c(Same tests as in #XXX.pp
test/unit/math/mix/mat/fun/multiply_test.cpp

Divide tests:
test/unit/math/prim/mat/fun/divide_test.cpp
test/unit/math/mix/mat/fun/divide_test.cpp

Divide had no real primitive mode tests, so I added some.

## Side Effects

None.

## Checklist

- [X] Math issue #1490

- [X] Copyright holder: Peter Wicks Stringfield

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [X] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [X] the code is written in idiomatic C++ and changes are documented in the doxygen

- [X] the new changes are tested

PR #1492, opened 2019-12-07
Feature/issue 988 matrix divide

### LINKED ISSUE #988: prim/mat/fun implementation of divide is too specialized
## Description
One argument is templated and one is not. It always returns an Eigen::Matrix with double scalars.

See: https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/fun/divide.hpp#L22

It's probably compiling fine because the other necessary specializations exist to take over, but I do not think this is properly templated.

Either both arguments should be templated (if this is expected to handle some of the high order stuff) or both the arguments should be doubles (if the other implementations cover all the necessary templated types)

## Current Math Version
v2.18.0

peterwicksstringfield:
## Summary

Analytic derivatives for M/c where M is matrix T1 and c is scalar T2. 

(The discussion in #988 is a bit out of date, the templating on the prim version was cleaned up by 3b459c9553; so the promotion to var is the only issue mentioned in that ticket that remained to be addressed. Closes #988.)

The key bit of math:
```
...
class matrix_scalar_divide_vari_vv : public vari {
  ...
  virtual void chain() {
    ...
    adjC_->adj_ += -1.0 * invc
                   * (adjResult.adj().array() * adjResult.val().array()).sum();
    adjM.adj() += invc * adjResult.adj();
...
```
f[M, c] = M/c
f_c[M, c] = -M/c^2 = (-1/c) * f[M, c].
f_M[M, c] = 1/c

I wanted to say
```
TEST(MathMixMatFun, divide) {
    ...
    Eigen::VectorXd u(4);
    u << nan, 0, -3, 4;
    stan::test::expect_ad(f, u, 2.0);
...
```

But of course, nans propagate through the analytic derivatives and the finite differences differently, so this fails. Adding check_finite to every implementation of divide would ensure consistency, but as I understand, there's a thread on the mailing list for that discussion. And that would probably call for a performance test, since division is supposed to be kind of free. So I didn't do that here. (But its easy enough to add those checks if desired.)

In particular:
```
f[x,y] = x/y
f_x[x,y] = 1/y
f_x[nan,1] = 1 (analytic/autodiff answer)
f_x[nan,1] ~ (1/n) (f[nan+1/n,1] - f[nan,1])
           = (1/n) (nan - nan)
           = nan (finite diff answer)
```

## Tests

This function had a nice set of expect_ad tests already. There were no real primitive mode tests, so I added some. (Same tests as in #1491.)

## Side Effects

None.

## Checklist

- [X] Math issue #988

### DIFF
--- a/stan/math/rev/mat/fun/divide.hpp
+++ b/stan/math/rev/mat/fun/divide.hpp
@@ -28,9 +28,8 @@ class matrix_scalar_divide_dv_vari : public vari {
         adjResultRef_(ChainableStack::instance_->memalloc_.alloc_array<vari*>(
             m.rows() * m.cols())),
         invc_(1.0 / c.val()) {
-    Eigen::Matrix<double, R, C> result = invc_ * m;
     Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
-        = result.unaryExpr([](double x) { return new vari(x, false); });
+        = (invc_ * m).unaryExpr([](double x) { return new vari(x, false); });
   }
 
   virtual void chain() {
@@ -60,9 +59,10 @@ class matrix_scalar_divide_vd_vari : public vari {
             m.rows() * m.cols())),
         invc_(1.0 / c) {
     Eigen::Map<matrix_vi>(adjMRef_, rows_, cols_) = m.vi();
-    Eigen::Matrix<double, R, C> result = invc_ * m.val();
     Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
-        = result.unaryExpr([](double x) { return new vari(x, false); });
+        = (invc_ * m.val()).unaryExpr([](double x) {
+            return new vari(x, false);
+          });
   }
 
   virtual void chain() {
@@ -94,9 +94,10 @@ class matrix_scalar_divide_vv_vari : public vari {
             m.rows() * m.cols())),
         invc_(1.0 / c.val()) {
     Eigen::Map<matrix_vi>(adjMRef_, rows_, cols_) = m.vi();
-    Eigen::Matrix<double, R, C> result = invc_ * m.val();
     Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
-        = result.unaryExpr([](double x) { return new vari(x, false); });
+        = (invc_ * m.val()).unaryExpr([](double x) {
+            return new vari(x, false);
+          });
   }
 
   virtual void chain() {
--- a/stan/math/rev/mat/fun/divide.hpp
+++ b/stan/math/rev/mat/fun/divide.hpp
@@ -10,11 +10,14 @@ namespace stan {
 namespace math {
 
 /**
- * Return the division of the specified column vector by
- * the specified scalar.
- * @param[in] v Specified vector.
- * @param[in] c Specified scalar.
- * @return Vector divided by the scalar.
+ * Return matrix divided by scalar.
+ * @tparam T1 TODO deleting this anyway
+ * @tparam T2 TODO deleting this anyway
+ * @tparam R number of rows or Eigen::Dynamic
+ * @tparam C number of columns or Eigen::Dynamic
+ * @param[in] m specified matrix
+ * @param[in] c specified scalar
+ * @return matrix divided by the scalar
  */
 template <typename T1, typename T2, int R, int C,
           typename = require_any_var_t<T1, T2>>
@@ -22,7 +25,6 @@ inline Eigen::Matrix<var, R, C> divide(const Eigen::Matrix<T1, R, C>& v,
                                        const T2& c) {
   return to_var(v) / to_var(c);
 }
-
 }  // namespace math
 }  // namespace stan
 #endif
--- a/stan/math/rev/mat/fun/divide.hpp
+++ b/stan/math/rev/mat/fun/divide.hpp
@@ -9,22 +9,167 @@
 namespace stan {
 namespace math {
 
+namespace internal {
+template <int R, int C>
+class matrix_scalar_divide_vari_dv : public vari {
+ public:
+  int rows_;
+  int cols_;
+  vari* adjCRef_;
+  vari** adjResultRef_;
+  double invc;
+
+  explicit matrix_scalar_divide_vari_dv(const Eigen::Matrix<double, R, C>& m,
+                                        const var& c)
+      : vari(0.0),
+        rows_(m.rows()),
+        cols_(m.cols()),
+        adjCRef_(c.vi_),
+        adjResultRef_(ChainableStack::instance_->memalloc_.alloc_array<vari*>(
+            m.rows() * m.cols())) {
+    double c_d = c.val();
+    invc = 1.0 / c_d;
+    Eigen::Matrix<double, R, C> result = invc * m;
+    Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
+        = result.unaryExpr([](double x) { return new vari(x, false); });
+  }
+
+  virtual void chain() {
+    Eigen::Map<matrix_vi> adjResult(adjResultRef_, rows_, cols_);
+    adjCRef_->adj_
+        += -1.0 * invc
+           * (adjResult.adj().array() * adjResult.val().array()).sum();
+  }
+};
+
+template <int R, int C>
+class matrix_scalar_divide_vari_vd : public vari {
+ public:
+  int rows_;
+  int cols_;
+  vari** adjMRef_;
+  vari** adjResultRef_;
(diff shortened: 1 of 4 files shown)


stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
 namespace stan {
 namespace math {
 
+namespace internal {
+template <int R, int C>
+class matrix_scalar_divide_vari_dv : public vari {
```
bob-carpenter: [optional]

I believe the convention we've used elsewhere is to have the `vari` extensions end in `vari`, so I'd suggst `_dv_vari` for the name and similarly elsewhere.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
+ public:
+  int rows_;
+  int cols_;
+  vari* adjCRef_;
+  vari** adjResultRef_;
+  double invc;
+
+  explicit matrix_scalar_divide_vari_dv(const Eigen::Matrix<double, R, C>& m,
+                                        const var& c)
+      : vari(0.0),
```
bob-carpenter: [optional]
`vari(0)` is fine---we don't need the `.0` everywhere. The only place it should be necessary is when you don't want integer rounding when dividing two integers.  If it's necessary elsewhere, that's a bug that should be reported.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
 
+namespace internal {
+template <int R, int C>
+class matrix_scalar_divide_vari_dv : public vari {
+ public:
+  int rows_;
+  int cols_;
+  vari* adjCRef_;
+  vari** adjResultRef_;
+  double invc;
```
bob-carpenter: All member variables should end in `_`.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
+                                        const var& c)
+      : vari(0.0),
+        rows_(m.rows()),
+        cols_(m.cols()),
+        adjCRef_(c.vi_),
+        adjResultRef_(ChainableStack::instance_->memalloc_.alloc_array<vari*>(
+            m.rows() * m.cols())) {
+    double c_d = c.val();
+    invc = 1.0 / c_d;
+    Eigen::Matrix<double, R, C> result = invc * m;
```
bob-carpenter: [no action required]
This is a neat way to code this.  I'm guessing the matrix operation will more than make up for constructing the various matrices and maps and functors.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
+    Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
+        = result.unaryExpr([](double x) { return new vari(x, false); });
+  }
+
+  virtual void chain() {
+    Eigen::Map<matrix_vi> adjResult(adjResultRef_, rows_, cols_);
+    adjCRef_->adj_
+        += -1.0 * invc
```
bob-carpenter: Replace this line with `-= invc` --- negation is preferred over multiplying by `-1` and it can be built straight into the operator.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
+};
+
+template <int R, int C>
+class matrix_scalar_divide_vari_vd : public vari {
+ public:
+  int rows_;
+  int cols_;
+  vari** adjMRef_;
+  vari** adjResultRef_;
+  double invc;
```
bob-carpenter: Same request here to use `invc_` and an initializer.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
+
+template <int R, int C>
+class matrix_scalar_divide_vari_vv : public vari {
+ public:
+  int rows_;
+  int cols_;
+  vari** adjMRef_;
+  vari* adjC_;
+  vari** adjResultRef_;
+  double invc;
```
bob-carpenter: same again.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,165 @@
...
+    invc = 1.0 / c_d;
+    Eigen::Matrix<double, R, C> result = invc * m.val();
+    Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
+        = result.unaryExpr([](double x) { return new vari(x, false); });
+  }
+
+  virtual void chain() {
+    Eigen::Map<matrix_vi> adjM(adjMRef_, rows_, cols_);
+    Eigen::Map<matrix_vi> adjResult(adjResultRef_, rows_, cols_);
+    adjC_->adj_ += -1.0 * invc
```
bob-carpenter: same comment again here.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: You're right, changed for consistency.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: vari(0) it is then; thanks.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: Fixed.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: Thanks, not sure why I wrote it with += to begin with.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: Fixed.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: Fixed.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: Fixed underscore and moved invc_'s assignment to initializer.

bob-carpenter:
> +    adjCRef_->adj_
> +        += -1.0 * invc
> 
> Thanks, not sure why I wrote it with += to begin with.

That's the idiom---result adjoint times partial gets added to operand adjoint.  It's just not always the most efficient implementation.

stan/math/rev/mat/fun/divide.hpp
peterwicksstringfield: Thanks. I adapted the pattern from the matrix-matrix multiply in multiply.hpp.

bob-carpenter:
Thanks!  Is this ready to review again?

peterwicksstringfield:
I think its ready to review again.

Also, I wanted to thank you for the feedback, on the this and a couple other PRs.

bob-carpenter:
You're welcome!  I prioritize feedback on PRs from new contributors.

stan/math/rev/mat/fun/divide.hpp
```diff
@@ -9,18 +9,162 @@
...
+      : vari(0),
+        rows_(m.rows()),
+        cols_(m.cols()),
+        adjCRef_(c.vi_),
+        adjResultRef_(ChainableStack::instance_->memalloc_.alloc_array<vari*>(
+            m.rows() * m.cols())),
+        invc_(1.0 / c.val()) {
+    Eigen::Matrix<double, R, C> result = invc_ * m;
+    Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
+        = result.unaryExpr([](double x) { return new vari(x, false); });
```
andrjohns: Super minor thing that I noticed, but you could also write this as:
```
Eigen::Map<matrix_vi>(adjResultRef_, rows_, cols_)
= (invc_ * m).unaryExpr([](double x) { return new vari(x, false); })
```
To avoid the creation of the ```result``` temporary

(Not a review, feel free to ignore)

bob-carpenter: @andrjohns:  Thanks --- that sounds like a good idea.

@peterwicksstringfield :  This is optional---we have a general policy of not double reviewing or unapproving unless absolutely necessary for a bug.  Let me know if you'd like to make this update or just merge.

peterwicksstringfield: Thanks, that transformation avoids collapsing an expression template.

Issue #1496, opened 2019-12-09
The way neg_binomial_2_lpmf delegates to Poisson is broken


martinmodrak:
## Description
The `neg_binomial_2_lpmf` delegates to `poisson_lpmf` for large values of the `phi` parameter. Currently the cutoff is set at `phi > 1e5`. This introduces discontinuities into the function - I am nut sure whether this is because the cutoff is too optimistic or too pessimistic.

## Example
```
neg_binomial_2_lpmf(181, 13, 1e5 - 1e-8) == -312.05065787932836
neg_binomial_2_lpmf(181, 13, 1e5 + 1e-8) == -312.19077641210203

neg_binomial_2_lpmf(121358, 13, 1e5 - 1e-8) == -933542.7362432084
neg_binomial_2_lpmf(121358, 13, 1e5 + 1e-8) == -988062.09132748179
```

I've written a failing test for the cutoff at: https://github.com/stan-dev/math/compare/develop...martinmodrak:bugfix/1496-poisson-phi-cutoff

## Expected Output
Values at `cutoff - 1e-8` and `cutoff + 1e-8` are almost constant, i.e. changes in `phi` no longer influence the lpmf value and delegating to Poisson is safe.

#### Current Version:
v3.0.0

martinmodrak:
It gets even more funny - the cutoff is implemented as 
```
    // if phi is large we probably overflow, defer to Poisson:
    if (phi__[i] > 1e5) {
      logp = poisson_lpmf(n_vec[i], mu__[i]);
    }
```
Note the `=`!!! So when a single element of a vector provokes cutoff, it _overwrites_ the `logp` accumulator. Also, the way `poisson_lpmf` is called means that the `propto` information cannot be conserved and thus the final `logp` will wildly change around the cutoff under certain `propto` combinations. On the same branch I wrote a failing test for this behavior as well.

bob-carpenter:
> It gets even more funny

That's just a flat-out bug.  The conditional should be at the top of the loop and have the rest of the body of the loop in the alternative branch.

Thanks for tracking this down!

bob-carpenter:
> On Dec 9, 2019, at 12:42 PM, Martin Modrák <notifications@github.com> wrote:
> 
> It gets even more funny - the cutoff is implemented as
> 
>     // if phi is large we probably overflow, defer to Poisson:
>     if (phi__[i] > 1e5) {
>       logp = poisson_lpmf(n_vec[i], mu__[i]);
>     }
[...]

Issue #1495, opened 2019-12-09
neg_binomial_2_log_lpmf is not stable for large phi


martinmodrak:
## Description
For some values, `neg_binomial_2_log_lpmf` would return values larger than 0 (e.g. `neg_binomial_2_log_lpmf(39, 4, 6e16) == 256`) for other, it just returns noticeably different values than the corresponding `neg_binomial_2_lpmf` call. In all cases this is related to large `phi` values (noticeable discrepancies manifest already at around `phi = 2e5`).

This seems to be related to/a regression of #428

## Example
`neg_binomial_2_log_lpmf(39, 4, 6e16) == 256`.
I've created a failing test on the branch: https://github.com/stan-dev/math/compare/develop...martinmodrak:bugfix/1495-neg_binomial_2_log_large_phi

## Expected Output
The same as `neg_binomial_2_lpmf(39, exp(4), 6e16)`, which is  `-5.22991`

#### Current Version:
v3.0.0

martinmodrak:
#1496 seems related as the phi cutoff for when to delegate to Poisson is probably not correctly chosen.

jtimonen:
Is it possible to pin down the part of `math/stan/math/prim/prob/neg_binomial_2_lpmf.hpp` where the numerical problems occur with large phi? My guess is that it's in computing the binomial coefficient, which you seem to be doing by computing the logarithms of the factorials using `lgamma`. Can the problem be just numerical inaccuracy of `lgamma` with large input? Are you using Stirling's approximation or something to compute the logarithm of the gamma function?

jtimonen:
Here is my idea:

<img width="742" alt="nb_stirling" src="https://user-images.githubusercontent.com/18327296/73704814-f1c2cd80-46fc-11ea-87de-a56964c25e20.png">

martinmodrak:
You are correct my work with the non-log version in #1497 shows that the Poisson branch can be eliminated with a better implementation of `binomial_coefficient_log` - this better implementation is almost ready in #1614 and so all this should bubble up eventually also here. Note that my experience with reimplementing `lbeta` is that Stirling approximation is not enough in and of itself and corrections are needed (see #1612) to maintain precision when `phi` is large and `x` is small.

wds15:
[...]

PR #1255, opened 2019-05-30
Bugfix/issue 1250 lgamma

wds15:
## Summary

The `std::lgamma` is not reentrant safe and has been shown to avoid performance scalability in **threaded** stan programs using `map_rect`-type  threading on macOS. This PR replaces this with a call to the reentrant safe variants with one of these two:

- `lgamma_r` the platform is not `__MINGW32__`
- `boost::math::lgamma` on MINGW32

Both versions are reentrant safe. The `lgamma_r` is a C library call which is only available upon defining `-D_REENTRANT`, but this is not working on Windows with MinGW32. This is why the boost version is used on Windows. However, the boost lgamma implementation currently uses as default policy to propagate its input arguments from double to long double. This leads a serious performance regression as the boost lgamma is unacceptably slow in comparison to the `std::lgamma` (see below). Changing the default policy of boost math functions to avoid the propagation to long double leads to speed of the same order as `std::lgamma`, see below.

The proposed log gamma implementations solve the performance problems seen with threading under macOS.

## Tests

The makefile target `test-math-dependencies` has been extended to check in `stan/math` for use of `std::lgamma`. In the commit 6521f3f the reviewer can verify that a

`make test-math-dependencies`

will rightfully detect the use of `std::lgamma` which is put in this commit for testing purposes into the source.

The boost math lgamma implementation does overflow for arguments greater than 1E25 whenever the double propagation to long double is disabled. This is a bug in boost 1.69 and will be fixed past boost 1.70.0. This PR includes the fix from the boost math sources (https://github.com/boostorg/math/commit/fd07b121a619e0429e189edb4ff50f5b4e9efe58). In order to check that the boost version is fixed a test is included in

`test/unit/math/prim/scal/fun/lgamma_test.cpp`

to check for large arguments to return a finite value. The bug appeared for arguments greater than 1E25 (around the sqrt of the largest double of 1E308). Other functions from boost math which use the `stan::math::boost_policy_t` (see below) are also tested to work for large arguments.

### Change to the default `boost_policy_t`

We change the default boost_policy_t within stan math to avoid the double to long double propagation (this being considered to be the new default in future boost versions due to the performance regression caused, see https://github.com/boostorg/math/issues/241). As a side effect all functions from boost math which use this policy are affected. These are:

```
stan/math/prim/scal/fun/rising_factorial.hpp
stan/math/prim/scal/fun/falling_factorial.hpp
stan/math/prim/scal/fun/digamma.hpp
[...]

PR #1497, opened 2019-12-09
Fixing negative binomial phi cutoff

martinmodrak:
## Summary

This introduces more tests on the behavior of negative binomial 2 distribution when `phi` is large and will try to fix issues discovered along the way. Currently:

* improved numerical stability of the density by using `binomial_coefficient_log` and rearranging some operations
* improved numerical stability of the derivative w.r.t `phi`
* this obviated the necessity to delegate to Poisson density for large phi and the whole code branch was removed
* addressed potential size mismatch noticed in #1531 
* renamed local variables with double underscores in names

This PR requires the improvements from PR #1614 to be merged before it is merged.

## Tests

New tests:
- Created a set of test values and derivatives in Mathematica (Wolfram Cloud) and test against those
- Changed the finite differences test for derivatives to cover more of the parameter space and use complex step differentiation instead of finite diffs. This test requires somewhat looser tolerance for derivative wrt. `phi` for `n > 100000` and large `phi`. Importantly, the sign of the derivative does not match (e.g. expected = -3.24e-10, actual = 9e-12), not sure if the complex step or the function is to blame, so I assume it is OK for now.
- Extended the test for extreme `phi` values to cover more of the parameter space
- Tests of values and derivatives against analytical solutions for `n = 0` and `n = 1` (this allows me to test even `mu` and `phi` values where Mathematica will not calculate exact solutions).

## Side Effects

- The way the density for `neg_binomial_2_lpmf` is computed changed to improve stability, at the cost of tangling `mu` and `phi` together earlier, so less computation can be avoided when `propto=true` and `phi` is fixed.
- Generally I would expect a very minor speed drop for some parameter values (when previously we delegated to poisson which is probably cheaper).

## Checklist

- [x] Math issue #1496


martinmodrak:
So, fixing the phi cutoff triggered some failures in the finite differences test. I tried to fix those and wrote more tests to make sure I am on the right track. Those tests revealed other issues. I failed to fix all of them so far.

In particular, there are numerical differences between the values we compute and what I get from Mathematica and the finite differences test does not pass.

bob-carpenter:
How far off is the finite diff test?  Let me know if I can help somehow with the testing.

martinmodrak:
Will continue the discussion here and not at the issue (hope that's OK). 

There are currently two failing tests a `precomputed` test, using numbers from Mathematica and the original `finiteDiffs` test, which I expanded to test over wider range of values (but it fails even for the original values tested).

The numbers I get from Mathematica for the `precomputed` test are computed in this notebook:  https://www.wolframcloud.com/obj/martin.modrak/Published/NegBinomial2_Tests.nb - I am still a Mathematica beginner, so hope I made no silly mistake, but since I am using what they call "Exact numbers" the outputs I get should allow arbitrary precision (I instruct it to work with 40 digits, then convert to what they call "CForm" which - hopefully - stores as much precision as C would accept, the number-to-string formatting in Mathematica is complicated). The `precomputed` test fails for some with relative errors up to 1e-5, all happening for either small `mu` or small `phi` here is a sample of the test output:

```
test/unit/math/rev/scal/prob/neg_binomial_2_test.cpp:84: Failure
The difference between gradients[1] and t.grad_phi is 1.7763568394002505e-015, which exceeds fabs(t.grad_phi * 1e-8), where
gradients[1] evaluates to -2.6984512402350447e-009,
t.grad_phi evaluates to -2.6984530165918841e-009, and
fabs(t.grad_phi * 1e-8) evaluates to 2.6984530165918841e-017.
grad_phi n = 14, mu = 1.5, phi = 162345

test/unit/math/rev/scal/prob/neg_binomial_2_test.cpp:82: Failure
The difference between gradients[0] and t.grad_mu is 4.0061531662412956e-017, which exceeds fabs(t.grad_mu * 1e-8), where
gradients[0] evaluates to -1.2600009924312872e-009,
t.grad_mu evaluates to -1.2600010324928188e-009, and
fabs(t.grad_mu * 1e-8) evaluates to 1.2600010324928188e-017.
grad_mu n = 10233, mu = 10586, phi = 0.00040000000000000002
```

The `finiteDiffs` test on the other hand fails with many values being off by 1 or more. I added one of the failing values (`n=7, mu = 8, phi = 150000`) to the `precomputed` test and there it passes for `mu` but doesn't pass for `phi`, but the error for `phi` is at the 8th digit, so I am starting to think this might be a problem with the finite diff. 

Here is the relevant error output from the `precomputed` test:

```
gradients[1] evaluates to 1.333262389380252e-010,
t.grad_phi evaluates to   1.333280152948646e-010
```

And here is from the `finiteDiffs` test:

```
gradients[0] evaluates to -0.12499333368886989,
finite_diffs[0] evaluates to -1.4533085845869209, and

gradients[1] evaluates to 1.333262389380252e-010,
finite_diffs[1] evaluates to 1.1641532182693481, and
```

Strangely, the test passed before my code changes. So not sure what is going on in there.

The full test output is seen in the failed Jenkins build.

martinmodrak:
Also, I've run out of both ideas and time to improve the actual computation (if the test failures above are convincing enough to show that it needs improvement). Will be able to work on this more after New Year (unless I'll need very hard to procrastrinate on the stuff I should be doing :-) ).

martinmodrak:
OK, obviously, I can't stop procrastrinating :-) For the differences in derivative wrt. `phi` at least part of the problem is different results of `digamma` between boost, R and Mathematica. R code to show the differences:

```
n <- 7
mu <- 0.0000256
phi <- 15324.0

digamma_phi_boost <-        9.6371428769091523 
digamma_phi_mathematica <-  9.6371428769091527082 
digamma_n_phi_boost <-        9.6375995872973039
digamma_n_phi_mathematica <-  9.6375995872973037537

base <- 1.0 - (n + phi) / (mu + phi) + log(phi) - log(mu + phi)

#Expected value (by mathematica)
-8.9402263370175206e-008
#Computed by Stan math
-8.9402261593818366e-008

base - digamma_phi_boost + digamma_n_phi_boost #Matches Stan
base - digamma(phi) + digamma(n + phi) #Matches mathematica 
base - digamma_phi_mathematica + digamma_n_phi_mathematica #Matches mathematica 
```

martinmodrak:
So it turns out I was chasing very small errors. Taking absolute tolerance into account (i.e. making the tolerance to be `max(true_value * 1e-8, 1e-14)`) made the tests pass for almost all values. Slightly rearranging the computation of the density removed the remaining test failures, at the cost of being able to save less computation when `propto=true` and `phi` is constant.

rok-cesnovar:
@martinmodrak one of the PR merges caused a conflict so I went ahead and fixed it as it was simple.

martinmodrak:
@bob-carpenter I changed the finite differences test to a test against complex-step differentiation (with a hack to get `lgamma(std::complex<double>)`). The test now passes with a relative tolerance of `1e-5`. I think the PR is ready for review and merge.

@rok-cesnovar Thanks! Sorry, the PR was in a messy state overnight, hope that didn't cause trouble.

rok-cesnovar:
There was no trouble. I just saw the conflict that was caused by a merge I did earlier and tried to fix it.

rok-cesnovar:
@martinmodrak mind tagging someone for review here? This would be great to get in for the next release.

martinmodrak:
@rok-cesnovar  I tagged @bob-carpenter before Christmas, so I assume he's busy. I am not sure who else can/should review this, I didn't find a list of possible reviewers.

rok-cesnovar:
Ok, we are in the process of removing the scal/mat/arr folder so there are bound to be some conflicts in the next few days. I can fix those as it will mostly be simple stuff like moving files, fixing includes.

stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -79,33 +84,60 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
+               / (2 * square(phi__[i]));
+      }
+
+    } else {
+      if (include_summand<propto, T_precision>::value) {
+        logp += binomial_coefficient_log(n_plus_phi[i] - 1, n_vec[i]);
+      }
+
+      logp += phi__[i] * (log_phi[i] - log_mu_plus_phi[i])
+              - (n_vec[i]) * log_mu_plus_phi[i];
```
bob-carpenter: This doesn't need parens around `n_vec[i]`.

test/unit/math/prim/prob/neg_binomial_2_test.cpp
```diff
@@ -239,16 +239,64 @@ TEST(ProbDistributionsNegBinomial, chiSquareGoodnessFitTest4) {
...
 
 TEST(ProbDistributionsNegBinomial, extreme_values) {
-  int N = 100;
-  double mu = 8;
-  double phi = 1e12;
-  for (int n = 0; n < 10; ++n) {
-    phi *= 10;
-    double logp = stan::math::neg_binomial_2_log<false>(N, mu, phi);
-    EXPECT_LT(logp, 0);
+  std::array<unsigned int, 5> n_to_test = {1, 5, 100, 12985, 1968422};
```
bob-carpenter: [optional]

```
std::vector<unsigned int> n_to_test = ...
```
is shorter and easier to modify.  And speed won't matter here in the tests.

test/unit/math/prim/prob/neg_binomial_2_test.cpp
```diff
@@ -239,16 +239,64 @@ TEST(ProbDistributionsNegBinomial, chiSquareGoodnessFitTest4) {
...
+  for (double mu : mu_to_test) {
+    for (unsigned int n : n_to_test) {
+      // Test just before cutoff
+      double logp
+          = stan::math::neg_binomial_2_log<false>(n, mu, phi_cutoff - 1e-8);
+      EXPECT_LT(logp, 0) << "n = " << n << ", mu = " << mu
+                         << ", phi = " << (phi_cutoff - 1e-8);
+
+      // Test across a range of phi
+      double phi = 1e12;
```
bob-carpenter: Loop directly on `phi`:

```
for (double phi = 1e12; phi < 1e22; phi *= 10) {
  ...
}
```

test/unit/math/rev/prob/neg_binomial_2_test.cpp
```diff
@@ -1,20 +1,500 @@
...
+//          WriteString[out,"  TestValue(",CForm[cn],",",CForm[cmu],",",
+//            CForm[cphi],",", CForm[val],","CForm[ddmu],",",CForm[ddphi],"),"]
+//        }]
+//      }]
+//   }]
+//    }];
+//  WriteString[out,"};"];
+//  Close[out];
+//  FilePrint[%]
+std::array<TestValue, 210> testValues = {
```
bob-carpenter: When the size gets to 210, this should be `std::vector`---it's way too many to count by hand.

stan/math/prim/scal/fun/lgamma.hpp
```diff
@@ -89,6 +89,17 @@ inline double lgamma(int x) {
 #endif
 }
 
+inline double lgamma(unsigned int x) {
+#if !__MINGW32__
+  int sign = 1;
+  return ::lgamma_r(x, &sign);
```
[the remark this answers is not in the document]
martinmodrak: I have literally no idea :-) I just copied the code for `inline double lgamma(int x)`, (just above in the file) all the quirks are inherited from there. The code seems to have been written by @wds15 in https://github.com/stan-dev/math/pull/1255. It is possible that duplicating the code was not the best way to do this. Alternatively, I can probably force conversion to signed `int` somewhere and remove this call, but that felt less elegant.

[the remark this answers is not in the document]
martinmodrak: > The logic is simpler if the loops are rearranged and the negation is eliminated.

Not sure what you are referring to - perhaps you ment a diffent chunk of code?

stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -79,33 +84,60 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
-    if (!is_constant_all<T_location>::value) {
-      ops_partials.edge1_.partials_[i]
-          += n_vec[i] / mu__[i] - (n_vec[i] + phi__[i]) / (mu__[i] + phi__[i]);
-    }
-    if (!is_constant_all<T_precision>::value) {
-      ops_partials.edge2_.partials_[i]
-          += 1.0 - n_plus_phi[i] / (mu__[i] + phi__[i]) + log_phi[i]
-             - log_mu_plus_phi[i] - digamma(phi__[i]) + digamma(n_plus_phi[i]);
+    if (phi__[i] > internal::neg_binomial_2_phi_cutoff) {
+      // Phi is large, deferring to Poisson.
```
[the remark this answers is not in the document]
martinmodrak: I kind of like more verbose comments (possibly because I am new to the codebase and had to learn a lot from gleaning at the code). Instead, I moved all the comments together to keep the code clean and tight while still letting people read more details about the code. Do you think that's better? (I don't insist on this in any way)

stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -79,33 +84,60 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
+      // Copying the code here as just calling
+      // poisson_lpmf does not preserve propto logic correctly
+      if (include_summand<propto>::value) {
+        logp -= lgamma(n_vec[i] + 1.0);
+      }
+      if (include_summand<propto, T_location>::value) {
+        logp += multiply_log(n_vec[i], mu__[i]) - mu__[i];
+      }
+
```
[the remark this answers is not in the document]
martinmodrak: Done

stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp
martinmodrak: Thanks for spotting this, removed.

test/unit/math/prim/prob/neg_binomial_2_test.cpp
martinmodrak: Sure, I kind of missed you can do this with `std::vector`. Modified everywhere.

test/unit/math/prim/prob/neg_binomial_2_test.cpp
martinmodrak: Good point, modified.

test/unit/math/rev/prob/neg_binomial_2_test.cpp
martinmodrak: Modified.

martinmodrak:
Thanks @bob-carpenter for the notes, it definitely made the code better. I however realized one thing that is weird in the current implementation and I would like your input on this. 

The thing is: I return non-zero derivative wrt. `phi` when deferring to Poisson, because otherwise the code doesn't pass the `derivativesAtCutoff` test - the derivative wrt. `phi` just before cutoff can be around `1e-6` which is deemed different enough from `0`. However, the computation of the density does not actually involve `phi`, so finite diffs are zero. Now I wasn't able to write a sensible test that would force me to add the second-order term to the density (which would involve `phi`), neither did this second term improve approximation (letting me to decrease `neg_binomial_2_phi_cutoff`). 

If I understand it correctly, the reason the tests pass without the second-order term in the density approximation is that the absolute value of the log density is large enough that changes on the order of `1/phi` are negligible _relative_ to the total log density.

This leaves us with 3 options:
1) the non-zero derivative wrt. `phi` after cutoff is not a problem 
2) that `derivativesAtCutoff` test is misguided and should no look at the absolute values of the derivatives but rather at the values of the derivatives relative to the actual log density
3) we should include the second-order term (ideally finding a test that would fail without this term being included).

What do you think?

Just for reference, the second order term of the log density approximation is `(mu^2 -n -2*mu*n +n ^2)/phi`.

martinmodrak:
Further, I added very large `n` values to the complex step test. Turns out it fails for derivative wrt. `phi` for `n > 100000` and large `phi`. Importantly, the sign of the derivative does not match (e.g. expected = `-3.24e-10`, actual = `9e-12`), not sure if the complex step or the function is to blame, so I assume it is OK for now and just increase the tolerance a notch.

bob-carpenter:
When values get near zero, the test framework switches from relative tolerance to absolute tolerance.  Relative tolerance doesn't make sense when one value is zero and the other non-zero.

In addition to quality of the approximation, we'd ideally like the interpolation to be continuously differentiable.  It doesn't have to be perfect, but if it gets too far off you can wind up having problems with HMC at the boundaries.

When delegating to Poisson, can you also delegate derivatives?  Then the derivatives match the function and finite diffs will work everywhere except across the boundary.  If the parameter gets transformed, you'd need the Jacobian adjustment.

To summarize, it shouldn't be a problem.  If we find cases in practice where it is, we can rethink.

bob-carpenter:
The test framework goes over to absolute difference for small values because relative is so finicky around zero.  In practice, when derivatives are near zero, sign differences aren't a big deal.  You go the wrong way, but not far.

stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp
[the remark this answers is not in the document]
bob-carpenter: Here's [Google style on implementation comments](https://google.github.io/styleguide/cppguide.html#Implementation_Comments).

It's no big deal either way.  The general rule for comments is to assume the reader of the code is better than the author at the language, so only non-idiomatic or non-obvious things get documented.  A good example is the comment about why there is cut-and-paste code.

The reason comments should be minimal is that it helps with code readability, which is the true goal.  That's largely because the code is the only thing you can ever really trust---comments can be obsolete or just plain wrong.  Too many comments get in the way of code readability, especially when they eat up a lot of vertical space and make it hard to take in the structure of the code at a glance.

P.S.  Also, that should be "delegate" instead of "defer";  my bad in just condensing rather than editing!

stan/math/prim/scal/fun/lgamma.hpp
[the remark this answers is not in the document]
bob-carpenter: If there's cut-and-pasted code, please cross-link so we can fix both at the same time.  Hopefully @wds15 can explain what's going on and we can either leave the cut and paste or change them both later.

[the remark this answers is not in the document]
bob-carpenter: I meant the logic's easier on the reader without the negation.  In general, it's easier to follow

```
#if C
...C is true...
#else
...C is false...
```

than

```
#if !C
...C is false...
#else
... C is true...
```

On the other hand, we usually want to peel off exceptional behavior so that the main line of execution is not nested:

```
if (bad input) {
  ... handle with return or excpetion
}
... main code ...
```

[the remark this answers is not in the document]
rok-cesnovar: This is the C library version of lgamma that is reentrant safe as opposed to `std::lgamma`. 

See https://github.com/stan-dev/math/pull/1255 for details, but the main point is: 

> The lgamma_r is a C library call which is only available upon defining -D_REENTRANT, but this is not working on Windows with MinGW32. This is why the boost version is used on Windows.

bob-carpenter:
Thanks.  A line to that effect in the doc (both places it shows up if it's cut and paste) would be super helpful and stop the next person from asking the same thing.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -49,21 +58,17 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
   size_t len_ep = max_size(mu, phi);
   size_t len_np = max_size(n, phi);
 
-  VectorBuilder<true, T_partials_return, T_location> mu__(size(mu));
-
-  for (size_t i = 0, max_size_seq_view = size(mu); i < max_size_seq_view; ++i) {
+  size_t len_mu = size(mu);
+  VectorBuilder<true, T_partials_return, T_location> mu__(len_mu);
+  for (size_t i = 0; i < len_mu; ++i) {
```
mcol: I've been removing uses of trailing underscores for local variables. You could rename this to `mu_val`, for example.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -49,21 +58,17 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
     mu__[i] = value_of(mu_vec[i]);
   }
 
-  VectorBuilder<true, T_partials_return, T_precision> phi__(size(phi));
-  for (size_t i = 0, max_size_seq_view = size(phi); i < max_size_seq_view;
-       ++i) {
+  size_t len_phi = size(phi);
+  VectorBuilder<true, T_partials_return, T_precision> phi__(len_phi);
```
mcol: Rename this to `phi_val` or something to that extent.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -79,33 +84,60 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
-    if (!is_constant_all<T_precision>::value) {
-      ops_partials.edge2_.partials_[i]
-          += 1.0 - n_plus_phi[i] / (mu__[i] + phi__[i]) + log_phi[i]
-             - log_mu_plus_phi[i] - digamma(phi__[i]) + digamma(n_plus_phi[i]);
+    if (phi__[i] > internal::neg_binomial_2_phi_cutoff) {
+      // Phi is large, deferring to Poisson.
+      // Copying the code here as just calling
+      // poisson_lpmf does not preserve propto logic correctly.
+      // Note that Poisson can be seen as first term of Taylor series for
+      // phi -> Inf. Similarly, the derivativew wrt. mu and phi can be obtained
```
mcol: Typo: "derivativew". Also I would use "wrt" rather than "wrt." as the latter makes me think the sentence is finished (there are other uses).

stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp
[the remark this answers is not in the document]
martinmodrak: I really want to keep the comments on how I got the formulae as I am confident about the code and not very confident about the math, so I want to "show my work" and make it easier for others to evaluate if it makes sense.

martinmodrak:
Thanks, @bob-carpenter.

First a procedural question: is up to me or up to you to "resolve conversation" around suggested changes once I incorporate the suggestions?

> When values get near zero, the test framework switches from relative tolerance to absolute tolerance. Relative tolerance doesn't make sense when one value is zero and the other non-zero.

Just to clarify - are you describing something that happens automatically in `EXPECT_NEAR` or similar calls (which I don't think is the case), or something I should be doing myself in the code? Or is there a preferred/default way to enforce this? Currently I do something to that effect in my code by hand.

> In addition to quality of the approximation, we'd ideally like the interpolation to be continuously differentiable. It doesn't have to be perfect, but if it gets too far off you can wind up having problems with HMC at the boundaries. When delegating to Poisson, can you also delegate derivatives? ... To summarize, it shouldn't be a problem. If we find cases in practice where it is, we can rethink.

Not sure what to take from this - what shouldn't be a problem? If I delegate the derivatives to Poisson derivatives, they are *slightly* discontinuous (there is a sudden step from around 1e-6 to 0). So I have to choose either continuity OR matching finite diffs OR both, but additional computation (adding the second-order approximation term to the log density). Or did you mean that any of the three solutions is OK and I should just choose the one I like most?

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
martinmodrak: Thanks, will do.

martinmodrak:
Thanks @mcol for the suggestions, I've incorporated all of them.

stan/math/prim/scal/fun/lgamma.hpp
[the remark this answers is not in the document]
martinmodrak: Turns out, I only needed the copy-pasted code because I was using `unsigned int` in my tests, but all of Stan uses signed ints, so I was able to remove the code.

bob-carpenter:
> I've been removing uses of trailing underscores for local variables. You could rename this to mu_val,

Thanks!

> for example.
> 
> In stan/math/prim/prob/neg_binomial_2_lpmf.hpp:
> 
> >      mu__[i] = value_of(mu_vec[i]);

Two double underscores are illegal in C++ as they're reserved for system functions.  We violate this in our parser code to make sure not to conflict with user code, but it actually violates the C++ language.  We definitely shouldn't have double underscores in regular functions.

For Stan C++ code, we've conventionally reserved trailing underscore for member variables and it'd be nice to stick to that convention.

> Typo: "derivativew". Also I would use "wrt" rather than "wrt." as the latter makes me think the sentence is finished (there are other uses).

Spell it out or use the standard lower-case acronym "w.r.t.".

bob-carpenter:
> On Jan 6, 2020, at 5:29 AM, Martin Modrák <notifications@github.com> wrote:
> 
> Thanks, @bob-carpenter.
> 
> First a procedural question: is up to me or up to you to "resolve conversation" around suggested changes once I incorporate the suggestions?

Either of us can if we think it's resolved, but nothing hinges on having all conversations marked resolved.  As reviewer, I get final approval the way we're doing things now, but all of this is fairly minor, so I don't see a problem if you object or want to do things differently

> 
> When values get near zero, the test framework switches from relative tolerance to absolute tolerance. Relative tolerance doesn't make sense when one value is zero and the other non-zero.
> 
> Just to clarify - are you describing something that happens automatically in EXPECT_NEAR or similar calls

In the AD test framework, the relevant code is part of expect_near_rel in test/unit/math/expect_near_rel.hpp.  It also handles doing the right thing for NaN values, which don't play nicely with standard comparison operators.

> (which I don't think is the case), or something I should be doing myself in the code? Or is there a preferred/default way to enforce this? Currently I do something to that effect in my code by hand.
> 
> In addition to quality of the approximation, we'd ideally like the interpolation to be continuously differentiable. It doesn't have to be perfect, but if it gets too far off you can wind up having problems with HMC at the boundaries. When delegating to Poisson, can you also delegate derivatives? ... To summarize, it shouldn't be a problem. If we find cases in practice where it is, we can rethink.
> 
> Not sure what to take from this - what shouldn't be a problem?

If they're only slightly discontinuous (in a numerical sense).  When they get too discontinuous, we wind up almost never accepting anything that crosses a discontinuity.

> If I delegate the derivatives to Poisson derivatives, they are slightly discontinuous (there is a sudden step from around 1e-6 to 0). So I have to choose either continuity OR matching finite diffs OR both, but additional computation (adding the second-order approximation term to the log density). Or did you mean that any of the three solutions is OK and I should just choose the one I like most?

Yes, you should choose which solution you like with an awareness of the goals.  I don't know the numerics for this particular case and was just trying to lay out the bigger picture.

martinmodrak:
Just a quick update, using `expect_near_rel` showed a weakness in one of my tests, which then showed more failures, which led me to realize the Poisson approximation breaks for small `n`. In turn, I discovered bug #1592  and it seems that delegating to Poisson is not necessary if the computation is further reordered and #1592 is resolved. (not sure how to resolve it, though, but for all currently failing tests, manually recomputing with the correct value for `binomial_coefficient_log` produces good results).

bob-carpenter:
Cool.  Looks like you figured out how to fix #1592 (N >> n binom coeff imprecision) downstream.

martinmodrak:
>  Looks like you figured out how to fix #1592 (N >> n binom coeff imprecision) downstream.

Unfortunately, I did not.

stan/math/prim/scal/fun/lgamma_stirling.hpp
```diff
@@ -10,6 +10,7 @@ namespace math {
 
 template <typename T>
 T lgamma_stirling(const T x) {
+  using std::log;
   return 0.5 * log(2 * stan::math::pi()) + (x - 0.5) * log(x) - x;
```
mcol: Note that we have `LOG_TWO_PI` defined in `stan/math/prim/scal/fun/constants.hpp`.

martinmodrak: Thanks for the input. Sorry, the code is now a bit of a mess, #1592 sent me on a wild tangent and the current state is more of a prototype than polished code.

mcol: No problem, I was not trying to review, just browsed the code and saw that! Ah, perhaps there'll be a HALF_LOG_TWO_PI constant soon. :)

martinmodrak:
This PR requires #1614 and #1612 to be merged before it can be merged (otherwise tests do not pass).

bbbales2:
@martinmodrak I was messing around with the distribution testing framework for a pull @mcol was doing and thought we ran across a bug with how the propto checks work: https://github.com/stan-dev/math/issues/1763

I put a fix in and am now getting errors in the negative binomial case.

The test that is failing is that (which will be the log of the normalizing constant for the propto distribution):

```
stan::math::neg_binomial_lpmf<false>(10, var(2.5), var(1.5)) -
    stan::math::neg_binomial_lpmf<true>(10, var(2.5), var(1.5))
```

Is a different number than:

```
stan::math::neg_binomial_lpmf<false>(10, var(1e11), var(1e10)) -
    stan::math::neg_binomial_lpmf<true>(10, var(1e11), var(1e10))
```

There's the switch to the Poisson stuff at some point. The normalizing constant should stay the same though right? The approximation has to be continuous when it kicks in?

Would the sequence of pull requests here make this error I'm getting go away?

martinmodrak:
@bbbales2 I can imagine the switch to Poisson to break the `propto` logic. But I believe the function can be implemented without any switch so once this all goes through, it should not be an issue. This is a good nudge for me to move this forward a bit.

mcol:
@martinmodrak let me know if you need a hand in solving the conflict, quite a few of them were caused by changes I made, so if I can help... :)

martinmodrak:
@bob-carpenter So that code evolved quite a bit and the required changes to `binomial_coefficient_log` have been merged. I believe this is now in good shape for a re-review.

bbbales2:
I can take the review over if that's convenient since I did the last binomial one

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -49,66 +48,62 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
   size_t size_n_phi = max_size(n, phi);
   size_t max_size_seq_view = max_size(n, mu, phi);
 
-  VectorBuilder<true, T_partials_return, T_location> mu_val(size_mu);
-  for (size_t i = 0; i < size_mu; ++i) {
+  size_t len_ep = max_size(mu, phi);
```
mcol: This is a duplicate of `size_mu_phi` above, which has a clearer name.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -49,66 +48,62 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
+  size_t len_np = max_size(n, phi);
```
mcol: This duplicates `size_n_phi`.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -49,66 +48,62 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
+
+  size_t len_mu = size(mu);
```
mcol: This duplicates `size_mu`. By the way, note that for compatibility with C++17 compilers, all uses of `size()` should actually be `stan::math::size()` even if we are inside the `stan::math` namespace.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -49,66 +48,62 @@ return_type_t<T_location, T_precision> neg_binomial_2_lpmf(
...
+  VectorBuilder<true, T_partials_return, T_location> mu_val(len_mu);
+  for (size_t i = 0; i < len_mu; ++i) {
     mu_val[i] = value_of(mu_vec[i]);
   }
 
-  VectorBuilder<true, T_partials_return, T_precision> phi_val(size_phi);
-  VectorBuilder<true, T_partials_return, T_precision> log_phi(size_phi);
-  for (size_t i = 0; i < size_phi; ++i) {
+  size_t len_phi = size(phi);
```
mcol: We already have `size_phi`.

stan/math/prim/prob/neg_binomial_2_lpmf.hpp
```diff
@@ -3,13 +3,12 @@
 
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/binomial_coefficient_log.hpp>
 #include <stan/math/prim/fun/digamma.hpp>
 #include <stan/math/prim/fun/lgamma.hpp>
-#include <stan/math/prim/fun/log.hpp>
-#include <stan/math/prim/fun/max_size.hpp>
 #include <stan/math/prim/fun/multiply_log.hpp>
-#include <stan/math/prim/fun/size.hpp>
```
mcol: You removed headers that are still used.

martinmodrak:
@mcol thanks for catching those - I was obviously sloppy with the merge, sorry for that. I resolved those.

bbbales2:
@bob-carpenter can I take this over?

bob-carpenter:
Yes, please!

test/unit/math/prim/prob/neg_binomial_2_log_test.cpp
```diff
@@ -212,7 +212,10 @@ TEST(ProbNegBinomial2, log_matches_lpmf) {
 TEST(ProbDistributionsNegBinomial2Log, neg_binomial_2_log_grid_test) {
   std::vector<double> mu_log_to_test
       = {-101, -27, -3, -1, -0.132, 0, 4, 10, 87};
-  std::vector<double> phi_to_test = {2e-5, 0.36, 1, 2.3e5, 1.8e10, 6e16};
+  // TODO(martinmodrak) Reducing the span of the test, should be fixed
+  // along with #1495
+  // std::vector<double> phi_to_test = {2e-5, 0.36, 1, 10, 2.3e5, 1.8e10, 6e16};
```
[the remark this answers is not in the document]
martinmodrak: The other pull (#1495) is for the log version (`neg_binomial_2_log`). This test checks that `neb_binomial_2(n, mu, phi)` and `neg_binomial_2_log(n, log(mu), phi)` agree. The fixes in this PR change `neg_binomial_2` but ignore `neg_binomial_2_log` so they now diverge (previously they were both equally wrong). They should agree again once I rewrite the log version with the tricks I learned while implementing this PR.

test/unit/math/rev/prob/neg_binomial_2_test.cpp
```diff
@@ -27,22 +456,140 @@ TEST(ProbDistributionsNegBinomial2, derivatives) {
...
+      double expected_dphi_0
+          = mu_dbl / (mu_dbl + phi_dbl) - log1p(mu_dbl / phi_dbl);
+      expect_near_rel("dmu, n = 0 " + msg.str(), gradients0[0], expected_dmu_0);
+      expect_near_rel("dphi, n = 0 " + msg.str(), gradients0[1],
+                      expected_dphi_0);
+
+      double expected_dmu_1
+          = (phi_dbl * (1 - mu_dbl)) / (mu_dbl * (mu_dbl + phi_dbl));
+      expect_near_rel("dmu, n = 1 " + msg.str(), gradients1[0], expected_dmu_1);
+    }
```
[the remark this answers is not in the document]
martinmodrak: I swear I had `expected_d_phi_p` in there before. Must have got lost in some of the merges. Will add.

martinmodrak:
Thanks @bbbales2 will try to resolve shortly. Note to self: #1679 was fixed and merged, can also expand the lbeta tests.

bbbales2:
Quick ping.

> can also expand the lbeta tests.

I don't think there's a need to add many new tests here if that's what you're getting at! Just those couple things, and really I'd be fine with the pull as-is.

martinmodrak:
OK, just added the the test for d/dphi at n = 1. Also the lbeta thing was literally that I had a line of code saying "Delete this once #1679 is merged", so I deleted it :-) Tests pass on my computer, hopefully they pass on Jenkins as well.

bob-carpenter:
Sorry I missed this was blocked on me.

Issue #1342, opened 2019-09-06
Implement OpenCL kernel generator


t4c1:
## Description
Now we have separate kernels for simple operations (such as matrix addition or transposition). These kernels do low amount of computation, compared to amouunt of data transfered from global GPU memory. So vast majority of time spent executing such kernels is needed for data transfers, instead of actual computations.

Usually when multiple simple operations are executed one after another they could be combined into a single kernel so that data is loaded only once. Less data transfers result in faster execution. 

We should combine multiple operations on `matrix_cl` into expressions in similar way as Eigen does it. When evaluated or assigned to another matrix_cl an expression generates OpenCL kernel source, compiles and executes it.

Envisioned pull requests:
- [x] basic kernel generator with matrices, scalars and binary element-wise operations (+, -, *, /)
- [x] unary functions
- [x] submatrix blocks
- [x] selction operator
- [x] comparisons and logical operators
- [x] reductions(sum, min, max)
- [x] multiple results in a single kernel
- [x] broadcasting
- [x] transposition
- [ ] more?

## Example

The code below should result in execution of a single kernel.
```
matrix_cl<double> a;
matrix_cl<double> b;
matrix_cl<double> c;
matrix_cl<double> d = a + b - 3 * c;
```

## Expected Output

Expressions of simple matrix operations on GPUs are faster. Creating simple OpenCL kernels  does not require writing OpenCL source and can be done by non-GPU developers.

#### Current Version:
v2.20.0

rok-cesnovar:
@t4c1 we can probably close this right? If other issues/feature requests pop up we can open additional issue.

Issue #1485, opened 2019-12-05
include-what-you-use errors


bob-carpenter:
## Description

The following files (at least) have include-what-you-use violations:

* `rev/mat/fun/ordered_constrain.hpp`
* `rev/mat/fun/positive_ordered_constrain.hpp`

They use `exp(double)` without including a definition or bringing `std::exp` into scope with a using statement.  The fix is to include `<cmath>` and add a `using std::exp` in functions that use it.  

The header tests didn't catch it because there's an `exp()` definition lurking in `::` for C99 compatibility.  These have caused us enormous headaches in the past for this reason.

I found this bug when introducing `exp(const std::complex<T>&)`, which broke the header tests.   

#### Current Version:
v3.0.0

syclik:
Thanks for trapping this! And for writing a clear issue.

bob-carpenter:
I'm finding more and more of them as I expose std:: functions like sin and exp operating on complex numbers.  I'm just fixing them as I go, but I can probably pull them off into their own PR when I'm done.

rok-cesnovar:
There is also an include of `<math.h.>` here https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/fun/gp_periodic_cov.hpp#L4 

Not sure why that is there.

PR #1500, opened 2019-12-11
Add unary functions to kernel generator

t4c1:
## Summary

Adds unary functions to kernel generator - ones that are defined by OpenCL standard. To prevent code duplication implementation and tests use macros.

Also fixes a bug in operation_cl that limited operations to ones with 2 arguments. To prevent clashes a require was added to `apply_scalar_unary`.

List of all added functions: rsqrt, sqrt, cbrt, exp, exp2, expm1, log, log2, log10, log1p, sin, cos, tan, sinh, cosh, tanh, asin, acos, atan, asinh, acosh, atanh, tgamma, lgamma, erf, erfc, floor, round, ceil.

## Tests

Every added function has a simple test.

## Side Effects

None.

## Checklist

- [x] Math issue #1342 


SteveBronder:
Nice! This should be a much quicker review I'll take a look today or tomorrow. Does the unary also do the subexpression elimination aka if we do

```
auto tmp= elewisemult(a,b);
matrix_cl = sin(tmp) + tmp;
```

t4c1:
Subexpression elimination is part of `operation_cl`, so it works for all existing and future opeartions.

t4c1:
This PR seems to be hitting into more instances of this issue: https://github.com/stan-dev/math/issues/1485

rok-cesnovar:
@t4c1 the merge of the include-what-you-use-branch was weird somehow, showing a bunch of changes, so I reset those and just merged develop in.

stan/math/opencl/kernel_generator/unary_function_cl.hpp
```diff
@@ -0,0 +1,152 @@
...
+   public:                                                                     \
+    explicit fun##__(T&& a)                                                    \
+        : unary_function_cl<fun##__<T>, T>(std::forward<T>(a), #fun) {}        \
+  };                                                                           \
+                                                                               \
+  template <typename T, typename Cond                                          \
+                        = require_all_valid_expressions_and_none_scalar_t<T>>  \
+  inline fun##__<as_operation_cl_t<T>> fun(T&& a) {                            \
+    return fun##__<as_operation_cl_t<T>>(as_operation_cl(std::forward<T>(a))); \
+  }
```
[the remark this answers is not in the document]
t4c1: The only difference is in that view. `ADD_UNARY_FUNCTION()` has its own definition of `.view()` that always returns `Entire`. This is used for unary functions with no special properties. 

However many functions `f` pass trought zero unchanged - `f(0)=0`. For these fuctions we can propagate triangular view of its argument. So `ADD_UNARY_FUNCTION_PASS_ZERO()` uses definition from `class unary_function_cl` that does that.

t4c1:
> one thing I'm wondering about is whether the class should take in opencl code for functions that don't exist in the opencl spec like lbeta or other stuff in scale. similarly for stuff like fma for binary ops we could pass opencl code representing that op

We can definitely do that. However the code would be a bit different than for functions from OpenCL spec, so I suggest a separate class and separate PR(s).

If you have a list of functions and their implementations I can add them in next PR. Otherwise I would suggest adding them as we need them.

stan/math/opencl/kernel_generator/unary_function_cl.hpp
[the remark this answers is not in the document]
SteveBronder: > However many functions f pass trought zero unchanged - f(0)=0. For these fuctions we can propagate triangular view of its argument. So ADD_UNARY_FUNCTION_PASS_ZERO() uses definition from class unary_function_cl that does that.

I'm cool with that but let's add it to the docs so people in the future can understand the diff between the two

stan/math/opencl/kernel_generator/unary_function_cl.hpp
```diff
@@ -0,0 +1,152 @@
...
+ * Represents a unary function in kernel generator expressions.
+ * @tparam Derived derived type
+ * @tparam T type of argument
+ */
+template <typename Derived, typename T>
+class unary_function_cl
+    : public operation_cl<
+          Derived, typename std::remove_reference_t<T>::ReturnScalar, T> {
+ public:
+  using ReturnScalar = typename std::remove_reference_t<T>::ReturnScalar;
```
SteveBronder: Whups, just realized these won't work with value_type and scalar_type will they? I think you can just add a `Scalar` and `value_type` alias here and in binary_op and that should be fine

SteveBronder:
> We can definitely do that. However the code would be a bit different than for functions from OpenCL spec, so I suggest a separate class and separate PR(s).

The only difference I can think of would be

1. Having a new argument in function signature
2. Putting OpenCL funcs above function definition

The only thing I can think of that would be weird is supporting multiple types where we'd have to do something like templating and greps during construction. So like say we didn't have square function we would want something that looks like


```cpp
square_kernel_cl = STRINGIFY(
template <typename T>
T square(T a) {
  return a*a;
}
)

// in unary fun
  std::string code_{""};
  unary_function_cl(T&& a, const std::string& fun, const std::string& code)
      : base(std::forward<T>(a)), fun_(fun), code_(code) {}

// would have to change the def of kernel_parts to support adding stuff above signature
  inline kernel_parts generate(const std::string& i, const std::string& j,
                               const std::string& var_name_arg) const {
    kernel_parts res{};
   // parse here would
   // 1. search for the template head
   // 2. get the type identifier and replace it with `T`
    if (check_for_code(code_)) {
      res.funcs = parse<value_type<T>>(code_);   
   }
    res.body = type_str<ReturnScalar>() + " " + var_name + " = " + fun_ + "("
               + var_name_arg + ");\n";
    return res;
  }
```

Though supporting a subset of C++11 seems like, a lot haha. So yeah maybe this should go in a separate PR. We would probs need an if check to see if `code_` exists (or is not a nullptr), though I think it would work generally. Pretty sure the form above should also work nicely for binary ops 

If we just wanted to support explicitly written funcs though then it's the `parse` thing goes away and it's pretty straightforward

stan/math/opencl/kernel_generator/unary_function_cl.hpp
```diff
@@ -0,0 +1,152 @@
...
+  std::string fun_;
+};
+
+/**
+ * generates a class and function for an unary function.
+ * @param fun function
+ */
+#define ADD_UNARY_FUNCTION(fun)                                                \
+  template <typename T>                                                        \
+  class fun##__ : public unary_function_cl<fun##__<T>, T> {                    \
```
SteveBronder: this comes out to `sin__` right? C++ reserves anything with double underscores in it for the compiler

t4c1:
Parsing templates in OpenCL is a huge overkill. Even if we need functions with variable input types we can just construct function signature in similar way to how kernel signature is generated.

stan/math/opencl/kernel_generator/unary_function_cl.hpp
t4c1: I checked the code and both `value_type` and `scalar_type` will need specializations for kernel generator expressions. For now I am renaming ReturnScalar to Scalar. Changes to metaprogramming can come as a separate PR.

stan/math/opencl/kernel_generator/unary_function_cl.hpp
t4c1: Faixed.

stan/math/opencl/kernel_generator/unary_function_cl.hpp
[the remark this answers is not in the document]
t4c1: Done.

SteveBronder:
> If you have a list of functions and their implementations I can add them in next PR (or you can try). Otherwise I would suggest adding them as we need them.

I can give it a go this weekend

PR #1507, opened 2019-12-12
Bugfix/include what you use

### LINKED ISSUE #1485: include-what-you-use errors

rok-cesnovar:
## Summary

This PR fixes the files that were not respecting include-what-you-use with respect to <cmath> and the lack of `using std::*`.
It also removes an include of `<math.h> `and does some minor include cleanup.

## Tests

/

## Side Effects

/

## Checklist

- [x] Math issue Fixes #1485 


stan/math/fwd/scal/fun/inv_logit.hpp
```diff
@@ -4,6 +4,7 @@
 #include <stan/math/fwd/meta.hpp>
 #include <stan/math/fwd/core.hpp>
 #include <stan/math/prim/scal/fun/inv_logit.hpp>
+#include <cmath>
```
bob-carpenter: [question]
Any idea why cpplint didn't catch this?  I thought they were supposed to catch problems with include-what-you-use from the std:: lib.  Anyway, thanks for finding and fixing all this!

stan/math/opencl/opencl_context.hpp
```diff
@@ -107,7 +107,7 @@ class opencl_context_base {
         in_order_ = CL_TRUE;
       }
       int thread_block_size_sqrt
-          = static_cast<int>(sqrt(static_cast<double>(max_thread_block_size_)));
+          = static_cast<int>(std::sqrt(static_cast<double>(max_thread_block_size_)));
```
bob-carpenter: [optional]
With an explicit qualification for `std::sqrt`, we shouldn't need the cast to `double`---the implementation of `std::sqrt` in C++11 has all the primitives as overloads.  See:  https://en.cppreference.com/w/cpp/numeric/math/sqrt

stan/math/prim/scal/fun/erfc.hpp
```diff
@@ -21,7 +21,7 @@ inline double erfc(double x) { return std::erfc(x); }
 
 /**
  * Return the error function of the specified argument.  This
- * version is required to disambiguate <code>erf(int)</code>.
+ * version is required to disambiguate <code>erfc(int)</code>.
```
bob-carpenter: [comment]
Nice catch!

bob-carpenter:
Thanks.  And feel free to merge when checks pass.

rok-cesnovar:
I am actually just about to push some more missing <cmath> includes and then its done

stan/math/fwd/scal/fun/inv_logit.hpp
rok-cesnovar: I was amazed also, so dont have an answer. I made a simple Python script and just ran it through  `stan/math` and fixed the files the script flagged.

stan/math/opencl/opencl_context.hpp
rok-cesnovar: Done! Thanks.

rok-cesnovar:
Thanks for the review! I fixed the static cast and added the rest of missing `<cmath>` includes. Will tag you if/when it passes tests to get this merged.

rok-cesnovar:
Something weird happened here. I merged this PR, the commit is on develop (https://github.com/stan-dev/math/commit/a2ef4ff23d293c4004877fd2af4ad126f31e6ffc) but it didn't close this PR and it's now showing like it has merge conflicts.

rok-cesnovar:
Probably related to https://www.githubstatus.com/incidents/4g79klqzy99s better leave it and wait for github to resolve it.

PR #1482, opened 2019-12-02
Rename stan::math::domain_error to throw_domain_error

mcol:
## Summary

This renames `domain_error` and `domain_error_vec` to `throw_domain_error` and `throw_domain_error_vec`. Fixes #339.

## Tests

No new tests, this is just clean up.

## Side Effects

None.

## Checklist

- [X] Math issue #339

- [X] Copyright holder: Marco Colombo

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [X] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [X] the code is written in idiomatic C++ and changes are documented in the doxygen

- [X] the new changes are tested

PR #1508, opened 2019-12-13
style - opencl header update

alashworth:
## Summary

Changes the OpenCL header includes from <cl.hpp> to <CL/cl2.hpp>. 

Reason: this naming convention is the way upstream, all major vendors (Intel, AMD, NVidia), and distributions package their header files. This change is useful for anybody packaging or distributing Stan downstream, since they no longer need to patch Stan's file names.

## Side Effects

While making these changes, I also moved one file-local function into an anonymous namespace. It should not affect any current user of Stan. I moved it because I am performing unity/jumbo builds and the function was violating the one definition rule.

## Checklist
- [x] Copyright holder: Andrew Ashworth


stan/math/opencl/matrix_cl.hpp
```diff
@@ -9,7 +9,8 @@
 #include <stan/math/opencl/err/check_opencl.hpp>
 #include <stan/math/prim/arr/fun/vec_concat.hpp>
 #include <stan/math/prim/scal/err/check_size_match.hpp>
-#include <cl.hpp>
+#include <stan/math/prim/scal/err/domain_error.hpp>
```
rok-cesnovar: Why do we need this header here? I see no domain_error (deprecated since [yesterday](https://github.com/stan-dev/math/pull/1482)) or throw_domain_error here?

alashworth: Oops, good catch. I wasn't paying attention when rebasing onto HEAD.

rok-cesnovar:
It also seems that the clang-format cant push changes to your fork. Not sure why. Please check that you ticked the allow maintainers box on this PR. Otherwise, you will probably need to format it locally.

alashworth:
Hmm, I'm formatting it according to the clang-format file in the repository. Is there a difference on the Jenkins side that would cause it to fail?

Issue #1490, opened 2019-12-07
Some loops may need to be interchanged; column major


peterwicksstringfield:
## Description
There are a few places in the code where we walk a column-major Eigen matrix row-major style, when we could interchange the loops. Probably the compiler does it for us anyway, but interchanging the loops would make the code more self-documenting. This might be worth it? Feel free to close this ticket and trust the compiler to handle it.

Two such interchanges are performed in PR #1491.

The diffs would be trivial, but the change would not. A certain amount of close-reading is required to be sure the interchange is safe and the surrounding code is tested.

- [X] stan/math/prim/mat/prob/lkj_cov_lpdf.hpp
```
  for (int m = 0; m < y.rows(); ++m) {
    for (int n = 0; n < y.cols(); ++n) {
      check_finite(function, "Covariance matrix", y(m, n));
      ...
```
(Don't even need the loops here? Looks like this predates a generalization of check_finite, which can accept matrices.)

- [ ] stan/math/prim/mat/prob/gaussian_dlm_obs_lpdf.hpp
122 and 283
```
    for (int i = 0; i < C0.rows(); i++) {
      for (int j = 0; j < C0.cols(); j++) {
        C(i, j) = C0(i, j);
        ...
```
(See #1466.)

- [X] stan/math/prim/mat/prob/matrix_normal_prec_rng.hpp
102
```
  for (int row = 0; row < m; ++row) {
    for (int col = 0; col < n; ++col) {
      double stddev = row_stddev(row) * col_stddev(col);
      ...
      X(row, col) = stddev * std_normal_rng();
```

- [X] stan/math/fwd/mat/fun/to_fvar.hpp
52
```
  for (int i = 0; i < val.rows(); i++) {
    for (int j = 0; j < val.cols(); j++) {
      ret(i, j).val_ = val(i, j);
      ret(i, j).d_ = deriv(i, j);
      ...
```

- [X] stan/math/fwd/mat/fun/mdivide_left_ldlt.hpp
```
  for (int i = 0; i < b.rows(); i++) {
    for (int j = 0; j < b.cols(); j++) {
      b_val(i, j) = b(i, j).val_;
      b_der(i, j) = b(i, j).d_;
      ...
```

- [X] stan/math/fwd/mat/fun/inverse.hpp
22
```
  for (size_type i = 0; i < m.rows(); i++) {
    for (size_type j = 0; j < m.cols(); j++) {
      m_inv(i, j) = m(i, j).val_;
[...]

PR #1516, opened 2019-12-14
Feature/issue 1490 loop order

peterwicksstringfield:
## Summary

Partially addresses #1490.

Eliminates instances of row-major iteration across a column-major Eigen matrix in six places:
stan/math/prim/mat/prob/lkj_cov_lpdf.hpp
stan/math/prim/mat/prob/matrix_normal_prec_rng.hpp
stan/math/fwd/mat/fun/to_fvar.hpp
stan/math/fwd/mat/fun/mdivide_left_ldlt.hpp
stan/math/fwd/mat/fun/inverse.hpp
stan/math/rev/mat/functor/kinsol_data.hpp

I can trivially break this into multiple PRs if that would make it easier.

This pattern occurs frequently when handling matrices of fvars (especially in the mdivide_foo family):
```
       b_val(i, j) = b(i, j).val_;
       b_der(i, j) = b(i, j).d_;
```

There were three places where it was row-col instead of col-row. I fixed that.

It could be:
```
b_val = b.val();
b_der = b.d();
```

Which would be two walks across b. Well, technically val and d both return CWiseUnaryOps, and the walks happen when they are turned into dense matrices.

kinsol_data is tested indirectly, via its uses in the algebra solvers.

In Stan, lkj_cov_lpdf is depreciated in favor of lkj_corr_lpdf. Also lkj_cov_lpdf had no tests. The file that was supposed to contain its tests just had a copy of the lkj_corr_lpdf tests. I've added a few tests for error conditions in lkj_cov_lpdf, but its still not proper coverage.

When reviewing the lkj tests, note the following: if y is a covariance matrix, and D is a diagonal matrix with Dii = 1/sqrt[yii], then DyD is the associated correlation matrix. Now if yii = 1, mu_i = 0, sd_i = 1, then:
```
lkj_cov_lpdf[y, mu, sigma, eta]
    ~= lkj_corr_lpdf[DyD, eta] + sum_i lognormal_lpdf[sqrt[yii], mu_i, sigma_i]
    = lkj_corr_lpdf[y, eta] + sum_i lognormal_lpdf[1, 0, 1]
    = lkj_corr_lpdf[y, eta] + size_of_y * lognormal_lpdf[1, 0, 1]
```

## Tests

Added tests:
test/unit/math/prim/mat/prob/lkj_cov_test.cpp

Old tests:
test/unit/math/mix/mat/fun/inverse_test.cpp
test/unit/math/prim/mat/fun/inverse_test.cpp
test/unit/math/prim/mat/prob/matrix_normal_prec_rng_test.cpp
test/unit/math/mix/mat/fun/to_fvar_test.cpp
test/unit/math/fwd/mat/fun/to_fvar_test.cpp
test/unit/math/mix/mat/fun/mdivide_left_ldlt_test.cpp
test/unit/math/prim/mat/fun/mdivide_left_ldlt_test.cpp
test/unit/math/rev/mat/functor/algebra_solver_fp_test.cpp
test/unit/math/rev/mat/functor/algebra_solver_test.cpp

## Side Effects

None.

## Checklist

- [X] Math issue #1490

### DIFF
--- a/stan/math/fwd/mat/fun/inverse.hpp
+++ b/stan/math/fwd/mat/fun/inverse.hpp
@@ -15,12 +15,13 @@ namespace math {
 template <typename T, int R, int C>
 inline Eigen::Matrix<fvar<T>, R, C> inverse(
     const Eigen::Matrix<fvar<T>, R, C>& m) {
+  check_nonempty("inverse", "m", m);
   check_square("inverse", "m", m);
   Eigen::Matrix<T, R, C> m_deriv(m.rows(), m.cols());
   Eigen::Matrix<T, R, C> m_inv(m.rows(), m.cols());
 
-  for (size_type i = 0; i < m.rows(); i++) {
-    for (size_type j = 0; j < m.cols(); j++) {
+  for (size_type j = 0; j < m.cols(); j++) {
+    for (size_type i = 0; i < m.rows(); i++) {
       m_inv(i, j) = m(i, j).val_;
       m_deriv(i, j) = m(i, j).d_;
     }
--- a/stan/math/fwd/mat/fun/mdivide_left_ldlt.hpp
+++ b/stan/math/fwd/mat/fun/mdivide_left_ldlt.hpp
@@ -26,8 +26,8 @@ inline Eigen::Matrix<fvar<T2>, R1, C2> mdivide_left_ldlt(
 
   Eigen::Matrix<T2, R2, C2> b_val(b.rows(), b.cols());
   Eigen::Matrix<T2, R2, C2> b_der(b.rows(), b.cols());
-  for (int i = 0; i < b.rows(); i++) {
-    for (int j = 0; j < b.cols(); j++) {
+  for (int j = 0; j < b.cols(); j++) {
+    for (int i = 0; i < b.rows(); i++) {
       b_val(i, j) = b(i, j).val_;
       b_der(i, j) = b(i, j).d_;
     }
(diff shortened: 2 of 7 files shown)


stan/math/fwd/mat/fun/inverse.hpp
```diff
@@ -15,12 +15,13 @@ namespace math {
 template <typename T, int R, int C>
 inline Eigen::Matrix<fvar<T>, R, C> inverse(
     const Eigen::Matrix<fvar<T>, R, C>& m) {
+  check_nonempty("inverse", "m", m);
```
rok-cesnovar: @mcol you are working on these a lot. Just double checking with you if this is what we want or would returning an empty matrix be better. I do think we should check that the matrix is square before the nonempty check.

mcol: For inverse we always throw if it's empty, so the order of the checks doesn't quite matter (see #1464 for example).

rok-cesnovar: Sweet. All good then. Thanks @mcol

peterwicksstringfield: Looks like you've figured it out, but that function threw on non empty anyway. Inverse<fwd<T> > calls inverse<T>, which eventually terminates at either inverse<prim> or inverse<var>. Prim and var threw on non-empty, so fwd did too. This change just makes it explicit.

PR #1518, opened 2019-12-15
Bugfix/fix test headers with opencl

rok-cesnovar:
## Summary

We have a bunch of `make test-headers` errors if you run it with STAN_OPENCL=true. I am fixing those issues and adding the test-headers check with STAN_OPENCL to Jenkins.

## Tests

This adds a `sh "make -j${env.PARALLEL} test-headers"` to the Full unit with GPU stage in order to prevent this from happening again. This adds 4-5 minutes to testing but the entire testing procedure takes 7-8 hours so I think we can live with that. We could also split it out.

## Side Effects

/

## Checklist

- [x] Math issue #1485 

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -142,15 +142,32 @@ pipeline {
                 }
             }
         }
-        stage('Headers check') {
-            agent any
-            steps {
-                deleteDir()
-                unstash 'MathSetup'
-                sh "echo CXX=${env.CXX} -Werror > make/local"
-                sh "make -j${env.PARALLEL} test-headers"
-            }
-            post { always { deleteDir() } }
+        stage('Headers checks') {
+            parallel {
+              stage('Headers check') {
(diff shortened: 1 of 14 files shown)


Jenkinsfile
```diff
@@ -175,6 +175,7 @@ pipeline {
                         sh "echo STAN_OPENCL=true>> make/local"
                         sh "echo OPENCL_PLATFORM_ID=0>> make/local"
                         sh "echo OPENCL_DEVICE_ID=${OPENCL_DEVICE_ID}>> make/local"
+                        sh "make -j${env.PARALLEL} test-headers"
```
t4c1: [optional] Maybe OpenCL headers tests should nrun in parallel with existing headers tests.

rok-cesnovar: Yeah, that may be better. Will do.

Issue #1519, opened 2019-12-15
Unnecessary checks in *_lp[dm]f functions


mcol:
## Description
Inspired by @bbbales2's observation in the first paragraph of https://github.com/stan-dev/math/issues/1498#issuecomment-564581852, I noticed that there are a number of checks for an `include<...>::value` inside a for loop although at the beginning of the function we had returned if that same check was false. 

Edit by @rok-cesnovar : The same changes need to be done in GPU supported _lpdfs.

## Example
In `poisson_log_lpmf` we have this check towards the beginning:
```
if (!include_summand<propto, T_log_rate>::value) {
  return 0.0;
}
```
Therefore, further down the following code
```
for (size_t i = 0; i < length(alpha); i++) {
  if (include_summand<propto, T_log_rate>::value) {
    exp_alpha[i] = exp(value_of(alpha_vec[i]));
  }
}
```
can be simplified to
```
for (size_t i = 0; i < length(alpha); i++) {
  exp_alpha[i] = exp(value_of(alpha_vec[i]));
}
```

## Expected Output
This should perhaps produce a minimal speedup, unless compilers are already taking care of that. Other than that, these functions should produce the same output as before.

rok-cesnovar:
This do get compiled out but that is not a reason to not clean it.

PR #1520, opened 2019-12-15
Clean up unnecessary checks in _lpmf/_lpdf functions

mcol:
## Summary

I've gone through all `*_lpmf` and `*_lpdf` functions we have in math. For cases where we have
```
if (!include_summand<propto, ...>::value) {
  return 0;
}
```
this removes any identical check (where the terms in `...` match exactly) from the subsequent code. This is a somewhat conservative fix, as there may be more cases that could be cleaned up if I had better understanding of the `include_summand` function. 

It probably helps to ignore space changes when reviewing. The only cases in which I moved a bit of code are in `stan/math/prim/scal/prob/neg_binomial_2_log_lpmf.hpp` and `stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp`, as it looked weird to have a single line of code in the middle of a long sequence of if blocks.

## Tests

The current distribution tests should be enough to test that there aren't unwanted changes being introduced.

## Side Effects

None.

## Checklist

- [X] Math issue #1519

### DIFF
--- a/stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
+++ b/stan/math/prim/mat/prob/neg_binomial_2_log_glm_lpmf.hpp
@@ -158,10 +158,8 @@ neg_binomial_2_log_glm_lpmf(
                  - lgamma(forward_as<double>(phi_val)));
     }
   }
-  if (include_summand<propto, T_x_scalar, T_alpha, T_beta,
-                      T_precision>::value) {
-    logp -= sum(y_plus_phi * logsumexp_theta_logphi);
-  }
+  logp -= sum(y_plus_phi * logsumexp_theta_logphi);
+
   if (include_summand<propto, T_x_scalar, T_alpha, T_beta>::value) {
     logp += sum(y_arr * theta);
   }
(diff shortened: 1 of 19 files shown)


bob-carpenter:
> understanding of the include_summand

The `include_summand<propto, T1, ..., TN>::value` traits metaprogram evaluates to `false` if

* `propto` is true and all of the scalar types involved are primitives.

That is, we drop a term if it's constant and we're working up to a proportion.  Being a static check, if it evaluates to `false`, the code in the conditional drops out and nothing gets executed (even the check) at runtime.

It would also be nice to get rid of the `::value` using the same trick as the standard library uses for `enable_if_t` with some parameter pack magic for good measure.

```
template <bool propto, typename ...Ts>
using include_summand_b = include_summand<propto, Ts...>::value;
```

Then the usage could go from `include_summand<propto, T1, ..., TN>::value` to
`include_summand_b<propto, T1, ..., TN>`.

I chose `_b` because the result is a boolean, rather than the `_t` for a type in `enable_if_t`.

rok-cesnovar:
These changes look great. Thanks for cleaning all these up.
Feel free to merge once it passes. But don't close the issue the same changes need to be done to the GPU GLMs. I can handle those in the next few days.

mcol:
Thanks Bob for the explanation, I may get back to that at some point, although knowing that these are static checks makes this less pressing. I've opened #1521 for your suggestion regarding `include_summand_b`.

PR #1522, opened 2019-12-16
Add include_summand_b as replacement for include_summand<...>::value

### LINKED ISSUE #1521: get rid of the ::value in include_summand<propto, T1, ..., TN>::value
## Description

This was suggested by @bob-carpenter in https://github.com/stan-dev/math/pull/1520#issuecomment-565819683:

It would also be nice to get rid of the `::value` using the same trick as the standard library uses for `enable_if_t` with some parameter pack magic for good measure.

```
template <bool propto, typename ...Ts>
using include_summand_b = include_summand<propto, Ts...>::value;
```

Then the usage could go from `include_summand<propto, T1, ..., TN>::value` to
`include_summand_b<propto, T1, ..., TN>`.

I chose `_b` because the result is a boolean, rather than the `_t` for a type in `enable_if_t`.

mcol:
## Summary

This introduces `include_summand_b` (where the `_b` marks the fact that this function returns a boolean value) to clean up places where we've been using `include_summand<...>::value`. This was done semi-automatically with a regexp replacement, plus a few manual adjustments. Fixes #1521.

## Tests

Existing tests for `include_summand` have been changed to use `include_summand_b`. In `test/unit/math/mix/meta/include_summand_test.cpp` where I kept both side by side, as perhaps we also want to test `include_summand` directly: let me know if it's not necessary!

## Side Effects

None.

## Checklist

- [X] Math issue #1521

### DIFF
--- a/stan/math/opencl/prim/bernoulli_logit_glm_lpmf.hpp
+++ b/stan/math/opencl/prim/bernoulli_logit_glm_lpmf.hpp
@@ -71,7 +71,7 @@ return_type_t<T_alpha, T_beta> bernoulli_logit_glm_lpmf(
     return 0;
   }
 
-  if (!include_summand<propto, T_alpha, T_beta>::value) {
+  if (!include_summand_b<propto, T_alpha, T_beta>) {
     return 0;
   }
 
--- a/stan/math/opencl/prim/categorical_logit_glm_lpmf.hpp
+++ b/stan/math/opencl/prim/categorical_logit_glm_lpmf.hpp
@@ -63,7 +63,7 @@ return_type_t<T_alpha_scalar, T_beta_scalar> categorical_logit_glm_lpmf(
     return 0;
   }
 
-  if (!include_summand<propto, T_alpha_scalar, T_beta_scalar>::value) {
+  if (!include_summand_b<propto, T_alpha_scalar, T_beta_scalar>) {
     return 0;
   }
 
--- a/stan/math/opencl/prim/neg_binomial_2_log_glm_lpmf.hpp
+++ b/stan/math/opencl/prim/neg_binomial_2_log_glm_lpmf.hpp
@@ -124,7 +124,7 @@ return_type_t<T_alpha, T_beta, T_precision> neg_binomial_2_log_glm_lpmf(
       need_phi_derivative ? (need_phi_derivative_sum ? wgs : N) : 0, 1);
   const bool need_logp1 = include_summand_b<propto>;
   const bool need_logp2
-      = include_summand_b<propto, T_precision> && is_vector<T_precision>;
+      = include_summand_b<propto, T_precision> && is_vector<T_precision>::value;
   const bool need_logp3
       = include_summand_b<propto, T_alpha, T_beta, T_precision>;
   const bool need_logp4 = include_summand_b<propto, T_alpha, T_beta>;
--- a/stan/math/opencl/prim/neg_binomial_2_log_glm_lpmf.hpp
+++ b/stan/math/opencl/prim/neg_binomial_2_log_glm_lpmf.hpp
@@ -123,7 +123,7 @@ return_type_t<T_alpha, T_beta, T_precision> neg_binomial_2_log_glm_lpmf(
   matrix_cl<double> phi_derivative_cl(
       need_phi_derivative ? (need_phi_derivative_sum ? wgs : N) : 0, 1);
   const bool need_logp1 = include_summand_b<propto>;
-  const bool need_logp
+  const bool need_logp2
       = include_summand_b<propto, T_precision> && is_vector<T_precision>;
   const bool need_logp3
       = include_summand_b<propto, T_alpha, T_beta, T_precision>;
--- a/stan/math/opencl/prim/neg_binomial_2_log_glm_lpmf.hpp
+++ b/stan/math/opencl/prim/neg_binomial_2_log_glm_lpmf.hpp
@@ -90,7 +90,7 @@ return_type_t<T_alpha, T_beta, T_precision> neg_binomial_2_log_glm_lpmf(
     return 0;
   }
 
-  if (!include_summand<propto, T_alpha, T_beta, T_precision>::value) {
+  if (!include_summand_b<propto, T_alpha, T_beta, T_precision>) {
     return 0;
   }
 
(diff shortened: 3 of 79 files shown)


SteveBronder:
std tends to use _v for this, do we want to follow that?

bob-carpenter:
Yes, I just crossposted the same suggestion on the issue.  Sorry for the misleading suggestion of _b in the first place (I was thinking "boolean" and should've been thinking "value").

mcol:
I'll be happy to rename to `include_summand_v` when I get this to pass tests. At this stage, it fails on Windows with messages such as the following, which don't give me much idea on how to solve:
```
./stan/math/prim/meta/include_summand.hpp:53:16: error: template declaration of 'constexpr const bool stan::math::include_summand_b'
 constexpr bool include_summand_b = include_summand<propto, T_pack...>::value;
```
or also
```
In file included from ./stan/math/prim/mat/prob/bernoulli_logit_glm_log.hpp:5:0,
                 from ./stan/math/prim/mat.hpp:268,
                 from <command-line>:0:
./stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp: In function 'stan::return_type_t<T_x_scalar, T_alpha, T_beta> stan::math::bernoulli_logit_glm_lpmf(const T_y&, const Eigen::Matrix<T_x_scalar, T_x_rows, -1>&, const T_alpha&, const T_beta&)':
./stan/math/prim/mat/prob/bernoulli_logit_glm_lpmf.hpp:84:8: error: 'include_summand_b' was not declared in this scope
   if (!include_summand_b<propto, T_x_scalar, T_alpha, T_beta>) {
        ^
```
If anyone has a hint of where I should look, I'd be very grateful, as I can't reproduce any failures on Linux with gcc.

bob-carpenter:
The [example of is_arithmetic on cppreference](https://en.cppreference.com/w/cpp/types/is_arithmetic) has the `_v` version defined with `inline constexpr` instead of just `constexpr`, but that'd just solve linking problems.

I find g++ provides better error messages for templates than clang++, but it sounds like you're only running into this problem on Windows.  

That first message doesn't say what the error is.  Is that all there is?  In the second one, it doesn't seem to be finding the deifnition.  Does `bernoulli_logit` have the appropriate include of `include_summand`?

mcol:
I've copied a couple of errors (they are mostly the same repeated over and over) from the Jenkins logs. None of them seemed to point to anything specific. I haven't added or removed any includes here, so I would be surprised if that was the real cause.

Currently I've used this definition:
```
template <bool propto, typename... T_pack>
constexpr bool include_summand_b = include_summand<propto, T_pack...>::value;
```
Is it possible that I have to use a different version of `include_summand<>`? There are two more, the one I used seemed to me the more general.

andrjohns:
If its working on Linux but not Windows, it might be a consequence of the older gcc version available in RTools (only 4.9.3), it's caused me issues before! It also looks like others have reported [problems with constexpr in 4.9.3](https://stackoverflow.com/questions/27946241/workaround-for-gcc-4-9-constexpr-bug), not sure whether that's related

mcol:
We are not using RTool on Jenkins as far as I know. @serban-nicusor-toptal could you remind me what compiler versions we have for Windows?

[Here are some logs](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1522/5/pipeline/146) for the failed run.

rok-cesnovar:
We are using RTools on windows, so the compile is 4.9.3. 

Wouldnt just going with

`using include_summand_b = include_summand<propto, T_pack...>::value;`

also work here?

mcol:
Thanks, I thought we had a different compiler pipeline.

The alternative suggestion does not compile (g++ Linux):
```
./stan/math/prim/meta/include_summand.hpp:53:27: error: need ‘typename’ before ‘stan::math::include_summand<propto, T_pack ...>::value’ because ‘stan::math::include_summand<propto, T_pack ...>’ is a dependent scope
 using include_summand_b = include_summand<propto, T_pack...>::value;
```
I had tried a number of variation on that theme, and I didn't find any that worked (but then I also didn't fully understand the error message).

mcol:
I've installed RTools 3.5 and reproduced the error. I tried playing around with constexpr/using and haven't got anywhere. If there are no other ideas, perhaps we may have to drop this until we move to RTools 4.0.

bob-carpenter:
> I had tried a number of variation on that theme

Can you report what you tried that didn't work?  That `using` statement should have a template declaration before it.

mcol:
This is with RTools 3.5:
```
template <bool propto, typename... T_pack>
using include_summand_b = include_summand<propto, T_pack...>::value;
```
produces
```
In file included from ./stan/math/prim/meta.hpp:16:0,
                 from ./stan/math/fwd/core/fvar.hpp:4,
                 from ./stan/math/fwd/core.hpp:4,
                 from ./stan/math/fwd/scal.hpp:4,
                 from test/unit/math/fwd/meta/include_summand_test.cpp:1:
./stan/math/prim/meta/include_summand.hpp:54:27: error: need 'typename' before 'stan::math::include_summand<propto, T_pack ...>::value' because 'stan::math::include_summand<propto, T_pack ...>' is a dependent scope
 using include_summand_b = include_summand<propto, T_pack...>::value;
                           ^
```

Adding `typename` as suggested by the error message:
```
template <bool propto, typename... T_pack>
using include_summand_b = typename include_summand<propto, T_pack...>::value;
```
produces
```
./stan/math/prim/meta/include_summand.hpp: In substitution of 'template<bool propto, class ... T_pack> using include_summand_b = typename stan::math::include_summand<propto, T_pack ...>::value [with bool propto = true; T_pack = {stan::math::fvar<double>}]':
test/unit/math/fwd/meta/include_summand_test.cpp:11:3:   required from here
./stan/math/prim/meta/include_summand.hpp:54:77: error: no type named 'value' in 'struct stan::math::include_summand<true, stan::math::fvar<double> >'
 using include_summand_b = typename include_summand<propto, T_pack...>::value;
```

rok-cesnovar:
Ah yeah, there is probably not a way to do this without constexpr as the result of the expression is a boolean not a type.

mcol:
This won't happen until we move to Rtools 4.0.

PR #1524, opened 2019-12-17
Upgrade to boost 1.72.0

### LINKED ISSUE #1523: boost 1.72.0
## Description
We should move to boost 1.72.0 as the CRAN package BH just moved to this version.

## Example

Things to watch out for:

- possibly the patches to avoid overflow of lgamma are still needed (our unit tests catch this)
- possibly there are issues with the RK45 integrators, see https://www.r-project.org/nosvn/R.check/r-devel-windows-ix86+x86_64-gcc8/OncoBayes2-00install.html

This could be a consideration for @bgoodri to get StanHeaders to CRAN.

#### Current Version:
v3.0.0

rok-cesnovar:
## Summary

Upgrades to Boost 1.72 as that is what the CRAN Boost Headers package upgraded to.

I just ran the upgrade script, nothing else.

## Tests

/

## Side Effects

/

## Checklist

- [x] Math issue Closes #1523 


wds15:
Yack. MPI stuff to look into. I can do that.

rok-cesnovar:
And some gamma behavior seems to change: 

```
[ RUN      ] mathMixScalFun.gammaP
./test/unit/math/expect_near_rel.hpp:92: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(0, nan)
expect_near_rel; require items x1(i) = x2(i): gradient() grad

./test/unit/math/expect_near_rel.hpp:92: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true

expect_near_rel(0, nan)
expect_near_rel; require items x1(i) = x2(i): gradient_fvar() grad

./test/unit/math/expect_near_rel.hpp:92: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(0, nan)
expect_near_rel; require items x1(i) = x2(i): hessian() grad

./test/unit/math/expect_near_rel.hpp:92: Failure

Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
```

I will look into that. The MPI one seems to be something related to c+11 or something

wds15:
The MPI build issues should be fixed now. I needed to switch on C++11 for the boost compilation in order to have `std::shared_ptr` available which is used in the sources. I was able to compile the `mpi.so` library on our system.

wds15:
For the gamma stuff: You should look into the custom patches we had to do for 1.69.0; maybe that is still needed to carry around.

alashworth:
Hmm, it shouldn't. I was going to PR Boost 1.72 too since it would remove the need for custom patches, AFAIK looking at the backported patches to 1.69 there shouldn't be a need to keep them around. I will see if I can reproduce this on my end.

rok-cesnovar:
Yeah, I also dont recall having to do any other modifications for 1.69.

rok-cesnovar:
The problem is this example:
```
std::cout << stan::math::gamma_p(1.01, stan::math::positive_infinity()) << std::endl;
```
Boost 1.69 returns `-nan` and 1.72 returns `1`.

rok-cesnovar:
In the [Boost docs](https://www.boost.org/doc/libs/1_46_0/libs/math/doc/sf_and_dist/html/math_toolkit/special/sf_gamma/igamma.html) they reference to http://mathworld.wolfram.com/RegularizedGammaFunction.html and looking at wolfram alpha 1 is the correct result https://www.wolframalpha.com/input/?i=GammaRegularized%5B1.01%2C+0%2C+Inf%5D

So we will probably need to fix the gradient of gamma_p for these edge cases. I will do that separately.

rok-cesnovar:
Actually what I found was that the implemented gamma_p derivative is just fine and outputs the same result as 
`std::cout <<  boost::math::gamma_p_derivative( 0.5001, stan::math::positive_infinity()) << std::endl;`

Both rev/gamma_p and the boost derivative returns nan. Finite grad however returns zeros.

@bob-carpenter would you have an idea what would be the best case of handling this?

bob-carpenter:
I'd suggest just replacing the finite diff test with the explicit gradient test.  If things other than gradients work in finite diff, you can just up tolerance for that and keep the rest, but I've never seen gradients fail and Hessians work.

rok-cesnovar:
Ok this PR is now the commits from the upgrade script + https://github.com/stan-dev/math/pull/1534 (approved by Bob) + resolving a conflict in README. Ready for review.

syclik:
My phone crashed while trying to add a quick request for change. In the
read me, the links changed from https to http. Can you change that back?

Otherwise, it looks great! Thank you very much for splitting the PR into
manageable commits so the review was painless.

rok-cesnovar:
@syclik thanks. Fixed those URLs.

Issue #123, opened 2015-07-18
complex numbers with vars


syclik:
_From @bgoodri on July 18, 2015 15:10_

The following C++ program compiles but segfaults (with clang++-3.6 and g++-5)

```
#include <stan/math.hpp>
#include <iostream>
#include <complex>

int main() {
  stan::math::var x = 1;
  std::complex<stan::math::var> z = x;
  std::cout << "imaginary part is " << z.imag().val() << std::endl;
  return 0;
}
```

I think we should try to prevent that from compiling, or at least dying gracefully, but failing that I guess we should initialize the imaginary part to zero?

_Copied from original issue: stan-dev/stan#1554_

-----

There were a lot of things learned by @ChrisChiasson's effort to implement `std::complex` for Stan's autodiff variables in PR #789. Some of the important takeaways:

- The scope is for both the real and imaginary parts of `std::complex<T>` to be of type `T`. We can either autodiff the real component, the imaginary component, or the result of a function that takes a `std::complex<T>` and returns a `T`. This means we don't have to extend autodiff to deal with adding a `.grad()` function to `std::complex<T>` and we can focus on making sure `std::complex<T>` compiles and behaves appropriately.

- `std::complex<T>` where `T` is our autodiff types (e.g. `stan::math::var`, `stan::math::fvar<U>`) are explicitly unspecified (as of C++11, 08/2018). "The effect of instantiating the template complex for any other type is unspecified." https://en.cppreference.com/w/cpp/numeric/complex

- Given that `std::complex<T>` is unspecified, it is a requirement for any implementation to have tests for the basic operations of `std::complex<T>`. The list of functions can be found here: https://en.cppreference.com/w/cpp/numeric/complex. That means that we should be testing for construction, the member functions, the non-member functions, etc. It's ok to do this in small chunks, but since we're relying on unspecified behavior, we need to be able to tell when our implementation fails in the basic tests before trying to push this through Eigen functions.

[...]

PR #1525, opened 2019-12-18
Generic var templates for operators and std::iterator_trait var/fvar specialization

SteveBronder:
## Summary

This pulls out / cleans up some of the stuff in the complex branch. Namely

1. Pulls out and tidies up the templates that are added to the `var` operators
2. Adds `std::iterator_traits` specializations for `var` and `fvar`
3. Adds cmath member functions to the stan::math namespace for autodiff types

## Tests

@bob-carpenter do need additional tests for the iterators?

## Side Effects

std will have `std::iterator_trait specializations for var and fvar

## Checklist

- [x] Math issue #123 


SteveBronder:
@bob-carpenter ready for review!

bob-carpenter:
@SteveBronder Should I really be reviewing this given that I wrote some of it?

SteveBronder:
Good point! Would anyone else mind taking a look at this?

bob-carpenter:
OK.  I'll just go ahead and review today or tomorrow---it will have two of us looking at all the code.

stan/math/prim/meta/require_generics.hpp
```diff
@@ -37,7 +37,7 @@ struct is_double_or_int
  * @tparam The type to check
  */
 template <typename T>
-struct is_var_or_fvar
+struct is_autodiff
```
mcol: Such a nicer name!

stan/math/rev/core/cmath.hpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * Overloads `std::isinf` from `<cmath>` for argument-dependent
+ * lookup.
+ *
+ * @tparam T type of argument
+ * @param[in] v argument
+ * @return true if argument is infinite
+ */
+template <typename T, require_autodiff_t<T>...>
+inline bool isinf(T&& v) {
```
mcol: This could help me in #1545!

stan/math/rev/core/cmath.hpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * Overloads `std::isfinite` from `<cmath>` for argument-dependent
+ * lookup.
+ *
+ * @tparam T type of argument
+ * @param[in] v argument
+ * @return true if argument is finite
+ */
+template <typename T, require_autodiff_t<T>...>
+inline bool isfinite(T&& v) {
```
mcol: And this too! :)

stan/math/rev/mat/functor/algebra_solver_fp.hpp
```diff
@@ -71,11 +71,11 @@ struct KinsolFixedPointEnv {
       : f_(f),
         y_dummy(),
         y_(y),
+        N_(x.size()),
+        M_(y.size()),
         x_r_(x_r),
         x_i_(x_i),
         msgs_(msgs),
-        N_(x.size()),
-        M_(y.size()),
```
mcol: Well spotted!

stan/math/rev/core/operator_addition.hpp
```diff
@@ -73,11 +74,14 @@ class add_vd_vari : public op_vd_vari {
    \end{cases}
    \f]
  *
+ * @tparam Var1 value type of a var
+ * @tparam Var2 value type of a var
```
mcol: Thanks for documenting all these!

SteveBronder:
> I've read through this mainly because I happened to see how you dealt with std::isinf(), which hit me in #1545 (and I had worked around it in not such a nice way). It turned out that I think I've learned a lot of new things, so it was time well spent.

Glad to hear and thanks for looking this over!

> As you'll see, most of comments concern the ordering of typenames that doesn't match the ordering of function arguments. Of course that's not a big deal, so I leave it up to you what do about that.

I think that's v reasonable to fix up

> I totally skipped stan/math/rev/core/std_iterator_traits.hpp because I can't even pretend to know what use it has, and also stan/math/rev/scal/fun/pow.hpp because I haven't cracked std::forward yet. So you should get a second pair of more experienced eyes on this! :)

That's fair, after snooping around online the link below makes me think they are _much_ harder to implement than I originally thought. I'll probably have to rework these

https://github.com/tzlaine/stl_interfaces

SteveBronder:
> OK. I'll just go ahead and review today or tomorrow---it will have two of us looking at all the code. This is going to take some reviewing and I have to head out soon, so I can get to it tomorrow.

Worst case we had 3 sets of eyes on it now (thanks again @mcol!) so I think that's fine

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -80,16 +83,22 @@ class precomputed_gradients_vari : public vari {
  * specified value, vector of operands, and vector of partial
  * derivatives of value with respect to the operands.
  *
+ * @tparam Arith An arithmetic type
+ * @tparam VecVar A vector of vars
+ * @tparam VecArith A vector of
```
mcol: Oops...

stan/math/rev/core/operator_division.hpp
```diff
@@ -96,11 +100,15 @@ inline var operator/(const var& a, const var& b) {
...
+ * @tparam Var value type of a var
+ * @tparam Arith An arithmetic type
  * @param a Variable operand.
  * @param b Scalar operand.
  * @return Variable result of dividing the variable by the scalar.
  */
-inline var operator/(const var& a, double b) {
+template <typename Var, typename Arith, require_var_t<Var>...,
+          require_arithmetic_t<Arith>...>
+inline var operator/(Var&& a, Arith b) {
```
bob-carpenter: [optional]
The right doc for `a` is `dividend`, for `b` is `divisor`, and for the result is `quotient`.

stan/math/rev/core/operator_equal.hpp
```diff
@@ -18,35 +19,50 @@ namespace math {
...
  *
+ * @tparam Var1 value type of a var
+ * @tparam Var2 value type of a var
  * @param a First variable.
  * @param b Second variable.
  * @return True if the first variable's value is the same as the
  * second's.
  */
-inline bool operator==(const var& a, const var& b) {
+template <typename Var1, typename Var2, require_all_var_t<Var1, Var2>...>
```
bob-carpenter: [optional]
This could be taken to prim and generalized with `value_of_rec` to arbitrary autodiff and scalar types in a single definition.   That assumes `value_of_rec(1.0)` is just `1.0`---if not, we should probalby do that with `value_of_rec`, too.

Same goes for `!=` and the other comparisons (`>, <, >=, <=`)

stan/math/rev/core/operator_logical_or.hpp
```diff
@@ -10,39 +11,48 @@ namespace math {
...
  *
+ * @tparam Var1 value type of a var
+ * @tparam Var2 value type of a var
  * @param[in] x first argument
  * @param[in] y second argument
  * @return disjuntion of the argument's values
  */
-inline bool operator||(const var& x, const var& y) {
+template <typename Var1, typename Var2, require_all_var_t<Var1, Var2>...>
+inline bool operator||(Var1&& x, Var2&& y) {
```
bob-carpenter: [optional]
The boolean ops can also be generalized in `prim` for all autodiff types now.

stan/math/rev/core/operator_unary_negative.hpp
```diff
@@ -44,10 +45,12 @@ class neg_vari : public op_v_vari {
...
    \f]
  *
+ * @tparam A var autodiff type
  * @param a Argument variable.
  * @return Negation of variable.
  */
-inline var operator-(const var& a) {
+template <typename Var, require_var_t<Var>...>
+inline var operator-(Var&& a) {
   return var(new internal::neg_vari(a.vi_));
```
bob-carpenter: [optional as it wasn't in your code]
Just wanted to throw this out for the future.  When we don't return refs, we can just use the braces constructor:

```
return { new internal::neg_vari(a.vi_) };
```

stan/math/rev/core/operator_unary_not.hpp
```diff
@@ -10,10 +11,14 @@ namespace math {
  * Return the negation of the value of the argument as defined by
  * <code>!</code>.
  *
+ * @tparam A var autodiff type
  * @param[in] x argument
  * @return negation of argument value
  */
-inline bool operator!(const var& x) { return !x.val(); }
+template <typename Var, require_var_t<Var>...>
+inline bool operator!(Var&& x) {
```
bob-carpenter: Because there are no derivatives, this can also be done as a single primitive template using `value_of_rec`.

stan/math/rev/core/std_isinf.hpp
```diff
@@ -13,7 +13,8 @@ namespace std {
  * @param a Argument.
  * @return 1 if argument is infinite and 0 otherwise.
  */
-inline int isinf(const stan::math::var& a) {
+template <typename Var, stan::require_var_t<Var>...>
+inline bool isinf(Var&& a) {
```
bob-carpenter: This can also be made fully generic in `prim`.

stan/math/rev/mat/fun/gp_periodic_cov.hpp
```diff
@@ -257,7 +260,7 @@ class gp_periodic_cov_vari<T_x, double, T_l, T_p> : public vari {
         sin_2_dist_[pos] = sin(2.0 * pi_div_p * dist);
         sin_dist_sq_[pos] = sin_dist_sq;
         cov_lower_[pos] = new vari(
-            sigma_sq_d_ * std::exp(sin_dist_sq * neg_two_inv_l_sq), false);
```
bob-carpenter: I'm OK weither way with this,  but if everything's guaranteed to be `int` or `double`, there's no reason not to directly qualify the function.

stan/math/rev/scal/fun/pow.hpp
```diff
@@ -121,26 +122,29 @@ inline var pow(const var& base, const var& exponent) {
...
-    return inv(base);
+    return inv(std::forward<Var>(base));
   }
   if (exponent == -0.5) {
-    return inv_sqrt(base);
+    return inv_sqrt(std::forward<Var>(base));
   }
-  return var(new internal::pow_vd_vari(base.vi_, exponent));
+  return var(
+      new internal::pow_vd_vari(base.vi_, static_cast<double>(exponent)));
```
bob-carpenter: Even though you're allowing `exponent` to be `int` now, I don't think you need the static cast to double if the only constructor available for `pow_vd_vari` does it for you.

stan/math/rev/core/cmath.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * lookup.
+ *
+ * @tparam T type of argument
+ * @param[in] v argument
+ * @return `true` if the argument is negative
+ */
+template <typename T, require_autodiff_t<T>...>
+inline bool signbit(T&& v) {
+  using std::signbit;
+  return signbit(v.val());
```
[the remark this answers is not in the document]
SteveBronder: Yeah since fvar and var have `.val()` we can move this to `prim`. Should we also have each of these for arithmetic types or just rely on the cmath impl for those (in case someone calls `stan::math::signbit(a)` where a is double)?

If we wanted to be v smart we could break these up to have signatures for 

1. Arithmetic types
2. AD types with an arithmetic return from `.val()`
3. AD types with heavier types in `.val()`

Then 1 and 2 would pass by copy while 3 would be able to forward along / pass a reference to `.val()`

stan/math/prim/meta/require_generics.hpp
```diff
@@ -473,24 +473,22 @@ using require_any_not_fvar_t
     = require_any_not_t<is_fvar<std::decay_t<Types>>...>;
 
 template <typename T>
-using require_var_or_fvar_t = require_t<is_var_or_fvar<T>>;
+using require_autodiff_t = require_t<is_autodiff<T>>;
 
 template <typename T>
-using require_not_var_or_fvar_t = require_not_t<is_var_or_fvar<T>>;
+using require_not_autodiff_t = require_not_t<is_autodiff<T>>;
```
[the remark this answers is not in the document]
SteveBronder: > removing the `require_` prefix because that's implying a use, not a behavior

Can you clarify? I'm not sure i understand how the require part of the name is implying a use over a behaviour. The intent with the `require_` in the name is because all those `require_*`s are trying to emulate a legacy version of C++20 concepts. It's for if you want a universal reference in the function signature for all the weird value types and const combinations. In terms of use/behavior I think my answer is that I think about C++ concepts like compile time contracts (which feels like a behaviour, though idk what that word means in context here)

stan/math/rev/core/cmath.hpp
```diff
@@ -0,0 +1,119 @@
+#ifndef STAN_MATH_REV_CORE_CMATH_HPP
```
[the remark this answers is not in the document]
SteveBronder: looking at [cmath](http://www.cplusplus.com/reference/cmath/) it looks like all these are in the cmath header? For clarification you want `signbit`, `isinf`, `isfinite`, `isnan`, `copysign`, and `isnormal` to be in their own header files?

stan/math/prim/meta/require_generics.hpp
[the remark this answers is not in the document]
SteveBronder: > replace _t with _v because it denotes a boolean value, not a type; this follows the behavior of is_arithmetic_v; enable_if_t ends in _t because it denotes a type (void by default), not a value.
> For example, that'd be not_autodiff_v<T> for true if T is not an autodiff type, and usage would be, for example enable_if_t<not_autodiff_v<T>, U>.

The `require_*_t` metaprogramming stuff is essentially a convoluted / compact way to write `std::enable_if_t<...>` i.e. they also return `void`.

You can think of `require_autodiff_t` as an alias for `enable_if_t<not_autodiff_v<T>, U>`.

Also idt we can use C++ compile time constexpr objects with c++1y (anything _v like how std does) since the windows compiler we require compatibility with for Rtools has a bug for those

stan/math/prim/meta/require_generics.hpp
```diff
@@ -473,24 +473,22 @@ using require_any_not_fvar_t
...
 
 template <typename... Types>
-using require_all_var_or_fvar_t = require_all_t<is_var_or_fvar<Types>...>;
+using require_all_autodiff_t = require_all_t<is_autodiff<Types>...>;
```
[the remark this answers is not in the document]
SteveBronder: I've been throwing this back and forth with myself. I've leaned towards not doc since aliases are inlined in the docs you can see the full definition

http://mc-stan.org/math/d0/dfb/group__require__base__types.html#gac55eae5c1fc2f38a96336b5f0bc94449

stan/math/rev/core/operator_addition.hpp
```diff
@@ -73,11 +74,14 @@ class add_vd_vari : public op_vd_vari {
...
  * @param a First variable operand.
  * @param b Second variable operand.
  * @return Variable result of adding two variables.
  */
-inline var operator+(const var& a, const var& b) {
+template <typename Var1, typename Var2, require_all_var_t<Var1, Var2>...>
```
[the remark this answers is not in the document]
SteveBronder: The direct answer to this is that a function signature with `const var&&` would only accept constant rvalue types. The `Var1&&` and `Var2&&` in the function signature are universal references. Combined with `require_all_var_t` the operators here accept `var&`, `const var&`, and `var&&` (they could accept `const var&&` though idt const rvalues are a thing). It allows the caller (callee?) to do something like `auto the_sign = copysign(std::forward<FVar>(an_fvar));` to move a fvar, for example, when the call site can tell `a_fvar` is movable.

The longer indirect answer is that if we want "universal references" in function signatures for perfect forwarding then you can't have any declaration specifiers alongside the type specifier for each function parameter. 

Would we ever allow a `var` to be moved?

stan/math/rev/core/operator_divide_equal.hpp
```diff
@@ -3,16 +3,18 @@
...
 #include <stan/math/rev/core/var.hpp>
 #include <stan/math/rev/core/operator_division.hpp>
+#include <stan/math/prim/meta.hpp>
 
 namespace stan {
 namespace math {
-
-inline var& var::operator/=(const var& b) {
+template <typename Var, require_var_t<Var>...>
+inline var& var::operator/=(Var&& b) {
```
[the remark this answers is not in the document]
SteveBronder: > What is different between require_var_t<Var> and just removing the template parameter and using var instead of Var? I'd think given that it's an overload, we'd also prefer the more specific type to match.

As with the comment above about [universal references](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers) the `Var&&` and `require_var_t` combo here will match against any `var` type with whatever [deceleration specifiers](https://en.cppreference.com/w/cpp/language/declarations#Specifiers) it has on it (const etc.). So it gives us the most specific type as long as the paremeters decayed type is a `var`.

Does that make sense? ftr I find the universal reference thing in C++the most weird/legalize part of the language

stan/math/rev/core/operator_division.hpp
SteveBronder: Word!

stan/math/rev/core/operator_equal.hpp
SteveBronder: I agree but let's flag this as an issue and do it in another PR once this goes through. 

On a side note, do we think of `fvar`'s as a "heavy" type? we may want specializations that forward for `fvar`'s with `var`s or `fvar`'s inside

stan/math/rev/core/operator_unary_negative.hpp
SteveBronder: Will do!

stan/math/rev/core/operator_unary_not.hpp
SteveBronder: Yep! I think with all these it would be good to do them as a seperate PR else this is going to blast off into like a 800-1000 line change PR

stan/math/rev/mat/fun/gp_periodic_cov.hpp
SteveBronder: It feels like there's a reason i did this but I can't remember. I'll change it back and if something breaks we'll know what's up

bob-carpenter:
> looking at cmath it looks like all these are in the cmath header? For clarification you want signbit, isinf, isfinite, isnan, copysign, and isnormal to be in their own header files?

Sorry, but yes.  That's how we've laid out the other cmath function overloads.

stan/math/prim/meta/require_generics.hpp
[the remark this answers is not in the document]
bob-carpenter: I misunderstood and thought this was a boolean.  Now that I see what it's doing, I still think it needs to be renamed.  How about `enable_if_autodiff_t` to enable if it's an autodiff type?  That follows the standard library naming conventions more closely than `require_autodiff_t`.  

We can then use `disable_if_autodiff_t` instead of `enable_if_not_autodiff_t`.  

We can build similar ones for primitives, `enable_if_arithmetic_t` and `disable_if_arithmetic_t`.

SteveBronder:
> Sorry, but yes.  That's how we've laid out the other cmath function overloads.

lol wugh aight I'll get to this. They also need tests

bob-carpenter:
> +using require_all_autodiff_t = require_all_t<is_autodiff<Types>...>;

Now that I see what they're doing I think these should be enable_if_all_autodiff_t.

> ... I've leaned towards not doc since aliases are inlined in the docs you can see the full definition

Nice.  So that'll follow cppreference's definition, which is OK by me.

bob-carpenter:
var is just a pointer to vari, so can be copied by value efficient.  An fvar<T> is just two Ts held by value, not reference.  So I don't see anything to move anywhere.

bob-carpenter:
I think I get what perfect forwarding is doing after reading a couple explanations.  I should've thought through why T&& would be more general than var&&---I always forget that sometimes T matches more than just a base type---it picks up qualifiers. 

I don't think we ever need to pass non-constant references for autodiff variables or move them.  Where would the extra generality get used?  Or the other way around, is there an example where (const var&) as an argument for a var won't work?

bob-carpenter:
fvar<T> is 16 bytes and fvar<fvar<T>> is 32 bytes whether T is var or double.  But everything lives on the function call stack, so I'm not seeing what can get usefully forwarded.

bob-carpenter:
> 
> Yep! I think with all these it would be good to do them as a seperate PR else this is going to blast off into like a 800-1000 line change PR

Agreed.

bob-carpenter:
I can help write the tests if you want to give me some or all of them to do.

SteveBronder:
Actually I think the only ones we need testing for are

- `iterator_traits` for var / fvar (? idk if this needs tests but we could check we can call and declare it with types we expect
- ad types, and arithmetic for
    - signbit
    - isinf
    - isfinite
    - isnan
    - isnormal
    - copysign

I can probs get to them this weekend if you can do it earlier then :+1:

SteveBronder:
> Now that I see what they're doing I think these should be enable_if_all_autodiff_t.

I'll make a seperate issue for this

SteveBronder:
> I don't think we ever need to pass non-constant references for autodiff variables or move them.  Where would the extra generality get used?  Or the other way around, is there an example where (const var&) as an argument for a var won't work?

Tadej and I had a big discussion on ref vs const ref for Eigen matrices [here](https://github.com/stan-dev/math/pull/1344#issuecomment-530832942). There's also [this stackoverflow](https://stackoverflow.com/questions/46168394/when-to-prefer-const-lvalue-reference-over-rvalue-reference-templates) post I found interesting talking about casting rvalue's to const lvalue's in function calls. 

For `var` I think it might be a different story. Though if you look at fvar's inverse function in mat it has these eigen matrix fvars multiplied together.

`m_deriv = multiply(multiply(m_inv, m_deriv), m_inv);`

I _think_ the above stackoverflow's rvalue / const lvalue example would apply here

I have to think about this. `var` seems to be [trivially copyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable) so idk.  If `var` ever became non-trivial then these could matter.

Do non-const references hurt in some way?

I'll look over the C++ templates book tmrw, section 7 has a bunch of stuff about passing ref, const ref, etc. stuff 

> fvar<T> is 16 bytes and fvar<fvar<T>> is 32 bytes whether T is var or double.  But everything lives on the function call stack, so I'm not seeing what can get usefully forwarded.

I'm not following why everything sitting on the function call stack makes their sizes not relevant

bob-carpenter:
> I'm not following why everything sitting on the function call stack makes their sizes not relevant

Because there's no memory allocated with malloc to steal.  

Compare this with the behavior of std::vector, which allocates memory using malloc and frees it in the destructor.  When you deep copy a standard vector, it requires a malloc and copy, whereas when you move it's just an assignment.

With var, there's nothing in memory to be moved.  If I write

var a = 3.2;

A var is just a struct containing a vari*.  So a copy is just copying a pointer.

An fvar is just a struct containing two T type objects.  Again no malloc going on and there's no difference between a copy and a deep copy.

SteveBronder:
> Because there's no memory allocated with malloc to steal.

gah! brainfart alright I'll swap these back (though I'd like to leave the templates for Arith)

bob-carpenter:
I think the templates for arithmetic make sense---it cuts out a lot of code.

bob-carpenter:
Is this ready to review again?

SteveBronder:
Not yet! Sorry gonna try to get around to to tonight. Didn't have time to write tests yet for the new functions. I'll ping you when this is ready for review

SteveBronder:
@bob-carpenter this still needs the following tests if you have any spare time

1.  iterator_traits (not sure if this one is necessary)
2.  `signbit`
3. `isinf`
4.  `isfinite`
5. `isnan`
6.  `isnormal`
7. `copysign`

We've had 3 sets of eyes look at this PR. If you have any time to write the tests I'll review them and I think we'll all be on the same page to have this merged

bob-carpenter:
I'll write those ASAP.  I thought I already had tests for some of this, but if not, they'll be easy to write.

bob-carpenter:
Looks like I'm also going to have to break `cmath` down into its bits as we don't want a `cmath.hpp`.

SteveBronder:
Much appreciated!

bob-carpenter:
Don't worry---that didn't take 5 hours---lots of stuff came up in between!

SteveBronder:
Hmm, that's a really weird error I'll look into what's happening here tmrw

SteveBronder:
Actually the only way I know how to fix this 

```
./test/test-models/good/optimization/rosenbrock.hpp:177:73: warning: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second:
             lp_accum__.add(-((pow((1 - x), 2) + (100 * pow((y - pow(x, 2)), 2)))));
```

Is to make the `var pow()` use a `require` so that it get's kicked out of the choices during SFINAE. This collision happens because the right hand side is an int instead of a double&. Is that cool? If we do it for one should we go back to doing that for all of them?

SteveBronder:
Actually Bob, kind of a wild take but should we pass all these vars by value? They are 8 bytes and like a double can just fit in a word

SteveBronder:
Uh, passing by value seems very good

Running performance-tests-cmdstan with

```
./compare-git-hashes.sh "stat_comp_benchmarks -j4 --runs 1" develop develop develop 993ceaa8629c59bfe1cebab746cdc50a7e894a4d
```

```
+ ./comparePerformance.py develop_performance.csv performance.csv markdown
---RESULTS---
| stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan | 3.22 | 3.21 | 1.0 | 0.44% faster |
| stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan | 0.01 | 0.01 | 1.11 | 9.7% faster |
| stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan | 0.06 | 0.06 | 1.0 | -0.18% slower |
| stat_comp_benchmarks/benchmarks/arK/arK.stan | 1.38 | 1.35 | 1.02 | 1.85% faster |
| stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan | 4.95 | 4.85 | 1.02 | 2.1% faster |
| performance.compilation | 73.0 | 66.67 | 1.1 | 8.68% faster |
| stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan | 6.86 | 6.93 | 0.99 | -1.08% slower |
| stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan | 15.27 | 15.66 | 0.98 | -2.55% slower |
| stat_comp_benchmarks/benchmarks/sir/sir.stan | 76.34 | 85.63 | 0.89 | -12.16% slower |
| stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan | 0.23 | 0.24 | 0.98 | -2.49% slower |
| stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan | 2.19 | 2.1 | 1.04 | 3.85% faster |
| stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan | 0.15 | 0.12 | 1.22 | 17.96% faster |
| stat_comp_benchmarks/benchmarks/arma/arma.stan | 0.53 | 0.52 | 1.02 | 2.13% faster |
| stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan | 0.03 | 0.03 | 1.08 | 7.67% faster |
| stat_comp_benchmarks/benchmarks/garch/garch.stan | 0.5 | 0.39 | 1.27 | 21.37% faster |
Mean result: 1.04750942953
```

bob-carpenter:
We should be able to pass var by value.  I assume 1.05 means it's faster?  What's this testing?

SteveBronder:
This is just running the basic cmdstan performance tests that get kicked off with every PR, but locally on my home computer

bob-carpenter:
I brought this up to date with develop.  I think the last failure upstream was some kind of CI failure, not a problem with this PR.

bob-carpenter:
@SteveBronder : There's now an ambiguity in `pow(double, int)` surfacing in this PR:

```
./test/test-models/good/optimization/rosenbrock.hpp:177:31: error: call to 'pow' is ambiguous
            lp_accum__.add(-((pow((1 - x), 2) + (100 * pow((y - pow(x, 2)), 2)))));
                              ^~~
...
/usr/include/x86_64-linux-gnu/bits/mathcalls.h:153:17: note: candidate function
__MATHCALL_VEC (pow,, (_Mdouble_ __x, _Mdouble_ __y));
                ^
lib/stan_math/stan/math/rev/fun/pow.hpp:125:12: note: candidate function [with Arith = int, $1 = <>]
inline var pow(const var& base, Arith exponent) {
           ^
lib/stan_math/stan/math/rev/fun/pow.hpp:160:12: note: candidate function [with Arith = double, $1 = <>]
inline var pow(Arith base, const var& exponent) {
           ^
lib/stan_math/stan/math/rev/fun/pow.hpp:108:12: note: candidate function
inline var pow(const var& base, const var& exponent) {
           ^
```

`mathcalls.h` is coming in from the standard library math header. 

All of these require one `int` to `double` promotion, hence the ambiguity. I'm not sure why it's not matching overload (7) mentioned in this doc:

https://en.cppreference.com/w/cpp/numeric/math/pow

It seems to be on my machine because I'm not getting the ambiguity.

The ambiguity arose in this PR because the `pow` function now matches the `int` or `double` with no promotion.  

Would it be possible to replace the `var` with a template parameter that is required to be a var?  Then it wouldn't match an `int`.

We might also consider adding our own `pow(arithmetic, arithmetic)` with templates.   I'm not sure what kind of havoc that'd wreak with proper implementations of `std::pow()` overload (7) or with other implementations.

bob-carpenter:
If you don't have time for this, do you mind if I cherry pick the bits I need for `std::complex` so that I can start staging that?

bob-carpenter:
I tried the fully templating `var` approach.  At least it doesn't seem to mess up math.

bob-carpenter:
The problem now is that the test server is running out of memory.  @serban-nicusor-catena ---should we just restart the tests or does something else need to happen?

```
g++  -std=c++1y -m64 -D_REENTRANT -Wall -Wno-unused-function -Wno-uninitialized -Wno-unused-but-set-variable -Wno-unused-variable -Wno-sign-compare -Wno-unused-local-typedefs      -I lib/stan_math/lib/tbb_2019_U8/include -O3 -I src -I . -I lib/stan_math/ -I lib/stan_math/lib/eigen_3.3.3 -I lib/stan_math/lib/boost_1.72.0 -I lib/stan_math/lib/sundials_5.1.0/include -I lib/stan_math/lib/gtest_1.8.1/include -I lib/stan_math/lib/gtest_1.8.1   -D_USE_MATH_DEFINES  -DBOOST_DISABLE_ASSERTS       -c src/test/unit/lang/parser/assignment_statement_test.cpp -o test/unit/lang/parser/assignment_statement_test.o

cc1plus.exe: out of memory allocating 2097112 bytes

cc1plus.exe: out of memory allocating 1442192 bytes
make/tests:13: recipe for target 'test/unit/lang/generator/generate_idxs_test.o' failed
mingw32-make: *** [test/unit/lang/generator/generate_idxs_test.o] Error 1
mingw32-make: *** Waiting for unfinished jobs....
make/tests:13: recipe for target 'test/unit/lang/generator/generate_cpp_test.o' failed
mingw32-make: *** [test/unit/lang/generator/generate_cpp_test.o] Error 1
```

rok-cesnovar:
Argh, Nic did apply this fix https://www.intel.com/content/www/us/en/programmable/support/support-resources/knowledge-base/embedded/2016/cc1plus-exe--out-of-memory-allocating-65536-bytes.html but it seems to not do the trick. 

It only occurs occasionally on the recently added Windows machine (running in our lab in Ljubljana). Let me know @serban-nicusor-toptal if I can do something you cant remotely. Will also buy some additional RAM. 

I restarted the upstream tests https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1525/30/pipeline
If you only restart a stage Github doesnt show the yellow dot but if it will pass it will go green.

serban-nicusor-toptal:
Thanks for the help here @rok-cesnovar 
Ram and swap are more than enough I think so it may be a software configuration which doesn't let it allocate more maybe ... I'll look more into it tonight.

bob-carpenter:
Yatta!  The fully-templated version works.  This is useful as I'll be able to do this elsewhere I've been having trouble with unwanted promotion of primitives to `var`.   

 @SteveBronder or @rok-cesnovar : would one of you review my changes to `pow()` here---nothing else has changed since it's been reviewed.  I also added tests for instantiating `pow`. 

I approved the changes I was waiting for.

SteveBronder:
Sorry for the delay I'll take a look at this today

stan/math/rev/fun/pow.hpp
```diff
@@ -117,12 +117,18 @@ inline var pow(const var& base, const var& exponent) {
...
+ * @tparam Arith arithmetic type
  * @param base Base variable.
  * @param exponent Exponent scalar.
  * @return Base raised to the exponent.
  */
-template <typename Arith, require_arithmetic_t<Arith>...>
-inline var pow(const var& base, Arith exponent) {
+template <typename Var, typename Arith, require_var_t<Var>...,
+          require_arithmetic_t<Arith>...>
+inline var pow(const Var& base, Arith exponent) {
```
SteveBronder: I'm fine with merging as is. Should we make a separate issue to have all these be passed by value?

bob-carpenter: Yes, that should definitely go  in a separate issue, along with performance evaluations.

PR #1528, opened 2019-12-19
Make use of new Windows agent in Jenkins GPU steps

serban-nicusor-toptal:
Modified Jenkinsfile to check OS type and then based on the agent OS decide which stage it should run. This was done so we can offload jobs that use a GPU on the new windows instance.

## Summary

This PR will allow Jenkins to run `Headers check with OpenCL` and `Full unit with GPU` stages on the new windows machine. This was done to speedup a bit the CI/CD by offloading GPU tests to this machine too.

## Side Effects

Not that I'm aware of but @rok-cesnovar can confirm, thanks!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,6 +64,7 @@ pipeline {
                 script{
                     env.OS = checkOs()
                     env.ASSIGNED_NODE = "${NODE_NAME}"
+                    println("${NODE_NAME}")
                 }
             }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,9 +59,11 @@ pipeline {
     }
     stages {
         stage("Determine OS"){
+            agent { label "gpu" }
             steps{
                 script{
                     env.OS = checkOs()
+                    env.ASSIGNED_NODE = "${NODE_NAME}"
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
Thanks for doing this! 

To provide a bit more context, we now have another windows node, provided by our Ljubljana group. Its been running Jenkins jobs for a while now but it wasnt running GPU jobs (it has a Titan XP, donated by NVIDIA).

So I asked @serban-nicusor-toptal to add the "gpu" label Windows machine. But that doesnt work as simple because on Windows you need to add an additional line to the make/local, which means we need a separate stage.

@serban-nicusor-toptal we need to add the OPENCL_DEVICE_ID environment variable from which jenkins reads the device id. You could also hardcode it to 0 for the Windows machine.

serban-nicusor-toptal:
I was looking for `OPENCL_DEVICE_ID` value, thanks!

serban-nicusor-toptal:
I've kept testing this today and I can't get it to work, it may be a dependency version mismatch.
Tomorrow I'll compare the old windows with the new one to see if that's the case.

rok-cesnovar:
Lets close this for now and revisit this later.

PR #1530, opened 2019-12-20
reverts changes to beta binomial

SteveBronder:
## Summary

See discourse discussion [here](https://discourse.mc-stan.org/t/rstan-works-but-cmdstan-does-not-for-a-beta-binomial-model/12364/3)

It looks like I goofed up beta_binomial_lpmf with this change. 

## Checklist

- [x] Math issue #1308 

### DIFF
--- a/stan/math/prim/scal/prob/beta_binomial_lpmf.hpp
+++ b/stan/math/prim/scal/prob/beta_binomial_lpmf.hpp
@@ -41,13 +41,12 @@ return_type_t<T_size1, T_size2> beta_binomial_lpmf(const T_n& n, const T_N& N,
                                                    const T_size1& alpha,
                                                    const T_size2& beta) {
   static const char* function = "beta_binomial_lpmf";
-  using T_partials = partials_return_t<T_size1, T_size2>;
+  using T_partials_return = partials_return_t<T_size1, T_size2>;
 
-  if (size_zero(n, N, alpha, beta)) {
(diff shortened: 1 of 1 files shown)


stan/math/prim/scal/prob/beta_binomial_lpmf.hpp
```diff
@@ -56,68 +55,96 @@ return_type_t<T_size1, T_size2> beta_binomial_lpmf(const T_n& n, const T_N& N,
...
+      normalizing_constant(max_size(N, n));
+  for (size_t i = 0; i < max_size(N, n); i++)
+    if (include_summand<propto>::value)
+      normalizing_constant[i] = binomial_coefficient_log(N_vec[i], n_vec[i]);
+
+  VectorBuilder<include_summand<propto, T_size1, T_size2>::value,
+                T_partials_return, T_n, T_N, T_size1, T_size2>
+      lbeta_numerator(size);
+  for (size_t i = 0; i < size; i++)
+    if (include_summand<propto, T_size1, T_size2>::value)
```
rok-cesnovar: This line and lines 93 and 135 could be removed because in line 58 we have
```
if (!include_summand<propto, T_size1, T_size2>::value)
    return 0.0;	    return 0.0;
```
which means that if this isnt true we never reach this anyways. This was cleaned in #1520 by @mcol. 

These all get compiled out so there is no performance penalty and I would rather fix the bug first and then clean this in subsequent PR.

PR #1542, opened 2019-12-23
Bugfix/issue 1533 test ad, attempt 2

### LINKED ISSUE #1533: Some expect_ad calls fail to compile
## Description
I think that there is a problem with expect_ad, preventing certain expect_ad calls which would otherwise work as intended from compiling.

This test case for example:
```
TEST(test_unit_math_test_ad, test_ad_show_problem) {
  auto f = [](auto& x, auto& y) { return x + y(0); };
  Eigen::VectorXd v(3);
  v << 1, 2, 3;
  stan::test::expect_ad(f, 0, v);
}
```

Produces this compilation error:
```
[...]
./test/unit/math/test_ad.hpp:493:21: error: no matching function for call to ‘expect_all_throw(const test_unit_math_test_ad_test_ad_show_problem_Test::TestBody()::<lambda(auto:50&, auto:51&)>&, int&, const Eigen::Matrix<double, -1, 1>&)’
     expect_all_throw(f, x1, x2);
[...]
```

expect_ad here delegates to expect_ad_vv(tols, f, int, T2), and expect_ad_vv(tols, f, int, T2) tries to call expect_all_throw(f, int, T2), but expect_all_throw only has the following signatures:
```
void expect_all_throw(const F& f, const Eigen::VectorXd& x) {
void expect_all_throw(const F& f, double x1) {
void expect_all_throw(const F& f, double x1, double x2) {
void expect_all_throw(const F& f, double x1, double x2, double x3) {
```

I think expect_all_throw is only supposed to be called on the serialized form of the functor f?

After trying to call expect_all_throw, expect_ad_vv(tols, f, int, T2) delegates to expect_ad_vv(tols, f, T1, T2) which serializes f and delegates to expect_ad_helper, which itself calls expect_all_throw on the serialized form of f.

My suggested fix is to delete the call to expect_all_throw inside expect_ad_vv(tols, f, int, T2). This allows the test case given above to compile, and does not reduce test coverage, because expect_all_throw gets called by expect_ad_helper later. (I think.) Also, delete the corresponding expect_all_throw calls inside the other expect_ad_* variations. (The ones introduced by 5d49688eb.)

## Example
Above.

## Expected Output
expect_ad compiles in this case, test coverage is not reduced.

#### Current Version:
v3.0.0

peterwicksstringfield:
#  Summary

Fixes #1533.

I also introduced a function expect_expect_ad_failure(f, xs...) which is just a negated form of expect_ad. It succeeds if and only if expect_ad fails. This function is currently confined to test_ad_test.cpp, where it is used to verify that expect_ad catches certain types of failures. GTEST provides only lukewarm support for writing such a negated test, so I had to jump through a couple hoops.

## Tests

Test that expect_ad can be called in all the intended ways, with various combinations of int/double/whatever. Test that mismatched exceptions and improper int handling are still being caught.

## Side Effects

None.

## Checklist

- [X] Math issue #1533 

### DIFF
--- a/test/unit/math/test_ad.hpp
+++ b/test/unit/math/test_ad.hpp
@@ -487,13 +487,6 @@ void expect_ad_vv(const ad_tolerances& tols, const F& f, const T1& x1,
 
 template <typename F, typename T2>
 void expect_ad_vv(const ad_tolerances& tols, const F& f, int x1, const T2& x2) {
-  try {
-    f(x1, x2);
-  } catch (...) {
-    expect_all_throw(f, x1, x2);
-    return;
-  }
-
   double x1_dbl = static_cast<double>(x1);
 
   // expect same result with int or and cast to double
@@ -509,13 +502,6 @@ void expect_ad_vv(const ad_tolerances& tols, const F& f, int x1, const T2& x2) {
 
 template <typename F, typename T1>
 void expect_ad_vv(const ad_tolerances& tols, const F& f, const T1& x1, int x2) {
-  try {
-    f(x1, x2);
-  } catch (...) {
-    expect_all_throw(f, x1, x2);
-    return;
-  }
-
   double x2_dbl = static_cast<double>(x2);
 
   // expect same result with int or and cast to double
@@ -647,13 +633,6 @@ void expect_ad_vvv(const ad_tolerances& tols, const F& f, const T1& x1,
 template <typename F, typename T3>
 void expect_ad_vvv(const ad_tolerances& tols, const F& f, int x1, int x2,
                    const T3& x3) {
-  try {
-    f(x1, x2, x3);
-  } catch (...) {
-    expect_all_throw(f, x1, x2, x3);
-    return;
-  }
-
   double x1_dbl = static_cast<double>(x1);
   double x2_dbl = static_cast<double>(x2);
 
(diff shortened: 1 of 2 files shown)


test/unit/math/test_ad.hpp
```diff
@@ -509,18 +505,14 @@ void expect_ad_vv(const ad_tolerances& tols, const F& f, int x1, const T2& x2) {
...
 template <typename F, typename T1>
 void expect_ad_vv(const ad_tolerances& tols, const F& f, const T1& x1, int x2) {
+  double x2_dbl = static_cast<double>(x2);
+
+  // test value, exception handling checked recursively
   try {
-    f(x1, x2);
+    expect_near_rel("expect_ad_vv(T1, int)", f(x1, x2), f(x1, x2_dbl));
   } catch (...) {
-    expect_all_throw(f, x1, x2);
```
syclik: @peterwicksstringfield, thanks for adding this code. For the most part it looks ok, but I'm wondering if this is missing a test case.

It looks like with this change, if `f(x1, x2)` or `f(x1, x2_dbl)` throws, but not both, then we miss marking that as an error. (The old code had `expect_all_throw(f, x1, x2)`.)

I saw that there were some tests below, so maybe this has the right behavior and I missed it. So... question: will this fail a test if either one throws but not the other?

peterwicksstringfield: It is missing a testcase, just as you say. It will NOT fail a test if one throws and the other does not. The negated version of expect_ad in this PR is broken too.

Thank you for taking the time to review and for spotting this.

PR #1551, opened 2019-12-27
Allow kappa=0 in von_mises_lpdf and von_mises_rng

### LINKED ISSUE #564: von Mises distribution fails for kappa=0
#### Summary:
The von Mises distribution returns a non-finite value when kappa=0.  This should result in a uniform distribution and likely needs to be special cased.

#### Reproducible Steps:
The following model should produce a uniform distribution of test_var but instead, fails at initialization:

```
parameters {
    real test_var;
}
model {
    test_var ~ von_mises(0,0);
}
```

#### Current Output:
```
Initialization between (-2, 2) failed after 100 attempts. 
 Try specifying initial values, reducing ranges of constrained values, or reparameterizing the model.

```

#### Expected Output:
Should sample easily and produce a uniform distribution on test_var.

#### Current Version:
v2.15.0

mcol:
## Summary

Relaxed the boundary check in `stan/math/prim/scal/prob/von_mises_lpdf.hpp` and `von_mises_rng` for `kappa` so that 0 is a valid value. Fixes #564.

## Tests

Added tests for `kappa = 0`.

## Side Effects

None.

## Checklist

- [X] Math issue #564


mcol:
To clarify the test I've added: the Von Mises distribution for `kappa = 0` corresponds to the circular uniform distribution with density `1 / (2 * PI)` (independently of the values for `y` or `mu`). And `log(1 / (2 * PI)) = -1.837877` (the value checked in the test).

nhuurre:
Nice. I think this should have an autodiff test too? Note that computing `ops_partials.edge3_.partials_[n]` divides by `kappa`. That's easy to fix, though.

Might be worth fixing `von_mises_rng` too. The algorithm there looks stable until `rho` rounds to zero, which happens around `kappa<1e-8`. At that point the distribution is indistinguishable from uniform so no clever math needed.

mcol:
Thanks @nhuurre for your suggestions. I've fixed the expression for the derivative to work also in the case of `kappa=0`. I'll file an issue for the `von_mises_rng` when I'm done with this. Perhaps you know the answer to my question below. :)

Locally I have added a set of parameters to the distribution test (`test/prob/von_mises/von_mises_test.hpp`) for `kappa=0`, but it throws in the finite differences part because `kappa` becomes `-1e-08`. So is it correct to write a custom test using the autodiff framework instead of relying on the distribution tests? Any pointers as to where I could look for an example would be very helpful.

mcol:
I've pushed a distribution test for `kappa` as close to zero as the finite differences would allow, and another hand-coded test inspired by `test/unit/math/mix/scal/prob/normal_test.cpp`. The latter doesn't fully work, there's a bit at the end for which I get `-nan`, so there's something I must be doing wront, but to be honest I don't fully follow the autodiff intricacies.

@bob-carpenter could you have a look if you see what of those tests should be kept and what should be fixed?

mcol:
I went ahead and fixed `von_mises_rng` too (and updated the description accordingly). @nhuurre is this what you had in mind?

nhuurre:
Yes, although that 1e-8 was just an order-of-magnitude guess. Testing a bit with Python it seems the bound needs to be 1.4e-8 to ensure nonzero `rho`. Also, the main branch does some "fmod with positivity constraint" adjustment to `mu` before centering the distribution. Just copying that blindly the uniform branch would be
```
output[n]
    = (uniform_rng() - 0.5) * TWO_PI
      + std::fmod(std::fmod(mu_vec[n], TWO_PI) + TWO_PI,TWO_PI);
```

mcol:
Thanks! I'll post an updated patch after the rest of the flattening PRs go in as they may cause further conflicts.

test/unit/math/mix/scal/prob/von_mises_test.cpp
```diff
@@ -0,0 +1,87 @@
+#include <stan/math/mix/scal.hpp>
```
rok-cesnovar: This file should be test/unit/math/mix/prob/von_mises_test.cpp, without the scal.

mcol: Thanks!

mcol:
This is ready for review.

stan/math/prim/prob/von_mises_rng.hpp
```diff
@@ -26,16 +26,17 @@ namespace math {
  * Journal of the Royal Statistical Society. Series C (Applied Statistics),
  * Vol. 28, No. 2 (1979), pp. 152-157
  *
- * See licenses/stan-license.txt for Stan license.
+ * For kappa < 1.4e-8, this reduces to a uniform distribution.
```
bob-carpenter: Thanks for both of these secondary doc fixes (plus the main ones).

Issue #1425, opened 2019-10-28
Use Eigen::Map to replace arr functions


andrjohns:
## Description
As initially discussed in [this PR](https://github.com/stan-dev/math/pull/1373#issuecomment-541424024), there is some code duplication in having the same functions implemented for ```std::vector``` and ```Eigen``` separately. This could be avoided by using a wrapper which compiles to a no-op for ```Eigen``` types and returns an ```Eigen::Map``` for ```std::vector``` types prior to calling the ```mat``` definition:

```
template <typename Derived>
const auto& as_eigen(const Eigen::MatrixBase<Derived>& v) {
  return v;
}

template <typename T>
const auto as_eigen(const std::vector<T>& v) {
  return Eigen::Map<const Eigen::Matrix<T, Eigen::Dynamic, 1>>(v.data(),
                                                               v.size());
}
```

This also means that the templating for the ```mat``` definitions needs to be changed from ```Eigen::Matrix<double, R, C>``` to ```Eigen::MatrixBase<Derived>``` so that it works with ```Eigen::Map``` inputs.

After some initial testing, the speed gains are minor:
```
2019-10-28 20:40:26
Running ./log_sum_exp_arr
Run on (8 X 4000 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
-----------------------------------------------------
Benchmark              Time           CPU Iterations
-----------------------------------------------------
LogSumExp_Old    8175394 ns    8175190 ns         92
LogSumExp_New    5511970 ns    5511907 ns        124
DotSelf_Old      5136511 ns    5136391 ns        132
DotSelf_New      4497041 ns    4496989 ns        157
Sum_Old          4868005 ns    4867939 ns        139
Sum_New          4608184 ns    4608127 ns        149
```

Benchmarking code [here](https://github.com/andrjohns/perf-math/blob/feature/tbb_test/log_sum_exp_arr.cpp)

The main benefits would be in code reduction for functions with existing ```arr``` definitions, as well as making it easier to extend other functions to take ```std::vector``` inputs (without requiring additional definitions).

Does this seem like a worthwhile change?

#### Current Version:
v3.0.0

rok-cesnovar:
[...]

PR #1558, opened 2019-12-29
Generalised unary vector function framework

andrjohns:
## Summary

This pull introduces a framework for generalising unary vector functions (e.g. ```log_sum_exp``` or ```log_softmax```) to work with Eigen column and row vectors, ```std::vectors```, and containers of these.

The motivations behind the framework were discussed in #1425, and the design document for the framework [is here](https://github.com/stan-dev/design-docs/pull/12). There has also been additional discussion [over in the forums](https://discourse.mc-stan.org/t/vectorised-ad-testing-nested-containers/12345/5).

A new vectorisation framework, ```apply_vector_unary```, is proposed. The framework has two functions to address the different types of vector functions:

- ```apply_vector_unary<T>::apply()``` for:  ```f(vector) -> vector```
- ```apply_vector_unary<T>::reduce()``` for:  ```f(vector) -> scalar```

This pull also introduces an example use of each type of this vectorisation:
- ```log_softmax``` (```apply()```)
- ```log_sum_exp``` (```reduce()```)

For each of these functions, their ```prim```, ```rev```, and ```fwd``` definitions have been re-written in the new framework, and their associated tests expanded.

## Tests

Both prim and mix tests have been added to check that the same values and derivatives are returned across the new vector inputs.

## Side Effects

I had to update the templating in one header from ```Eigen::Matrix<T, R, C>``` to ```Eigen::MatrixBase<Derived>``` so that they could work with ```Eigen::Map``` inputs:

- ```matrix_vari```

## Checklist

- [x] Math issue #1425 


andrjohns:
I've removed the ```apply_scalar``` and ```head``` implementations, since they don't line up with the expected functionality of ```head``` in the integration tests. Under ```apply_scalar```, when passing head a container of vectors (e.g. ```head(std::vector<VectorXd>, 2)```), the framework would apply ```head``` to each of those vectors in the container. However, the integration tests expect that ```head(std::vector<VectorXd>, 2)``` would return the first two vectors in the container (in full).

This was me misunderstanding the implementation of ```head``` and so I've not touched it in this PR.

stan/math/rev/mat/fun/log_sum_exp.hpp
```diff
@@ -14,27 +14,12 @@ namespace math {
...
-}
-
 class log_sum_exp_matrix_vari : public op_matrix_vari {
  public:
-  template <int R, int C>
-  explicit log_sum_exp_matrix_vari(const Eigen::Matrix<var, R, C>& x)
-      : op_matrix_vari(log_sum_exp_as_double(x), x) {}
+  template <typename T>
+  explicit log_sum_exp_matrix_vari(T&& x)
+      : op_matrix_vari(log_sum_exp(std::forward<T>(x).val()),
```
t4c1: This call to `std::forward` does nothing as as you only call a membor function on forwarded `x` (forwarding only makes sense if you use it when calling a function with different overloads for rvalues and lvalues or one that accepts a forwarding reference - `T&&`). If you actually forwarded one variable into two functions/classes it would not work correctly for rvalues as the first one would move content the variable and the second one would only get empty container!

stan/math/fwd/mat/fun/log_softmax.hpp
```diff
@@ -6,44 +6,47 @@
...
+ * Return the log softmax of the specified vector or container of vectors.
+ *
+ * @tparam T Type of input vector or matrix.
+ * @param[in] x Unconstrained input vector.
+ * @return Softmax of the input.
+ * @throw std::domain_error If the input vector is size 0.
+ */
+template <typename T, require_t<is_fvar<scalar_type_t<T>>>...>
+inline auto log_softmax(T&& x) {
+  return apply_vector_unary<T>::apply(std::forward<T>(x), [&](auto& alpha) {
```
t4c1: Perfect forwarding brings no benefits here. This would work as well with no additional copys:

```

**Comment by rok-cesnovar** [comment]
@andrjohns I will make a PR to your fork to fix these merge conflicts. This wont be completely trivial changes so I think a PR is needed. I missed that you are touching fwd files, else I would have hold off on the flattening.

**Comment by andrjohns** [comment]
Thanks for sorting out those conflicts @rok-cesnovar!

@t4c1 I'll update this PR (and the tests) once #1471 gets merged in. Thanks!

File: stan/math/fwd/mat/fun/log_softmax.hpp (hunk 2 above)
Comment (SteveBronder) [reply stan/math/fwd/mat/fun/log_softmax.hpp]: > Perfect forwarding is better than just const refs only if a variable can outlive the function in question. For example if the function can directley return the variable or construct some object that holds the variable.

Can you link me to a doc that talks about that? My understanding of PF is that it tells the compiler "I'm okay with moving ownership of this memory and leaving the original object in an undefined state"

In particular, for the below think about [a recursive function that can have a long ref stack that the compiler probably can't sort through](https://github.com/stan-dev/math/blob/develop/stan/math/prim/arr/fun/vec_concat.hpp#L29). PF gives the compiler the ability to move ownership of that memory. So we don't have to worry about whether the compiler will have alias issues when deciding to inline recurses.

> For example if the function can directly return the variable or construct some object that holds the variable.

The CPP guides have a lot on this with a nice little table in F.15 has some good heuristics

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#fcall-parameter-passing

In particular this seems to recommend Andrew keep it as a forwarding reference

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f19-for-forward-parameters-pass-by-tp-and-only-stdforward-the-parameter

It feels like the above func is just something we want to allow the compiler to move through if it's able

File: stan/math/prim/mat/vectorize/apply_vector_unary.hpp [hunk 3]
```diff
@@ -0,0 +1,158 @@
...
+ * the same vector type as the input is returned.
+ *
+ * Two taxonomies of unary vector functions are implemented:
+ * - f(vector) -> vector
+ * - f(vector) -> scalar
+ *
+ * This base template class takes (and returns) Eigen expression templates.
+ */
+template <typename T>
+struct apply_vector_unary<T, require_eigen_t<T>> {
```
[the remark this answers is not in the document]
andrjohns: Yep, this should also accept full matrices (which makes me think the name should be different, like ```apply_container_unary``` or something), since it allows for the easy extension of matrix functions to accept arrays of matrices. For example, this could be used to extend ```cholesky_decompose``` to also take ```std::vector<MatrixXd>``` inputs with minimal change to the code

stan/math/prim/mat/vectorize/apply_vector_unary.hpp
```diff
@@ -0,0 +1,158 @@
...
+   * @tparam F Type of functor to apply.
+   * @param x std::vector input to which operation is applied.
+   * @param f functor to apply to vector input.
+   * @return std::vector with result of applying functor to input.
+   */
+  template <typename F>
+  static inline std::vector<T_vt> apply(const T& x, const F& f) {
+    Eigen::Matrix<T_vt, -1, 1> result
+        = apply_vector_unary<T_map>::apply(as_column_vector_or_scalar(x), f);
+    return std::vector<T_vt>(result.data(), result.data() + result.size());
```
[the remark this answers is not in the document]
andrjohns: Good call, don't know how I didn't think of this

stan/math/rev/core/matrix_vari.hpp
```diff
@@ -15,13 +17,12 @@ class op_matrix_vari : public vari {
...
-  template <int R, int C>
-  op_matrix_vari(double f, const Eigen::Matrix<var, R, C>& vs)
+  template <typename Derived, require_t<is_var<typename Derived::Scalar>>...>
+  op_matrix_vari(double f, const Eigen::MatrixBase<Derived>& vs)
       : vari(f), size_(vs.size()) {
-    vis_ = reinterpret_cast<vari**>(operator new(sizeof(vari*) * vs.size()));
-    for (int i = 0; i < vs.size(); ++i) {
-      vis_[i] = vs(i).vi_;
-    }
+    vis_ = ChainableStack::instance_->memalloc_.alloc_array<vari*>(size_);
```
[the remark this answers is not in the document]
andrjohns: This was a TODO in the header, figured I'd fix while I was there (similar changes have been made in the ```log_softmax``` header)

stan/math/fwd/mat/fun/log_sum_exp.hpp
```diff
@@ -2,19 +2,39 @@
...
+ * \max(x))\f$.
+ *
+ * @tparam T Type of input vector or matrix.
+ * @param[in] x Matrix of specified values.
+ * @return The log of the sum of the exponentiated vector values.
+ */
+template <typename T, require_t<is_fvar<scalar_type_t<T>>>...>
+inline auto log_sum_exp(T&& x) {
+  return apply_vector_unary<T>::reduce(std::forward<T>(x), [&](auto& v) {
+    using T_fvar_inner = typename value_type_t<decltype(v)>::Scalar;
```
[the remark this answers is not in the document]
andrjohns: This won't work, since ```scalar_type_t<fvar<T>>``` returns ```fvar<T>```, whereas we need to return ```T```

stan/math/fwd/fun/log_sum_exp.hpp
```diff
@@ -33,37 +32,35 @@ inline fvar<T> log_sum_exp(double x1, const fvar<T>& x2) {
...
-  }
-  T deriv(0.0);
-  T denominator(0.0);
-  for (size_t i = 0; i < v.size(); ++i) {
-    T exp_vi = exp(vals[i]);
-    denominator += exp_vi;
-    deriv += v[i].d_ * exp_vi;
-  }
-  return fvar<T>(log_sum_exp(vals), deriv / denominator);
+  return log_sum_exp(x2, x1);
```
andrjohns: I've removed a redundant definition here (and in the rev) header. Originally, there was a definition for ```log_sum_exp(const fvar<T>& x1, double x2)``` and ```log_sum_exp(double x1, const fvar<T>& x2)```, but we can just have one definition and change the order of arguments as needed.

andrjohns:
@t4c1 This is ready for another look-over

stan/math/fwd/mat/fun/log_softmax.hpp
t4c1: > Can you link me to a doc that talks about that?

Sorry, that is just my understanding and I a quick search does not find anything on whet perfect forwarding is benefitial and when it is not. I can't think of any other example, but please correct me if I am wrong.

> My understanding of PF is that it tells the compiler "I'm okay with moving ownership of this memory and leaving the original object in an undefined state"

I would say more accurate would be "I am CAPABLE of either moving a temporary object or using a reference to lvalue object." In this PR no function can actually move the parameter, so const refs are just as good.

>PF gives the compiler the ability to move ownership of that memory. So we don't have to worry about whether the compiler will have alias issues when deciding to inline recurses.

I don't know anything about that. Can you link any doc or benchmark that supports this?

Thanks for the links to the guidelines. However I disagree with your conclusion that they suggest we should use perfect forwarding. The first link points to section that says simple means of parameter passing (which exclude perfect forwarding) should be used, unless benefits of something more advanced are demonstrated.

The second link suggests nothing on whether perfect forwarding should be used or not. It only explains what a function accepting a forwarding reference should do with it. It says that a function should be flagged (as violatin those guidelines) if it does anything with this argument except `std::forward`-ing it exactley once. 

I would add to this that `std::forward`-ing an argument to function is only beneficial if that function either uses perfect forwarding or has two overloads accepting const lvalue ref and rvalue ref.

stan/math/rev/fun/log_sum_exp.hpp
```diff
@@ -69,70 +57,16 @@ inline var log_sum_exp(const var& a, double b) {
...
-}
-
 class log_sum_exp_matrix_vari : public op_matrix_vari {
  public:
-  template <int R, int C>
-  explicit log_sum_exp_matrix_vari(const Eigen::Matrix<var, R, C>& x)
-      : op_matrix_vari(log_sum_exp_as_double(x), x) {}
+  template <typename T>
+  explicit log_sum_exp_matrix_vari(T&& x)
+      : op_matrix_vari(log_sum_exp(x.val()), std::forward<T>(x)) {}
```
t4c1: Perfect forwarding to a function that accepts a const& is pointless. Const ref should be used here as well.

stan/math/rev/core/matrix_vari.hpp
```diff
@@ -15,13 +16,11 @@ class op_matrix_vari : public vari {
...
  public:
-  template <int R, int C>
-  op_matrix_vari(double f, const Eigen::Matrix<var, R, C>& vs)
-      : vari(f), size_(vs.size()) {
-    vis_ = reinterpret_cast<vari**>(operator new(sizeof(vari*) * vs.size()));
-    for (int i = 0; i < vs.size(); ++i) {
-      vis_[i] = vs(i).vi_;
-    }
+  template <typename T, require_eigen_vt<is_var, T>...>
+  op_matrix_vari(double f, T&& vs) : vari(f), size_(vs.size()) {
```
t4c1:
```suggestion
  op_matrix_vari(double f, const T& vs) : vari(f), size_(vs.size()) {
```

stan/math/fwd/mat/fun/log_softmax.hpp
bob-carpenter: I agree that we shouldn't clutter the code with perfect forwarding examples where they can't be used.  This came up in another PR with autodiff types, which have no data content to move.

SteveBronder: @t4c1 at work but ill link some stuff tonight 

Maybe we should write up an rfc to talk about what types and sizes should use certain argument types. I tend to be way to pf giddy

andrjohns:
@rok-cesnovar I'm guessing this means that the segfaults have been fixed? Is this safe to merge?

SteveBronder:
I think either way we need to wait until next release since we are in code freeze atm

andrjohns:
Good catch, forgot about that part

rok-cesnovar:
Yes, the segfault were resolved (see https://github.com/stan-dev/math/issues/1615), but we have to wait until Sunday/Monday to merge after the release.

PR #1563, opened 2019-12-29
Flatten /prob tests

rok-cesnovar:
## Summary

Next in line for the flatten PRs. This one flattens the `{prim,rev,mix,fwd}/*/prob` tests to {prim,rev,mix,fwd}/prob. It also builds a master header file.

There were around 20 tests I had to merge, other changes are just moves.

## Tests

This PR adds or modifies no tests, just merges the files.

## Side Effects

/

## Checklist

- [x] Math issue #937 


test/unit/math/prim/prob/poisson_log_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+TEST(ProbPoisson, log_matches_lpmf) {
+  int y = 3;
+  double lambda = 2.3;
+
+  EXPECT_FLOAT_EQ((stan::math::poisson_lpmf(y, lambda)),
+                  (stan::math::poisson_log(y, lambda)));
+  EXPECT_FLOAT_EQ((stan::math::poisson_lpmf<true>(y, lambda)),
+                  (stan::math::poisson_log<true>(y, lambda)));
+  EXPECT_FLOAT_EQ((stan::math::poisson_lpmf<false>(y, lambda)),
+                  (stan::math::poisson_log<false>(y, lambda)));
```
mcol: This block is in the wrong place. It's very confusing, but we have tests for `poisson_log_lpmf` (Poisson with log parametrization) and tests for `poisson_log` as the old alias for `poisson_lpmf`. Since with this patch the convention is to call tests as `distribution_name_test.cpp`, `poisson_log_test.cpp` should be about `poisson_log_lpmf` and family.

So this block should probably go into `poisson_test`, although doing so won't match what's done for other `log_matches_lpmf`-type tests that are in separate files. Let's put it at the very end of that file, similar as to what we'll do for `neg_binomial_2`.

test/unit/math/prim/prob/scaled_inv_chi_square_test.cpp
```diff
@@ -1,11 +1,52 @@
+
+#include <stan/math/prim/mat.hpp>
```
mcol: Note extra empty line.

test/unit/math/prim/prob/normal_test.cpp
```diff
@@ -1,7 +1,11 @@
-#include <gtest/gtest.h>
-#include <boost/math/distributions.hpp>
+
 #include <stan/math/prim/mat.hpp>
```
mcol: Extra white line.

test/unit/math/prim/prob/neg_binomial_2_log_test.cpp
```diff
@@ -1,9 +1,42 @@
...
+
+TEST(ProbDistributionsNegativeBinomial2Log, errorCheck) {
+  check_dist_throws_all_types(NegativeBinomial2LogTestRig());
+}
+
+TEST(ProbDistributionsNegativeBinomial2Log, distributionCheck) {
+  check_counts_real_real(NegativeBinomial2LogTestRig());
+}
+
 TEST(ProbDistributionsNegBinomial2Log, error_check) {
```
mcol: There's a pre-existing inaccuracy in this file (`neg_binomial_2_log_test.cpp`): at the end (past the part we can see here) there's a `log_matches_lpmf` block that has tests for `neg_binomial_2_lpmf`, but this file is meant for `neg_binomial_2_log_lpmf`. So, it'd be nice if you moved that block into `neg_binomial2_test.cpp`.

mcol:
Also, I think I missed why we should have a prob.hpp master file. I think you've used it only a handful of times to replace where a specific prob-function header was included, but I don't see any advantages in that, as we already have `prim/scal.hpp`. Could you clarify that for me?

test/unit/math/prim/prob/poisson_log_test.cpp
rok-cesnovar: Agree. Done.

test/unit/math/prim/prob/neg_binomial_2_log_test.cpp
rok-cesnovar: Done. I move the `NegativeBinomial2LogTestRig` in a separate header file and include that header in both test files to avoid duplication.

rok-cesnovar:
> Also, I think I missed why we should have a prob.hpp master file. I think you've used it only a handful of times to replace where a specific prob-function header was included, but I don't see any advantages in that, as we already have `prim/scal.hpp`. Could you clarify that for me?

I created `prob.hpp` for future use here. After all these flattening PRs will get through, we will remove `prim/scal.hpp`, `prim/arr.hpp`, etc.
They will be replaced by `prim/prob.hpp`, `err.hpp`, `fun.hpp` etc. and a master header file `stan/math/prim/prim.hpp` or `stan/math/prim.hpp`. We havent discussed the location of the master header file so dont want to assume anything.

test/unit/math/prim/prob/neg_binomial_2_log_test.cpp
mcol: Nice!

PR #1567, opened 2019-12-30
Cleanup the use of find in makefile

rok-cesnovar:
## Summary

This PR fixes the Stan Math side of https://github.com/stan-dev/cmdstan/issues/786

The issue is that our makefiles rely on the gnu `find` command that is not available on Windows. 

~~This PR introduces a makefile subroutine that calls `dir` on Windows and find on Unix systems.~~

~~ifeq ($(OS),Windows_NT)~~
~~findfiles=$(shell cmd.exe /C "dir /b /s $(subst /,\,$1)\$2 2>NUL")~~
~~else~~
~~findfiles=$(shell find $1 -type f -name $2)~~
~~endif~~

~~The 2>NUL is there because if you search for files that dont exist, `dir `will output a `File not found` error. So if you called `make clean` twice in a row it would output a bunch of `File not found`s.~~

This PR introduces a findfiles function that is a recursive version of the makefiles wildcard function. 
`findfiles=$(foreach d,$(wildcard $(addsuffix *,$(1))),$(call findfiles,$(d)/,$(2)) $(filter $(subst *,%,$(2)),$(d)))`

This PR also excludes the generaton of distribution tests on Windows. This means that `make generate-tests` will not be available on Windows by default. This functionality is currently not working on Windows and just produces a bunch of warnings. Making this functionality available on Windows should be addressed separately. That requires a larger overhaul and since its a functionality aimed at developers its not something that needs to be addressed before the next release. The unnecessary warnings seen by users compiling models is a priority in my opinion.

## Tests

/

## Side Effects

/


rok-cesnovar:
Now that the Windows tests actually try to run tests that need kinsol we need to add that to the Path variable in Jenkins.

Issue #1578, opened 2020-01-03
The derivatives of neg_binomial_lpmf are incorrect for large alpha


martinmodrak:
## Description
The `neg_binomial_lpmf(n, alpha, beta)` is approximated by Poisson for `alpha > 1e10`. While this approximation is OK for the density, the derivatives get messed up.

## Example
Analytically, the derivative wrt. `alpha` for `n = 0` is `log(beta / (1 + beta))`. So, for `beta = 0.8` the derivative should be roughly `-0.81093` irrespective of `alpha`. However, when we set `alpha = 1e10 + 1`, the current implementation evaluates the gradient as `-1.25`.

Similarly (and perhaps more worrying) the derivative wrt. `beta` for `n = 0` is `(alpha * (1 + beta) *(-(beta/(1 + beta)^2) + 1/(1 + beta)))/beta`, so for `beta = 0.8`  and `alpha = 1e10 + 1` the analytic derivative is `6.94e9` while the current implementation in Stan returns `1.56e10`.

(see bottom of https://www.wolframcloud.com/obj/martin.modrak/Published/NegBinomial_test.nb for Mathematica code I used to obtain those).

## Expected Output
I wrote a test that compares the derivatives to those computed by complex step. Unfortunately Mathematica (at least the free cloud version) refuses to compute exact derivatives for `n > 0` and `alpha = 1e10`, but the complex step results match exact solutions for `n = 0` and the test passes for values just below the `alpha` cutoff, so I believe, this is a true failure. I would expect the test to pass. 

#### Current Version:
v3.0.0

PR #1579, opened 2020-01-03
[On hold] New formulas for derivatives of neg_binomial_lpmf after alpha cutoff

martinmodrak:
## Summary

*Currently on-hold. If PR #1497 gets merged, `neg_binomial_lpmf` should probably be rewritten using the same approach*

The formulas for derivatives of `neg_binomial_lpmf` after delegating to poisson (for `alpha > 1e10`) are replaced with Taylor expansions of the exact derivatives for `alpha -> Inf` obtained in Mathematica via:

```
     nb[n_,alpha_,beta_]:= LogGamma[n + alpha] - LogGamma[n + 1] - 
        LogGamma[alpha ] + alpha * Log[beta/ (1 + beta)] - n * Log[1 + beta];
      nbdalpha[n_,alpha_,beta_]= D[nb[n, alpha, beta],alpha];
      nbdbeta[n_,alpha_,beta_]= D[nb[n, alpha, beta],beta];
      Series[nbdalpha[n, alpha, beta],{alpha, Infinity, 1}]
      Series[nbdbeta[n, alpha, beta],{alpha, Infinity, 0}]
```

(the lowest order of the series that let's me pass the tests was chosen)

## Tests

Tests are adapted from #1497 where new tests for `neg_binomial_2_lpmf` were written. Namely

- Created a set of test values and derivatives in Mathematica (Wolfram Cloud) and test against those
- Testing the derivatives against complex step derivatives. This is important for large `alpha`, as Mathematica refuses to compute exact derivatives in this case.
- Over a grid of `beta` and `n` values test whether the change in the derivatives of `neg_binomial_lpmf` at `alpha_cutoff +/- 1e-8` is negligible.
- Test whether there is no drastic change in `neg_binomial_lpmf<propto=true, int, var, double>` between `alpha` just below and just above the cutoff.

## Side Effects

Not any.

## Checklist

- [x] Math issue #1578 


martinmodrak:
@bob-carpenter May I ask for review on this? The tests here are similar to those written for #1497 (although the problem addressed is different).

stan/math/prim/prob/neg_binomial_lpmf.hpp
```diff
@@ -15,6 +15,12 @@
 namespace stan {
 namespace math {
 
+namespace internal {
+// Exposing to let me use in tests
```
bob-carpenter: Comments shouldn't be first person.  I'd suggest just eliminating line 19.

Line 20 should say what the value is for---in this case, the threshold beyond which a Poisson approximation is used.  It's also OK to say why it's 1e10, but I couldn't follow the logic of why it's set to 1e10.

stan/math/prim/prob/neg_binomial_lpmf.hpp
```diff
@@ -102,19 +108,30 @@ return_type_t<T_shape, T_inv_scale> neg_binomial_lpmf(const T_n& n,
   }
 
   for (size_t i = 0; i < max_size_seq_view; i++) {
-    if (alpha_vec[i] > 1e10) {  // reduces numerically to Poisson
+    if (alpha_vec[i] > internal::neg_binomial_alpha_cutoff) {
+      // reduces numerically to Poisson
+      // The derivatives are obtained via Taylor series at alpha -> Inf
+      // via Mathematica as:
+      // nb[n_,alpha_,beta_]:= LogGamma[n + alpha] - LogGamma[n + 1] -
```
bob-carpenter: I don't think we should assume the readers of comments understand mathematica.  So this should be put in Stan notation or just eliminated.  I couldn't understand what `LogGamma` is or `Log` or `D`.  If these comments stay, the functions used need to be explained.

stan/math/prim/prob/neg_binomial_lpmf.hpp
syclik: @bob-carpenter, would it be ok if L20 said something like:
``` // Set to 1e10 based on empirical evaluation. At 1e9, the approximation isn't close enough```

bob-carpenter:
> 
> @bob-carpenter, would it be ok if L20 said something like:
> // Set to 1e10 based on empirical evaluation. At 1e9, the approximation isn't close enough

Yes, that's clear.

martinmodrak:
Thx @syclik and @bob-carpenter for the notes. I should've marked this PR as "on hold" because if #1497 turns out to be on the right track, I would probably rewrite `neg_binomial_lpmf` in the same spirit (eliminating the Poisson branch completely). So sorry for wasting your time here.

syclik:
@martinmodrak, do you want to close this PR? #1497 was merged and from reading through the PR, it seems like you are favoring a different implementation.

martinmodrak:
Thanks for reminidng me, yes this should be closed.

Issue #1611, opened 2020-01-13
lbeta is unstable when the arguments differ a lot


martinmodrak:
## Description
When the arguments are on wildly different scales, `lbeta` suffers from catastrophic cancellation.

## Example + Expected output

https://www.wolframalpha.com/input/?i=LogGamma%2810%5E20%29+%2B+LogGamma%283%29+-+LogGamma%2810%5E20+%2B+3%29
```
lbeta(1e20, 3)
Stan:         0
Mathamatica: -137.46195839908279
```
https://www.wolframalpha.com/input/?i=LogGamma%283*10%5E15%29+%2B+LogGamma%2812895%29+-+LogGamma%283*10%5E15+%2B+12895%29
```
lbeta(3e15, 12895)
Stan:        -350384
Mathamatica: -350396.98895556212
```

#### Current Version:
v3.0.0

bob-carpenter:
Nice catch.

Is there a known fix here?  We're currently using this:

```
lgamma(a) + lgamma(b) - lgamma(a + b);
```

martinmodrak:
Fixed along #1592 in #1497 (if it is preferable, I might split this fix from the #1497 PR into a separate one , but so far it is IMHO all related as failures in `neg_binomial_2_lpmf` motivated the improvements for `binomial_coefficient_log` and `lbeta`).

It is a neat trick I found in the R codebase - you first compute the `lbeta` via Stirling approximation to `lgamma` which allows you to analytically manipulate the expression and avoid the cancellation for the "bulk" of the value. Then you compute the difference between `lgamma` and the Stirling approximation and add this to the result to restore precision.

The code in R for the difference between `lgamma` and Stirling was opaque and used some tricks I didn't really get, but it turns out just using 3 additional terms of the Stirling series as found on Wikipedia is enough for all the tests I could come up with :-)

mcol:
Assuming it doesn't take an inordinate amout of effort, having it separate makes it easier to review, and will help if in the future this needs to be revisited for whatever reason.

bob-carpenter:
Thanks for the explanation.  This numerics stuff is tricky.

martinmodrak:
So, I separated this into PR #1612, should be ready for review.

PR #1612, opened 2020-01-14
Fix lbeta for large arguments

martinmodrak:
## Summary
Reimplementing `lbeta` to be stable when one argument is large and other small.

It is a neat trick I found in the R codebase - you first compute an approximation to `lbeta` via Stirling approximation to `lgamma` which allows you to analytically manipulate the expression and avoid catastrophic cancellation for the "bulk" of the value. Then you compute the difference between `lgamma` and the Stirling approximation (the function `lgamma_stirling_diff`) and add this to the result to restore precision.

The code in R for the difference between `lgamma` and Stirling was opaque and used some tricks I didn't really get, but it turns out just using 3 additional terms of the Stirling series as found on Wikipedia is enough for all the tests I could come up with :-)

**Possible issues:**
- Not sure if `lgamma_stirling_diff` is a good name for the function.
- The R code that inspired this implementation (https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/nmath/lbeta.c) is GPLv2. While I reimplemented the code from scratch, some portions of `lbeta.hpp` are quite similar to the R code as there are only so many ways you can rearrange the Stirling approximation to get stable results. I believe this should not be a copyright issue, but I am not a lawyer. 

## Tests

- A test against values and derivatives precomputed in Mathematica
- A test against several identities 
- A test for extreme values, infinity
- A test for stability of value around the cutoff for using the Stirling approximation

## Side Effects

Hopefully none. I have not done any performance comparison of the old vs. new implementation. I can imagine the new is faster (can avoid calls to `lgamma`) or slower (needs quite a bunch of arithmetic operations).

I did not touch the way gradients are computed (as those look mostly OK, although for very small arguments I needed to increase the test tolerance to `1e-7`).

## Checklist

- [x] Math issue #1611 

- [x] Institute of Microbiology of the Czech Academy of Sciences


martinmodrak:
In #1579 @bob-carpenter inquired about Mathematica code. I will address it here, as this PR needs to be merged before #1579. 

>Also, have you checked that the IP is OK for us to include Mathematica code in comments and use the output of their approximations? I can't see why it wouldn't be, but it'd be good to check before we go too far down the road of using a proprietary software package.

I am not a lawyer, but the code is IMHO easy: I've written the code, I (or my employer) have the IP - this has nothing to do with what you do with the code after it's written.

For ouptut of software, the case is less clear cut, but AFAIK the important part is who did a "creative act" in creating the outputs. Referring to for example [this source](https://www.lexology.com/library/detail.aspx?g=eb52567a-2104-4034-8daf-bf6c5cb9cd45) (but others agree): If the user provides non-trivial inputs to a software, they can usually claim copyright to its outputs (the same way Microsoft cannot claim copyright over stuff written in Word / Excel or Adobe cannot claim copyright to images created in Illustrator). In some cases, just running a software does not give you copyright over the results - for example for software with no inputs designed to produce the outputs. While there is an interesting little explored field over stuff like AI-created artwork, I think we are safely on the good side of the law - my input to Mathematica was creative and non-trivial.

test/unit/math/prim/fun/lbeta_test.cpp
```diff
@@ -21,3 +24,79 @@ TEST(MathFunctions, lbeta_nan) {
...
+    msg << "inv: x = " << x;
+    expect_near_rel(msg.str(), lbeta(x, 1.0), -log(x));
+  }
+}
+
+TEST(MathFunctions, lbeta_stirling_cutoff) {
+  using stan::test::expect_near_rel;
+
+  double after_stirling
+      = std::nextafter(stan::math::lgamma_stirling_diff_useful,
```
mcol: I would put a `using stan::math::lgamma_stirling_diff_useful` since it's used a few times and lines get very long.

test/unit/math/prim/fun/lgamma_stirling_diff_test.cpp
```diff
@@ -0,0 +1,103 @@
+#include <gtest/gtest.h>
+#include <test/unit/math/expect_near_rel.hpp>
+#include <stan/math/prim.hpp>
+#include <stan/math/prim/fun/lgamma_stirling_diff.hpp>
+#include <limits>
+#include <vector>
+#include <cmath>
```
mcol: Here we generally order them as stan, test, gtest, then the rest. Also, no need to include a specific prim header if you already have `prim.hpp`.

test/unit/math/rev/fun/lbeta_test.cpp
```diff
@@ -0,0 +1,191 @@
+#include <test/unit/math/expect_near_rel.hpp>
+#include <stan/math/prim.hpp>
+#include <gtest/gtest.h>
+#include <limits>
+#include <vector>
+#include <string>
+#include <cmath>
```
mcol: Same comment on ordering.

stan/math/prim/fun/lgamma_stirling_diff.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Return the difference between log of the gamma function and it's Stirling
+ * approximation.
+ * This is useful to stably compute log of ratios of gamma functions with large
+ * arguments where the Stirling approximation allows for analytic solution
+ * and the (small) differences can be added afterwards.
+ * This is for example used in the implementation of lbeta.
+ *
+ * The function will return correct value for all arguments, but the can add
```
mcol: Typo: "it can".

test/unit/math/rev/fun/lbeta_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+  double y;
+  double val;
+  double dx;
+  double dy;
+};
+
+constexpr double NaN = std::numeric_limits<double>::quiet_NaN();
+// Test values generated in Mathematice, reproducible notebook at
+// https://www.wolframcloud.com/obj/martin.modrak/Published/lbeta.nb
+// Mathematica Code reproduced below for convenience:
```
mcol: I don't have a strong opinion, but perhaps this should be recorded in the issue rather than committed here.

stan/math/prim/fun/lgamma_stirling_diff.hpp
```diff
@@ -0,0 +1,79 @@
...
+  } else if (value_of(x) < lgamma_stirling_diff_useful) {
+    return lgamma(x) - lgamma_stirling(x);
+  } else {
+    // Using the Stirling series as expressed in formula 5.11.1. at
+    // https://dlmf.nist.gov/5.11
+    constexpr double stirling_series[] = {
+        0.0833333333333333333333333,
+        -0.00277777777777777777777778,
+        0.000793650793650793650793651,
+        -0.000595238095238095238095238,
```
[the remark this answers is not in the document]
martinmodrak: Yeah, from theory I thought the fourth term would be needed, but I couldn't find a test that would require it, so I didn't use it. Added a comment in this spirit.

test/unit/math/rev/fun/lbeta_test.cpp
martinmodrak: I didn't realize putting it into an issue is an option, that's a good point! Still I slightly prefer having it in code - it can evolve with the code and so it might make sense to version it along the code. And the code might be around longer than GitHub...

martinmodrak:
Thanks @mcol for the suggestions, I've implemented all of them and learned a bit more about Stan's style.

bob-carpenter:
> On Jan 14, 2020, at 4:51 AM, Martin Modrák <notifications@github.com> wrote:
> 
> In #1579 @bob-carpenter inquired about Mathematica code. I will address it here, as this PR needs to be merged before #1579.
> 
> Also, have you checked that the IP is OK for us to include Mathematica code in comments and use the output of their approximations? I can't see why it wouldn't be, but it'd be good to check before we go too far down the road of using a proprietary software package.
> 
> I am not a lawyer, but the code is IMHO easy: I've written the code, I (or my employer) have the IP -

So far so good.  You own the IP to what you write.

> this has nothing to do with what you do with the code after it's written.

It does, as this is where the legal restrictions come in.

License to distribute does not follow from owning copyright any more than owning a patent on a technology gives you the right to create it. Patents and copyright are about stopping other people from doing things with your work.  For example, you can write code that's considered a derivative work of other code and then your distribution and licensing is bound by the original code's licensing terms.  Or you can write code  that's covered by a patent you don't own the rights to, so it would be illegal to distribute.

Anyway, I think we should be OK with including code snippets given that Mathematica's a programming language.


> For ouptut of software, the case is less clear cut, but AFAIK the important part is who did a "creative act" in creating the outputs. Referring to for example this source (but others agree): If the user provides non-trivial inputs to a software, they can usually claim copyright to its outputs (the same way Microsoft cannot claim copyright over stuff written in Word / Excel or Adobe cannot claim copyright to images created in Illustrator).

That makes sense.  Thanks for looking into this.

test/unit/math/prim/fun/lgamma_stirling_diff_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+  double value_at = lgamma_stirling_diff(lgamma_stirling_diff_useful);
+  double value_after = lgamma_stirling_diff(after_useful);
+  double diff_before = value_at - value_before;
+  double diff_after = value_after - value_at;
+  double tol
+      = std::max(1e-14 * (0.5 * (fabs(diff_before) + fabs(diff_after))), 1e-14);
+  EXPECT_NEAR(diff_before, diff_after, tol) << "diff around useful cutoff";
+}
+
+namespace lgamma_stirling_diff_test_internal {
```
mcol: There's no need to namespace this in a test file.

stan/math/prim/fun/lgamma_stirling.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/fun/constants.hpp>
+#include <stan/math/prim/fun/lgamma.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return the Stirling approximation to the gamma function.
```
mcol: Shouldn't it say `lgamma`?

test/unit/math/prim/fun/lgamma_stirling_diff_test.cpp
martinmodrak: I've heard there is a plan to run all unit tests in  a single translation unit to save on compilation time (https://discourse.mc-stan.org/t/requiring-global-unique-names-for-unit-tests-in-stan-math/11270). Since I do have `TestValue` and `testValues` in other tests, this seemed like a reasonable precaution.

martinmodrak:
Thanks @mcol for the suggestions, I've implemented them, except for the test namespace (see the comment).

I would tag @bbbales2 for review (but this does not need to make it into the upcoming release, so no rush).

martinmodrak:
Pinging @bbbales2 - do you have time for review or should I ask someone else? Thx!

bbbales2:
Also, I asked Bob about:

> While I reimplemented the code from scratch, some portions of lbeta.hpp are quite similar to the R code as there are only so many ways you can rearrange the Stirling approximation to get stable results.

He said algorithms can't be copyrighted, so we're good.

bob-carpenter:
To clarify, algorithms themselves are like recipes for food---they can't be copyrighted.  

The confusing thing is that the expression of the algorithm in code, like the expression of a recipe in text, can be copyrighted.  

The grey area is whether a reimplementation in another language infringes on the copyright.  The feedback we got from IP lawyers working for NumFOCUS is that we should be OK if we rewrite the algorithm in another language.

Patents are another story.  Algorithms can be patented, but it's much harder to do now than it was 10 or 20 years ago.  The person holding the patent would be able to sue someone using the patent without their permission.

PR #1627, opened 2020-01-19
Eliminate unnecessary loop.

### LINKED ISSUE #1490: Some loops may need to be interchanged; column major

peterwicksstringfield:
## Summary

Eliminate this last loop and close #1490. This is trivial when C and C0 are the same type, sometimes C0 is double and C is an autodiff type, and the automatic conversion is correct. Likewise m and m0.

## Tests

This function is tested in:
test/unit/math/prim/prob/gaussian_dlm_obs_test.cpp

These tests don't exercise the condition in which C and C0 are different types.

## Side Effects

None.

## Checklist

- [X] Math issue #1490

### DIFF
--- a/stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
+++ b/stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
@@ -112,15 +112,8 @@ gaussian_dlm_obs_lpdf(
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m(n);
     Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C(n, n);
 
-    // TODO(anyone): how to recast matrices
-    for (int i = 0; i < m0.size(); i++) {
-      m(i) = m0(i);
-    }
-    for (int i = 0; i < C0.rows(); i++) {
-      for (int j = 0; j < C0.cols(); j++) {
-        C(i, j) = C0(i, j);
-      }
-    }
+    m = m0;
+    C = C0;
 
     Eigen::Matrix<return_type_t<T_y>, Eigen::Dynamic, 1> yi(r);
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> a(n);
@@ -272,15 +265,8 @@ gaussian_dlm_obs_lpdf(
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m(n);
     Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C(n, n);
 
-    // TODO(anyone): how to recast matrices
-    for (int i = 0; i < m0.size(); i++) {
-      m(i) = m0(i);
-    }
-    for (int i = 0; i < C0.rows(); i++) {
-      for (int j = 0; j < C0.cols(); j++) {
-        C(i, j) = C0(i, j);
-      }
-    }
+    m = m0;
+    C = C0;
 
     for (int i = 0; i < y.cols(); i++) {
       // Predict state
--- a/stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
+++ b/stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
@@ -109,12 +109,8 @@ gaussian_dlm_obs_lpdf(
   }
 
   if (include_summand<propto, T_y, T_F, T_G, T_V, T_W, T_m0, T_C0>::value) {
-    Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m(n);
-    Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C(n, n);
-
-    m = m0;
-    C = C0;
-
+    Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m{m0};
+    Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C{C0};
     Eigen::Matrix<return_type_t<T_y>, Eigen::Dynamic, 1> yi(r);
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> a(n);
     Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> R(n, n);
@@ -262,11 +258,8 @@ gaussian_dlm_obs_lpdf(
     T_lp e;
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> A(n);
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> Fj(n);
-    Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m(n);
-    Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C(n, n);
-
-    m = m0;
-    C = C0;
+    Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m{m0};
+    Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C{C0};
 
     for (int i = 0; i < y.cols(); i++) {
       // Predict state


stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
```diff
@@ -112,15 +112,8 @@ gaussian_dlm_obs_lpdf(
...
-    // TODO(anyone): how to recast matrices
-    for (int i = 0; i < m0.size(); i++) {
-      m(i) = m0(i);
-    }
-    for (int i = 0; i < C0.rows(); i++) {
-      for (int j = 0; j < C0.cols(); j++) {
-        C(i, j) = C0(i, j);
-      }
-    }
+    m = m0;
```
bob-carpenter: Line 112 should have all the logic in it:

```
Eigen::Matrix<T_lp, -1, 1> m{m0};
```

This should work if they obeyed the rule of three.  If not, use a compound declare/assign:

```
Eigen::Matrix<T_lp, -1, 1> m = m0;
```

Also, the same should be done for `C`.

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
```diff
@@ -112,15 +112,8 @@ gaussian_dlm_obs_lpdf(
     Eigen::Matrix<T_lp, Eigen::Dynamic, 1> m(n);
     Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> C(n, n);
 
-    // TODO(anyone): how to recast matrices
```
bob-carpenter: Is it guaranteed at this point `m0` doesn't have scalar `T_lp`?  If it does, the copy's unnecessary.

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
```diff
@@ -272,15 +265,8 @@ gaussian_dlm_obs_lpdf(
...
-    for (int i = 0; i < m0.size(); i++) {
-      m(i) = m0(i);
-    }
-    for (int i = 0; i < C0.rows(); i++) {
-      for (int j = 0; j < C0.cols(); j++) {
-        C(i, j) = C0(i, j);
-      }
-    }
+    m = m0;
```
bob-carpenter: Same comment for these assignments---move to compound declare/assign.

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
peterwicksstringfield: m0 can have scalar type T_lp. And m is modified later so we do need a copy anyway. If I'm understanding correctly.

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
peterwicksstringfield: Yeah definitely want to combine those lines, thank you. And direct-list-initialization does work here. 

In this case, direct-initialization, m(m0), would have the same effect; since Eigen::Matrix isn't aggregate, and there are no std::initializer_list constructors or narrowing conversions in sight. Do we prefer direct-list-initialization over direct-initialization when both are possible? Because lines 114-121 (among others) could be using direct-list-initialization instead as well. It looks a little odd to use braces just for C and m.

```
    Eigen::Matrix<return_type_t<T_y>, Eigen::Dynamic, 1> yi{r};
    Eigen::Matrix<T_lp, Eigen::Dynamic, 1> a{n};
    Eigen::Matrix<T_lp, Eigen::Dynamic, Eigen::Dynamic> R{n, n};
    ...
```

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
bob-carpenter: OK.

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
bob-carpenter: So far we haven't developed a standard.  I thought the standard practice in C++ was going to be curly braces for constructors, but I have no idea why I thought that.

bob-carpenter:
Thanks.  This is good to merge when tests pass.

stan/math/prim/prob/gaussian_dlm_obs_lpdf.hpp
peterwicksstringfield: O I'm not disagreeing! I don't know what to do with the whole braces thing and I was hoping their was a policy. Constructors in C++ are kind of weird, they pretend to be functions but then do a bunch of special stuff, so it makes a certain amount of sense for them to just get their own syntax.

For what it's worth, the "cpp core guidelines" have this to say:
"Prefer {}. The rules for {} initialization are simpler, more general, less ambiguous, and safer than for other forms of initialization."

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list

So I think you're right; braces for constructors. And just fix up old code whenever we touch it.

peterwicksstringfield:
As always, thanks for the code review. :)

And for the book recommendation from last time. "The Pragmatic Programmer" was time well spent.

wds15:
btw, how come we keep on merging to develop if the release freeze is in place? Maybe I misunderstood.

SteveBronder:
pretty sure this was considered a bug-fix

bob-carpenter:
Yikes.  That would explain the pileup of PRs in the math lib (over 25 last I checked).   

Sorry for merging other things.  I had mistakenly thought the release had already happened.  When will it be and how will I know when to start merging again?

rok-cesnovar:
I think everything is ready now, we were waiting for some cmdstan-stanc3 docs. Though we could have probably just released Math, since it was ready a few days ago. 

I dont think its worth to revert this, since its a pretty simple change and can technically be seen as a bugfix. It is a precedent and we do need to think about the process for the next release. This one wasnt ideal, though if it werent for the stanc parser switch we would have been on schedule probably.

PR #1628, opened 2020-01-20
Generalize operator-like functions

t4c1:
## Summary

Generalizes function signatures of functions that act kind of like operators (`add`, `subtract`, `multiply`, `divide`, `elt_multiply`, `elt_divide`). That is they can now accept and return general Eigen expressions.
EDIT: I also had to generalize functions that were called directely by return value of one of generalized functions. Those were: `trace`, `trace_gen_quad_form`, `transpose` and `matrix_exp`. 

In many cases I replaced function specializations with overloads. This results in more `require`s, but it also allowed me to replace multiple specializations with one general overload, reducing total amount of code.

There are also some changes to autodiff testing framework that allow it to work with Eigen expressions. I did some of those in earlier PR #1471, but I missed some places.

## Tests

This is mostly a refactor so no new tests.

## Side Effects

None.

## Checklist

- [x] Math issue #1470 


stan/math/prim/fun/dot_product.hpp
```diff
@@ -17,11 +18,9 @@ namespace math {
...
-template <int R1, int C1, int R2, int C2>
-inline double dot_product(const Eigen::Matrix<double, R1, C1> &v1,
-                          const Eigen::Matrix<double, R2, C2> &v2) {
-  check_vector("dot_product", "v1", v1);
-  check_vector("dot_product", "v2", v2);
+template <typename T1, typename T2,
+          typename = require_all_eigen_vector_t<T1, T2>>
+inline auto dot_product(const T1 &v1, const T2 &v2) {
   check_matching_sizes("dot_product", "v1", v1, "v2", v2);
   return v1.dot(v2);
```
t4c1: These `check_vector` checks were redundant, as Eigen's `.dot()` is only defined for compile time vectors, so there is no way this could compile and check fail.

stan/math/prim/fun/multiply.hpp
```diff
@@ -93,22 +117,22 @@ inline Eigen::Matrix<return_type_t<T1, T2>, R1, C2> multiply(
...
-                                      const Eigen::Matrix<T2, R2, 1>& v) {
+template <
+    typename T1, typename T2,
+    typename = require_all_not_var_t<scalar_type_t<T1>, scalar_type_t<T2>>,
+    typename
+    = require_t<conjunction<is_eigen_row_vector<T1>, is_eigen_col_vector<T2>>>>
+inline auto multiply(const T1& rv, const T2& v) {
   check_matching_sizes("multiply", "rv", rv, "v", v);
-  return rv.dot(v);
+  return dot_product(rv, v);
```
t4c1: This change allows using same overload for prim and fwd.

stan/math/prim/fun/trace_gen_quad_form.hpp
```diff
@@ -68,10 +60,11 @@ inline return_type_t<TD, TA, TB> trace_gen_quad_form(
  * @throw std::domain_error if A cannot be multiplied by B or B cannot
  * be multiplied by D.
  */
-template <int RD, int CD, int RA, int CA, typename TB, int RB, int CB>
```
t4c1: `typename TB` is never used and therefore can not be deduced. Due to the bug this overload was never chosen.

test/unit/math/mix/fun/trace_gen_quad_form_test.cpp
```diff
@@ -19,8 +19,8 @@ TEST(mathMixMatFun, traceGenQuadForm) {
   stan::test::expect_ad(f, c11, a11, b11);
 
   stan::test::ad_tolerances tols;
-  tols.hessian_hessian_ = 1e-1;
-  tols.hessian_fvar_hessian_ = 1e-1;
+  tols.hessian_hessian_ = 0.15;
+  tols.hessian_fvar_hessian_ = 0.15;
```
t4c1: I fixed a bug that prevented all-double overload of `trace_gen_quad_form()` to be ever chosen. While faster, that overload seems to be a bit less accurate, so I had to increase these tolerances.

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,58 @@
...
-  return multiply(m, c);
-}
-
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> multiply(
-    const Eigen::Matrix<fvar<T>, R1, C1>& m1,
-    const Eigen::Matrix<fvar<T>, R2, C2>& m2) {
+template <typename T1, typename T2,
+          typename = require_all_eigen_vt<is_fvar, T1, T2>,
+          typename = require_same_t<typename value_type_t<T1>::Scalar,
```
SteveBronder: There is require_same_st and require_same_vt

https://github.com/stan-dev/math/blob/develop/stan/math/prim/meta/require_generics.hpp#L283

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,58 @@
...
+                                    typename value_type_t<T2>::Scalar>,
+          typename = require_not_t<
+              conjunction<is_eigen_row_vector<T1>, is_eigen_col_vector<T2>>>,
```
SteveBronder: [optional]
We could have something like `require_eigen_dot_product (I'm not sure if there is some other math word for general ops with a row and column vector.

Also I think this could use `require_any_not_t`

SteveBronder: Also names like RowVec and ColVec may be good here

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,58 @@
...
+  return m1 * m2;
 }
 
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> multiply(
-    const Eigen::Matrix<fvar<T>, R1, C1>& m1,
-    const Eigen::Matrix<double, R2, C2>& m2) {
+template <typename T1, typename T2, typename = require_all_eigen_t<T1, T2>,
+          typename = require_fvar_t<value_type_t<T1>>,
+          typename = require_same_vt<double, T2>,
```
SteveBronder: Could shorten this up with
```
template <typename Mat1, typename Mat2, typename = require_eigen_vt<is_fvar, Mat1>,
  typename = require_eigen_vt<std::is_floating_point, Mat2>
```

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,58 @@
...
+          typename = require_not_t<
+              conjunction<is_eigen_row_vector<T1>, is_eigen_col_vector<T2>>>,
```
SteveBronder: Maybe just a meta function for `is_dot_product` for these instead of a requires specialization

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,58 @@
...
+          int = 0>
+inline auto multiply(const T1& m1, const T2& m2) {
   check_multiplicable("multiply", "m1", m1, "m2", m2);
-  Eigen::Matrix<fvar<T>, R1, C2> result(m1.rows(), m2.cols());
+  Eigen::Matrix<value_type_t<T1>, T1::RowsAtCompileTime, T2::ColsAtCompileTime>
+      result(m1.rows(), m2.cols());
   for (size_type i = 0; i < m1.rows(); i++) {
-    Eigen::Matrix<fvar<T>, 1, C1> crow = m1.row(i);
+    Eigen::Matrix<value_type_t<T1>, 1, T1::ColsAtCompileTime> crow = m1.row(i);
```
SteveBronder: Should we define a helper type traits that wraps around eigen's internal traits so we can do like `eigen_rows_i<T1>`

stan/math/prim/fun/multiply.hpp
```diff
@@ -15,71 +16,90 @@ namespace math {
...
-          typename = require_all_arithmetic_t<T1, T2>>
-inline Eigen::Matrix<return_type_t<T1, T2>, R1, C2> multiply(
-    const Eigen::Matrix<T1, R1, C1>& m1, const Eigen::Matrix<T2, R2, C2>& m2) {
+template <typename T1, typename T2, typename = require_all_eigen_t<T1, T2>,
+          typename
+          = require_all_arithmetic_t<value_type_t<T1>, value_type_t<T2>>,
+          typename
+          = require_any_not_same_t<double, value_type_t<T1>, value_type_t<T2>>,
+          typename = require_not_t<
+              conjunction<is_eigen_row_vector<T1>, is_eigen_col_vector<T2>>>>
```
SteveBronder: I'm not totally understanding the concept on this one

```
template <typename Mat1, typename Mat2, typename = require_all_eigen_vt<std::is_arithmetic, Mat1, Mat2>
```

Then specialize for one being an integral? (since the above previously said this is for a subset of arithmetic)

Let's add that `is_dot_product` to clean these up

stan/math/prim/fun/multiply.hpp
```diff
@@ -15,71 +16,90 @@ namespace math {
...
+ *
+ * @param m1 first matrix or expression
+ * @param m2 second matrix or expression
+ * @return the product of the first and second matrices
+ * @throw <code>std::invalid_argument</code> if the number of columns of m1 does
+ * not match the number of rows of m2.
+ */
+template <typename T1, typename T2, typename = require_all_eigen_t<T1, T2>,
+          typename
+          = require_all_same_t<double, value_type_t<T1>, value_type_t<T2>>,
```
SteveBronder:
```
template <typename Mat1, typename Mat2, typename = require_all_eigen_vt<std::is_arithmetic, Mat1, Mat2>
```

stan/math/prim/fun/multiply.hpp
SteveBronder: I wonder if the dot product specialization for multiply was because previously we were only allowing Eigen dynamic matrices in some places and internally eigen would just call dot product if it saw a row and column vector

stan/math/prim/fun/trace.hpp
```diff
@@ -16,12 +17,12 @@ namespace math {
  * @param[in] m Specified matrix.
  * @return Trace of the matrix.
  */
-template <typename T>
-inline T trace(const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& m) {
+template <typename T, typename = require_eigen_t<T>>
+inline value_type_t<T> trace(const T& m) {
```
SteveBronder: Does auto deduction not work here?

stan/math/prim/fun/trace_gen_quad_form.hpp
```diff
@@ -34,11 +28,13 @@ namespace math {
...
-          int RB, int CB, typename = require_all_not_var_t<TD, TA, TB>>
-inline return_type_t<TD, TA, TB> trace_gen_quad_form(
-    const Eigen::Matrix<TD, RD, CD> &D, const Eigen::Matrix<TA, RA, CA> &A,
-    const Eigen::Matrix<TB, RB, CB> &B) {
+template <typename TD, typename TA, typename TB,
+          typename = require_all_not_var_t<value_type_t<TD>, value_type_t<TA>,
+                                           value_type_t<TB>>,
+          typename = require_any_not_same_t<double, value_type_t<TD>,
+                                            value_type_t<TA>, value_type_t<TB>>,
+          typename = require_all_eigen_t<TD, TA, TB>>
```
SteveBronder: Can we flip these? So users would read this is for 

- Eigen expressions
- Where not all value types vars
- and at least one is not a double

stan/math/prim/fun/trace_gen_quad_form.hpp
```diff
@@ -34,11 +28,13 @@ namespace math {
...
-template <typename TD, int RD, int CD, typename TA, int RA, int CA, typename TB,
-          int RB, int CB, typename = require_all_not_var_t<TD, TA, TB>>
-inline return_type_t<TD, TA, TB> trace_gen_quad_form(
-    const Eigen::Matrix<TD, RD, CD> &D, const Eigen::Matrix<TA, RA, CA> &A,
-    const Eigen::Matrix<TB, RB, CB> &B) {
+template <typename TD, typename TA, typename TB,
+          typename = require_all_not_var_t<value_type_t<TD>, value_type_t<TA>,
+                                           value_type_t<TB>>,
+          typename = require_any_not_same_t<double, value_type_t<TD>,
+                                            value_type_t<TA>, value_type_t<TB>>,
```
SteveBronder: It might be a weird way to express it but one way to look at this is 

```cpp
require_not_same_t<double, return_type_t<TD, TA, TB>>
```

stan/math/prim/fun/trace_gen_quad_form.hpp
```diff
@@ -68,10 +60,11 @@ inline return_type_t<TD, TA, TB> trace_gen_quad_form(
...
  * be multiplied by D.
  */
-template <int RD, int CD, int RA, int CA, typename TB, int RB, int CB>
-inline double trace_gen_quad_form(const Eigen::Matrix<double, RD, CD> &D,
-                                  const Eigen::Matrix<double, RA, CA> &A,
-                                  const Eigen::Matrix<double, RB, CB> &B) {
+template <typename TD, typename TA, typename TB,
+          typename = require_all_same_t<double, value_type_t<TD>,
+                                        value_type_t<TA>, value_type_t<TB>>,
+          typename = require_all_eigen_t<TD, TA, TB>>
```
SteveBronder: `require_all_eigen_vt<std::is_floating_point, TD, TA, TB>`

If we ever worry about mixing types of floats we could also have
```
require_same_same_vt<TD, TA, TB>
```

SteveBronder: Though the bottom may be better as a static assert to the user

stan/math/prim/fun/transpose.hpp
```diff
@@ -1,13 +1,21 @@
...
-Eigen::Matrix<T, C, R> inline transpose(const Eigen::Matrix<T, R, C>& m) {
+/**
+ * Transposes a matrix.
+ * @tparam T type of the matrix or expression
+ * @param m matrix or expression
+ * @return transposed matrix
+ */
+template <typename T, typename = require_eigen_t<T>>
+Eigen::Matrix<value_type_t<T>, T::ColsAtCompileTime,
+              T::RowsAtCompileTime> inline transpose(const T& m) {
```
SteveBronder: Transpose is a rather cheap op no? Would `auto` not work here because of something with Eigen's `Transpose` class being returned?

stan/math/rev/fun/divide.hpp
```diff
@@ -115,18 +115,18 @@ class matrix_scalar_divide_vv_vari : public vari {
...
-                                       const var& c) {
-  internal::matrix_scalar_divide_dv_vari<R, C>* baseVari
-      = new internal::matrix_scalar_divide_dv_vari<R, C>(m, c);
-  Eigen::Matrix<var, R, C> result(m.rows(), m.cols());
+template <typename T, typename = require_eigen_vt<std::is_arithmetic, T>>
+inline auto divide(const T& m, const var& c) {
+  auto* baseVari
+      = new internal::matrix_scalar_divide_dv_vari<T::RowsAtCompileTime,
+                                                   T::ColsAtCompileTime>(m, c);
+  Eigen::Matrix<var, T::RowsAtCompileTime, T::ColsAtCompileTime> result(
```
SteveBronder: Reading these more I think a `compile_rows_i<T>` would look nice here

stan/math/rev/fun/divide.hpp
```diff
@@ -135,18 +135,19 @@ inline Eigen::Matrix<var, R, C> divide(const Eigen::Matrix<double, R, C>& m,
...
  * @return matrix divided by the scalar
  */
-template <int R, int C>
-inline Eigen::Matrix<var, R, C> divide(const Eigen::Matrix<var, R, C>& m,
-                                       const double& c) {
-  internal::matrix_scalar_divide_vd_vari<R, C>* baseVari
-      = new internal::matrix_scalar_divide_vd_vari<R, C>(m, c);
-  Eigen::Matrix<var, R, C> result(m.rows(), m.cols());
+template <typename T, typename = require_eigen_vt<is_var, T>>
+inline Eigen::Matrix<var, T::RowsAtCompileTime, T::ColsAtCompileTime> divide(
```
SteveBronder: Does auto not work here?

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,64 @@
...
-    const Eigen::Matrix<fvar<T>, R2, C2>& m2) {
+template <
+    typename T1, typename T2, typename = require_all_eigen_vt<is_fvar, T1, T2>,
+    typename = require_same_t<typename value_type_t<T1>::Scalar,
+                              typename value_type_t<T2>::Scalar>,
+    typename
+    = require_t<bool_constant<T1::ColsAtCompileTime == T2::RowsAtCompileTime>>,
+    typename = require_not_t<
+        conjunction<is_eigen_row_vector<T1>, is_eigen_col_vector<T2>>>,
+    int = 0>
```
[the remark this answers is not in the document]
peterwicksstringfield: ! O that's how you do that. I was fighting with that compile error like a week ago. Thank you. :)

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,64 @@
...
-  Eigen::Matrix<fvar<T>, R1, C2> result(m1.rows(), m2.cols());
-  for (size_type i = 0; i < m1.rows(); i++) {
-    Eigen::Matrix<fvar<T>, 1, C1> crow = m1.row(i);
-    for (size_type j = 0; j < m2.cols(); j++) {
-      Eigen::Matrix<fvar<T>, R2, 1> ccol = m2.col(j);
-      result(i, j) = dot_product(crow, ccol);
-    }
-  }
-  return result;
+  return m1 * m2;
```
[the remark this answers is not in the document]
SteveBronder: Is a `.cast<T>` when the matrix already has scalar type T a no-op? If so maybe one thing could do here is `m1.cast<return_type_t<T1, T2>>() * m2.cast<return_type_t<T1, T2>>();` or something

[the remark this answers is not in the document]
andrjohns: Yep, that's a no-op. Got an example on godbolt here where it compiles down to the same assembly [with a cast](https://godbolt.org/z/flQN9u) and [without a cast](https://godbolt.org/z/CnbEev), for a simple assign, at least

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
t4c1: To me this solution seems a bit hacky, but I don't know any better one.

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
t4c1: Unfortunately that being a no-op is not enough to guarantee speedup here. The problem is that matrix multiplication can not be combined with (most) other operations, so if cast is not a no-op it results in a copy of the matrix. 

I benchmarked this. For two square matrices it is comparable to existing code. If the matrix that is being typecast is significantly larger it gets slower.

stan/math/prim/fun/multiply.hpp
t4c1: I am pretty sure the reason was to return a scalar instead of 1x1 matrix.

stan/math/fwd/fun/multiply.hpp
t4c1: I am not sure. This does not simplify code in any way - just hides what is done in a metaprogram. If it was not for consistency with other code, I would also prefere to use `T::Scalar`, wherever `T` could only be Eigen type.

stan/math/prim/fun/trace.hpp
t4c1: It would work. But if the type is not complex or long I prefere not to use auto.

stan/math/prim/fun/transpose.hpp
t4c1: It would work. But than I would have to also generalize any function that is called directely by passing result of transpose. Instead I will do that in next PR.

stan/math/fwd/fun/multiply.hpp
t4c1: The problem here is that `scalar_t<T1>` is `fvar<X>` and I want to compare those `X`es.

t4c1: Whoops, I was wrong. Of course `require_same_vt` is fine.

stan/math/fwd/fun/multiply.hpp
t4c1: Here we have require_NOT, so these are not row and col vec.

stan/math/fwd/fun/multiply.hpp
t4c1: done

stan/math/prim/fun/multiply.hpp
t4c1: We can't use specialization, since return type is different (this one returns expression, all doubles returns a matrix). So it must be an overload.

stan/math/prim/fun/multiply.hpp
t4c1: We have OpenCL implementation only for `double`s, so `is_arithmetic` is not enough.

stan/math/prim/fun/trace_gen_quad_form.hpp
t4c1: Static assert is not what we want here - if that require is not satisfied, other overload must be chosen.

stan/math/prim/fun/trace_gen_quad_form.hpp
t4c1: While this is shorter, I find it less clear.

stan/math/prim/fun/trace_gen_quad_form.hpp
t4c1: Anyway I would not complicate here, since this overload can be removed once transpose is generalized too.

stan/math/rev/fun/divide.hpp
t4c1: It does. I missed that one.

SteveBronder:
For the errors up in the stan repo I started working on generalizing the `rvalue` and `assign` functions. Should those all take in `Eigen::MatrixBase<>` objects (or just a `EigMat` template with a `require_*`?) and call eval if it's being assigned to or we need to do coefficient level access?

https://github.com/stan-dev/stan/blob/refactor/rvalue-pass/src/stan/model/indexing/rvalue.hpp#L114

If that branch is helpful and you want to make any changes to that branch feel free to push straight to it

SteveBronder:
I think the lockstep here is updating the stan repo first then merged this

t4c1:
I agree. `Eigen::MatrixBase<>` is not as general. I think we should either have a general template and require or `Eigen::EigenBase<>`. Let's just go with template + require. 

Looking over the branch you linked I see many changes, none of which seem like they could fix test failures in this PR. Did you link the right branch?

SteveBronder:
> Looking over the branch you linked I see many changes, none of which seem like they could fix test failures in this PR. Did you link the right branch?

Crap actually sorry this branch doesn't cover that. I thought I had a branch with these but looking on my local I must not have saved it

If we use the generic template with require should we have inside each function that does coeff access like 

```cpp
if (is_eigen_expression<EigMat>::value) {
  the_mat.eval();
}
```

Or is there some better pattern?

t4c1:
What you suggested has multiple issues. The correct pattern would be: 
```
const auto& mat = expr_argument.eval();
```
 where `expr_argument` is function argument and `mat` is used for reminder of the function (if is not necessary). However we can do even better and use `Eigen::Ref` to avoid evaling expression if it is just a sub block operation, which allows element access. So the pattern becomes: 
```
const Eigen::Ref<const typename T::PlainObject>& mat = expr_argument;
```
where T is the type of `expr_argument`.

SteveBronder:
Odd, I can get the 

```
  const auto& vec = v.eval();
```
to work but not 
```
const Eigen::Ref<const typename T::PlainObject>& mat = expr_argument;
```

SteveBronder:
Oh, nvm I got it

SteveBronder:
Aight check out rvalue below

https://github.com/stan-dev/stan/blob/stan/generic-rvalue/src/stan/model/indexing/rvalue.hpp#L74

I think we need a `require_eigen_matrix` or something like that

SteveBronder:
I think we need to do the same thing still for `assign`

t4c1:
I checked it out. If you use `Eigen::Ref` don't call `.eval()` - that evaluates any expression. I see you joined two versions for row and col vector into a single overload. Make sure that you make its return value have correct row and col type. Also please open a PR, so I can give you comments there, instead of littering here.

I agree on `assign`.

t4c1:
Due to the bug #1643 I changed return values back to Eigen matrices. We can only return expressions once all functions accept them. This also makes the PR @SteveBronder is preparing for stan repo not needed for this PR, but we will still need it in the long run.

stan/math/prim/fun/trace_gen_quad_form.hpp
mcol: Could you also remove the doxygen `@tparam TB` line?

t4c1:
This is ready for review.

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,50 @@
...
-}
-
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> multiply(
-    const Eigen::Matrix<fvar<T>, R1, C1>& m1,
-    const Eigen::Matrix<fvar<T>, R2, C2>& m2) {
+template <typename T1, typename T2,
+          typename = require_all_eigen_vt<is_fvar, T1, T2>,
+          typename = require_same_vt<T1, T2>,
+          typename = require_not_eigen_row_and_col_t<T1, T2>, unsigned int = 0>
```
SteveBronder: Also why is unsigned int here?

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
t4c1: I am not sure if this can be reproduced with just built-in types. What I had in mind is:
```
Eigen::MatrixXd a;
Eigen::Matrix<fvar, Eigen::Dynamic, Eigen::Dynamic> b;
Eigen::Matrix<fvar, Eigen::Dynamic, Eigen::Dynamic> c = a.cast<fvar>() * b;
```

stan/math/prim/fun/add.hpp
```diff
@@ -11,56 +12,49 @@ namespace math {
...
  * dimensions.
  */
-template <typename T1, typename T2, int R, int C>
-inline Eigen::Matrix<return_type_t<T1, T2>, R, C> add(
-    const Eigen::Matrix<T1, R, C>& m1, const Eigen::Matrix<T2, R, C>& m2) {
+template <typename T1, typename T2, typename = require_all_eigen_t<T1, T2>>
+inline auto add(const T1& m1, const T2& m2) {
   check_matching_dims("add", "m1", m1, "m2", m2);
-  return m1 + m2;
+  return (m1 + m2).eval();
```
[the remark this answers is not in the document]
t4c1: Since there is no `.array()` in this expression and we assume all input are matrices the output will also be matrix. If someone passes in two arrays I would expect this to return the same type - an array.

stan/math/prim/fun/matrix_exp.hpp
```diff
@@ -12,19 +13,25 @@ namespace math {
...
  * @throw <code>std::invalid_argument</code> if the input matrix
  * is not square.
  */
-template <typename T>
-inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> matrix_exp(
-    const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& A) {
+template <typename T, typename = require_eigen_t<T>>
+inline plain_type_t<T> matrix_exp(const T& A_in) {
+  using std::exp;
+  const auto& A = A_in.eval();
```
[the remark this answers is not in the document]
t4c1: Yes. 
```
const auto& a = b.eval();
```
will evaluate any expression `b`.

```
const Eigen::Ref<const typename T::PlainObject>& a = b;
```
will evaluate if `b`'s elements can not be directley accessed. In other words if `b` is something like `c+d` or `3*c` it will evaluate it. If `b` is something like `c.block(...).transpose()` it will not be evaluated.

```
const auto& a = b;
```
This will never evaluate anything. `a` will just be a reference to `b`, regardless of what `b` is.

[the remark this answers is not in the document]
t4c1: Actually I should use `Ref` here.

stan/math/fwd/fun/multiply.hpp
```diff
@@ -10,120 +11,50 @@
...
+      result(m1.rows(), m2.cols());
   for (size_type i = 0; i < m1.rows(); i++) {
-    Eigen::Matrix<double, 1, C1> crow = m1.row(i);
+    Eigen::Matrix<double, 1, T1::ColsAtCompileTime> crow = m1.row(i);
     for (size_type j = 0; j < m2.cols(); j++) {
-      Eigen::Matrix<fvar<T>, R2, 1> ccol = m2.col(j);
+      auto ccol = m2.col(j);
       result(i, j) = dot_product(crow, ccol);
     }
   }
```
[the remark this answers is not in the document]
t4c1: This is a case where less is not better.

Doing `Eigen::Matrix<double, 1, T1::ColsAtCompileTime> crow = m1.row(i);` in outer loop benefits performance for large matrices. Consecutive elements in same row of the matrix are not in consecutive memory locations, so accessing them in order uses cache poorly. Copying them to a vector makes them consecutive in memory and accessing them faster. 

Since we go over them in every iteration of inner loop it is better to inefficiently access them just once and make remaining accesses efficient.

If you want I can make line 51 na 52 into a single line, but it won't affect performance in any way.

stan/math/fwd/fun/multiply.hpp
t4c1: see https://github.com/stan-dev/math/pull/1628#discussion_r368457618

stan/math/fwd/fun/columns_dot_product.hpp
```diff
@@ -16,8 +16,8 @@ inline Eigen::Matrix<fvar<T>, 1, C1> columns_dot_product(
   check_matching_dims("columns_dot_product", "v1", v1, "v2", v2);
   Eigen::Matrix<fvar<T>, 1, C1> ret(1, v1.cols());
   for (size_type j = 0; j < v1.cols(); ++j) {
-    Eigen::Matrix<fvar<T>, R1, C1> ccol1 = v1.col(j);
-    Eigen::Matrix<fvar<T>, R2, C2> ccol2 = v2.col(j);
+    Eigen::Matrix<fvar<T>, R1, 1> ccol1 = v1.col(j);
+    Eigen::Matrix<fvar<T>, R2, 1> ccol2 = v2.col(j);
     ret(0, j) = dot_product(ccol1, ccol2);
   }
```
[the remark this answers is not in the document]
t4c1: I did not intend to include `columns_dot_product` in this PR. This is just a bugfix, that was required for generalization some other functions. Now that functions don't return expressions anymore it is not needed, but I would leave it in.

stan/math/fwd/fun/rows_dot_product.hpp
```diff
@@ -16,8 +16,8 @@ inline Eigen::Matrix<fvar<T>, R1, 1> rows_dot_product(
   check_matching_dims("rows_dot_product", "v1", v1, "v2", v2);
   Eigen::Matrix<fvar<T>, R1, 1> ret(v1.rows(), 1);
   for (size_type j = 0; j < v1.rows(); ++j) {
-    Eigen::Matrix<fvar<T>, R1, C1> crow1 = v1.row(j);
-    Eigen::Matrix<fvar<T>, R2, C2> crow2 = v2.row(j);
+    Eigen::Matrix<fvar<T>, 1, C1> crow1 = v1.row(j);
+    Eigen::Matrix<fvar<T>, 1, C2> crow2 = v2.row(j);
```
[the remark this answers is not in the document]
t4c1: Same answer too :)

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
SteveBronder: > Doing Eigen::Matrix<double, 1, T1::ColsAtCompileTime> crow = m1.row(i); in outer loop benefits performance for large matrices. Consecutive elements in same row of the matrix are not in consecutive memory locations, so accessing them in order uses cache poorly. Copying them to a vector makes them consecutive in memory and accessing them faster.

> Since we go over them in every iteration of inner loop it is better to inefficiently access them just once and make remaining accesses efficient.

Word that's sensible to me.

> If you want I can make line 51 na 52 into a single line, but it won't affect performance in any way.

idt it will effect performance though I think zipping 51 into 52 would be nice

stan/math/prim/fun/matrix_exp.hpp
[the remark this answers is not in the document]
SteveBronder: Ty for the 'splainer!

Side note wish they had some way to do Ref but on the rhs so auto& and auto&& could be used in these cases

stan/math/prim/fun/add.hpp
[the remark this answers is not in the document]
SteveBronder: We previously returned a matrix so I think it would be good to return a matrix now. When we start unwinding these `evals()` we can let this return an array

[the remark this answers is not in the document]
t4c1: Previously this function did not accept arrays, so there is no previous behavour we would have to be consistent with.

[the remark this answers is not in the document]
SteveBronder: The problem here is that if we miss another function that ends up returning an array, then pump that function into multiply, then pump multiply into a function that only accepts a matrix this is gonna throw when we try to compile. Until the rest of stan math can accept non-matrix eigen expressions being safe feels much better here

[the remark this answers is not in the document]
t4c1: Not compiling or throwing if a user makes a bug seems fine to me. Anyway using `Eigen::Array` is not possible from stan language so potential for using Arrays is internal to Stan Math. I think we should follow the convention that functions return same type as they got in, if that makes sense. Std math functions do this, Eigen does it ...

Also adding `.matrix()` would not work for matrices, as they do not define this member function.  Instead we would have to write separate set of overloads for `Eigen::Array`s, which we don't actually need.

t4c1:
@SteveBronder This is still waiting for a review.

SteveBronder:
Sorry for the delay it's been a v busy week. I should have time tomorrow night to look at this

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
SteveBronder: I'm not a fan of this approach. It's feels like it fights against the language. I think it would be good to post a stackoverflow Q about this

stan/math/fwd/fun/multiply.hpp
```diff
@@ -1,133 +1,68 @@
...
 
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> multiply(
-    const Eigen::Matrix<fvar<T>, R1, C1>& m1,
-    const Eigen::Matrix<double, R2, C2>& m2) {
+template <typename Mat1, typename Mat2,
+          typename = require_eigen_vt<is_fvar, Mat1>,
+          typename = require_eigen_vt<std::is_floating_point, Mat2>,
+          typename = require_not_eigen_row_and_col_t<Mat1, Mat2>, int = 0>
+inline auto multiply(const Mat1& m1, const Mat2& m2) {
```
SteveBronder: I think we need to do these multiplies differently. This and the ones in prim are pretty confusing

stan/math/prim/fun/divide.hpp
```diff
@@ -33,19 +34,17 @@ inline int divide(int x, int y) {
...
  */
-template <int R, int C, typename T1, typename T2,
-          typename = require_all_arithmetic_t<T1, T2>>
-inline Eigen::Matrix<return_type_t<T1, T2>, R, C> divide(
-    const Eigen::Matrix<T1, R, C>& m, T2 c) {
-  return m / c;
+template <typename Mat, typename Scal, typename = require_eigen_t<Mat>,
+          typename = require_stan_scalar_t<Scal>,
+          typename = require_all_not_var_t<scalar_type_t<Mat>, Scal>>
+inline auto divide(const Mat& m, Scal c) {
```
SteveBronder: Is this so that divide works for fvar?

stan/math/prim/fun/multiply.hpp
```diff
@@ -15,77 +16,96 @@ namespace math {
...
  */
-template <int R1, int C1, int R2, int C2, typename T1, typename T2,
-          typename = require_all_arithmetic_t<T1, T2>>
-inline Eigen::Matrix<return_type_t<T1, T2>, R1, C2> multiply(
-    const Eigen::Matrix<T1, R1, C1>& m1, const Eigen::Matrix<T2, R2, C2>& m2) {
+template <typename Mat1, typename Mat2,
+          typename = require_all_eigen_vt<std::is_arithmetic, Mat1, Mat2>,
+          typename = require_any_not_same_t<double, value_type_t<Mat1>,
+                                            value_type_t<Mat2>>,
+          typename = require_not_eigen_row_and_col_t<Mat1, Mat2>>
```
SteveBronder: [sidenote]

We should probably have a `require_*_vt` for checking the value types of any arbitrary container

stan/math/prim/fun/trace_gen_quad_form.hpp
```diff
@@ -68,10 +59,10 @@ inline return_type_t<TD, TA, TB> trace_gen_quad_form(
  * @throw std::domain_error if A cannot be multiplied by B or B cannot
  * be multiplied by D.
  */
-template <int RD, int CD, int RA, int CA, typename TB, int RB, int CB>
-inline double trace_gen_quad_form(const Eigen::Matrix<double, RD, CD> &D,
-                                  const Eigen::Matrix<double, RA, CA> &A,
-                                  const Eigen::Matrix<double, RB, CB> &B) {
+template <typename TD, typename TA, typename TB,
+          typename = require_all_eigen_t<TD, TA, TB>,
+          typename = require_all_same_vt<double, TD, TA, TB>>
```
SteveBronder: Should we not do this for all arithmetics?

stan/math/prim/fun/transpose.hpp
SteveBronder: We can't just use .eval() or .matrix() here?

SteveBronder:
I looked this over and like a lot of it! But the multiply needs cleaned up. The templates there are v complicated, the requires stuff is pretty complicated to read for those. Do you have any ideas on how else to do those? I can look at this again on Thursday and see if I can hack up something else

SteveBronder:
If @syclik or @bob-carpenter sign off on the multiply templates then I'm fine with it but I'm not sure about them

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
t4c1: I found a nice solution here: https://stackoverflow.com/questions/38305222/default-template-argument-when-using-stdenable-if-as-templ-param-why-ok-wit

The trick is to use requires as type of template parameters instead of defaults to typename parameters - `require_t<...>* = nullptr` instead of `typename = require_t<...>`

stan/math/fwd/fun/multiply.hpp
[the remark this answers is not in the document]
t4c1: done

stan/math/fwd/fun/multiply.hpp
t4c1: done

stan/math/prim/fun/divide.hpp
t4c1: This one works for fwar. I just generalized it the same way as other functions.

stan/math/prim/fun/trace_gen_quad_form.hpp
t4c1: You are right. Done.

stan/math/prim/fun/transpose.hpp
t4c1: You mean to use auto? Unfortunately not yet. `m.transpose().eval()` returns a row major matrix. This would not work with other functions untill they are generalized.

t4c1:
@SteveBronder Now everything you mentioned is fixed, including multiply templates, and tests passed so this is ready for next review.

PR #1630, opened 2020-01-20
make read_corr_L deal with size-1 correlation matrices

### LINKED ISSUE #1629: Eigen assertion triggered in read_corr_L
## Description
The current implementation of `read_corr_L` silently assumes that `K`, the size of the correlation matrix, is at least 2. This means that for `K = 1` the following line will try to read from a negative index, which triggers an assertion in Eigen :
https://github.com/stan-dev/math/blob/10cc6ba675743f09832c6749fbb1a92d74888bd2/stan/math/prim/fun/read_corr_L.hpp#L61

This was originally observed in https://github.com/stan-dev/stan/issues/2742, where a potential segmentation fault was also reported.

## Expected Output
As a size 1 correlation matrix has to be the matrix `[1]`, its Cholesky factor `L` is also `[1]`. So, in `read_corr_L` we should just return that if `K = 1`. This will allow running models that declare a correlation matrix of size 1, such as the one in https://github.com/stan-dev/stan/issues/2742.

#### Current Version:
v3.0.0

mcol:
## Summary

This adds an early return in `read_corr_L` so that if the input matrix has size 1, it returns the correct Cholesky factor without triggering assertions for out of bounds indices. Fixes #1629.

## Tests
Added  test for size-1 input.

## Side Effects

None.

## Checklist

- [X] Math issue #1629

### DIFF
--- a/stan/math/prim/fun/read_corr_L.hpp
+++ b/stan/math/prim/fun/read_corr_L.hpp
@@ -36,10 +36,11 @@ template <typename T>
 Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
     const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs,  // on (-1, 1)
     size_t K) {
+  if (K == 0) {
+    return {};
+  }
   if (K == 1) {
-    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> L(1, 1);
-    L(0) = 1;
-    return L;
+    return Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>::Identity(1, 1);
   }
 
   using std::sqrt;
@@ -95,10 +96,11 @@ Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
 template <typename T>
 Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
     const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs, size_t K, T& log_prob) {
+  if (K == 0) {
+    return {};
+  }
   if (K == 1) {
-    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> L(1, 1);
-    L(0) = 1;
-    return L;
+    return Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>::Identity(1, 1);
   }
 
   Eigen::Matrix<T, Eigen::Dynamic, 1> values(CPCs.rows() - 1);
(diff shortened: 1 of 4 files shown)


stan/math/prim/fun/read_corr_L.hpp
```diff
@@ -36,6 +36,12 @@ template <typename T>
 Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
     const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs,  // on (-1, 1)
     size_t K) {
+  if (K == 1) {
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> L(1, 1);
```
bob-carpenter: This should be:

```
return Eigen::Matrix<T, -1, -1>::Identity(1, 1);
```

Or with `Eigen::Dynamic` instead of `-1`.

stan/math/prim/fun/read_corr_L.hpp
```diff
@@ -89,6 +95,12 @@ Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
 template <typename T>
 Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
     const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs, size_t K, T& log_prob) {
+  if (K == 1) {
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> L(1, 1);
```
bob-carpenter: Also use constant one-liner here.

stan/math/prim/fun/read_corr_L.hpp
```diff
@@ -36,6 +36,12 @@ template <typename T>
 Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_corr_L(
     const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs,  // on (-1, 1)
     size_t K) {
```
bob-carpenter: [question]
Given that there's a `K == 0` case in other situations, why not here?

stan/math/prim/fun/read_corr_L.hpp
mcol: Thanks, my code looked a bit awkward but couldn't think of a different way!

stan/math/prim/fun/read_corr_L.hpp
mcol: I didn't put it in `read_corr_L` originally because the `K == 0` was already handled in `read_corr_matrix`. I'm now adding it to `read_corr_L` too, as this function is called also from other places that don't seem to do that check.

bob-carpenter:
Thanks for the quick patches.  Ready to merge when tests pass.

Issue #1631, opened 2020-01-20
speedup ode integration


wds15:
## Description

ODE integration currently wastes a few `chain` calls which we can avoid to make ODE integration faster.

## Example

NA
## Expected Output
NA

#### Current Version:
v3.0.0

PR #1634, opened 2020-01-20
avoid unneccessary chain calls on the state used in ODE functor

wds15:
## Summary

Speedup ODE integration by keeping the var chain stack a bit smaller by not putting the states onto the chain stack, but the nochain stack.

## Tests

NA

## Side Effects

## Checklist

- [X] Math issue #1631 

### DIFF
--- a/stan/math/rev/functor/coupled_ode_system.hpp
+++ b/stan/math/rev/functor/coupled_ode_system.hpp
@@ -119,7 +119,10 @@ struct coupled_ode_system<F, double, var> {
     try {
       start_nested();
 
-      const vector<var> y_vars(z.begin(), z.begin() + N_);
+      vector<var> y_vars;
+      y_vars.reserve(N_);
+      for (std::size_t i = 0; i < N_; ++i)
+        y_vars.emplace_back(var(new vari(z[i], false)));
 
       vector<var> dy_dt_vars = f_(t, y_vars, theta_nochain_, x_, x_int_, msgs_);
 
@@ -276,7 +279,10 @@ struct coupled_ode_system<F, var, double> {
     try {
       start_nested();
 
-      const vector<var> y_vars(z.begin(), z.begin() + N_);
+      vector<var> y_vars;
+      y_vars.reserve(N_);
+      for (std::size_t i = 0; i < N_; ++i)
+        y_vars.emplace_back(var(new vari(z[i], false)));
 
       vector<var> dy_dt_vars = f_(t, y_vars, theta_dbl_, x_, x_int_, msgs_);
 
@@ -454,7 +460,10 @@ struct coupled_ode_system<F, var, var> {
     try {
       start_nested();
 
-      const vector<var> y_vars(z.begin(), z.begin() + N_);
+      vector<var> y_vars;
+      y_vars.reserve(N_);
+      for (std::size_t i = 0; i < N_; ++i)
+        y_vars.emplace_back(var(new vari(z[i], false)));
 
       vector<var> dy_dt_vars = f_(t, y_vars, theta_nochain_, x_, x_int_, msgs_);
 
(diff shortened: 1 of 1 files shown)


stan/math/rev/functor/coupled_ode_system.hpp
```diff
@@ -119,7 +119,10 @@ struct coupled_ode_system<F, double, var> {
     try {
       start_nested();
 
-      const vector<var> y_vars(z.begin(), z.begin() + N_);
+      vector<var> y_vars;
+      y_vars.reserve(N_);
+      for (std::size_t i = 0; i < N_; ++i)
+        y_vars.emplace_back(var(new vari(z[i], false)));
```
bob-carpenter: This should use a range-based for-loop and a simplifed emplace-back:

```
for (double zi : z)
  y_vars.emplace_back(new vari(zi, false)));
```

The reason to simplify `emplace_back` is that it is able to takes the arguments of a consructor for `var` and do the construction in place.

stan/math/rev/functor/coupled_ode_system.hpp
```diff
@@ -276,7 +279,10 @@ struct coupled_ode_system<F, var, double> {
...
```
bob-carpenter: Same here.

stan/math/rev/functor/coupled_ode_system.hpp
```diff
@@ -454,7 +460,10 @@ struct coupled_ode_system<F, var, var> {
...
```
bob-carpenter: Same.  

[optional from here on]

Which is making me think this should all be encapsulated in a construction function:

```
std::vector<var> to_vector_var(const std::vector<double>& x) {
  vector<var> y;
  y.reserve(x.size());
  for (double xi : x)
    y.emplace_back(new vari(xi, false));  
  return y;     
}
```

The  compiler would presumably be able to remove the copy on return.

bbbales2:
This warning applies here I think also: https://github.com/stan-dev/math/blob/develop/stan/math/rev/functor/coupled_ode_system.hpp#L147

bob-carpenter:
Regarding @bbbales2's message above, https://github.com/stan-dev/math/pull/1634#issuecomment-576389272, this actually applies **everywhere** we convert a `vector<double>` to a `vector<var>` and should also apply anywhere we convert a `Matrix<double, R, C>` to a `Matrix<var, R, C>`.

What do people think about pulling out construction functions as I suggested in https://github.com/stan-dev/math/pull/1634#discussion_r368681698?

wds15:
@bbbales2 I think the adjoints are restored to zero:

https://github.com/stan-dev/math/blob/10cc6ba675743f09832c6749fbb1a92d74888bd2/stan/math/rev/core/set_zero_all_adjoints_nested.hpp#L34

The `y`s are constructed inside the nested AD stack and are thus reset to zero like they should.

@bob-carpenter We should definitely have construction utilities to make use of this optimization everywhere. `map_rect` should use this as well, for example (the root solver the same).

stan/math/rev/functor/coupled_ode_system.hpp
wds15: Thanks. Fixed. For this one and for the placement of the theta's on the outer stack as well.

stan/math/rev/functor/coupled_ode_system.hpp
```diff
@@ -119,7 +119,10 @@ struct coupled_ode_system<F, double, var> {
     try {
       start_nested();
 
-      const vector<var> y_vars(z.begin(), z.begin() + N_);
+      vector<var> y_vars;
+      y_vars.reserve(N_);
+      for (std::size_t i = 0; i < N_; ++i)
```
bob-carpenter: I'd strongly prefer this to be done with range iterators, as in:

```
for (double zi : z)
  y_vars.emplace_back(new vari(zi, false));
```

wds15: No, that does not work, since we do not iterate over the entire z vector. Only the first N elements of it.

bob-carpenter:
Thanks for patching.  Ready to merge once tests pass.

PR #1636, opened 2020-01-21
Add row vector, array and int_array construction utilities

### LINKED ISSUE #692: Container construction utilities
#### Summary:

Provide functions to construct commonly used containers.

#### Description:

Here are suggested signatures:

```
// [1/K, ..., 1/K]'
vector uniform_simplex(int K);
```

```
// [1, ..., 1]'
vector ones_vector(int K);
```

```
// [0, ..., 0]'
vector zeros_vector(int K);
```

```
// [c, ..., c]'
vector constant_vector(double c);
```

```
// diag_matrix(ones_vector(K))
matrix identity_matrix(int K);
```

```
// [low, ..., high]' with equal spacing between values in ...
vector set_spaced_vector(real low, real high, int K);
```

```
// [0, ..., 1, ..., 0]'  with 1 in k-th position (indexing from 1)
vector one_hot_vector(int k, int K);
```

These all need to validate sizes are non-negative and validate other bounds like the index being in ragne for `one_hot_vector`.

We could also have versions for `_row_vector` and `_array`, and `_int_array`.  

This lines up with what's [available in Eigen](https://eigen.tuxfamily.org/dox/group__QuickRefPage.html#title3) [we already have `rep_vector` for their constant vectors and specific RNGs].

#### Additional Information:

These only require double-based implementations in C++, because in Stan there are only integer arguments, so there can't be any derivatives.  For example, this function will have to be added to `stan/math/prim/mat/fun/uniform_simplex.hpp` and then included in the header for `prim/mat.hpp`.

```
Eigen::VectorXd uniform_simplex(long K) {
  // validate that K is positive
  // return K-vector filled with 1.0 / K (the 1.0 avoids int division)
```

The implementation of `uniform_simplex` can call `const_vector`.

#### Current Version:
v2.17.0

mcol:
## Summary

This adds the code to implement the following function signatures:
- `row_vector one_hot_row_vector(int K, int k)`
- `row_vector ones_row_vector(int K, int k)`
- `row_vector spaced_row_vector(int K, real low, real high)`
- `row_vector zeros_row_vector(int K)`
- `real[] one_hot_array(int K, int k)`
- `real[] ones_array(int K, int k)`
- `real[] spaced_array(int K, real low, real high)`
- `real[] zeros_array(int K)`
- `int[] one_hot_int_array(int K, int k)`
- `int[] ones_int_array(int K, int k)`
- `int[] zeros_int_array(int K)`

Fixes #692. 

## Tests

Added unit tests for each function.

## Side Effects

None.

## Checklist

- [X] Math issue #692


bob-carpenter:
Before reviewing, I have some comments on scope.

All of the `constant_XXX` functions are duplicates of existing `rep_XXX` functions.  For instance, `constant_int_array` and `const(_real)_array` are currently overloaded as `rep_array(double, int)` and `rep_array(int, int)`.  `const_vector` and `const_row_vector` are `rep_vector` and `rep_row_vector`.  

The matrix versions are missing from this list.  We currently have rep_matrix and rep_array that works in 2D.  We also have 2D and 3D array constants as overloads of `rep_array`.

`set_XXX` functions should not have `set_` as a prefix as nothing is getting set.

`constant_XXX` functions should not have `constant_` as a prefix because the result isn't constant in any sense.  I mean `constant_XXX(...)` is not an lvalue, but as soon as you assign it, which you have to do to use it, it's not constant any more.  This is why Eigen just uses `Zero` and `One`.  Note that they do not use `Zeros` and `Ones`, but I'm OK with its usage in the functions here, though I'm also OK with `one_array` and `zero_array`.

mcol:
I don't have any particular attachment to these functions or names, so I'm happy to remove those that duplicate existing functionality. I followed the indications in #692, and when reviewing #1620 none of this came up. So whatever you settle on here will potentially affect the other functions (I'm totally fine with it).

bob-carpenter:
Thanks.  And sorry about not catching this the first time.  I've been doing
far too many reviews these days.  I have to slow down and be more careful.

Yes, I think we should remove the constant_ functions because they're duplicates and they're not constant.

I also think set_ should be removed from the names.

Given that none of this has gone into the Stan language yet, I don't think it should be a problem.

mcol:
No problem at all, it's better to get these right while they are not exposed yet.

bob-carpenter:
Thanks.  And sorry again for not being sharper on the first review!

stan/math/prim/fun/one_hot_row_vector.hpp
```diff
@@ -0,0 +1,33 @@
...
+#define STAN_MATH_PRIM_FUN_ONE_HOT_ROW_VECTOR_HPP
+
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return a row vector with 1 in the k-th position and zero elsewhere
```
bob-carpenter: [optional]
Minor detail, but these are full imperative sentences that should end with periods.

stan/math/prim/fun/one_hot_int_array.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * Return an integer array with 1 in the k-th position and zero elsewhere
+ *
+ * @param K size of the array
+ * @param k position of the 1 (indexing from 1)
+ * @return An integer array of size K with all elements initialised to zero
+ * and a 1 in the k-th position.
+ * @throw std::domain_error if K is not positive, or if k is less than 1 or
+ * greater than K.
+ */
+inline std::vector<int> one_hot_int_array(int K, int k) {
```
bob-carpenter: [optional]
Probably not worth refactoring, but a single implementation would work here.  I realize there'd need to be two fronting functions to work easily with the Stan language, but the core implementation is shared.  They'd probably need an enable-if to ensure the type is arithmetic.

That's true of most of these functions.

stan/math/prim/fun/ones_array.hpp
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Return an array of ones
+ *
+ * @param K size of the array
+ * @return An array of size K with all elements initialised to 1.
+ * @throw std::domain_error if K is negative.
+ */
+inline std::vector<double> ones_array(int K) {
+  check_nonnegative("ones_array", "size", K);
+  return std::vector<double>(K, 1);
```
bob-carpenter: This can just be `return {K, 1};`  You can always return curly braces around constructor args for the return type.

And same for next definition.

stan/math/prim/fun/spaced_array.hpp
```diff
@@ -0,0 +1,52 @@
...
+ * if high is nan or infinite, or if high is less than low.
+ */
+inline std::vector<double> spaced_array(int K, double low, double high) {
+  static const char* function = "spaced_array";
+  check_nonnegative(function, "size", K);
+  check_finite(function, "low", low);
+  check_finite(function, "high", high);
+  check_greater_or_equal(function, "high", high, low);
+
+  if (K == 0) {
```
bob-carpenter: [optional]

This could be implemented as

```
Eigen::VectorXd a = Eigen::VectorXd::LinSpaced(K, low, high);
return {&a[0], &a[0] + K};
```

It's an extra copy but a lot less code and ensures we get the same answers in all of our functions.
It should match the input iterator constructor.

stan/math/prim/fun/zeros_array.hpp
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Return an array of zeros
+ *
+ * @param K size of the array
+ * @return an array of size K with all elements initialised to 0.
+ * @throw std::domain_error if K is negative.
+ */
+inline std::vector<double> zeros_array(int K) {
+  check_nonnegative("zeros_array", "size", K);
+  return std::vector<double>(K, 0);
```
bob-carpenter: Should use braces constructor return.

stan/math/prim/fun/zeros_int_array.hpp
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Return an integer array of zeros
+ *
+ * @param K size of the array
+ * @return an integer array of size K with all elements initialised to 0.
+ * @throw std::domain_error if K is negative.
+ */
+inline std::vector<int> zeros_int_array(int K) {
+  check_nonnegative("zeros_int_array", "size", K);
+  return std::vector<int>(K, 0);
```
bob-carpenter: Same here.

test/unit/math/prim/fun/zeros_row_vector_test.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <gtest/gtest.h>
+
+TEST(MathFunctions, zeros_row_vector) {
+  using Eigen::RowVectorXd;
+  using stan::math::zeros_row_vector;
+
+  RowVectorXd u0 = zeros_row_vector(0);
+  EXPECT_EQ(0, u0.size());
+
+  for (int K = 1; K < 5; K++) {
```
bob-carpenter: Start the loop at K = 0.  The general test will hold.

test/unit/math/prim/fun/zeros_row_vector_test.cpp
```diff
@@ -0,0 +1,20 @@
...
+    Eigen::RowVectorXd v = Eigen::RowVectorXd::Zero(K);
```
bob-carpenter: Stick to either the `using` with no qualification or qualification.  As is, there's one `Eigen::RowVectorXd` and one just `RowVetorXd`.

test/unit/math/prim/fun/zeros_int_array_test.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+TEST(MathFunctions, zeros_int_array) {
+  using stan::math::zeros_int_array;
+
+  std::vector<int> u0 = zeros_int_array(0);
+  EXPECT_EQ(0, u0.size());
+
+  for (int K = 1; K < 5; K++) {
+    std::vector<int> u = zeros_int_array(K);
+    std::vector<int> v(K);
```
bob-carpenter: `v` needs ot explicitly initialize to 0.  That's not guaranteed by the spec.

This also doesn't need a separate 0 case---it can get rolled into the loop.  At that point, you can save some vertical space by explicitly qualifing the call to `zeros_int_array`.

test/unit/math/prim/fun/zeros_array_test.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+TEST(MathFunctions, zeros_array) {
+  using stan::math::zeros_array;
+
+  std::vector<double> u0 = zeros_array(0);
+  EXPECT_EQ(0, u0.size());
+
+  for (int K = 1; K < 5; K++) {
+    std::vector<double> u = zeros_array(K);
+    std::vector<double> v(K);
```
bob-carpenter: Same here as below---`v` needs to be initialized to 0 and the 0 case can be rolled into the loop.

test/unit/math/prim/fun/spaced_array_test.cpp
```diff
@@ -0,0 +1,52 @@
+#include <stan/math/prim.hpp>
+#include <gtest/gtest.h>
+#include <limits>
+#include <vector>
+
+TEST(MathFunctions, spaced_array) {
+  using stan::math::spaced_array;
+
+  std::vector<double> u0 = spaced_array(0, 1, 2);
```
bob-carpenter: [optional]

With this kind of function, it can be helpful to write a general test function, as in:

```
void expect_spaced_array(int K, double low, double high, const std::vector<double>& expected) {
  std::vector<double> found = spaced_array(K, low, high);
  expect_float_eq_array(expected, found);
}
```

Where the final test is a general test to compare arrays.  We should add that ASAP to avoid all this redundancy.  it could go in with the matrix comparisons in

```
test/unit/math/prim/fun/expect_matrix_eq.hpp
```

stan/math/prim/fun/ones_array.hpp
mcol: This doesn't work:
```
./stan/math/prim/fun/ones_array.hpp:19:11: error: non-constant-expression cannot be
      narrowed from type 'int' to 'double' in initializer list [-Wc++11-narrowing]
  return {K, 1};
          ^
./stan/math/prim/fun/ones_array.hpp:19:11: note: insert an explicit cast to silence
      this issue
  return {K, 1};
          ^
          static_cast<double>( )
```
If I use the static cast around `K`, it builds a vector of two elements.

stan/math/prim/fun/spaced_array.hpp
mcol: This works, but the `K == 0` case has to be kept separate.

stan/math/prim/fun/zeros_array.hpp
mcol: Same as in the ones case.

test/unit/math/prim/fun/spaced_array_test.cpp
mcol: I've added `expect_std_vector_eq` to the header. With that I could simplify most array tests, and implement your suggesion for the `spaced_array` tests.

stan/math/prim/fun/one_hot_int_array.hpp
mcol: I'll pass on this for now.

mcol:
I've addressed your request, apart from those that I couldn't get to compile. To be honest those don't seem a big deal, as my code was not much longer anyway. The suggestion to merge the array/int array implementations is good, but there are a number of moving pieces that I'm not familiar with just yet, so I may revisit it in the future.

stan/math/prim/fun/ones_array.hpp
bob-carpenter: Thanks for trying.  I've never seen that error and it's suprrising in that I thought you could always replace return-type constructor calls with braces.   It clearly requires a stricter match.

bob-carpenter: It does look like declaring `const int K` might also work, but I'd rather not sprinkle constants around.

Issue #1637, opened 2020-01-22
Intermittent segfaults with map_rect and TBB


Pyrena:
## Description
Running a model with map_rect and multithreading enabled results in segfaults in around 5% of the cases. The issue only occurs on v2.21.0, not on v2.20.0 and could be reproduced by @wds15.
When the segfault occurs, results are still written to disk.

## Example

To reproduce, compile the attached model (taken from the stan user manual) and repeatedly run it with the attached data (though the segfaults seem to occur independently of the model or data). 
Observed output (~5% of the runs):

> Elapsed Time: 0.482551 seconds (Warm-up)
> 0.6037 seconds (Sampling)
> 1.08625 seconds (Total)
> 
> [2] 88862 segmentation fault /home/hbr/sync/test/map_rect_model sample data

Link to discourse thread: https://discourse.mc-stan.org/t/intermittent-segfaults-with-map-rect-and-tbb/12750

g++ -v

> Using built-in specs.
> COLLECT_GCC=g++
> COLLECT_LTO_WRAPPER=/usr/lib64/gcc/x86_64-solus-linux/9/lto-wrapper
> Target: x86_64-solus-linux
> Configured with: …/configure --prefix=/usr --with-pkgversion=Solus --libdir=/usr/lib64 --libexecdir=/usr/lib64 --with-system-zlib --enable-shared --enable-threads=posix --enable-gnu-indirect-function --enable-__cxa_atexit --enable-plugin --enable-gold --enable-ld=default --enable-clocale=gnu --enable-multilib --with-multilib-list=m32,m64 --enable-lto --with-gcc-major-version-only --with-bugurl=https://dev.getsol.us/ --with-arch_32=i686 --enable-linker-build-id --with-linker-hash-style=gnu --with-gnu-ld --build=x86_64-solus-linux --target=x86_64-solus-linux --enable-languages=c,c++,fortran
> Thread model: posix
> gcc version 9.2.0 (Solus)

## Expected Output
No intermittent segfaults

[map_rect_model.stan.txt](https://github.com/stan-dev/math/files/4096720/map_rect_model.stan.txt)
[data.Rdump.txt](https://github.com/stan-dev/math/files/4096721/data.Rdump.txt)

rok-cesnovar:
Does this mean any map_rect model is afected when threading is ok?

wds15:
See here:

https://github.com/stan-dev/cmdstan/issues/802#issuecomment-577207184

[...]

PR #1638, opened 2020-01-23
avoid random segfaults at program exit

wds15:
## Summary

Avoid random segfaults at program exit due to use of already deallocated objects in the global TBB thread pool observer.

## Tests

I was not able to write a reproducible test as this is similar to the static init order problem and has a random nature. The model referred to in the issue #1637 now runs without segfaults for me. 

## Side Effects

No

## Checklist

- [X] Math issue #1637 

### DIFF
--- a/stan/math/rev/core/init_chainablestack.hpp
+++ b/stan/math/rev/core/init_chainablestack.hpp
@@ -33,9 +33,7 @@ class ad_tape_observer : public tbb::task_scheduler_observer {
     observe(true);             // activates the observer
   }
 
-  ~ad_tape_observer() {
-    observe(false);
-  }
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
I think a reproducible test (or at least one that fails 1/10 runs or so) can only be done on the cmdstan repo (or even maybe on the performance cmdstan repo). I can see if I can get something like that after the release.

I was able to reproduce this bug easily with 8 threads on my workstation. Will double check later that this fixes it (definitely before all the tests pass) and then approve. Thanks for the quick fix.

bob-carpenter:
Thanks much for fixing this so quickly.

Issue #1643, opened 2020-01-27
Stan Math functions should not return Eigen expressions


avehtari:
#### Summary:
Using cmdstan-2.22.0.tar.gz, fixed stan-dev/cmdstan#804 so that make build works.
EDIT: the same problem with the latest git commit e9075d0
EDIT: by @wds15 suggestion transferred tgo stanc3 repo

Using cmdstanr (but this is clearly an issue in cmdstan or stanc3)

One model compiles and runs successfully, but other one one fails the compilation.

#### Description:

This Stan model compiles and runs with previous cmdstan and 2.22.0
```
data {
  int<lower=0> D;           // number of dimensions
}
parameters {
  vector[D] mu; 
}
model {
  mu ~ normal(0, 1);
}
```

This model compiles and runs with the previous cmdstan, but doesn't compile with 2.22.0
```

data {
  int<lower=0> n;				      // number of observations
  int<lower=0> d;             // number of predictors
  int<lower=0,upper=1> y[n];	// outputs
  matrix[n,d] x;				      // inputs
  real<lower=0> scale_icept;	// prior std for the intercept
  real<lower=0> scale_global;	// scale for the half-t prior for tau
  real<lower=1> nu_global;	  // degrees of freedom for the half-t priors for tau
  real<lower=1> nu_local;		  // degrees of freedom for the half-t priors for lambdas
                              // (nu_local = 1 corresponds to the horseshoe)
  real<lower=0> slab_scale;   // for the regularized horseshoe
  real<lower=0> slab_df;
}

parameters {
  real beta0;
  vector[d] z;                // for non-centered parameterization
  real <lower=0> tau;         // global shrinkage parameter
  vector <lower=0>[d] lambda; // local shrinkage parameter
  real<lower=0> caux;
}

transformed parameters {
  vector[d] beta;                     // regression coefficients
  {
    vector[d] lambda_tilde;   // 'truncated' local shrinkage parameter
    real c = slab_scale * sqrt(caux); // slab scale
    lambda_tilde = sqrt( c^2 * square(lambda) ./ (c^2 + tau^2*square(lambda)));
    beta = z .* lambda_tilde*tau;
  }
}

model {
  // half-t priors for lambdas and tau, and inverse-gamma for c^2
  z ~ std_normal();
  lambda ~ student_t(nu_local, 0, 1);
  tau ~ student_t(nu_global, 0, scale_global*2);
[...]

PR #1644, opened 2020-01-27
fix returned expressions

t4c1:
## Summary

PR #1471 generalized some functions to return Eigen expressions. In #1643 it was found that the change prevents compilation of some models. This PR fixes the problem by making those functions return `Eigen::Matrix` again.

## Tests

No tests, as this kind of problem is hard to test in general. But now I know this is an issue It will be simple not to repeat it.

## Side Effects

Maybe the code will be a bit slower, but the difference should be less than the speedup introduced in #1471, if it is present at all.

## Checklist

- [x] Math issue #1643

### DIFF
--- a/stan/math/prim/fun/log_softmax.hpp
+++ b/stan/math/prim/fun/log_softmax.hpp
@@ -41,7 +41,7 @@ template <typename T, require_t<std::is_arithmetic<scalar_type_t<T>>>...>
 inline auto log_softmax(const T& x) {
   return apply_vector_unary<T>::apply(x, [&](const auto& v) {
     check_nonzero_size("log_softmax", "v", v);
-    return (v.array() - log_sum_exp(v)).matrix();
+    return (v.array() - log_sum_exp(v)).matrix().eval();
   });
 }
 }  // namespace math
--- a/stan/math/prim/fun/acos.hpp
+++ b/stan/math/prim/fun/acos.hpp
@@ -45,7 +45,7 @@ inline auto acos(const T& x) {
 template <typename Derived,
           typename = require_eigen_vt<std::is_arithmetic, Derived>>
 inline auto acos(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().acos().matrix();
+  return x.derived().array().acos().matrix().eval();
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/asin.hpp
+++ b/stan/math/prim/fun/asin.hpp
@@ -45,7 +45,7 @@ inline auto asin(const T& x) {
 template <typename Derived,
           typename = require_eigen_vt<std::is_arithmetic, Derived>>
 inline auto asin(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().asin().matrix();
+  return x.derived().array().asin().matrix().eval();
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/atan.hpp
+++ b/stan/math/prim/fun/atan.hpp
@@ -45,7 +45,7 @@ inline typename apply_scalar_unary<atan_fun, T>::return_t atan(const T& x) {
 template <typename Derived,
           typename = require_eigen_vt<std::is_arithmetic, Derived>>
 inline auto atan(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().atan().matrix();
+  return x.derived().array().atan().matrix().eval();
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/ceil.hpp
+++ b/stan/math/prim/fun/ceil.hpp
@@ -45,7 +45,7 @@ inline auto ceil(const T& x) {
 template <typename Derived,
           typename = require_eigen_vt<std::is_arithmetic, Derived>>
 inline auto ceil(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().ceil().matrix();
+  return x.derived().array().ceil().matrix().eval();
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/cos.hpp
+++ b/stan/math/prim/fun/cos.hpp
@@ -45,7 +45,7 @@ inline auto cos(const T& x) {
 template <typename Derived,
           typename = require_eigen_vt<std::is_arithmetic, Derived>>
 inline auto cos(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().cos().matrix();
+  return x.derived().array().cos().matrix().eval();
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/cosh.hpp
+++ b/stan/math/prim/fun/cosh.hpp
@@ -45,7 +45,7 @@ inline typename apply_scalar_unary<cosh_fun, T>::return_t cosh(const T& x) {
 template <typename Derived,
           typename = require_eigen_vt<std::is_arithmetic, Derived>>
 inline auto cosh(const Eigen::MatrixBase<Derived>& x) {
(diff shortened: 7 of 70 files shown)


t4c1:
@avehtari Can you check that using this branch fixes the problem?

avehtari:
Seems to work.
`bin/stanc --version` gives `stanc acfb4612 (Unix)` and `cmdstanr` using that compiles and runs the longest example in that issue.

rok-cesnovar:
@serban-nicusor-toptal just tagging you that the merge commit of this PR should be cherry-picked to Stan Math 3.1.1 

This has to first pass tests. Let me know if you need any help regarding the patch release.

bob-carpenter:
The model that failed to compile should be added to the set of good test models.  Those are tested for compilability.

avehtari:
> The code that failed to compile should be added as a test. An alternative is to add the problematic model to the good models in the stan-dev/stan repo.

Additional info: That model is our (Aki, Juho Piironen) recommended reference code for logistic regression with regularized horseshoe prior. This will be also included in posteriordb as with some data sets, we get very nice examples of high dimensional multimodal distributions for which dynamic HMC in Stan works well.

t4c1:
> The code that failed to compile should be added as a test. An alternative is to add the problematic model to the good models in the stan-dev/stan repo.

A test like this will only test a single combination of nested function calls `a(b(...))`. The problem could potentially happen again for any combination of functions `a` and `b` from stan math, where `b` can return and `a` can accept `Eigen::Matrix`. If we want to make tests in a way that will prevent this from happening again all combinations should be tested. And I would prefere not to write these tests manually ...

However not testing that is not that large problem, since avoiding this is easy once one knows it could be an issue. Also once all functions are generalized to accept any Eigen expressions this issue can not happen again.

bob-carpenter:
I understand that it's hard to test in full generality.  Rather than waiting for an automated general solution, I'd like to see the tests added for the cases being fixed to at least ensure the problem is fixed and there are no regressions.

SteveBronder:
Having no testing for this because we can't test everything seems like not a great answer

rok-cesnovar:
> That model is our (Aki, Juho Piironen) recommended reference code for logistic regression with regularized horseshoe prior.

I will make a Stan PR to add this to the good-models. Do I have your permission to add it? If yes, I will add you both as copyright holder for the PR. Do you want to add any description as a commented-out line in the stan model?

avehtari:
You can add the code. 

Useful comment would be
> The regularized horseshoe prior code from Appendix C.1 of Piironen and Vehtari (2017). Sparsity information and regularization in the horseshoe and other shrinkage priors. In Electronic Journal of Statistics, 11(2):5018-5051. https://projecteuclid.org/euclid.ejs/1513306866

as this paper and the appendix explains what that model is.

As the code is very direct implementation of math in that paper, it's likely that someone else writing the code just using the equations would get the same result and thus there is no creativity and it should not be copyrightable. I think that comment should be sufficient, but I and Juho don't mind if you want to have us as the copyright holders in the PR.

rok-cesnovar:
Added Aki's and Juho's model to Stan and it fails on current math develop and passes with this PR as Math submodule. Once this PR is green we can proceed then.

failing: 
https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/PR-2882/3/pipeline
passing:
https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/PR-2882/4/pipeline/
parameters of the passing: https://jenkins.mc-stan.org/job/Stan/job/PR-2882/4/parameters/

bob-carpenter:
> On Jan 28, 2020, at 2:28 PM, Rok Češnovar <notifications@github.com> wrote:
> 
> Added Aki's and Juho's model to Stan and it fails on current math develop and passes with this PR as Math submodule. 

Thanks!

SteveBronder:
Should we just revert #1471 and #1558, do a hotfix release as 1.22.1 and make revert revert PRs for both of those? We need to do a 1.22.1 for cmdstan anyway

I think we went around this the opposite way we should have. We should have the stan repo first accept generic eigen expressions, then do this in math

stan/math/prim/fun/log_softmax.hpp
```diff
@@ -41,7 +41,7 @@ template <typename T, require_t<std::is_arithmetic<scalar_type_t<T>>>...>
 inline auto log_softmax(const T& x) {
   return apply_vector_unary<T>::apply(x, [&](const auto& v) {
     check_nonzero_size("log_softmax", "v", v);
-    return (v.array() - log_sum_exp(v)).matrix();
+    return (v.array() - log_sum_exp(v)).matrix().eval();
```
andrjohns: Rather than adding ```.eval()``` to each of these functions, you could instead modify the ```apply_vector_unary``` header to eval the eigen returns:
```
  template <typename F>
  static inline auto apply(const T& x, const F& f) {
    return f(x).eval();
  }
```

t4c1: I went this way, so we don't need to first eval and than copy matrices in std::vector overloads that assign result of `apply_vector_unary::apply()` instead of returning it.

rok-cesnovar:
> Should we just revert #1471 and #1558, do a hotfix release as 1.22.1 and make revert revert PRs for both of those? We need to do a 1.22.1 for cmdstan anyway

We need to do 3.1.1 with either this PR or revert of #1471. I am fine either way. The main benefit of #1471 will be seen after we have everything using eigen expressions anyways I think.

t4c1:
Actually the main benefit of #1471 is using Eigens's implementations of exp that is way faster than std implementation.

rok-cesnovar:
Reverting #1471 might however be an issue due to the flatten.

SteveBronder:
Why do the test values need changed?

SteveBronder:
I'll try the revert tonight

t4c1:
> Why do the test values need changed?

At first I just copy-pasted same values for all functions. But it turns out for some functions they were out of range where functions are defined.

bob-carpenter:
One possible fix for this kind of thing is define a function that composes a variable transform with the function being tested so that it can take unconstrained input values to test.

rok-cesnovar:
> I'll try the revert tonight

We either add the merge commit of this PR to master or the revert merge commit PR to master and that becomes Math 3.1.1. Not sure its worth another set of Jenkins tests. 

I am also not sure what you mean by 

> We should have the stan repo first accept generic eigen expressions

There is no issue in the Stan repo. Can you explain?

SteveBronder:
>> We should have the stan repo first accept generic eigen expressions

> There is no issue in the Stan repo. Can you explain?

In [#1628](https://github.com/stan-dev/math/pull/1628#issuecomment-577360793) We hit some errors upstream in the Stan repo related to passing arbitrary eigen expressions.

I'm working on the indexing functions right now but we should check what else we would want to make sure takes in Eigen matrices. We'll probably need to do this `.eval()` on all the returns first while we get the signatures lined up. Then have one repo where we remove all the .evals()

SteveBronder:
If all checks are passing and this fixes all the upstream issues then let's approve and merge

rok-cesnovar:
> We'll probably need to do this `.eval()` on all the returns first while we get the signatures lined up. Then have one repo where we remove all the .evals()

Yep, I see no other way of doing this. We also need to come up with a "eigen expression mother model" that would make sure all functions signatures that accept eigen matrices, then work with eigen expressions.

> If all checks are passing and this fixes all the upstream issues then let's approve and merge

It does yes, and I have reviewed changes here and they look fine. I will however go through #1471 and this PR manually to double check no function from the first PR now returns eigen expressions. Then I will merge and that is where @serban-nicusor-toptal comes in :)

PR #1650, opened 2020-01-28
Add reverse

### LINKED ISSUE #344: add reverse() function for arrays and vectors
#### Summary:

`reverse(x)` should return the elements of x in reversed order.
#### Description:

This allows, for example, a decreasing vector type by reversing an increasing one.
#### Current Version:

v2.11.0

mcol:
## Summary

This add the implementation of `reverse` for vectors, row vectors and arrays. Fixes #344.

## Tests

Added some basic prim and mix tests.

## Side Effects

None.

## Checklist

- [X] Math issue #344


stan/math/prim/fun/reverse.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * @param x array to reverse
+ * @return Array in reversed order.
+ */
+template <typename T>
+inline std::vector<T> reverse(const std::vector<T>& x) {
+  std::vector<T> rev(x.size());
+  std::reverse_copy(x.begin(), x.end(), rev.begin());
+  return rev;
+}
+
```
andrjohns: You could replace the three definitions with a single ```apply_vector_unary``` here:
```
template <typename T>
inline auto reverse(const T& x) {
    return v.reverse();
  });
}
````

File: test/unit/math/mix/fun/reverse_test.cpp [hunk 2]
```diff
@@ -0,0 +1,19 @@
+#include <test/unit/math/test_ad.hpp>
+
+TEST(MathMixMatFun, reverse) {
+  auto f = [](const auto& x) { return stan::math::reverse(x); };
+
+  // 0 x 0
+  Eigen::VectorXd x0(0);
```
andrjohns: Should probably add tests for ```RowVectorXd``` and ```std::vector<double>``` as well

stan/math/prim/fun/reverse.hpp
andrjohns: I should note that this would also implicitly allow for ```reverse``` with nested arrays (e.g. ```reverse(std::vector<VectorXd>)```) which returns an array with each vector reversed

mcol: Good to know! But for `std::vector`, will this make a copy?

andrjohns: It has the same cost as the current implementations, where the result is evaluated into a vector which is then returned. ```Eigen::Map``` is used to evaluate directly into an ```std::vector``` and avoid the need for an intermediary copy

bob-carpenter: > I should note that this would also implicitly allow for reverse with nested arrays

I think this would be super confusing.  I think there should be a reverse function that works the same for any type `T[]`, namely by reversing the elements.  If the `T` is a container type, I don't think it should switch to working elementwise.

If there really needs to be a form of `reverse` that reverses containers within a list, I think it should get a different name.

I feel very strongly that this pattern of behavior should hold everywhere.

stan/math/prim/fun/reverse.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <stan/math/prim/vectorize/apply_vector_unary.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return a copy of the specified vector, row vector or array
+ * in reversed order.
+ *
+ * This implicitly allows for reverse with nested arrays (e.g.
```
bob-carpenter: This needs to be changed so that it's not vectorized.  The function `reverse({a1, a2, a3})` should always return `{a3, a2, a1}` no matter what the types are.  That is, it should *not* return `{reverse(a1), reverse(a2), reverse(a3)}`.  If you want that, define a different function.

bob-carpenter:
I'm overreviewing this one because I feel very strongly it would be wrong to have `reverse` apply elementwise to a `real[ , ]` or to `vector[]` or `row_vector[]` in Stan.

stan/math/prim/fun/reverse.hpp
andrjohns: Sounds good to me, @mcol can you revert this change (sorry for the detour!). You could instead collapse the two Eigen definitions, so that you just have one definition for Eigen vectors (row or col) and one for std::vectors:
```
template <typename T>
inline std::vector<T> reverse(const std::vector<T>& x) {
  std::vector<T> rev(x.size());
  std::reverse_copy(x.begin(), x.end(), rev.begin());
  return rev;
}

template <typename T, require_eigen_vector_t<T>>
inline auto reverse(const T& x) {
  return x.reverse();
}
```

mcol: No worries, if anything I learned about `apply_vector_unary`! :)

stan/math/prim/fun/reverse.hpp
mcol: I've reverted the change, so it has the behaviour you describe.

stan/math/prim/fun/reverse.hpp
andrjohns: Sorry to be a pain, but that eigen definition should have had an ```.eval()``` at the end, otherwise it could cause the same issues as #1644:

```
template <typename T, require_eigen_vector_t<T>>
inline auto reverse(const T& x) {
  return x.reverse().eval();
}
```

bob-carpenter: Thanks!  After posting, I wanted to make sure I wasn't around the bend on this one, so I took an informal poll of the non-computer scientists in my office (I won't name names) and they expected a top-level reverse only, not an elementwise one.

Issue #1656, opened 2020-01-29
Make the relative tolerance in expect_near_rel continuous


martinmodrak:
## Description
As discussed [on Discourse](https://discourse.mc-stan.org/t/expect-near-rel-behaves-weirdly-for-values-around-1e-8/12573/6), `expect_near_rel` behaves weirdly for values around zero. Most importantly, the relative tolerance jumps up at the tolerance value - here is the relative tolerance vs. average of the tested values in the current implementation, assuming `tol = 1e-8`:

![image](https://user-images.githubusercontent.com/9483603/73368491-591def00-42b1-11ea-9ca7-6c2e494419a8.png)

This makes it hard to write tests that are strict for values near zero while not being too strict for large values.

## Proposed solution

I propose that `expect_near_rel` has two tolerances _relative_ (`tol_rel`) and _minimal_ (`tol_min`). The final tolerance  is then computed as 
```
auto avg = 0.5 * (fabs(x1) + fabs(x2));
tol_final = max(tol_min, tol_rel * avg);
EXPECT_NEAR(x1, x2, tol_final);
```
where `x1` and `x2` are the values we are comparing.

The relative tolerance now increases smoothly when approaching zero as seen in this plot made with `tol_rel = 1e-8, tol_min = 1e-16`.
![image](https://user-images.githubusercontent.com/9483603/73368840-f8db7d00-42b1-11ea-8b99-536d4dfecdda.png)

It is slightly unclear what should be the default value for `tol_min`. Using `tol_min = tol_rel` means the tolerance stays the same or is increased (relative to the current implementation). Using `tol_min = tol_rel ^ 2` means the tolerance stays the same or decreases.  Values in between would make the tolerance sometimes increase sometimes decrease.
Currently the default tolerance is `1e-8`. Having `tol_min = 1e-8` is quite lax, while having `tol_min = 1e-16` is very strict (I can't make almost any slightly strong test pass with `tol_min = 1e-16`).

I would currently prefer somehing in between, possibly `tol_min = max(tol_rel ^ 2,1e-14)`  (as `1e-14` is IMHO achievable for many functions).

#### Current Version:
v3.1.0

mcol:
Fixed by #1657.

PR #1657, opened 2020-01-29
Improved behavior of expect_near_rel

martinmodrak:
## Summary
Improved the behavior of `expect_near_rel` as proposed in #1656.

Wherever a tolerance of type `double` was used, there is a new struct `relative_tolerance` that gathers relative (`tol`) and minimal tolerance (`tol_min`). The final tolerance is computed as `std::max(tol * fabs(x), tol_min)` where `x` is either an exact value to be tested against (I will use this in the precomputed tests) or average of the two values to be compared (this is what `expect_near_rel` uses).

To make the transition easy `relative_tolerance` has an _implicit_ constructor from double (which is treated as relative tolerance), I instructed the linter to be quiet about it.

Default `tol_rel` is `1e-8` and default `tol_min` is `std::max(tol_ * tol_, 1e-14)`. The `tol_min` default is slightly weird but it is a compromise between a desire to be strict by default while acknowledging that `1e-16` tolerance is almost impossible to achieve in many cases.

## Tests

There is a test for `relative_tolerance` - since the tolerance code is decoupled from `expect_near_rel` it can be tested more directly than in the previous version. 

I had to tweak the tolerances for a bunch of tests.

## Side Effects

Most tests got stricter or stayed the same for all/most values, but some test/value combinations got a bit more relaxed. Specifically:

* Any test that had  `tol > 1e-7` and the tolerance was not explicitly touched in this PR is stricter for `fabs(value) < tol` and the same for `fabs(value) > tol`.
* Any test that had  `1e-14 < tol < 1e-7` and the tolerance was not explicitly touched in this PR is stricter for `fabs(value) < tol`, slightly relaxed for `tol < fabs(value) < 1e-14/tol` and the same for `fabs(value) > 1e-14/tol`.
* In tests I touched I always aimed to got stricter somewhere, i.e. if I had to increase `tol_min` above it's default, I would decrease `tol` or vice versa. Mostly tests that modified their tolerance actually needed to modify `tol_min` and not `tol`.
* The default relative tolerance for `hessian_hessian_` and `hessian_fvar_hessian_` used to be `1e-3`. I noticed that this was necessary primarily because of bad behaviour around zero. The new default is `tol = 1e-4, tol_min = 1e-3`, so those tests got stricter above `1e-1` and weaker between `1e-3` and `1e-1`. 

Most places where I had to tweak the tolerance already had tolerance tweaked before. The exceptions are 
* `test/unit/math/mix/fun/matrix_exp_test.cpp`
* `test/unit/math/mix/fun/quad_form_diag_test.cpp`

where I had to slightly increase tolerances for at least one `test_ad` call that used default tolerance before. This was due to the change in default tolerance described above.

I also removed functions `relative_diff` and `expect_near_relative` from `test/unit/math/rev/fun/util.hpp` and replaced their use (only in 1 file) by `expect_near_rel`.

## Checklist

- [x] Math issue #1656 


martinmodrak:
@bob-carpenter could you review this? Or should I ask someone else?

bob-carpenter:
I can review it.  Thanks for documenting so clearly what's going on.

test/unit/math/relative_tolerance.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * tolerance that should be tested against. The final tolerance is computed as
+ * max(tol * fabs(x), tol_min)
+ * Where x is either an exact value to be tested against or average of
+ * two inexact values to be compared.
+ */
+struct relative_tolerance {
+  /**
+   * The relative tolerance
+   */
+  double tol;
```
bob-carpenter: The convention we've adopted is that member variables get underscore suffixes and nothing else does.  So these should be `tol_` and `tol_min_` and the args to the constructor should be the ones without underscores.

test/unit/math/ad_tolerances.hpp
```diff
@@ -50,11 +53,11 @@ struct ad_tolerances {
 
         hessian_val_(1e-8),
         hessian_grad_(1e-4),
-        hessian_hessian_(1e-3),
+        hessian_hessian_(1e-4, 1e-3),
```
bob-carpenter: [comment]
Thanks for making this so easily backward compatible.

martinmodrak:
So I checked the coding guidelines and I realized `relative_tolerance` should have really been a `class`, so I changed it to one. Also added some accessor/modifier functions while I was at it. I briefly documented the changes at https://github.com/stan-dev/stan/wiki/Testing-framework but I mostly link to code as this will be less stale.

martinmodrak:
@bob-carpenter tests are passing, should be ready for re-review

PR #1663, opened 2020-01-31
correctness fixes in neg_binomial_* functions

### LINKED ISSUE #595: Make stan/math/prim/scal/prob/neg_binomial_lccdf.hpp more efficient
#### Summary:
There are terms that are being computed multiple times and we should be using `inv(x)` instead of `1 / x`.

See @bob-carpenter's [comment](https://github.com/stan-dev/math/pull/588#discussion_r130223121)

#### Current Version:
v2.16.0

### LINKED ISSUE #1531: Cleanup code in neg_binomial_2_lpmf and neg_binomial_2_log_lpmf
## Description

There is some potentially dangerous code in neg_binomial_2_lpmf and neg_binomial_2_log_lpmf that I think should be made more robust. There is also possible there is a bug.

## Example

This example is from [neg_binomial_2_lpmf](https://github.com/stan-dev/math/blob/develop/stan/math/prim/scal/prob/neg_binomial_2_lpmf.hpp) but `neg_binomial_2_log_lpmf ` has a similar example.

In line 47 we have

```cpp
size_t size = max_size(n, mu, phi);
```

A bit later down we have

```cpp
  for (size_t i = 0, size = length(mu); i < size; ++i) {
    mu__[i] = value_of(mu_vec[i]);
  }

  for (size_t i = 0, size = length(phi); i < size; ++i) {
    phi__[i] = value_of(phi_vec[i]);
  }

  for (size_t i = 0, size = length(phi); i < size; ++i) {
    log_phi[i] = log(phi__[i]);
  }
```
So we basically overwrite the size we calculated in line 47 which is never used. And we then have

```cpp
for (size_t i = 0; i < size; i++) {
    if (include_summand<propto>::value) {
      logp -= lgamma(n_vec[i] + 1.0);
    }
    if (include_summand<propto, T_precision>::value) {
      logp += multiply_log(phi__[i], phi__[i]) - lgamma(phi__[i]);
    }
  // much more code
```

Can someone with the knowledge of this codebase double-check. It does seem at least a bit fishy.

#### Current Version:
v3.0.0

### LINKED ISSUE #1662: wrong gradient computations in neg_binomial_*cdf functions
## Description
All cumulative distribution functions for `neg_binomial` contain code such as the following:
https://github.com/stan-dev/math/blob/dc096aa6a66f7cd95cb29b50fe47680556187aa1/stan/math/prim/prob/neg_binomial_lccdf.hpp#L62-L71

Note that while the loop goes up to the `size(alpha)`, the code also uses elements from `n_vec`. This means that if `n` is an array while `alpha` is a scalar, only the first element of `n` will be used, while the rest are going to be ignored. The quantities computed there are then used in gradients, so those will be wrong.

A similar bug is present also in `neg_binomial_2_cdf`, while `neg_binomial_2_lccdf` is also affected since it delegates to `neg_binomial_lccdf`. On the other hand, `neg_binomial_2_lcdf` seems to be fine.

This was initially discussed in https://discourse.mc-stan.org/t/suspicious-code-in-neg-binomial-lccdf/12889.

#### Current Version:
v3.1.0

mcol:
## Summary
The various `neg_binomial` and `neg_binomial_2` functions suffer from several cases in which the loops that construct VectorBuilders are executed the wrong number of times (see examples pointed out in #1662 and #1531). Besides that, these functions can be cleaned up a bit so that intermediate computations are stored in VectorBuilders if they are reused later on, as well as using `inv` and `square`, and a few other touches here and there.

If it helps, I've split the correctness fixes and the rest of cleanup in two different commits. Fixes #1662, fixes #1531, fixes #595.

## Tests
Existing tests should continue to pass. I'm not sure that the current distribution test framework allows to test inputs of different sizes.

## Side Effects

None.

## Checklist

- [X] Math issue #1662


stan/math/prim/prob/neg_binomial_2_cdf.hpp
```diff
@@ -48,18 +52,18 @@ return_type_t<T_location, T_precision> neg_binomial_2_cdf(
 
   VectorBuilder<!is_constant_all<T_precision>::value, T_partials_return,
                 T_precision>
-      digamma_phi_vec(size(phi));
-
-  VectorBuilder<!is_constant_all<T_precision>::value, T_partials_return,
+      digamma_phi_vec(size_phi);
+  VectorBuilder<!is_constant_all<T_precision>::value, T_partials_return, T_n,
```
bob-carpenter: `T_n` has to be constant because it's an integer or array of integers, so it doesn't need to show up in `!is_constant`.

stan/math/prim/prob/neg_binomial_2_lcdf.hpp
```diff
@@ -29,23 +30,25 @@ return_type_t<T_location, T_precision> neg_binomial_2_lcdf(
...
 
-  size_t size_n = size(n);
   VectorBuilder<true, return_type_t<T_n>, T_n> np1(size_n);
   for (size_t i = 0; i < size_n; i++) {
-    if (n_vec[i] < 0) {
-      return log(0.0);
-    } else {
-      np1[i] = n_vec[i] + 1.0;
-    }
+    np1[i] = n_vec[i] + 1.0;
```
bob-carpenter: [optional]
These don't need the `.0` --- just `1` is fine.  The  compiler will promote.

stan/math/prim/prob/neg_binomial_2_log_lpmf.hpp
```diff
@@ -48,58 +48,63 @@ return_type_t<T_log_location, T_precision> neg_binomial_2_log_lpmf(
...
 
   operands_and_partials<T_log_location, T_precision> ops_partials(eta, phi);
 
-  size_t len_ep = max_size(eta, phi);
-  size_t len_np = max_size(n, phi);
-
-  VectorBuilder<true, T_partials_return, T_log_location> eta__(size(eta));
-  for (size_t i = 0, max_size_seq_view = size(eta); i < max_size_seq_view;
-       ++i) {
-    eta__[i] = value_of(eta_vec[i]);
```
bob-carpenter: [comment]
Thanks for cleaning up the bad naming.

stan/math/prim/prob/neg_binomial_lcdf.hpp
```diff
@@ -34,39 +36,46 @@ return_type_t<T_shape, T_inv_scale> neg_binomial_lcdf(const T_n& n,
...
   // Explicit return for extreme values
   // The gradients are technically ill-defined, but treated as zero
-  for (size_t i = 0; i < size(n); i++) {
+  for (size_t i = 0; i < size_n; i++) {
     if (value_of(n_vec[i]) < 0) {
       return ops_partials.build(negative_infinity());
     }
   }
 
+  VectorBuilder<!is_constant_all<T_shape>::value, T_partials_return, T_n>
```
bob-carpenter: The `T_n` shouldn't  be necessary here or elsewhere for constant.

stan/math/prim/prob/neg_binomial_2_cdf.hpp
mcol: Just to clarify: the  `T_n` is not inside `is_constant`, but among the typenames of the variables used in the `VectorBuilder`. So I think I should keep it.

bob-carpenter: Thanks.  I initially misread where the closing `>` was.

mcol:
I've added a bunch of tests (one revealed that one of my patches was incomplete). Ideally this sort of things would be tested with a specific framework or inside the distribution tests, but that would be quite a large undertaking.

These test revealed that for inputs of different lengths:
- `neg_binomial_cdf`, `neg_binomial_lcdf`, `neg_binomial_lccdf`, `neg_binomial_2_cdf` and `neg_binomial_2_lccdf` had buggy derivatives
- `neg_binomial_lpmf`, `neg_binomial_2_lpmf`, `neg_binomial_2_lcdf` and `neg_binomial_2_log_lpmf` were fine

The issues reported in #1531, albeit very confusing, were actually benign, as the variables that appeared to be reassigned in the for loop, were actually local variables within the loop, so they had no negative impact on the rest of the function. Those got cleaned up anyway by earlier patches.

mcol:
I'm getting a failure in the `neg_binomial_2_lcdf` distribution tests:
```
unknown file: Failure
C++ exception with description "beta_lcdf: Second shape parameter[1] is -2.14748e+09, but must be > 0!" thrown in the test body.
[  FAILED  ] AgradCdfLogNegBinomial2_ffv_80/AgradCdfLogTestFixture/0.UpperBound, where TypeParam = boost::mpl::vector<AgradCdfLogNegBinomial2, boost::mpl::vector<Eigen::Matrix<int, -1, 1, 0, -1, 1>, Eigen::Matrix<stan::math::fvar<stan::math::fvar<stan::math::var> >, -1, 1, 0, -1, 1>, Eigen::Matrix<stan::math::fvar<stan::math::fvar<stan::math::var> >, -1, 1, 0, -1, 1>, empty, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na> (0 ms)
```
The only change I had in that file was replacing a 1.0 with a 1. Locally, switching it back makes the test pass. I don't understand why this would be a problem, but I'm reverting that optional change and hope tests will pass.

bob-carpenter:
As long as that's not an argument, you should be OK.  If it succeeds with 1.0 as an argument to the lcdf but not 1, that's a bug.

mcol:
It was in an assigmnent to a VectorBuilder, see the last commit.

PR #1575, opened 2020-01-02
Additional tests for distributions

martinmodrak:
## Summary

Additional tests for distributions to increase coverage. All tests pass with the current codebase.

## Tests

- `bernoulli_logit_lpmf` involves a cutoff, continuity around the cutoff is tested
- `inv_gamma_lpmf` and `weibull_lpmf` had no tests for derivatives. A new test testing all derivatives  against values precomputed byt Mathematica was added.
- adapted the tests from #1497 to `neg_binomial_lpmf` (test against precomputed values from Mathematica, test against complex step, testing continuity at cutoff for `alpha`, testing extreme values of `alpha`)

## Side Effects

None

## Checklist

- [x] Math issue: I decided not to open an math issue for this, as it seemed like clutter. Will open, if desired.

- [x] Copyright holder: Institute of Microbiology of the Czech Academy of Sciences

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1664, opened 2020-02-01
Revert "Additional tests for distributions"

mcol:
Reverts stan-dev/math#1575. Sorry, I hadn't noticed that it still contained some paths from before the end of flatten.


mcol:
I'm not sure what the best approach is, so I posted #1665 to just fix those headers without doing a full revert, which seems much more work for such an easy fix.

rok-cesnovar:
I think this is fine. Its an easy and quick fix. We revert when there is something to investigate. If the time to fix = time to revert I see no point of reverting.

mcol:
Cool, so I'm closing this, we'll proceed with #1665.

PR #1666, opened 2020-02-02
add the range_* functions

### LINKED ISSUE #838: regular array and vector constructor functions
#### Summary:

Create functions to create regular arrays. 

Here's one from MATLAB (also related to Python `range`):

```
/**
 * Return vector of evenly spaced elements with given minimum and maximum.  Behavior for size = 0 and size = 1 to be determined.  Requires min <= max.
 */
vector linspace_vector(real min, real max, int size);
```
There could also be `linspace_array` and `linspace_row_vector` versions. It might be possible to call the vector or array version just `linspace`.  

Python provides `range`, which is a bit different.

```
/**
 * Return sequence of numbers min, min + by, min + 2*by, ...
 * up to but not including max.  Requires min <= max and by > 0.
 */
vector range_vector(real min, real max, real by);
```

In R, there's a function `seq` that allows either of these parameterizations, but we don't want to follow that redundant pattern.

This one's a suggestion from peterzh on http://discourse.mc-stan.org/t/create-vector-from-a-number-range/3760.

This will then get pushed through to Stan.

#### Current Version:
v2.17.0

mcol:
## Summary

This adds the `range_array`, `range_vector` and `range_row_vector` functions to help construct containers of linearly spaced elements between `low` and `high` and separated by `by`. Fixes #838.

This also renames the spaced_vector functions to linspaced_vector, following the suggestion and discussion in #838.

## Tests
Added tests.

## Side Effects

None.

## Checklist

- [X] Math issue #838

### DIFF
--- a/stan/math/prim/fun.hpp
+++ b/stan/math/prim/fun.hpp
@@ -137,6 +137,9 @@
 #include <stan/math/prim/fun/ldexp.hpp>
 #include <stan/math/prim/fun/LDLT_factor.hpp>
 #include <stan/math/prim/fun/lgamma.hpp>
+#include <stan/math/prim/fun/linspaced_array.hpp>
+#include <stan/math/prim/fun/linspaced_row_vector.hpp>
+#include <stan/math/prim/fun/linspaced_vector.hpp>
 #include <stan/math/prim/fun/lmgamma.hpp>
 #include <stan/math/prim/fun/log.hpp>
 #include <stan/math/prim/fun/log10.hpp>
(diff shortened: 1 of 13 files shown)


mcol:
This also renames the `spaced_vector` functions to `linspaced_vector`, following the suggestion and discussion in #838.

nhuurre:
I'm sorry, but would it be possible to _not_ add new nondifferentiable functions for no reason?
This range function coerces a real to an int, which is something Stan has intentionally not allowed before.

I don't see any use case for the range. If the size of the resulting vector is known ahead of time then `range_vector(low, high, by)` is the same as `linspaced_vector(N, low, low + by*(N-1))`.
If the size is not known then it's not even possible to store the vector in a local variable [because those must have a declared size.](https://discourse.mc-stan.org/t/ragged-arrays-bug-in-the-language/12566)

mcol:
The requirement for having sizes defined ahead of time is a killer for `range`, so I'm closing this. When I get to document `linspaced` I can mention how it can be used as a workaround for `range`. Thanks @nhuurre for your input!

bob-carpenter:
> If the size is not known then it's not even possible to store the vector in a local variable because those must have a declared size.

The one workaround to this general rule is calling functions.  You can have an expression with no declared size and then pass it as an argument to a function with no declared size.  The function argument acts like an unsized local variable if you really do need one.

linwangidd:
Thanks so much to have `linspaced_vector`!

PR #1667, opened 2020-02-02
Refactor Eigen Type traits

SteveBronder:
## Summary

Cleans up and expands the type traits for selecting certain Eigen types. This uses Eigen's inheritance structure to determine for each type trait whether the type we are checking inherits a particular Eigen base class. In order, the hierarchy for types is essentially

- `EigenBase` 
    - `DenseBase`
        - `MatrixBase`
            - `Matrix`
        - `ArrayBase`
            - `Array`
        - `MapBase`
        - `LDLT`
        - `LLT`
    - `SparseMatrixBase`
        - `SparseCompressedMatrixBase` 
            - `SparseMatrix`

This PR includes functions to check types that inherit from `EigenBase`, `DenseBase`, `MatrixBase`, `ArrayBase`, `MapBase`, and `SparseMatrixBase`.

This PR also changes the definition of `is_eigen_matrix` to be anything that inherits from `Eigen::MatrixBase` with more than one row and column. There were places where we want to specialize for eigen vectors while having other specialization for general eigen matrices. This meant we had to do something like

```cpp
template <typename EigMat, typename = require_eigen_t<EigMat>,
  typename = require_not_eigen_vector_t<EigMat>>
```

Which is annoying. With this PR we can just write

```cpp
template <typename EigMat, typename = require_eigen_matrix_t<EigMat>>
```

## Tests

Tests are done for generic eigen expressions inheriting from each type above and for all stan scalar types. They can be run with

```cpp
./runTests.py ./test/unit/math/mix/meta/ -f eigen
```

## Side Effects

Yes, the definition of is_eigen_matrix will now accept more generic eigen expressions.

## Checklist

- [x] Math issue #1308


syclik:
Conceptually, wouldn’t it be better if we called this `is_eigen_dense` or even something more verbose like `is_eigen_densebase`? That would allow us to specialize for `Eigen::Matrix` if we ever needed to. And if we don’t, at least it’s clear that we don’t actually mean `Eigen::Matrix`.

SteveBronder:
> Conceptually, wouldn’t it be better if we called this is_eigen_dense or even something more verbose like is_eigen_densebase? That would allow us to specialize for Eigen::Matrix if we ever needed to. And if we don’t, at least it’s clear that we don’t actually mean Eigen::Matrix.

What if we just switched it to `Eigen::MatrixBase`? That type has all the attributes of an Eigen matrix. My only concern is that there are functions in Eigen that return `DenseBaseCoeff` classes, I'm not sure if those would be caught if we look for just things that inherit from `MatrixBase`.

I think with the direction things are going we wouldn't want an explicit specialization for `Eigen::Matrix`. imo it's about deciding whether a DenseBase is conceptually enough to say, "This is matrix-ish"

@t4c1 ?

t4c1:
I agree with @syclik. I would prefere to keep the name `is_eigen_matrix` free in case we ever need it to distinguish between Eigen `Matrix` and `Array`. I would propose to name the thing you are trying to do here `is_eigen_and_not_vector` or something similar.

> What if we just switched it to Eigen::MatrixBase?

What exactley do you mean by switch to it?

bob-carpenter:
What are you trying to match.  What will is_eigen_and_not_vector match besides Matrix<T, -1, -1>?

t4c1:
It will also match any expression that evaluates to `Matrix<T, -1, -1>`. Hierarchy of Eigen classes should help determining how things can be matched: https://eigen.tuxfamily.org/dox/TopicClassHierarchy.html

bob-carpenter:
How about is_eigen_matrix_assignable instead or is_assignable_eigen_matrix to keep our naming in line with std::is_assignable?

Or given that we don't have other types of matrices, we could just drop the "eigen" here.

SteveBronder:
> I agree with @syclik. I would prefer to keep the name is_eigen_matrix free in case we ever need it to distinguish between Eigen Matrix and Array. I would propose to name the thing you are trying to do here is_eigen_and_not_vector or something similar.

idk if I like having the concept be named by what it's not. I have to think about that. How about `is_eigen_dense_matrix`? I can clean up the definition here to be

" `is_eigen_dense_matrix` will hold a static bool defined as true if the template parameter T both inherits from `Eigen::DenseBase` and neither the compile time number of rows and columns is 1"

So the above fits for dynamic Eigen matrices and fixed size matrices. Does that conceptually/practically seem clear enough? Though it leaves a `Eigen::Matrix<double, 1, 1>` as not a matrix (a row _and_ column vector I guess?) I'm not sure how I feel about that. 

>> What if we just switched it to Eigen::MatrixBase?

> What exactly do you mean by switch to it?

Jus switching the above definition to say Eigen::MatrixBase

> How about is_eigen_matrix_assignable instead or is_assignable_eigen_matrix to keep our naming in line with std::is_assignable?

Are there other places we use `std::is_assignable`?

> Or given that we don't have other types of matrices, we could just drop the "eigen" here.

`matrix_cl<>`!

bob-carpenter:
@SteveBronder :  I think our suggestions crossed in the ether.  I prefer `is_eigen_matrix_assignable` or `is_assignable_eigen_matrix` (or without the `eigen_`) to follow `std::is_assignable`.  And that's what we're talking about here---assignability.  That's what an `eval()` does.

t4c1:
> idk if I like having the concept be named by what it's not.

I think that is the only way not to be ambiguaus. The problem is that row and column vectors are special cases of matrices and I don't think we have a word for a matrix that is not a vector. 

Calling it either `is_eigen_matrix`, `is_dense_matrix` or `is_matrix_assignable` does not convey that vectors are not accepted.

bob-carpenter:
Despite the superficial similarity, there are three distinct overloads of `Matrix<T, R, C>` that we use.  Given how C++ works, the only relation among these classes is that they all overload the base template class.  Each has its own distinct implementation and produces different return types for operators like `operator*`.    I suggest we settle on the terminology:

* matrix:  `Matrix<T, -1, -1>`
* column vector: `Matrix<T, -1, 1>`
* row vector: `Matrix<T, 1, -1>

We can use "vector" for either row or column vectors, though in informal math writing, when we say "vector" we usually mean column vectors.

t4c1:
If we use suggested terminology the term "matrix" becomes ambiguous. It could mean any matrix or just `Matrix<T,-1,-1>`.

bob-carpenter:
There's only one matrix, Matrix<T, -1, -1>.  The other two classes are vectors.

SteveBronder:
> If we use suggested terminology the term "matrix" becomes ambiguous. It could mean any matrix or just Matrix<T,-1,-1>.

I think there's two points to call here

1. it matters whether you consider whether a row/col vector is also a matrix. In math row/col vectors are certainly matrices, but when doing code stuff it feels like we handle them pretty differently. Hence the definition/PR. 
2. The definition I give below explicitly allows for fixed matrices like `Eigen::Matrix<double, 10, 10>`. At this time we don't use those (Unless a computation gives back a `[1, 1]`, but with the new compiler if a user tells us the size of a vector/matrix during compile time we could use fixed matrices and do much fewer dynamic allocations

how would you want to correct the below definition? Or is it just the name?

>> > " is_eigen_dense_matrix will hold a static bool defined as true if the template parameter T both inherits from Eigen::DenseBase and neither the compile time number of rows ~and~ _or_ columns is 1"

ftr I'm thinking along Bob's line in the below where computationally we treat vectors different conceptually in the codebase

> Despite the superficial similarity, there are three distinct overloads of Matrix<T, R, C> that we use. Given how C++ works, the only relation among these classes is that they all overload the base template class. Each has its own distinct implementation and produces different return types for operators like operator*. I suggest we settle on the terminology:

> - matrix: Matrix<T, -1, -1>
> - column vector: Matrix<T, -1, 1>
> - row vector: `Matrix<T, 1, -1>
> We can use "vector" for either row or column vectors, though in informal math writing, when we say "vector" we usually mean column vectors.

bob-carpenter:
> In math row/col vectors are certainly matrices, but when doing code stuff it feels like we handle them pretty differently. 

Exactly.  It's like the relation between integers and real numbers, or between reals and complex.

> but with the new compiler if a user tells us the size of a vector/matrix during compile time we could use fixed matrices and do much fewer dynamic allocations

This is tricky.  Eigen itself only recommends doing this for very small matrices (single digit sizes) because everything would otherwise be allocated on the stack, which is of very limited size, especially with multithreading (on the order of 20KB, so anything beyond about 50 x 50 is going to blow out stack memory).

> how would you want to correct the below definition? 

`is_eigen_matrix_assignable<T>::value` is `true` if the type `T` is assignable to type `Eigen::Matrix<U, 1, -1>`, where `U` is the scalar type of `T`.

SteveBronder:
Mulling this over I like this definition

> is_eigen_matrix_assignable<T>::value is true if the type T is assignable to type Eigen::Matrix<U, -1, -1>, where U is the scalar type of T.

Though I changed it slightly to be just

> Checks whether type T is derived from Eigen::MatrixBase and has columns and
 * rows not equal to 1.

imo I think this makes sense because this is the base class for eigen matrices. See below for Eigen's reference on that

https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

SteveBronder:
This is ready for review

stan/math/prim/meta/is_eigen.hpp
```diff
@@ -33,6 +34,35 @@ struct is_eigen_base<Eigen::EigenBase<T>> : std::true_type {};
...
+ */
+template <typename T>
+struct is_eigen_dense_base
+    : std::integral_constant<bool,
+                             std::is_base_of<Eigen::DenseBase<T>, T>::value> {};
+
+template <typename T>
+struct is_eigen_dense_base<Eigen::DenseBase<T>> : std::true_type {};
+
+}  // namespace internal
```
bob-carpenter: This can all go in the same namespace.  

But then our convention has been one function per file with the file named after the function.  Let's ask @syclik what he thinks about changing that pattern.  I'm OK either way.

stan/math/prim/meta/is_eigen.hpp
```diff
@@ -43,14 +73,20 @@ struct is_eigen<
     : std::true_type {};
 
 namespace internal {
-template <typename T>
+template <typename T, typename Enable = void>
```
bob-carpenter: This new template parameter needs doc.  If it's called Enable, why set it to void by default?  Shouldn't it be boolean?

stan/math/prim/meta/require_generics.hpp
```diff
@@ -102,6 +102,13 @@ template <template <class...> class TypeCheck, class... Check>
 struct is_eigen_value_check
     : container_value_type_check_base<is_eigen, TypeCheck, Check...> {};
 
+/**
+ * Meta type trait to check if type inherits from EigenBase.
```
bob-carpenter: These should have their template parameters documented as well as what kind of value they're going to produce.   Same for all the rest of these traits classes.

stan/math/prim/meta/require_generics.hpp
```diff
@@ -629,6 +644,26 @@ using require_all_not_eigen_t = require_all_not_t<is_eigen<Types>...>;
 template <typename... Types>
 using require_any_not_eigen_t = require_any_not_t<is_eigen<Types>...>;
 
+template <typename T>
```
bob-carpenter: These are really neat.  

Was the decision that they dont need doc because they just follow the definitions and the standard library doesn't doc the `_t` types?  These are a bit different in that there's not underlying simple structs they'e wrapping.   I'd rather err in favor of at least simple doc for compounds like this.  Especially since the whole notion of "Eigen matrix" is fundamentally confused given that vectors and row vectors use a specialization of the `Matrix` type.

stan/math/prim/meta/require_generics.hpp
```diff
@@ -799,6 +834,30 @@ template <template <class...> class TypeCheck, class... Check>
 using require_all_not_eigen_vt
```
bob-carpenter: This really needs doc because I have no idea what the `v` is supposed to signal in `vt`.

stan/math/prim/meta/require_generics.hpp
```diff
@@ -935,6 +994,30 @@ template <template <class...> class TypeCheck, class... Check>
 using require_all_not_eigen_st
     = require_all_not_t<is_eigen_scalar_check<TypeCheck, Check>...>;
 
+template <template <class...> class TypeCheck, class... Check>
```
bob-carpenter: What does the `s` mean?  I see `is_eigen_matrix_scalar_check` but I don't know what that means, either as it seems to be mixing types (matrix and scalar).

test/unit/math/prim/meta/is_eigen_test.cpp
```diff
@@ -39,17 +39,25 @@ TEST(MathMetaPrim, expression) {
   EXPECT_FALSE((is_eigen_matrix<Eigen::EigenBase<Eigen::MatrixXd>>::value));
```
bob-carpenter: Every metaprogram needs tests.

bob-carpenter:
@SteveBronder :  Is something holding this up?  I'd rather not leave PRs dangling for weeks with no activity.

SteveBronder:
Apologies I've just been running around with job stuff I'll put fix this up today

SteveBronder:
I got a lot of it fixed tonight but didn't get it all so I'm going to close this and open it up later

stan/math/prim/meta/is_eigen.hpp
```diff
@@ -2,84 +2,40 @@
...
-/**
- * Checks whether type T is derived from EigenBase. If true this will have a
- * static member function named value with a type of true, else value is false.
- */
 template <typename T>
-struct is_eigen<
-    T, std::enable_if_t<internal::is_eigen_base<std::decay_t<T>>::value>>
+struct is_eigen<T, std::enable_if_t<std::is_base_of<
+                       Eigen::EigenBase<typename std::decay_t<T>::PlainObject>,
+                       typename std::decay_t<T>::PlainObject>::value>>
```
andrjohns: Just a general comment, but it might make these templates more readable if the ```typename std::decay_t<T>::PlainObject``` is moved up to the template definition:
```
template <typename T, typename PlainType = typename std::decay_t<T>::PlainObject>
struct is_eigen<T, std::enable_if_t<std::is_base_of<
                       Eigen::EigenBase<PlainType>,
                       PlainType>::value>>
```

SteveBronder: Agree!!

SteveBronder: Actually idt this would work since it would set the default of the template value but I have to grab `PlainObject` in one specialization and `MatrixType` in another

stan/math/prim/meta/require_generics.hpp
SteveBronder: The `st` stands for `scalar_type` it's doc'd on the website

http://mc-stan.org/math/d1/db9/group__require__meta.html

This PR is already pretty big so I'm going to open up another PR to move these

stan/math/prim/meta/require_generics.hpp
SteveBronder: Yeah in another PR I'm going to break these out into the respective meta functions so that the requires base stuff just sits in `require_generics.hpp` and the specializations are in each `is_*` file

stan/math/prim/meta/is_eigen.hpp
SteveBronder: Added doc. `Enable` is set to void here since when the other enable_if specializations fail the base class needs to be able to have all of it's template parameters resolved. T will be resolved by the user calling `is_eigen<Type>` but the Enable needs to also have a valid type to work.

What's the reason behind using boolean?

SteveBronder:
This is ready for review

stan/math/prim/meta/is_eigen_matrix_or_array.hpp
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+/**
+ * Check if a type satisfies either of @c is_eigen_matrix or @c is_eigen_array
+ * @tparam T Type to check if it is derived from `EigenBase`
+ */
+template <typename T>
+using is_eigen_matrix_or_array
+    = math::disjunction<is_eigen_matrix<std::decay_t<T>>,
+                        is_eigen_array<std::decay_t<T>>>;
```
t4c1: With the change to `is_eigen_matrix` this does not make sense anymore. It should be renamed to `is_eigen_matrix_or_vector_or_array` ant the test should be changed appropriately.

SteveBronder: It seems like the only place we are using `is_eigen_matrix_or_array` was in matrix_cl to check whether we should call .eval() or not. `is_eigen_matrix_or_vector_or_array` seems like kind of a mouthful. Should we just do `is_eigen_dense` in `matrix_cl` and get rid of this type trait?

Code that uses `is_eigen_matrix_or_array:
https://github.com/stan-dev/math/blob/develop/stan/math/opencl/matrix_cl.hpp#L312

t4c1: `is_eigen_dense` is not the same thing! But I agree we can find a better name. This is actually checking whether something is a plain type so maybe `is_eigen_plain_type`?

bob-carpenter:
I'm a little confused as to who's reviewing this now.  Should I do the final review now?

t4c1:
@bob-carpenter go ahead with the review. I just had one comment.

bob-carpenter:
I don't understand what any of those metaprograms do, so I still want doc. 

For example, I don't know what what the `_vt` means from my original review that's all over the place in `require_genercs`.  

And I'm afraid I can't even parse these headers:

```
template <template <class...> class ContainerCheck,
          template <class...> class TypeCheck, class... Check>
```

I think you may need someone with more skill in templates to review this. I'm totally slammed right now and this whole file's full of things I don't understand.

SteveBronder:
@t4c1 if this is too large to be nice to review I think I could break it up into at least 2 PRs with

1. The testing framework and moving 1 or 2 type traits (like eig row and col vector)
2. Then the rest

Or (2) could be broken up into 2 more PRs as well

bob-carpenter:
I can take a look at this if @t4c1 doesn't have time.  Let me know.

t4c1:
@bob-carpenter I you want to review, go ahead. Otherwise I can do it.
@SteveBronder If I am reviewing, I would prefere this to be split in 2 or 3 PRs.

SteveBronder:
I'm very fine with splitting this into 2 or 3 PRs

bob-carpenter:
If you could review, @tadej, that'd be great as I'm swamped as usual.

PR #1680, opened 2020-02-06
Add discrete_range_rng(lower, upper) and discrete_range_lpmf

### LINKED ISSUE #719: Add discrete uniform rng for stan#2457
#### Summary:
Add a discrete uniform rng for https://github.com/stan-dev/stan/issues/2457.

#### Additional Information:
Can use boost.

#### Current Version:
v2.17.0

mcol:
## Summary

This adds support for new functions `discrete_range_rng(lower, upper)` and `discrete_range_lpmf(y, lower, upper)`, which allow to return a random discrete uniform variate between lower an upper (both extremes inclusive) and its log PMF. Fixes #719.

## Tests

Added. As far as I understand, there's no need to enforce `lower` and `upper` to be integers or integer arrays here, as that will be done when we expose the signatures, so tests don't include the case for which the arguments are not integers.

## Side Effects

None.

## Checklist

- [X] Math issue #719

### DIFF
--- a/stan/math/prim/prob/discrete_range_lpmf.hpp
+++ b/stan/math/prim/prob/discrete_range_lpmf.hpp
@@ -1,0 +1,96 @@
+#ifndef STAN_MATH_PRIM_PROB_DISCRETE_RANGE_LPMF_HPP
+#define STAN_MATH_PRIM_PROB_DISCRETE_RANGE_LPMF_HPP
+
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/constants.hpp>
+#include <stan/math/prim/fun/inv.hpp>
+#include <stan/math/prim/fun/log.hpp>
+#include <stan/math/prim/fun/max_size.hpp>
+#include <stan/math/prim/fun/size_zero.hpp>
+#include <stan/math/prim/fun/value_of.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/** \ingroup prob_dists
+ * Return the log PMF of a discrete range for the given y, lower and upper
+ * bound.
+ *
+ * `lower` and `upper` can each be a scalar or a one-dimensional container.
+ * Any non-scalar inputs must be the same size.
+ *
+ * @tparam T_y type of scalar, either int or std::vector<int>
+ * @tparam T_lower type of lower bound, either int or std::vector<int>
+ * @tparam T_upper type of upper bound, either int or std::vector<int>
+ *
+ * @param y random variable
+ * @param lower lower bound
+ * @param upper upper bound
+ * @return Log probability. If containers are supplied, returns the log sum
+ * of the probabilities.
+ * @throw std::domain_error if upper is smaller than lower.
+ * @throw std::invalid_argument if non-scalar arguments are of different
+ * sizes.
+ */
+template <bool propto, typename T_y, typename T_lower, typename T_upper>
+double discrete_range_lpmf(const T_y& y, const T_lower& lower,
+                           const T_upper& upper) {
+  static const char* function = "discrete_range_lpmf";
+
+  using std::log;
+
+  if (size_zero(y, lower, upper)) {
+    return 0.0;
+  }
+
+  check_not_nan(function, "Random variable", y);
+  check_consistent_sizes(function, "Lower bound parameter", lower,
+                         "Upper bound parameter", upper);
+  check_greater_or_equal(function, "Upper bound parameter", upper, lower);
+
+  if (!include_summand<propto>::value) {
+    return 0.0;
+  }
+
+  double logp(0.0);
+
+  scalar_seq_view<T_y> y_vec(y);
+  scalar_seq_view<T_lower> lower_vec(lower);
+  scalar_seq_view<T_upper> upper_vec(upper);
+  size_t size_lower_upper = max_size(lower, upper);
+  size_t N = max_size(y, lower, upper);
+
+  for (size_t n = 0; n < N; ++n) {
+    const double y_dbl = value_of(y_vec[n]);
+    if (y_dbl < value_of(lower_vec[n]) || y_dbl > value_of(upper_vec[n])) {
+      return LOG_ZERO;
+    }
+  }
+
+  VectorBuilder<true, double, T_lower, T_upper> log_upper_minus_lower(
+      size_lower_upper);
+
+  for (size_t i = 0; i < size_lower_upper; i++) {
+    const double lower_dbl = value_of(lower_vec[i]);
+    const double upper_dbl = value_of(upper_vec[i]);
+    log_upper_minus_lower[i] = log(upper_dbl - lower_dbl + 1);
+  }
+
+  for (size_t n = 0; n < N; n++) {
+    logp -= log_upper_minus_lower[n];
+  }
+
+  return logp;
+}
+
+template <typename T_y, typename T_lower, typename T_upper>
+inline double discrete_range_lpmf(const T_y& y, const T_lower& lower,
+                                  const T_upper& upper) {
+  return discrete_range_lpmf<false>(y, lower, upper);
+}
+
+}  // namespace math
+}  // namespace stan
+#endif
--- a/stan/math/prim/prob/discrete_range_lpmf.hpp
+++ b/stan/math/prim/prob/discrete_range_lpmf.hpp
@@ -16,18 +16,25 @@ namespace math {
 
 /** \ingroup prob_dists
  * Return the log PMF of a discrete range for the given y, lower and upper
- * bound.
+ * bound (all integers).
+ *
+ \f{eqnarray*}{
+   y &\sim& \mbox{\sf{discrete\_range}}(lower, upper) \\
+     \log(p (y \, |\, lower, upper))
+        &=& \log \left( \frac{1}{upper - lower + 1} \right) \\
+        &=& -\log (upper - lower + 1)
+ \f}
  *
  * `lower` and `upper` can each be a scalar or a one-dimensional container.
- * Any non-scalar inputs must be the same size.
+ * Any container arguments must be the same size.
  *
  * @tparam T_y type of scalar, either int or std::vector<int>
  * @tparam T_lower type of lower bound, either int or std::vector<int>
  * @tparam T_upper type of upper bound, either int or std::vector<int>
  *
- * @param y random variable
- * @param lower lower bound
- * @param upper upper bound
+ * @param y integer random variable
+ * @param lower integer lower bound
+ * @param upper integer upper bound
  * @return Log probability. If containers are supplied, returns the log sum
  * of the probabilities.
  * @throw std::domain_error if upper is smaller than lower.
@@ -38,7 +45,6 @@ template <bool propto, typename T_y, typename T_lower, typename T_upper>
 double discrete_range_lpmf(const T_y& y, const T_lower& lower,
                            const T_upper& upper) {
   static const char* function = "discrete_range_lpmf";
-
   using std::log;
 
   if (size_zero(y, lower, upper)) {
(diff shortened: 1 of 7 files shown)


stan/math/prim/prob/discrete_range_rng.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * Any non-scalar inputs must be the same size.
+ *
+ * @tparam T_lower type of lower bound
+ * @tparam T_upper type of upper bound
+ * @tparam RNG type of random number generator
+ *
+ * @param lower lower bound
+ * @param upper upper bound
+ * @param rng random number generator
+ * @return A (sequence of) discrete random variate(s) between `lower` and
```
bob-carpenter: The return needs to be a single integer when both arguments are integers, and a standard vector of integers otherwise.  Is that what this is supposed to mean?  Is it how the other RNGs are documnted?  I find it a bit confusing, but if it's like all the other ones, we should go with consistency for now.

stan/math/prim/prob/discrete_range_rng.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * `upper`, both bounds included.
+ * @throw std::domain_error if lower or upper are non-finite, or if upper
+ * is smaller than lower
+ * @throw std::invalid_argument if non-scalar arguments are of different
+ * sizes
+ */
+template <typename T_lower, typename T_upper, class RNG>
+inline typename VectorBuilder<true, int, T_lower, T_upper>::type
+discrete_range_rng(const T_lower& lower, const T_upper& upper, RNG& rng) {
```
bob-carpenter: The documentation needs to indicate what `T_lower` and `T_upper` can be (presumably either `int` or `std::vector<int>`).  

It should not be possible to compile an instantiation where the lower or upper bounds are `double`.  It's OK if that's enforced by not being able to compile someting inside---it's not like we have to guard against ambiguity.

stan/math/prim/prob/discrete_range_rng.hpp
```diff
@@ -0,0 +1,65 @@
...
+  using boost::random::uniform_int_distribution;
+  using boost::variate_generator;
+
+  static const char* function = "discrete_range_rng";
+
+  check_finite(function, "Lower bound parameter", lower);
```
bob-carpenter: The `check_finite` tests are not necessary because all integers are finite.

stan/math/prim/prob/discrete_range_rng.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/max_size.hpp>
+#include <boost/random/uniform_int_distribution.hpp>
+#include <boost/random/variate_generator.hpp>
+
+namespace stan {
+namespace math {
+
+/** \ingroup prob_dists
+ * Return a discrete random variate for the given lower and upper bounds
```
bob-carpenter: I think it'd help to say that it returns integer(s) between the lower and upper bound(s) inclusive and then elaborate on the return type in the longer doc.  As is, it's very unclear from reading this doc what the return type actually is.   I imagine a lot of the other RNG doc probaby has this same problem.

test/unit/math/prim/prob/discrete_range_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  double nan = std::numeric_limits<double>::quiet_NaN();
+  double inf = std::numeric_limits<double>::infinity();
+
+  Eigen::VectorXd lower(3);
+  Eigen::VectorXd upper(3);
+
+  lower << 5.0, 11.0, -15.0;
+  upper << 5.0, 15.0, -10.0;
+  EXPECT_NO_THROW(discrete_range_rng(lower, upper, rng));
```
bob-carpenter: This should be illegal, so it looks like that Boost thing lets integers get assigned.  So the signature should prevent this with an enable_if on the template parameters.  The tests need to be updated then to use integers.

test/unit/math/prim/prob/discrete_range_test.cpp
```diff
@@ -0,0 +1,73 @@
+#include <stan/math/prim.hpp>
+#include <boost/math/distributions.hpp>
+#include <boost/random/mersenne_twister.hpp>
+#include <gtest/gtest.h>
+#include <limits>
+#include <vector>
+
+TEST(ProbDistributionsDiscreteRange, error_check) {
```
bob-carpenter: The tests need to consider boundary conditions like (lower = 17, upper = 17) and (lower = 17, upper = 18) to make sure we get the right boundary behavior.

test/unit/math/prim/prob/discrete_range_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+  boost::random::mt19937 rng;
+
+  int N = 10000;
+  int lower = -3;
+  int upper = 20;
+
+  int K = upper - lower + 1;
+  boost::math::chi_squared mydist(K - 1);
+
+  int bin[K];
```
bob-carpenter: I'd prefer std::vector here with initializers.  E.g., 

```
std::vector<int> bin(K, 0);  //  initialize bins to 0
std::vector<double> expect(K, static_cast<double>(N) / K);
```

test/unit/math/prim/prob/discrete_range_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+  double expect[K];
+  double prop = static_cast<double>(N) / K;
+  for (int i = 0; i < K; i++) {
+    bin[i] = 0;
+    expect[i] = prop;
+  }
+
+  int count = 0;
```
bob-carpenter: This should be  a for loop:

```
for (int count = 0; count < N; ++count) {  ... }
```

test/unit/math/prim/prob/discrete_range_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+  while (count < N) {
+    int a = stan::math::discrete_range_rng(lower, upper, rng);
+    bin[a - lower]++;
```
bob-carpenter: The `++` should come before the variable if the variable before the increment's not being used.

test/unit/math/prim/prob/discrete_range_test.cpp
mcol: No other distribution uses any `enable_if` or `require_` on the template parameters or elsewhere. So my understanding was that this is  enforced by the signatures we expose in the compiler. I can try to concoct something, but it'd be a first in the codebase. I'll update the tests.

stan/math/prim/prob/discrete_range_rng.hpp
mcol: I've updated the docs. See my other answer as for the instantiation with non-int arguments.

mcol:
Thanks @bob-carpenter for the review. As for restricting instantiations to int or ints[], I have applied no changes as no other probability function does that. If you think it's really necessary, could you show me how to do it? I don't think we have `require_` for int or int array/vectors.

I went ahead and added the lpmf here, following what is done in `hypergeometric_lpmf`, which is the only other function with all int arguments. Since it's quite new territory for me, this deserves extra attention.

stan/math/prim/prob/discrete_range_lpmf.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <stan/math/prim/fun/max_size.hpp>
+#include <stan/math/prim/fun/size_zero.hpp>
+#include <stan/math/prim/fun/value_of.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/** \ingroup prob_dists
+ * Return the log PMF of a discrete range for the given y, lower and upper
```
bob-carpenter: The documentation should mention that only integers and integer arrays are legal arguments.  This doesn't necessarily have to be enforced anywhere by the function, but it should be documented.

stan/math/prim/prob/discrete_range_lpmf.hpp
```diff
@@ -0,0 +1,96 @@
...
+ * bound.
+ *
+ * `lower` and `upper` can each be a scalar or a one-dimensional container.
+ * Any non-scalar inputs must be the same size.
```
bob-carpenter: [optional]
I prefer "container" to "non-scalar" and "argument" to "input".

stan/math/prim/prob/discrete_range_lpmf.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <stan/math/prim/fun/value_of.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/** \ingroup prob_dists
+ * Return the log PMF of a discrete range for the given y, lower and upper
+ * bound.
+ *
```
bob-carpenter: This should define the density, which is p(x | lower, upper) = 1 / (upper - lower + 1)` or on the log scale, `log p(x | lower, upper) = -log(upper - lower + 1)`.

stan/math/prim/prob/discrete_range_lpmf.hpp
```diff
@@ -0,0 +1,96 @@
...
+ * of the probabilities.
+ * @throw std::domain_error if upper is smaller than lower.
+ * @throw std::invalid_argument if non-scalar arguments are of different
+ * sizes.
+ */
+template <bool propto, typename T_y, typename T_lower, typename T_upper>
+double discrete_range_lpmf(const T_y& y, const T_lower& lower,
+                           const T_upper& upper) {
+  static const char* function = "discrete_range_lpmf";
+
```
bob-carpenter: [optional]
Vertical whitespace is precious!  I'd get rid of all these extra spaces to allow myself to view more of the function at once.

test/unit/math/prim/prob/discrete_range_test.cpp
bob-carpenter: The way to enforce that all arguments are integers or integer arrays, add a final template parameter:

```
typename = std::enable_if_t<std::is_integral<scalar_type_t<T1>>::value
                            && std::is_integral<scalar_type_t<T2>>::value
                            && std::is_integral<scalar_type_t<T3>>::value>
```

where `scalar_type_t` is a Stan meta function.

This isn't strictly necessary, but it should be clear in the doc that the use is intended to only be for integers.

stan/math/prim/prob/discrete_range_lpmf.hpp
mcol: I cut a couple of empty lines. I still prefer keeping logically separate block also separate visually.

mcol:
Thanks Bob! This should be closer to what you expected. :)

stan/math/prim/prob/discrete_range_lpmf.hpp
bob-carpenter: That's why I made it optional.  No promises I won't remove vertical space if I ever touch the files in the future :-)

test/unit/math/prim/prob/discrete_range_test.cpp
bob-carpenter: Just to be clear here, our `binomial_lpmf` and our `hypergeometric_lpmf` are broken in that they'll accept real-valued arguments.  I'm going to open a separate issue to try to deal with some of this as it'll be easy to call them erroneously as they're written with real values, but those don't make sense (they don't normalize properly).

test/prob/discrete_range/discrete_range_test.hpp
```diff
@@ -0,0 +1,62 @@
+// Arguments: Ints, Ints, Ints
+#include <stan/math/prim.hpp>
+
+using std::vector;
+
+class AgradDistributionsHypergeometric : public AgradDistributionTest {
```
bob-carpenter: The name should match what's being tested.  

We should also follow uniform_lpdf and throw if the parameters are not properly ordered.

I believe we should also be throwing a domain error when there's a domain error, that is when the variant isn't within the bounds.  That's not how uniform_lpdf or pareto_lpdf currently work, so I'm going to open another issue to fix those and this one.

I also believe we should be throwing when the variant is not in domain.  That's what we do for other distributions.  The `uniform_lpdf`, but it should be.  

Now that I see these tests, I realize that we should be throwing when there's a domain error, not just returnign -infinity.  For instance, consider `uniform_lpdf`, which throws if the parameters aren't properly ordered.  I'm going to open yet another issue to fix all this in the other distributions.

PR #1684, opened 2020-02-07
Explicitly call stan::math::size (fix ambiguous call versus std::size)

kedartal:
## Summary

I'm getting ambiguous call errors when compiling `stan/math/prim/err/check_consistent_size.hpp` and `stan/math/prim/err/check_consistent_sizes.hpp` using a very recent LLVM clang and libc++

    $ .../bin/clang++ --version
    clang version 11.0.0 (https://github.com/llvm/llvm-project.git 8acedb595d039f68ad15f9e5f2e6cb79729307e4)
    Target: x86_64-apple-darwin19.3.0
    Thread model: posix

Explicitly calling `stan::math::size` seems both the intended behavior as well as solving the issue.

## Tests

N/A

## Side Effects

Shouldn't be any.

## Checklist

I did not create an issue.

Tal Kedar

### DIFF
--- a/stan/math/prim/err/check_consistent_size.hpp
+++ b/stan/math/prim/err/check_consistent_size.hpp
@@ -24,7 +24,7 @@ template <typename T>
 inline void check_consistent_size(const char* function, const char* name,
                                   const T& x, size_t expected_size) {
   if (!is_vector<T>::value
-      || (is_vector<T>::value && expected_size == size(x))) {
+      || (is_vector<T>::value && expected_size == stan::math::size(x))) {
     return;
   }
 
@@ -36,7 +36,7 @@ inline void check_consistent_size(const char* function, const char* name,
       << "multidimensional values of the same shape.";
   std::string msg_str(msg.str());
 
-  invalid_argument(function, name, size(x),
+  invalid_argument(function, name, stan::math::size(x),
                    "has dimension = ", msg_str.c_str());
 }
 
--- a/stan/math/prim/err/check_consistent_sizes.hpp
+++ b/stan/math/prim/err/check_consistent_sizes.hpp
@@ -26,8 +26,8 @@ template <typename T1, typename T2>
 inline void check_consistent_sizes(const char* function, const char* name1,
                                    const T1& x1, const char* name2,
                                    const T2& x2) {
-  size_t max_size = std::max(is_vector<T1>::value * size(x1),
-                             is_vector<T2>::value * size(x2));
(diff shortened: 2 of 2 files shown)


kedartal:
Please let me know if you want me to create an issue, and/or if you prefer that the pull request be against the `develop` branch.

mcol:
Do you have this problem only on master or also the current develop? I don't know if we take patches for master or if this warrants a fix release, @bob-carpenter or @syclik may know better.

If you have the same problem on the current develop, then we'd take a fix. I've recently moved `size.hpp` on develop, and all's well with the set of compilers we use, but perhaps on a very recent clang something is different?

rok-cesnovar:
I think @ChrisChiasson was experiencing something similar. This ambiguity with size() seems to be an issue with some newer compilers. If this warrants another patch release is something we can discuss, but I think this PR needs to be done to develop first.

ChrisChiasson:
The problem is C++ 17 defines a function with the nearly same signature. The solution chosen in the complex thread on discourse was to directly define stan's size function on the types on which it is intended to operate (std::vector and Eigen types), rather than allow the catch-all template to activate if a non-stan container type is detected. The relevant code may be in @bob-carpenter 's feature/0123-complex-var branch.

kedartal:
@rok-cesnovar I moved the PR to be against the `develop` branch.
The fix is rather trivial for exactly the kind of people who are likely to encounter the issue, so not sure there's a lot of value in a release...

@ChrisChiasson Can you please share a link to the discussion?
Specializing `stan::math::size` to a new type (or concept) in response to a concrete compiler error should be very straightforward -- what am I missing? Is there a reason to expect people will try using such new types often, so we're trying to avoid needing the specialization?

ChrisChiasson:
https://github.com/stan-dev/math/commit/4857c338aebdd628efbbc64ca2c2d07bcd667af2

kedartal:
@ChrisChiasson Thanks! Having the compiler enforce a requirement that any type used in the PR code paths is actually supported by `stan::math::size` may be beneficial.

kedartal:
I'm not sure what failed exactly -- there's a `Cannot contact old-imac: java.lang.InterruptedException` on line 60 of https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/downstream_tests/1239/pipeline -- but looks like this is something in the CI environment, not an issue with the code, right?

bob-carpenter:
I added issue #1691 to deal with the remaining issues of `size()`---we need to do this everywhere.

PR #1688, opened 2020-02-09
Add positive definite checks to mdivide_left_spd

### LINKED ISSUE #1687: mdivide_left_spd doesn't check for positive definite argument in rev
## Description
`mdivide_left_spd` should throw if it's called on a non-positive definite matrix. This doesn't happen for the rev version as it's lacking some `check_pos_definite` calls.

## Example

The following currently doesn't throw but should:
```
Eigen::Matrix<var, -1, -1> A = Eigen::Matrix<var, -1, -1>::Zero(3, 3);
Eigen::VectorXd b(3);
b << 1, 2, 3;
mdivide_left_spd(A, b); // produces -nan, -nan, -nan
```

#### Current Version:
v3.1.0

mcol:
## Summary

This makes `mdivide_left_spd` throw if the argument is not a positive definite matrix. Fixes #1687.

## Tests
There were already tests commented out for this, so I only had to uncomment them.

## Side Effects

None.

## Checklist

- [X] Math issue #1687

### DIFF
--- a/stan/math/rev/fun/mdivide_left_spd.hpp
+++ b/stan/math/rev/fun/mdivide_left_spd.hpp
@@ -147,6 +147,7 @@ inline Eigen::Matrix<var, R1, C2> mdivide_left_spd(
 
   check_square("mdivide_left_spd", "A", A);
   check_multiplicable("mdivide_left_spd", "A", A, "b", b);
+  check_pos_definite("mdivide_left_spd", "A", A);
 
   // NOTE: this is not a memory leak, this vari is used in the
   // expression graph to evaluate the adjoint, but is not needed
@@ -167,6 +168,7 @@ inline Eigen::Matrix<var, R1, C2> mdivide_left_spd(
 
   check_square("mdivide_left_spd", "A", A);
   check_multiplicable("mdivide_left_spd", "A", A, "b", b);
+  check_pos_definite("mdivide_left_spd", "A", A);
 
   // NOTE: this is not a memory leak, this vari is used in the
   // expression graph to evaluate the adjoint, but is not needed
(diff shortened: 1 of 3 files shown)


mcol:
There are mix tests for `mdivide_right_spd` that can be turned on as well. I'll push a commit after this CI cycle has finished, so that if there are problems in the upstream tests I can fix those at the same time.

rok-cesnovar:
Lets say we dont add this check and a non-spd matrix is put on the input. What would happen? Would we get Eigen exceptions or would we see nans and/or zeros on the diagonal? 

What I am wondering here is can we let the function roll and check the result and then throw if we see things went south? The problem I am seeing here is that `check_pos_definite` basically does a cholesky decomposition, meaning that mdivide_left_spd would get a huge performance hit here.

mcol:
We already do the `check_pos_definite` in all other entry points to `mdivide_left_spd`, so this does not depart from what is expected. The documentation also expects the throwing behaviour: "If the matrix dividend argument is not symmetric and positive definite, these will reject and print warnings" (https://mc-stan.org/docs/2_22/functions-reference/linear-algebra-functions-and-solvers.html).

But you are making a good point about the expensiveness of the check, so perhaps it's worth considering removing all those checks. The Eigen documentation is not clear as to what happens if `llt` is called on a matrix that's not positive definite. In my tests, it seems to silently return the original matrix, so all subsequent computations will be wrong.

This [has been discussed on StackOverflow](https://stackoverflow.com/questions/35227131/eigen-check-if-matrix-is-positive-semi-definite), where they mention throwing if `llt.info() == Eigen::NumericalIssue`. So implementing this should allow us to keep the current expected behaviour, fix the inconsistency between prim and rev, and also get rid of the expensive check. I'll have a go at it, and if it works I'll repurpose this PR accordingly.

rok-cesnovar:
> We already do the `check_pos_definite` in all other entry points to `mdivide_left_spd`, so this does not depart from what is expected.

Oh, so this is used everywhere else in `mdivide_left_spd` apart from these cases you added? I am fine with adding the rest then to have a consistent solution and then revisit if what I am proposing would make sense. Sounds good to you?

mcol:
As I have a few things spinning at the moment, I'll merge as it is and open an issue regarding the option of testing the error code after the computation instead of checking for positive definiteness on entry.

Issue #1639, opened 2020-01-23
How do we do move parts of the autodiff stack to the GPU?


SteveBronder:
## Description

I was going to email @t4c1 and @rok-cesnovar seperately but I figured it would be better to discuss this in an open issue for clarity.

With the new compiler and a lot of @t4c1's work I think we can move whole chunks of the autodiff stack onto the GPU. So in the compiler when it sees a whole set of functions that can take in a `matrix_cl<var>` I think we could get it to generate something like

```cpp
// If a, b, and c are all large Eigen matrices of vars
auto a_cl = to_matrix_cl(a);
auto b_cl = to_matrix_cl(b);
auto c_cl = to_matrix_cl(c);
Eigen::Matrix<var, -1, -1> stuff = from_matrix_cl(add(a_cl, multiply(c_cl, d_cl)));
```

But I'm not totally sure what that would look like, maybe something like the below? We want a vari pointer to sit in the `matrix_cl<var>` and the autodiff stack so when the reverse pass is called the correct chain method is also called

## Example

```cpp

namespace internal {

// Adding two var matrices on gpu
class add_vv_vari_cl : public op_vv_vari_cl {
 public:
  add_vv_vari_cl(const matrix_cl<var>& avi, const matrix_cl<var>& bvi)
      : op_vv_vari_cl(avi.val() + bvi.val(), avi, bvi) {}
  void chain() {
      avi_.adj() += adj_;
      bvi_.adj() += adj_;
  }
};

matrix_cl<var> operator+(const matrix_cl<var>& a, const matrix_cl<var>& b) {
  // need a new vari constructor for matrix_cl<var>
  return matrix_cl<var>(new add_vv_vari_cl(a, b));
}
```

I think that makes sense? Figuring out which variables are candidates to be moved over to the GPU on the compiler side would probably be a bigger issue. It would be something like how we would check for a `static_matrix` like bob mentioned

#### Current Version:
v3.0.0

rok-cesnovar:
Thanks for bringing this up. 

I have been thinking about this a lot and I think how you drew it up is more or less how I envisioned it. I never got to writing the code out, but this example looks great. Adding var overloads and adding compiler support is the last phase to a fully OpenCL supported Stan. With Tadej's @t4c1 advances this seems like a great time to start.

Regarding when to move to the OpenCL device or not I think there are 2 stages:

[...]

PR #1690, opened 2020-02-09
[WIP] Refactor var ops to use one op_vari class instead of multiple classes

SteveBronder:
## Summary

Related to #1639 in that I'm trying to figure out how to use `matrix_cl<var>` types to push parts of the autodiff stack onto the gpu. It's not totally clear to me yet but I think the first step is making the vari operations more generic. 

For adding two vars, this PR uses a new class called `op_vari` that stores all of the objects needed for the reverse pass's `chain()` call into a tuple. This means we can remove all the specializations like `vv_vari`, `vd_vari`, `vvv_vari`, etc. The classes `add_vv_vari` and `add_vd_vari` have been replaced with a templated `add_vari` with specializations for the case of both inputs being `vari` and one input being a double.  This also adds an `is_vari` struct that checks if an input class is derived from a `vari`.

I think with the code here we can start thinking about how to call 

```cpp
add_vari<matrix_cl<var>, matrix_cl<var>>(a, b);
```

and what that should do. Though it's a separate issue than what this PR does, it is the intent

I'll review inline which pieces of the code I don't like or at least think can be improved.

This is just for adding with var types, then when we think this looks good enough we can add it for the rest of the operators.

## Tests

atm there are no new tests, but I will move the tests for `vv_vari` etc. into one test for `op_vari`

## Side Effects

Almost surely. The biggest things for devs are

1. Accessing the elements in the tuple for the reverse mode. We now have to access them either via `std::get<Ind>(this->vi())` or `this->template get<Ind>()` where `Ind` is a `size_t`. Tbh I don't enjoy either of how those look. Though I'm not sure if there are other alternatives.

2. Because the classes that hold the chain method for var operators are now templated we have to call `internal::add_vari<vari*, vari*>(a.vi_, b.vi_)}` instead of `internal::add_vv_vari(a.vi_, b.vi_)}`, etc.

## Checklist

- [x] Math issue #1639

### DIFF
--- a/stan/math/rev/core/operator_addition.hpp
+++ b/stan/math/rev/core/operator_addition.hpp
@@ -4,38 +4,55 @@
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/rev/core/var.hpp>
 #include <stan/math/rev/core/op_vari.hpp>
-#include <stan/math/rev/core/vd_vari.hpp>
 #include <stan/math/prim/fun/constants.hpp>
 #include <stan/math/prim/fun/is_any_nan.hpp>
+#include <type_traits>
 
 namespace stan {
 namespace math {
 
 namespace internal {
-template <typename Av, typename Bv>
-class add_vv_vari : public op_vari<Av, Bv> {
+
+template <typename T>
+struct is_vari : bool_constant<std::is_base_of<vari, std::remove_pointer_t<std::decay_t<T>>>::value> {};
+
+template <typename T>
+using require_vari_t = require_t<is_vari<T>>;
+
+template <typename... Types>
+using require_all_vari_t = require_all_t<is_vari<Types>...>;
+
+template <typename T1, typename T2, typename= void>
+class add_vari  {
+  static_assert(1, "If you see this please report a bug!");
+};
+
+template <typename T1, typename T2>
+class add_vari<T1, T2, require_all_vari_t<T1, T2>> : public op_vari<T1, T2> {
  public:
-  add_vv_vari(vari* avi, vari* bvi)
-      : op_vari<Av, Bv>(avi->val_ + bvi->val_, avi, bvi) {}
+  add_vari(T1 avi, T2 bvi)
+      : op_vari<T2, T1>(avi->val_ + bvi->val_, avi, bvi) {}
   void chain() {
-    if (unlikely(is_any_nan(this->template get<0>()->val_, this->template get<1>()->val_))) {
-      this->template get<0>()->adj_ = NOT_A_NUMBER;
-      this->template get<1>()->adj_ = NOT_A_NUMBER;
+    if (unlikely(is_any_nan(std::get<0>(this->vi())->val_, std::get<1>(this->vi())->val_))) {
+      std::get<0>(this->vi())->adj_ = NOT_A_NUMBER;
+      std::get<1>(this->vi())->adj_ = NOT_A_NUMBER;
     } else {
-      this->template get<0>()->adj_ += this->adj_;
-      this->template get<1>()->adj_ += this->adj_;
+      std::get<0>(this->vi())->adj_ += this->adj_;
+      std::get<1>(this->vi())->adj_ += this->adj_;
     }
   }
 };
 
-class add_vd_vari : public op_vd_vari {
+template <typename T1, typename T2>
+class add_vari<T1, T2,
+  require_t<conjunction<is_vari<T1>, std::is_floating_point<T2>>>> : public op_vari<T1, T2> {
  public:
-  add_vd_vari(vari* avi, double b) : op_vd_vari(avi->val_ + b, avi, b) {}
+  add_vari(T1 avi, T2 b) : op_vari<T1, T2>(avi->val_ + b, avi, b) {}
   void chain() {
-    if (unlikely(is_any_nan(avi_->val_, bd_))) {
-      avi_->adj_ = NOT_A_NUMBER;
+    if (unlikely(is_any_nan(std::get<0>(this->vi())->val_, std::get<1>(this->vi())))) {
+      std::get<0>(this->vi())->adj_ = NOT_A_NUMBER;
     } else {
-      avi_->adj_ += adj_;
+      std::get<0>(this->vi())->adj_ += this->adj_;
     }
   }
 };
@@ -80,7 +97,7 @@ class add_vd_vari : public op_vd_vari {
  * @return Variable result of adding two variables.
  */
 inline var operator+(var a, var b) {
-  return {new internal::add_vv_vari<vari*, vari*>(a.vi_, b.vi_)};
+  return {new internal::add_vari<decltype(a.vi_), decltype(b.vi_)>(a.vi_, b.vi_)};
 }
 
 /**
@@ -100,7 +117,7 @@ inline var operator+(var a, Arith b) {
   if (b == 0.0) {
     return a;
   }
-  return {new internal::add_vd_vari(a.vi_, b)};
+  return {new internal::add_vari<decltype(a.vi_), double>(a.vi_, b)};
 }
 
 /**
@@ -120,7 +137,7 @@ inline var operator+(Arith a, var b) {
   if (a == 0.0) {
     return b;
   }
-  return {new internal::add_vd_vari(b.vi_, a)};  // by symmetry
+  return {new internal::add_vari<decltype(b.vi_), double>(b.vi_, a)};  // by symmetry
 }
 
 }  // namespace math
--- a/stan/math/rev/core/operator_addition.hpp
+++ b/stan/math/rev/core/operator_addition.hpp
@@ -145,6 +145,7 @@ inline var operator+(Arith a, var b) {
                                                           a)};  // by symmetry
 }
 
+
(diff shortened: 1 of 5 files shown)


SteveBronder:
Something I'm unclear about is how OpenCL will react with this. Like if we have something that looks like

```cpp
inline matrix_cl<var> operator+(const matrix_cl<var>& a, const matrix_cl<var>& b) {
  return {new internal::add_vari<matrix_cl<var>*, matrix_cl<var>*>(a, b)};
}
```

I think that implies we need a `vari*` type in the `matrix_cl<var>`. Though I'm having a hard time figuring the memory implications of calling `grad()` with a matrix_cl<var>.

stan/math/rev/core/operator_plus_equal.hpp
```diff
@@ -9,7 +9,7 @@ namespace stan {
 namespace math {
 
 inline var& var::operator+=(var b) {
-  vi_ = new internal::add_vv_vari(vi_, b.vi_);
+  vi_ = new internal::add_vari<vari*, vari*>(vi_, b.vi_);
```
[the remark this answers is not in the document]
bob-carpenter: Yes, if they're always pointes, then the template parameter can just be the type without the pointer.

But when is the type not going to be `vari` for `add_vari`?

stan/math/rev/core/operator_addition.hpp
```diff
@@ -99,7 +121,7 @@ inline var operator+(var a, Arith b) {
   if (b == 0.0) {
     return a;
   }
-  return {new internal::add_vd_vari(a.vi_, b)};
+  return {new internal::add_vari<decltype(a.vi_), double>(a.vi_, b)};
```
bob-carpenter: Won't these types be inferred?

stan/math/rev/core/operator_addition.hpp
```diff
@@ -3,38 +3,59 @@
...
-      bvi_->adj_ = NOT_A_NUMBER;
+    if (unlikely(is_any_nan(std::get<0>(this->vi())->val_,
+                            std::get<1>(this->vi())->val_))) {
+      std::get<0>(this->vi())->adj_ = NOT_A_NUMBER;
+      std::get<1>(this->vi())->adj_ = NOT_A_NUMBER;
     } else {
-      avi_->adj_ += adj_;
-      bvi_->adj_ += adj_;
+      std::get<0>(this->vi())->adj_ += this->adj_;
+      std::get<1>(this->vi())->adj_ += this->adj_;
```
bob-carpenter: This all seems much more complicated than the code was before.  What's the advantage of this new scheme?  As far as I can tell, efficiency will be the same.

stan/math/rev/core/operator_plus_equal.hpp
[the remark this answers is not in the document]
bob-carpenter: Oh, I see the example on line 21.  This is going to store everything.  This seems like it could get very confusing for the code for the functions.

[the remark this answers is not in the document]
SteveBronder: For `add_vari<vari*, double>`. Since we want the double to be the actual type here and not a pointer it could be better to specify this at the call site. Otherwise in the code we'll have to have some wrapper around the signatures to make them a pointer or not for vari which I think would be weird

stan/math/rev/core/operator_addition.hpp
SteveBronder: What do you mean by inferred? For classes I think you always need to declare the template types. In C++17 you can just call the constructor and it will infer the type

stan/math/rev/core/operator_addition.hpp
SteveBronder: It's wayyy more complicated than before. There's no computational advantage here, the idea is that if we can get this to work for `matrix_cl<var>` types then we won't have to rewrite a lot of things and can push elementwise operations onto the gpu.

There's a real tradeoff here between duplication and complication. imo I'd rather have a new pattern that's more complicated but consistent over having to maintain duplicate code. The other way to do this would be to template vv_vari etc., but I'm not sure if I like that. `std::get<>` is certainly weird, but while the call site is weird the pattern itself is simple and removes a good few specialized classes

t4c1: This could be simplified with a pair of functions in `op_vari`:
```
template <std::size_t Ind>
auto& arg_val (){
    return std::get<Ind>(this->vi())->val_;
}


template <std::size_t Ind>
auto& arg_adj (){
    return std::get<Ind>(this->vi())->adj_;
}
```

Regardless it is hard to see the point of this PR. I know it is a lot to ask, but a prototype of how this would work with `matrix_cl<var>` would be super helpful.

t4c1: Or if you don't like writing `this->template arg_val<2>()` you can even hardcode options for lets say up to 3 arguments.

SteveBronder: > Regardless it is hard to see the point of this PR. I know it is a lot to ask, but a prototype of how this would work with matrix_cl<var> would be super helpful.

Yeah it's still very vague. My intent here was to try to write out the operator methods such that we could pass in a `matrix_cl<var>`. It's reasonable to want to see how we would use this before taking on any complication costs

What I'm thinking is that for 

```cpp
inline matrix_cl<var> operator+(const matrix_cl<var>& a, const matrix_cl<var>& b) {
  return {new internal::add_vari<matrix_cl<var>*, matrix_cl<var>*>(a, b)};
}
```

Say the stan compiler could generate the silly/trivial code below

```cpp
Eigen::Matrix<var, -1, -1> x = from_matrix_cl(to_matrix_cl(A) + to_matrix_cl(B))
```

1. `to_matrix_cl(A/B)` gives back a `matrix_cl<var>` that then calls the above `operator+`. 
2. `operator+` would add that `add_vari` to the returned `matrix_cl<var>` stored as `vari*`. When we do the reverse mode that `add_vari` exists on the autodiff stack. 
4. When `grad()` is called and starts executing `.chain()` methods on the autodiff stack...
5. (???)

For (5) my main Q is where those `matrix_cl<var>`s we point to stored so RAII doesn't clean up a matrix_cl<var> before we call chain

  - One thing we could do is store the `matrix_cl<var>`'s memory in the context (kind of abusing the context but we made it a singleton and it's there)
  - Is there something smarter we can do since we know that an `Eigen::Matrix<var,-1,-1>`'s data is stored on our stack? The `val_` part is `const` so we'll have to copy that, but we could take ownership of the `adj_`
  - And do we want to try to be lazy about data movement here so that the GPU doesn't crap out from lack of dram?

bob-carpenter:
> Or if you don't like writing this->template arg_val<2>() you can even hardcode options for lets say up to 3 arguments.

If we could encapsulate this in nicer syntax, that'd make me happier.  I'm still worried about not having variable names and just having numbers, but shorter syntax would be a good start.

stan/math/rev/core/operator_addition.hpp
t4c1: > One thing we could do is store the matrix_cl<var>'s memory in the context (kind of abusing the context but we made it a singleton and it's there)

I think having a separate chain stack for `matrix_cl<var>`s would be preferable.

> The val_ part is const so we'll have to copy that, but we could take ownership of the adj_

I don't follow that part.

> And do we want to try to be lazy about data movement here so that the GPU doesn't crap out from lack of dram?

That is probably a good idea. It might help to wait until reverse sweep before adjoint matrix_cl is allocated.

SteveBronder: >> The val_ part is const so we'll have to copy that, but we could take ownership of the adj_

> I don't follow that part.

If we wanted to not have an OpenCL stack and use stan's stack I think we could do some trick with [enqueueMapBuffer](https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clEnqueueMapBuffer.html) and enqueueUnMapBuffer. Though I'm not sure if that's a thing we can do

t4c1: We could do it, but I think OpenCL stack is the way to go. This mapping seems to be just a conveniance function that hides host<->device transfers.

SteveBronder: For GPUs it's just a wrapper effectively. But for cpus the map buffer can take ownership of the memory so we avoid a copy. I think using them would let us use the OpenCL stuff for smaller problems while still getting parallelization. Though what we really want is shared virtual memory. It looks like Nvidia has started supporting that (from windows [release notes](http://us.download.nvidia.com/Windows/442.19/442.19-win10-win8-win7-release-notes.pdf))

https://software.intel.com/en-us/articles/getting-the-most-from-opencl-12-how-to-increase-performance-by-minimizing-buffer-copies-on-intel-processor-graphics

t4c1: > I think using them would let us use the OpenCL stuff for smaller problems while still getting parallelization

Running one chain per core is far easier. I think it is also likely to be more efficient.

> Though what we really want is shared virtual memory

That just means same address space. Unless you are thinking about CPUs or integrated GPUs copying still needs to happen.

Issue #1689, opened 2020-02-09
remove size zero checks in `check_multiplicable`


mitzimorris:
## Description

Currently `check_multiplicable(A, B)` checks that both `A` and `B` are non-empty.  This needs to be changed so that either or both can be empty.  Fixing `check_multiplicable` will cause other functions to break rather than throw exceptions for size zero inputs when Eigen hits an assert error.   Each function that uses `check_multiplicable` needs to be checked as to whether it also requires one or both of `A` and `B` to be empty.  

The `multiply` function itself was updated to remove the `check_multiplicable` tests.  These should be returned after `check_multiplicable` is fixed.

A new function `check_nonempty()` should be added to deal with checking for emptiness or size checks can be done on the number of rows or columns. 

 Documentation should be updated in both the C++ and the Stan function reference to reflect the changed boundary conditions.

More generally, all of Stan's matrix functions should be audited for boundary behavior.  Nothing should be able to get through to an Eigen assert.  The entire library does not need to be audited for this issue.

This is a generalization of issue  #1669.

#### Examples

[`matrix_exp_multiply`](https://github.com/stan-dev/math/blob/100c9850a0e4da99ebb9a497ce5902c7340de0a3/stan/math/prim/fun/matrix_exp_multiply.hpp#L20-L27) explicitly checks whether matrix B is 0x0, then calls `check_multiplicable`.

`A * B` currently requires `A` and `B` to be non-empty, but this expression is well-defined for empty matrices.  For example, if `A` is 5 x 0 and `B` is 0 x 4, then `A * B` is 5 x 4 (and contains all zero entries, as that's the boundary case for a summation, which is how entries are defined).

`A / B = A * inverse(B)` and `B \ A = inverse(B) * A` should be able to handle empty matrices in either position, because the inverse of a 0 x 0 matrix is just a 0 x 0 matrix (it satisifes the relevant identity laws).

#### Current Version:
v3.1.0

dpsimpson:
If this is something that isn't particularly urgent, I can do it in my spare time over the next week or two.

bob-carpenter:
We have a quarterly release schedule, so there's a couple months and change to get this in before the next release :-)  I was just saying we need someone like Dan Simpson or Ben Goodrich to figure out the boundary conditions of all these matrix ops.

dpsimpson:
Tbh I’ve been looking for some busywork so this is perfect for me.

mcol:
Some notes regarding the issue description:
[...]

PR #1701, opened 2020-02-11
Don't throw when computing the inverse of a size zero matrix

### LINKED ISSUE #1432: inverse_spd should have consistent errors on size 0
## Description

This test should pass in `test/unit/math/mix/mat/fun/inverse_spd_test.cpp`:

```
Eigen::MatrixXd m00(0, 0);
stan::test::expect_ad(f, m00);
```

Edit:  I previous indicated this was an error, but I don't see why the unique 0 x 0 matrix isn't considered the inverse of the 0 x 0 matrix.  It satisfies the identity laws.  Also, I'd think that the determinant of an empty matrix would be one as it's a product, and the empty product is one.

#### Current Version:
v3.0.0

### LINKED ISSUE #1463: inverse should have consistent errors on size 0 inputs

mcol:
## Summary
This reverts the earlier decision to throw when computing the inverse of a size zero matrix, as discussed in #1689. We need to return an empty matrix ourselves because Eigen asserts when computing the inverse of an empty matrix. Fixes #1463. Fixes #1432.

## Tests

Updated accordingly. There were also a couple of errors in those tests concerning the type of exception thrown.

## Side Effects

None.

## Checklist

- [X] Math issue #1463, #1432

### DIFF
--- a/stan/math/rev/fun/inverse.hpp
+++ b/stan/math/rev/fun/inverse.hpp
@@ -70,12 +70,17 @@ class inverse_vari : public vari {
 /**
  * Reverse mode specialization of calculating the inverse of the matrix.
  *
- * @param m Specified matrix.
- * @return Inverse of the matrix.
+ * @param m specified matrix
+ * @return Inverse of the matrix (an empty matrix if the specified matrix has
+ * size zero).
+ * @throw std::invalid_argument if the matrix is not square.
  */
 inline matrix_v inverse(const matrix_v &m) {
   check_square("inverse", "m", m);
-  check_nonempty("inverse", "m", m);
+  if (m.size() == 0) {
+    return matrix_v(0, 0);
+  }
+
   matrix_v res(m.rows(), m.cols());
   internal::inverse_vari *baseVari = new internal::inverse_vari(m);
   res.vi() = Eigen::Map<matrix_vi>(baseVari->vari_ref_A_inv_, res.rows(),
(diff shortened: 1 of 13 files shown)


stan/math/rev/fun/inverse.hpp
```diff
@@ -70,12 +70,17 @@ class inverse_vari : public vari {
...
+ * @param m specified matrix
+ * @return Inverse of the matrix (an empty matrix if the specified matrix has
+ * size zero).
+ * @throw std::invalid_argument if the matrix is not square.
  */
 inline matrix_v inverse(const matrix_v &m) {
   check_square("inverse", "m", m);
-  check_nonempty("inverse", "m", m);
+  if (m.size() == 0) {
+    return matrix_v(0, 0);
```
bob-carpenter: [optional]
This could be `return {0, 0};` to match the other definitions.

bob-carpenter:
I approved because I don't think it's worth changing return style, but I'll let @mcol decide.   Feel free to merge as is.

mcol:
Since I didn't think it was worth just that one case, I corrected that pattern everywhere.

Issue #1692, opened 2020-02-10
Investigate which int overloads are still needed (exp, pow, ...)


rok-cesnovar:
## Description

My lab mate asked me about the reasoning behind having `exp(int)` defined in the Math library. 

Example: `inline double exp(int x) { return std::exp(x); }` in https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/exp.hpp#L18

I could not find a reason so I am wondering myself also. Can anyone help me understand the reasoning behind this.

This came up while trying to add a custom function in stan math and using something like `exp(3.3)` inside the stan::math namespace was being truncated down to `exp(3)`. This was due to a missing `using std::exp;` that would have been picked if he used cpplint.

#### Current Version:
v3.1.0

bob-carpenter:
That problem hadn't occurred to me!

The reason we have the `int` signature is to avoid ambiguity with `exp(int)` when `std::exp()` and `stan::math::exp()` are in scope.  None of them have integer signatures, so an `int` either needs to be promoted to `double` or to `var` or `fvar`. 

We could get around this by having our math lib version be defined as something like:

```
template <typename T, typename = std::enable_if_t<std::is_arithmetic<scalar_type_t<T>>::value>>
inline T exp(T x) {
  return std::exp(x);
}
```

Then it won't try to cast `double` to `int`, while still being a better match for `exp(int)` than `std::exp()`.

bob-carpenter:
OK, I just went and looked up the [signatures of `exp`](https://en.cppreference.com/w/cpp/numeric/math/exp) and as of C++11, there's an `exp(IntegralType)`.  That must not cause ambiguities with our `exp(int)`, but it probably means we can get rid of our `exp(int)` definition without problems.

rok-cesnovar:
> while still being a better match for `exp(int)` than `std::exp()`

But stan::math::exp still uses std::exp so these would be the same in the end right? So I guess this was to prevent int being promoted to var?

bob-carpenter:
Right---the exp(int) signature was just to disambiguate calls to exp(int) in Stan, which were otherwise ambiguous betwene promotion to double and promotion to var.

rok-cesnovar:
This was fixed in #1711

PR #1702, opened 2020-02-12
Cleanup unnecessary prim/fun signatures and checks (redundant as of C++11)

### LINKED ISSUE #1704: Scalar lgamma_stirling and lbeta truncate log input to integer
## Description

https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/lbeta.hpp does not include `<cmath>` and does not declare `using std::log` anywhere. Which probably means that that would use stan::math::log. But, given that only declare stan::math::log(int) that means that the input gets truncated to integer. Luckily this doesnt effect vectorized versions and eigen versions.

We have the same issue in https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/lgamma_stirling.hpp where there is a missing using statement.

This came up when trying to fix https://github.com/stan-dev/math/issues/1692

cc: @martinmodrak  I think you were having issue with these two functions in the past if I remember correctly, but cant find the thread/PR right now. Hopefully the fix of this wil help in any way.

Will fix in #1702 

#### Current Version:
v3.1.0

rok-cesnovar:
## Summary

C++11 added overloads for integers for some of the base functions. Previously Stan Math had its own overloads for these to avoid ambiguity with (var) overloads as it wasnt clear whether to promote to double or var. 

This PR removes the following:
    - asinh(double) and asinh(int)
    - cbrt(double), cbrt(int)
    - erf(double), erf(int)
    - erfc(double), erf(int)
    - exp(int)
    - expm1(int)
    - log(int)
    - sqrt(int)
    - round(double), round(int)
    - trunc(double), trunc(int)

The following integer overloads will stay in: 
- acosh, atanh, log1p (I cleaned up these a bit because I dont think is_nan on integer values is needed). The overloads have to stay because Stan Math throws for some input values, std:: returns nan and std:: throws for value we want to throw ourselves.
- lgamma (no changes here, this will stay because std::lgamma is not thread-safe)

Due to removing some signatures I had to fix the functions to now use std for double/int inputs. And that then led to a minor code cleanup in gp_matern* files.

This now also fixes #1704 by including `cmath` and adding using `std::log` statements in lgamma_stirling and lbeta.

## Tests

No new tests, existing tests cover all the cases.

## Side Effects

Hopefully none.

## Checklist

- [x] Math issue #1692 

### DIFF
--- a/stan/math/prim/fun/acosh.hpp
+++ b/stan/math/prim/fun/acosh.hpp
@@ -19,16 +19,8 @@ namespace math {
  * @throw std::domain_error If argument is less than 1.
  */
 inline double acosh(double x) {
-  if (is_nan(x)) {
-    return x;
-  } else {
-    check_greater_or_equal("acosh", "x", x, 1.0);
-#ifdef _WIN32
-    if (is_inf(x))
-      return x;
-#endif
-    return std::acosh(x);
-  }
+  check_greater_or_equal("acosh", "x", x, 1.0);
+  return std::acosh(x);
 }
 
 /**
@@ -39,16 +31,8 @@ inline double acosh(double x) {
  * @throw std::domain_error If argument is less than 1.
  */
 inline double acosh(int x) {
-  if (is_nan(x)) {
-    return x;
-  } else {
-    check_greater_or_equal("acosh", "x", x, 1);
-#ifdef _WIN32
-    if (is_inf(x))
-      return x;
-#endif
-    return std::acosh(x);
-  }
+  check_greater_or_equal("acosh", "x", x, 1);
+  return std::acosh(x);
 }
 
 /**
--- a/stan/math/prim/fun/acosh.hpp
+++ b/stan/math/prim/fun/acosh.hpp
@@ -19,11 +19,11 @@ namespace math {
  * @throw std::domain_error If argument is less than 1.
  */
 inline double acosh(double x) {
-  if (is_nan(x)) {	  
+  if (is_nan(x)) {
     return x;
-  } else {	
-    check_greater_or_equal("acosh", "x", x, 1.0);	
-    return std::acosh(x);	
+  } else {
+    check_greater_or_equal("acosh", "x", x, 1.0);
+    return std::acosh(x);
   }
 }
 
--- a/stan/math/prim/fun/acosh.hpp
+++ b/stan/math/prim/fun/acosh.hpp
@@ -19,8 +19,12 @@ namespace math {
  * @throw std::domain_error If argument is less than 1.
  */
 inline double acosh(double x) {
-  check_greater_or_equal("acosh", "x", x, 1.0);
-  return std::acosh(x);
+  if (is_nan(x)) {	  
+    return x;
+  } else {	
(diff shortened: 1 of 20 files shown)


stan/math/prim/fun/gp_matern32_cov.hpp
```diff
@@ -66,8 +65,8 @@ gp_matern32_cov(const std::vector<T_x> &x, const T_s &sigma,
   check_positive_finite(function, "length scale", length_scale);
 
   T_s sigma_sq = square(sigma);
-  T_l root_3_inv_l = sqrt(3.0) / length_scale;
-  T_l neg_root_3_inv_l = -1.0 * sqrt(3.0) / length_scale;
+  T_l root_3_inv_l = std::sqrt(3.0) / length_scale;
+  T_l neg_root_3_inv_l = -1.0 * std::sqrt(3.0) / length_scale;
```
mcol: This should just be `T_l neg_root_3_inv_l = -root_3_inv_l`, and similarly for the other places in this and other `gp_matern` files.

rok-cesnovar: Oh yeah, good call. Didnt look at this hard enough to see that. Was just trying to fix the issue at hand.

stan/math/prim/fun/acosh.hpp
```diff
@@ -23,10 +23,6 @@ inline double acosh(double x) {
     return x;
   } else {
     check_greater_or_equal("acosh", "x", x, 1.0);
-#ifdef _WIN32
-    if (is_inf(x))
-      return x;
-#endif
```
rok-cesnovar: This worked fine on my Windows box, but will let Jenkins give the final verdict. Will just revert back if Jenkins flags this.

stan/math/prim/fun/acosh.hpp
```diff
@@ -39,16 +35,8 @@ inline double acosh(double x) {
...
-    return x;
-  } else {
-    check_greater_or_equal("acosh", "x", x, 1);
-#ifdef _WIN32
-    if (is_inf(x))
-      return x;
-#endif
-    return std::acosh(x);
-  }
+  check_greater_or_equal("acosh", "x", x, 1);
```
rok-cesnovar: I think is_nan for integer inputs is redundant, isnt it?

stan/math/prim/fun/gp_matern32_cov.hpp
rok-cesnovar: Fixed in https://github.com/stan-dev/math/pull/1702/commits/652bed1fc4032658bed2271095d34b22b2eb041d, the neg_ version is only used in one place so I ommited the extra local variable.

rok-cesnovar:
Closing this for now as there are some additional question that need to be solved.

stan/math/prim/fun/acosh.hpp
mcol: Is this what made the test fail on Jenkins?

rok-cesnovar: Yep, will remove once I update the PR. Was hoping it could go away :)

PR #1705, opened 2020-02-12
Check for positive definiteness only after computing the LLT factors

### LINKED ISSUE #1693: Avoid check for positive definiteness in mdivide_left_spd
## Description
Currently we always check for positive definiteness of the matrix argument when `mdivide_left_spd` is called. These happen twice for `mdivide_right_spd`, as it checks it then delegates to `mdivide_left_spd` which checks it again.

Discussion in #1688 led us to consider whether this check could be skipped, as it computes a Cholesky factorization, which can be expensive. The Eigen documentation is not clear as to what happens if their `.llt()` method is called on a matrix that's not positive definite. In my tests, it seems to silently return the original matrix, so all subsequent computations will be wrong if we don't throw.

A thread on StackOverflow (https://stackoverflow.com/questions/35227131/eigen-check-if-matrix-is-positive-semi-definite) mentions that one could check `llt.info() == Eigen::NumericalIssue` after the LLT factorization is performed and throw if that's the case. So implementing this should allow us to keep the current expected behaviour of throwing if the matrix is not positive definite and also get rid of the expensive check.

## Expected Output
The functions should still be throwing if the matrix argument is not positive definite, but this will happen after the LLT computation rather than on entry to the function. This should also improve performance of these functions as there will be fewer wasted computations.

#### Current Version:
v3.1.0

mcol:
## Summary
This uses the `check_pos_definite` overload to check an LLT decompositions after it's computed, rather than using the generic `check_pos_definite` on the original matrix, as that would compute and then discard an LDLT decomposition of the matrix. Fixes #1693.

Since this other overload doesn't first check for NaNs or symmetry, I added those where necessary. Note that it's not necessary to check explicitly for a square matrix if we also check for symmetry, because the latter implies the former, that's why I removed some `check_square` in a couple of places.

I had a very brief look at the performance impact with this code:
```
TEST(MathMatrixPrim, mdivide_spd) {
  const int size = 100; // 200
  stan::math::vector_d d = Eigen::VectorXd::Ones(size) * 20;
  stan::math::matrix_d diag = d.asDiagonal();
  stan::math::matrix_d R = Eigen::MatrixXd::Random(size, size);
  stan::math::matrix_d A = 0.5 * (R + R.transpose()) + diag;

  for (int i = 0; i < 100; i++) {
    stan::math::matrix_d m1 = stan::math::mdivide_left_spd(A, A);
    stan::math::matrix_d m2 = stan::math::mdivide_right_spd(A, A);
  }
}
```
size 100: from 8706ms to 7011ms
size 200: from 49912ms to 40712ms

## Tests

No new tests.

## Side Effects

None.

## Checklist

- [X] Math issue #1693

### DIFF
--- a/stan/math/prim/err/check_pos_definite.hpp
+++ b/stan/math/prim/err/check_pos_definite.hpp
@@ -15,14 +15,19 @@ namespace math {
 
 /**
  * Check if the specified square, symmetric matrix is positive definite.
- * @tparam T_y Type of scalar of the matrix
- * @param function Function name (for error messages)
- * @param name Variable name (for error messages)
- * @param y Matrix to test
- * @throw <code>std::invalid_argument</code> if the matrix is not square
+ *
+ * This computes an LDLT decomposition to establish positive definiteness,
+ * so it can be expensive. If an LDLT or LLT decomposition is available,
+ * that should be tested instead.
+ *
+ * @tparam T_y type of elements in the matrix
+ * @param function function name (for error messages)
+ * @param name variable name (for error messages)
+ * @param y matrix to test
+ * @throw std::invalid_argument if the matrix is not square
  * or if the matrix has 0 size.
- * @throw <code>std::domain_error</code> if the matrix is not symmetric,
- * if it is not positive definite, or if any element is <code>NaN</code>
+ * @throw std::domain_error if the matrix is not symmetric,
+ * if it is not positive definite, or if any element is NaN
  */
 template <typename T_y>
 inline void check_pos_definite(const char* function, const char* name,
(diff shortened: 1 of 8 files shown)


stan/math/prim/err/check_pos_definite.hpp
```diff
@@ -15,14 +15,19 @@ namespace math {
 
 /**
  * Check if the specified square, symmetric matrix is positive definite.
- * @tparam T_y Type of scalar of the matrix
- * @param function Function name (for error messages)
- * @param name Variable name (for error messages)
- * @param y Matrix to test
- * @throw <code>std::invalid_argument</code> if the matrix is not square
+ *
+ * This computes an LDLT decomposition to establish positive definiteness,
```
bob-carpenter: [comment]
Thanks for improving the doc with useful dev-facing comments.

stan/math/prim/fun/mdivide_left_spd.hpp
```diff
@@ -30,14 +30,15 @@ namespace math {
...
-              b));
+  static const char *function = "mdivide_left_spd";
+  check_multiplicable(function, "A", A, "b", b);
+  check_positive(function, "rows", A.rows());
+  check_symmetric(function, "A", A);
+  check_not_nan(function, "A", A);
+
+  auto llt = Eigen::Matrix<return_type_t<T1, T2>, R1, C1>(A).llt();
+  check_pos_definite(function, "A", llt);
+  return llt.solve(Eigen::Matrix<return_type_t<T1, T2>, R2, C2>(b));
```
bob-carpenter: [comment]
As an added bonus the new code's also readable!

Issue #1703, opened 2020-02-12
Use RAII for nested autodiff


martinmodrak:
## Description
Use the RAII idiom to make sure that nested autodiff is always properly deinitialized. More thorough discussion at 
https://discourse.mc-stan.org/t/proposal-raii-for-nested-autodiff/12861

## Example

A code using autodiff would then look like:
```
{
    local_nested_autodiff nested();
    //Do stuff
    nested.set_zero_all_adjoints();
    //Do more stuff
}
```

With the destructor of `local_nested_autodiff` stopping the nested autodiff.

#### Current Version:
v3.1.0

wds15:
Do we really want to make `set_zero_all_adjoints` a member of this local nested thing?

I am not opposed to it... this is just a radical change.

Or asked differently: Are u also suggesting to **remove** the global `recover_memory_nested()` and all the other *global* nested functions?

Going in the direction of representing the entire AD tape as an instance is a good thing from my perspective. It's just that we need to be careful with this given that right now things are thread local. So passing any instance around between threads is a bad idea right now...

martinmodrak:
> Are u also suggesting to remove the global recover_memory_nested() and all the other global nested functions?

No, that would probably constitute a breaking API change. I would however suggest the RAII as "recommended". Also right now my implementation of `local_nested_autodiff` just calls the underlying functions.

> Do we really want to make set_zero_all_adjoints a member of this local nested thing?

Not 100% sure about this, it just felt natural to allow this use so that calling global functions and using the RAII class does not mix. To me this is a minor aesthetic choice on which I do not insist in any way.

PR #1706, opened 2020-02-12
Using local_nested_autodiff for all instances of nested autodiff

martinmodrak:
## Summary

Implemented a RAII-style class `local_nested_autodiff` (are we happy with the name?) that makes it easier to ensure that nested stack is correctly recovered on leaving of the scope. The class also has a `set_zero_all_adjoints` member, so that one can handle everything around nested autodiff via this class and without using global function calls.

Removed all uses of `start_nested`, `recover_memory_nested` and `set_zero_all_adjoints_nested` and use `local_nested_autodiff` instead. The only exception are some of the tests of the autodiff where those functions are tested separately.

The class itself is just a thin wrapper around the original calls.

## Tests

A new test was written that replicates the tests for nested autodiff using global functions with using `local_nested_autodiff`.

I also moved the `gradable` struct used in `rev/core/var_test.cpp` to its own header so that I can reuse it.

## Side Effects

Hopefully no. However I am worried that some of the functions I touched appear to have no tests.

## Checklist

- [x] Math issue #1703 


wds15:
I am not sure if I would put the nested thing in the most inner loops. This creates many additional start_nested calls which we did avoid before.

What is the rationale for that?

martinmodrak:
> I am not sure if I would put the nested thing in the most inner loops. This creates many additional start_nested calls which we did avoid before.

I definitely didn't want to change the location of any nested autodiff and aimed to put the constructor at exactly the line the `start_nested` was before. I briefly rechecked and it seems to me that all instances match the previous code. Could you point me the specific case where you see a problem?

wds15:
Argh... I misread. Matching the old order of calls is the right thing to do. Sorry for the noise. My bad.

bob-carpenter:
@wds15 --- can you put in the formal review?  It sounds like you did the audit for location, which is the most important part.  I'm going to drop some comments here rather than in a review.

I'm OK with the name as is.  If it's not too much trouble, I'd suggest somethihg a bit more descriptive like `begin_nested_rev_autodff` to indicate it's starting the scope.  

The usage can be just this:

```
begin_nested_rev_autodiff();
```

instead of

```
begin_nested_rev_autodiff nested;
```

Doesn't the latter give you lint warnings for unused variables?

The RAII class itself is very elegantly coded and documented.  It's making me so happy that devs are figuring out these kinds of patterns that make the code both easier to use and more robust.  

The only thing I would've required in a review is doc for the constructor and destructor, because that's where the action is.

wds15:
Ok, I can dive into this.

bob-carpenter:
I checked everything other than that the new calls are in the right place.  If you don't have time, I can go through and do that, too.

wds15:
So there is one thing which does concern me and I think we can solve it in a nice way.

The problem is with threads. The issue is that the instance of the local nested thing can get created in thread A, but destructed in thread B... and then we are in no-mans land!

However, there is a clean way to handle this. We can *define*  that instantiating the local nested thing in thread A will *always* tear down the nested context in thread A - no matter which thread is running the destructor.

Does that make sense to everyone? I am happy to code this up (either directly on this branch or on an extra branch based on this one).

I think we should include this in this PR to make it thread safe.

martinmodrak:
@wds15 I lack deep knowledge of threads, so I'd defer to you on this. But I am honestly intrigued - how do you imagine this being created and destroyed on different threads - can a scope begin on a thread and end on another thread? Note that I purposefully deleted the copy constructor, assignment operator and the `new` operator to avoid weirdness, but I certainly won't be completely shocked if that is not enough.

@bob-carpenter  
> I'm not sure how unassigned classes are handled for scoping---they might be destroyed right away since it's not going to be used again. Doesn't the latter give you unused variable warnings?

In all examples of using RAII, I've seen people declare the vars, so I would stick to this even if it is not strictly necessary (e.g. https://en.cppreference.com/w/cpp/language/raii). Though I think temporaries go out of scope sooner (https://rules.sonarsource.com/cpp/RSPEC-5184). Also I like having `nested.set_zero_all_adjoints();` but that's a minor point.

>Doesn't the latter give you unused variable warnings?

It does not - the compiler detects the constructor has side effects and realizes this might be the desired behavior.

Regarding the name - it is easy to change it. I slightly don't light having `begin` in the name - once again most RAII code I've seen doesn't do this and the classes are just called `lock`, not `begin_lock` - also there is no `end_nested_autodiff` to match which might be confusing. I like having the explicit `rev`, so I would currently vote for `nested_rev_autodiff` or `local_nested_rev_autodiff` (to emphasize it is scope dependent). No strong opinions on this and will happily comply with your view on this.

wds15:
Example:

```c++
// in global scope:

local_nested* some_nested_ptr = 0;

// somewhere in the program start thread A
// thread A does this in its own running context:
some_nested_ptr = new local_nested();

// another thread B then does
delete some_nested_ptr;
```

And there you have the issue.

I am happy to code up something up which will avoid this. Essentially you have to store upon construction of the object a pointer to the actual AD tape which we want to refer to - and store it in the instance. That would be my suggestion.

martinmodrak:
@wds15 The code you proposed will not compile, because I marked the `new` operator as deleted - precisely to avoid such issues.

bob-carpenter:
> On Feb 14, 2020, at 6:02 AM, Martin Modrák <notifications@github.com> wrote:
> 
> @wds15 I lack deep knowledge of threads,

Very few people do.  Nobody on our team does.


> so I'd defer to you on this. But I am honestly intrigued - how do you imagine this being created and destroyed on different threads -

You're right that this can't happen because RAII is controlled by scope and scope is a per-thread thing.

> can a scope begin on a thread and end on another thread? Note that I purposefully deleted the copy constructor, assignment operator and the new operator to avoid weirdness, but I certainly won't be completely shocked if that is not enough.

That should be enough to avoid unintended shenanigans.  But it's really more a programming idiom thing we have to introduce.  Given that it's the wild west of C underneath, you can do things like take a reference using & then pass it around as a pointer.  So you could maliciously try to run the destructor in a different chain if you could somehow get ahold of a reference.

> 
> @bob-carpenter
> 
> I'm not sure how unassigned classes are handled for scoping---they might be destroyed right away since it's not going to be used again. Doesn't the latter give you unused variable warnings?
> 
> In all examples of using RAII, I've seen people declare the vars, so I would stick to this even if it is not strictly necessary

OK.  Good to know.

> (e.g. https://en.cppreference.com/w/cpp/language/raii). Though I think temporaries go out of scope sooner (https://rules.sonarsource.com/cpp/RSPEC-5184). Also I like having nested.set_zero_all_adjoints(); but that's a minor point.

That's a relatively expensive operation that we shouldn't be doing if we don't have to.

> 
> Doesn't the latter give you unused variable warnings?
> 
> It does not - the compiler detects the constructor has side effects and realizes this might be the desired behavior.

Cool!  The C++ compilers never cease to amaze me.

> 
> Regarding the name - it is easy to change it. I slightly don't light having begin in the name - once again most RAII code I've seen doesn't do this and the classes are just called lock, not begin_lock -

Great.  I'm 100% behind following standard idioms for naming.

> also there is no end_nested_autodiff to match which might be confusing. I like having the explicit rev, so I would currently vote for nested_rev_autodiff or local_nested_rev_autodiff (to emphasize it is scope dependent). No strong opinions on this and will happily comply with your view on this.

I agree---the "local" bit is implied by scope anyway and I prefer shorter to longer identifiers.

bob-carpenter:
Will the following compile?

THREAD 0:
local_nested* ptr;  // not thread local

THREAD 1:
local_nested a;
ptr = &a;

THREAD 0:
ptr -> ~local_nested();

Even if it does, I'm not convinced it's something we need to try to programatically guard against as it would pretty much have to be done maliciously as I can't imagine anyone doing this kind of thing on purpose even by accident.

martinmodrak:
Name changed to `nested_rev_autodiff`.

> Will the following compile?

I guess that it will, but I agree that we shouldn't worry about this - we want to prevent inadvertent errors. Anyone wanting to do more complex nesting logic would use the global functions directly anyway.

bob-carpenter:
That's also my take on the situation.  There's only so far we need to go in defense vs. other developers.

martinmodrak:
@wds15 just pinging this, I'd be glad to get this merged, it is starting to see conflicts with develop. But if there is anything you have still doubts about, I'll be happy to work through it.

stan/math/rev/functor/idas_forward_system.hpp
```diff
@@ -121,56 +121,48 @@ class idas_forward_system : public idas_system<F, Tyy, Typ, Tpar> {
...
+      stan::math::jacobian(fyp, vec_yp, f_val, J);
+      r += J * yps_mat;
+
+      if (dae->is_var_par) {
+        auto fpar
+            = [&t, &vyy, &vyp, &N, &M, &dae](const matrix_v& x) -> vector_v {
+          std::vector<var> par(x.data(), x.data() + M);
           auto eval
-              = dae->f_(t, vyy, yp, vtheta, dae->x_r_, dae->x_i_, dae->msgs_);
+              = dae->f_(t, vyy, vyp, par, dae->x_r_, dae->x_i_, dae->msgs_);
```
bob-carpenter: [question]
Why did this call change when this PR is just about nested autodiff?

wds15:
@bob-carpenter can u please finish the review? Thanks. If you want me to look at something specific, I can do that.

stan/math/rev/functor/idas_forward_system.hpp
martinmodrak: It didn't change - note that the diff is misaligned and line 145 on the left should actually be paired with line 143 on the right. Possibly the reformatting by clangformat misled the diff algorithm. If you view the diff ignoring whitespace (can be set via the cog icon at the top), it is visible that the code didn't change.

bob-carpenter:
I reviewed all the scopes and the class and submitted the review.  The "change request" was just for an answer to my question of why one of the function's arguments changed when this was just about scoping memory.

Was there an answer somewhere?

rok-cesnovar:
The answer is here https://github.com/stan-dev/math/pull/1706#discussion_r382938707

wds15:
Thanks @bob-carpenter

Issue #56, opened 2015-07-06
Incorrect behaviour of CCDFs when there is a negative random variable


syclik:
_From @randommm on November 20, 2014 1:55_

The following program:

    transformed data {
      int y[2];
      y[1] <- 0;
      y[2] <- 1;
      print(poisson_ccdf_log(0, 5));
      print(poisson_ccdf_log(1, 5));
      print(poisson_ccdf_log(y, 5));
    }
    parameters {
      real any;
    }
    model {
      any ~ normal(0, 1);  
    }

Outputs: 

   -0.00676075
   -0.0412676
   -0.0480283

(the last number is the sum of the first two as expected)

However, the following program:

    transformed data {
      int y[2];
      y[1] <- -1;
      y[2] <- 1;
      print(poisson_ccdf_log(-1, 5));
      print(poisson_ccdf_log(1, 5));
      print(poisson_ccdf_log(y, 5));
    }
    parameters {
      real any;
    }
    model {
      any ~ normal(0, 1);  
    }

Outputs: 

    0
    -0.0412676
    0

Which doesn't make sense: the last number should have been -0.0412676 too.

The way the CCDFs are currently designed makes them ignore everything when there is a negative random variable inside the vector of random variables:

      // Explicit return for extreme values
      // The gradients are technically ill-defined, but treated as neg infinity
      for (size_t i = 0; i < stan::length(n); i++) {
        if (value_of(n_vec[i]) < 0) 
          return operands_and_partials.to_var(0.0);
      }

_Copied from original issue: stan-dev/stan#1142_

syclik:
_From @randommm on November 20, 2014 2:5_

Another closelly related bug, for neg_binomial_ccdf and neg_binomial_cdf, it was used <= instead of <

      for (size_t i = 0; i < stan::length(n); i++) {
        if (value_of(n_vec[i]) <= 0) 
          return operands_and_partials.to_var(0.0);
      }

Because of that:

neg_binomial_ccdf_log(0, 5, 5) == 0 == log(1)
but this is incorrect since P(X > 0) != 1

and

neg_binomial_cdf_log(0, 5, 5) == -inf == log(0)
but this is incorrect since P(X <= 0) != 0


    transformed data {
      print(neg_binomial_ccdf_log(0, 5, 5));
      print(neg_binomial_cdf_log(0, 5, 5));
    }
    parameters {
      real any;
    }
    model {
      any ~ normal(0, 1);  
    }

outputs:

    0
    -inf

[...]

PR #1716, opened 2020-02-15
Add discrete_range_cdf, discrete_range_lcdf and discrete_range_lccdf

### LINKED ISSUE #1683: Complete the discrete_range functions
## Description
To complete the work started in #1680, we should add `discrete_range_{cdf|lcdf|lccdf}`.

#### Current Version:
v3.1.0

mcol:
## Summary

This completes the set of functions for `discrete_range`. Fixes #1683.

## Tests

Added.

## Side Effects

None.

## Checklist

- [X] Math issue #1683


stan/math/prim/prob/discrete_range_lcdf.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+  scalar_seq_view<T_y> y_vec(y);
+  scalar_seq_view<T_lower> lower_vec(lower);
+  scalar_seq_view<T_upper> upper_vec(upper);
+  size_t size_y_lower = max_size(y, lower);
+  size_t N = max_size(y, lower, upper);
+
+  for (size_t n = 0; n < N; ++n) {
+    const double y_dbl = value_of(y_vec[n]);
+    if (y_dbl < value_of(lower_vec[n])) {
```
bob-carpenter: These are inconsistent.  Both this and the next conditional should return `LOG_ZERO` if anythng, but really they should be throwing exceptions because `y` is out of scope.  Also, all this should be int, not double.

stan/math/prim/prob/discrete_range_lcdf.hpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  double cdf(0.0);
+  for (size_t n = 0; n < N; n++) {
+    const double y_dbl = value_of(y_vec[n]);
+    const double lower_dbl = value_of(lower_vec[n]);
+    const double upper_dbl = value_of(upper_vec[n]);
+
+    cdf += log(y_dbl - lower_dbl + 1) - log(upper_dbl - lower_dbl + 1);
```
bob-carpenter: Integer arithmetic is fine in side these logs.  Preferable, actually, as we know it won't overflow.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
```diff
@@ -0,0 +1,65 @@
...
+ public:
+  void valid_values(vector<vector<double>>& parameters,
+                    vector<double>& ccdf_log) {
+    vector<double> param(3);
+
+    param[0] = 3;  // y
+    param[1] = 1;  // lower
+    param[2] = 5;  // upper
+    parameters.push_back(param);
+    ccdf_log.push_back(-0.916290731874154995573);  // expected ccdf_log
```
bob-carpenter: Rather than magic numbers like this, the expected value should be `log(2.0 / 5)`.  It's easy to see that's the right ansewr because there are 2 elements strictly greater than y.

Same for others.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
```diff
@@ -0,0 +1,65 @@
...
+    param[1] = 5;   // lower
+    param[2] = 15;  // upper
+    parameters.push_back(param);
+    ccdf_log.push_back(-0.606135803570315601085);  // expected ccdf_log
+
+    param[0] = 0;   // y
+    param[1] = -4;  // lower
+    param[2] = 5;   // upper
+    parameters.push_back(param);
+    ccdf_log.push_back(-0.693147180559945286227);  // expected ccdf_log
```
bob-carpenter: The upper boundary should be tested for completeness, y = upper.  There the value should be `LOG_ZERO`. It's in support, but has a zero CCDF.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  void invalid_values(vector<size_t>& /*index*/, vector<double>& /*value*/) {
+    // y
```
bob-carpenter: These are going to need the out-of-support tests.

In general, this shouldn't have three lines of comments double spaced.  A simple `// all values valid` comment is better because it doesn't look like someone forgot to fill in values here.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // lower
+
+    // upper
+  }
+
+  bool has_lower_bound() { return false; }
```
bob-carpenter: I'm not exactly sure how these tests work, but this distribution has a lower and upper bound.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
```diff
@@ -0,0 +1,65 @@
...
+    return stan::math::discrete_range_lccdf(y, lower, upper);
+  }
+
+  template <class T_y, class T_lower, class T_upper, typename T3, typename T4,
+            typename T5>
+  stan::return_type_t<T_y, T_lower, T_upper> ccdf_log_function(
+      const T_y& y, const T_lower& lower, const T_upper& upper, const T3&,
+      const T4&, const T5&) {
+    if (y < lower || y > upper) {
+      return stan::math::LOG_ZERO;
```
bob-carpenter: These should throw.

test/prob/discrete_range/discrete_range_cdf_log_test.hpp
```diff
@@ -0,0 +1,63 @@
...
+using stan::math::var;
+using std::numeric_limits;
+using std::vector;
+
+class AgradCdfLogDiscreteRange : public AgradCdfLogTest {
+ public:
+  void valid_values(vector<vector<double>>& parameters,
+                    vector<double>& cdf_log) {
+    vector<double> param(3);
+
```
bob-carpenter: The boundary conditions `y = lower` and `y = upper` should be tested.  This applies to the other tests, too.

test/unit/math/prim/prob/discrete_range_ccdf_log_test.cpp
```diff
@@ -0,0 +1,15 @@
+#include <stan/math/prim.hpp>
+#include <gtest/gtest.h>
+
+TEST(ProbDiscreteRange, cdf_log_matches_lccdf) {
```
bob-carpenter: We know the implementation just delegates, but these tests are to prevent future breakage, so they should be as throough as everything else.    One way to do this is something like this:

```
for (int lower = 0; lower < 5; ++lower) {
  for (int upper = lower; upper < 5; ++upper)
    for (int y = lower; y <= upper; ++y)
       ... test that they do the same thing...
```

Same for `discrete_range_cdf`.

stan/math/prim/prob/discrete_range_cdf.hpp
```diff
@@ -0,0 +1,81 @@
...
+    if (y_dbl < value_of(lower_vec[n])) {
+      return 0;
+    }
+    if (y_dbl > value_of(upper_vec[n])) {
+      return 1;
+    }
+  }
+
+  double cdf(1.0);
+  for (size_t n = 0; n < N; n++) {
```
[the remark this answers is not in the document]
mcol: I also thought that the operations that could be precomputed here were only sums, rather than more involved operations, so I'll leave this. I have a PR for other distributions that tries to clean up this kind of operations.

stan/math/prim/prob/discrete_range_lcdf.hpp
mcol: Following what is done in other lcdfs, if `y` is below the lower bound, the lcdf returns `LOG_ZERO`, but if it's above the upper boound it returns `log(1)`, by definition of a log CDF. But I'll start throwing, so this is moot.

stan/math/prim/prob/discrete_range_lccdf.hpp
```diff
@@ -0,0 +1,81 @@
...
+      return 0;
+    }
+    if (y_dbl > value_of(upper_vec[n])) {
+      return LOG_ZERO;
+    }
+  }
+
+  double ccdf(0.0);
+  for (size_t n = 0; n < N; n++) {
+    const double y_dbl = value_of(y_vec[n]);
```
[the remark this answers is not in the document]
mcol: I can make them int, `scalar_seq_view` can deal with them already.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
mcol: Those are used for distributions that have a fixed support, say the `beta` that is in [0,1]. There is a kludge for `uniform` that sets only the upper bound. So I'm not sure it's worth trying to use those here, I can add some prim tests about boundary behaviour.

bob-carpenter:
> 
> Following what is done in other lcdfs, if y is below the lower bound, the lcdf returns LOG_ZERO, but if it's above the upper boound it returns log(1), by definition of a log CDF. But I'll start throwing, so this is moot.

Thanks.

The goal is to have all of the CDF functions throw if the argument y is outside of support.

Here's a mathy way to think about it:  if the CDF is defined as Pr[X < x], the integral (infinite sum) won't be well defined unless the pdf (pmf) is defined.  We've taken the pdfs/pmfs to be undefined and hence throw rather than be zero outside of support.

From a computational point of view, it's about error handling.  It's almost always going to be a programming mistake to try to evaluate a CDF out of support.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
mcol: I've added prim tests for boundaries. I'm struggling with the distribution test framework to behave as desired for boundaries/exceptions, it just doesn't seem designed to support those. I'll try to summarise my findings about that when I've got a better handle on thosse.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
mcol: I've added prim test for out-of-support values. The invalid_values section of distribution test doesn't let you express that easily, they are designed to capture values that are always invalid (say a negative rate parameter in a Poisson distribution). The formatting follows what's done in all other tests.

mcol:
I've only addressed comments not related to throwing behaviour. While changing the code to do that is trivial, the distribution test framework is not setup to deal with that. I'll try to explain what gets in the way, considering the `lccdf` test for clarity (same issues appear for other tests):

1. If we don't specify that the distribution has lower/upper bounds (leaving the `has_[lower|upper]_bound()` to return `false`), the test will try to evaluate the lccdf at minus infinity: however such value is outside the range we specify, so we throw and the test fails

2. If we specify a lower bound setting `has_lower_bound()` to return `true`, and adding a `double lower_bound()` that returns a specified value), the test will evaluate the lccf at that value, expecting it to return 0: however this is a discrete distribution, to the lccdf at the lower bound is not 0 as for a continuous distribution, but corresponds to the lpmf evaluated at that point. This is is fixable by adding an `is_discrete()` method and modify that check if the distribution is discrete.

3. Even if we fix point 2, then all lower bounds in the set of valid values must match: this is because the distribution test will try all combinations of valid values, expecting all to be compatible. But now one of the bounds might contradict the value reported by `lower_bound()` and the test will fail.

4. Same for upper bounds.

So at this stage I can see the following: don't throw if values are outside the support and deal with that in #56 (and whatever other PR is needed to make distribution tests more flexible), or don't write distribution tests for these functions but rely on hand-written tests.

mcol:
Hi @bob-carpenter, at the math development meeting I discussed these shortcomings in the distribution test framework with @bbbales2 and @syclik. Their take was that trying to doing further work on that framework was built before C++11 and extending it to capture correctly the new requirements will take energies that would be better spent elsewhere.

In particular, @bbbales2 suggested extending the `expect_ad` framework to cover distributions too, and thus potentially help us move away from the distribution test framework. The point is: what would be the cost of such an undertaking? What's your opinion concerning this line of thought? Would you have alternate suggestions?

I'll be happy to move this discussion elsewhere if necessary.

As for this PR, would you be happy to review it as is for the fixes and the additional tests I've already added?

bob-carpenter:
I would suggest just writing a few extra tests on the outside to test throwing behavior. 

We'll be rewriting the whole distribution test framework at some point, so it's probably not worth trying to extend what we have.

We can also continue to not throw outside of support if none of the other fixes work.  The doc's just going to have to match.

bob-carpenter:
I'm totally swamped with math PRs.  I can get to this some time next week if that's soon enough.

test/prob/discrete_range/discrete_range_cdf_log_test.hpp
mcol: I'm testing boundaries in prim tests.

test/unit/math/prim/prob/discrete_range_ccdf_log_test.cpp
mcol: I've changed tests to do as suggested.

mcol:
As I mentioned above, I can't introduce throwing behaviour unless we go all the way without distribution tests. So I'd rather not throw until we have a test framework that can deal with that. The docs for this function are already consistent with not throwing. I've now added prim tests for out of support values.

There's no hurry for this, it won't get conflicts.

bob-carpenter:
> I'm testing boundaries in prim tests.

The boundaries should get tested for all types.

bob-carpenter:
OK.  We need to fix all of the CDFs, so no big deal if this one doesn't get fixed for a while longer.  And it sounds like the tests all need to get fixed, too.  Sorry for the horrible mess that's our existing CDF code and testing code.

mcol:
In https://github.com/stan-dev/math/pull/1716#discussion_r380837157 you suggested that I should use ints and remove all `value_of`. In order to add mix tests, I'll have to undo that and generalize again the return type and all assignments that now don't use `value_of`. Before I start doing that, could you confirm that it's the right direction?

bob-carpenter:
My point was just that if you know the type is `int`, you'll never need `value_of`.  So I was just suggesting getting rid of `value_of` where it's not needed.

mcol:
All parameters are int or vector<int> (although we can't enforce it just yet in the function signature): that's why I had only added prim tests. To make these functions work with var/fvar I'm reintroducing `value_of` everywhere with rev/fwd tests. I'm still not clear if that's what you want and why, but my understanding is still patching in some of these areas.

bob-carpenter:
Sorry if I've been sending mixed messages.  

If all arguments are int, then we don't need derivative tests because there are no arguments to differentiate!

mcol:
Cool, so the current version should work and prim tests are enough. As far as I can see there's nothing more to be done, but let me know if you spot anything.

bob-carpenter:
I can review this Friday.

stan/math/prim/prob/discrete_range_ccdf_log.hpp
```diff
@@ -0,0 +1,20 @@
...
+namespace stan {
+namespace math {
+
+/** \ingroup prob_dists
+ * @deprecated use <code>discrete_range_lccdf</code>
+ */
+template <typename T_y, typename T_lower, typename T_upper>
+double discrete_range_ccdf_log(const T_y& y, const T_lower& lower,
+                               const T_upper& upper) {
+  return discrete_range_lccdf<T_y, T_lower, T_upper>(y, lower, upper);
```
bob-carpenter: [optional]
the template parameters on the function call will be inferred so you don't need to explicitly specify them.

stan/math/prim/prob/discrete_range_cdf.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * supplied, returns the product of the CDFs.
+ * @throw std::domain_error if upper is smaller than lower.
+ * @throw std::invalid_argument if non-scalar arguments are of different
+ * sizes.
+ */
+template <typename T_y, typename T_lower, typename T_upper>
+double discrete_range_cdf(const T_y& y, const T_lower& lower,
+                          const T_upper& upper) {
+  static const char* function = "discrete_range_cdf";
+  if (size_zero(y, lower, upper)) {
```
bob-carpenter: This should come after the consistent size checks.  It should be illegal to call with mismatched arguments even one of the sizes is zero.

stan/math/prim/prob/discrete_range_cdf.hpp
```diff
@@ -0,0 +1,75 @@
...
+                         "Upper bound parameter", upper);
+  check_greater_or_equal(function, "Upper bound parameter", upper, lower);
+
+  scalar_seq_view<T_y> y_vec(y);
+  scalar_seq_view<T_lower> lower_vec(lower);
+  scalar_seq_view<T_upper> upper_vec(upper);
+  size_t N = max_size(y, lower, upper);
+
+  for (size_t n = 0; n < N; ++n) {
+    const int y_dbl = y_vec[n];
```
bob-carpenter: [optional]
We have not in general marked our local args as `const`.  It's not a mistake per se, but it clutters the code up a lot.

stan/math/prim/prob/discrete_range_lccdf.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * supplied, returns the sum of the log CCDFs.
+ * @throw std::domain_error if upper is smaller than lower.
+ * @throw std::invalid_argument if non-scalar arguments are of different
+ * sizes.
+ */
+template <typename T_y, typename T_lower, typename T_upper>
+double discrete_range_lccdf(const T_y& y, const T_lower& lower,
+                            const T_upper& upper) {
+  static const char* function = "discrete_range_lccdf";
+  if (size_zero(y, lower, upper)) {
```
bob-carpenter: after checks

stan/math/prim/prob/discrete_range_lcdf.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * supplied, returns the sum of the log CDFs.
+ * @throw std::domain_error if upper is smaller than lower.
+ * @throw std::invalid_argument if non-scalar arguments are of different
+ * sizes.
+ */
+template <typename T_y, typename T_lower, typename T_upper>
+double discrete_range_lcdf(const T_y& y, const T_lower& lower,
+                           const T_upper& upper) {
+  static const char* function = "discrete_range_lcdf";
+  if (size_zero(y, lower, upper)) {
```
bob-carpenter: after checks

mcol:
Thank you!

PR #1715, opened 2020-02-15
Cleanup mdivide_* and tests

mcol:
## Summary

This cleans up the `mdivide` functions, mainly to use `return_type_t` instead of `promote_common`. There are a couple of unused header removed and a correction to the return type declared for the fwd version of `mdivide_left_tri_low`. Fixes #1714.

## Tests
Reworked and expanded so that the left and right versions contain the same tests. Some boundary condition tests will be completed in #1689.

## Side Effects

None.

## Checklist

- [X] Math issue #1714

- [X] Copyright holder: Marco Colombo

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [X] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [X] the code is written in idiomatic C++ and changes are documented in the doxygen

- [X] the new changes are tested

PR #1698, opened 2020-02-11
Improve tests and documentation for quad_form and quad_form_sym

mcol:
## Summary
This started out to be a simple clean up of the `quad_form` and `quad_form_sym` tests, but it turned into a rabbit hole. There were a few issues around those functions:
- the prim tests were in a single file: I've now split them into one file per function
- the mix tests for `quad_form_sym` were broken because their input was not symmetric: I now symmetrize the input matrix so that tests execute as expected (this also required fixing some tolerances)
- again in that test, variable `b22` was not initialized, and variable `b23` was not used
- there were no mix tests for `quad_form`: this is because there was no fwd version of that function, which I've now added
- some implementations could be consolidated by better template metaprogramming
- there was no doxygen documentation, and now all files have some.

Fixes #1695.

## Tests
Added and fixed as mentioned above.

## Side Effects

None.

## Checklist

- [X] Math issue #1695

- [X] Copyright holder: Marco Colombo

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [X] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [X] the code is written in idiomatic C++ and changes are documented in the doxygen

- [X] the new changes are tested

PR #1718, opened 2020-02-16
Remove size zero checks from check_multiplicable

### LINKED ISSUE #1689: remove size zero checks in `check_multiplicable`

mcol:
## Summary
This removes the `check_positive` calls from `check_multiplicable`, so that we can multiply matrices/vectors even if they have size 0. This means that for several functions, we need to return early for size 0 inputs with the correct result, as otherwise Eigen might raise some assertions. This restores the `check_multiplicable` tests inside `multiply.hpp` that were removed in #1685.

Perhaps, the only less obvious change concerns the following functions:
- `matrix_exp_multiply` and `scale_matrix_exp_multiply`
- `mdivide_left_ldlt` and `mdivide_right_ldlt`
- `trace_gen_inv_quad_form_ldlt`, `trace_inv_quad_form_ldlt`
For these functions, we move the `check_multiplicable` earlier as now it doesn't interferes with the size 0 check. The latter can then be simplified since we know that A and B are multiplicable.

Instead, for `quad_form_sym` we now move `check_multiplicable` before `check_symmetric` because it's cheaper.

Fixes #1689.

## Tests
Tests have been changed to account for the fact that now these functions don't throw. They have also been expanded to account for more boundary conditions. Note that some tests had already been added in #1698 and #1715, especially the mix tests for `mdivide_*`.

## Side Effects

None?

## Checklist

- [X] Math issue #1689

### DIFF
--- a/stan/math/fwd/fun/mdivide_left.hpp
+++ b/stan/math/fwd/fun/mdivide_left.hpp
@@ -19,6 +19,9 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
     const Eigen::Matrix<fvar<T>, R2, C2> &b) {
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
+  if (A.size() == 0) {
+    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
+  }
 
   Eigen::Matrix<T, R1, C2> inv_A_mult_b(A.rows(), b.cols());
   Eigen::Matrix<T, R1, C2> inv_A_mult_deriv_b(A.rows(), b.cols());
@@ -58,6 +61,9 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
     const Eigen::Matrix<fvar<T>, R2, C2> &b) {
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
+  if (A.size() == 0) {
+    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
+  }
 
   Eigen::Matrix<T, R2, C2> val_b(b.rows(), b.cols());
   Eigen::Matrix<T, R2, C2> deriv_b(b.rows(), b.cols());
@@ -78,6 +84,9 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
     const Eigen::Matrix<double, R2, C2> &b) {
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
+  if (A.size() == 0) {
+    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
+  }
 
   Eigen::Matrix<T, R1, C2> inv_A_mult_b(A.rows(), b.cols());
   Eigen::Matrix<T, R1, C1> inv_A_mult_deriv_A(A.rows(), A.cols());
--- a/stan/math/fwd/fun/mdivide_left.hpp
+++ b/stan/math/fwd/fun/mdivide_left.hpp
@@ -20,7 +20,7 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
   if (A.size() == 0) {
-    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
+    return {0, b.cols()};
   }
 
   Eigen::Matrix<T, R1, C2> inv_A_mult_b(A.rows(), b.cols());
@@ -62,7 +62,7 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
   if (A.size() == 0) {
-    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
+    return {0, b.cols()};
   }
 
   Eigen::Matrix<T, R2, C2> val_b(b.rows(), b.cols());
@@ -85,7 +85,7 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
   if (A.size() == 0) {
-    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
+    return {0, b.cols()};
   }
 
   Eigen::Matrix<T, R1, C2> inv_A_mult_b(A.rows(), b.cols());
--- a/stan/math/prim/fun/mdivide_left_tri.hpp
+++ b/stan/math/prim/fun/mdivide_left_tri.hpp
@@ -37,7 +37,7 @@ inline Eigen::Matrix<return_type_t<T1, T2>, R1, C2> mdivide_left_tri(
   check_square("mdivide_left_tri", "A", A);
   check_multiplicable("mdivide_left_tri", "A", A, "b", b);
   if (A.rows() == 0) {
-    return Eigen::Matrix<return_type_t<T1, T2>, R1, C2>(0, b.cols());
+    return {0, b.cols()};
   }
 
   return Eigen::Matrix<return_type_t<T1, T2>, R1, C1>(A)
@@ -97,7 +97,7 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
   check_square("mdivide_left_tri", "A", A);
   check_multiplicable("mdivide_left_tri", "A", A, "b", b);
   if (A.rows() == 0) {
-    return Eigen::MatrixXd(0, b.cols());
+    return {0, b.cols()};
   }
 
 #ifdef STAN_OPENCL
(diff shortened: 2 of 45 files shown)


mcol:
@dpsimpson This is my take on the size zero checks, if you wanted to have a look.

In any case, this is ready for review.

stan/math/fwd/fun/mdivide_left.hpp
```diff
@@ -19,6 +19,9 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
     const Eigen::Matrix<fvar<T>, R2, C2> &b) {
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
+  if (A.size() == 0) {
+    return Eigen::Matrix<fvar<T>, R1, C2>(0, b.cols());
```
bob-carpenter: This should just be `return {0, b.cols()};` The constructor is redundant given that it's the return type.

I'm only going to marke this one, but this is a problem with almost every return defined in this PR.

stan/math/prim/fun/mdivide_left_tri.hpp
```diff
@@ -89,6 +96,10 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
     const Eigen::Matrix<double, R2, C2> &b) {
   check_square("mdivide_left_tri", "A", A);
   check_multiplicable("mdivide_left_tri", "A", A, "b", b);
+  if (A.rows() == 0) {
+    return Eigen::MatrixXd(0, b.cols());
```
bob-carpenter: This should be `return {0, b.cols()}` so that the object constructed matches the return type, which is `Eigen::Matrix<double, R1, C2>` here.  I don't see any requirement that `R1 = C2 = -1` as implied by `Eigen::MatrixXd`.  

It'll work as written because Eigen ignores the size templates in assignment, but it's confusing that the types don't match.

stan/math/prim/fun/mdivide_right_spd.hpp
```diff
@@ -33,9 +33,12 @@ inline Eigen::Matrix<return_type_t<T1, T2>, R1, C2> mdivide_right_spd(
...
   static const char *function = "mdivide_right_spd";
   check_multiplicable(function, "b", b, "A", A);
-  check_positive(function, "rows", A.rows());
   check_symmetric(function, "A", A);
   check_not_nan(function, "A", A);
+  if (A.size() == 0) {
+    return Eigen::Matrix<return_type_t<T1, T2>, R1, C2>(b.rows(), 0);
+  }
+
   // FIXME: After allowing for general MatrixBase in mdivide_left_spd,
```
bob-carpenter: [optional---this PR didn't introduce the problem]
These should say TODO and have a developer name attached, as in `TODO(mcol)`.  At least that's the Google coding standard we've adopted.

stan/math/prim/fun/trace_gen_inv_quad_form_ldlt.hpp
```diff
@@ -42,12 +42,11 @@ inline return_type_t<T1, T2, T3> trace_gen_inv_quad_form_ldlt(
     const Eigen::Matrix<T1, R1, C1> &D, const LDLT_factor<T2, R2, C2> &A,
     const Eigen::Matrix<T3, R3, C3> &B) {
   check_square("trace_gen_inv_quad_form_ldlt", "D", D);
-  if (D.size() == 0 && A.cols() == 0 && B.rows() == 0 && B.cols() == 0) {
-    return 0;
-  }
-
   check_multiplicable("trace_gen_inv_quad_form_ldlt", "A", A, "B", B);
   check_multiplicable("trace_gen_inv_quad_form_ldlt", "B", B, "D", D);
+  if (D.size() == 0 && A.cols() == 0) {
```
bob-carpenter: This should be `(D.size() == 0 || A.size() == 0 || B.size() == 0)`.  If any of them are size zero, the result is going to either be empty or have all zero entries, so the trace is going to be 0.  There's another instance of this, too.

stan/math/rev/fun/mdivide_left.hpp
```diff
@@ -179,6 +179,9 @@ inline Eigen::Matrix<var, R1, C2> mdivide_left(
 
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
+  if (A.size() == 0) {
```
bob-carpenter: [optional, because it's a minor point]

This should be `return {0,  b.cols()};` and `res` should be defined after this test.  Same elsewhere.

stan/math/rev/fun/mdivide_left_ldlt.hpp
```diff
@@ -240,13 +239,12 @@ inline Eigen::Matrix<var, R1, C2> mdivide_left_ldlt(
 template <int R1, int C1, int R2, int C2>
 inline Eigen::Matrix<var, R1, C2> mdivide_left_ldlt(
     const LDLT_factor<var, R1, C1> &A, const Eigen::Matrix<double, R2, C2> &b) {
+  check_multiplicable("mdivide_left_ldlt", "A", A, "b", b);
   Eigen::Matrix<var, R1, C2> res(b.rows(), b.cols());
-  if (A.cols() == 0 && b.rows() == 0) {
```
bob-carpenter: [comment]
good catch on the redundant (given `check_multiplicable`) test

stan/math/rev/fun/trace_gen_inv_quad_form_ldlt.hpp
```diff
@@ -38,12 +38,11 @@ inline var trace_gen_inv_quad_form_ldlt(const Eigen::Matrix<T1, R1, C1> &D,
                                         const LDLT_factor<T2, R2, C2> &A,
                                         const Eigen::Matrix<T3, R3, C3> &B) {
   check_square("trace_gen_inv_quad_form_ldlt", "D", D);
-  if (D.size() == 0 && A.cols() == 0 && B.rows() == 0 && B.cols() == 0) {
-    return 0;
-  }
-
   check_multiplicable("trace_gen_inv_quad_form_ldlt", "A", A, "B", B);
   check_multiplicable("trace_gen_inv_quad_form_ldlt", "B", B, "D", D);
+  if (D.size() == 0 && A.cols() == 0) {
```
bob-carpenter: This is another case where it should be `||` and maybe include `B`.

test/unit/math/prim/fun/mdivide_left_tri_low_test.cpp
```diff
@@ -0,0 +1,38 @@
+#include <stan/math/prim.hpp>
+#include <test/unit/math/prim/fun/expect_matrix_eq.hpp>
+#include <gtest/gtest.h>
+
+#define EXPECT_MATRIX_NEAR(A, B, DELTA) \
+  for (int i = 0; i < A.size(); i++)    \
+    EXPECT_NEAR(A(i), B(i), DELTA);
+
+TEST(MathMatrixPrim, mdivide_left_tri_low_val) {
```
bob-carpenter: Thanks for filling in unrelated missing tests!

There should be a size 1 test as well as size 2 and size 0, just to check that boundary conditions are working.

stan/math/rev/fun/mdivide_left.hpp
mcol: Cool, I was going to do that, but then went back as it looked as I was moving unrelated code. I'll be happy to change these..

stan/math/prim/fun/mdivide_right_spd.hpp
mcol: That goes back to 2015 or so.

bob-carpenter:
> On Feb 18, 2020, at 1:13 PM, Marco Colombo <notifications@github.com> wrote:
> 
> @mcol commented on this pull request.
> 
> In stan/math/prim/fun/mdivide_right_spd.hpp:
> 
> >    check_symmetric(function, "A", A);
>    check_not_nan(function, "A", A);
> +  if (A.size() == 0) {
> +    return Eigen::Matrix<return_type_t<T1, T2>, R1, C2>(b.rows(), 0);
> +  }
> +
>    // FIXME: After allowing for general MatrixBase in mdivide_left_spd,
> 
> That goes back to 2015 or so.

Before we'd enabled C++11 constructor patterns, so we couldn't have done this the right way then.

stan/math/fwd/fun/mdivide_left.hpp
mcol: I'm glad this works! When I tried something similar in a different PR, it wouldn't compile. Perhaps because here the return type is spelled out unambiguously.

stan/math/prim/fun/trace_gen_inv_quad_form_ldlt.hpp
mcol: Thanks for spotting this! I think it can be just `(D.size() == 0 || A.size() == 0)`, as either case implies `B.size() == 0`.

PR #1720, opened 2020-02-16
Feature/0123 complex spec

bob-carpenter:
## Summary

This is the first pull request of two for adding support for autodiff of `std::complex` values in Stan.  This PR has 

* a utility base class for std::complex implemented with a reduced-type form of the CRTP
* specializations for `std:complex<var>` and `std::complex<fvar<T>>`
* upgrades to the ad testing framework to deal with serializing/deserializing complex numbers
* metaprogram extra case for complex number scalar type

## Tests

The AD test framework was extended to deal with complex numbers and used to test the member operators (compound arithmetic assignments only).

Specialized tests were implemented for the constructors and defaults.

The tests for std::complex classes are in `test/unit/math/mix/core/std_complex_test.cpp`.  Tests for additions to the testing framework itself are in the testing framework at `test/unit/math/`.

## Side Effects

No

## Checklist

- [x] Math issue #123 [but this doesn't solve the whole issue, this is just PR 1 of 2 to make reviewing easy]

- [x] Copyright holder: (fill in copyright holder information):  Columbia University

- [x] the basic tests are passing  (I only unit tested new additions; test-math-dependencies is broken)


bob-carpenter:
@serban-nicusor-toptal There was a failure that looked like it was due to email:

https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-1720/1/pipeline/173

Here's the text from the failure from Always-run tests part 2, Windows Headers & Unit.

```
C:\Jenkins\workspace\Math_Pipeline_PR-1720>attrib -r -s /s /d 
File not found - C:\Jenkins\workspace\Math_Pipeline_PR-1720\*.*
java.nio.channels.ClosedChannelException[Clang (LLVM based)] Skipping execution of recorder since overall result is 'FAILURE'
An attempt to send an e-mail to empty list of recipients, ignored.
```

serban-nicusor-toptal:
Hey Bob,
`java.nio.channels.ClosedChannelException` in Jenkins [points](https://www.inf.unibz.it/~calvanese/teaching/java-docs/5.0/api/java/nio/channels/class-use/ClosedChannelException.html) to the fact that the machine running your tests closed the channel, happens when a machine suddenly dies or network connectivity issues. Sometimes in rare cases a job can kill a machine ( cpu/gpu temps as an example  ). See logs at the bottom [here](https://jenkins.mc-stan.org/job/Math%20Pipeline/view/change-requests/job/PR-1720/1/console).

I can't be sure now if it's a Jenkins issue, I see that it's now running. Please let me know if it fails again with the same reason and I'll look into it.

Thanks

bob-carpenter:
@serban-nicusor-toptal --- yes, it failed again with the same error.  I think that may be because the windows machine was offline.  @bgoodri --- do you know if it's online again?

bgoodri:
Back online. Had to pull its power cord out of the socket.

bob-carpenter:
> Back online. Had to pull its power cord out of the socket.

Thanks, @bgoodri.  Now all I need is a review.

stan/math/fwd/core/std_complex.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+namespace std {
+
+/**
+ * Specialization of the standard libary complex number type for
+ * reverse-mode autodiff type `stan::math::fvar<T>`.
+ */
+template <typename T>
+class complex<stan::math::fvar<T>>
+    : public stan::math::complex_base<stan::math::fvar<T>> {
```
SteveBronder: If this is a crtp should this be

```
public stan::math::complex_base<complex<stan::math::fvar<T>>>
```

So the type is recurrent in the base class?

stan/math/prim/core/complex_base.hpp
```diff
@@ -0,0 +1,275 @@
...
+   * Construct a complex base with the specified real and imaginary
+   * parts.
+   *
+   * @tparam T real type (must be assignable to `V`)
+   * @tparam U imaginary type (must be assignable to `V`)
+   * @param[in] re real part
+   * @param[in] im imaginary part (default 0)
+   */
+  template <typename T, typename U>
+  complex_base(const T& re, const U& im) : re_(re), im_(im) {}
```
SteveBronder: Should these be passed by value or reference?

stan/math/prim/core/complex_base.hpp
```diff
@@ -0,0 +1,275 @@
...
+   * Divides this by other.
+   *
+   * @tparam X value type of complex argument (assignable to value type)
+   * @param[in] other a complex value of compatible type
+   * @return this complex number
+   */
+  template <typename X>
+  complex_type& operator/=(const std::complex<X>& other) {
+    using stan::math::square;
+    value_type sum_sq_im = square(other.real()) + square(other.imag());
```
SteveBronder: Can/should these be `auto`?

stan/math/fwd/core/std_complex.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <stan/math/fwd/core/fvar.hpp>
+#include <complex>
+#include <type_traits>
+
+namespace std {
+
+/**
+ * Specialization of the standard libary complex number type for
+ * reverse-mode autodiff type `stan::math::fvar<T>`.
+ */
```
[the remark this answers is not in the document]
bob-carpenter: It's no big deal either way, but I'd rather not.  We haven't linked other specializations to the C++ reference and that's not the official library reference doc, just a presentation of some of it.  The fact that it's a specialization is in the template declaration.

I'll leave you to decide whether I should add the link.

stan/math/fwd/core/std_complex.hpp
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Construct a complex number with zero real and imaginary
+   * components.
+   */
+  complex() : base_t() {}
+
+  /**
+   * Destroy this complex number.
+   */
+  ~complex() {}
```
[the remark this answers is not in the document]
bob-carpenter: I was trying to follow the "rule of three".  From the Wikipedia:

> The rule of three (also known as the Law of The Big Three or The Big Three) is a rule of thumb in C++ (prior to C++11) that claims that if a class defines any of the following then it should probably explicitly define all three: destructor. copy constructor. copy assignment operator.

I'm overloading the define of the copy operator to deal with non-matching types (like assigning `std::complex<double>` to `std::complex<var>`.   So I'm not sure where this falls.  I'm happy to get rid of it if it's OK style.

stan/math/fwd/core/std_complex.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * Constructs complex number from argument, which may be a scalar or
+   * a complex number.  If it is a scalar, the real part is set to the
+   * scalar value and the imaginary part to zero;  if it is complex,
+   * the real and imaginary parts are set to those of the argument.
+   *
+   * @tparam U type of argument
+   * @param[in] x the scalar or complex argument
+   */
+  template <typename U>
+  complex(const U& x) : base_t(x) {}  // NOLINT(runtime/explicit)
```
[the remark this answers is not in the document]
bob-carpenter: It won't work if it's marked `explicit`.  We need to be able to promote arithmetic types to complex if necessary.

It will only compile if `U` is something assignable to the class template `T`, which means the restriction is to arithmetic types (primitives) or `var` (or to `fvar<U>` for the forward version).  Elsehwere, the decision has been not to tighten down the signature with type traits unless needed for ambiguity resolution.

stan/math/fwd/core/std_complex.hpp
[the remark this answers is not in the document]
SteveBronder: 👍 I'm fine with not having it

stan/math/fwd/core/std_complex.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Construct a complex number with zero real and imaginary
+   * components.
+   */
+  complex() : base_t() {}
```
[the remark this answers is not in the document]
bob-carpenter: The default constructor would have the wrong behavior.  `std::complex()` needs to initialize the real and imaginary components to zero.

There is nothing to move, so move semantics won't do anything.  Should I still define the move copy constructor and copy assignment?

stan/math/fwd/core/std_complex.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * number and set imaginary part to zero.
+   *
+   * @tparam V type of value
+   * @param[in] x value to assign
+   * @return this complex number after assigning other
+   */
+  template <typename V>
+  complex_type& operator=(const V& x) {
+    return base_t::operator=(x);
+  }
```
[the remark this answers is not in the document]
bob-carpenter: `V` must be assignable to the value type of the complex number.  It won't compile otherwise.  Adding an `enable_if` restriction would just cause the compiler to fail in a different way.  We don't need the restriction for disambiguation.

My understanding is that we generally do not want to try to explicitly enumerate all the possible instantiations of a template parameter with type traits.  We don't do that on any of the `prim` functions in our math library, for example.  You can try to pass a `std::string` to `log1m`, but it won't compile.

stan/math/prim/core/complex_base.hpp
```diff
@@ -0,0 +1,275 @@
...
+   * @param[in] other another complex to use as source
+   */
+  template <typename T>
+  complex_base(const std::complex<T>& other)  // NOLINT(runtime/explicit)
+      : complex_base(other.real(), other.imag()) {}
+
+  /**
+   * Destroy this complex number.
+   */
+  ~complex_base() {}
```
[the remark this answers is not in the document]
bob-carpenter: We'll need it for the rule of 5 if we want to implement copy constructors in cases where they don't do anything.

stan/math/prim/core/complex_base.hpp
```diff
@@ -0,0 +1,275 @@
...
+ protected:
+  /**
+   * Real part.
+   */
+  V re_;
+
+  /**
+   * Imaginary part.
+   */
+  V im_;
```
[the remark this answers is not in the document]
bob-carpenter: Yes, `re_` and `im_` is what the standard library uses.  We don't have to follow that, but I didn't see any reason to innovate on member variable names in specializations.

stan/math/prim/core/complex_base.hpp
```diff
@@ -0,0 +1,275 @@
...
+
+  /**
+   * Return a reference to thi class cast to the derived complex
+   * type.
+   *
+   * @return reference to this class cast to the complex return type
+   */
+  complex_type& derived_complex_ref() {
+    return static_cast<complex_type&>(*this);
+  }
```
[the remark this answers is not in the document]
bob-carpenter: I didn't do that because it's not an exact use of the CRTP.  It's just CRTP-like, so there's no template parameter `Derived`.  Instead there's a template parameter `T` and the equivalent of derived is `std::complex<T>`, which is typedef-ed as `complex_type`.  

But I'm happy to change it if you think that'd be clearer.

stan/math/fwd/core/std_complex.hpp
bob-carpenter: No.  The `complex` (which would have to be `std::complex`) is redundant.  I only take the value type `T` as the template parameter, but the "derived" class is `std::complex<T>`.  If I suppoied all of `complex<fvar<T>>`, then I'd just have to go fishing the `T` out with traits again.

stan/math/prim/core/complex_base.hpp
bob-carpenter: By reference.  They might be primitive, `var`, or `fvar`.   Passing by value would be performant for primitives and `var`, but `fvar<double>` and `fvar<var>` are 16 bytes, and `fvar<fvar<double>>` and `fvar<fvar<var>>` are 32 bytes.

I could break out an arithmetic type or `var` base class that would deal with the 8-byte ones separately.  But I don't think there's a huge saving from passing primitives by value vs. constant reference and would rather not have the complication in the constructors that would be required to do all that type splitting on primitive or var vs. fvar.

stan/math/prim/core/complex_base.hpp
bob-carpenter: What's the feeling about `auto` when there's a simple type we can use?  I find the explicit types simpler if they're small.  I'm not bothered either way, so let me know.  Our existing code is all over the place on this issue.

bob-carpenter:
@SteveBronder :  I replied to all the questions.  I think everything's OK as is, but let me know about rule of 5, naming the derived class, specializing for arithmetic types, etc.

stan/math/fwd/core/std_complex.hpp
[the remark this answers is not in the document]
SteveBronder: > It won't work if it's marked explicit. We need to be able to promote arithmetic types to complex if necessary.

As is yes but if you added the `operator=` for scalar types it would work right?

I'm fine with it not being explicit but I'll leave this open for now until I have time to sit down and play with this / look at the spec

(also fyi my review is a comment and not a request for changes so everything here is optional)

stan/math/fwd/core/std_complex.hpp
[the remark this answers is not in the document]
SteveBronder: >  Should I still define the move copy constructor and copy assignment and just define them = default?

I think it's fine to add them if the compiler wants to use them and they can just be default

stan/math/prim/core/complex_base.hpp
SteveBronder: > By reference. They might be primitive, var, or fvar. Passing by value would be performant for primitives and var, but fvar<double> and fvar<var> are 16 bytes, and fvar<fvar<double>> and fvar<fvar<var>> are 32 bytes.

I've been flip flopping on this too much to make a declarative statement, but I think it comes down to cache locality whether it's good to pass by value or reference. Most modern processors have some KB size of L1 cache with 64 byte cache lines. In section 7.2.1 of the 2nd edition of the C++ templates book he comments that if a signature with a const& argument is not inlined then the compiler can end up clearing cache out

> However, passing an address can create uncertainties for the compiler when it compiles the caller's code: What is the callee doing with that address? In theory, the callee can change all the values that are "reachable" through that address. That means, that the compiler has to assume that all the values it may have cached (usually, in machine registers) are invalid after the call. Reloading all those values can be quite expensive. You may be thinking that we are passing by _constant_ reference: Cannot the compiler deduce from that that no change can happen? Unfortunately, that is not the case because the caller may modify the referenced object through it own, non-const reference.

And if the compiler can't deduce that none of the methods a method calls uses const_cast then I think the cache is almost guaranteed to be cleared out

> I could break out an arithmetic type or var base class that would deal with the 8-byte ones separately. But I don't think there's a huge saving from passing primitives by value vs. constant reference and would rather not have the complication in the constructors that would be required to do all that type splitting on primitive or var vs. fvar.

idt that's necessary right now. lemme leave this open but consider this optional unless I have time to sit down and do enough tests to have some declarative statement about performance

SteveBronder:
^word! Yeah consider all that optional. I can also goof around and make a branch with those things to see how they'd look

stan/math/prim/core/complex_base.hpp
SteveBronder: > Most modern processors have some KB size of L1 cache

Also relevant to this, if the machine can make those copies in cache then it can cost a few less cycles, Dan Lemire has a [blog post](https://lemire.me/blog/2019/09/05/passing-integers-by-reference-can-be-expensive/) with an example of that but I'm not sure if the example he uses matches up here (or generally)

stan/math/fwd/core/std_complex.hpp
SteveBronder: It is definitely redundant, though if I use a pattern I like keeping it standard instead of 'ish. But whether a strict CRTP pattern here would be better is totally subjective. 

> then I'd just have to go fishing the T out with traits again.

Would argue that type traits change here would be fishing in a barrel

```cpp
  using value_type = value_type_t<V>;

  /**
   * Derived complex type used for function return types.
   */
  using complex_type = V;
```

stan/math/prim/meta/scalar_type.hpp
```diff
@@ -53,11 +53,22 @@ struct scalar_type<T, std::enable_if_t<is_std_vector<T>::value>> {
...
 };
+
+/** \ingroup type_trait
+ * Template metaprogram defining the base scalar type of
+ * values stored in a complex number.
+ *
+ * @tparam T value type of complex number
+ */
+template <typename T>
+struct scalar_type<std::complex<T>> {
```
bbbales2: This could be a bit weird. The complex number is the scalar type. I feel like the way this is written it should be called autodiff_scalar_type.

return_type leans on scalar_type (https://github.com/stan-dev/math/blob/ed61d817eb71f36fa2cd35f75b7c65c3135a817a/stan/math/prim/meta/return_type.hpp#L33), and so ```return_type<std::complex<var>, double>``` would be ```var```, which isn't what I'd expect mathematically.

bob-carpenter:
> On Feb 19, 2020, at 3:43 PM, Steve Bronder <notifications@github.com> wrote:
> 
> @SteveBronder commented on this pull request.
> 
> In stan/math/fwd/core/std_complex.hpp:
> 
> > +  template <typename U>
> +  complex(const U& x) : base_t(x) {}  // NOLINT(runtime/explicit)
> 
> It won't work if it's marked explicit. We need to be able to promote arithmetic types to complex if necessary.
> 
> As is yes but if you added the operator= for scalar types it would work right?

That won't participate in promotion for function calls.  But maybe we could still get away without it.  I'm a bit reluctant given that the std::complex base class uses an implicit constructor:

https://en.cppreference.com/w/cpp/numeric/complex/complex

stan/math/prim/core/complex_base.hpp
bob-carpenter: If we can see this in profiling, then it's definitely time to change the thousands of `const var&`  and `const fvar<T>&` args everywhere.

stan/math/prim/core/complex_base.hpp
```diff
@@ -0,0 +1,275 @@
...
+   * @return the real part
+   */
+  value_type real() const { return re_; }
+
+  /**
+   * Set the real part to the specified value.
+   *
+   * @param[in] x the value to set the real part to
+   */
+  void real(const value_type& x) { re_ = x; }
```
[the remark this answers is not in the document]
bob-carpenter: Yes, I can generalize that and avoid some casting to the `value_type`.

test/unit/math/mix/core/std_complex_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+  EXPECT_EQ(a, z.real());
+  EXPECT_EQ(0, z.imag());
+}
+
+template <typename T1, typename T2, typename T>
+void test_binary_constructor(const T1& x, const T2& y) {
+  using stan::math::value_of_rec;
+  std::complex<T> z(x, y);
+  EXPECT_EQ(1.1, value_of_rec(z.real()));
+  EXPECT_EQ(2.3, value_of_rec(z.imag()));
```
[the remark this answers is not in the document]
bob-carpenter: Will do.

bob-carpenter:
Looks like we're going to put this PR on hold yet again while I rewrite the meta libs.  Then I have to come backand fix this.  So please don't merge this.

stan/math/prim/core/complex_base.hpp
SteveBronder: I personally like `auto` so that type deduction just rides through instead of performing an implicit cast if there is a mismatch. I'm fine either way here

stan/math/prim/core/complex_base.hpp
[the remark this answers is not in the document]
SteveBronder: I'm _okay_ with it as is, but imo I'd rather just have a CRTP than CRTP-like.

stan/math/fwd/core/std_complex.hpp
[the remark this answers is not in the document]
SteveBronder: I need to look around, somewhere there is a rule set of which constructors getting created for free when you define other ones. Here I just meant

`~complex() = default;`

bbbales2:
This doesn't need put on hold. It's the next pull that has all the complexity, right? And even then we can redo the metaprograms afterwards. It doesn't change anything.

The only thing I'd say is this function:

```
void real(const value_type& x)
```

should accept more than a value type since apparently that makes sense and then this goes in.

Like:
```
std::complex<var> x;

x.real(5.0)
```

Should work and just cast that double to a var. If this already happens and I'm just not understanding the C++, lemme know and I'll approve (and merge after tests pass).

bob-carpenter:
From @SteveBronder:

> I'm okay with it as is, but imo I'd rather just have a CRTP than CRTP-like.

The path of least resistance is to change the code so I can get this thing out before more people pile on with change requests.  I'll do this:

* in the constructor of `T`, I'll have everywhere I have `T` as a template parameter to the base class, use `U = std::complex<T>` instead

* defining `using T = U::value_type;` right up front in the class because it's the `T` I actually need everywhere in the class

SteveBronder:
@mitzimorris it looks like this is failing because Jenkins can't find stanc in cmdstan? Do you know what's up with that?

https://jenkins.mc-stan.org/job/CmdStan/job/downstream_tests/1453/console

```
--- Translating Stan model to C++ code ---
bin/stanc  --o=src/test/test-models/transformed_data_rng_test.hpp src/test/test-models/transformed_data_rng_test.stan
--- Translating Stan model to C++ code ---
bin/stanc  --o=src/test/test-models/print_uninitialized.hpp src/test/test-models/print_uninitialized.stan

bin/stanc: 1: bin/stanc: Not: not found
make/program:42: recipe for target 'src/test/test-models/transformed_data_rng_test.hpp' failed
```

serban-nicusor-toptal:
@SteveBronder noticed a bit earlier too, on it

SteveBronder:
@serban-nicusor-catena much appreciated!

bob-carpenter:
@bbbales2 I'm afraid we need another review for this one after changes.  There's not a huge rush here because I'm working on a branch that has already merged this.

stan/math/prim/core/complex_base.hpp
SteveBronder: Had some time to sit down today and workout a benchmark for this. The file [here](https://github.com/seantalts/perf-math/blob/test/eigen_pf/constref_move_copy_var.cpp) is a google benchmark that tries to just isolate the cost of moving by const ref, perfect forwarding, and by a straight copy. The `noinline` and `no_icf` in the attributes is to mock a large function that wouldn't be inlined and to stop any of the functions in there from getting merged together during the Identical Code Folding stage of most compilers.

Do you think the above example code fairly estimates the effect of passing arguments by certain types? Open to changes! Below is the results from running 

```bash
make constref_move_copy_var.o && make constref_move_copy_var
./constref_move_copy_var --benchmark_repetitions=50 --benchmark_report_aggregates_only=true
```

The `benchmark_repetitions` flag means to run the entire benchmark process N times. A single benchmark looks like this

```
----------------------------------------------------------
Benchmark                Time             CPU   Iterations
----------------------------------------------------------
add_const_bench       14.7 ns         14.7 ns     42042642
add_pf_bench          12.1 ns         12.1 ns    130300991
add_copy_bench        5.40 ns         5.40 ns    128300668
```

Where Benchmark is each individual test, time wall time, cpu is cpu time, and iterations is the number of iterations the benchmark was run. The `benchmark_repititions` flag does the above N times then takes the mean, median, and stddev

```
2020-03-03 12:12:29
Running ./constref_move_copy_var
Run on (32 X 3500 MHz CPU s)
CPU Caches:
  L1 Data 32K (x16)
  L1 Instruction 64K (x16)
  L2 Unified 512K (x16)
  L3 Unified 8192K (x4)
Load Average: 0.50, 0.42, 0.28
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
add_const_bench_mean         5.57 ns         5.57 ns           50
add_const_bench_median       5.37 ns         5.37 ns           50
add_const_bench_stddev       1.31 ns         1.31 ns           50
add_pf_bench_mean            5.57 ns         5.57 ns           50
add_pf_bench_median          5.43 ns         5.43 ns           50
add_pf_bench_stddev         0.954 ns        0.954 ns           50
add_copy_bench_mean          5.31 ns         5.31 ns           50
add_copy_bench_median        5.25 ns         5.25 ns           50
add_copy_bench_stddev       0.359 ns        0.358 ns           50
```

Suprising! Doing values by copy seems to be about 4% faster! tbh I'm kind of suprised by that result, if you can find anything in the benchmark that makes the above seem fishy I'm open to any changes.

I haven't done a fresh install in a while but I think you should be able to copy and paste the below to run the benchmark

```bash
git clone --recursive https://github.com/seantalts/perf-math.git
git checkout eigen_pf
cd perf-math/benchmark
git clone https://github.com/google/googletest
mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=RELEASE
make
cd ..
make constref_move_copy_var.o && make constref_move_copy_var
./constref_move_copy_var --benchmark_repetitions=50 --benchmark_report_aggregates_only=true
```

SteveBronder: (also fyi this is just a side convo if we change anything because of the above we should do it in a separate PR)

bob-carpenter: Can you please say what's being tested here @SteveBronder.

And can we move this over to Discourse rather than having this long discussion via a pending issue?

SteveBronder: Yeah sure I'll close this and do something on discourse

bob-carpenter:
Just the constructor?  For what type U?  And what are you comparing it to?

stan/math/prim/core/complex_base.hpp
```diff
@@ -2,270 +2,232 @@
...
-   */
-  V im_;
-
  public:
   /**
-   * Type of real and imaginary parts.
+   * Type of real and imaginary parts
    */
-  using value_type = V;
+  using value_type = ValueType;
```
bbbales2: Is our standard to do underscore_naming for usings and CamelCase for template parameters?

bbbales2: That's consistent with how I understand the code but it's weird to see if directly in a situation like this :/.

bob-carpenter:
The using descriptors should be underscore separated.  The template
parameters should be capitalized.

The C++ standard library uses single capital letters for template
parameters and underscore separated lowercase for everything else.  I've
always tried to follow that.

More recently, Stan devs have started using more descriptive template
parameter names which will indeed call for camel case.  The standard
library doc just uses `T` for the value type of a complex number, not
`ValueType`.

PR #1722, opened 2020-02-17
remove discontinuity in lub_constrain; fixes #1441

### LINKED ISSUE #1441: lub_constrain boundary behaviour
## Description

Remove the conditional

```
 if ((x > NEGATIVE_INFTY) && (inv_logit_x == 0))
    inv_logit_x = 1e-15;
```

from the definition of `lub_constrain()` in `stan/math/prim/scal/fun/lub_constrain.hpp`.

### Background

The parameter constraining transform `lub_constrain()` has an code path with a comment
```cpp
// Prevent x from reaching zero unless it really really should.
```
The limiting value `1e-15` chosen is so large as to make the transform non-monotone. The branch also cannot prevent rounding when the lower bound is not zero. e.g. `<lower=100,upper=101>` does not see any effect, other than it stopping derivative propagation.
Similar code is _not_ present in `lb_constrain()` which is allowed to round to the boundary value.

A related question is what to do about the unconstraining transform `lub_free()`. The `check_bounded()` is inclusive so values at boundaries are accepted but the logit function maps boundary to infinity so it will fail later. A reasonable choice might be to map the boundary values to some finite number beyond the rounding limit. However, the rounding limit depends on the bounds in a complex fashion.

#### Current Version:
v3.0.0

pgree:
## Summary

Fixed a discontinuity in `lub_constrain`
Generalized the `lp` template parameters in `{lub,lb,ub,identity}_constrain`

## Tests

Added test of boundary conditions in `lub_transform_test.cpp`

## Side Effects

Now lub_constrain underflows instead of having a discontinuity

## Checklist

- [x] Math issue #1441

### DIFF
--- a/stan/math/prim/fun/identity_constrain.hpp
+++ b/stan/math/prim/fun/identity_constrain.hpp
@@ -30,12 +30,13 @@ inline T identity_constrain(const T& x) {
  * <p>This method is effectively a no-op and mainly useful as a
  * placeholder in auto-generated code.
  *
- * @tparam T type of scalar.
+ * @tparam T type of scalar
+ * @tparam S type of log probability
  * @param[in] x scalar
  * @return transformed input
  */
-template <typename T>
-inline T identity_constrain(const T& x, T& /*lp*/) {
+template <typename T, typename S>
+inline T identity_constrain(const T& x, S& lp) {
   return x;
 }
 
--- a/stan/math/prim/fun/lub_constrain.hpp
+++ b/stan/math/prim/fun/lub_constrain.hpp
@@ -46,7 +46,6 @@ namespace math {
 template <typename T, typename L, typename U>
 inline return_type_t<T, L, U> lub_constrain(const T& x, const L& lb,
                                             const U& ub) {
-  using std::exp;
   check_less("lub_constrain", "lb", lb, ub);
   if (lb == NEGATIVE_INFTY) {
     return ub_constrain(x, ub);
@@ -54,29 +53,14 @@ inline return_type_t<T, L, U> lub_constrain(const T& x, const L& lb,
   if (ub == INFTY) {
     return lb_constrain(x, lb);
   }
-
-  T inv_logit_x;
-  if (x > 0) {
-    inv_logit_x = inv_logit(x);
-    // Prevent x from reaching one unless it really really should.
-    if (x < INFTY && inv_logit_x == 1) {
-      inv_logit_x = 1 - 1e-15;
-    }
-  } else {
-    inv_logit_x = inv_logit(x);
-    // Prevent x from reaching zero unless it really really should.
-    if (x > NEGATIVE_INFTY && inv_logit_x == 0) {
-      inv_logit_x = 1e-15;
-    }
-  }
-  return fma(ub - lb, inv_logit_x, lb);
+  return fma(ub - lb, inv_logit(x), lb);
 }
 
 /**
  * Return the lower- and upper-bounded scalar derived by
  * transforming the specified free scalar given the specified
  * lower and upper bounds and increment the specified log
- * probability with the log absolute Jacobian determinant.
+ * density with the log absolute Jacobian determinant.
  *
  * <p>The transform is as defined in
  * <code>lub_constrain(T, double, double)</code>.  The log absolute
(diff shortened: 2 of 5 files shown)


stan/math/prim/fun/identity_constrain.hpp
```diff
@@ -30,12 +30,13 @@ inline T identity_constrain(const T& x) {
...
- * @tparam T type of scalar.
+ * @tparam T type of scalar
+ * @tparam S type of log probability
  * @param[in] x scalar
  * @return transformed input
  */
-template <typename T>
-inline T identity_constrain(const T& x, T& /*lp*/) {
+template <typename T, typename S>
+inline T identity_constrain(const T& x, S& lp) {
```
mcol: This should still be `/*lp*/` as the `lp` variable is unused in the function body and it may produce a warning.

stan/math/prim/fun/lub_constrain.hpp
```diff
@@ -130,18 +114,10 @@ inline return_type_t<T, L, U> lub_constrain(const T& x, const L& lb,
...
   } else {
     T exp_x = exp(x);
     inv_logit_x = inv_logit(x);
     lp += log(ub - lb) + x - 2 * log1p(exp_x);
-    // Prevent x from reaching zero unless it really really should.
-    if (x > NEGATIVE_INFTY && inv_logit_x == 0) {
-      inv_logit_x = 1e-15;
-    }
   }
   return fma(ub - lb, inv_logit_x, lb);
```
mcol: I know this is preexisting, but the computation of `inv_logit(x)` could be done in just one place (such as in the return line) rather that in both branches of the `if`.

stan/math/prim/fun/identity_constrain.hpp
mcol: We already have a version without the `lp` argument. This other one is for symmetry with the other constraint functions which have that argument and pass it to `identity_constraint`, even if it's not used. I don't feel I understand this stuff well enough to depart from the existing signatures.

charlesm93: Ok. I don't think symmetry alone is a good enough reason to keep a piece of code. @bob-carpenter, I think you understand this better than I do -- is there any good reason for overloading `identity_constraint` here?

bob-carpenter: > This should still be /*lp*/ as the lp variable is unused in the function body

If the types get relaxed, we need it for type inference.  That's done statically, so there's no runtime overhead.

> I don't feel I understand this stuff well enough to depart from the existing signatures.

I built the constrain/unconstrain idiom for doing this when we first built the language.  That was very early on in my understanding of C++ and we were having a lot of trouble getting the simplest things to compile.

The current change strictly generalizes the set of instantiations allowed.  In Stan as currently deployed, all arguments will be at least type `double`.  

> is there any good reason for overloading identity_constraint here?

The goal is to relax the type instantiations here to allow for `int` arguments to constraints with `double` or `var` or `fvar` arguments to `lp`.  That will come up if these ever get exposed to the language, as we've been proposing to do forever. 

As is, we can't write tests with `int` arguments along with `double` arguments for `lp`.  

I opened issue https://github.com/stan-dev/math/issues/1733 to request a universal fix.

Whoever's reviewing this needs to decide if it's OK to have a partial solution.  I've been of a fix-it-as-you-go mindset with this kind of stuff, but I can see the argument for consistency across constraints and it's up to whoever is reviewing to decide.

Thanks for looking at this so closely, by the way!

stan/math/prim/fun/lub_constrain.hpp
bob-carpenter: It needs to be pulled up above the conditional because the value is used in the condition.

bob-carpenter:
There were no replies to my earlier comments, so I suggested to @pgree that he could finish this by:

1. cleaning up the duplicated `inv_logit` call, and

2. leaving the generalized signatures with arguments that are now needed for type inference.

If that's not OK, could you please clarify what you are requesting to change, @charlesm93?

charlesm93:
@bob-carpenter just coming back from a trip, will respond to your comments tomorrow morning.

stan/math/prim/fun/identity_constrain.hpp
charlesm93: I'm ok with the the fix-it-as-you-go approach, especially since this PR is straightforward and we have an issue for the universal fix.

stan/math/prim/fun/lub_constrain.hpp
charlesm93: Ok if we implement @bob-carpenter 's suggestion.

pgree:
@charlesm93 made the requested changes, ready for review

charlesm93:
Looks good to me. If the continuous integration passes, we can merge.

charlesm93:
@pgree, I'll leave you the honor to press the "merge pull request" button. Welcome to the dev team!

bob-carpenter:
I don't think he has permission to merge yet.

> @pgree, I'll leave you the honor to press the "merge pull request" button.

Issue #1723, opened 2020-02-18
Elementwise functions & Eigen::Map


andrjohns:
## Description
Now that the ```apply_vector_unary``` framework has gone through, I want to use it to extend the elementwise functions(e.g. ```exp```, ```acos```). At the moment, Eigen's SIMD-optimised functions are only used for ```Eigen::MatrixBase``` objects (e.g. ```Eigen::Matrix``` and ```Eigen::Map```), not for ```std::vectors``` (containing either scalars or Eigen objects) or for ```Eigen::Array``` objects (unless an additional specialisation is added, as was done for some functions). This leaves a bit of speed on the table, as these objects then get called with the less optimised versions of the function.

Extending these functions to be agnostic between```std::vector``` and ```Eigen::Array``` is simple with the ```apply_vector_unary``` approach. The difficulty comes in the return type. Because the rest of the Stan library is not yet able to take Eigen expressions as inputs, the function needs to be ```.eval```-ed to return either an ```Eigen::Matrix``` or an ```Eigen::Array```, depending on the input.

From what I can see, there are two ways of achieving this. We could have a different specialisation for ```MatrixBase``` and ```ArrayBase```, however this means the entire function would be duplicated, with the only difference being the return. 

Using ```exp``` as an example:
```
template <typename Derived,
          typename = require_eigen_vt<std::is_arithmetic, Derived>>
inline auto exp(const Eigen::MatrixBase<Derived>& x) {
  return x.derived().array().exp().matrix().eval();
}

template <typename Derived,
          typename = require_eigen_vt<std::is_arithmetic, Derived>>
inline auto exp(const Eigen::ArrayBase<Derived>& x) {
  return x.derived().array().exp().eval();
}
```

Alternatively, we could use a helper function that uses the input type to determine the right wrapper to call (```.matrix()``` or ```.array()```) prior to the call to ```.eval()```. As an initial prototype, I've created the function ```match_wrapper<T>()``` which will call the ```.array()``` wrapper for any ```Eigen::Array``` objects, and the ```.matrix()``` wrapper otherwise.

Using ```exp``` as an example, the following function will use Eigen's ```exp``` function for all containers (and containers of containers) and return either an ```Eigen::Matrix``` or ```Eigen::Array``` (or container of these), depending on the input type:
```
template <typename T, require_container_st<is_container, std::is_arithmetic, T>>
inline auto exp(const T& x) {
  return apply_vector_unary<T>::apply(x, [&](const auto& v) {
[...]

PR #1727, opened 2020-02-20
Eigen::Map and elementwise functions

andrjohns:
## Summary

This pull extends the existing elementwise math functions to use Eigen's optimised functions for ```std::vector```, ```Eigen::Array```, and nested containers of these.

As part of this, the ```apply_vector_unary``` framework was modified to ```.eval()``` Eigen returns (rather than doing so in the individual function headers) and to deduce the return scalar types (for when integers are promoted to doubles).

## Tests
Optimising existing behaviour, so no new tests added

## Side Effects

N/A

## Checklist

- [x] Math issue #1723 


SteveBronder:
Is this ready for review?

andrjohns:
Just fixed up the merge conflicts, ready for review. Thanks!

stan/math/prim/fun/acos.hpp
```diff
@@ -30,21 +30,24 @@ struct acos_fun {
  * @param x container
  * @return Arc cosine of each variable in the container, in radians.
  */
-template <typename T, typename = require_not_eigen_vt<std::is_arithmetic, T>>
+template <typename T,
+          require_not_container_st<is_container, std::is_arithmetic, T>...>
```
t4c1: `require_ ... _st<is_container, ...>` does not make sense - scalar type of something is never a container. Did you mean to use `require_not_container_vt`? Even with this change I don't understand. Actually I think even the previous requirement was wrong - it should be `require_arithmetic_vt<T>` and `require_not_eigen_t<T>` separately.

Is this now only used for scalars? Than you can use `require_arithmetic_t<T>`. In this case also `apply_scalar_unary` can be removed.

stan/math/prim/fun/acos.hpp
```diff
@@ -30,21 +30,24 @@ struct acos_fun {
...
+ * @tparam T Type of x
+ * @param x Eigen Matrix/Array or expression
  * @return Arc cosine of each variable in the container, in radians.
  */
-template <typename Derived,
-          typename = require_eigen_vt<std::is_arithmetic, Derived>>
-inline auto acos(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().acos().matrix().eval();
+template <typename T,
+          require_container_st<is_container, std::is_arithmetic, T>...>
```
t4c1: Another `require_..._st<is_container, ...>`. Same in other files.

stan/math/prim/fun/acos.hpp
```diff
@@ -30,21 +30,24 @@ struct acos_fun {
...
+          require_not_container_st<is_container, std::is_arithmetic, T>...>
 inline auto acos(const T& x) {
   return apply_scalar_unary<acos_fun, T>::apply(x);
 }
 
 /**
- * Version of acos() that accepts Eigen Matrix or matrix expressions.
- * @tparam Derived derived type of x
- * @param x Matrix or matrix expression
+ * Version of acos() that accepts Eigen Matrix/Array objects or expressions.
```
t4c1: I think this now also accepts `std::vector`s.

stan/math/prim/fun/inv_square.hpp
```diff
@@ -18,11 +18,27 @@ inline double inv_square(double x) { return inv(square(x)); }
...
+ * @param x Eigen Matrix/Array or expression
+ * @return 1 / the square of each value in x.
+ */
+template <typename T,
+          require_container_st<is_container, std::is_arithmetic, T>...>
+inline auto inv_square(const T& x) {
+  return apply_vector_unary<T>::apply(
+      x, [&](const auto& v) { return v.derived().array().square().inverse(); });
+}
+
```
t4c1: This one is not needed as both `inv` and `square` use `apply_vector_unary`. So the simple implementation of `inv_square` already works for all input types.

stan/math/prim/fun/log_softmax.hpp
```diff
@@ -41,7 +41,7 @@ template <typename T, require_t<std::is_arithmetic<scalar_type_t<T>>>...>
 inline auto log_softmax(const T& x) {
   return apply_vector_unary<T>::apply(x, [&](const auto& v) {
     check_nonzero_size("log_softmax", "v", v);
-    return (v.array() - log_sum_exp(v)).matrix().eval();
+    return (v.array() - log_sum_exp(v));
```
t4c1: This one should use `Eigen::Ref` so as not to evaluate ˙`v` twice.

stan/math/prim/fun/logit.hpp
```diff
@@ -80,11 +80,27 @@ struct logit_fun {
...
+ *
+ * @tparam T Type of x
+ * @param x Eigen Matrix/Array or expression
+ * @return the logit of each variable in the container.
+ */
+template <typename T,
+          require_container_st<is_container, std::is_arithmetic, T>...>
+inline auto logit(const T& x) {
+  return apply_vector_unary<T>::apply(x, [&](const auto& v) {
+    return (v.derived().array() / (1 - v.derived().array())).log();
```
t4c1: This one should use `Eigen::Ref` so as not to evaluate ˙`v` twice.

stan/math/prim/meta/apply_vector_unary.hpp
```diff
@@ -82,10 +96,12 @@ struct apply_vector_unary<T, require_std_vector_vt<is_stan_scalar, T>> {
    * @return std::vector with result of applying functor to input.
    */
   template <typename F>
-  static inline std::vector<T_vt> apply(const T& x, const F& f) {
-    std::vector<T_vt> result(x.size());
-    Eigen::Map<Eigen::Matrix<T_vt, -1, 1>>(result.data(), result.size())
-        = apply_vector_unary<T_map>::apply(as_column_vector_or_scalar(x), f);
+  static inline auto apply(const T& x, const F& f) {
+    using T_return = typename plain_type_t<decltype(
+        f(as_column_vector_or_scalar(x)))>::Scalar;
```
t4c1: This can be simplified to: 
```
    using T_return = value_type_t<decltype(f(as_column_vector_or_scalar(x)))>;
```

stan/math/prim/meta/require_generics.hpp
```diff
@@ -840,6 +840,11 @@ template <template <class...> class ContainerCheck,
 using require_container_st = require_t<
     container_scalar_type_check_base<ContainerCheck, TypeCheck, Check...>>;
 
+template <template <class...> class ContainerCheck,
+          template <class...> class TypeCheck, class... Check>
+using require_not_container_st = require_not_t<
+    container_scalar_type_check_base<ContainerCheck, TypeCheck, Check...>>;
+
```
t4c1: This should have a test or two.

stan/math/prim/fun/acos.hpp
```diff
@@ -30,21 +30,24 @@ struct acos_fun {
...
  */
-template <typename Derived,
-          typename = require_eigen_vt<std::is_arithmetic, Derived>>
-inline auto acos(const Eigen::MatrixBase<Derived>& x) {
-  return x.derived().array().acos().matrix().eval();
+template <typename T,
+          require_container_st<is_container, std::is_arithmetic, T>...>
+inline auto acos(const T& x) {
+  return apply_vector_unary<T>::apply(
+      x, [&](const auto& v) { return v.derived().array().acos(); });
```
t4c1: Thanks for simplifying `Eigen::MatrixBase<Derived>`. With this change `.derived()` call is also not needed anymore.

stan/math/prim/fun/acos.hpp
andrjohns: With ```require_container_st<is_container, std::is_arithmetic, T>```, it checks whether ```T``` is a container (either ```std::vector``` or ```Eigen```) whose base scalar type is ```arithmetic```. For example, this would capture the case where ```T``` is an ```std::vector<Eigen::MatrixXd>```.

The ```require_not_container_st``` is required because ```apply_scalar_unary``` handles all other types (scalar and containers of autodiff types)

stan/math/prim/fun/inv_square.hpp
andrjohns: Good call, will revert

andrjohns: Actually, I feel like this would be more efficient. Since with ```inv(square(x))```, wouldn't the matrix ```x``` would get traversed twice? Once for ```square``` and once for ```inv```?

stan/math/prim/fun/acos.hpp
t4c1: I went over the code of `require_not_container_st`. Now this makes a bit more sense. I was confused by name, as what it does is not consistent with other requires.

stan/math/prim/meta/require_generics.hpp
t4c1: Also what this does is not consistent with other requires that just accept scalar check as a parameter. To be consistent with other requires I would suggest to change it to:
```
template <template <class...> class TypeCheck, class... Check>
using require_not_container_st = require_not_t<
    container_scalar_type_check_base<is_container, TypeCheck, Check...>>;
```

stan/math/prim/fun/inv_square.hpp
t4c1: Yes, but just until we remove all those `eval()`s. I am leaving it up to you if you want to leave it in.

stan/math/prim/fun/logit.hpp
SteveBronder: Does this need & in the lambda? And why `const auto&` We don't reuse v in the lambda so I think we could just  `auto&&`

stan/math/prim/fun/acos.hpp
SteveBronder: [optional] I like having more helpful names than `T` for templates when we do specializations like this. Though that may not be a thing here since the only name I can think of is `Container`

stan/math/prim/fun/inv_square.hpp
andrjohns: That's the case for Eigen objects, but not for ```std::vector<double>``` or ```std::vectors``` of Eigen objects, since they'll get evaluated and returned between each call

stan/math/prim/meta/require_generics.hpp
andrjohns: It needs to be written this way to be consistent with ```require_container_st```, which allows for the specification of the container type (e.g., ```require_container_st<is_std_vector, is_integral, T>```)

andrjohns:
@t4c1 this is ready for re-review

stan/math/prim/fun/logit.hpp
t4c1: @andrjohns Sorry, I overlooked this reply. I disagree with @SteveBronder here. In my opinion we should always pass arguments as `const&` if possible, to make it obvious the function does not change them. `auto&&` or any other forwarding reference should only be used when we actually need perfect forwarding.

stan/math/prim/meta/require_generics.hpp
t4c1: [optional] I see. I still think we should make a separate type trait for `require_container_st<is_container, ...>`. Though to be honest I can't think of a good name for it.

Issue #912, opened 2018-06-22
Beta Binomial CDF gives non-monotone results


wds15:
#### Summary:
CDF functions should always return monotone output with increasing input value. The BetaBinomial CDF does violate this.

#### Description:
See Stan program below which fails under cmdstan 2.17.1 and develop (as of 22nd June 18).

#### Reproducible Steps:

```stan
parameters {
  real<lower=0,upper=1> dummy;
}
generated quantities {
  real m = log(100);
  real s= log(100);
  real a = lognormal_rng(m, s);
  real b = lognormal_rng(m, s);
  int N = poisson_log_rng(normal_rng(m, s));
  
  for(n in 0:(N-1)) {
    real cdfn = beta_binomial_lcdf(n| N, a, b);
    real cdfn1 = beta_binomial_lcdf(n+1| N, a, b);
    if(cdfn1 < cdfn) {
      print("n = ", n);
      print("N = ", N);
      print("a = ", a);
      print("b = ", b);
      reject("Investigate me please!");
    }
  }
}
```

#### Current Output:
The current output. Knowing what is the current behavior is useful.

#### Expected Output:
Describe what you expect the output to be. Knowing the correct behavior is also very useful.

#### Additional Information:
Provide any additional information here.

#### Current Version:
v2.17.0

bob-carpenter:
This should be tackled at the same time as #688

mcol:
The problem here seems to be more complex than the one in #688. I don't fully understand it yet, but running @wds15's model led me to a set values that demonstrate the problem, and after exposing the beta-binomial lcdf function in R via rstan, I got the following:
```
> betabinomial(2, 10, 148036, 1.98592)
[1] -21.62329
> betabinomial(3, 10, 148036, 1.98592)
[1] NaN
```

With respect to these lines: https://github.com/stan-dev/math/blob/490c18d18d349a27dc782bf80e059ca1f11fc524/stan/math/prim/scal/prob/beta_binomial_lcdf.hpp#L105:L107, the `NaN` case happens because `C = 1.0000000000204339` so that `Pi = -0.0000000000204339`, and taking the log fails.

My current feeling is that there is some loss of precision somewhere, and `C` goes above 1 when it should have stayed strictly below that (at least the code assumes that).

nhuurre:
The `_lccdf` computation is identical to `_lcdf` except that `Pi = C` instead of `Pi = 1-C`.
[...]

PR #1728, opened 2020-02-20
Keep computations in log space and simplify expressions in beta binomial_*cdf

mcol:
## Summary
In #912 @nhuurre noticed that computations for the `beta_binomial` *cdf functions switch from log space to linear space halfway through. He also noticed that there is no gain in using the `lgamma` instead of `lbeta`, as we are effectively still computing all terms: switching to `lbeta` makes the code a bit more readable. This also does a bit of general cleanup to save on some intermediate computations.

This doesn't solve the issue of loss of precision, but will make it easier to investigate that going forward.

## Tests

No new tests, this is just cleanup.

## Side Effects

None.

## Checklist

- [X] Math issue #912

### DIFF
--- a/stan/math/prim/prob/beta_binomial_cdf.hpp
+++ b/stan/math/prim/prob/beta_binomial_cdf.hpp
@@ -9,7 +9,7 @@
 #include <stan/math/prim/fun/exp.hpp>
 #include <stan/math/prim/fun/F32.hpp>
 #include <stan/math/prim/fun/grad_F32.hpp>
-#include <stan/math/prim/fun/lgamma.hpp>
+#include <stan/math/prim/fun/lbeta.hpp>
 #include <stan/math/prim/fun/max_size.hpp>
 #include <stan/math/prim/fun/size.hpp>
 #include <stan/math/prim/fun/size_zero.hpp>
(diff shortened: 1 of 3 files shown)


stan/math/prim/prob/beta_binomial_cdf.hpp
```diff
@@ -87,46 +87,37 @@ return_type_t<T_size1, T_size2> beta_binomial_cdf(const T_n& n, const T_N& N,
...
+    C = F * exp(C) / (N_dbl + 1);
 
     const T_partials_return Pi = 1 - C;
 
     P *= Pi;
 
     T_partials_return dF[6];
-    T_partials_return digammaOne = 0;
-    T_partials_return digammaTwo = 0;
+    T_partials_return digammaDiff = 0;
```
bob-carpenter: [optional]

This pattern of setting and then resetting is something we want to avoid for both clarity and efficiency.  As is, it gets set to 0 then reset if size1 or size2 is not constant.  The setting allocates memory, which resetting throws away without recovering.

A better pattern for this is:

```
T_partials_return digammaDiff
  = is_constant_all<T_size1, T_size2>::value ? 0
                                             : digamma(mu + nu) - digamma(alpha_dbl + beta_dbl);
```

It's not incorrect as written and the inefficiency is minor, so I'm marking this optional.

bob-carpenter: [optional]

Rather than defining `digammaDiff = A - B` and then using `- digammaDiff`, it'd be more efficient to define `digammaDiff = B - A` and then use without the negation.

mcol: Thanks, for the time being I'm merging as is, I've added them to my other branch in which I'm cleaning up code like this, so they won't get lost.

PR #1732, opened 2020-02-21
Generalize */fun starting with a-chol2

t4c1:
## Summary

Generalizes inputs to functions that start with a - chol2. `chol2inv` depends on `mdivide_left`, `mdivide_left_tri` and `mdivide_left_tri_low` so these were generalized as well.

## Tests
`assign()` had different requirements wrt compile time size for Eigen matrices and Eigen blocks. I changed that so it accepts all assignments Eigen does (It previously did that for blocks. For matrices it required same compile time sizes). That also changed some tests.

## Side Effects
None.

## Checklist

- [x] Math issue #1470

### DIFF
--- a/stan/math/fwd/fun/mdivide_left.hpp
+++ b/stan/math/fwd/fun/mdivide_left.hpp
@@ -54,7 +54,8 @@ mdivide_left(const T1& A, const T2& b) {
   return to_fvar(inv_A_mult_b, deriv);
 }
 
-template <typename T1, typename T2, require_eigen_vt<std::is_arithmetic, T1>* = nullptr,
+template <typename T1, typename T2,
+          require_eigen_vt<std::is_arithmetic, T1>* = nullptr,
           require_eigen_vt<is_fvar, T2>* = nullptr>
 inline Eigen::Matrix<value_type_t<T2>, T1::RowsAtCompileTime,
                      T2::ColsAtCompileTime>
--- a/stan/math/fwd/fun/mdivide_left.hpp
+++ b/stan/math/fwd/fun/mdivide_left.hpp
@@ -13,10 +13,18 @@
 namespace stan {
 namespace math {
 
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
-    const Eigen::Matrix<fvar<T>, R1, C1> &A,
-    const Eigen::Matrix<fvar<T>, R2, C2> &b) {
+template <typename T1, typename T2,
+          require_all_eigen_vt<is_fvar, T1, T2>* = nullptr,
+          require_same_vt<T1, T2>* = nullptr>
+inline Eigen::Matrix<value_type_t<T1>, T1::RowsAtCompileTime,
+                     T2::ColsAtCompileTime>
+mdivide_left(const T1& A, const T2& b) {
+  using T = typename value_type_t<T1>::Scalar;
+  constexpr int R1 = T1::RowsAtCompileTime;
+  constexpr int C1 = T1::ColsAtCompileTime;
+  constexpr int R2 = T2::RowsAtCompileTime;
+  constexpr int C2 = T2::ColsAtCompileTime;
+
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
 
@@ -28,17 +36,19 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   Eigen::Matrix<T, R2, C2> val_b(b.rows(), b.cols());
   Eigen::Matrix<T, R2, C2> deriv_b(b.rows(), b.cols());
 
+  const Eigen::Ref<const plain_type_t<T1>>& A_ref = A;
   for (int j = 0; j < A.cols(); j++) {
     for (int i = 0; i < A.rows(); i++) {
-      val_A(i, j) = A(i, j).val_;
-      deriv_A(i, j) = A(i, j).d_;
+      val_A(i, j) = A_ref(i, j).val_;
+      deriv_A(i, j) = A_ref(i, j).d_;
     }
   }
 
+  const Eigen::Ref<const plain_type_t<T2>>& b_ref = b;
   for (int j = 0; j < b.cols(); j++) {
     for (int i = 0; i < b.rows(); i++) {
-      val_b(i, j) = b(i, j).val_;
-      deriv_b(i, j) = b(i, j).d_;
+      val_b(i, j) = b_ref(i, j).val_;
+      deriv_b(i, j) = b_ref(i, j).d_;
     }
   }
 
@@ -52,30 +62,47 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   return to_fvar(inv_A_mult_b, deriv);
 }
 
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
-    const Eigen::Matrix<double, R1, C1> &A,
-    const Eigen::Matrix<fvar<T>, R2, C2> &b) {
+template <typename T1, typename T2, require_eigen_t<T1>* = nullptr,
+          require_same_vt<double, T1>* = nullptr,
+          require_eigen_vt<is_fvar, T2>* = nullptr>
+inline Eigen::Matrix<value_type_t<T2>, T1::RowsAtCompileTime,
+                     T2::ColsAtCompileTime>
+mdivide_left(const T1& A, const T2& b) {
+  using T = typename value_type_t<T2>::Scalar;
+  constexpr int R1 = T1::RowsAtCompileTime;
+  constexpr int C1 = T1::ColsAtCompileTime;
+  constexpr int R2 = T2::RowsAtCompileTime;
+  constexpr int C2 = T2::ColsAtCompileTime;
+
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
 
   Eigen::Matrix<T, R2, C2> val_b(b.rows(), b.cols());
   Eigen::Matrix<T, R2, C2> deriv_b(b.rows(), b.cols());
 
+  const Eigen::Ref<const plain_type_t<T2>>& b_ref = b;
   for (int j = 0; j < b.cols(); j++) {
     for (int i = 0; i < b.rows(); i++) {
-      val_b(i, j) = b(i, j).val_;
-      deriv_b(i, j) = b(i, j).d_;
+      val_b(i, j) = b_ref(i, j).val_;
+      deriv_b(i, j) = b_ref(i, j).d_;
     }
   }
 
   return to_fvar(mdivide_left(A, val_b), mdivide_left(A, deriv_b));
 }
 
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
-    const Eigen::Matrix<fvar<T>, R1, C1> &A,
-    const Eigen::Matrix<double, R2, C2> &b) {
+template <typename T1, typename T2, require_eigen_vt<is_fvar, T1>* = nullptr,
+          require_eigen_t<T2>* = nullptr,
+          require_same_vt<double, T2>* = nullptr>
+inline Eigen::Matrix<value_type_t<T1>, T1::RowsAtCompileTime,
+                     T2::ColsAtCompileTime>
+mdivide_left(const T1& A, const T2& b) {
+  using T = typename value_type_t<T1>::Scalar;
+  constexpr int R1 = T1::RowsAtCompileTime;
+  constexpr int C1 = T1::ColsAtCompileTime;
+  constexpr int R2 = T2::RowsAtCompileTime;
+  constexpr int C2 = T2::ColsAtCompileTime;
+
   check_square("mdivide_left", "A", A);
   check_multiplicable("mdivide_left", "A", A, "b", b);
 
@@ -84,10 +111,11 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   Eigen::Matrix<T, R1, C1> val_A(A.rows(), A.cols());
   Eigen::Matrix<T, R1, C1> deriv_A(A.rows(), A.cols());
 
+  const Eigen::Ref<const plain_type_t<T1>>& A_ref = A;
   for (int j = 0; j < A.cols(); j++) {
     for (int i = 0; i < A.rows(); i++) {
-      val_A(i, j) = A(i, j).val_;
-      deriv_A(i, j) = A(i, j).d_;
+      val_A(i, j) = A_ref(i, j).val_;
+      deriv_A(i, j) = A_ref(i, j).d_;
     }
   }
 
--- a/stan/math/fwd/fun/mdivide_left.hpp
+++ b/stan/math/fwd/fun/mdivide_left.hpp
@@ -20,9 +20,7 @@ inline Eigen::Matrix<value_type_t<T1>, T1::RowsAtCompileTime,
                      T2::ColsAtCompileTime>
 mdivide_left(const T1& A, const T2& b) {
   using T = typename value_type_t<T1>::Scalar;
-  constexpr int R1 = T1::RowsAtCompileTime;
-  constexpr int C1 = T1::ColsAtCompileTime;
-  constexpr int R2 = T2::RowsAtCompileTime;
+  constexpr int S1 = T1::RowsAtCompileTime;
   constexpr int C2 = T2::ColsAtCompileTime;
 
   check_square("mdivide_left", "A", A);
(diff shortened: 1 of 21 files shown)


stan/math/prim/fun/append_row.hpp
```diff
@@ -161,21 +55,21 @@ inline Eigen::Matrix<T, Eigen::Dynamic, 1> append_row(
...
-    const T1& A, const Eigen::Matrix<T2, R, C>& B) {
+template <typename Scal, typename ColVec,
+          require_stan_scalar_t<Scal>* = nullptr,
+          require_t<is_eigen_col_vector<ColVec>>* = nullptr>
+inline Eigen::Matrix<return_type_t<Scal, ColVec>, Eigen::Dynamic, 1> append_row(
+    const Scal& A, const ColVec& B) {
   using Eigen::Dynamic;
   using Eigen::Matrix;
-  using return_type = return_type_t<T1, T2>;
+  using return_type = return_type_t<Scal, ColVec>;
```
andrjohns: This should be ```T_return``` for consistency with the above functions

stan/math/prim/fun/chol2inv.hpp
```diff
@@ -20,22 +20,22 @@ namespace math {
...
  *  lower triangular
  */
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> chol2inv(
-    const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& L) {
-  check_square("chol2inv", "L", L);
-  check_lower_triangular("chol2inv", "L", L);
+template <typename T, require_eigen_t<T>* = nullptr>
+plain_type_t<T> chol2inv(const T& L) {
+  const Eigen::Ref<const plain_type_t<T>>& L_ref = L;
```
andrjohns: Same comment here about having ```Eigen::Ref``` in the input type

stan/math/prim/fun/chol2inv.hpp
```diff
@@ -20,22 +20,22 @@ namespace math {
...
+    return L_ref;
   }
   if (K == 1) {
     matrix_t X(1, 1);
-    X.coeffRef(0) = inv_square(L.coeff(0));
+    X.coeffRef(0) = inv_square(L_ref.coeff(0, 0));
     return X;
   }
-  matrix_t L_inv = mdivide_left_tri_low(L, matrix_t::Identity(K, K).eval());
+  matrix_t L_inv = mdivide_left_tri_low(L_ref, matrix_t::Identity(K, K).eval());
```
andrjohns: Probably don't need the ```.eval()``` here since the ```mdivide_*``` were generalised above

stan/math/prim/fun/chol2inv.hpp
```diff
@@ -20,22 +20,22 @@ namespace math {
...
-  check_square("chol2inv", "L", L);
-  check_lower_triangular("chol2inv", "L", L);
+template <typename T, require_eigen_t<T>* = nullptr>
+plain_type_t<T> chol2inv(const T& L) {
+  const Eigen::Ref<const plain_type_t<T>>& L_ref = L;
+  check_square("chol2inv", "L", L_ref);
+  check_lower_triangular("chol2inv", "L", L_ref);
   int K = L.rows();
-  using matrix_t = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>;
+  using matrix_t = plain_type_t<T>;
```
andrjohns: typedefs in the rest of the library are generally ```T_```

stan/math/prim/fun/mdivide_left_tri.hpp
```diff
@@ -16,57 +16,54 @@ namespace math {
...
   check_square("mdivide_left_tri", "A", A);
   if (A.rows() == 0) {
     return {};
   }
 
   int n = A.rows();
-  Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> b;
-  b.setIdentity(n, n);
+  plain_type_t<T> b = plain_type_t<T>::Identity(n, n);
   A.template triangularView<TriView>().solveInPlace(b);
```
andrjohns: Could probably have the ```Identity``` call within the ```solveInPlace``` to avoid a copy (depending on whether Eigen would optimise that, I guess)

stan/math/prim/fun/mdivide_left_tri.hpp
```diff
@@ -77,23 +74,20 @@ inline Eigen::Matrix<T, R1, C1> mdivide_left_tri(
...
  * @throws std::domain_error if A is not square or the rows of b don't
  * match the size of A.
  */
-template <Eigen::UpLoType TriView, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
-    const Eigen::Matrix<double, R1, C1> &A,
-    const Eigen::Matrix<double, R2, C2> &b) {
+template <Eigen::UpLoType TriView, typename T1, typename T2,
+          require_all_eigen_t<T1, T2> * = nullptr,
+          require_all_same_vt<double, T1, T2> * = nullptr>
```
andrjohns: Probably simpler to use ```require_all_eigen_vt``` here

stan/math/prim/fun/mdivide_left_tri.hpp
```diff
@@ -122,16 +116,15 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
...
  *
  * @param A Triangular matrix.
  * @return x = A^-1 .
  * @throws std::domain_error if A is not square
  */
-template <Eigen::UpLoType TriView, int R1, int C1>
-inline Eigen::Matrix<double, R1, C1> mdivide_left_tri(
-    const Eigen::Matrix<double, R1, C1> &A) {
+template <Eigen::UpLoType TriView, typename T, require_eigen_t<T> * = nullptr,
+          require_same_vt<double, T> * = nullptr>
```
andrjohns: Could use ```require_eigen_vt``` here

stan/math/prim/fun/mdivide_left_tri.hpp
```diff
@@ -122,16 +116,15 @@ inline Eigen::Matrix<double, R1, C2> mdivide_left_tri(
...
+inline plain_type_t<T> mdivide_left_tri(const T &A) {
```
andrjohns: Why the inconsistency between a ```Eigen::Matrix``` or ```plain_type_t``` return? Harder to deduce the return for mixed scalar types?

stan/math/rev/fun/mdivide_left.hpp
```diff
@@ -185,19 +186,22 @@ inline Eigen::Matrix<var, R1, C2> mdivide_left(
...
   return res;
 }
 
-template <int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<var, R1, C2> mdivide_left(
-    const Eigen::Matrix<var, R1, C1> &A,
-    const Eigen::Matrix<double, R2, C2> &b) {
+template <typename T1, typename T2, require_eigen_vt<is_var, T1> * = nullptr,
+          require_eigen_t<T2> * = nullptr,
+          require_same_vt<double, T2> * = nullptr>
```
andrjohns: ```require_eigen_vt``` here too

test/unit/math/fwd/fun/assign_test.cpp
```diff
@@ -73,9 +73,6 @@ TEST(AgradFwdMatrixAssign, eigen_row_vector_fvar_double_shape_mismatch) {
   Matrix<fvar<double>, Dynamic, Dynamic> zzz(3, 1);
   zzz << 1, 2, 3;
   EXPECT_THROW(assign(x, zzz), std::invalid_argument);
-
-  Matrix<fvar<double>, Dynamic, Dynamic> zzzz(1, 3);
-  EXPECT_THROW(assign(x, zzzz), std::invalid_argument);
```
andrjohns: This is a tricky one. Size-wise the two objects are the same, but one is declared as a matrix and one a vector. This means that throwing in the math library will be dimension-based, but throwing in the Stan language will be type-dependent (although I'm not sure if ```assign``` is even exposed?). Not sure if that's an important distinction to make somewhere.

stan/math/fwd/fun/mdivide_left.hpp
```diff
@@ -31,17 +39,19 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   Eigen::Matrix<T, R2, C2> val_b(b.rows(), b.cols());
   Eigen::Matrix<T, R2, C2> deriv_b(b.rows(), b.cols());
 
+  const Eigen::Ref<const plain_type_t<T1>>& A_ref = A;
   for (int j = 0; j < A.cols(); j++) {
     for (int i = 0; i < A.rows(); i++) {
-      val_A(i, j) = A(i, j).val_;
-      deriv_A(i, j) = A(i, j).d_;
+      val_A(i, j) = A_ref(i, j).val_;
```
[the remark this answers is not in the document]
t4c1: I am pretty sure if a do `val_A = A.val(); deriv_A = A.d()`, A would be evaluated twice. I think we can not avoid either a copy or two evaluations.

stan/math/prim/err/check_lower_triangular.hpp
```diff
@@ -23,13 +23,13 @@ namespace math {
...
  *   portion is NaN
  */
-template <typename T_y>
-inline void check_lower_triangular(
-    const char* function, const char* name,
-    const Eigen::Matrix<T_y, Eigen::Dynamic, Eigen::Dynamic>& y) {
+template <typename T_y, require_eigen_t<T_y>* = nullptr>
+inline void check_lower_triangular(const char* function, const char* name,
+                                   const T_y& y) {
+  const Eigen::Ref<const plain_type_t<T_y>>& y_ref = y;
```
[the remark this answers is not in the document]
t4c1: I think I made no changes to `throw_domain_error`. Anyway it just writes a message and throws. There is no iteration over Eigen objects in `throw_domain_error`.

stan/math/prim/fun/accumulator.hpp
```diff
@@ -73,15 +73,14 @@ class accumulator {
...
    * @param m Matrix of values to add
    */
-  template <typename S, int R, int C>
-  void add(const Eigen::Matrix<S, R, C>& m) {
+  template <typename S, require_eigen_t<S>* = nullptr>
+  void add(const S& m) {
+    const auto& m_eval = m.eval();
     for (int i = 0; i < m.size(); ++i) {
-      this->add(m(i));
+      this->add(m_eval.coeff(i));
```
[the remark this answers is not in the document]
t4c1: I would use a `Eigen::Ref`, but it does not allow linear indexing. Using two loops could be inefficient if we get in a row-major matrix (such as a result of `.transpose()`).

stan/math/fwd/fun/mdivide_left.hpp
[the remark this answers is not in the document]
andrjohns: Ah gotcha, makes sense

stan/math/prim/err/check_lower_triangular.hpp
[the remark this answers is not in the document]
andrjohns: I meant down on line 38:
```
throw_domain_error(function, name, y(m, n), msg_str.c_str())
```
Shouldn't that be:
```
throw_domain_error(function, name, y_ref(m, n), msg_str.c_str())
```

stan/math/prim/fun/accumulator.hpp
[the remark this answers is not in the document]
andrjohns: That makes sense to me

stan/math/fwd/fun/mdivide_left.hpp
```diff
@@ -13,10 +13,18 @@
...
-    const Eigen::Matrix<fvar<T>, R2, C2> &b) {
+template <typename T1, typename T2,
+          require_all_eigen_vt<is_fvar, T1, T2>* = nullptr,
+          require_same_vt<T1, T2>* = nullptr>
+inline Eigen::Matrix<value_type_t<T1>, T1::RowsAtCompileTime,
+                     T2::ColsAtCompileTime>
+mdivide_left(const T1& A, const T2& b) {
+  using T = typename value_type_t<T1>::Scalar;
+  constexpr int R1 = T1::RowsAtCompileTime;
+  constexpr int C1 = T1::ColsAtCompileTime;
```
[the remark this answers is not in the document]
t4c1: fixed.

stan/math/fwd/fun/mdivide_left.hpp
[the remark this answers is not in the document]
t4c1: Hey, I think you have some experience with adding to Eigen library. Do you think it would be possible to make some kind of expression that would make a `fvar` "Matrix" by referencing two matrices of `double`s. So that we could directley evaluate a fvar expression into two matrices like this:
```
make_fvar_ref_matrix(val_A, deriv_A) = A.eval();
```

stan/math/fwd/fun/mdivide_left.hpp
```diff
@@ -55,10 +65,18 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
   return to_fvar(inv_A_mult_b, deriv);
 }
 
-template <typename T, int R1, int C1, int R2, int C2>
-inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
-    const Eigen::Matrix<double, R1, C1> &A,
-    const Eigen::Matrix<fvar<T>, R2, C2> &b) {
+template <typename T1, typename T2, require_eigen_t<T1>* = nullptr,
+          require_same_vt<double, T1>* = nullptr,
```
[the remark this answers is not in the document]
t4c1: fixed.

stan/math/fwd/fun/mdivide_left.hpp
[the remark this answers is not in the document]
andrjohns: Yep I could do that. I think the easiest approach would be a custom ```CwiseBinaryOp```, something like:
```
val_A.make_fvar(deriv_A)
```
I could just drop that into the current matrixbase addons. The tricky part will be the order of includes, since the ```fvar``` header will need to be included before the ```eigen_plugins.h``` header. but I can do some testing and see what works.

stan/math/fwd/fun/mdivide_left.hpp
```diff
@@ -31,17 +39,19 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left(
...
+      deriv_A(i, j) = A_ref(i, j).d_;
     }
   }
 
+  const Eigen::Ref<const plain_type_t<T2>>& b_ref = b;
   for (int j = 0; j < b.cols(); j++) {
     for (int i = 0; i < b.rows(); i++) {
-      val_b(i, j) = b(i, j).val_;
-      deriv_b(i, j) = b(i, j).d_;
+      val_b(i, j) = b_ref(i, j).val_;
```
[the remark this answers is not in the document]
t4c1: fixed.

stan/math/fwd/fun/mdivide_left_tri_low.hpp
```diff
@@ -56,10 +66,18 @@ inline Eigen::Matrix<fvar<T>, R1, C2> mdivide_left_tri_low(
   return to_fvar(inv_A_mult_b, deriv);
```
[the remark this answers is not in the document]
t4c1: I'll do that, but don't expect any significant performance benefit in this case. The majority of time is spent in those O(n**3) LU decompositions in `mdivide_left`.

stan/math/prim/err/check_lower_triangular.hpp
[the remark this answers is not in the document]
t4c1: fixed

stan/math/prim/fun/add_diag.hpp
```diff
@@ -13,49 +13,28 @@ namespace math {
...
  * @param mat a matrix
- * @param to_add value to add along the diagonal
+ * @param to_add scalar value or column vector or row vector to add along the
+ * diagonal
  * @return a matrix with to_add added along main diagonal
  */
-template <typename T_m, typename T_a>
+template <typename T_m, typename T_a, typename = require_eigen_t<T_m>,
+          typename
+          = require_t<disjunction<is_eigen_vector<T_a>, is_stan_scalar<T_a>>>>
```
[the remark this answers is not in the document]
t4c1: fixed

stan/math/prim/fun/add_diag.hpp
```diff
@@ -13,49 +13,28 @@ namespace math {
...
-         const Eigen::Matrix<T_a, R, C> &to_add) {
-  const size_t length_diag = std::min(mat.rows(), mat.cols());
-  check_consistent_size("add_diag", "number of elements of to_add", to_add,
-                        length_diag);
-
-  Eigen::Matrix<return_type_t<T_m, T_a>, Eigen::Dynamic, Eigen::Dynamic> out
-      = mat;
-  out.diagonal() += to_add;
+  out.diagonal().array()
+      += as_array_or_scalar(as_column_vector_or_scalar(to_add));
```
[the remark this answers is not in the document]
t4c1: Similar situations came up a lot in GLMs, that is why I added these two functions.

stan/math/prim/fun/append_row.hpp
t4c1: fixed

stan/math/prim/fun/chol2inv.hpp
t4c1: Also same dilemma between a copy and double evaluation.

stan/math/prim/fun/chol2inv.hpp
t4c1: fixed

stan/math/prim/fun/chol2inv.hpp
t4c1: fixed

stan/math/prim/fun/mdivide_left_tri.hpp
t4c1: This works in place, that is in memory, allocated by `b` so there is no copy. And if we did not make `b` a variable first we could not return it.

stan/math/prim/fun/mdivide_left_tri.hpp
t4c1: It would be if we had a `is_double` type trait. But we don't.

stan/math/prim/fun/mdivide_left_tri.hpp
t4c1: Only if we had `is_double` type trait. Maybe we should add it?

stan/math/prim/fun/mdivide_left_tri.hpp
t4c1: I can't use `plain_type_t` for previous function as it has rows and cols numbers from different input matrices.

stan/math/rev/fun/mdivide_left.hpp
t4c1: same answer too

stan/math/prim/fun/mdivide_left_tri.hpp
t4c1: Most OpenCL implementations only work with double. Passing in int would fail to compile.

stan/math/rev/fun/mdivide_left.hpp
t4c1: fixed

test/unit/math/fwd/fun/assign_test.cpp
t4c1: Indeed. I think there is no reason to have different requirements for `Eigen::Block` and `Eigen::Matrix`. I assume it is better to allow assigning of matrices woth different compile time sizes. Do you think it is prefereble to fail assigments for blocks with different sizes? Or is there a reason for difference in handling of matrices and blocks?

stan/math/prim/fun/mdivide_left_tri.hpp
SteveBronder: We have require_floating_point and I'm pretty sure the OpenCL will case float to double if it sees a float

t4c1: > case float to double if it sees a float

I don't think so. It seem it would just copy raw data resulting in garbage output with no errors.

t4c1:
@andrjohns This is ready for another review.

PR #1747, opened 2020-02-25
Generalize */fun startimg with chole-cov

t4c1:
## Summary

Generalizes inputs to functions with names starting with chol-cov.

## Tests
Just a refactor. No new tests

## Side Effects
None.

## Checklist

- [x] Math issue #1470


stan/math/prim/err/check_cholesky_factor.hpp
```diff
@@ -25,14 +25,14 @@ namespace math {
  *   if there are 0 columns, or if any element in matrix is NaN
  */
 template <typename T_y>
-inline void check_cholesky_factor(
-    const char* function, const char* name,
-    const Eigen::Matrix<T_y, Eigen::Dynamic, Eigen::Dynamic>& y) {
+inline void check_cholesky_factor(const char* function, const char* name,
+                                  const T_y& y) {
```
SteveBronder: Q/Optional: There's no other specializations of this but would we want T_y to fail during sfinae instead of after it selected the function? I'm not sure which one is nicer

stan/math/prim/err/check_lower_triangular.hpp
```diff
@@ -15,27 +15,27 @@ namespace math {
...
-    const char* function, const char* name,
-    const Eigen::Matrix<T_y, Eigen::Dynamic, Eigen::Dynamic>& y) {
+template <typename T_y, require_eigen_t<T_y>* = nullptr>
+inline void check_lower_triangular(const char* function, const char* name,
+                                   const T_y& y) {
+  const Eigen::Ref<const plain_type_t<T_y>>& y_ref = y;
   for (int n = 1; n < y.cols(); ++n) {
     for (int m = 0; m < n && m < y.rows(); ++m) {
-      if (y(m, n) != 0) {
+      if (y_ref(m, n) != 0) {
```
SteveBronder: Optional:

Performance wise, do you think it's worth replacing this loop with something like

```cpp
if((A.template triangularView<Eigen::StrictlyUpper>().nestedExpression().array() == 0).any()) {
/// ...
}
```

https://godbolt.org/z/u2oMaB

stan/math/prim/fun/cholesky_corr_constrain.hpp
```diff
@@ -12,20 +12,22 @@
 namespace stan {
 namespace math {
 
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_corr_constrain(
-    const Eigen::Matrix<T, Eigen::Dynamic, 1>& y, int K) {
+template <typename T, require_t<is_eigen_vector<T>>* = nullptr>
+Eigen::Matrix<value_type_t<T>, Eigen::Dynamic, Eigen::Dynamic>
+cholesky_corr_constrain(const T& y, int K) {
```
SteveBronder: Think I'd rather have a nicer name like `EigVec` here

stan/math/prim/fun/cholesky_corr_constrain.hpp
```diff
@@ -12,20 +12,22 @@
...
   int k_choose_2 = (K * (K - 1)) / 2;
   check_size_match("cholesky_corr_constrain", "y.size()", y.size(),
                    "k_choose_2", k_choose_2);
-  Matrix<T, Dynamic, 1> z(k_choose_2);
+  Matrix<T_scalar, Dynamic, 1> z(k_choose_2);
+  const Eigen::Ref<const plain_type_t<T>>& y_ref = y;
   for (int i = 0; i < k_choose_2; ++i) {
-    z(i) = corr_constrain(y(i));
+    z[i] = corr_constrain(y_ref[i]);
   }
```
SteveBronder: Could we remove the loop and `Eigen::Ref` here through

```cpp
Matrix<T_scalar, Dynamic, 1> z = corr_constrain(y.head(k_choose_2));
```

with

```cpp
template <typename T>
inline auto corr_constrain(T&& x) {
  using std::tanh;
  return tanh(std::forward<T>(x));
}
```

stan/math/prim/fun/cholesky_corr_constrain.hpp
```diff
@@ -45,20 +47,22 @@ Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_corr_constrain(
 }
 
 // FIXME to match above after debugged
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_corr_constrain(
-    const Eigen::Matrix<T, Eigen::Dynamic, 1>& y, int K, T& lp) {
+template <typename T, require_t<is_eigen_vector<T>>* = nullptr>
```
SteveBronder: `require_eigen_vector<EigVec>`?

stan/math/prim/fun/cholesky_corr_constrain.hpp
```diff
@@ -45,20 +47,22 @@ Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_corr_constrain(
...
   int k_choose_2 = (K * (K - 1)) / 2;
   check_size_match("cholesky_corr_constrain", "y.size()", y.size(),
                    "k_choose_2", k_choose_2);
-  Matrix<T, Dynamic, 1> z(k_choose_2);
+  Matrix<T_scalar, Dynamic, 1> z(k_choose_2);
+  const Eigen::Ref<const plain_type_t<T>>& y_ref = y;
   for (int i = 0; i < k_choose_2; ++i) {
-    z(i) = corr_constrain(y(i), lp);
+    z[i] = corr_constrain(y_ref[i], lp);
   }
```
SteveBronder: Same as other looper comment

stan/math/prim/fun/cholesky_corr_free.hpp
```diff
@@ -10,24 +10,25 @@
...
+  const Eigen::Ref<const plain_type_t<T>> x_ref = x;
   int K = (x.rows() * (x.rows() - 1)) / 2;
-  Matrix<T, Dynamic, 1> z(K);
+  Matrix<value_type_t<T>, Dynamic, 1> z(K);
   int k = 0;
   for (int i = 1; i < x.rows(); ++i) {
-    z(k++) = corr_free(x(i, 0));
-    double sum_sqs = square(x(i, 0));
+    z(k++) = corr_free(x_ref(i, 0));
+    double sum_sqs = square(x_ref(i, 0));
```
SteveBronder: Should these all be `coeffRef`?

stan/math/prim/fun/cholesky_corr_free.hpp
```diff
@@ -10,24 +10,25 @@
 namespace stan {
 namespace math {
 
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, 1> cholesky_corr_free(
-    const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& x) {
+template <typename T, require_eigen_t<T>* = nullptr>
+Eigen::Matrix<value_type_t<T>, Eigen::Dynamic, 1> cholesky_corr_free(
```
SteveBronder: Optional: No auto?

stan/math/prim/fun/cholesky_decompose.hpp
```diff
@@ -20,21 +21,27 @@ namespace math {
  * <p>\f$A = L \times L^T\f$.
  *
  * @tparam T type of elements in the matrix
```
SteveBronder:
```suggestion
 * @tparam T type deriving from `Eigen::MatrixBase`
```

stan/math/prim/fun/cholesky_decompose.hpp
```diff
@@ -20,21 +21,27 @@ namespace math {
...
+inline Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
+                     T::ColsAtCompileTime>
+cholesky_decompose(const T& m) {
+  eval_return_type_t<T>& m_eval = m.eval();
+  check_symmetric("cholesky_decompose", "m", m_eval);
+  check_not_nan("cholesky_decompose", "m", m_eval);
+  Eigen::LLT<Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
+                           T::ColsAtCompileTime> >
+      llt(m.rows());
+  llt.compute(m_eval);
```
SteveBronder: Is there a reason not to do

```cpp
Eigen::LLT<Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
                           T::ColsAtCompileTime>> llt = m_eval.llt();

File: stan/math/prim/fun/cholesky_factor_constrain.hpp [hunk 11]
```diff
@@ -17,39 +17,39 @@ namespace math {
  * specified vector.  A total of (N choose 2) + N + (M - N) * N
  * elements are required to read an M by N Cholesky factor.
  *
- * @tparam T type of elements in the matrix
+ * @tparam T type of the vector
```
SteveBronder:
```suggestion
 * @tparam T type that inherits from `Eigen::MatrixBase` and has one row or column of 1 at compile time.
```

stan/math/prim/fun/cholesky_factor_constrain.hpp
```diff
@@ -17,39 +17,39 @@ namespace math {
...
-  T zero(0);
+  Eigen::Matrix<T_scalar, Eigen::Dynamic, Eigen::Dynamic> y(M, N);
+  T_scalar zero(0);
   int pos = 0;
 
+  const Eigen::Ref<const plain_type_t<T>>& x_ref = x;
   for (int m = 0; m < N; ++m) {
     for (int n = 0; n < m; ++n) {
-      y(m, n) = x(pos++);
+      y(m, n) = x_ref(pos++);
```
SteveBronder: Is there a way in Eigen to fill a lower triangular of a matrix with an array?

stan/math/prim/fun/cholesky_factor_constrain.hpp
```diff
@@ -64,25 +64,24 @@ Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_factor_constrain(
  *
  * @tparam T type of elements in the matrix
```
SteveBronder: Pretty sure you can copy pasta from the other comments for all of these

stan/math/prim/fun/columns_dot_self.hpp
```diff
@@ -9,17 +10,16 @@ namespace math {
...
  * @param x Matrix.
  * @return Vector containing the dot product of each column of the matrix
  * with itself.
  */
-template <typename T, int R, int C>
-inline Eigen::Matrix<T, 1, C> columns_dot_self(
-    const Eigen::Matrix<T, R, C>& x) {
+template <typename T, require_eigen_t<T>* = nullptr,
+          require_not_eigen_vt<is_var, T>* = nullptr>
+inline Eigen::Matrix<value_type_t<T>, 1, T::ColsAtCompileTime> columns_dot_self(
```
SteveBronder: One reason I'd like to have auto on these is that we can add .eval() or .matrix() now and once we get these all working with generic eigen types we can just remove the .eval() and let it all rock and roll

stan/math/prim/fun/cov_matrix_constrain_lkj.hpp
```diff
@@ -68,18 +68,19 @@ Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cov_matrix_constrain_lkj(
...
   int pos = 0;
   for (size_t i = 0; i < k_choose_2; ++i) {
-    cpcs[i] = corr_constrain(x[pos++], lp);
+    cpcs[i] = corr_constrain(x_ref[pos++], lp);
   }
-  Eigen::Array<T, Eigen::Dynamic, 1> sds(k);
+  Eigen::Array<value_type_t<T>, Eigen::Dynamic, 1> sds(k);
   for (size_t i = 0; i < k; ++i) {
-    sds[i] = positive_constrain(x[pos++], lp);
+    sds[i] = positive_constrain(x_ref[pos++], lp);
```
SteveBronder: Same thing for positive constrain. I was goofing around with the free/constrain function in this branch though it's not ready for a PR for some performance stuff I goofed up

https://github.com/stan-dev/math/blob/refactor/cov-corr-constrain/stan/math/prim/fun/positive_constrain.hpp

stan/math/prim/fun/read_cov_matrix.hpp
```diff
@@ -13,36 +13,42 @@ namespace math {
...
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_cov_matrix(
-    const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs,
-    const Eigen::Array<T, Eigen::Dynamic, 1>& sds, T& log_prob) {
-  Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> L
+template <typename T_CPCs, typename T_sds,
+          require_all_eigen_vector_t<T_CPCs, T_sds>* = nullptr,
+          require_same_vt<T_CPCs, T_sds>* = nullptr>
+Eigen::Matrix<value_type_t<T_CPCs>, Eigen::Dynamic, Eigen::Dynamic>
+read_cov_matrix(const T_CPCs& CPCs, const T_sds& sds,
```
SteveBronder: This feels like a good candidate for forwarding

stan/math/prim/fun/factor_cov_matrix.hpp
```diff
@@ -18,35 +18,36 @@ namespace math {
...
-template <typename T>
-bool factor_cov_matrix(
-    const Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>& Sigma,
-    Eigen::Array<T, Eigen::Dynamic, 1>& CPCs,
-    Eigen::Array<T, Eigen::Dynamic, 1>& sds) {
+template <typename T_Sigma, typename T_CPCs, typename T_sds,
+          require_eigen_t<T_Sigma>* = nullptr,
+          require_all_eigen_vector_t<T_CPCs, T_sds>* = nullptr,
+          require_all_same_vt<T_Sigma, T_CPCs, T_sds>* = nullptr>
+bool factor_cov_matrix(const T_Sigma& Sigma, T_CPCs& CPCs, T_sds& sds) {
```
SteveBronder: Side Note: Another thing I goofed / am goofing with is removing CPCs and sds here as in/out parameters

https://github.com/stan-dev/math/blob/refactor/cov-corr-constrain/stan/math/prim/fun/factor_cov_matrix.hpp

stan/math/prim/fun/read_cov_L.hpp
```diff
@@ -14,17 +14,19 @@ namespace math {
...
  */
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> read_cov_L(
-    const Eigen::Array<T, Eigen::Dynamic, 1>& CPCs,
-    const Eigen::Array<T, Eigen::Dynamic, 1>& sds, T& log_prob) {
+template <typename T_CPCs, typename T_sds,
+          require_all_eigen_vector_t<T_CPCs, T_sds>* = nullptr,
+          require_same_vt<T_CPCs, T_sds>* = nullptr>
+Eigen::Matrix<value_type<T_CPCs>, Eigen::Dynamic, Eigen::Dynamic> read_cov_L(
+    const T_CPCs& CPCs, const T_sds& sds, value_type_t<T_CPCs>& log_prob) {
```
SteveBronder: idt we want log_prob to be var if T_CPCs is Eigen::Matrix<var> no?

SteveBronder: I know that's how we did it before but idt it needs to do that. Should it use `return_type_t` for T_cpcs and T_sds?

stan/math/prim/err/check_cholesky_factor.hpp
t4c1: Which error condition is this about?

stan/math/prim/err/check_lower_triangular.hpp
t4c1: Yes, but if we do it that way we lose information at which index the error is detected. So if we detect the error we than have to repeat what the function does now to find the index. 

I am not sure if it is worth complicating the code unless we find a use case where this function is something close to a bottleneck. I think it is mostly used next to matrix decompositions, which are an order of magnitude slower than this.

stan/math/prim/fun/cholesky_corr_constrain.hpp
t4c1: done.

stan/math/prim/fun/cholesky_corr_constrain.hpp
t4c1: done

stan/math/prim/fun/cholesky_corr_constrain.hpp
t4c1: fixed

stan/math/prim/fun/cholesky_corr_constrain.hpp
t4c1: fixed

stan/math/prim/fun/cholesky_corr_free.hpp
t4c1: `coeffRef` is required when writing values. For reading I think `coeff` is prefered.

stan/math/prim/fun/cholesky_corr_free.hpp
t4c1: fixed

stan/math/prim/fun/cholesky_decompose.hpp
t4c1: I can't think of any. Fixed.

stan/math/prim/fun/cholesky_factor_constrain.hpp
t4c1: I don't think so. I am changing this to use eigen expressions instead of inner loops.

stan/math/prim/fun/columns_dot_self.hpp
t4c1: This is again a case where eval returns a row major matrix.

stan/math/prim/fun/read_cov_matrix.hpp
t4c1: I see no need for perfect forwarding here.

stan/math/prim/fun/factor_cov_matrix.hpp
t4c1: I prefere in/out parameters to returning tuples. I made some more changes so that these two need not be copied to x after this function returns. Instead this function gets segments of that x.

stan/math/prim/fun/read_cov_L.hpp
t4c1: Honestly I have no idea. I just tried to make it do the same thing as before. Anyway that is probably better put in a separate PR.

stan/math/prim/fun/cholesky_factor_constrain.hpp
t4c1: fixed

stan/math/prim/fun/cholesky_factor_constrain.hpp
t4c1: done

stan/math/prim/err/check_lower_triangular.hpp
SteveBronder: Cool cool yeah I think having that info is important

stan/math/prim/err/check_cholesky_factor.hpp
SteveBronder: If T_y was not an eigen matrix

stan/math/prim/fun/read_cov_matrix.hpp
SteveBronder: Why not? It looks like they are just being forwarded along to read_cov_L?

I started trying to write a test to look at const& vs. &&, what do you think about the below?

https://github.com/seantalts/perf-math/blob/test/eigen_pf/eigen_pf.cpp

I used `noinline` to act like a larger function that would not be an inline candidate and no_icf because I think preserving the original functions is a thing that makes sense (mby idk?)

Besides that it just does a mul passing the values down two layers of functions. You should be able to compile and run it with

```
make eigen_pf && ./eigen_pf
```

I'm running a longer test ala
```
./eigen_pf --benchmark_repetitions=50 --benchmark_report_aggregates_only=true
```

To get a better idea of how the large data benchmarks run on average

SteveBronder: One run on mine says it's about 1-2% faster as a ballpark, though I'd rather wait for the numbers from a few repetitions to get an idea of pf is useful in these situations

```
Running ./eigen_pf
Run on (32 X 3500 MHz CPU s)
CPU Caches:
  L1 Data 32K (x16)
  L1 Instruction 64K (x16)
  L2 Unified 512K (x16)
  L3 Unified 8192K (x4)
Load Average: 1.44, 1.60, 1.41
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
add_copy_bench/2           176 ns          176 ns      4040759
add_copy_bench/4           320 ns          320 ns      2192642
add_copy_bench/8          1112 ns         1112 ns       626119
add_copy_bench/16         4275 ns         4274 ns       163690
add_copy_bench/32        17538 ns        17537 ns        39809
add_copy_bench/64        78332 ns        78331 ns         8988
add_copy_bench/128      412057 ns       412041 ns         1702
add_copy_bench/256     2411324 ns      2411265 ns          298
add_copy_bench/512    14329605 ns     14329181 ns           48
add_copy_bench/1024   92212267 ns     92210599 ns            7
add_copy_bench/2048  667639201 ns    667627868 ns            1
add_copy_bench/4096 5055086122 ns   5054647216 ns            1
add_copy_bench/8192 38807557827 ns   38803388581 ns            1
add_pf_bench/2             175 ns          175 ns      3985378
add_pf_bench/4             318 ns          318 ns      2196480
add_pf_bench/8            1161 ns         1161 ns       611528
add_pf_bench/16           4424 ns         4424 ns       157600
add_pf_bench/32          18263 ns        18262 ns        38474
add_pf_bench/64          81782 ns        81761 ns         8585
add_pf_bench/128        388528 ns       388511 ns         1791
add_pf_bench/256       2120808 ns      2120470 ns          335
add_pf_bench/512      13709277 ns     13707653 ns           49
add_pf_bench/1024     89777511 ns     89767389 ns            7
add_pf_bench/2048    701121667 ns    701069035 ns            1
add_pf_bench/4096   4939446339 ns   4939085649 ns            1
add_pf_bench/8192   38407552743 ns   38404496372 ns            1
```

t4c1: Benchmark code seems fine for testing passing in rvalues. For completness maybe you should also test lvalues. I ran 50 iterations over a bit sparser range of sizes and it seems pf is a bit than passing by const ref. Makes sense, since pf needs to invoke move constructor and alternative just passes pointer. Running the benchmark a few more times differencer vary. I think it is safe to say these differences are most likely just noise.

```
Running F:\Users\tadej\Programming\bstatcomp\perf-math\eigen_pf.exe
Run on (4 X 3293 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
------------------------------------------------------------------
Benchmark                           Time           CPU Iterations
------------------------------------------------------------------
add_copy_bench/2_mean             453 ns        452 ns    1544828
add_copy_bench/2_median           452 ns        455 ns    1544828
add_copy_bench/2_stddev             1 ns          5 ns    1544828
add_copy_bench/4_mean             832 ns        831 ns     896000
add_copy_bench/4_median           831 ns        837 ns     896000
add_copy_bench/4_stddev             5 ns          9 ns     896000
add_copy_bench/16_mean           8900 ns       8862 ns      89600
add_copy_bench/16_median         8859 ns       8894 ns      89600
add_copy_bench/16_stddev          123 ns         91 ns      89600
add_copy_bench/64_mean         156052 ns     155538 ns       3733
add_copy_bench/64_median       154832 ns     154869 ns       3733
add_copy_bench/64_stddev         2452 ns       2294 ns       3733
add_copy_bench/256_mean       3902117 ns    3877444 ns        179
add_copy_bench/256_median     3873370 ns    3840782 ns        179
add_copy_bench/256_stddev       63451 ns      53170 ns        179
add_copy_bench/1024_mean    137456676 ns  137062500 ns          5
add_copy_bench/1024_median  136603340 ns  137500000 ns          5
add_copy_bench/1024_stddev    1902190 ns    1895088 ns          5
add_copy_bench/2048_mean    941109244 ns  936875000 ns          1
add_copy_bench/2048_median  934882400 ns  937500000 ns          1
add_copy_bench/2048_stddev   13821822 ns   12611113 ns          1
add_pf_bench/2_mean               458 ns        456 ns    1544828
add_pf_bench/2_median             454 ns        455 ns    1544828
add_pf_bench/2_stddev               7 ns          7 ns    1544828
add_pf_bench/4_mean               838 ns        834 ns     896000
add_pf_bench/4_median             830 ns        837 ns     896000
add_pf_bench/4_stddev              14 ns         16 ns     896000
add_pf_bench/16_mean             8931 ns       8873 ns      74667
add_pf_bench/16_median           8862 ns       8789 ns      74667
add_pf_bench/16_stddev            129 ns        164 ns      74667
add_pf_bench/64_mean           155515 ns     155064 ns       4480
add_pf_bench/64_median         154515 ns     153460 ns       4480
add_pf_bench/64_stddev           2200 ns       2462 ns       4480
add_pf_bench/256_mean         3894863 ns    3877005 ns        187
add_pf_bench/256_median       3867772 ns    3843583 ns        187
add_pf_bench/256_stddev         62613 ns      53382 ns        187
add_pf_bench/1024_mean      138726830 ns  137375000 ns          5
add_pf_bench/1024_median    138238250 ns  137500000 ns          5
add_pf_bench/1024_stddev      1994051 ns    2358901 ns          5
add_pf_bench/2048_mean      948947858 ns  941562500 ns          1
add_pf_bench/2048_median    942359850 ns  937500000 ns          1
add_pf_bench/2048_stddev     22353437 ns   16345018 ns          1
```

t4c1: > Why not? It looks like they are just being forwarded along to read_cov_L?

In this case using pf does not avoid any copying. It is just more code.

stan/math/prim/err/check_cholesky_factor.hpp
t4c1: I think it does not matter. For consistency with other functions I think it is fine as it is.

t4c1:
I added a fix for `is_eigen` wich did not work with segments of Eigen vectors.

andrjohns:
> call .matrix() before .val() and .d() 

We could avoid the need for this by exposing those methods for ```Array``` types as well. In ```prim/fun/Eigen.hpp```, just need to add the following line:
```
#define EIGEN_ARRAYBASE_PLUGIN "stan/math/prim/eigen_plugins.h"
```

t4c1:
@andrjohns I tried doing that, but for some reason I can't get it to work. When I remove `.matrix()` I keep gettin `error: no member named 'val' in Eigen::Ref<const Eigen::Array< stan::math::fvar< stan::math::fvar< double> >, -1, 1, 0, -1, 1>, 0, Eigen::InnerStride<1> >`. If you want feel free to push the solution into this branch.

andrjohns:
The problem was the header guards in ```eigen_plugins.h```, it meant that nothing was being included in the plugin definition for ArrayBase. I don't have permissions to push to your branch, but I've put up an example fix [on this branch](https://github.com/stan-dev/math/compare/develop...andrjohns:array_plugins)

stan/math/prim/fun/read_cov_matrix.hpp
SteveBronder: I ran this for 50 iterations for each (data [here](https://gist.github.com/SteveBronder/5d0c064f3b951b3dcd554601b660ee11)) and it looks like moves were pretty beneficial!

image


I can't explain the big spike at around 1000, I actually ran this twice and saw that same spike both times. Maybe something to do with my CPU?

Also fyi I consider this a side convo it doesn't have to be resolved for the PR to go through

SteveBronder: > In this case using pf does not avoid any copying. It is just more code.

It's not that pf avoids copying it's that pf transfers ownership of the memory and avoids aliasing issues. See the comment below

https://github.com/stan-dev/math/pull/1720#discussion_r381542520

bob-carpenter:
Big spikes in this kind of thing may just be related to cache thresholds.

What's the context here?  I can't tell what's being modified to what.  I
was commenting that our autodiff variables don't manage memory that can be
moved.  Eigen's a different story.

SteveBronder:
> What's the context here?  I can't tell what's being modified to what.

This PR changes the signatures for a bunch of functions to use generic eigen types instead of only taking in Eigen matrices or arrays. The benchmarks above are to check whether we should also do perfect forwarding (`&&` in the signature) vs. keeping things `const &`

>  I was commenting that our autodiff variables don't manage memory that can be
moved.  Eigen's a different story.

Yep! The above tests are to check moves vs. passing by const ref for Eigen matrices of doubles while trying to account for the aliasing issues non-inlined functions can have with const-ref. It may be a good idea to file an issue or comment on Eigen's forums to check those benchmarks make sense

stan/math/prim/fun/read_cov_matrix.hpp
SteveBronder: Bob asked that I close these and move them over to discourse so I'm gonna post over there and close this

stan/math/prim/err/check_cholesky_factor.hpp
SteveBronder: Fix docs for T_y

stan/math/prim/fun/cholesky_decompose.hpp
```diff
@@ -46,35 +52,36 @@ inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> cholesky_decompose(
  * original matrix \f$A\f$ is given by
  * <p>\f$A = L \times L^T\f$.
  *
- * @param m Symmetrix matrix.
+ * @param m Symmetric matrix.
```
SteveBronder: Also update docs for T

SteveBronder: Also instead of T maybe EigMat or something

stan/math/prim/meta/is_eigen.hpp
```diff
@@ -18,18 +18,12 @@ namespace internal {
...
-struct is_eigen_base<Eigen::MatrixBase<T>> : std::true_type {};
-
-template <typename T>
-struct is_eigen_base<Eigen::EigenBase<T>> : std::true_type {};
+struct is_eigen_base {
+  static std::false_type f(...);
+  template <typename Derived>
+  static std::true_type f(const Eigen::EigenBase<Derived>&);
+  enum { value = decltype(f(std::declval<T>()))::value };
+};
```
SteveBronder: I agree the old one is not good, is there a way that does not do `...` for a variadic argument list in `f()`. What about

```cpp
/**
 * Check if type derives from EigenBase
 **/
template <typename T, typename Enable = void>
struct is_eigen : std::false_type {};
// Check plain object for Eigen type fits in EigenBase
template <typename T>
struct is_eigen<T,
 std::enable_if_t<std::is_base_of<Eigen::EigenBase<typename std::decay_t<T>::PlainObject>, typename std::decay_t<T>::PlainObject>::value>> : std::true_type {};

// For LDLT and LLT
template <typename T>
struct is_eigen<T,
std::enable_if_t<std::is_base_of<Eigen::EigenBase<typename std::decay_t<T>::MatrixType>, typename std::decay_t<T>::MatrixType>::value>> : std::true_type {};

template <typename T>
struct is_eigen<Eigen::EigenBase<T>, void>: std::true_type {};
```

stan/math/rev/fun/sum.hpp
```diff
@@ -94,8 +94,8 @@ class sum_eigen_v_vari : public sum_v_vari {
  * @param m Specified matrix or vector.
  * @return Sum of coefficients of matrix.
  */
-template <int R, int C>
-inline var sum(const Eigen::Matrix<var, R, C>& m) {
+template <typename T, require_eigen_vt<is_var, T>* = nullptr>
```
SteveBronder: Remove docs for R, C and add doc for T

stan/math/fwd/fun/sum.hpp
```diff
@@ -34,25 +34,18 @@ inline fvar<T> sum(const std::vector<fvar<T> >& m) {
...
- * @tparam C number of columns, can be Eigen::Dynamic
+ * @tparam T type of the matrix
  *
  * @param m Matrix.
  * @return Sum of matrix entries.
  */
-template <typename T, int R, int C>
-inline fvar<T> sum(const Eigen::Matrix<fvar<T>, R, C>& m) {
+template <typename T, require_eigen_vt<is_fvar, T>* = nullptr>
+inline value_type_t<T> sum(const T& m) {
```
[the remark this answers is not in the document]
t4c1: Having a trait metaprogram as return type tells someone who reads the code more than just `auto`. Since `value_type_t<T>` is not long and unreadable I prefere it to `auto`. 
But I agree that duplicating it in return statement does not make sense so I am changing the return statement to use implicit conversion.

stan/math/prim/eigen_plugins.h
```diff
@@ -1,6 +1,3 @@
-#ifndef STAN_MATH_PRIM_EIGEN_PLUGINS_H
-#define STAN_MATH_PRIM_EIGEN_PLUGINS_H
-
```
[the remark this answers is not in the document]
t4c1: These are gone. We are now including this file twice - both in `MatrixBase` and `ArrayBase` so it can not have header guards anymore.

stan/math/prim/err/check_cholesky_factor.hpp
t4c1: done

stan/math/prim/meta/is_eigen.hpp
t4c1: I know that "elipses are bad". But in this case the function is never called. It is not even defined, just declared to implement a trait metaprogram. So we can not hit any of the dangers of using an elipsis.

While your suggestion would work for the segment case I am not sure if there are some Eigen types that have plain type that indirectly inherits `EigenBase`. If there are, the suggested implementation would break again.

I found an elegant way to avoid elipsis by using pointers. Since many people are not familiar with elipses I am changing it.

stan/math/prim/fun/cholesky_decompose.hpp
t4c1: done

stan/math/rev/fun/sum.hpp
t4c1: done

stan/math/prim/eigen_plugins.h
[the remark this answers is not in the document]
SteveBronder: Hmm, okay. I have to think about this for 5 minutes. As long as it's isolated in those two files then I think it's fine

SteveBronder:
@t4c1 looks like there is a merge conflict on cholesky_decompose. If you afk I can fix that up real quick

Issue #1230, opened 2019-05-08
Make use of composed functions and temporaries


andrjohns:
## Description
There are several places throughout the math library where arithmetic is being used in place of the composed functions and constants. There are also places where temporary variables could be used in place of repeated calculations.

An example of this can be seen in: ```prim/scal/fun/lub_constrain```:
```
  if (x > 0) {
    T exp_minus_x = exp(-x);
    inv_logit_x = 1.0 / (1.0 + exp_minus_x);
    // Prevent x from reaching one unless it really really should.
    if ((x < std::numeric_limits<double>::infinity()) && (inv_logit_x == 1))
      inv_logit_x = 1 - 1e-15;
  } else {
    T exp_x = exp(x);
    inv_logit_x = 1.0 - 1.0 / (1.0 + exp_x);
    // Prevent x from reaching zero unless it really really should.
    if ((x > -std::numeric_limits<double>::infinity()) && (inv_logit_x == 0))
      inv_logit_x = 1e-15;
  }
```
Which could be updated to:
```
  T inv_logit_x;
  if (x > 0) {
    inv_logit_x = inv_logit(x);
    // Prevent x from reaching one unless it really really should.
    if ((x < INFTY) && (inv_logit_x == 1))
      inv_logit_x = 1 - 1e-15;
  } else {
    inv_logit_x = inv_logit(x);
    // Prevent x from reaching zero unless it really really should.
    if ((x > NEGATIVE_INFTY) && (inv_logit_x == 0))
      inv_logit_x = 1e-15;
  }
```

I'll be working through this in my spare time, and will break the PRs into multiple parts. I'll update this issue as I go

- [ ]  ```prim```
   * [ ]  ```scal```
      * [X]  ```fun```
      * [ ]  ```prob```
   * [ ]  ```mat```
      * [ ]  ```fun```
      * [ ]  ```prob```
   * [ ]  ```arr```
      * [ ]  ```fun```

#### Current Version:
v2.19.1

PR #1744, opened 2020-02-25
Reuse intermediate computations in distributions part 1

mcol:
## Summary
This is the first of a few PRs that aim to clean up the distributions so that we make better use of intermediate computations and composed functions.

This covers the following distributions:
- bernoulli
- bernoulli_logit_glm
- beta
- beta_binomial
- beta_proportion
- binomial
- double_exponential
- exponential
- exp_mod_normal

## Tests

None, this is just cleanup.

## Side Effects

None.

## Checklist

- [X] Math issue: #1230

- [X] Copyright holder: Marco Colombo

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [X] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [X] the code is written in idiomatic C++ and changes are documented in the doxygen

- [X] the new changes are tested

PR #1331, opened 2019-09-01
Moves if statements for scal/prob/beta-binomial out of for loops

SteveBronder:
## Summary

While looking at how some of the internals worked I noticed beta binomials lpmf function has a bunch of if statements in it's for loops. This refactors the code to remove that and remove a bunch of the conditionals.

## Tests
refactor so no new tests

## Side Effects
nope
## Checklist

- [x] Math issue #1308 

- [x] Copyright holder: Steve Bronder

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

Issue #596, opened 2017-08-01
a / (b * c) is more efficient than a / b / c


syclik:
#### Summary:
See @bob-carpenter's [comment](https://github.com/stan-dev/math/pull/588#discussion_r130223133)

This appears in at least:
- [ ] stan/math/prim/scal/prob/neg_binomial_lccdf.hpp

#### Description:
This will make the math library a little faster.

#### Current Version:
v2.16.0

seantalts:
On my machine with primitives, these takes exactly the same amount of time and we can do a billion of them in a second. Is this more important with autodiff variables for some reason?

bob-carpenter:
Two things.  (1) Yes, more important with autodiff.  Follow the chain rule---the expression for the derivatives involves more operations.

(2)  Maybe you're doing the evaluation wrong.  I used this code to make sure the compiler doesn't optimize away anything.  Integer arithmetic in parallel is mainly free, so I wasn't worried about that.  The eval underestimates the difference as there's also addition.  What surprised me is that addition wasn't faster than multiplication---I'm guessing this test isn't very fine grained either.

```
#include <iostream>

int main() {
  double a;
  for (double x = 1; x < 100000; ++x)
    for (double y = 1; y < 100000; ++y)
      a += x / y;
  std::cout << "a = " << a << std::endl;
}
```

Division:

```
~/temp2$ clang++ -O3 race.cpp
~/temp2$ time ./a.out
a = 6.04501e+10

real	0m45.319s
user	0m45.218s
sys	0m0.069s
```

Multiplication:

```
~/temp2$ time ./a.out
a = 2.49995e+19

real	0m9.608s
user	0m9.593s
sys	0m0.010s
```

bob-carpenter:
I didn't test `a / b / c`---that might get unfolded into `a / (b * c)`.  They're not quite equivalent w.r.t. overflow behavior at the size limits.

seantalts:
I tested this as well with the following code:
```cpp
#include <stdlib.h>

int main (int argc, char* argv[]) {
  double a = atof(argv[0]);
  double b = atof(argv[1]);
  double c = atof(argv[2]);
  double d = 0;
  for (int i = 0; i < 1000000000; ++i) {
    //d += a / b / c;
    d += a / (b * c);
  }
  return d;
}
```
[...]

PR #1752, opened 2020-02-28
Reuse intermediate computations in distributions part 2

mcol:
This is the second of a few PRs that aim to clean up the distributions so that we make better use of intermediate computations and composed functions.

This covers the following distributions:
- chi_square
- gamma
- inv_chi_square
- inv_gamma
- scaled_inv_chi_square

## Tests

None, this is just cleanup.

## Side Effects

None.

## Release Notes

Cleaned up implementations of `chi_square`, `gamma`, `inv_chi_square`, `inv_gamma`, and `scaled_inv_chi_square` to make use of intermediate calculations more.

## Checklist

- [X] Math issue: #1230


stan/math/prim/prob/chi_square_lccdf.hpp
```diff
@@ -51,59 +51,58 @@ return_type_t<T_y, T_dof> chi_square_lccdf(const T_y& y, const T_dof& nu) {
...
 
+  VectorBuilder<true, T_partials_return, T_dof> half_nu(size_nu);
+  VectorBuilder<!is_constant_all<T_y, T_dof>::value, T_partials_return, T_dof>
+      tgamma_vec(size_nu);
   VectorBuilder<!is_constant_all<T_dof>::value, T_partials_return, T_dof>
-      gamma_vec(size(nu));
-  VectorBuilder<!is_constant_all<T_dof>::value, T_partials_return, T_dof>
-      digamma_vec(size(nu));
-
-  if (!is_constant_all<T_dof>::value) {
```
bbbales2: Was there a bug here? It feels like the logic changed from checking only T_dof to T_dof and T_y

stan/math/prim/prob/chi_square_lcdf.hpp
```diff
@@ -51,59 +51,58 @@ return_type_t<T_y, T_dof> chi_square_lcdf(const T_y& y, const T_dof& nu) {
...
 
+  VectorBuilder<true, T_partials_return, T_dof> half_nu(size_nu);
+  VectorBuilder<!is_constant_all<T_y, T_dof>::value, T_partials_return, T_dof>
+      tgamma_vec(size_nu);
   VectorBuilder<!is_constant_all<T_dof>::value, T_partials_return, T_dof>
-      gamma_vec(size(nu));
-  VectorBuilder<!is_constant_all<T_dof>::value, T_partials_return, T_dof>
-      digamma_vec(size(nu));
-
-  if (!is_constant_all<T_dof>::value) {
```
bbbales2: Same question as above with T_dof and T_y

stan/math/prim/prob/chi_square_lpdf.hpp
```diff
@@ -61,64 +69,59 @@ return_type_t<T_y, T_dof> chi_square_lpdf(const T_y& y, const T_dof& nu) {
...
                 T_y>
-      log_y(size(y));
-  for (size_t i = 0; i < stan::math::size(y); i++) {
-    log_y[i] = log(value_of(y_vec[i]));
-  }
-
+      log_y(size_y);
   VectorBuilder<include_summand<propto, T_y>::value, T_partials_return, T_y>
-      inv_y(size(y));
-  for (size_t i = 0; i < stan::math::size(y); i++) {
```
bbbales2: Looks like the template logic changed. Now we're checking T_dof. Was there a bug?

stan/math/prim/prob/chi_square_lpdf.hpp
```diff
@@ -61,64 +69,59 @@ return_type_t<T_y, T_dof> chi_square_lpdf(const T_y& y, const T_dof& nu) {
...
-    if (include_summand<propto, T_dof>::value) {
+      digamma_half_nu(size_nu);
+  if (include_summand<propto, T_dof>::value) {
+    for (size_t i = 0; i < size_nu; i++) {
+      T_partials_return half_nu = 0.5 * value_of(nu_vec[i]);
       lgamma_half_nu[i] = lgamma(half_nu);
-    }
-    if (!is_constant_all<T_dof>::value) {
-      digamma_half_nu_over_two[i] = digamma(half_nu) * 0.5;
+      if (!is_constant_all<T_dof>::value) {
```
bbbales2: Template logic changed?

stan/math/prim/prob/gamma_cdf.hpp
```diff
@@ -75,16 +73,19 @@ return_type_t<T_y, T_shape, T_inv_scale> gamma_cdf(const T_y& y,
...
+      tgamma_vec(size_alpha);
   VectorBuilder<!is_constant_all<T_shape>::value, T_partials_return, T_shape>
-      gamma_vec(size(alpha));
-  VectorBuilder<!is_constant_all<T_shape>::value, T_partials_return, T_shape>
-      digamma_vec(size(alpha));
+      digamma_vec(size_alpha);
 
-  if (!is_constant_all<T_shape>::value) {
-    for (size_t i = 0; i < stan::math::size(alpha); i++) {
+  if (!is_constant_all<T_y, T_shape, T_inv_scale>::value) {
```
bbbales2: This is different template logic.

stan/math/prim/prob/chi_square_lccdf.hpp
mcol: For `digamma_vec` nothing has changed, only the diff is a bit hard to read.

What changed is the logic for `tgamma_vec`: I've done this so that this precomputed term can be used further down in a `if (!is_constant_all<T_y>::value)` block, where before we were recomputing that term.

stan/math/prim/prob/chi_square_lcdf.hpp
mcol: Same stuff, we can reuse a precomputed `tgamma` term.

stan/math/prim/prob/chi_square_lpdf.hpp
mcol: I don't think so. I've moved the `if (include_summand<propto, T_dof>::value)` test outside the for loop (this case is marginally easier to follow if you ignore space changes).

stan/math/prim/prob/gamma_cdf.hpp
mcol: This is another situation in which a `tgamma` function could be precomputed for two extra cases (see the `if (!is_constant_all<T_y>::value)` and the `if (!is_constant_all<T_inv_scale>::value)` tests in which before we were recomputing `tgamma(alpha_dbl)`).

stan/math/prim/prob/chi_square_lpdf.hpp
mcol: Earlier we had:
```
VectorBuilder<include_summand<propto, T_y, T_dof>::value, T_partials_return, T_y>
    log_y(size(y));
for (size_t i = 0; i < stan::math::size(y); i++) {
  log_y[i] = log(value_of(y_vec[i]));
}
```
With this change I'm doing the for loop if `include_summand<propto, T_y, T_dof>::value` is true, as per that VectorBuilder definition. But it doesn't matter, as if `include_summand<propto, T_y, T_dof>::value` is true, we've already returned 0 before reaching there. So neither the old version not this one is buggy, but it can be further simplified.

mcol:
All these are tested in the distribution tests. For chi_square_lcdf, for example, all the following are being tested:
```double, double
double, std::vector<double>
double, Eigen::Matrix<double, Eigen::Dynamic, 1>
double, var
double, std::vector<var>
double, Eigen::Matrix<var, Eigen::Dynamic, 1>
std::vector<double>, double
std::vector<double>, std::vector<double>
std::vector<double>, Eigen::Matrix<double, Eigen::Dynamic, 1>
std::vector<double>, var
std::vector<double>, std::vector<var>
std::vector<double>, Eigen::Matrix<var, Eigen::Dynamic, 1>
Eigen::Matrix<double, Eigen::Dynamic, 1>, double
Eigen::Matrix<double, Eigen::Dynamic, 1>, std::vector<double>
Eigen::Matrix<double, Eigen::Dynamic, 1>, Eigen::Matrix<double, Eigen::Dynamic, 1>
Eigen::Matrix<double, Eigen::Dynamic, 1>, var
Eigen::Matrix<double, Eigen::Dynamic, 1>, std::vector<var>
Eigen::Matrix<double, Eigen::Dynamic, 1>, Eigen::Matrix<var, Eigen::Dynamic, 1>
var, double
var, std::vector<double>
var, Eigen::Matrix<double, Eigen::Dynamic, 1>
var, var
var, std::vector<var>
var, Eigen::Matrix<var, Eigen::Dynamic, 1>
std::vector<var>, double
std::vector<var>, std::vector<double>
std::vector<var>, Eigen::Matrix<double, Eigen::Dynamic, 1>
std::vector<var>, var
std::vector<var>, std::vector<var>
std::vector<var>, Eigen::Matrix<var, Eigen::Dynamic, 1>
Eigen::Matrix<var, Eigen::Dynamic, 1>, double
Eigen::Matrix<var, Eigen::Dynamic, 1>, std::vector<double>
Eigen::Matrix<var, Eigen::Dynamic, 1>, Eigen::Matrix<double, Eigen::Dynamic, 1>
Eigen::Matrix<var, Eigen::Dynamic, 1>, var
Eigen::Matrix<var, Eigen::Dynamic, 1>, std::vector<var>
Eigen::Matrix<var, Eigen::Dynamic, 1>, Eigen::Matrix<var, Eigen::Dynamic, 1>
```
Same for fvar<double>, fvar<fvar<double>>, fvar<var> and fvar<fvar<var>>, so I think we are well covered.

I'm going to push a further commit to ensures that the `size_zero` checks happen after we checked the consistency of the other arguments, along with a few more cleanups that I've noticed while having another look at this.

Thanks for looking at this PR too!

bbbales2:
> All these are tested in the distribution tests.

Thanks for checking!

SteveBronder:
Should this branch be tested by all the models in the cmdstan performance test suit? I only say that because I've tried one of these before that led to a revert

bbbales2:
The code here looks good (tests pass and looking over it there's nothing obviously wrong), but I'm worried (as Steve points out) that we've made changes to these distribution functions before and it's broken stuff downstream unexpectedly.

Regarding this scope/objectives of pull/how to get it merged, I have a few questions:

1. This is part of a larger pull request broken into pieces (including one that was already merged: https://github.com/stan-dev/math/pull/1744). How many pieces are there? Are they just touching all the distributions?

2. I don't think issue https://github.com/stan-dev/math/issues/1230 justifies the changes here.

It looks like the point of these changes is to make the lpdfs faster in situations where not all of the arguments are autodiff variables (and so there are optional calculations). Is that closer to the mark?

I guess the additional question here is how will I know the problem fixed?

3. What are the specific issues where this came up and what were the fixes? Specifically what broke for you @SteveBronder? I remember the glms have broken before as well. I'd like to know this before figuring out if we're tested enough or need more tests.

I'll look into #3 myself (feel free to add links if you know about stuff). #1 and #2 are for you @mcol.

mcol:
My plan is to cover the large majority of distributions, excluding the multivariate distributions and the glms (for both these types of distributions, the code is quite different from the rest, and I'm not familiar with that). If we go with PRs of sizes such as this and #1744 (which seems a reasonable size to me, but let me know if smaller would be desirable), then we will have a part 3 and a part 4. Note that #1758 will reduce the size of these PRs, so I will wait until it goes in before I pick this one up again.

This set of PRs does indeed more that what #1230 required. I can open a new issue to describe that, if that's what you'd like to see. Overall, this is not motivated by performance gains, as I expect them to be generally tiny if measurable at all, but from the attempt to reuse some intermediate computations and simplify existing ones. In a few cases some computations can be avoided for non-autodiff types, but those are the minority.

bbbales2:
> I can open a new issue to describe that, if that's what you'd like to see

Just talk them out here for now. We can just copy-paste them over to an issue easily.

> Overall, this is not motivated by performance gains, as I expect them to be generally tiny if measurable at all, but from the attempt to reuse some intermediate computations and simplify existing ones.

Oh, is it for numerical reasons? I don't think simplifying these things just for the sake of simplification really justifies the work or dangers, honestly.

mcol:
This is an example of what I mean by simplification, from stan/math/prim/prob/chi_square_cdf.hpp (the first one in the diff):
```
-    
-      ops_partials.edge1_.partials_[n] += beta_dbl * exp(-beta_dbl * y_dbl)
-                                          * pow(beta_dbl * y_dbl, alpha_dbl - 1)
-                                          / tgamma(alpha_dbl) / Pn;
+      ops_partials.edge1_.partials_[n] += 0.5 * exp(-half_y_dbl)
+                                          * pow(half_y_dbl, half_nu_dbl - 1)
+                                          / (tgamma_vec[n] * Pn);
```
What's going on is the following:
- `beta_dbl` is a constant of value 0.5 defined a few lines above this block, so I just replace it with 0.5
- `beta_dbl * y_dbl` is used already in two places in this block, so I precompute it in variable `half_y_dbl`
- `alpha_dbl` is defined a few lines above this block to be `value_of(nu_vec[n]) * 0.5` (would you have guessed it?), and for clarity I rename it as `half_nu_dbl`
- `tgamma(alpha_dbl)` was actually precomputed in a VectorBuilder but not used here, so I replace it with the VectorBuilder version called `tgamma_vec[n]`
- instead of the two divisions `/ tgamma(alpha_dbl) / Pn`, I follow the suggestion from #596 and rewrite is as `/ (tgamma_vec[n] * Pn)`

In this example, only the last point could potentially change the numerics, and only skipping the extra `tgamma` call could potentially help performance. The rest is pretty trivial stuff that makes the code a bit more clear.

There may be cases in which what's going on is a bit more involved than this: I could go and find an example if you'd find it helpful.

bbbales2:
Yeah the numbers make sense and weren't too bad to check.

The things that scare me are the template logic/the VectorBuilder logic/the ```loop``` and ```if``` placement. That stuff I have trouble following.

Looking at the pull request that got reverted for Steve, I think the culprit line is: https://github.com/stan-dev/math/pull/1331/files#diff-565dac80cbb935e5597e584799cc9220R76

```if (include_summand<propto>::value)``` should be ```if (include_summand<propto, T_size1, T_size2>::value)```, I think (didn't test this though, so quite likely wrong). I will look more tomorrow to understand what happened.

**Comment by mcol** [comment]
That line is part of the problem. In case of Steve's bug, I think what happened is that the normalizing constants were originally computed in three blocks:
1. the binomial coefficient term if `include_summand<propto>::value`
2. the lbeta numerator if `include_summand<propto, T_size1, T_size2>::value`
3. the lbeta denominator again if `include_summand<propto, T_size1, T_size2>::value`

Further down, those terms were summed up again respecting those `include_summand` conditions.

In the buggy version, they got computed only if `include_summand<propto>::value`, as you pointed out. Which means that for `propto=true` (that is using the `~` notation) the contribution of the `lbeta` terms was lost, even if `T_size1` and `T_size2` were autodiff types.

In general, the rule for the `include_summand` or `!is_constant_all` checks is that if you add a term to their parameter list, that block will be run in more situations (whenever any of the parameters is an autodiff type): so the addition of a term in general is not dangerous (it may be detrimental to performance, but not to correctness). What is potentially more risky is the removal of a term, and this should be double-checked against the rest of the changes. 

For example, looking at this code block (again, the first I found):
```
-  if (!is_constant_all<T_dof>::value) {
-    for (size_t i = 0; i < stan::math::size(nu); i++) {
-      const T_partials_return alpha_dbl = value_of(nu_vec[i]) * 0.5;
-      gamma_vec[i] = tgamma(alpha_dbl);
-      digamma_vec[i] = digamma(alpha_dbl);
-      digamma_vec(size_nu);
+  if (!is_constant_all<T_y, T_dof>::value) {
+    for (size_t i = 0; i < size_nu; i++) {
+      const T_partials_return half_nu_dbl = 0.5 * value_of(nu_vec[i]);
+      tgamma_vec[i] = tgamma(half_nu_dbl);
+      if (!is_constant_all<T_dof>::value) {
+        digamma_vec[i] = digamma(half_nu_dbl);
+      }
```
Before it was run only if `T_dof` is an autodiff type, but now if any of `T_y` or `T_dof` is. This change was done so that the `tgamma_vec` could be reused in a term further down that is computed if `T_y` is not constant. On the other hand, the `digamma_vec` term is only ever needed if `T_dof` is an autodiff type, hence the second extra check. Overall, `digamma_vec` is computed as often as before, and `tgamma_vec` potentially more often.

As for moving these check out of for loops, that should be neutral. Both examples below will do the same amount of work as the `include_summand` lines are compile-time checks, and will be removed by the compiler if they always evaluate to false. I tend to prefer the second form (I take Steve does too), as that's what would be used if those ifs were evaluated at runtime.
```
  for (size_t n = 0; n < size_beta; n++) {
    if (include_summand<propto, T_shape, T_inv_scale>::value) {
      log_beta[n] = log(value_of(beta_vec[n]));
```

```
  if (include_summand<propto, T_shape, T_inv_scale>::value) {
    for (size_t n = 0; n < size_beta; n++) {
      log_beta[n] = log(value_of(beta_vec[n]));
```

Sorry for the long message, especially because I'm probably telling you stuff you already know by heart!

**Comment by bbbales2** [comment]
> so the addition of a term in general is not dangerous

Good point.

Our problems are with ```propto = true```

1. The autodiff framework won't work since it assumes ```value_of(f(var(1.0))) == f(1.0)```. This won't work for ```propto = true```

2. We can't even expect ```f<propto = true>(var(1.0)) == f<propto = false>(1.0)``` (binomial has non-parameter terms that can still get dropped).

3. Because we have the true densities (```propto = false```), we can write a test to verify that the normalizing constant is not a function of any of the autodiff variables.

So if ```f(x)``` is a function proportional to the true distribution ```g(x)``` we have:

```
log(g(x)) = K + log(f(x))
```

So we can compute K at a couple different values of x and make sure they're the same. This would have caught Steve's bug (f in that case was constant, but g would have changed).

4. The gradient templating of the log density functions can be checked by verifying the gradients of all combinations of autodiff variables against the finite difference versions with ```propto = false```

5. The cdfs and ccdfs don't deal with propto, so the autodiff tester would be fine for them (not sure we've had any problems with cdfs and ccdfs though).

I think there is a bug in the probability test framework in handling 3 (https://github.com/stan-dev/math/issues/1763). I think the probability test framework should be doing 4 but I didn't check.

> I'm probably telling you stuff you already know by heart!

I don't :D. This is why I'm so scared to move forward with this if it's purely code-style changes. We're really playing with fire.

Do you know how the bug you found in https://github.com/stan-dev/math/issues/1662 got through the testing framework? I see that the glms aren't tested in the testing framework so it's no surprise that bugs made it through there. That's a problem.

**Comment by mcol** [comment]
#1662 was caused by VectorBuilder loops that ran for the wrong number of times if one of the parameters was a vector and the rest were not. The distribution tests admittedly contains test for this case, but perhaps there's a subtle bug there too?

**Comment by bbbales2** [comment]
> #1662 was caused by VectorBuilder loops that ran for the wrong number of times if one of the parameters was a vector and the rest were not

@mcol if you get a chance, mind taking a look at the test framework to see if this is something it tests for? If it doesn't, that's okay, just report back and we'll make an issue. If it does, then presumably there's another bug with the test framework?

**Comment by mcol** [comment]
@bbbales2 There's a `test_repeat_as_vector` function that presumably shoud do that. But (after only a quick skim), it only seems to repeat the same value multiple times, while bug in #1662 would appear if different values were evaluated (as if a vector is not filled up correctly and the only first value is used, this is the same as repeating the first value mutiple times).

**Comment by bbbales2** [comment]
@mcol is there an obvious way to change it so that it goes through multiple values?

And if so, does it catch the #1662 error?

You can run specific probability tests with code like:
```
./runTests.py test/prob/neg_binomial/neg_binomial_cdf_00000_generated_v_test
```

which is much faster than running them all.

syclik:
@bbbales2, mind taking a few minutes to see if your comments have been addressed? (Either confirm the PR shouldn't be in, approve the PR, or remove the review.)

bbbales2:
@syclik yeah sorry.

I originally hesitated on this pull requestion because it touches a lot of difficult code and so I was scared a cleanup for the sake of a cleanup is kinda risky cause something might break.

The fear was that shuffling things around might break something that the tests don't catch. Even though we think our tests are fairly comprehensive, we found a bug in the test framework in the process of reviewing this code: https://github.com/stan-dev/math/issues/1763

But now I'm thinking the reverse logic applies as well. If someone goes through and shuffles around the code they might also find bugs we didn't notice before. And it is useful when people do this. So I think I should accept this if tests pass and the code looks good.

What's your opinion on this? I think I accept it if it passes tests (and [this](https://github.com/stan-dev/math/pull/1764) goes through) and I review it -- I think I only made it part way through the first roung.

bbbales2:
@mcol any chance you could take a look at the `y_dbl == INFTY` things?

mcol:
All those infinity cases were pre-existing, I only touch those lines for variable renames or some similar cleanups. Since those you pointed out are lccdfs, at infinity they will take on the value of log(0), which is negative infinity. At least, that's how I justified to myself that code and kept it around. If that's not correct, then there must be something that we are not testing or not testing correctly.

bbbales2:
@mcol oh yeah you are right on all counts

bbbales2:
Merged in the new develop since this has been sitting awhile (just wanna make sure nothing went funky). I'll merge when the tests pass thanks!

wds15:
@bbbales2 you have reviewed this thoroughly and want to merge it? I am not sure if you can given that you pushed commits, so let me know if you need a hand.

bbbales2:
@wds15 what I did was merge in develop and add a missing `using std::pow` ([here](https://github.com/stan-dev/math/pull/1752/commits/60a574d53adacdda69bf1d93d8ecb003e017bdc8)).

If you wanna click the merge button I'm fine with that. The rest of this is reviewed (as best I could). The discussion in here is mostly two problems with the distribution testing framework we dug up working on this.

SteveBronder:
Should we wait for the fixes to the testing framework before approving this? (i.e. this would wait till the next release most likely)

bbbales2:
@SteveBronder well I was thinking really conservatively on this back in March and that stalled the pull request quite a bit (though we did find a couple problems with the test framework).

I don't think we'll ever be totally happy with the distribution tests. There's at least one fix (https://github.com/stan-dev/math/pull/1764) that'll probably make it in this release (and will presumably catch in merge any #1764 bugs this introduces if it does).

But there are also problems/weaknesses with the tests that won't be fixed this release: https://github.com/stan-dev/math/issues/1978 and https://github.com/stan-dev/math/issues/1976

And that's just the stuff I know about. I'm more leaning now we accept what's passing our tests, otherwise it's kindof an impossible moving bar to hop over.

wds15:
You guys decide on this one, please.

SteveBronder:
imo I'd rather wait till the distribution tests are beefier till we merge this

bbbales2:
Sure, but we should decide what beefier is so we don't just leave it hanging.

SteveBronder:
By "beefier" I just meant merging in #1764. Ben I can look this over if you want another set of eyes but otherwise I'm fine with it (though I'd classify this as not a bugfix so it would wait till the next release)

bbbales2:
We can wait. I reran the tests after 1764 merged and it's green light. We can press the merge button next Tuesday.

rok-cesnovar:
Is this good to go now?

bbbales2:
I'm working on the test framework now so we can just merge it after https://github.com/stan-dev/math/pull/1989 is through.

I think it would pass everything fine, but I am working on the test framework (and there are problems with it I'm finding), so this may as well wait. I'm trying to get that done asap.

rok-cesnovar:
Cool cool. Test away!

bbbales2:
@mcol The distributions in this pull request ended up all getting rewritten before I got the tests in place (still open: https://github.com/stan-dev/math/pull/2085).

I guess this in particular didn't go through because my testing standards lowered between July and now. Apologies for that.

Mind if I close this (I'll go ahead and do it in a few days if I don't hear anything back)?

mcol:
No problem, let's close this now!

Issue #1710, opened 2020-02-13
von Mises circular distribution cdf


bob-carpenter:
## Description

It'd be nice to have the cdf and ccdf for the von Mises distribution.  

This came up in this Discourse topic:  https://discourse.mc-stan.org/t/cdf-of-a-von-mises-distribution/12953

#### Current Version:
v3.1.0

pgree:
I've started to work on this.

PR #1753, opened 2020-02-28
Feature/1710 von mises cdf

pgree:
## Summary

I added a function to compute the cdf of the von Mises distribution.

## Tests

The test evaluates the von Mises cdf at a number of points in the parameter space and checks that those values are equal, to 12 digits, to hard-coded values obtained from scipy's implementation of the same function. 

## Side Effects

None

## Checklist

- [x] Math issue #1710 


stan/math/prim/fun/modified_bessel_first_kind.hpp
```diff
@@ -40,6 +40,12 @@ inline T2 modified_bessel_first_kind(int v, const T2 z) {
   return boost::math::cyl_bessel_i(v, z);
 }
 
+inline double modified_bessel_first_kind(int v, int z) {
+  check_not_nan("modified_bessel_first_kind", "z", z);
+
+  return boost::math::cyl_bessel_i(v, double(z));
+}
+
```
SteveBronder: Why the need for a specialization here? I think you can just change the first one to

```cpp
template <typename T2, require_arithmetic_t<T2>* = nullptr>
inline auto modified_bessel_first_kind(int v, const T2 z) {
  check_not_nan("modified_bessel_first_kind", "z", z);
  return boost::math::cyl_bessel_i(v, static_cast<double>(z));
}
```

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+namespace stan {
+namespace math {
+
+namespace internal {
+
+template <typename T_x, typename T_k, typename T_p>
+return_type_t<T_x, T_k, T_p> von_mises_cdf_series(const T_x& x, const T_k& k,
+                                                  const T_p& p) {
+  using stan::math::cos;
+  using stan::math::sin;
```
SteveBronder: You are already inside of the math namespace so these should be visible. You may actually want to do 
```suggestion
  using std::cos;
  using std::sin;
```

to make sure those are visible to the compiler

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  double pi = stan::math::pi();
+  auto s = sin(x);
+  auto c = cos(x);
+  auto sn = sin(p * x);
+  auto cn = cos(p * x);
+
+  return_type_t<T_x, T_k, T_p> R = 0;
+  return_type_t<T_x, T_k, T_p> V = 0;
```
SteveBronder:
```suggestion
  using return_t = return_type_t<T_x, T_k, T_p>;
  return_t R(0.0);
  return_t V(0.0);
```

I think we usually prefer calling the constructor over the assignment operator

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  int n;
+  for (n = p - 1; n > 0; n--) {
+    auto sn_tmp = sn * c - cn * s;
+    cn = cn * c + sn * s;
+    sn = sn_tmp;
+    R = 1 / (2.0 * n / k + R);
+    V = R * (sn / n + V);
+  }
```
SteveBronder: Can you do `for (int n = p - 1; n > 0; n--) {`?

Also optional, it's usually easier to read code that's done as a forward iterator instead of reverse. idk if that's possible here but would be nice if so

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+  return 0.5 + x / (2 * pi) + V / pi;
+}
+
+template <typename T_x, typename T_k>
+return_type_t<T_x, T_k> von_mises_cdf_normalapprox(const T_x& x, const T_k& k) {
+  using std::exp;
+  using std::sqrt;
+  double pi = stan::math::pi();
```
SteveBronder:
```suggestion
  const double pi = stan::math::pi();
```

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  auto b = sqrt(2 / pi) * exp(k) / modified_bessel_first_kind(0, k);
+  auto z = b * sin(x / 2);
+  double mu = 0;
+  double sigma = 1;
```
SteveBronder: all const

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  return normal_cdf(z, mu, sigma);
+}
+
+template <typename T_x, typename T_k>
+return_type_t<T_x, T_k> von_mises_cdf_centered(const T_x& x, const T_k& k) {
+  // if the scale is sufficiently small, approximate the cdf with a
+  // normal distribution, otherwise, use the expansion from scipy
```
SteveBronder: Usually we don't require docs for internal functions, but this seems to be doing something pretty specific and you have docs already under the signature so it may be good to include docs for this

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+    f = von_mises_cdf_series(x, k, p);
+    if (f < 0)
+      f = 0;
+    if (f > 1)
+      f = 1;
+  } else {
+    f = von_mises_cdf_normalapprox(x, k);
+  }
+  return f;
+}
```
SteveBronder: How about just hardcoding the stuff above?
```suggestion
using return_t = return_type_t<T_x, T_k>;
return_t f; // Should f be initialized to some value?
if (k < 50.0) {
  const auto p = value_of_rec(28.0 + 0.5 * k - 100.0 / (k + 5.0) + 1.0);
  f = von_mises_cdf_series(x, k, p);
  if (f < 0) {
    f = 0.0
    return f;
  } else if (f > 1.0) {
    f = 1.0;
    return f;
  }
} else {
  f = von_mises_cdf_normalapprox(x, k);
  return f;
}
return f;
}
```

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * @param k The inverse scale of the distriubtion
+ * @return The von Mises cdf evaluated at the specified arguments
+ * @tparam T_x Type of x
+ * @tparam T_mu Type of mean parameter
+ * @tparam T_k Type of inverse scale parameter
+ */
+template <typename T_x, typename T_mu, typename T_k>
+inline return_type_t<T_x, T_mu, T_k> von_mises_cdf(const T_x& x, const T_mu& mu,
+                                                   const T_k& k) {
+  static char const* const function = "von_mises_cdf";
```
SteveBronder:
```suggestion
  static char const* function = "von_mises_cdf";
```

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+  using T_partials_return = partials_return_t<T_x, T_mu, T_k>;
+  using internal::von_mises_cdf_centered;
+
+  check_not_nan(function, "Random variable", x);
+  check_not_nan(function, "Scale parameter", k);
+  check_not_nan(function, "Location parameter", mu);
+  check_positive(function, "Scale parameter", k);
+
+  // shift x so that mean is 0
+  return_type_t<T_x, T_mu, T_k> x2 = x - mu;
```
SteveBronder:
```suggestion
  T_partials_return x2 = x - mu;
```

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  // x is on an interval (2*n*pi, (2*n + 1)*pi), move it to (-pi, pi)
+  double pi = stan::math::pi();
+  x2 += pi;
+  auto x_floor = floor(x2 / (2 * pi));
+  auto x_moded = x2 - x_floor * 2 * pi;
+  x2 = x_moded - pi;
```
SteveBronder:
```suggestion
  const double pi = stan::math::pi();
  x2 += pi;
  const auto x_floor = floor(x2 / (2 * pi));
  const auto x_moded = x2 - x_floor * 2 * pi;
  x2 = x_moded - pi;
```

stan/math/prim/fun/modified_bessel_first_kind.hpp
bbbales2: @pgree was talking to me about this yesterday.

Apparently ```boost::math::cyl_bessel_i``` explodes if the second argument is an int, but it's fine to cast it to a double (and that makes sense).

I think we need the other version in case we pass in something that'll get autodiffed (fvars, or whatnot). I think what you have here is a suitable way to code the alternative.

The need for the alternative should definitely be doc'd in the function comments though.

stan/math/prim/prob/von_mises_cdf.hpp
bbbales2: Do ```--n``` too. I think we favor that style?

stan/math/prim/prob/von_mises_cdf.hpp
bbbales2: Also we're in stan::math still so let's just use ```pi()``` everywhere

stan/math/prim/fun/modified_bessel_first_kind.hpp
SteveBronder: Cool beans feel free to chose this after docs are added

stan/math/prim/prob/von_mises_cdf.hpp
SteveBronder: it's nanoseconds faster but I'm indifferent either way

bbbales2: I feel the same.

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+  check_not_nan(function, "Random variable", x);
+  check_not_nan(function, "Scale parameter", k);
+  check_not_nan(function, "Location parameter", mu);
+  check_positive(function, "Scale parameter", k);
+
+  // shift x so that mean is 0
+  return_type_t<T_x, T_mu, T_k> x2 = x - mu;
+
+  // x is on an interval (2*n*pi, (2*n + 1)*pi), move it to (-pi, pi)
+  double pi = stan::math::pi();
```
bbbales2: I'm not sure we want the cdf to be periodic. I'm not sure why this doesn't have the support cleanly defined. @bgoodri do you have an opinion here? This seems a bit sketchy. But also saying the support is ```(mu - pi, to mu + pi)``` is weird cause then the support is a function of the parameters.

There's a little note in the docs, but it's also weirdly vague: https://mc-stan.org/docs/2_22/functions-reference/von-mises-distribution.html (starting with: "In order for this density to properly normalize")

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+  x2 += pi;
+  auto x_floor = floor(x2 / (2 * pi));
```
bbbales2: Looks like TWO_PI is a constant we can use: https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/von_mises_lpdf.hpp#L70

stan/math/prim/prob/von_mises_cdf.hpp
bbbales2: Yeah this have docs. This looks like the von-mises for mu = 0 and assuming a strict -pi to pi support right?

Just copy paste the docs from the other function and start with that. You don't need to repeat the equations in here.

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+#include <stan/math/prim/fun/modified_bessel_first_kind.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+namespace internal {
+
+template <typename T_x, typename T_k, typename T_p>
+return_type_t<T_x, T_k, T_p> von_mises_cdf_series(const T_x& x, const T_k& k,
```
bbbales2: Add short doc to describe where this series thing is useful, if possible put a citation

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,124 @@
...
+    cn = cn * c + sn * s;
+    sn = sn_tmp;
+    R = 1 / (2.0 * n / k + R);
+    V = R * (sn / n + V);
+  }
+  return 0.5 + x / (2 * pi) + V / pi;
+}
+
+template <typename T_x, typename T_k>
+return_type_t<T_x, T_k> von_mises_cdf_normalapprox(const T_x& x, const T_k& k) {
```
bbbales2: Add short doc, also a citation if possible. It's hard to understand why code that is doing a numerical trick is doing that trick otherwise.

stan/math/prim/prob/von_mises_cdf.hpp
bbbales2: We don't want the std ones though? We really do want the stan::math ones. Just leaving stan::math out is enough.

test/unit/math/prim/prob/von_mises_cdf_test.cpp
```diff
@@ -0,0 +1,189 @@
+#include <stan/math/prim.hpp>
+#include <stan/math/prim/fun/constants.hpp>
+#include <gtest/gtest.h>
+
+TEST(ProbVonMises, random_cdf_test) {
+  using stan::math::von_mises_cdf;
+  double pi = stan::math::pi();
+  // check that our von_mises_cdf coincides with scipy's
+  double ABS_TOL = 1e-12;
+  ASSERT_NEAR(0.0000157567385742, von_mises_cdf(-3.141492653589793, 0, 0.01),
```
bbbales2: The prim tests should test all the throws happen in the right cases: https://github.com/stan-dev/math/blob/f24720ca194057e7e27e8306cc6cb09a23c25b2f/stan/math/prim/prob/von_mises_cdf.hpp#L103

Like: https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/fun/diag_post_multiply_test.cpp#L40

test/unit/math/rev/prob/von_mises_cdf_test.cpp
```diff
@@ -0,0 +1,49 @@
+#include <stan/math/rev.hpp>
```
bbbales2: I think the mix tests cover this file. I think we can delete this.

stan/math/prim/prob/von_mises_cdf.hpp
pgree: Yes, that's right. Will add the docs.

stan/math/prim/prob/von_mises_cdf.hpp
pgree: The docs are a little unclear and also don't reflect what the pdf is actually doing. I made the cdf periodic in order to be consistent with the pdf. We could just define all the von mises functions so that they have mu=0 and support on (-pi, pi).

bbbales2:
How annoying would it be if I said let's not do the normal approximation? I don't think what scipy is doing is great.

Try this, for instance:
```
>>> import scipy.stats
>>> a = 49.99999999999999
>>> print('{:.20}'.format(a))
49.999999999999992895
>>> b = 49.99999999999998
>>> print('{:.20}'.format(b))
49.999999999999978684
>>> scipy.stats.distributions.vonmises.cdf(0.16534171860998947, a)
0.87796123244401891
>>> scipy.stats.distributions.vonmises.cdf(0.16534171860998947, b)
0.87796123244401891
>>> scipy.stats.distributions.vonmises.cdf(0.16534171860998947, 50)
0.8779632968295088
```

Either that or we should smooth it out. Maybe interpolate between k = 49 and k = 50.

Is this approximation because the numerics of the series blow up for k > 50 or is this a performance thing? If it's numerics, I'm tempted to try to figure out the interpolation. If it's performance, don't worry about this approximation.

Ofc. I don't know if that really matters, but these are just the sorts of things that will probably never get checked again once they're in hence the nitpickyness.

test/unit/math/prim/prob/von_mises_cdf_test.cpp
```diff
@@ -0,0 +1,247 @@
...
+  }
+  sum = sum / ((n - 1) / (xmax - xmin));
+
+  // and compare to cdf
+  double val2 = von_mises_cdf(xmax, mu, k) - von_mises_cdf(xmin, mu, k);
+
+  ASSERT_NEAR(val2, sum, abs_tol_int);
+}
+
+TEST(ProbVonMises, pointwise_cdf_test) {
```
bbbales2, line 42: It looks like the tests to make sure prim is throwing in the right places are missing.

All these checks should get checked that they fail in at least one way:

```
check_not_nan(function, "Random variable", x);
check_not_nan(function, "Scale parameter", k);
check_not_nan(function, "Location parameter", mu);
check_positive(function, "Scale parameter", k);
check_less_or_equal(function, "Random variable", x, pi);
check_greater_or_equal(function, "Random variable", x, -pi);
```

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,172 @@
...
+  check_positive(function, "Scale parameter", k);
+  check_less_or_equal(function, "Random variable", x, pi);
+  check_greater_or_equal(function, "Random variable", x, -pi);
+
+  // shift x so that mean is 0
+  return_t x2 = x - mu;
+
+  // x is on an interval (2*n*pi, (2*n + 1)*pi), move it to (-pi, pi)
+  x2 += pi;
+  const auto x_floor = floor(x2 / TWO_PI);
```
bbbales2, line 139: Do we still need this modulus logic?

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,172 @@
...
+ * https://github.com/scipy/scipy/blob/8dba340293fe20e62e173bdf2c10ae208286692f/scipy/stats/vonmises.py
+ *
+ * When k < 50, approximate the von Mises cdf
+ * with the following expansion that comes from
+ * scipy.
+ */
+template <typename T_x, typename T_k>
+return_type_t<T_x, T_k> von_mises_cdf_series(const T_x& x, const T_k& k) {
+  const double pi = stan::math::pi();
+  int p = value_of_rec(28 + 0.5 * k - 100 / (k + 5) + 1);
```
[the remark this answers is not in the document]
pgree: Rounding down should be fine here

stan/math/prim/prob/von_mises_cdf.hpp
```diff
@@ -0,0 +1,172 @@
...
+ * the cdf with a Gaussian. Otherwise, use the expansion from scipy.
+ */
+template <typename T_x, typename T_k>
+return_type_t<T_x, T_k> von_mises_cdf_centered(const T_x& x, const T_k& k) {
+  double ck = 50;
+  using return_t = return_type_t<T_x, T_k>;
+  return_t f;
+  if (k < 49) {
+    f = von_mises_cdf_series(x, k);
+    if (f < 0) {
```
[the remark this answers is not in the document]
pgree: these are both relics of the scipy implementation. I haven't checked  whether these conditions are ever actually met or not. I did test that the CDF agrees with the lpdf already  in Stan and they do to at least ~7 digits.

stan/math/prim/prob/von_mises_cdf.hpp
pgree: unfortunately, I think we still do. I played around with it a bit in the hopes of deleting some of these lines, but I could not find a good solution.

pgree: The comment is now wrong, though, and I'll fix that

bbbales2:
@pgree I just realized this function doesn't take vectorized things. Let me just take this pull over and put all this stuff in. Also right now all our CDFs are tested in math/test/prob/ (check [here](https://github.com/stan-dev/math/blob/develop/test/prob/normal/normal_cdf_test.hpp) for instance).

It's gonna be easier if I just code all this stuff up and then you or someone else reviews it. The numerical stuff make sense and the rest of this is just annoying boilerplate.

bbbales2:
Sorry for the long delay, finally got back to this. I pushed three files to the probability test directory.

They are (have a quick look at them to see what they're doing):
```
test/prob/von_mises/von_mises_cdf_test.hpp
test/prob/von_mises/von_mises_cdf_log_test.hpp
test/prob/von_mises/von_mises_ccdf_log_test.hpp
```

They feed into the probability test framework, which calls pdfs/cdf/lcdfs/lccdfs with all combinations of arguments to make sure they work.

To run the basic tests you can do:
```
./runTests.py test/prob/von_mises/von_mises_cdf_00000_generated_v_test
./runTests.py test/prob/von_mises/von_mises_cdf_log_00000_generated_v_test
./runTests.py test/prob/von_mises/von_mises_ccdf_log_00000_generated_v_test
```

From this, I think the immediate things that are obvious are we should add a cdf_log implementation and a ccdf_log implementation. If these are wrappers around cdf, so be it, but they should probably exist to make this consistent with all the other interfaces.

When I ran ```./runTests.py test/prob/von_mises/von_mises_cdf_00000_generated_v_test```, it looks like the lpdf there wasn't accepting arrays for one of the arguments properly. The error lists are long cause it's compiling a bajillion tests. When you find an error here, it might be easier to try to re-create the error in the other unit tests and then develop those till they pass (in which case the simpler unit tests are better as well), but if you just want to work against these tests that's fine.

When you have those three tests working, you can run:
```
./runTests.py test/prob/von_mises/
```

And that'll test a bunch more stuff on those same files.

bbbales2:
@pgree: @bob-carpenter mentioned over [here](https://github.com/stan-dev/math/issues/1976) that all the pdfs/cdfs are dong with operands_and_partials ([here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/bernoulli_cdf.hpp) for instance).

That basically means we've written in custom gradients for everything.

Can we do that here too?

bob-carpenter:
I did not mean to imply all the cdfs are using operands-and-partials.  Just that it was a case where it'd make sense to write a simpler implementation to test a more complex operands-and-partials version or a more complex version that drops constants, etc.

The log cdfs just wrap the cdfs in a call to log.  Those could really use some more attention.  The log ccdfs just use log1m.

bbbales2:
I suppose if it's too hard here we won't do it.

pgree:
> I suppose if it's too hard here we won't do it.

I looked into this when I started writing this CDF. I determined at the time that it wasn't worth it since the partial with respect to the scale parameter is an ugly integral.

bbbales2:
Cool, we'll leave it autodiffed then.

syclik:
@pgree and @bbbales2, I just fixed the merge conflict (I hope). After this passes, @bbbales2, can you walk through your review and see what else needs to be addressed?

bbbales2:
This is failing the vectorized distribution checks. @pgree we can meet and discuss weee

bbbales2:
@pgree I vectorized the cdf implementation (in a very lazy way) and I also added lazy lcdf and lccdf implementations. I figure it's better to have them than not have them. If someone wants to do better later they are more than welcome.

I also added a bunch of tests to `test/prob`. These are failing and need some work.

So if you run:
```
./runTests.py test/prob/von_mises/von_mises_cdf_00000_generated_fd_test.cpp
```

And then you look at the output errors you will see the output.

It will probably be fastest if we work through this together. I'll share my screen and you can advise me on where the errors might be.

bbbales2:
I wasn't able to figure o ut the last thing with the lccdf. We're gonna have to do a round 2 codin' session

bbbales2:
@pgree remember to check out doing a von mises log cdf/log ccdfs. It should be possible to replace the log(cdf) terms with log_cdf terms at least when the normal approximation kicks in.

The tests to run are:
```
./runTests.py test/prob/von_mises/von_mises_cdf_00000_generated_v_test.cpp
./runTests.py test/prob/von_mises/von_mises_cdf_log_00000_generated_v_test.cpp
./runTests.py test/prob/von_mises/von_mises_ccdf_log_00000_generated_v_test.cpp
```

But if you hit numerical difficulties it might be faster to copy the mix `von_mises_cdf` tests (since these probability tests take a crazy long time to compile)

bbbales2:
@pgree yoyoyo want to pair program this?

pgree:
> @pgree yoyoyo want to pair program this?

I do!

bbbales2:
@serban-nicusor-toptal I merged develop into this a couple times but the Jenkins test never shows up. I can just close a reopen this pull. Is that the best way to get this rolling again?

serban-nicusor-toptal:
Hey @bbbales2 it should work fine now. For some reason, it went into a deadlock for this job, sorry for the inconvenience.

bbbales2 (commented):
I fixed a thing with one test. I'm gonna merge this if it passes tests. I think everything is here.

bbbales2:
@pgree you happy with this last test change? If so I'll merge it

bbbales2:
This [one](https://github.com/stan-dev/math/pull/1753/commits/8a2c5892387b1b9805abc85e67f8f643c680b049)

pgree:
> @pgree you happy with this last test change? If so I'll merge it

Yes, I agree that's a better test

bbbales2 (approved):

PR #1754, opened 2020-03-02
Generalize */fun starting with cr-d

t4c1:
## Summary

Generalizes functions with names starting with cr-d.

## Tests
Some checks that an input is a vector are now done at compile time instead of runtime. Tests fpr these checks are removed.

## Side Effects

None.

## Checklist

- [x] Math issue #1470


stan/math/fwd/fun/tcrossprod.hpp
```diff
@@ -1,20 +1,22 @@
...
 
 namespace stan {
 namespace math {
 
-template <typename T, int R, int C>
-inline Eigen::Matrix<fvar<T>, R, R> tcrossprod(
-    const Eigen::Matrix<fvar<T>, R, C>& m) {
+template <typename T, require_eigen_vt<is_fvar, T>* = nullptr>
+inline Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
+                     T::RowsAtCompileTime>
```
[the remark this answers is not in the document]
SteveBronder: fyi everywhere T is an Eigen Matrix would be good to have a nicer name like `EigMat`

stan/math/prim/fun/squared_distance.hpp
```diff
@@ -12,13 +12,14 @@ namespace math {
...
  *
- * @param x1 First vector.
- * @param x2 Second vector.
- * @return Dot product of the vectors.
- * @throw std::domain_error If the vectors are not the same
- * size or if they are both not vector dimensioned.
+ * @param x1 First scalar.
+ * @param x2 Second scalar.
+ * @return Squared distance between scalars
+ * @throw std::domain_error Any scalar is not finite.
```
SteveBronder: Would you mind documenting T1 and T2 as well. 

[Optionally] Could also make the names something nicer like `Scalar1` and `Scalar2`

stan/math/rev/fun/dot_self.hpp
```diff
@@ -19,34 +19,16 @@ class dot_self_vari : public vari {
...
-  }
-  template <int R, int C>
-  explicit dot_self_vari(const Eigen::Matrix<var, R, C>& v)
-      : vari(var_dot_self(v)), size_(v.size()) {
+      : vari(Eigen::Map<vector_vi>(v, size).val().squaredNorm()),
+        v_(v),
+        size_(size) {}
+  template <typename T, require_eigen_t<T>* = nullptr>
+  explicit dot_self_vari(const T& v)
+      : vari(v.val().squaredNorm()), size_(v.size()) {
```
SteveBronder: Nice!

stan/math/rev/fun/squared_distance.hpp
```diff
@@ -130,27 +124,23 @@ class squared_distance_vd_vari : public vari {
...
 }
-template <int R1, int C1, int R2, int C2>
-inline var squared_distance(const Eigen::Matrix<var, R1, C1>& v1,
-                            const Eigen::Matrix<double, R2, C2>& v2) {
-  check_vector("squared_distance", "v1", v1);
-  check_vector("squared_distance", "v2", v2);
+template <typename T1, typename T2,
+          require_eigen_vector_vt<is_var, T1>* = nullptr,
+          require_eigen_vector_vt<std::is_arithmetic, T2>* = nullptr>
+inline var squared_distance(const T1& v1, const T2& v2) {
```
SteveBronder: [Optional] For something like this I think `EigenVar` and `EigenArith` would be useful names

test/unit/math/prim/fun/dot_self_test.cpp
```diff
@@ -56,21 +56,4 @@ TEST(MathMatrixPrimMat, dot_self) {
   Eigen::Matrix<double, 1, Eigen::Dynamic> rv3(3);
   rv3 << 2.0, 3.0, 4.0;
   EXPECT_NEAR(29.0, dot_self(rv3), 1E-12);
-
-  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> m1(1, 1);
-  m1 << 2.0;
-  EXPECT_NEAR(4.0, dot_self(m1), 1E-12);
```
SteveBronder: Why gone?

SteveBronder: Oh because it only takes in vectors now? I wonder if this will break user code tho?

andrjohns:
The rev stuff all looks pretty great. Should we also replace the:
```
reinterpret_cast<vari**>(ChainableStack::instance_->memalloc_.alloc()
```
with 
```
ChainableStack::instance_->memalloc_.alloc_array<vari*>();
```
Or save that for a separate issue?

stan/math/fwd/fun/determinant.hpp
```diff
@@ -1,19 +1,20 @@
...
 
-template <typename T, int R, int C>
-inline fvar<T> determinant(const Eigen::Matrix<fvar<T>, R, C>& m) {
+template <typename T, require_eigen_vt<is_fvar, T>* = nullptr>
+inline value_type_t<T> determinant(const T& m) {
   check_square("determinant", "m", m);
 
-  const T vals = m.val().determinant();
-  return fvar<T>(vals, vals * (m.val().inverse() * m.d()).trace());
+  const typename value_type_t<T>::Scalar vals = m.val().determinant();
```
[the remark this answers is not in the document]
t4c1: Nope. `value_type_t<T>` is `fvar<T>` and I need `T` here.

stan/math/fwd/fun/determinant.hpp
```diff
@@ -1,19 +1,20 @@
...
 #include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/fwd/core.hpp>
 
 namespace stan {
 namespace math {
 
-template <typename T, int R, int C>
-inline fvar<T> determinant(const Eigen::Matrix<fvar<T>, R, C>& m) {
+template <typename T, require_eigen_vt<is_fvar, T>* = nullptr>
+inline value_type_t<T> determinant(const T& m) {
```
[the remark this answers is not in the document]
t4c1: done

stan/math/fwd/fun/tcrossprod.hpp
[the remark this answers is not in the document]
t4c1: Done. I know about template names, but they are not very high on my priority list, so I tend to forget to change them. You can either keep reminding me about every single instance I miss or we can leave that for a separate series of PRs.

stan/math/prim/fun/squared_distance.hpp
t4c1: done

stan/math/fwd/fun/tcrossprod.hpp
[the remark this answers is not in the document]
t4c1: One of the return statements is invoking default constructor, so compiler can not deduce `auto`.

stan/math/rev/fun/squared_distance.hpp
t4c1: done

t4c1:
I would prefere to leave this one out of this one. I am already doing quite a few things at once here.

stan/math/fwd/fun/tcrossprod.hpp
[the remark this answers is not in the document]
andrjohns: Except at moment the function will take any eigen type as an input but can only return ```Eigen::Matrix``` types (which could cause some compile errors down the line). You could restrict the inputs to only take matrix types, or since the result will need to be evaluated anyway, this could be written to be more flexible:
```
template <typename T, require_eigen_vt<is_fvar, T>* = nullptr>
inline auto tcrossprod(const T& m) {
  using T_return = decltype(multiply(m, m.transpose()));
  if (m.rows() == 0) {
    return plain_type_t<T_return>{};
  }
  return multiply(m, m.transpose()).eval();
}
```

test/unit/math/prim/fun/dot_self_test.cpp
t4c1: Right. I should probably let this accept matrices too, even if matrix version is not exposed to stan language.

stan/math/fwd/fun/tcrossprod.hpp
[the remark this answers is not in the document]
t4c1: Many functions still assume input is derived from `Eigen::MatrixBase` and will not work with `Eigen::ArrayBase`-derived types. Or are you thinking about some other type this would fail with?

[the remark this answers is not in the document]
andrjohns: No it was just ```ArrayBase``` that I was thinking of, so if that's a non-issue then feel free to ignore

andrjohns:
@t4c1 There's a problem with using ```Eigen::Ref``` for some expression types, since the default Stride length can be wrong. You can verify in your branch with this test for ```cumulative_sum```:
```
TEST(MathMatrixPrimMat, cumulative_sum_expr) {
  using stan::math::cumulative_sum;

  stan::math::vector_d x2(2);
  x2 << -1.0, 1.0;
  stan::math::matrix_d m2(2, 2);
  m2 << -1.0, 1.0, -1.0, 1.0;

  stan::math::vector_d x2_out = cumulative_sum(x2);
  stan::math::vector_d m2_out = cumulative_sum(cumulative_sum(m2.diagonal()));

  EXPECT_FLOAT_EQ(x2_out[1],m2_out[1]);
}
```

Which returns:
```
test/unit/math/prim/fun/cumulative_sum_test.cpp:61: Failure
  x2_out[1]
    Which is: 0
  m2_out[1]
    Which is: -1
```

The only fix I've found is to change the ```Ref``` call to ```Ref<const T,0,InnerStride<> >&```, but this ends up disabling vectorisation ([more info in the Eigen doc here](https://eigen.tuxfamily.org/dox/classEigen_1_1Ref.html)). Not sure if there's a better approach to be had here

andrjohns:
Noticed the error after I hit send, the test has a typo (```cumulative_sum``` called twice). Looks like I'm chasing down a different error.

andrjohns:
Alright, I wasn't entirely crazy, there is an issue with ```Eigen::Ref``` and expressions as inputs, I've got a minimal example [here on godbolt](https://godbolt.org/z/gG4BDM) that replicates the problem. It may have something to do with the ```Eigen::Ref``` object falling out of scope, would be my hunch.

bob-carpenter:
With the example:

```
decltype(auto) test_fun(const T& x) {
    const Eigen::Ref<const Eigen::VectorXd>& x_ref = x;
    return (x_ref.array() - 5.0).matrix();
}
```

I'm not exactly sure what the reference structure for expressions is, but I'm guessing that `x_ref.array()` creates a local structure and then the `- 5` creates an expression that references it.  That array goes out of scope when the function returns.  The local that gets created for `.matrix()` holds a reference to the subtraction expression, and that will also go away.  The matrix created by `.matrix()` will be an expression, but it gets copied.  It's the references it's holding that can go bad.

Also, why is `x_ref` being defined as a `Ref` and as a reference (`&`)?

andrjohns:
That makes sense, going to make generalising these functions safely an interesting proposition.

> Also, why is ```x_ref``` being defined as a ```Ref``` and as a reference (```&```)?

More info on that [here](https://stackoverflow.com/a/21138316/4362612)

andrjohns:
One solution I've found is to detect whether an expression has been passed to the function, and if so, evaluate the expression prior to calling the function: https://godbolt.org/z/jqoFZi

Whether this is the best (or most efficient) solution is another question

t4c1:
First thanks for finding this. I don't like your suggestion. It gives up all the benefits of using `Eigen::Ref` in the first place. 

I guess we could instead place that `Eigen::Ref` on the nochain stack. Than it would not go out of scope. This gives up on using references (`&`), but I think reconstructing `Eigen::Ref`s is not very expensive.

Also it might be better to put this discussion on `Eigen::Ref` in a separate issue.

bob-carpenter:
Yes, passing Eigen structures around efficiently deserves its own independent discussion. 

I don't see a good option.  The two under consideration are.

1.  Do an `eval()` and lose the benefits of `Ref` by creating a copy.  This is what we're doing now implicitly by making arguments const references to `Eigen::Matrix`.

2.  Use placement new on the autodiff stack for every (sub)expression template used in a function.  This leaks memory because the locals never get recovered.  And I don't see how to do everything in a nicely readable way with placement new because you need to know the type.

What we really need is a kind of inlining through expression templates.

t4c1:
@SteveBronder this is waiting for another review. The issue with local variables only blocks returning expressions.

stan/math/prim/fun/squared_distance.hpp
```diff
@@ -29,45 +33,24 @@ inline return_type_t<T1, T2> squared_distance(const T1& x1, const T2& x2) {
...
  */
-template <int R1, int C1, int R2, int C2>
-inline double squared_distance(const Eigen::Matrix<double, R1, C1>& v1,
-                               const Eigen::Matrix<double, R2, C2>& v2) {
-  check_vector("squared_distance", "v1", v1);
-  check_vector("squared_distance", "v2", v2);
+template <typename EigVec1, typename EigVec2,
+          require_all_eigen_vector_t<EigVec1, EigVec2>* = nullptr,
+          require_all_not_eigen_vt<is_var, EigVec1, EigVec2>* = nullptr>
+inline return_type_t<EigVec1, EigVec2> squared_distance(const EigVec1& v1,
```
SteveBronder: Can you double check these? It looks like you want
```suggestion
          require_all_eigen_vector_t<EigVec1, EigVec2>* = nullptr,
          require_all_not_var<value_type_t<EigVec1>, value_type_t<EigVec2>>* = nullptr>
inline return_type_t<EigVec1, EigVec2> squared_distance(const EigVec1& v1,
```

Since the first checks for eigen types but the second fails for eigen types

SteveBronder: (this week I'm going to start fixing up the requires so these things are more clear)

stan/math/prim/fun/tcrossprod.hpp
```diff
@@ -11,21 +12,24 @@ namespace math {
...
- * @tparam C number of columns, can be Eigen::Dynamic
+ * @tparam T type of the matrix (must be derived from \c Eigen::MatrixBase)
  * @param M Matrix to multiply.
  * @return M times its transpose.
  */
-template <int R, int C>
-inline Eigen::MatrixXd tcrossprod(const Eigen::Matrix<double, R, C>& M) {
+template <typename T, require_eigen_vt<std::is_arithmetic, T>* = nullptr>
+inline Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
+                     T::RowsAtCompileTime>
```
SteveBronder: [optional]

If you want to make this a bit prettier I think you could declare this type in the template like

```
, ReturnType = Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
                     T::RowsAtCompileTime>>
```

stan/math/rev/fun/squared_distance.hpp
```diff
@@ -62,29 +58,23 @@ class squared_distance_vv_vari : public vari {
...
+      require_all_eigen_vector_vt<is_var, EigVecVar1, EigVecVar2>* = nullptr>
+  squared_distance_vv_vari(const EigVecVar1& v1, const EigVecVar2& v2)
+      : vari((as_column_vector_or_scalar(v1).val()
+              - as_column_vector_or_scalar(v2).val())
+                 .squaredNorm()),
+        length_(v1.size()) {
     v1_ = reinterpret_cast<vari**>(
         ChainableStack::instance_->memalloc_.alloc(length_ * sizeof(vari*)));
     v2_ = reinterpret_cast<vari**>(
         ChainableStack::instance_->memalloc_.alloc(length_ * sizeof(vari*)));
```
SteveBronder: [optional]

I think these can be replaced by alloc_array<vari*>

stan/math/rev/fun/squared_distance.hpp
```diff
@@ -130,27 +115,26 @@ class squared_distance_vd_vari : public vari {
...
-inline var squared_distance(const Eigen::Matrix<var, R1, C1>& v1,
-                            const Eigen::Matrix<var, R2, C2>& v2) {
-  check_vector("squared_distance", "v1", v1);
-  check_vector("squared_distance", "v2", v2);
+template <
+    typename EigVecVar1, typename EigVecVar2,
+    require_all_eigen_vector_vt<is_var, EigVecVar1, EigVecVar2>* = nullptr>
+inline var squared_distance(const EigVecVar1& v1, const EigVecVar2& v2) {
   check_matching_sizes("squared_distance", "v1", v1, "v2", v2);
   return var(new internal::squared_distance_vv_vari(v1, v2));
```
SteveBronder: [optional]

return type can let this be 
```suggestion
  return {new internal::squared_distance_vv_vari(v1, v2)};
```

stan/math/fwd/fun/determinant.hpp
```diff
@@ -1,19 +1,20 @@
...
-template <typename T, int R, int C>
-inline fvar<T> determinant(const Eigen::Matrix<fvar<T>, R, C>& m) {
+template <typename EigMat, require_eigen_vt<is_fvar, EigMat>* = nullptr>
+inline value_type_t<EigMat> determinant(const EigMat& m) {
   check_square("determinant", "m", m);
 
-  const T vals = m.val().determinant();
-  return fvar<T>(vals, vals * (m.val().inverse() * m.d()).trace());
+  const typename value_type_t<EigMat>::Scalar vals = m.val().determinant();
+  return value_type_t<EigMat>(vals, vals * (m.val().inverse() * m.d()).trace());
```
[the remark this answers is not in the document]
t4c1: fixed

stan/math/fwd/fun/tcrossprod.hpp
```diff
@@ -1,20 +1,22 @@
...
 namespace stan {
 namespace math {
 
-template <typename T, int R, int C>
-inline Eigen::Matrix<fvar<T>, R, R> tcrossprod(
-    const Eigen::Matrix<fvar<T>, R, C>& m) {
+template <typename EigMat, require_eigen_vt<is_fvar, EigMat>* = nullptr>
+inline Eigen::Matrix<value_type_t<EigMat>, EigMat::RowsAtCompileTime,
+                     EigMat::RowsAtCompileTime>
+tcrossprod(const EigMat& m) {
```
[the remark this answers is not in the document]
t4c1: I don't think such a type trait would be used that much. We can't use auto as this has two return statements (unless we used some decltype magic for the first one, but that would be ugly).

stan/math/prim/fun/crossprod.hpp
```diff
@@ -11,11 +12,13 @@ namespace math {
...
  *
+ * @tparam T type of the matrix (must be derived from \c Eigen::MatrixBase)
  * @param M Matrix to multiply.
  * @return Transpose of M times M
  */
-inline matrix_d crossprod(const matrix_d& M) {
-  return tcrossprod(static_cast<matrix_d>(M.transpose()));
+template <typename T, require_eigen_t<T>* = nullptr>
+inline auto crossprod(const T& M) {
+  return tcrossprod(M.transpose());
```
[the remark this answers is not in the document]
t4c1: fixed

stan/math/prim/fun/cumulative_sum.hpp
```diff
@@ -31,28 +32,32 @@ inline std::vector<T> cumulative_sum(const std::vector<T>& x) {
...
- * @param m Matrix of values.
+ * @param m Vector of values.
  * @return Cumulative sum of values.
  */
-template <typename T, int R, int C>
-inline Eigen::Matrix<T, R, C> cumulative_sum(const Eigen::Matrix<T, R, C>& m) {
-  Eigen::Matrix<T, R, C> result(m.rows(), m.cols());
+template <typename T, require_eigen_vector_t<T>* = nullptr>
+inline Eigen::Matrix<value_type_t<T>, T::RowsAtCompileTime,
+                     T::ColsAtCompileTime>
```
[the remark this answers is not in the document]
t4c1: fixed

stan/math/prim/fun/diag_matrix.hpp
```diff
@@ -10,13 +11,14 @@ namespace math {
...
+ * have one compile time dimmension equal to 1)
  * @param[in] v Specified vector.
  * @return Diagonal matrix with vector as diagonal values.
  */
-template <typename T>
-inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> diag_matrix(
-    const Eigen::Matrix<T, Eigen::Dynamic, 1>& v) {
+template <typename T, require_eigen_vector_t<T>* = nullptr>
+inline Eigen::Matrix<value_type_t<T>, Eigen::Dynamic, Eigen::Dynamic>
+diag_matrix(const T& v) {
```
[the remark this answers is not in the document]
t4c1: fixed

stan/math/prim/fun/distance.hpp
```diff
@@ -30,24 +31,20 @@ inline return_type_t<T1, T2> distance(const T1& x1, const T2& x2) {
...
-inline return_type_t<T1, T2> distance(const Eigen::Matrix<T1, R1, C1>& v1,
-                                      const Eigen::Matrix<T2, R2, C2>& v2) {
+template <typename T1, typename T2,
+          require_all_eigen_vector_t<T1, T2>* = nullptr>
+inline return_type_t<T1, T2> distance(const T1& v1, const T2& v2) {
   using std::sqrt;
-  check_vector("distance", "v1", v1);
-  check_vector("distance", "v2", v2);
   check_matching_sizes("distance", "v1", v1, "v2", v2);
   return sqrt(squared_distance(v1, v2));
```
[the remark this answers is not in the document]
t4c1: What eigen calls? `squared_distance` returns a scalar.

stan/math/prim/fun/squared_distance.hpp
t4c1: Both sets of requires actually do the same thing! Since first one requires eigen types the second one only fails if the value types do not match. I did it this way as it is a bit shorter. If you think it is not clear I can still change it to what you suggested.

stan/math/prim/fun/tcrossprod.hpp
t4c1: Personally I would prefer not to put another thing in template lists. They already contain actual templates and requires. So in this case I think the code being clearer outweighs some code duplication. If you disagree I can still change that.

stan/math/rev/fun/squared_distance.hpp
t4c1: As I said earlier I am already changing quite a few things at once. If we do too much at once there is larger chance that something is missed. So I would leave these for a separate series of PRs.

stan/math/rev/fun/squared_distance.hpp
t4c1: fixed

t4c1:
@SteveBronder This is ready for review.

stan/math/prim/fun/determinant.hpp
```diff
@@ -10,16 +11,14 @@ namespace math {
...
  * @return Determinant of the matrix.
  * @throw std::domain_error if matrix is not square.
  */
-template <typename T, int R, int C>
-inline T determinant(const Eigen::Matrix<T, R, C>& m) {
+template <typename T, require_eigen_vt<std::is_arithmetic, T>* = nullptr>
+inline value_type_t<T> determinant(const T& m) {
   check_square("determinant", "m", m);
   if (m.size() == 0) {
     return 1;
```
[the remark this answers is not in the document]
t4c1: Caring about rvo when returning a scalar seems a bit pointless. I will check if removing if breaks anything.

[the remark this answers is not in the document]
SteveBronder: Yeah idt it's a big deal either though I think it's good for style imo. That's why it's optional

Issue #1757, opened 2020-03-04
Move size_zero checks after consistency checks on arguments


mcol:
## Description
Almost all of our distributions we do a `size_zero` check on the arguments, and if true we immediately return. However, this should happen after we check that the arguments are consistent with our assumptions (https://github.com/stan-dev/math/pull/1716#discussion_r385764877).

## Example
In `normal_lpdf`:
```
  if (size_zero(y, mu, sigma)) {
    return 0.0;
  }

  check_not_nan(function, "Random variable", y);
  check_finite(function, "Location parameter", mu);
  check_positive(function, "Scale parameter", sigma);
  check_consistent_sizes(function, "Random variable", y, "Location parameter",
                         mu, "Scale parameter", sigma);
```
should become:
```
  check_not_nan(function, "Random variable", y);
  check_finite(function, "Location parameter", mu);
  check_positive(function, "Scale parameter", sigma);
  check_consistent_sizes(function, "Random variable", y, "Location parameter",
                         mu, "Scale parameter", sigma);

  if (size_zero(y, mu, sigma)) {
    return 0.0;
  }
```
#### Current Version:
v3.1.0

PR #1758, opened 2020-03-04
Move size_zero() calls after other consistency checks (A-I)

mcol:
## Summary

This is the first of two PRs that deals with moving the `size_zero` calls after the other consistency checks on function arguments. This covers all distributions with names starting A-I.

This also improves the consistency of the placement of the `using std::` statements and the declaration of the `operands_and_partials` variable. This came about as in many cases these lines of code were in between the `size_zero` checks and the other consistency checks, so they had to be moved anyway. Now all distribution files look very similar in their first few chunks, which is a small added benefit, I think.

## Tests

No new tests. The distribution test framework already contains tests for input vectors of size zero (search for `test_length_0_vector`).

## Side Effects

None.

## Checklist

- [X] Math issue #1757

### DIFF
--- a/stan/math/prim/prob/cauchy_lccdf.hpp
+++ b/stan/math/prim/prob/cauchy_lccdf.hpp
@@ -32,30 +32,26 @@ template <typename T_y, typename T_loc, typename T_scale>
 return_type_t<T_y, T_loc, T_scale> cauchy_lccdf(const T_y& y, const T_loc& mu,
                                                 const T_scale& sigma) {
   using T_partials_return = partials_return_t<T_y, T_loc, T_scale>;
-
-  if (size_zero(y, mu, sigma)) {
-    return 0.0;
-  }
-
   static const char* function = "cauchy_lccdf";
-
-  T_partials_return ccdf_log(0.0);
-
   check_not_nan(function, "Random variable", y);
   check_finite(function, "Location parameter", mu);
   check_positive_finite(function, "Scale parameter", sigma);
   check_consistent_sizes(function, "Random variable", y, "Location parameter",
                          mu, "Scale Parameter", sigma);
 
+  if (size_zero(y, mu, sigma)) {
+    return 0;
+  }
+
+  using std::atan;
+  using std::log;
+  T_partials_return ccdf_log(0.0);
+  operands_and_partials<T_y, T_loc, T_scale> ops_partials(y, mu, sigma);
+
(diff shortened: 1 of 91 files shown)


mcol:
It turns out that moving the `size_zero` checks exposed a bug in `check_bounded`, in that it didn't handle correctly size zero vectors (or rather, `scalar_seq_view` doesn't, but fixing `check_bounded` has fewer performance implications). With that simple fix, all tests pass, and this is ready for review.

stan/math/prim/prob/cauchy_lccdf.hpp
```diff
@@ -32,31 +32,27 @@ template <typename T_y, typename T_loc, typename T_scale>
...
   check_finite(function, "Location parameter", mu);
   check_positive_finite(function, "Scale parameter", sigma);
   check_consistent_sizes(function, "Random variable", y, "Location parameter",
                          mu, "Scale Parameter", sigma);
 
+  if (size_zero(y, mu, sigma)) {
+    return 0;
+  }
+
+  using std::atan;
```
bob-carpenter: [comment]
Just a comment because you're fixing so much of the style, but not worth going in and changing for this PR.  

I'd like to see using statements either all piled together at the top of the function or pushed down to just before the symbol is used.

stan/math/prim/prob/cauchy_log.hpp
```diff
@@ -22,9 +22,9 @@ namespace math {
  * @param mu (Sequence of) location(s).
  * @param sigma (Sequence of) scale(s).
  * @return The log of the product of densities.
+ * @tparam T_scale Type of scale.
```
bob-carpenter: [comment]
We want to head toward lower case and no periods for phrases that are not complete sentences.

mcol: Yes, that's where I'm heading to. I think what happened here is that I started fixing this file, then noticed it was one of the deprecated ones and forgot to undo completely my changes.

stan/math/prim/prob/cauchy_lccdf.hpp
mcol: Up to now they were placed inconsistently, at least my change put them in the same place.  I can move them to the top in the future.

bob-carpenter:
Everything's been very inconsistent in terms of formatting.  It's
especially great that you're fixing the behavior.

Issue #1763, opened 2020-03-05
propto tests not testing the correct things in probability unit tests


bbbales2:
## Description

I was investigating why pull https://github.com/stan-dev/math/pull/1331 passed tests but was broken.

It seemed to be a problem with how the ```propto = true``` logic was handled.

There are checks in the probability test framework that take advantage of the fact that if ```f(x)``` is a function proportional to the true distribution ```g(x)``` we have:

```
log(g(x)) = K + log(f(x))
```

So we can compute K at a couple different values of x and make sure they're the same. This would have caught the error in pull #1331.

If we look at the propto code (shortening here for simplicity: https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L205):

```
    T_return_type reference_logprob_true;
    T_return_type reference_logprob_false;

    {
      Scalar0 p0 = get_param<Scalar0>(parameters[0], 0);
      Scalar1 p1 = get_param<Scalar1>(parameters[0], 1);
      Scalar2 p2 = get_param<Scalar2>(parameters[0], 2);
      Scalar3 p3 = get_param<Scalar3>(parameters[0], 3);
      Scalar4 p4 = get_param<Scalar4>(parameters[0], 4);
      Scalar5 p5 = get_param<Scalar5>(parameters[0], 5);

      reference_logprob_true
          = TestClass.template log_prob<true, Scalar0, Scalar1, Scalar2,
                                        Scalar3, Scalar4, Scalar5>(p0, p1, p2,
                                                                   p3, p4, p5);
      reference_logprob_false
          = TestClass.template log_prob<false, Scalar0, Scalar1, Scalar2,
                                        Scalar3, Scalar4, Scalar5>(p0, p1, p2,
                                                                   p3, p4, p5);
    }

    for (size_t n = 0; n < parameters.size(); n++) {
      Scalar0 p0 = select_var_param<T0>(parameters, n, 0);
      Scalar1 p1 = select_var_param<T1>(parameters, n, 1);
      Scalar2 p2 = select_var_param<T2>(parameters, n, 2);
      Scalar3 p3 = select_var_param<T3>(parameters, n, 3);
      Scalar4 p4 = select_var_param<T4>(parameters, n, 4);
      Scalar5 p5 = select_var_param<T5>(parameters, n, 5);

      T_return_type logprob_true
          = TestClass.template log_prob<true, Scalar0, Scalar1, Scalar2,
[...]

PR #1764, opened 2020-03-05
Fixing propto tests in probability test framework

bbbales2:
This solves issue #1763

## Tests

This is a bugfix to a function in the test framework. I did not add any tests for this. Do we have tests for the probability test framework?

## Release Notes

Fixed propto tests in probability test framework (and fixed bugs this revealed in `neg_binomial_lpdf` and `pareto_type_2_lpdf`).

## Checklist

- [x] Math issue #1763


mcol:
This makes the prop tests for `neg_binomial_lpdf`, but I see that `normal_lpdf` fails as well, and perhaps other distributions too. Looking at the normal is much easier, as the distribution is much simpler and there are no numerical cutoffs. Right now I don't see anything wrong there, which may indicate that there's still something wrong in the distribution test, but before reaching that conclusion could you have a look at that too?

mcol:
This is not an exhaustive list, but also `cauchy_lpdf`, `chi_square_lpdf`, `exponential` and `gamma_lpdf` fail, while `bernoulli_lpmf`, `binomial_lpmf`, `poisson_lpmf` and `std_normal_lpdf` pass.

bbbales2:
Good catch. I think the check here: https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L254

Needs changed from:

```
EXPECT_TRUE(reference_logprob_false - logprob_false
            == reference_logprob_true - logprob_true)
```

to:

```
EXPECT_NEAR(value_of(reference_logprob_false - reference_logprob_true),
            value_of(logprob_false - logprob_true), 1e-12)
```

This just checks to make sure the two numbers are within 1e-12 of each other (absolute tolerance).

With this in place only negative binomial fails.

bob-carpenter:
You might need `value_of_rec` to get down to a`double` value from an
arbitrary autodiff variable.

mcol:
> With this in place only negative binomial fails.

Great! I went back as far as 2.18.1 and the `neg_binomial` bug was already there, so it's nothing new. I couldn't go further back as I had some compilation errors I didn't think it was worth trying to fix (probably the makefiles for 2.17 were not as clever as what we are used to now).

bbbales2:
I suspect the negative binomial is something @martinmodrak might know about. He got super deep on these functions a while back. I asked a question over here: https://github.com/stan-dev/math/pull/1497#issuecomment-595839718.

martinmodrak:
Yes, I believe the problem might be fixed with the new code, which, unfortunately, got a bit stuck, will try to move it forward a bit.

bbbales2:
@martinmodrak I removed the poisson stuff from neg_binomial in this branch and then updated the numerics a bit. I'm not sure I did it right though. Mind taking a look when you get the chance (no hurry)?

For `n = 13` `alpha = 1e11` and `beta = 1e10`, the version of negative_binomial in this branch is giving me:
```
-2.6185576442208003
```

R gives me:
```
-2.618557685656014211
```

with this code:

```
n = 13
alpha = 1e11
beta = 1e10
lp = dnbinom(n, mu = alpha / beta, size = alpha, log = TRUE)
sprintf("%.18f", lp)
```

Develop math was giving:
```
-2.618380836440565
```

So we improved but I'm not sure it's fixed yet.

bbbales2:
@martinmodrak with these values:

```
n = 13
alpha = 1.0e11
beta = 1.0e10
```

Can you run the code with lots of precision in Mathematica and tell me the result:

```
LogGamma[n + alpha] - LogGamma[alpha] - LogGamma[n + 1] +  alpha * Log[beta / (beta + 1)] - n * Log[beta + 1]
```

I'm trying to figure out how off the version of `neg_binomial` I implemented here is.

martinmodrak:
@bbbales2 Unfortunately, Mathematica (at least the free version available in Wolfram Cloud) refuses to compute this (allowed computation time exceeded).... To test for those big values of `alpha` and `beta`, the best I could do was to find analytical solutions for `n=0` and `n=1` (which then can be computed).

andrjohns:
The raspberry pi has Mathematica freely available, I tested with those values (may have done it wrong, not at all familiar with mathematica) to 20 decimals:
```
In[1]:= n = 13                                                                          

Out[1]= 13

In[2]:= alpha = 100000000000                                                            

Out[2]= 100000000000

In[3]:= beta  = 10000000000                                                             

Out[3]= 10000000000
                                                                       
In[4]:= N[LogGamma[n + alpha] - LogGamma[alpha] - LogGamma[n + 1] +  alpha * Log[beta /  (beta + 1)] - n * Log[beta + 1],{\[Infinity],20}]                                       

Out[4]= -2.6185576442208289933
```

bbbales2:
@martinmodrak @andrjohns thanks!

What we have in this branch is: -2.6185576442208003
What Mathematica gives: -2.6185576442208289933

I'll look into the n = 0 / n = 1 thing. I just wanted a spot check that things weren't totally awry. The big fix was dropping the Poisson like @martinmodrak did previously so I think we're probably better off already.

bbbales2:
I added an n = 0 and n = 1 comparison against long double calculations to the tests but I only compared with EXPECT_FLOAT_EQ so it's not really testing for much precision.

The numbers are (printed out to 17 digits):

```
n = 0:
implementation: -9.9999999995
long double:    -9.9999999995
```

```
n = 1
implementation: -7.6974149066059496
long double:    -7.6974149066159543
```

rok-cesnovar:
Are we targeting this for the release? Given that its a test it probably doesnt matter?

bbbales2:
@rok-cesnovar it's bugfixes, so I wasn't worried about the feature freeze. If it passes it's ready for review though.

rok-cesnovar:
Oh right. Yeah its a bug fix definitely.

bbbales2:
@andrjohns @SteveBronder this is ready to be reviewed.

There's three things going on here:

1. A fix to the original issue: https://github.com/stan-dev/math/issues/1763

2. Changed how neg_binomial worked to fix a problem this caught (avoids the poisson approximation and uses some new numerics copying from: https://github.com/stan-dev/math/pull/1497)

3. A fix a problem with templating in pareto_type_2 this caught

andrjohns:
I can review this one tonight

stan/math/prim/prob/neg_binomial_lpmf.hpp
```diff
@@ -65,69 +65,46 @@ return_type_t<T_shape, T_inv_scale> neg_binomial_lpmf(const T_n& n,
...
                 T_shape, T_inv_scale>
       lambda_m_alpha_over_1p_beta(size_alpha_beta);
-  for (size_t i = 0; i < size_alpha_beta; ++i) {
-    const T_partials_return alpha_dbl = value_of(alpha_vec[i]);
-    const T_partials_return beta_dbl = value_of(beta_vec[i]);
-    lambda[i] = alpha_dbl / beta_dbl;
-    alpha_log_beta_over_1p_beta[i] = alpha_dbl * log_beta_m_log1p_beta[i];
-    if (!is_constant_all<T_inv_scale>::value) {
-      lambda_m_alpha_over_1p_beta[i] = lambda[i] - alpha_dbl / (1 + beta_dbl);
+  if (!is_constant_all<T_inv_scale>::value) {
```
andrjohns, line 79: Given that this check and loop already exists below, I'd move these calculations there. Otherwise there'd be the cost associated with traversing large inputs twice

andrjohns: Similarly with the calculations of ```digamma_alpha``` above

stan/math/prim/prob/pareto_type_2_lpdf.hpp
```diff
@@ -67,9 +67,10 @@ return_type_t<T_y, T_loc, T_scale, T_shape> pareto_type_2_lpdf(
...
   }
 
-  VectorBuilder<!is_constant_all<T_shape>::value, T_partials_return, T_shape>
+  VectorBuilder<include_summand<propto, T_shape>::value, T_partials_return,
+                T_shape>
       inv_alpha(size(alpha));
-  if (!is_constant_all<T_shape>::value) {
+  if (include_summand<propto, T_shape>::value) {
     for (size_t n = 0; n < stan::math::size(alpha); n++) {
       inv_alpha[n] = 1 / value_of(alpha_vec[n]);
```
andrjohns, line 75: This is unrelated, but I'd probably change this to ```inv``` or ```1.0 / ``` to avoid accidentally invoking an integer division

test/prob/test_fixture_distr.hpp
```diff
@@ -251,8 +252,8 @@ class AgradDistributionTestFixture : public ::testing::Test {
                                         Scalar3, Scalar4, Scalar5>(p0, p1, p2,
                                                                    p3, p4, p5);
 
-      EXPECT_TRUE(reference_logprob_false - logprob_false
-                  == reference_logprob_true - logprob_true)
+      EXPECT_NEAR(value_of_rec(reference_logprob_false - logprob_false),
+                  value_of_rec(reference_logprob_true - logprob_true), 1e-12)
```
andrjohns, line 256: I might have missed this in the other issues, but why would we no longer expect the log-probabilities to be identical between the ```propto=true``` and ```propto=false``` conditions?

test/unit/math/mix/prob/neg_binomial_test.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <gtest/gtest.h>
+#include <boost/math/differentiation/finite_difference.hpp>
+
+TEST(mathMixScalFun, neg_binomial_lpmf_derivatives) {
+  auto f1 = [](const auto& alpha, const auto& beta) {
+    return stan::math::neg_binomial_lpmf(0, alpha, beta);
+  };
+  auto f2 = [](const auto& alpha, const auto& beta) {
+    return stan::math::neg_binomial_lpmf(6, alpha, beta);
+  };
```
andrjohns, line 12: Can combine these into a single lambda:
```
  auto f = [](const int x1) {
    return
        [=](const auto& x2, const auto& x3) { return stan::math::neg_binomial_lpmf(x1, x2, x3); };
  };
```
Then just call:
```
stan::test::expect_ad(f(0), 1.5, 4.1);
```

test/unit/math/prim/prob/neg_binomial_log_test.cpp
```diff
@@ -6,6 +6,17 @@ TEST(ProbNegBinomial, log_matches_lpmf) {
   double alpha = 1.1;
   double beta = 2.3;
 
+  long double alpha2 = 1e11;
+  long double beta2 = 1e10;
```
andrjohns, line 10: Does this need to be a ```long double``` if the results are only tested against ```float``` precision? (Genuinely asking, not sure if I'm missing something here)

test/prob/test_fixture_distr.hpp
bbbales2: It's a floating point comparison thing. Different code runs with `propto=true`/`propto=false`, so things can round differently.

Here was the error from Jenkins:
```
./test/prob/test_fixture_distr.hpp:256: Failure
Value of: value_of_rec(reference_logprob_false - logprob_false) == value_of_rec(reference_logprob_true - logprob_true)
  Actual: false
Expected: true
Proportional test failed at index: 1
  reference params: (1.2, 0.29999999999999999, 2)
  current params:   (-1, 0.20000000000000001, 0.25)
  ref<true> = -1.4364978919949096
  cur<true> = -1.4901106714054309
  ref<false> = -2.129645072554855
  cur<false> = -2.1832578519653758
```

And then if you compute the difference in R:
```
> -1.4364978919949096 + 1.4901106714054309 -
  (-2.129645072554855 + 2.1832578519653758)
[1] 4.440892e-16
```

I just picked 1e-12 cause I think it'll catch real errors and avoid anyone worrying about finickiness like this.

test/unit/math/prim/prob/neg_binomial_log_test.cpp
bbbales2: The logic is that the long double reference calculations there will be super high precision, and our neg_binomial_lpmf better match it pretty well. Our neg_binomial_lpmf is doing something more complicated though (it has gamma functions and such), so it's good that it's able to keep precision.

We could turn the precision up here: https://github.com/stan-dev/math/pull/1764#issuecomment-661158937

martinmodrak: A very minor point: I tried using `long double` to avoid numerical issues a few times with `neg_binomial_2_lpmf` and was always disappointed - all of the issues I had were almost untouched by moving to `long double` (my _guess_ is that this was because of some exponential-like behaviour where `long double` helds only a tiny bit longer than `double`, but I don't know for sure). It certainly can help a bit, but I wouldn't put a lot of confidence in `long double` as a "safe" method.

bbbales2: Mmm, we could axe the tests then since they're not doing much other than telling us the new binomial_coefficient_log is working well (that's already tested separately).

martinmodrak: I think it is good the tests are there, I just don't think you get a lot of benefit with `long double` over `double`, so wanted to share my experience so that you are not overconfident :-)  (but having  `long double` can't hurt)

stan/math/prim/prob/neg_binomial_lpmf.hpp
bbbales2: It's possible that alpha and beta are both size 1 but y is length N > 1, so this is faster possibly.

stan/math/prim/prob/neg_binomial_lpmf.hpp
```diff
@@ -65,69 +65,46 @@ return_type_t<T_shape, T_inv_scale> neg_binomial_lpmf(const T_n& n,
...
+    for (size_t i = 0; i < size_alpha_beta; ++i) {
+      const T_partials_return alpha_dbl = value_of(alpha_vec[i]);
+      const T_partials_return beta_dbl = value_of(beta_vec[i]);
```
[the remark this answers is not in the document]
bbbales2: Have a look at what I did. I don't see this pattern elsewhere in the code, but I think I did the right thing.

Is this what you meant? If so I can bench is real quick since it doesn't look familiar but I agree it makes sense (we avoid a bunch of vari dereferences which are probably expensive)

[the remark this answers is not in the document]
bbbales2: I did a benchmark for a length N input (where alpha and beta are both length N vectors). There wasn't much of a difference. It was worth a try but I'm going to revert the change just so this code looks like all the other lpmf stuff.

With the new code:

```
----------------------------------------------------------------------
Benchmark                               Time           CPU Iterations
----------------------------------------------------------------------
neg_binomial/2/manual_time            509 ns        544 ns    1368991
neg_binomial/4/manual_time            822 ns        855 ns     841446
neg_binomial/8/manual_time           1483 ns       1517 ns     473999
neg_binomial/16/manual_time          2797 ns       2830 ns     247871
neg_binomial/32/manual_time          5425 ns       5458 ns     129481
neg_binomial/64/manual_time         10607 ns      10637 ns      66323
neg_binomial/128/manual_time        21443 ns      21485 ns      33076
neg_binomial/256/manual_time        42787 ns      42817 ns      16309
neg_binomial/512/manual_time        85370 ns      85407 ns       8247
neg_binomial/1024/manual_time      170181 ns     170208 ns       4116
```

And then the old code:
```
neg_binomial2/2/manual_time           487 ns        522 ns    1435154
neg_binomial2/4/manual_time           806 ns        840 ns     859681
neg_binomial2/8/manual_time          1505 ns       1540 ns     469482
neg_binomial2/16/manual_time         2836 ns       2870 ns     247975
neg_binomial2/32/manual_time         5403 ns       5436 ns     128561
neg_binomial2/64/manual_time        10801 ns      10834 ns      65618
neg_binomial2/128/manual_time       21387 ns      21429 ns      32484
neg_binomial2/256/manual_time       42970 ns      43009 ns      16359
neg_binomial2/512/manual_time       85111 ns      85145 ns       8208
neg_binomial2/1024/manual_time     169757 ns     169782 ns       4128
```

This is the vague benchmark code:
```
static void neg_binomial(benchmark::State& state) {
  using stan::math::var;
  std::vector<int> y(state.range(0), 7);
  Eigen::VectorXd alpha_vals = 0.5 * Eigen::VectorXd::Ones(state.range(0)) + 0.1 * Eigen::VectorXd::Random(state.range(0));
  Eigen::VectorXd beta_vals = 0.5 * Eigen::VectorXd::Ones(state.range(0)) + 0.1 * Eigen::VectorXd::Random(state.range(0));
  for (auto _ : state) {
    auto start = std::chrono::high_resolution_clock::now();
    Eigen::Matrix<var, -1, 1> alpha = alpha_vals;
    Eigen::Matrix<var, -1, 1> beta = beta_vals;
    var lp = neg_binomial_lpmf(y, alpha, beta);
    lp.grad();
    auto end = std::chrono::high_resolution_clock::now();
    auto elapsed_seconds =
      std::chrono::duration_cast<std::chrono::duration<double>>(end - start);
    state.SetIterationTime(elapsed_seconds.count());
    stan::math::recover_memory();
    benchmark::ClobberMemory();
  }
}
```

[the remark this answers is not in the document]
andrjohns: Neat, no differences, safe to ignore

stan/math/prim/prob/neg_binomial_lpmf.hpp
andrjohns: Ah, I missed that part, cool beans

PR #1765, opened 2020-03-06
Cleanup duplicated overloads of scalar std:: functions

### LINKED ISSUE #1692: Investigate which int overloads are still needed (exp, pow, ...)

rok-cesnovar:
## Summary

Closes #1692 by cleaning up unnecessary int overloads. 
This PR also fixes a few comments and removes functions where the only code is a call to a std function. Example: 

```
inline double asinh(double x) { return std::asinh(x); }

inline double asinh(int x) { return std::asinh(x); }
```
These 2 can be removed because they are covered by 

```
struct asinh_fun {
  template <typename T>
  static inline T fun(const T& x) {
    using std::asinh;
    return asinh(x);
  }
};

template <typename T>
inline auto asinh(const T& x) {
  return apply_scalar_unary<asinh_fun, T>::apply(x);
}
```
We could do this for a lot more functions, for now this just removes these simple ones.

## Tests

No new tests.

## Side Effects

/

## Checklist

- [x] Math issue #1692 

### DIFF
--- a/stan/math/prim/fun/acos.hpp
+++ b/stan/math/prim/fun/acos.hpp
@@ -9,7 +9,7 @@ namespace stan {
 namespace math {
 
 /**
- * Structure to wrap acos() so it can be vectorized.
+ * Structure to wrap <code>acos()</code> so it can be vectorized.
  *
  * @tparam T type of variable
  * @param x variable
@@ -24,7 +24,8 @@ struct acos_fun {
 };
 
 /**
- * Vectorized version of acos().
+ * Returns the elementwise <code>acos()</code> of the input,
+ * which may be a scalar or any Stan container of numeric scalars.
  *
  * @tparam T type of container
  * @param x container
--- a/stan/math/prim/fun/acos.hpp
+++ b/stan/math/prim/fun/acos.hpp
@@ -9,7 +9,7 @@ namespace stan {
 namespace math {
 
 /**
- * Structure to wrap <code>acos()</code> so it can be vectorized.
+ * Structure to wrap \c acos() so it can be vectorized.
  *
  * @tparam T type of variable
  * @param x variable
@@ -24,7 +24,7 @@ struct acos_fun {
 };
 
 /**
- * Returns the elementwise <code>acos()</code> of the input,
+ * Returns the elementwise \c acos() of the input,
  * which may be a scalar or any Stan container of numeric scalars.
  *
  * @tparam T type of container
--- a/stan/math/prim/fun/acos.hpp
+++ b/stan/math/prim/fun/acos.hpp
@@ -9,7 +9,7 @@ namespace stan {
 namespace math {
 
 /**
- * Structure to wrap \c acos() so it can be vectorized.
+ * Structure to wrap `acos()` so it can be vectorized.
  *
  * @tparam T type of variable
  * @param x variable
@@ -24,7 +24,7 @@ struct acos_fun {
 };
 
 /**
- * Returns the elementwise \c acos() of the input,
+ * Returns the elementwise `acos()` of the input,
  * which may be a scalar or any Stan container of numeric scalars.
  *
  * @tparam T type of container
@@ -37,7 +37,7 @@ inline auto acos(const T& x) {
 }
 
 /**
- * Version of acos() that accepts Eigen Matrix or matrix expressions.
+ * Version of `acos()` that accepts Eigen Matrix or matrix expressions.
  * @tparam Derived derived type of x
  * @param x Matrix or matrix expression
  * @return Arc cosine of each variable in the container, in radians.
--- a/stan/math/prim/fun/atanh.hpp
+++ b/stan/math/prim/fun/atanh.hpp
@@ -36,12 +36,8 @@ inline double atanh(double x) {
  * @throw std::domain_error If argument is less than 1.
  */
 inline double atanh(int x) {
-  if (is_nan(x)) {
-    return x;
-  } else {
-    check_bounded("atanh", "x", x, -1, 1);
-    return std::atanh(x);
-  }
+  check_bounded("atanh", "x", x, -1, 1);
+  return std::atanh(x);
 }
 
 /**
--- a/stan/math/prim/fun/log2.hpp
+++ b/stan/math/prim/fun/log2.hpp
@@ -3,27 +3,10 @@
 
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/fun/constants.hpp>
-#include <stan/math/prim/fun/log.hpp>
 #include <cmath>
 
 namespace stan {
 namespace math {
-
-/**
- * Returns the base two logarithm of the argument (C99, C++11).
- *
- * The function is defined by:
- *
- * <code>log2(a) = log(a) / std::log(2.0)</code>.
- *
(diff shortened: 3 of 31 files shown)


mcol:
Just a comment on the doxygen: I think we can use markdown backticks around variable/function names, rather than `<code>/</code>`.

rok-cesnovar:
Just double checking. You mean like this:

`Structure to wrap \c acos() so it can be vectorized.`

mcol:
I meant the using "`" as in markdown. I've used those for the `discrete_range` functions, and they render as expected.

rok-cesnovar:
Thanks, good to know.

bob-carpenter:
The reasons all those `double` and `int` overloads were introduced was to disambiguate uses of `asinh(3)`, which would otherwise be ambiguous between being promoted to `double` or `var`.  

As of C++11, there are now integral type overloads for all the primitive functions like `asinh` so this may not be necessary.

What needs to be checked is compilability of a Stan program with `asinh(int)`, etc.  That should happen in the Stan model checks, but I don't know if those are part of the upstream.  But you can test just in a simple file after including all of math:

```
auto x = asinh(2);
```

rok-cesnovar:
Yeah, I am not sure why I thought the 2 in the below test would already check that.
```cpp
TEST(mathMixMatFun, asinh) {
  auto f = [](const auto& x1) { return stan::math::asinh(x1); };
  stan::test::expect_common_unary_vectorized(f);
  stan::test::expect_unary_vectorized(f, -2.6, -1.2, -0.2, 0.5, 2, -1.2);
}
```
This does compile fine and the `test_ad.hpp` include brings in fwd,mix,rev and prim.
```cpp
TEST(mathMixMatFun, asinh_compile) {
  auto x = asinh(2);
}
```

But I think it will be simpler and cleaner to add a model to stan-dev/stan. Will do that.

bob-carpenter:
Cool.  That's great.

That test function's written with qualified calls.  It'd probably be better to do something like:

```
using std::asinh;
return asinh(x1);
```

and let `stan::math::` be brought in only by ADL.  I can't recall why I didn't write them that way to begin with because that's the way we generate C++ code for the language.

stan/math/prim/fun/log2.hpp
```diff
@@ -45,12 +28,13 @@ struct log2_fun {
    */
   template <typename T>
   static inline T fun(const T& x) {
```
SteveBronder: Unrelated to this PR but why are these defined as static?

SteveBronder:
> But I think it will be simpler and cleaner to add a model to stan-dev/stan. Will do that.

Is this PR waiting on the above?

stan/math/prim/fun/log2.hpp
bob-carpenter: They're static so that there's no object creation/destruction at run time.   Any time you can move something to a static function it's a win.  You lose generality in that we can't accept an arbitrary functor, but we never need to in this application.

rok-cesnovar:
Yep, I am working on that.

rok-cesnovar:
Here is the related [Stan PR](https://github.com/stan-dev/stan/pull/2899) with the test model. Its now running a test run with the Math branch set to this PR: https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/PR-2899/2/pipeline

SteveBronder:
Nice! I looked this over so let me check that out / when the tests pass I'll approve

rok-cesnovar:
Great. Both windows machines are offline, so I guess it wont be soon :) Will see if there is someone in the lab to reboot our machine. Unlikely tho, as the whole uni is shut down.

rok-cesnovar:
Upstream PR is merged, this is ready for review.

stan/math/prim/fun/acos.hpp
```diff
@@ -9,7 +9,7 @@ namespace stan {
 namespace math {
 
 /**
- * Structure to wrap acos() so it can be vectorized.
+ * Structure to wrap `acos()` so it can be vectorized.
```
SteveBronder: Thanks! Didn't know we could use markdown until Bob showed me. Much easier!

stan/math/prim/fun/atanh.hpp
```diff
@@ -36,12 +36,8 @@ inline double atanh(double x) {
...
  */
 inline double atanh(int x) {
-  if (is_nan(x)) {
-    return x;
-  } else {
-    check_bounded("atanh", "x", x, -1, 1);
-    return std::atanh(x);
-  }
+  check_bounded("atanh", "x", x, -1, 1);
+  return std::atanh(x);
```
SteveBronder: Just a sanity check, this was changed because the standard says below right?

> If a domain error occurs, an implementation-defined value is returned (NaN where supported)

SteveBronder: Figure nan is a domain error

stan/math/prim/fun/log2.hpp
SteveBronder: Cool cool I don't see any difference between making these static vs. inline

https://godbolt.org/z/fIlp6-

test/unit/math/prim/fun/exp_test.cpp
```diff
@@ -3,8 +3,8 @@
 
 TEST(MathFunctions, expInt) {
   using stan::math::exp;
-  using std::exp;
```
SteveBronder: Can you double check the other function tests also have these removed?

stan/math/prim/fun/atanh.hpp
rok-cesnovar: This was changed because I feel that its an unnecessary check. An int cant be NaN as NaN is defined only for floating point numbers.

`std::isnan(x)` that is used underneath will just cast the int to double and do a check.

rok-cesnovar: Not a huge deal probably as I doubt many call atanh with ints.

test/unit/math/prim/fun/exp_test.cpp
rok-cesnovar: Sure.

stan/math/prim/fun/acos.hpp
rok-cesnovar: Yeah, me neither. @mcol let me in on the secret :)

test/unit/math/prim/fun/exp_test.cpp
rok-cesnovar: I did a grep of `using std::` in test/unit/math/prim/fun and apart from log, sqrt, exp the only uses are with std::vector and a few tests that explicitly want std::exp, std::log or std::fabs. I think we are good.

PR #1766, opened 2020-03-09
Allow constrain and free functions to work on Eigen types and std vectors

SteveBronder:
## Summary

I was working on a PR for sparse matrices up in the Stan library and found a few places where we can use vectorized ops in the reader and writer. This PR adds vectorized versions of the `*_constrain` and `*_free` functions so they can take in eigen types and standard vectors. This also adds testing for the `*_free` functions which idt existed before.

## Tests

Tests can be run with

```
./runTests.py ./test/unit/math/mix/fun/ -f constrain_test
```
## Side Effects

The `*_constrain` and `*_free` can now take in vector like types.

## Checklist


SteveBronder:
This is ready for review! If the reviewer wants me to break this up into multiple PRs that's fine, though the changes are mostly the same across a lot of the files

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -16,28 +18,80 @@ namespace math {
...
+ *
+ * @tparam EigT type derived from EigenBase
+ * @tparam U type of lower bound
+ * @param[in] x Unconstrained scalar input
+ * @param[in] lb lower bound on constrained output
+ * @return lower bound constrained value corresponding to inputs
+ */
+template <typename EigT, typename U, require_eigen_t<EigT>* = nullptr>
+inline auto lb_constrain(EigT&& x, U&& lb) {
+  return x.unaryExpr([&lb](auto&& x_iter) { return lb_constrain(x_iter, lb); })
```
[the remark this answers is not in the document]
t4c1: Also use `apply_vector_unary` to combine with next overload.

stan/math/prim/fun/ordered_constrain.hpp
```diff
@@ -14,26 +14,22 @@ namespace math {
...
-  using Eigen::Dynamic;
-  using Eigen::Matrix;
+template <typename Vec, require_vector_like_t<Vec>* = nullptr>
+inline auto ordered_constrain(Vec&& x) {
   using std::exp;
-  using size_type = index_type_t<Matrix<T, Dynamic, 1>>;
 
-  size_type k = x.size();
-  Matrix<T, Dynamic, 1> y(k);
+  auto k = x.size();
```
t4c1: No need for auto here.

stan/math/prim/fun/positive_ordered_constrain.hpp
```diff
@@ -18,21 +18,16 @@ namespace math {
...
+  plain_type_t<Vec> y(k);
   if (k == 0) {
     return y;
   }
   y[0] = exp(x[0]);
-  for (size_type i = 1; i < k; ++i) {
+  for (auto i = 1; i < k; ++i) {
     y[i] = y[i - 1] + exp(x[i]);
   }
   return y;
```
t4c1: `return cumulative_sum(exp(x))`?

stan/math/prim/fun/simplex_constrain.hpp
```diff
@@ -20,28 +20,24 @@ namespace math {
  *
  * The transform is based on a centered stick-breaking process.
  *
- * @tparam T type of elements in the vector
+ * @tparam Vec type deriving from `Eigen::MatgrixBase` with rows or columns
```
t4c1:
```suggestion
 * @tparam Vec type deriving from `Eigen::MatrixBase` with rows or columns
```

stan/math/prim/fun/vec_concat.hpp
```diff
@@ -13,24 +13,25 @@ namespace math {
...
  */
-template <typename T, typename... Args>
-inline const std::vector<T> vec_concat(const std::vector<T>& v1,
-                                       const Args... args) {
-  std::vector<T> vec = vec_concat(args...);
+template <typename Vec, typename... VecArgs,
+          require_all_std_vector_t<Vec, VecArgs...>* = nullptr>
+inline auto vec_concat(Vec&& v1, VecArgs&&... args) {
+  std::vector<value_type_t<Vec>> vec
+      = vec_concat(std::forward<VecArgs>(args)...);
```
t4c1: [optional] This is a bit inefficient. The result vector `vec` should reserve its final size in advance.

test/unit/math/mix/fun/identity_constrain_test.cpp
```diff
@@ -2,22 +2,27 @@
...
+  return x_free;
 }
 template <typename T>
-T g3(const T& x) {
-  T lp = 0;
-  stan::math::identity_constrain(x, lp);
+auto g3(const T& x) {
+  stan::value_type_t<T> lp = 0;
+  auto x_cons = stan::math::identity_constrain(x, lp);
+  auto x_free = stan::math::identity_free(x_cons);
```
t4c1: You are only testing `identity_constrain` and `identity_free` together. Each should also be tested individually.

stan/math/prim/fun/dot_self.hpp
```diff
@@ -21,14 +29,14 @@ inline double dot_self(const std::vector<double>& x) {
...
+ *  rows or columns equal to 1.
  * @param v Vector.
  * @throw std::domain_error If v is not vector dimensioned.
  */
-template <int R, int C>
-inline double dot_self(const Eigen::Matrix<double, R, C>& v) {
-  check_vector("dot_self", "v", v);
+template <typename EigMat,
+          require_eigen_vt<std::is_arithmetic, EigMat>* = nullptr>
+inline double dot_self(EigMat&& v) {
```
bob-carpenter: I'm not reviewing, but would very much appreciate seeing this renamed to  `EigVec` (what I'd really like is `V`, but that's a bigger discussion we need to take up).  

It is *not* an Eigen matrix, it's an Eigen *vector*.  `Eigen::Matrix<T, 1, -1>` and `Eigen::Matrix<T, -1, 1>` and `Eigen::Matrix<T, -1, -1>` are three completely independent specializations of `Eigen::Matrix<T, R, C>`.  Only the last of the tree is a matrix.  The other two are vectors and do not even support `T& operator()(ptrdiff_t, ptrdiff_t);`.

stan/math/prim/fun/identity_free.hpp
```diff
@@ -13,13 +13,32 @@ namespace math {
  * <p>This function is a no-op and mainly useful as a placeholder
  * in auto-generated code.
  *
- * @tparam T type of value
- * @param[in] y value
+ * @tparam Scalar type of value
```
bob-carpenter: This is not a scalar any more, because now scalars include complex numbers.  Going back to `T` works.  Or if you insist on verbosely naming types, it should be `Real`.

stan/math/prim/fun/dot_self.hpp
t4c1: I find it weird, but there are even some test checking that `dot_self` works with (non-vector) matrices.

stan/math/prim/fun/identity_free.hpp
t4c1: Why not? Complex scalars are scalars as well.

bob-carpenter: The problem's the other way around.  These functions won't work for complex inputs, so the required type should be called `Real`, not `Scalar`.

stan/math/prim/err/check_ordered.hpp
```diff
@@ -13,48 +13,17 @@ namespace math {
...
- * @throw <code>std::domain_error</code> if the vector elements are
- *   not ordered, if there are duplicated values, or if any element
- *   is <code>NaN</code>.
- */
-template <typename T_y>
-void check_ordered(const char* function, const char* name,
-                   const std::vector<T_y>& y) {
-  for (size_t n = 1; n < y.size(); n++) {
+template <typename Vec, require_vector_like_t<Vec>* = nullptr>
+inline void check_ordered(const char* function, const char* name, Vec&& y) {
```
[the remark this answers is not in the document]
SteveBronder: What's the reason behind prefering const&? I kind of like universal references since they preserve value types

stan/math/prim/fun/divide.hpp
```diff
@@ -47,6 +49,24 @@ inline auto divide(const Mat& m, Scal c) {
...
+ * @param[in] m specified matrix or expression
+ * @param[in] c specified scalar
+ * @return matrix divided by the scalar
+ */
+template <typename Vec, typename Scal, require_std_vector_t<Vec>* = nullptr,
+          require_stan_scalar_t<Scal>* = nullptr>
+inline auto divide(const Vec& x, Scal c) {
+  std::vector<value_type_t<Vec>> ret_x(x.size());
+  std::transform(x.begin(), x.end(), ret_x.begin(),
+                 [&c](auto&& x_iter) { return x_iter / c; });
```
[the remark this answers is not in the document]
SteveBronder: Which overload are you referencing here?

stan/math/prim/fun/dot_self.hpp
```diff
@@ -10,9 +10,17 @@
...
-  for (double i : x) {
+/**
+ * Returns the dot product of the specified vector with itself.
+ *
+ * @tparam StdVec a standard vector.
+ * @param v Vector.
+ * @throw std::domain_error If v is not vector dimensioned.
+ */
+template <typename StdVec, require_std_vector_t<StdVec>* = nullptr>
+inline auto dot_self(StdVec&& x) {
```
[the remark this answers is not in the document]
SteveBronder: I saw you changed this in your PR so I'll wait on these till your PR is merged

stan/math/prim/fun/identity_constrain.hpp
```diff
@@ -13,32 +13,33 @@ namespace math {
...
- * @tparam T Type of scalar.
+ * @tparam Scalar Type of scalar.
  * @param[in] x free scalar
  * @return transformed input
  */
-template <typename T>
-inline T identity_constrain(const T& x) {
-  return x;
+template <typename Scalar, require_all_stan_scalar_t<Scalar>* = nullptr>
+inline decltype(auto) identity_constrain(Scalar&& x) {
```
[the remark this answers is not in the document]
SteveBronder: I'm going to make/run benchmarks on passing by value and reference for stan autodiff types tmrw. My gut here is that since passing var by value requires adding something to the autodiff stack (I think that also generates another `chain()` call) passing by reference may be better

stan/math/prim/fun/identity_constrain.hpp
```diff
@@ -13,32 +13,33 @@ namespace math {
...
  * @param[in] x scalar
- * @param[in] lp log density reference
+ * @param[in] args values to check for promotion rules.
  * @return transformed input
  */
-template <typename T, typename S>
-inline T identity_constrain(const T& x, S& lp) {
-  return x;
+template <typename Scalar, typename... Types,
+          require_all_stan_scalar_t<Scalar, Types...>* = nullptr>
```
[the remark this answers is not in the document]
SteveBronder: When I was first fixing these I thought we would need to do `identity_constrain(x, lb, lp)` or `identity_constrain(x, lb, ub, lp)` for some functions to get the correct return type and enable RVO. But `lp` is actually not needed in those so I can roll this back to just take in two types. imo idt the variadic hurts here

stan/math/prim/fun/lb_constrain.hpp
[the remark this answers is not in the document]
SteveBronder: Could you or @andrjohns show me an example of how to do apply_vector_unary here? Or is there some docs somewhere? I looked at log_sum_exp and log_softmax but couldn't get it to compile

stan/math/prim/fun/positive_ordered_constrain.hpp
SteveBronder: I need to look into this but that was giving me the test failure

```
test/unit/math/prim/fun/positive_ordered_transform_test.cpp:25: Failure
  exp(1.0) + exp(-2.0) + exp(-5.0)
    Which is: 2.86036
  y[2]
    Which is: 2.86036
```

stan/math/prim/fun/ordered_constrain.hpp
SteveBronder: I thought so too but it looks like users can change this and it's default is

> EIGEN_DEFAULT_DENSE_INDEX_TYPE - the type for column and row indices in matrices, vectors and array (DenseBase::Index). Set to std::ptrdiff_t by default.

So I know it sounds weird but I kind of like auto here. Though casting to int is no biggie (idt) so I'm fine with just calling it int

https://eigen.tuxfamily.org/dox/TopicPreprocessorDirectives.html

stan/math/prim/err/check_ordered.hpp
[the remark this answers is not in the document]
t4c1: It is simply the matter of using the simplest tool that does the job. It makes code easier to understand (especially to someone new to the codebase). By specifying const& anyone reading the code immediately knows that the function does not modify the argument.

What do you mean by preserving value types? In what cases is it beneficial?

stan/math/prim/fun/divide.hpp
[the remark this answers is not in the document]
t4c1:
```
template <typename Mat, typename Scal, typename = require_eigen_t<Mat>,
          typename = require_stan_scalar_t<Scal>,
          typename = require_all_not_var_t<scalar_type_t<Mat>, Scal>>
inline auto divide(const Mat& m, Scal c) {
  return (m / c).eval();
}
```

stan/math/prim/fun/identity_constrain.hpp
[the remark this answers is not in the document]
t4c1: Well, every function could have a variadic parameter pack it ignors after all arguments it uses and work fine. But I would not add it unless we needed it for some reason.

stan/math/prim/fun/lb_constrain.hpp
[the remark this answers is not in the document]
t4c1: I think it should be something like:

```
template <typename T, typename U, require_vector_like_t<T>* = nullptr, require_stan_scalar_t<U>* = nullptr>
inline plain_type_t<T> lb_constrain(const T& x, const U& lb) {
  if (lb == NEGATIVE_INFTY) {
    return identity_constrain(x);
  }
  return apply_vector_unary<T>::apply(x, [&lb](const auto& v) {
    return exp(v) + lb; 
  });
}
```

stan/math/prim/fun/ordered_constrain.hpp
t4c1: In that case you can use `Eigen::Index` to avoid the cast.

stan/math/prim/fun/positive_ordered_constrain.hpp
t4c1: Well the test uses `EXPECT_EQ` for floating point numbers. It would fail for anything that is not exactely the same expression. The test should be modified to have some tolerance.

stan/math/prim/err/check_ordered.hpp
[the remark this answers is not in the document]
syclik: I'll second @t4c1's comment. This is purely for readability. Yes, you can use universal references, but we don't have that propagated throughout the Math library. This PR doesn't need this particular change and if you wanted to change that all at once, we can do that (in a separate issue + PR).

stan/math/prim/err/check_positive_ordered.hpp
```diff
@@ -14,19 +14,17 @@ namespace math {
...
  */
-template <typename T_y>
-void check_positive_ordered(const char* function, const char* name,
-                            const Eigen::Matrix<T_y, Eigen::Dynamic, 1>& y) {
-  using Eigen::Dynamic;
-  using Eigen::Matrix;
-
+template <typename Vec, require_vector_like_t<Vec>* = nullptr>
+inline void check_positive_ordered(const char* function, const char* name,
+                                   Vec&& y) {
```
syclik, line 27: Same comment about using the "universal reference."

stan/math/prim/fun/divide.hpp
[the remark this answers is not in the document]
syclik: @SteveBronder, did you see @t4c1's last comment here?

stan/math/prim/fun/identity_constrain.hpp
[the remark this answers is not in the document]
syclik: Any results from benchmarking?

SteveBronder:
I got this to work on an upstream stan branch and didn't see any performance improvements over the basic benchmarks and a few in examples so going to close

PR #1616, opened 2020-01-15
[WIP] Parallel Prototype

SteveBronder:
This is just a PR for discussion of the tbb parallelism prototype

From Ben's last comment on discourse chat below

https://discourse.mc-stan.org/t/parallel-autodiff-v3/12474/40

Alright, stuff is up (https://github.com/stan-dev/math/compare/proto-parallel-v3...cleanup/proto-parallel-v3).

I screwed up autodiff on the sliced arguments. I was scared about what would happen if all the vars in the sliced arguments had the same vari, so I just didn’t implement that.

I broke the things into two functions – parallel_sum_var and parallel_sum. I think we need enable_if logic or something to distinguish between these things properly.

I got rid of the init argument.

I got rid of local_operands_and_partials.

Arguments are stored as a tuple of const references: https://github.com/stan-dev/math/blob/cbcf42fad798015b317ba6dab7a6ddc5d9983aa2/stan/math/rev/functor/reduce_sum.hpp#L24

Adjoints are accumulated in an eigen vector: https://github.com/stan-dev/math/blob/cbcf42fad798015b317ba6dab7a6ddc5d9983aa2/stan/math/rev/functor/reduce_sum.hpp#L27

The member function ‘deep_copy’ of the recursive_reducer that handles the deep copies: https://github.com/stan-dev/math/blob/cbcf42fad798015b317ba6dab7a6ddc5d9983aa2/stan/math/rev/functor/reduce_sum.hpp#L114
The member function ‘accumulate_adjoints’ collects the the adjoints from the nested autodiff and arranges them in the Eigen::VectorXd: https://github.com/stan-dev/math/blob/cbcf42fad798015b317ba6dab7a6ddc5d9983aa2/stan/math/rev/functor/reduce_sum.hpp#L131

The reduce function accumulates the sum and the adjoint: https://github.com/stan-dev/math/blob/cbcf42fad798015b317ba6dab7a6ddc5d9983aa2/stan/math/rev/functor/reduce_sum.hpp#L140
It seems like operator() of a single recursive_reducer might get called multiple times though? I was surprised by this.

The member function count_memory of parallel_sum_rev_impl counts the number of vars in all the input arguments so it can allocate memory to store the vari pointers: https://github.com/stan-dev/math/blob/cbcf42fad798015b317ba6dab7a6ddc5d9983aa2/stan/math/rev/functor/reduce_sum.hpp#L217

[...]

PR #1768, opened 2020-03-09
[WIP][WIP] Add docs and generic value type signatures for reduce_sum

SteveBronder:
## Summary

This is a PR that will go into #1616. It adds docs and refactors the signatures as [universal references](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers) to preserve value types.

## Tests

With `STAN_THREADS=true` in `make/local` run

```cpp
STAN_NUM_THREADS=4 ./runTests.py -j12 test/unit/math/ -f reduce_sum
```

## Side Effects

Where `reduce_sum_impl` previously stored `vmapped_` and `args_tuple` as `const Vec&` and `std::tuple<const Args&...>` universal references now allow these to be `Arg&`, `const Arg&`, `Arg&&`, etc.

## Checklist

- [x] the basic tests are passing (for reduce_sum yes)


bbbales2:
Is this merging in develop or something?

SteveBronder:
Yeah I think so which is why the num changed is so high. Lemme look at this

SteveBronder:
There we go

bbbales2:
Nice thanks. Oh man this has doc too! Thanks

What does the * mean and what does the nullptr mean?
```
require_std_vector_vt<is_container, VarVec>* = nullptr
```

bbbales2:
Actually just read me the whole thing in words if you don't mind.

SteveBronder:
lol fair

```
template <typename VarVec, require_std_vector_vt<is_container, VarVec>* = nullptr>
```

- Define a template with a type parameter `VarVec` and a `void*` parameter with a default value set to `nullptr`.When `VarVec` (1) is an `std::vector` and (2) has a `value_type` this is a container (either an eigen type or standard vector) this template is well defined and a candidate when choosing specializations after argument dependent lookup. If `VarVec` does not satisfy (1) and (2) this template fails during argument dependent lookup.


`require_std_vector_vt<is_container, VarVec>` is just shorthand for 
```cpp
std::enable_if_t<
  conjunction<is_container<value_type_t<T>>,
   is_std_vector<T>>::value>>`
```

Does that make sense?

SteveBronder:
Similarly, `require_std_vector_st` there would be

```cpp
std::enable_if_t<
  conjunction<is_container<value_type_t<T>>,
   is_std_vector<T>>::value>>`
```

I have some docs for these on the website

http://mc-stan.org/math/d1/db9/group__require__meta.html

SteveBronder:
Also when the require is on the rhs of something like `typename = require_vector_like_t<Vec>` that creates an unnamed type parameter with a value of void if the require is successful and fails compilation otherwise (aka does SFINAE)

bbbales2:
Got it. That reads pretty well. I'll have a look-see at this tomorrow.

SteveBronder:
Also I just realized this, but is there some way we could write functions like `normal_lpdf_par()` that takes in some chunk size for the first input and then would just do reduce_sum in the function? Cuz then we could pretty much parallelize all of the probability functions

SteveBronder:
Something like

```cpp
template <typename T_y, typename T_loc, typename T_scale>
inline auto normal_lpdf_par(size_t chunk_size, T_y&& y, T_loc&& mu, T_scale&& sigma) {
  // Make functor
  auto normal_lpdf_func = [&](std::size_t start, std::size_t end,
     auto&& sub_slice, std::ostream* msgs, auto&&... args) {
    return normal_lpdf<false>(slice(sub_slice, start, end), args...)
  };
  // Get this from somewhere
  std::ostream* msgs = nullptr;
  return reduce_sum<decltype(normal_lpdf_func)>(chunk(y, chunk_size), 8, msgs, mu, sigma)
}
```

bob-carpenter:
> is there some way we could write functions like normal_lpdf_par() that takes in some chunk size for the first input and then would just do reduce_sum in the function?

Yes.  But they'd have to all be rewritten.  Right now, they use the operators-and-partials structure that has the nice property of avoiding duplicate subexpression work and working at all levels of autodiff.

bob-carpenter:
Right now, we use operands-and-partials to deal with avoiding repeated
subexpressions and to work at all levels of autodiff.

That would all have to be rewritten unless somehow the sums in
operands-and-partials itself could be parallelized.

wds15:
@SteveBronder skimming over this things look really nice so far. Need to look more carefully.

Wrt. to making our lpdf's parallel: This will be more subtle, since it's not only the first argument which can be a vector-like structure. It can be any argument which is vector-like and the crux is that you will need to slice all of the vector-like arguments. So that's a bit more involved, I think.

SteveBronder:
@wds15 ty! One thing I'd like for you to do when you have a minute is go through and fill in the parts of the docs I missed (i.e. `operator()` for each class, returns, etc. etc.)

SteveBronder:
Before we merge any of the reduce_sum stuff should we make a seperate PR with all of the non reduce_sum changes? It looks like that's where the errors in the tests are popping up

bbbales2:
deep_copy, accumulate_adjoints, count_var, and save_varis should all be broken out and tested.

I use them in the ODE stuff as well. Can do this myself maybe tomorrow?

SteveBronder:
That would be rad!

SteveBronder:
Also apply. Could be nice to have the left_fold but we also don't use it

stan/math/prim/functor/reduce_sum.hpp
```diff
@@ -20,76 +20,113 @@ namespace math {
...
+ * `ReduceFunction::operator()`
+ */
 template <typename ReduceFunction, typename Enable, typename ReturnType,
           typename Vec, typename... Args>
 struct reduce_sum_impl {
-  return_type_t<Vec, Args...> operator()(const Vec& vmapped,
-                                         std::size_t grainsize,
-                                         std::ostream* msgs,
-                                         const Args&... args) const {
+  template <typename OpVec, typename... OpArgs,
```
bbbales2: Why template on OpVec and OpArgs if we enforce that they be the same types as Vec and Args?

stan/math/prim/functor/reduce_sum.hpp
```diff
@@ -20,76 +20,113 @@ namespace math {
...
-    recursive_reducer(const vmapped_t& vmapped, std::ostream* msgs,
-                      const Args&... args)
-        : vmapped_(vmapped), msgs_(msgs), args_tuple_(args...), sum_(0.0) {}
+    template <typename OpVec, typename... OpArgs,
+              typename = require_same_t<Vec, OpVec>,
+              typename = require_all_t<std::is_same<Args, OpArgs>...>>
+    recursive_reducer(OpVec&& vmapped, std::ostream* msgs, OpArgs&&... args)
+        : vmapped_(std::forward<OpVec>(vmapped)),
+          msgs_(msgs),
+          args_tuple_(std::forward<OpArgs>(args)...) {}
```
bbbales2: So with the forwarding and the && notation, if we pass in a reference, we'll get a tuple with a reference (not a copy of the object)?

stan/math/prim/functor/reduce_sum.hpp
```diff
@@ -20,76 +20,113 @@ namespace math {
...
-          args_tuple_);
+          this->args_tuple_);
     }
-
-    void join(const recursive_reducer& child) { sum_ += child.sum_; }
+    /**
+     * Join the sum of the reducers
+     * @param child A sub-reducer that is aggregated to make the final reduced
+     * value.
+     * @note side effect of updating `recursive_reducer->sum_`
```
bbbales2: This isn't a side effect, it's the effect!

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -18,227 +18,346 @@ namespace stan {
...
           typename... Args>
 struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
                        Vec, Args...> {
-  template <typename T, require_arithmetic_t<scalar_type_t<T>>...>
-  static inline const T& deep_copy(const T& arg) {
-    return arg;
-  }
+  /**
+   * Internal object used in `tbb::parallel_reduce`
+   * @note see link [here](https://tinyurl.com/vp7xw2t) for requirements.
```
bbbales2: I don't think we need to say more than it's for tbb::parallel_reduce. I think I've left links in docs before but that doesn't seem reliable.

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -18,227 +18,346 @@ namespace stan {
...
+   */
+  struct recursive_reducer {
+    size_t num_shared_terms_;  // Number of terms shared across threads
+    double* sliced_partials_;  // Points to adjoints of the partial calculations
+    Vec vmapped_;
```
bbbales2: So if I pass in a ```const std::vector<T>&```, Vec will be a ```const std::vector<T>&```?

What if I pass in a ```std::vector<T>```? Will I get a full vector here or a reference?

SteveBronder: Since this is used in parallel reduce, and it's signature is all references, the types we can get back are 

```cpp
std::vector<T>&
const std::vector<T>&
std::vector<T>&&
```

If you passed in an `std::vector<T>` to reduce_sum_impl then `Var` would be equal to `std::vector<T>`

stan/math/rev/functor/reduce_sum.hpp
SteveBronder: The url here is a wayback machine link. I think for future maintenance it would be good to know why it's built with an inner struct and has the types it does etc.

stan/math/prim/functor/reduce_sum.hpp
SteveBronder: I can change this, but I've always thought of side effects as "things manipulated outside of the function that is not the return". like R's `plot()`. does that make sense?

stan/math/prim/functor/reduce_sum.hpp
SteveBronder: If we pass in an lvalue reference (a single `&` like `std::vector<T>&` or `const std::vector<T>&`) it will pass a normal reference to the args. If we pass in an rvalue ( `std::vector<T>&&`) then it knows the ownership of the underlying data for the vector can be moved from the input argument to the class's member. The link below is useful for this concept

https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers

stan/math/prim/functor/reduce_sum.hpp
SteveBronder: So we could get rid of them, having them allows the value type of `Vec` and `OpVec` to differ. (i.e. `Vec` could be `std::vector<T>&` and `OpVec` could be `std::vector<T>&&`). Removing them means that we couldn't have the value types in the template of the class differ from the constructor's templates. But idt that would be a big deal

bbbales2: Given we do the calls like:

```
internal::reduce_sum_impl<ReduceFunction, void, return_type, Vec,
                                   Args...>()(
      std::forward<Vec>(vmapped), grainsize, msgs, std::forward<Args>(args)...)
```

I don't think they're ever different. It makes me wonder if reduce_sum_impl should just be a function.

bbbales2: It has no member variables, and we're only calling operator() right after construction.

SteveBronder: I'm afk but theres a link in the docs to the tbb site and I think reduce sum needs to be a struct/class

stan/math/prim/functor/reduce_sum.hpp
SteveBronder: But I do agree the whole class could be simplified. Maybe it could be a lambda?

bbbales2: I think it'll need to be a function. There's the generic version in prim (no threading, full high order autodiff support), a version in prim that supports only arithmetic types + is threaded, and then a version in rev specialized for vars.

I'm gonna leave it as is this pass through. After I push stuff tonight can you take a look at it? I think the type logic would be cleaner as a function as well.

stan/math/prim/functor/reduce_sum.hpp
wds15: The reason I went this way was to have the template magic pick the right version. This template magic does not work so well for functions as I recall.

@SteveBronder it's right that parallel_reduce needs a struct, but you could define that within the body of a function if you want to.

Somehow I prefer the struct; feels safer... but whatever works.

bbbales2: In the previous version there were member variables if I recall correctly. I think there was a thread_local pointer to copies of what were the shared arguments or something?

wds15: Yes, I tried that, but for the moment I do not see a safe way to make sure things stay (and get cleaned) from the thread local storage thing. I tried on the v6 branch to introduce a new `local_vari` thing for the deep copies. This is probably the way to go at some later point; right now I am still getting failed unit tests on that branch for reasons not so clear to me now... this can be revisited later.

PR #1769, opened 2020-03-10
Add transposition to kernel generator

t4c1:
## Summary
Adds transposition to kernel generator. Existing transposition kernel is removed.

## Tests
Added new tests for transposition. Existing tests in `opencl/prim/transpose_test.cpp` also test the new code.

## Side Effects
None.

## Checklist

- [x] Math issue #1342


stan/math/opencl/kernel_generator/transpose.hpp
```diff
@@ -0,0 +1,131 @@
...
+#include <type_traits>
+#include <utility>
+
+namespace stan {
+namespace math {
+
+/**
+ * Represents a transpose in kernel generator expressions.
+ *
+ * Warning: transposing this expression is not supported!
```
SteveBronder: This warning is kind of confusing, like we can't do `transpose(transpose(x))`?

stan/math/opencl/kernel_generator/transpose.hpp
```diff
@@ -0,0 +1,131 @@
...
+ * @tparam Derived derived type
+ * @tparam T_a type of first argument
+ * @tparam T_b type of second argument
```
SteveBronder: These don't match up

SteveBronder: Also this may be a better place for clearer names like calling `T` `Expr`

stan/math/opencl/kernel_generator/transpose.hpp
```diff
@@ -0,0 +1,131 @@
...
+  explicit transpose_(T&& a) : base(std::forward<T>(a)) {}
+
+  /**
+   * Creates a deep copy of this expression.
+   * @return copy of \c *this
+   */
+  inline transpose_<std::remove_reference_t<T>> deep_copy() {
+    return transpose_<std::remove_reference_t<T>>{
+        std::get<0>(arguments_).deep_copy()};
+  }
```
SteveBronder: This is kinda weird to me. Let's say we call 
```cpp
double b = 10;
transpose_<double> foo(b);
``` 
(or insert any non ref type). Then the constructor here is going to be 

```cpp
explicit transpose_(double&& a) : base(std::forward<double>(a)) {}
```

Where `std::forward<double>(a)` here is actually going to give back an rvalue reference since forward with an lvalue as the template and rvalue as the type still produces an rvalue (see godbolt below). 

I think you need to have a separate template here
```cpp
template <typename OpT, require_same_t<T, OpT>* = nullptr>
explicit transpose_(OpT&& a) : base(std::forward<OpT>(a)) {}
```

The godbolt here shows this effect.

https://godbolt.org/z/pWNzSj

stan/math/opencl/kernel_generator/transpose.hpp
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * generates kernel code for this and nested expressions.
+   * @param[in,out] generated set of already generated operations
+   * @param ng name generator for this kernel
+   * @param i row index variable name
+   * @param j column index variable name
+   * @return part of kernel with code for this and nested expressions
+   */
+  inline kernel_parts generate(const std::string& i, const std::string& j,
+                               const std::string var_name_arg) const {
```
SteveBronder: Why pass this string by value?

stan/math/opencl/kernel_generator/transpose.hpp
t4c1: Whoops, this is not true anymore. Removed.

stan/math/opencl/kernel_generator/transpose.hpp
t4c1: fixed

stan/math/opencl/kernel_generator/transpose.hpp
t4c1: Having rvalue if rvalue is passed in here is fine. Your example does not compile because you have wrong template argument in `S<int> a(b);` - this should be `S<int&> a(b);`

stan/math/opencl/kernel_generator/transpose.hpp
t4c1: No reason. Fixed.

stan/math/opencl/kernel_generator/transpose.hpp
t4c1: Also in your example in the comment you have an error - the constructor for 
```
double b = 10;
transpose_<double> foo(b);
```
would be:
```
explicit transpose_(double& a) : base(std::forward<double&>(a)) {}
```

SteveBronder:
Odd it looks like this is failing the cholesky test? Ping me when it's fixed and I'll do the review

SteveBronder:
(also fyi feel free to ping me whenever to review these)

t4c1:
@SteveBronder This is (finaly) ready for a review.

SteveBronder:
Cool! I'll take a look tmrw

stan/math/opencl/tri_inverse.hpp
```diff
@@ -81,7 +81,7 @@ inline matrix_cl<T> tri_inverse(const matrix_cl<T>& A) {
   zero_mat.template zeros<stan::math::matrix_cl_view::Entire>();
   inv_padded.template zeros<stan::math::matrix_cl_view::Entire>();
   if (tri_view == matrix_cl_view::Upper) {
-    inv_mat = transpose(inv_mat);
+    inv_mat = transpose(inv_mat).eval();
```
SteveBronder: Why is there a need for `.eval()` here? Shouldn't the kernel kick off once it's being assigned to an already constructed `matrix_cl`?

test/unit/math/opencl/kernel_generator/transpose_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <stan/math/opencl/kernel_generator.hpp>
+#include <stan/math/opencl/matrix_cl.hpp>
+#include <stan/math/opencl/copy.hpp>
+#include <test/unit/math/opencl/kernel_generator/reference_kernel.hpp>
+#include <stan/math.hpp>
+#include <gtest/gtest.h>
+#include <string>
+
+using Eigen::MatrixXd;
+using stan::math::matrix_cl;
```
SteveBronder: I think its usually good practice to put these inside of the tests than have them floating in global

stan/math/opencl/tri_inverse.hpp
t4c1: The problem is that without eval we have aliasing issues, similar as [Eigen has](https://eigen.tuxfamily.org/dox/group__TopicAliasing.html). Since source and destination is the same matrix threads that work on lower/upper triangular part can (and do) overwrite each other's input values with their outputs. With eval we create new matrix for destination. That is also how individual transpose kernel works.

t4c1:
Yep, the paper is almost complete, but it does not go in much more details than the design doc I wrote. I will think about adding something to the site. Remind me, where is the source for that?

test/unit/math/opencl/kernel_generator/transpose_test.cpp
t4c1: fixed

SteveBronder:
>  Remind me, where is the source for that?

The source for the site? 

I think if we had something like the below it would be fine

```
/**
 * \ingroup opencl
 * \defgroup opencl_kernel_generator OpenCL Kernel Generator
 * [Brief intro and link to paper]
 */
```

SteveBronder:
I think just including the paper is fine as long as you have some details on the different types of optimizations going on in here

SteveBronder:
glad we did that design doc it made my understanding of this whole thing a lot more clear!

stan/math/opencl/tri_inverse.hpp
SteveBronder: Okay if you can file an issue to write some docs about this (and then add the docs later) then I'm cool with approving this. We should probably have something like Eigen does if we also have the same aliasing issues.

You can either add that as a module (like in the below comment) or follow the instructions [here](https://github.com/stan-dev/math/blob/develop/doxygen/howto_add_page.md) on adding a new page

t4c1:
I added docs about aliasing to doxygen. Extended documentation about kernel generator will come in its own PR (I plan two more PRs before that).

Issue #1679, opened 2020-02-06
Derivatives of log_sum_exp are wrong for large arguments


martinmodrak:
## Description

When both arguments to `log_sum_exp` are large, the reverse mode derivatives can be wrong.

## Example & expected output

We can arrive at a simple identity:
`f(x) = d/dx(log_sum_exp(x,x)) == 1` for all `x`.

Currently Stan computes

```
f(1e10) = 0.99999952125830205
f(1e50) = 2
```

#### Current Version:
v3.1.0

PR #1772, opened 2020-03-10
Changed calculation of log_sum_exp(x1, x2)

pgree:
## Summary
The partial derivatives of `log_sum_exp(x, y)` were losing accuracy in the regime where `x` and `y` were equal and greater than around `1e14`. In the original `log_sum_exp(x1, x2)` function, the derivative of `log_sum_exp` with respect to `x1` was evaluated by this formula:

`exp(x1 - (y + log(exp(x1 - y) + exp(x2 - y))))`

 where `y` is the maximum of `x1` and `x2`. This will result in a loss of accuracy when `y` is close to `x1` and `log(exp(x1-y) + exp(x2 - y))` is small. For example, when `x1 = 1e20` and `x2 = 1e20`

`x1 - (y + log(exp(x1 - y) + exp(x2 - y)))`

evaluates to `0`.

I changed the calculation of the partial derivative of `log_sum_exp` with respect to `x1` to be 

`1 / (1 + exp(x2 - x1))` 

and the partial derivative of `log_sum_exp` with respect to `x2` to be 

`1 / (1 + exp(x1 - x2))`.

## Tests

I wrote a couple of new tests to check that the partials are accurate with large and equal argument values as in @martinmodrak 's issue raised in #1679.

## Side Effects

None

## Checklist

- [x] Math issue #1679 


stan/math/rev/fun/log_sum_exp.hpp
```diff
@@ -22,8 +22,8 @@ class log_sum_exp_vv_vari : public op_vv_vari {
   log_sum_exp_vv_vari(vari* avi, vari* bvi)
       : op_vv_vari(log_sum_exp(avi->val_, bvi->val_), avi, bvi) {}
   void chain() {
-    avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_);
```
bbbales2: Would it be possible to get rid of calculate_chain everywhere? It seems like a strange function.

Also used here (which should be fixed as part of this pull):
https://github.com/stan-dev/math/blob/b5133581d24162e75ca5d04a5a39d43205707767/stan/math/rev/fun/log_sum_exp.hpp#L37

In stan/math/rev/fun/log_diff_exp.hpp and in stan/math/rev/fun/log1p_exp.hpp.

test/unit/math/rev/fun/log_sum_exp_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+  using stan::math::var;
+
+  var a = 1e50;
+  var output = stan::math::log_sum_exp(a, a);
+  output.grad();
+  EXPECT_FLOAT_EQ(output.val(), log(2.0) + value_of(a));
+  EXPECT_FLOAT_EQ(a.adj(), 1.0);
+
+  var a1 = 1e50;
+  var a2 = 1;
```
bbbales2: Once you remove the other calculate_chain thing we'll need some tests where a1 is a var and a2 is a double.

bob-carpenter:
The first rule of naming is that names should be short conventional placeholders where possible and mean something everywhere else.  The second rule of naming is that names should be as short as possible, because big ones obfuscate the structure of your code and we want to do everything in our power to make the actual code readable.   The third rule is to use your judgement in balancing the first and second rules.

With `compute_chain`, the `compute` part isn't really adding anything because every function does computation.  That goes for other generic naming conventions like `_helper` and `do_it`.

The `chain` part is misleading because there's no object named "chain" being computed.  The method `chain()` is so named because it's responsible for applying the chain rule.  

What is being computed?  A partial derivative.  So we can call the function something like `log_sum_exp_partial`, which describes what it returns and will read well in code.  

Finally, always let your functions pull things apart if it doesn't involve extra overhead.  So you can pass the `avi_` pointer and let the function pull out `avi_->val_`.  That helps get rid of code duplication when it's called twice.

P.S.  This was just to give you an idea of how to think about naming.  We're not going to micromanage every name in every program!

pgree:
Thanks @bob-carpenter and @bbbales2 

Is it better to change the function `calculate_chain(x, y)` to compute the new formula for the derivative (and change the name of the function)? Or should I replace the call to the function everywhere it appears with something like

`1 / (1 + exp(x2 - x1))`?

bob-carpenter:
Keep the function (it removes code duplication) and rename it (so it self documents).

bbbales2:
> calculate_chain(x, y)

I think it makes sense to get rid of this one.

bob-carpenter:
Ben's the official reviewer, so his opinion trumps mine in this one!

It's really not as big a deal as we're making it here.  It's just good to set out conventions when you start coding so you don't have to think later.

bbbales2:
Mm, I'm not convinced we should keep it. Really not much code duplication.

```
stan/math/rev/fun/log1p_exp.hpp:  void chain() { avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_); }

stan/math/rev/fun/log_sum_exp.hpp:    avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_);
stan/math/rev/fun/log_sum_exp.hpp:    bvi_->adj_ += adj_ * calculate_chain(bvi_->val_, val_);
stan/math/rev/fun/log_sum_exp.hpp:      avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_);
stan/math/rev/fun/log_sum_exp.hpp:      bvi_->adj_ += adj_ * calculate_chain(bvi_->val_, val_);
stan/math/rev/fun/log_sum_exp.hpp:      vis_[i]->adj_ += adj_ * calculate_chain(vis_[i]->val_, val_);

stan/math/rev/fun/log_diff_exp.hpp:    avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_);
stan/math/rev/fun/log_diff_exp.hpp:      avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_);
```

It also lives in rev, which I think is the wrong place and the comment on the implementation doesn't seem right:
```
inline double calculate_chain(double x, double val) {
  return std::exp(x - val);  // works out to inv_logit(x)
}
```

bbbales2:
And honestly is that just wrong? It just seems wrong to me. What is happening here.

```exp(x - val)``` is not ```1 / (1 + exp(x - val))``` (or ```1 / (1 + exp(val - x))``` or whatever)

**Comment by pgree** [comment]
In `log_sum_exp.hpp` the function `calculate_chain(x, val)` was being called with 

`val = log_sum_exp(x, y)`. 

So it works out that 

`calculate_chain(x, log_sum_exp(x, y)) = 1 / (1 + exp(y - x))`.

I need to check how it's being used in the other functions to make sure that this switch would work.

**Comment by bbbales2** [comment]
Aaaah, thanks for that clarification.

**Comment by andrjohns** [comment]
Quick suggestion (not a blocking review comment), some of these could use ```inv_logit```, which would be more resistant to over-/under-flow.

For ```log_sum_exp```:
```
  void chain() {
    avi_->adj_ += adj_ * inv_logit(avi_->val_ - bvi_->val_);
    bvi_->adj_ += adj_ * inv_logit(bvi_->val_ - avi_->val_);
  }
```

For ```log1p_exp```:
```
  void chain() { avi_->adj_ += adj_ * inv_logit(avi_->val_); }
```

**Comment by martinmodrak** [comment]
Thanks for handling this! Few thoughts: 

I actually played with this issue in https://github.com/stan-dev/math/pull/1677 (as it was a conveninent example to play with). As the tools for testing identitites are not going to be ready soon (we decided with @bob-carpenter to take the longer route there), I am however not sure you can get much from it (although I do use a bit more identities for testing).

I second @andrjohns on the `inv_logit` suggestion. 

Generally, I would suggest to make each test loop over a range of values instead of using just one value.

File: stan/math/rev/fun/log_diff_exp.hpp [hunk 3]
```diff
@@ -17,7 +16,7 @@ class log_diff_exp_vv_vari : public op_vv_vari {
   log_diff_exp_vv_vari(vari* avi, vari* bvi)
       : op_vv_vari(log_diff_exp(avi->val_, bvi->val_), avi, bvi) {}
   void chain() {
-    avi_->adj_ += adj_ * calculate_chain(avi_->val_, val_);
+    avi_->adj_ -= adj_ / expm1(bvi_->val_ - avi_->val_);
```
bbbales2: There's no way to use inv_logit here? If so, that's fine, just checking.

bob-carpenter: I don't see how you could because one has the extra 1 term inside and the other outside.

```
1 / expm1(x) = 1 / exp(x - 1)
inv_logit(u) = 1 / (1 + exp(-u))
```

bbbales2: Oh good point. I was too lazy to actually think through the math at all.

Issue #1635, opened 2020-01-20
Deeply nested containers and error checks


peterwicksstringfield:
## Description
check_not_nan (and friends) work element-wise on containers, but we only allow them to work on containers that are one level deep. Its easy to allow them to work on deeper containers, but we want indices in the error message and that makes things a bit more complicated. Basically, recursively traverse the containers and check for nans at the bottom, then, if we find a nan, collect indices on the way back up the call stack and build the error message at the top.

If we don't find a nan, then we don't want to be building up error messages that we end up not using, which makes this sort of implementation unsatisfactory:
```
void check_not_nan(function, name, y) {
  for (size_t n = 0; n < size(y); n++) {
    std::string name_ = std::string(name) + "[" + std::to_string(n) + "]";
    check_not_nan(function, name_.c_str(), stan::get(y, n));
  }
}
```

Also, parameterize the check so the other check functions can use the same code.

I think I remember @bob-carpenter talking about this recently but I can't find where. Anyway it came up in #1626 

## Example
```
std::vector<std::vector<double> > x;
x.push_back({1, 2, 3});
x.push_back({1, 2, 3, 4});
x.push_back({1, 2, stan::math::NOT_A_NUMBER});
x.push_back({1, 2});
check_not_nan("some_function", "x", x);
```

## Expected Output
"some_function: x[2][2] is nan, but must not be nan!"

#### Current Version:
v3.0.0

bob-carpenter:
Yes, I'd really like to generalize all the check functions to deal with arbitrary containers.  Some of them might not even be vectorized.  I think the way to do it is to have a basic check function and then a general vectorized driver that keeps track of indices.  

What we'd talked about before is breaking these into boolean functions that are called by the check functions that throw.  

We definitely can't build up error messages proactively.  But I'm not sure how we can collect indices going backwards.  Performance isn't such a big deal after deciding to throw, because we should only be throwing in exceptional conditions. 

There's also been a lot of interest in turning them off altogether with some kind of flag, which could easily be rolled into a general implementation.

peterwicksstringfield:
I'm super on board with anything that cleans up the error checking and testing. The more little stuff like this we can clean up the easier it is to develop the new gpu/parallel stuff.

I was thinking we might have a signature like:
```
template <typename T_to_check, typename pred_type>
do_check( const T_to_check& to_check
        , const pred_type& predicate
        , const char* function
        , const char* name); 
```
[...]

Issue #381, opened 2016-09-03
Error functions for general std::vector types


rtrangucci:
#### Summary:

We need error functions for general std::vector types.

#### Description:
Currently, we don't have `check` functions that can be called directly on 

*`std::vector<Eigen::Matrix<T_s, R, C>`
*`std::vector<std::vector<T_s> >`

For example, it isn't possible to do:

```
void foo(const std::vector<Eigen::MatrixXd>& x) {
  check_not_nan(x);
}
```

Instead, we need to loop through the vector and call `check` on each element:

```
void foo(const std::vector<Eigen::MatrixXd>& x) {
  for (size_t i = 0; i < x.size(); ++i)  
    check_not_nan(x[i]);
}
```

This makes the exceptions that are thrown by the element-wise `check`s inscrutable because the index of the vector is lost to the exception.

#### Current Version:
v2.11.0

bob-carpenter:
If you create an issue, please add labels, milestone and assignee.  Thanks.  I added them for this one.

rtrangucci:
Thanks, Bob. Sorry about that.

bob-carpenter:
There are examples already of functions that test vectors that we can follow to extend the ones that aren't defined for vectors yet.  We should really extend all of our error functions to work componentwise on containers of all types.  A recursive template pattern should be able to generalize to all of our containers following the pattern we used for vectorizing scalar functions.  Somehow we need to figure out how to make the error messages lazy, though, because it'd be too expensive to start passing down strings that get constructed every time we test bounds.

bob-carpenter:
See #366 for a related issue.

PR #1780, opened 2020-03-15
Feature/elementwise check

peterwicksstringfield:
## Summary

First part of a PR to address #1635 and #381. Thanks @bob-carpenter for the use the parameter pack in raise_error.

Implement elementwise_check / elementwise_is, which take a predicate and apply it to every scalar inside a (nested) container. If the predicate fails, they throw / return false.

Part two of this PR reimplements check_not_nan, check_positive, check_nonnegative, check_finite, is_scal_finite, is_positive, and is_not_nan in terms of these functions and eliminates a bunch of boilerplate loops.

We can do this: check_not_nan(..., some_double).
And this: check_not_nan(..., some_vector_of_doubles).
But we have to do this:
```
for (size_t i = ...)
  check_not_nan(..., some_vector_of_vector_of_doubles[i])
```
And we really want to just do this: check_not_nan(..., some_vector_of_vector_of_doubles).
And now we can.

The diff got kinda big, so I split it into two parts.

(Likewise .)

```
some_function: x[2][4][row=1, col=3] is nan, but must not be nan!
```

## Tests

Tests for new functions elementwise_check and elementwise_is.

## Side Effects

None.

## Checklist

- [X] Math issue #1635 (don't close)

### DIFF
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -452,15 +452,15 @@ class matrix_cl<T, require_arithmetic_t<T>> {
   }
 
   /**
-   * Construct from a kernel generator expression. It evaluates the ixpression
+   * Construct from a kernel generator expression. It evaluates the expression
    * into \c this.
    * @tparam Expr type of the expression
    * @param expression expression
    */
   template <typename Expr,
             require_all_valid_expressions_and_none_scalar_t<Expr>* = nullptr>
-  matrix_cl(const Expr& expresion);  // NOLINT This constructor is intentionally
-                                     // implicit
+  matrix_cl(const Expr& expression);  // NOLINT This constructor is
+                                      // intentionally implicit
 
   /** \ingroup opencl
    * Move assignment operator.
@@ -492,14 +492,14 @@ class matrix_cl<T, require_arithmetic_t<T>> {
   }
 
   /**
-   * Assignment of a kernel generator expression evaluates the ixpression into
+   * Assignment of a kernel generator expression evaluates the expression into
    * \c this.
    * @tparam Expr type of the expression
    * @param expression expression
    */
   template <typename Expr,
             require_all_valid_expressions_and_none_scalar_t<Expr>* = nullptr>
-  matrix_cl<T>& operator=(const Expr& expresion);
+  matrix_cl<T>& operator=(const Expr& expression);
 
  private:
   /** \ingroup opencl
--- a/stan/math/opencl/matrix_cl.hpp
+++ b/stan/math/opencl/matrix_cl.hpp
@@ -459,8 +459,7 @@ class matrix_cl<T, require_arithmetic_t<T>> {
    */
   template <typename Expr,
             require_all_valid_expressions_and_none_scalar_t<Expr>* = nullptr>
-  matrix_cl(const Expr& expression);  // NOLINT This constructor is
-                                      // intentionally implicit
+  matrix_cl(const Expr& expression);  // NOLINT(runtime/explicit)
 
   /** \ingroup opencl
    * Move assignment operator.
--- a/stan/math/prim/err/elementwise_check.hpp
+++ b/stan/math/prim/err/elementwise_check.hpp
@@ -1,0 +1,135 @@
+#ifndef STAN_MATH_PRIM_ERR_ELEMENTWISE_ERROR_CHECKER_HPP
+#define STAN_MATH_PRIM_ERR_ELEMENTWISE_ERROR_CHECKER_HPP
+
+#include <stan/math/prim/err/throw_domain_error.hpp>
+#include <stan/math/prim/fun/get.hpp>
+#include <stan/math/prim/fun/size.hpp>
+#include <stan/math/prim/fun/value_of_rec.hpp>
+#include <string>
+#include <sstream>
+#include <vector>
+
+namespace stan {
+namespace math {
+namespace internal {
+
+template <typename T_is_good, typename T_exception>
+struct Checker {
+  const T_is_good& is_good;
+  const char* function;
+  const char* name;
+  const char* suffix;
+
+  void raise_error_impl(std::stringstream& ss) { throw T_exception{ss.str()}; }
+
+  template <typename... M>
+  void raise_error_impl(std::stringstream& ss, const char* message,
+                        const M&... messages) {
+    ss << message;
+    raise_error_impl(ss, messages...);
+  }
+
+  template <typename... M>
+  void raise_error_impl(std::stringstream& ss, double value_of_x,
+                        const M&... messages) {
+    ss << value_of_x;
+    raise_error_impl(ss, messages...);
+  }
+
+  template <typename... M>
+  void raise_error_impl(std::stringstream& ss, size_t index,
+                        const M&... messages) {
+    ss << index + 1;
+    raise_error_impl(ss, messages...);
+  }
+
+  template <typename... M>
+  void raise_error(const M&... messages) {
+    std::stringstream ss{};
+    raise_error_impl(ss, messages...);
+  }
+
+  template <typename T, typename = require_stan_scalar_t<T>,
+            typename... T_indices>
+  void check(const T& x, T_indices... indices) {
+    double xd = value_of_rec(x);
+    if (!is_good(xd))
+      raise_error(function, ": ", name, indices..., " is ", xd, suffix);
+  }
+
+  template <typename T, typename = require_eigen_vector_t<T>, typename = void,
+            typename... T_indices>
+  void check(const T& x, T_indices... indices) {
+    for (size_t i = 0; i < stan::math::size(x); ++i)
+      check(x(i), indices..., "[", i, "]");
+  }
+
+  template <typename T, typename... T_indices>
+  void check(const std::vector<T>& x, T_indices... indices) {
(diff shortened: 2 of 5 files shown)


stan/math/opencl/matrix_cl.hpp
```diff
@@ -452,15 +452,15 @@ class matrix_cl<T, require_arithmetic_t<T>> {
...
+   * Construct from a kernel generator expression. It evaluates the expression
    * into \c this.
    * @tparam Expr type of the expression
    * @param expression expression
    */
   template <typename Expr,
             require_all_valid_expressions_and_none_scalar_t<Expr>* = nullptr>
-  matrix_cl(const Expr& expresion);  // NOLINT This constructor is intentionally
-                                     // implicit
+  matrix_cl(const Expr& expression);  // NOLINT This constructor is
```
bob-carpenter: We've just used `NOLINT(runtime/explicit)` I think elsewhere to indicate we're explicitly turning off just that piece of lint.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <sstream>
+#include <vector>
+
+namespace stan {
+namespace math {
+namespace internal {
+
+template <typename T_is_good, typename T_exception>
+struct Checker {
+  const T_is_good& is_good;
```
bob-carpenter: Use `F` and `E` as template types (with longer names if you really insist---I'm trying to fight this trend in our code).  The `T_` is confusing as we've generally used that for some kind of arithmetic type everywhere.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <stan/math/prim/fun/value_of_rec.hpp>
+#include <string>
+#include <sstream>
+#include <vector>
+
+namespace stan {
+namespace math {
+namespace internal {
+
+template <typename T_is_good, typename T_exception>
```
bob-carpenter: Every class needs doc, including the template parameters.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+struct Checker {
+  const T_is_good& is_good;
+  const char* function;
+  const char* name;
+  const char* suffix;
+
+  void raise_error_impl(std::stringstream& ss) { throw T_exception{ss.str()}; }
```
bob-carpenter: Functions also need doc.  

Also, `_impl` should be something like `_stream` or `_ss` to give readers a hint at how this function is different than `raise_error`.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+  template <typename... M>
```
bob-carpenter: These next three functions can probably be replaced with one where `T&&` is the type of `message`.  That might not work for the `char*` version, but will definitely work for the arithmetic types and should also be generalized so that there's not ambiguity on other arithmetic types.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+      check(x(i), indices..., "[", i, "]");
+  }
+
+  template <typename T, typename... T_indices>
+  void check(const std::vector<T>& x, T_indices... indices) {
+    for (size_t i = 0; i < stan::math::size(x); ++i)
+      check(x[i], indices..., "[", i, "]");
+  }
+
+  template <typename T, typename... T_indices>
```
bob-carpenter: [question]
Would this not cover the Eigen vector case, too?

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+struct Iser {
+  const T_is_good& is_good;
+
+  template <typename T, typename = require_stan_scalar_t<T>>
+  bool is(const T& x) {
+    return is_good(value_of_rec(x));
+  }
+
+  template <typename T, typename = require_not_stan_scalar_t<T>,
+            typename = void>
```
bob-carpenter: [question]
Is this dummy just so `require` and `require_not` can be used with the same template signature?

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+  template <typename T, typename... T_indices>
+  void check(const Eigen::DenseBase<T>& x, T_indices... indices) {
+    for (size_t n = 0; n < x.cols(); ++n)
+      for (size_t m = 0; m < x.rows(); ++m)
+        check(x(m, n), indices..., "[row=", m, ", col=", n, "]");
+  }
+};
+
+template <typename T_is_good>
```
bob-carpenter: I don't understand why this is alled `Iser`.  The doc for the class should explain what it does.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -0,0 +1,136 @@
...
+ * @param is_good predicate to check, must accept doubles and produce bools
+ * @param function function name (for error messages)
+ * @param name variable name (for error messages)
+ * @param x variable to check, can be a scalar, a container of scalars, a
+ * container of containers of scalars, etc
+ * @param suffix message to print at end of error message
+ * @throw <code>std::domain_error</code> if is_good returns false for the value
+ * of any element in x the predicate
+ */
+template <typename T_is_good, typename T_x>
```
bob-carpenter: `F` for functions, and just `T` for `T_x`.  All that `_x` stuff in the densities is because there are multiple `T_` to disambiguate.  Here, we can be more informative in the basic names.

test/unit/math/prim/err/elementwise_check_test.cpp
```diff
@@ -0,0 +1,149 @@
+#include <test/unit/util.hpp>
+#include <stan/math/prim.hpp>
+#include <stan/math/prim/fun/is_nan.hpp>
+#include <gtest/gtest.h>
+#include <limits>
+#include <vector>
+
+template <typename T>
```
bob-carpenter: Thanks for testing the framework!

test/unit/math/prim/err/elementwise_check_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+bool example_predicate(const T& x) {
+  return !stan::math::is_nan(x);
+}
+
```
bob-carpenter: [optional]
This'd be easier just assigning a functor

```
auto p = [](const auto& x) { return !stan::math::is_nan(x); }
```

and then using it in the checks.

peterwicksstringfield: Yes it would. Thanks.

test/unit/math/prim/err/elementwise_check_test.cpp
peterwicksstringfield: :)

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: Fixed.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: By analogy with Checker. Checker does check_not_nan : Iser does is_not_nan. Check and Is are just standard Stan terminology for the two versions of the error handling functions.

Checker |-> ElementwiseCheck
Iser |-> ElementwiseIs

would be a possible renaming.

Doc added.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: Yes.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: Combined into one overload, guarded by require_vector_t. Thanks. I thought I tried that but obviously I did not.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: They can all be replaced with one function. I broke them out specifically to avoid being general with the types. The only things that get passed to raise_error are const char*, double, and size_t.

I went ahead and replaced them.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: Fixed. Its _ss now.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: Fixed.

stan/math/prim/err/elementwise_check.hpp
peterwicksstringfield: Fixed.

stan/math/opencl/matrix_cl.hpp
peterwicksstringfield: Fixed.

peterwicksstringfield:
Thanks for reviewing.

stan/math/prim/err/elementwise_check.hpp
bob-carpenter: :-)  OK.  I think just saying "elementwise is" in the documentation would've helped me.  Same for checker.

Issue #1783, opened 2020-03-18
Eigen elementwise functions with autodiff variables


andrjohns:
## Description
Now that Eigen's SIMD-vectorised elementwise functions have been introduced for containers of ```doubles``` (in #1471 and #1727), they should also be added for containers of ```fvar<T>``` and ```var```. 

## Example
A quick example for using ```exp``` with ```fvar<T>```:
```
template <typename Container,
          require_container_st<is_container, is_fvar, Container>...>
inline auto exp(const Container& x) {
  return apply_vector_unary<Container>::apply(
      x, [](const auto& v) {
        auto exp_val = exp(v.val());
        return to_fvar(exp_val,exp_val.cwiseProduct(v.d()));
});
}
```

#### Current Version:
v3.1.0

PR #1785, opened 2020-03-20
Eigen elementwise functions with autodiff variables (#1)

andrjohns:
## Summary

This is the first batch of elementwise functions to add the use of Eigen's elementwise functions for calculating the values and gradients of containers of autodiff variables. The following functions are included in this pull:

- ```acos```
- ```asin```
- ```atan```
- ```ceil```
- ```cos```
- ```cosh```
- ```exp```
- ```fabs```
- ```floor```
- ```inv```

I can also break this pull out into smaller chunks if there's too much to review easily in this one.

## Tests
No tests added, as current behaviour should not change

## Side Effects
N/A

## Checklist

- [x] Math issue #1783 


SteveBronder:
It doesn't seem like the benchmarks we generally run cover these functions. Would you mind writing up a benchmark for this in perf_math?

andrjohns:
Good call! Ill get on that

andrjohns:
Good thinking Steve, looks like this actually does worse:
```
andrew@Lappinux:~/perf-math$ ./matstuff
2020-03-21 11:20:09
Running ./matstuff
Run on (8 X 4000 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
--------------------------------------------------
Benchmark           Time           CPU Iterations
--------------------------------------------------
Exp_Old      31270268 ns   31274997 ns         24
Exp_New      38605496 ns   38612270 ns         14
```
I'd guess that the performance benefits are outweighed by having to repeatedly traverse the input matrix

SteveBronder:
Huh! That's suprising, traversal could be it. Could you share the branch / try running it with

```
./matstuff --benchmark_repetitions=20 --benchmark_report_aggregates_only=true
```

SteveBronder:
You can also test it with multiple sizes by using something like 

```cpp
// in benchmark
// use state.range(0) etc to set sizes
// https://github.com/google/benchmark#passing-arguments
Eigen::Matrix<stan::math::var, -1, -1> Sigma(state.range(0), state.range(0))
//...
// run for sizes 100, 200, ..., 1900, 2000
BENCHMARK(mat_bench)->DenseRange(100, 2000, 100);
```

t4c1:
We probably won't be able to improve on appy_scalar_unary performance for autodiff variables until we redesign matrices. We need to store values and derivatives/adjoints in separate matrices to reduce amount of copying needed and enable vectorization.

bbbales2:
@bob-carpenter is interested in this too (https://github.com/stan-dev/math/issues/1787)

Could you share your code you used to bench this stuff? I've been surprised trying to optimize some related things as well.

Maybe a couple things to try:
1. Instead of the vector expressions, write for loops

2. Instead of computing the gradients in the chain call, precompute them in the forward pass (as you have the stuff in memory -- may have to do this with for loops) and then just do the adjoint * Jacobian thing in the chain itself (which is fine cause the Jacobians are diagonal).

andrjohns:
Thanks all! Benchmarking code is up in this branch: https://github.com/andrjohns/perf-math/tree/autodiff_elem_testing

I'll do some exploring of the other options

andrjohns:
Closing this while I look into other options, will post any further discussion in the relevant issue (#1783)

bob-carpenter:
Thanks for posting the benchmarking code.

PR #1797, opened 2020-03-24
Add broadcasting to kernel generator

t4c1:
## Summary

Adds broadcasting to kernel generator. If an expression is broadcasted in a dimmension it can be used as if it has any size in that dimmension when used in an operation that accepts more than one argument. The size is determined from other arguments to same expression.

Also contains a bugfix for determining view of the result.

## Tests

Added tests for broadcasting and bug.

## Side Effects

None.

## Release notes

Added broadcasting to kernel generator.

## Checklist

- [x] Math issue #1342 


stan/math/opencl/kernel_generator/broadcast.hpp
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Generates kernel code for this and nested expressions.
+   * @param[in,out] generated set of already generated operations
+   * @param ng name generator for this kernel
+   * @param i row index variable name
+   * @param j column index variable name
+   * @return part of kernel with code for this and nested expressions
+   */
+  inline kernel_parts generate(const std::string& i, const std::string& j,
+                               const std::string& var_name_arg) const {
```
SteveBronder: [seperate pr]

Reading this over wrt `modify_argument_indices` it's kind of confusing to read `i` and `j` here as the names of the index but in `modify_argument_indices` as the actual index values. It may be good here to change the argument names to something like `i_name` and `j_name`. Or maybe like `row_idx_name` and `col_idx_name`

stan/math/opencl/kernel_generator/broadcast.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Brodcast an expression in specified dimension(s). If broadcasting rowwise,
+ * the argument must have single column. If broadcasting colwise, the argument
+ * must have single row. Further expressions can use this expression as if it
+ * had any size in broadcasted dimension, repeating the values.
+ *
+ * Broadcasting evaluates argument expression multiple times. For performance
+ * reasons don't broadcast slow operations. Instead evaluate them in a separate
+ * kernel.
```
SteveBronder:
```suggestion
 * Broadcast an expression in specified dimension(s). If broadcasting rowwise,
 * the argument must have a single column. If broadcasting colwise, the argument
 * must have a single row. Further expressions can use this expression as if it
 * had any size in broadcasted dimension, repeating the values.
 *
 * Broadcasting evaluates the argument expression multiple times. For performance
 * reasons don't broadcast slow operations. Instead evaluate them in a separate
 * kernel.
```

test/unit/math/opencl/kernel_generator/broadcast_test.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  MatrixXd m1(3, 3);
+  m1 << 1, 2.5, 3, 4, 5, 6.3, 7, -8, -9.5;
+  MatrixXd m2(1, 3);
+  m2 << 10, 100, 1000;
+
+  matrix_cl<double> m1_cl(m1);
+  matrix_cl<double> m2_cl(m2);
+
+  auto tmp = m1_cl + stan::math::broadcast<true, false>(m2_cl);
```
SteveBronder: Do you think we should follow Eigen's API here and have `rowwise()` and `colwise()` defined in the matrix_cl class?

SteveBronder: I think then this would turn into 
```cpp
auto tmp = m1_cl.rowwise() + m2_cl;
```

t4c1: While it makes the code look simpler in this case, I find it harder to understand. Also for more complex expressions it becomes less clear. For example:
```
matrix_cl m1(3,3), m2(3,3), v(3,1);
m1 + broadcast<true, false>(v) + m2;
```
becomes (I think):
```
m1.colwise() + v + m2.colwise();
```

If you want to avoud explicit template parameters I can instead do separate functions `rowwise_broadcast()` and `colwise_broadcast()`.

SteveBronder: Ah yes okay that makes a lot of sense! If you don't mind adding those I think they would be nice helper functions that would make the API a little cleaner

t4c1: done

stan/math/opencl/kernel_generator/broadcast.hpp
```diff
@@ -0,0 +1,208 @@
...
+ * @return broadcast expression
+ */
+template <typename T,
+          typename = require_all_valid_expressions_and_none_scalar_t<T>>
+inline auto rowwise_broadcast(T&& a) {
+  return broadcast<false, true>(std::forward<T>(a));
+}
+
+/**
+ * Brodcast an expression in colwise dimmension. The argument must have single
```
SteveBronder, line 188: Can you just find/replace `Bordcast` with `Broadcast`

PR #1005, opened 2018-08-27
Use Eigen's allFinite to cut execution time of normal_glm (and others?) by 27%

seantalts:
For background discussion, see [this thread](https://discourse.mc-stan.org/t/potential-slowness-in-operands-and-partials/5343/15)
## Checklist

- [x] Copyright holder: (fill in copyright holder information)

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses: Columbia University
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [ ] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1798, opened 2020-03-24
feature/elementwise_check (2)

peterwicksstringfield:
## Summary

Allow check_not_nan and friends to work on nested containers. The next PR will eliminate a bunch of boilerplate loops.

These functions now operate on nested containers:
- check_not_nan
- check_nonnegative
- check_positive
- check_finite
- check_positive_finite
- is_not_nan
- is_nonnegative
- is_positive
- is_scal_finite
- is_positive_finite

Adds new functions is_nonnegative and is_positive_finite. This is part of #50, but it fits in with this PR really well.

Fix a bug where check_positive("someFunction", "x", 0u) would not throw a domain_error.

Improve the error messages for csr_u_to_z when called with out of range indices.

## Release Notes

Features: 

New functions is_nonnegative and is_positive_finite to parallel check_nonnegative and check_positive_finite. They signal failure by returning false instead of by throwing std::domain_error.

Functions check_not_nan, check_nonnegative, check_positive, check_finite, check_positive_finite, is_not_nan, is_nonnegative, is_positive, is_scal_finite, and is_positive_finite now operate on nested containers.

Clearer error messages when csr_u_to_z is called with out of range indices.

check_positive now throws domain error when given an unsigned 0.

## Tests

Tests for (check|is)_(nonnegative|positive) at 0.

Tests for the new functions is_nonnegative and is_positive_finite.

Tests to ensure that the vectorization is working.

Update some hardcoded error messages in other tests.

Fix the tests for csr_u_to_z.

## Side Effects

Changes the format of some error messages.
Old:
```
function: y[1] is nan, but must not be nan!
```
New:
```
function: y[row=1, col=1] is nan, but must not be nan!
```
Where y is a matrix that contains a nan.

## Release notes

New functions is_nonnegative and is_positive_finite to parallel check_nonnegative and check_positive_finite. They signal failure by returning false instead of by throwing std::domain_error.

Functions check_not_nan, check_nonnegative, check_positive, check_finite, check_positive_finite, is_not_nan, is_nonnegative, is_positive, is_scal_finite, and is_positive_finite now operate on nested containers.

Clearer error messages when csr_u_to_z is called with out of range indices.

check_positive now throws domain error when given an unsigned 0.

## Checklist

- [X] Math issue #1635 (don't close)


stan/math/prim/err/check_finite.hpp
```diff
@@ -1,87 +1,27 @@
...
- * @param name variable name (for error messages)
- * @param y matrix to test
- * @return <code>true</code> if the matrix is finite
- **/
-namespace internal {
-template <typename T, int R, int C>
-struct finite<Eigen::Matrix<T, R, C>, true> {
-  static void check(const char* function, const char* name,
-                    const Eigen::Matrix<T, R, C>& y) {
-    if (!value_of(y).allFinite()) {
```
peterwicksstringfield: We want this allFinite call I think. (It does need to be value_of_rec(y).) This gives Eigen a chance to do something clever. See also: https://github.com/stan-dev/math/pull/1005. I'm not really sure what to do here?

syclik: @peterwicksstringfield, are you suggesting this should be specialized for Eigen types? That would make sense to me. Especially if we could have a microbenchmark showing the difference in performance.

syclik:
@peterwicksstringfield, this is awesome! I think you brought up a comment about performance. It seems like it's something we need to check before we consider this change, if I understood your comment correctly. Otherwise, this change cleans up a lot of code; thank you!

stan/math/prim/fun/csr_u_to_z.hpp
```diff
@@ -19,12 +19,12 @@ namespace math {
...
  * @return z[i] where z is conversion from u.
- * @throw std::domain_error if u is zero length.
+ * @throw std::domain_error if u does not contain at least 2 elements.
  * @throw std::out_of_range if i is out of range.
  */
 inline int csr_u_to_z(const std::vector<int>& u, int i) {
-  check_positive("csr_u_to_z", "u.size()", u.size());
-  check_range("csr_u_to_z", "i", u.size(), i + 1, "index out of range");
+  check_greater("csr_u_to_z", "u.size()", u.size(), 1);
+  check_range("csr_u_to_z", "i", u.size() - 1, i + 1, "index out of range");
```
bbbales2, line 27: I'm having trouble understanding why `u.size()` is wrong. I'm not very familiar with sparse matrix formats so I'll write out my logic. Tell me where I'm wrong.

If there are `M` rows, there are `M + 1` elements of `u.size()`.

Since `u` marks the beginnings of each of the rows (with a one-past-the-end index in the last element), this function appears to, for a given row `i`, compute the number of elements in that row.

Since `i` is a row number, `i < M`, and so `i + 1 < M + 1` and so as long as `i` is a row number no accesses are out of range.

The check as coded here is `i + 1 < M`. I think it should be `i + 1 < M + 1` or `i < M`.

bbbales2:
Yo @SteveBronder, do you know the answer to the question here: https://github.com/stan-dev/math/pull/1798#discussion_r449184254? I see you have a pull request open with something about `u` and `z`.

stan/math/prim/fun/csr_u_to_z.hpp
SteveBronder: Srry it took me a sec to see this

> I'm having trouble understanding why u.size() is wrong. I'm not very familiar with sparse matrix formats so I'll write out my logic. Tell me where I'm wrong.

idt it's sparse matrices. The accessible range of a container is `[0, ..., N - 1]` since it's zero indexed. Then it's `i+1` because we access that `u[i + 1]` right?
```
 check_range("csr_u_to_z", "i", u.size() - 1, i + 1, "index out of range");
  return u[i + 1] - u[i];
```

bbbales2: Well the `i + 1` didn't change. The `u.size()` -> `u.size() - 1` changed.

SteveBronder: Wouldn't x[x.size()] be a bug for any indexing? This sounds like a bug fix more than anything

bbbales2: This check:

```
check_range("csr_u_to_z", "i", u.size() - 1, i + 1, "index out of range");
```

is checking if `i + 1 < u.size() - 1`.

I think `i` is always less than `u.size()` from the calling code. I don't know what `z` is supposed to be though which is why I'm having trouble debugging this code and I saw you had this pull open: https://github.com/stan-dev/math/pull/1844

Do you know what `z` is? This code is from the sparse matrix stuff.

bbbales2: > This sounds like a bug fix more than anything

Yeah as part of this pull I think this is just a bug fix. But there isn't an issue in the pull description.

bbbales2: Anyway the rest of the pull is great so I wanted to figure out what this is cause I assume it's probably here for a reason and shouldn't be tossed.

SteveBronder: idt we should not check this because the places we currently call it from do checks. That could change! `operator[]` for std vectors doesn't do a bounds check to verify `i + 1` is valid so I think it's reasonable to check that `i + 1` is within the indexing range of `[0, ..., N - 1]`

```cpp
inline int csr_u_to_z(const std::vector<int>& u, int i) {
  check_greater("csr_u_to_z", "u.size()", u.size(), 1);
  // check_range("csr_u_to_z", "i", u.size() - 1, i + 1, "index out of range");
  // below would grab out of bounds memory if (i + 1) > (u.size() - 1)
  return u[i + 1] - u[i];
}
```

> I don't know what z is supposed to be though which is why I'm having trouble debugging this code and I saw you had this pull open: #1844

I'm not following what in #1844 is related to this.

> Do you know what z is? This code is from the sparse matrix stuff.

I've never used this code but if I had to guess it looks like it's grabbing part of the row index for the next nonzero element. This is probably used in `rstanarm` so @bgoodri may be better to ask about this

bbbales2: > I'm not following what in #1844 is related to this

Oh, the title of that pull is generalize u to z. This is csr_u_to_z. I didn't actually look through the source in that pull. These are unrelated and these are different us and zs I guess.

@bgoodri help

SteveBronder: lol icic. The code here looks reasonable to me tho'. What situation do you expect this to fail (or erroneously fail)?

bbbales2: I think u is of length M + 1 so the right check is `i + 1 < M + 1` unless there is something about `u` and `z` I don't understand (likely).

From our sparse matrix * vector docs:
```
 *    - u: index of where each row starts in w,  length
 *      is equal to the number of rows plus one.  Last entry is
 *      one-past-the-end in w, following the Eigen spec.
```

I just don't know what `z` is or what it's trying to do lol.

bgoodri: >  This is probably used in `rstanarm` so @bgoodri may be better to ask about this

We don't actually use any of the `csr_*` functions. We had to write our own to avoid doing all these checks on sparse matrices that never change.

bbbales2: Hmm, I guess I'll just take another look at this and if I can't figure it out I'll just put the u_to_z stuff in another pull.

SteveBronder: I talked to Ben and I think it's fine to just change the range check to be `u.size()`. I forgot that the check is a less than and not a less than or equal to.

bbbales2: I went digging with some `std::cout` s. Looks like `stan::error_index::value` is 1 by default, and so by default `check_range` is  treating things as if they are one-indexed. So I think this change is right.

SteveBronder: Indeed I approve

bbbales2:
@peterwicksstringfield I pulled out the release notes and put them in the pull request and merged develop into your branch.

@syclik can you give my changes a once over? I guess I'm not supposed to merge this now that I've touched it.

bbbales2:
@serban-nicusor-toptal Did the Windows test computer go down: https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/downstream_tests/1745/pipeline ?

serban-nicusor-toptal:
Hey @bbbales2, one of the windows machines was unstable. I've turned it off for the moment and spin up a temporary one to not block the builds.
I've restarted this build, it should complete just fine now, Sorry for the inconvenience!

bbbales2:
This is good, but someone should check my changes ([this](https://github.com/stan-dev/math/pull/1798/commits/e5a7fa8928c8b87046845d5cfda87090ee76e743) and [this](https://github.com/stan-dev/math/pull/1798/commits/7d8bb9df96dc997ffd016b692ed157c3cb77afcf)) before doing the final merge.

This will also conflict with the variadic ODEs. Both make some changes to `check_finite`. The changes here are better and we should have them.

syclik:
@bbbales2, I'm assuming the release notes are correct. The changes to the test look fine.

PR #1799, opened 2020-03-24
Be careful not to produce autodiff nodes while checking for errors.

### LINKED ISSUE #1779: change check behavior to only use double types
## Description

`stan/math/prim/err/check_simplex.hpp` is doing its checks on the autodiff type that it's supplied.  

Here's the current code for a vector of scalar type `T_prob`.

```
  check_nonzero_size(function, name, theta);
  if (!(fabs(1.0 - theta.sum()) <= CONSTRAINT_TOLERANCE)) {
    std::stringstream msg;
    T_prob sum = theta.sum();
    msg << "is not a valid simplex.";
    msg.precision(10);
    msg << " sum(" << name << ") = " << sum << ", but should be ";
    std::string msg_str(msg.str());
    throw_domain_error(function, name, 1.0, msg_str.c_str());
  }
```

That first sum and subtraction from 1 is being done with autodiff types.  

`check_cholesky_factor_corr` is just as bad.  This habit probably started because the first ones like positive and negative didn't require any intermediate expressions.  

Someone needs to go through all of these things and make sure that the values are extracted and that we don't factor an autodiff variable for a covariance matrix test.

This one's actually pretty important for performance since it hits every parameter defined with a constrained type other than the simple lower and upper bounds and orderings:  unit vectors, simplexes, correlation cholesky factors, and covariance and correlation matrices.

It'd be good to spot check the other ones so there aren't anything like this `fabs(1 - x)` thing going on anywhere, because that'll also drop things on the stack.

#### Current Version:
v3.1.0

peterwicksstringfield:
## Summary

Be careful not to create autodiff nodes inside check functions. Closes #1779.

## Tests

No new tests.

## Side Effects

Shoudn't be anyway.

## Release notes

Reduce wasteful creation of vars.

## Checklist

- [X] Math issue #1779


stan/math/prim/err/check_symmetric.hpp
```diff
@@ -42,7 +42,7 @@ inline void check_symmetric(
   }
   for (size_type m = 0; m < k; ++m) {
     for (size_type n = m + 1; n < k; ++n) {
-      if (!(fabs(value_of(y(m, n)) - value_of(y(n, m)))
+      if (!(fabs(value_of_rec(y(m, n)) - value_of_rec(y(n, m)))
```
peterwicksstringfield: Note: This calls value_of_rec once for each entry in the matrix, not twice. y(m, n) runs through the upper triangle and y(n, m) through the lower.

Issue #1787, opened 2020-03-20
autodiff for matrix add and subtract


bob-carpenter:
## Description

There are lots of functions that don't have reverse mode specializations.  

#### Matrix operations

All of these would be very useful, espcially the basic arithmetic of the first five and `inverse_spd` because it's so important.

* [ ] `add`
* [ ] `subtract`
* [ ] `minus`

* [ ] `elt_multiply`
* [ ] `elt_divide`

* [ ] `diag_post_multiply`
* [ ] `diag_pre_multiply`
* [ ] `quad_form_diag`

* [ ] `chol2inv`
* [ ] `inverse_spd

All of the constraints would be good to specialize.  These get used everywhere (the least important are `corr_constrain` and `cov_matrix_constrain`---the cholesky forms are more important for our recommended usages).

* [ ] `lb_constrain`
* [ ] `ub_constrain`
* [ ] `lub_constrain`
* [ ] `corr_constrain`
* [ ] `cov_matrix_constrain`
* [ ] `cholesky_factor_constrain`
* [ ] `cholesky_corr_constrain`
* [ ] `offset_multiplier_constrain`

Then there are some reductions:

* [ ] `mean`
* [ ] `sd`
* [ ] `variance`

There are probably more.  

Feel free to solve one or more of these in a PR.

#### Current Version:
v3.1.0

betanalpha:
While specializing it would also be useful to consider direct adjoint-Jacobian transpose
implementations which can provide significant memory savings for matrix operations.

wds15:
Sounds like there is a lot of performance on the table to grab here! Looking forward to those benchmarks.

bob-carpenter:
Definitely direct adjoint-Jacobian implementations are the way to go to avoid all the virtual function calls and collapse the autodiff graph.

dpsimpson:
If no one minds, I'll have a shot at some of these over the weekend.

bob-carpenter:
@dpsimpson:  All yours.

Issue #1805, opened 2020-03-28
How to add static matrix?


SteveBronder:
## Description

@bob-carpenter has a nice [discourse thread](https://discourse.mc-stan.org/t/efficient-static-matrix-type/2136) discussing how a `static_matrix` type would work in the stan language. This idea has come up again now for efficient autodiff with Eigen #1785 and I think is the path forward to use the OpenCL backend for autodiff #1639 

My main question is how we would go about implementing this. Should we make a stan `Matrix` type that for doubles just inherits from Eigen and for var is two Eigen matrices? Something like

```cpp
// pseudocode
class Matrix<double> : public Eigen::Matrix<double, -1, -1> {
// stuff
}
```

```cpp
class Matrix<var> {
  Eigen::Matrix<double, -1, -1> val_;
  Eigen::Matrix<double, -1, -1> adj_;
// stuff
}
```

One other fun idea with this is to add an `ExecutionPolicy` template argument sort of like in [std parallel](https://en.cppreference.com/w/cpp/algorithm/transform). This would be an enum with values like `sequential`, `parallel_opencl`, `parallel_tbb` etc. and could allow for pretty simple swaps of execution backends.

I'm open to any suggestions!

#### Current Version:
v3.1.0

t4c1:
I think the first suggestion could be simplified a bit by writing a specialization for `Eigen::Matrix<var>`. Also we can make non-static version with second member `Eigen::Matrix<vari*, -1, -1> adj_;`. This is not as efficient as what you suggested, but can be moddified.

 The second suggestion (`ExecutionPolicy`) seems very complex.

bob-carpenter:
Why not just something like the following, which would work for matrices, vectors, and row vectors?

```
template <typename T, int R, int C>
class static_matrix {
  Eigen::Matrix<T, R, C> val_;
  Eigen::Matrix<T, R, C> adj_;
};
```

For the dynamic case, both members need to be `Matrix<vari*, ...>` because both would change under an assignment.

SteveBronder:
> I think the first suggestion could be simplified a bit by writing a specialization for Eigen::Matrix<var>. Also we can make non-static version with second member Eigen::Matrix<vari*, -1, -1> adj_;. This is not as efficient as what you suggested, but can be moddified.

I'm not sure what you have in mind here can you elaborate?

> The second suggestion (ExecutionPolicy) seems very complex.

Maybe I need to think about it

[...]

Issue #1300, opened 2019-08-05
Move up to C++17


SteveBronder:
## Description

Jw, what are the blockers that stop us from moving to C++17?

There's a couple features that I think would be nice. From [here](https://en.wikipedia.org/wiki/C%2B%2B17#New_features)

- Nested namespace definitions, e.g., `namespace X::Y { … }` instead of `namespace X { namespace Y { … } }`
- Fold expressions, for variadic templates (ala `std::is_arithmetic<Args> &&...`)
- A compile-time static if with the form `if constexpr(expression)` (No more `#ifdefs` for the OpenCL code!)
- copy-initialization and direct-initialization of objects of type `T` from prvalue expressions of type `T` (ignoring top-level cv-qualifiers) shall result in no copy or move constructors from the prvalue expression.
- Additional mathematical special functions, including elliptic integrals and Bessel functions [see here](https://en.cppreference.com/w/cpp/numeric)

From that same wiki page it looks like we would have to bump up the supported compiler versions?

> GCC since version 7 has complete support for C++17. [GCC C++ standards status](http://gcc.gnu.org/projects/cxx-status.html#cxx17)
> Clang 5 and later implement all the features of C++17. [clang C++ standards status](https://clang.llvm.org/cxx_status.html)

#### Current Version:
v2.20.0

SteveBronder:
Another reason is better support for [devirtualization](https://marcofoco.com/the-power-of-devirtualization/)

I can run a branch to check the performance, but I think we can put a `final` over the `chain` methods

andrjohns:
The main thing at the moment would probably Windows, since that's restricted to the gcc version packaged with RTools, which is only 4.9.3 at the moment. Having said that, the next version of RTools will apparently ship with gcc 8 (https://cran.r-project.org/bin/windows/testing/rtools40.html)

alashworth:
Devirtualization should be pretty well supported in C++11; there's nothing stopping you from adding `final` in 4.9.3, AFAIK.

syclik:
> Jw, what are the blockers that stop us from moving to C++17?

This is really a question about what are the blockers from requiring a higher version C++ compiler. Something to note: different versions of compilers support incomplete parts of the spec. For example, see [C++ Standards Support in GCC](https://gcc.gnu.org/projects/cxx-status.html).

Although, I think technically we (the developers of the Math library) could pick a version that's different than what's currently supported across Stan, we shouldn't do that lightly. Here are some of the things we care about:

[...]

PR #1812, opened 2020-03-31
Reverse mode specializaiton for matrix add and subtract

dpsimpson:
## Summary

This is addresses part of #1787 by adding reverse mode specializations for the matrix `add` and `subtract` functions. This should reduce the size of the autodiff stack.

The implementation uses a lot of tricks copped from the specialization of `multiply`.

`subtract` uses a few constexpr ifs to avoid re-writing code. I'm getting a C++17 warning, and I can change them to ordinary ifs if it's easier.  They're there so I didn't have to implement different vari extensions for `A - c` and `c - A` (where `A` is a matrix and `c` is a scalar). I had honestly hoped I wouldn't need to write almost equivalent code for `subtract(Scalar,Matrix)` and `subtract(Matrix, Scalar)` (lines 549-600 of `rev/fun/subtract.hpp`) but I couldn't work out how to do that.

As always, any suggestions/substitutions/changes greatly appreciated.

## Tests
The following tests are implemented for both `add` and `subtract`:
- (double, double), (double, var), (var, double), (var, var) using random matrices.
- Bound checking tests

I'm happy to add more, but I couldn't think of any.
## Side Effects

This should be a plug in replacement.

## Release notes

## Checklist

- [x] Math issue #1787 (Partial)

- [x] Copyright holder: Daniel simpson

### DIFF
--- a/stan/math/prim/fun/add.hpp
+++ b/stan/math/prim/fun/add.hpp
@@ -23,8 +23,8 @@ namespace math {
  * dimensions.
  */
 template <typename Mat1, typename Mat2,
-        require_all_eigen_t<Mat1, Mat2>* = nullptr,
-        require_all_not_eigen_vt<is_var, Mat1, Mat2>* = nullptr>
+          require_all_eigen_t<Mat1, Mat2>* = nullptr,
+          require_all_not_eigen_vt<is_var, Mat1, Mat2>* = nullptr>
 inline auto add(const Mat1& m1, const Mat2& m2) {
   check_matching_dims("add", "m1", m1, "m2", m2);
   return (m1 + m2).eval();
@@ -39,8 +39,7 @@ inline auto add(const Mat1& m1, const Mat2& m2) {
  * @param c Scalar.
  * @return The matrix plus the scalar.
  */
-template <typename Mat, typename Scal,
-          require_eigen_t<Mat>* = nullptr,
+template <typename Mat, typename Scal, require_eigen_t<Mat>* = nullptr,
           require_stan_scalar_t<Scal>* = nullptr,
           require_not_var_t<Scal>* = nullptr,
           require_not_eigen_vt<is_var, Mat>* = nullptr>
@@ -57,11 +56,10 @@ inline auto add(const Mat& m, const Scal c) {
  * @param m Matrix.
  * @return The scalar plus the matrix.
  */
- template <typename Mat, typename Scal,
-           require_eigen_t<Mat>* = nullptr,
-           require_stan_scalar_t<Scal>* = nullptr,
-           require_not_var_t<Scal>* = nullptr,
-           require_not_eigen_vt<is_var, Mat>* = nullptr>
+template <typename Mat, typename Scal, require_eigen_t<Mat>* = nullptr,
+          require_stan_scalar_t<Scal>* = nullptr,
+          require_not_var_t<Scal>* = nullptr,
+          require_not_eigen_vt<is_var, Mat>* = nullptr>
 inline auto add(const Scal c, const Mat& m) {
   return (c + m.array()).matrix().eval();
 }
--- a/stan/math/prim/fun/add.hpp
+++ b/stan/math/prim/fun/add.hpp
@@ -4,6 +4,7 @@
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
+#include <type_traits>
 
 namespace stan {
 namespace math {
@@ -22,7 +23,8 @@ namespace math {
  * dimensions.
  */
 template <typename Mat1, typename Mat2,
-          typename = require_all_eigen_t<Mat1, Mat2>>
+        require_all_eigen_t<Mat1, Mat2>* = nullptr,
+        require_all_not_eigen_vt<is_var, Mat1, Mat2>* = nullptr>
 inline auto add(const Mat1& m1, const Mat2& m2) {
   check_matching_dims("add", "m1", m1, "m2", m2);
   return (m1 + m2).eval();
@@ -37,8 +39,11 @@ inline auto add(const Mat1& m1, const Mat2& m2) {
  * @param c Scalar.
  * @return The matrix plus the scalar.
  */
-template <typename Mat, typename Scal, typename = require_eigen_t<Mat>,
-          typename = require_stan_scalar_t<Scal>>
+template <typename Mat, typename Scal,
+          require_eigen_t<Mat>* = nullptr,
+          require_stan_scalar_t<Scal>* = nullptr,
+          require_not_var_t<Scal>* = nullptr,
+          require_not_eigen_vt<is_var, Mat>* = nullptr>
 inline auto add(const Mat& m, const Scal c) {
   return (m.array() + c).matrix().eval();
 }
@@ -52,8 +57,11 @@ inline auto add(const Mat& m, const Scal c) {
  * @param m Matrix.
  * @return The scalar plus the matrix.
  */
-template <typename Scal, typename Mat, typename = require_stan_scalar_t<Scal>,
-          typename = require_eigen_t<Mat>>
+ template <typename Mat, typename Scal,
+           require_eigen_t<Mat>* = nullptr,
+           require_stan_scalar_t<Scal>* = nullptr,
+           require_not_var_t<Scal>* = nullptr,
+           require_not_eigen_vt<is_var, Mat>* = nullptr>
 inline auto add(const Scal c, const Mat& m) {
   return (c + m.array()).matrix().eval();
 }
--- a/stan/math/rev/fun/add.hpp
+++ b/stan/math/rev/fun/add.hpp
@@ -26,19 +26,19 @@ namespace math {
  * @tparam C number of columns (common to A and B)
  */
 
- template <typename Ta, typename Tb, int R, int C>
- class add_mat_vari : public vari {
+template <typename Ta, typename Tb, int R, int C>
+class add_mat_vari : public vari {
  public:
-   int rows_;
-   int cols_;
-   int size_;
-   double* Ad_;
-   double* Bd_;
-   vari** variRefA_;
-   vari** variRefB_;
-   vari** variRefAplusB_;
+  int rows_;
+  int cols_;
+  int size_;
+  double* Ad_;
+  double* Bd_;
+  vari** variRefA_;
+  vari** variRefB_;
+  vari** variRefAplusB_;
 
-   /**
+  /**
    * Constructor for add_mat_vari.
    *
    * All memeory is allocated in ChainableStack's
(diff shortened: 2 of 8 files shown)


bob-carpenter:
Thanks for writing tests, but they're redundant given the testing framework tests which cover all levels of autodiff against values and finite differences of the `double` implementation.  For these functions, those are in:

```
test/unit/math/mix/fun/multiply_test.cpp
test/unit/math/mix/fun/add_test.cpp
```

The one thing that should be done is to add throw tests for mismatched sizes in `multiply_test.cpp` like the ones for `add_test.cpp`.

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+ *
+ * @param[in] A matrix
+ * @param[in] B matrix
+ * @return Sum of two matrices
+ */
+template <typename Mat1, typename Mat2,
+          require_all_eigen_t<Mat1, Mat2>* = nullptr,
+          require_any_eigen_vt<is_var, Mat1, Mat2>* = nullptr>
+auto add(const Mat1& A, const Mat2& B) {
+  using Ta = value_type_t<Mat1>;
```
andrjohns, line 250: Given that these have generalised input (i.e. can take Eigen expressions), the inputs should be passed to ```Eigen::Ref``` first, and the ```Ref``` object used throughout:

```
const Eigen::Ref<const plain_type_t<Mat1>>& A_Ref = A;
```
And then ```A_Ref``` used in the rest of the function. Otherwise, if an Eigen expression is passed as an input, it could evaluated multiple times

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+  using Tb = value_type_t<Mat2>;
+  constexpr int R = Mat1::RowsAtCompileTime;
+  constexpr int C = Mat1::ColsAtCompileTime;
+  check_matching_dims("add", "A", A, "B", B);
+  check_not_nan("add", "A", A);
+  check_not_nan("add", "B", B);
+
+  // Manage memory in arena allocator
+  add_mat_vari<Ta, Tb, R, C>* baseVari = new add_mat_vari<Ta, Tb, R, C>(A, B);
+  Eigen::Matrix<var, R, C> AplusB_v(A.rows(), A.cols());
```
andrjohns, line 260: I'm realising that this is a broader issue with the generalised Eigen rev specialisations, that they can take any kind of input, but always return a ```Matrix<var, R, C>``` as the output.

@t4c1 What are your thoughts here?

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+
+/**
+ * This is a subclass of the vari class for matrix
+ * addition A + c, where A is a N by M matrix and
+ * c is a scalar.
+ *
+ * The class stores the structure of each matrix,
+ * the double values of A and c, and pointers to
+ * the varis for A and c if A or c is a var. It
+ * also instatiates and stores pointers to varis
```
andrjohns, line 275:
```suggestion
 * also instantiates and stores pointers to varis
```

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+    using Eigen::Map;
+    Map<matrix_vi>(variRefA_, rows_, cols_) = A.vi();
+    Map<matrix_d> Ad(Ad_, rows_, cols_);
+    double cd = c.val();
+    Ad = A.val();
+    Map<matrix_vi>(variRefAplusc_, rows_, cols_)
+        = (Ad.array() + cd).matrix().eval().unaryExpr([](double x) {
+            return new vari(x, false);
+          });
+  }
```
andrjohns, line 333:
```suggestion
    Map<matrix_d> Ad(Ad_, rows_, cols_);
    Ad = A.val();
    Map<matrix_vi>(variRefAplusc_, rows_, cols_).array()
        = (Ad.array() + cd_).unaryExpr([](double x) {
            return new vari(x, false);
          });
  }
```

This cuts out the unnecessary ```cd``` temporary, and the creation of the temporary matrix caused by ```.matrix().eval()```

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+
+/**
+ * This is a subclass of the vari class for matrix
+ * addition A + c, where A is a N by M matrix and
+ * c is a double.
+ *
+ * The class stores the structure of each matrix,
+ * the double values of A and c, and pointers to
+ * the varis for A and c if A or c is a var. It
+ * also instatiates and stores pointers to varis
```
andrjohns, line 352:
```suggestion
 * also instantiates and stores pointers to varis
```

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+ * for all elements of A + c.
+ *
+ * @tparam Ta type of elements in matrix A
+ * @tparam Tc type of elements in matrix B
```
andrjohns, line 356: This template parameter isn't used

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+ * @tparam R number of rows (common to A and B)
```
andrjohns, line 357: This should just be the number of rows in A

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+            ChainableStack::instance_->memalloc_.alloc_array<vari*>(size_)) {
+    using Eigen::Map;
+    Map<matrix_vi>(variRefA_, rows_, cols_) = A.vi();
+    Map<matrix_d> Ad(Ad_, rows_, cols_);
+    double cd = c;
+    Ad = A.val();
+    Map<matrix_vi>(variRefAplusc_, rows_, cols_)
+        = (Ad.array() + cd).matrix().eval().unaryExpr([](double x) {
+            return new vari(x, false);
+          });
```
andrjohns, line 407: Same suggested changes as the above

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+  }
+
+  virtual void chain() {
+    using Eigen::Map;
+    matrix_d adjAplusc(rows_, cols_);
+    adjAplusc = Map<matrix_vi>(variRefAplusc_, rows_, cols_).adj();
+    Map<matrix_vi>(variRefA_, rows_, cols_).adj() += adjAplusc;
```
andrjohns, line 414:
```suggestion
    Map<matrix_vi> Aplusc(variRefAplusc_, rows_, cols_);
    Map<matrix_vi>(variRefA_, rows_, cols_).adj() += Aplusc.adj();
```

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+ * The class stores the structure of each matrix,
+ * the double values of A and c, and pointers to
+ * the varis for A and c if A or c is a var. It
+ * also instatiates and stores pointers to varis
+ * for all elements of A + c.
+ *
+ * @tparam Ta type of elements in matrix A
+ * @tparam Tc type of elements in matrix B
+ * @tparam R number of rows (common to A and B)
+ * @tparam C number of columns (common to A and B)
```
andrjohns, line 432: Need to update this doc for consistency with template parameters used

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+        = (Ad.array() + cd).matrix().unaryExpr([](double x) {
+            return new vari(x, false);
+          });
+  }
+
+  virtual void chain() {
+    using Eigen::Map;
+    matrix_d adjAplusc(rows_, cols_);
+    adjAplusc = Map<matrix_vi>(variRefAplusc_, rows_, cols_).adj();
+    variRefc_->adj_ += adjAplusc.sum();
```
andrjohns, line 487:
```suggestion
    variRefc_->adj_ += Map<matrix_vi>(variRefAplusc_, rows_, cols_).adj().sum();
```

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+ *
+ * @tparam Mat type of the matrix or expression
+ * @tparam Scal type of the scalar
+ * @param m Matrix or expression.
+ * @param c Scalar.
+ * @return The matrix plus the scalar.
+ */
+template <typename Mat, typename Scal, require_eigen_t<Mat>* = nullptr,
+          require_stan_scalar_t<Scal>* = nullptr,
+          require_any_var_t<typename value_type<Mat>::type, Scal>* = nullptr>
```
andrjohns, line 502:
```suggestion
template <typename Mat, typename Scal, typename Ta = value_type_t<Mat>,
          require_eigen_t<Mat>* = nullptr,
          require_stan_scalar_t<Scal>* = nullptr,
          require_any_var_t<Ta, Scal>* = nullptr>
```

Then you don't need the additional typedef for ```Ta``` below

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+auto add(const Mat& A, const Scal c) {
+  using Ta = value_type_t<Mat>;
+  constexpr int R = Mat::RowsAtCompileTime;
+  constexpr int C = Mat::ColsAtCompileTime;
+  check_not_nan("add matrix-scalar A + c", "A", A);
```
andrjohns, line 507: For consistency with other functions this should just be ```check_not_nan("add", "A", A);```. Possibly less informative, but more consistent, so I'm not sure which should be chosen

andrjohns:
I'll take the review for this one (if nobody else had started)

stan/math/rev/fun/subtract.hpp
```diff
@@ -0,0 +1,591 @@
+#ifndef STAN_MATH_REV_FUN_MATRIX_SUBTRACT_HPP
+#define STAN_MATH_REV_FUN_MATRIX_SUBTRACT_HPP
+
+#include <stan/math/rev/meta.hpp>
+#include <stan/math/rev/core.hpp>
+#include <stan/math/prim.hpp>
+#include <type_traits>
```
andrjohns, line 7: Don't need this header

stan/math/rev/fun/subtract.hpp
```diff
@@ -0,0 +1,591 @@
...
+namespace math {
+
+/**
+ * This is a subclass of the vari class for matrix
+ * subtraction A - B, whre A and B are N by M matrices.
+ *
+ * The class stores the structure of each matrix,
+ * the double values of A and B, and pointers to
+ * the varis for A and B if A or B is a var. It
+ * also instatiates and stores pointers to varis
```
andrjohns, line 19:
```suggestion
 * also instantiates and stores pointers to varis
```

stan/math/rev/fun/subtract.hpp
```diff
@@ -0,0 +1,591 @@
...
+    Bd = B.val();
+    Map<matrix_vi>(variRefAminusB_, rows_, cols_)
+        = (Ad - Bd).unaryExpr([](double x) { return new vari(x, false); });
+  }
+
+  virtual void chain() {
+    using Eigen::Map;
+    matrix_d adjAminusB(rows_, cols_);
+    adjAminusB = Map<matrix_vi>(variRefAminusB_, rows_, cols_).adj();
+    Map<matrix_vi>(variRefB_, rows_, cols_).adj() -= adjAminusB;
```
andrjohns, line 159:
```suggestion
    Map<matrix_vi> AminusB(variRefAminusB_, rows_, cols_);
    Map<matrix_vi>(variRefB_, rows_, cols_).adj() -= AminusB.adj();
```

stan/math/rev/fun/subtract.hpp
```diff
@@ -0,0 +1,591 @@
...
+            ChainableStack::instance_->memalloc_.alloc_array<vari*>(size_)),
+        variRefc_(c.vi_),
+        variRefAminusc_(
+            ChainableStack::instance_->memalloc_.alloc_array<vari*>(size_)) {
+    using Eigen::Map;
+    Map<matrix_vi>(variRefA_, rows_, cols_) = A.vi();
+    Map<matrix_d> Ad(Ad_, rows_, cols_);
+    double cd = c.val();
+    Ad = A.val();
+    if constexpr (MatMinusScal) {
```
andrjohns, line 331: ```if constexpr``` is c++17, and we're only up to c++14

stan/math/rev/fun/subtract.hpp
```diff
@@ -0,0 +1,591 @@
...
+  return Aminusc_v;
+}
+
+/**
+ * Return the difference c - A of a scalar c and a matrix A.
+ *
+ * @tparam Scal type of the scalar
+ * @tparam Mat type of the matrix or expression
+ * @param c Scalar.
+ * @param m Matrix.
```
andrjohns, line 566:
```suggestion
 * @param A Matrix.
```

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,538 @@
...
+            ChainableStack::instance_->memalloc_.alloc_array<vari*>(size_)) {
+    using Eigen::Map;
+
+    double cd = c.val();
+    Map<matrix_d> Ad(Ad_, rows_, cols_);
+    Ad = A.val();
+    Map<matrix_vi>(variRefAplusc_, rows_, cols_)
+        = (Ad.array() + cd).matrix().unaryExpr([](double x) {
+            return new vari(x, false);
+          });
```
andrjohns, line 480: Same suggested changes as the above

stan/math/rev/fun/subtract.hpp
andrjohns: Also, see #1300 for more info

stan/math/prim/fun/add.hpp
```diff
@@ -4,6 +4,7 @@
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
+#include <type_traits>
```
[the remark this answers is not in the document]
SteveBronder: I usually think it's good to include std stuff when you use it, include what you use and all that

stan/math/rev/fun/add.hpp
t4c1: That is true. Rev (anfd fwd) expressions would be quite complicated to implement with current rev matrix being implemented as array of structs. We have some plans for struct of arrays implementations in #1805. That will allow to return two eigen expressions without doing custom Eigen expressions.

andrjohns: As far as I understand, it shouldn't be that complicated. As long as ```add_mat_vari``` is generalised to work with ```Matrix```/```Array``` inputs, the only difficulty is whether the return should be a ```Matrix``` or ```Array```, since the autodiff implementations won't be returning expressions.

We could use something like:
```
template<typename EigT, int R, int C, typename Enable = void>
struct eigen_return_type {};

template<typename EigT, int R, int C>
struct eigen_return_type<EigT, R, C, require_t<is_eigen_matrix<plain_type_t<EigT>>>> {
  using type = Eigen::Matrix<value_type_t<EigT>, R, C>;
}

template<typename EigT, int R, int C>
struct eigen_return_type<EigT, R, C, require_t<is_eigen_array<plain_type_t<EigT>>>> {
  using type = Eigen::Array<value_type_t<EigT>, R, C>;
}

template<typename EigT, int R, int C>
using eigen_return_t = typename eigen_return_type<EigT, R, C>::type;
```

Then in the function, the returned object can be declared as:
```
eigen_return_t<Mat1, R, C> AplusB_v(A.rows(), A.cols());

AplusB_v.matrix().vi()
    = Eigen::Map<matrix_vi>(&baseVari->variRefAplusB_[0], A.rows(), A.cols());
```

andrjohns: We also need to look at restricting the current templates to working with the same Eigen types, since the current templating allows for ```multiply(Matrix, Array)```, which I don't think should be allowed

andrjohns: @dpsimpson Sorry for derailing the pull here! I think you can just ignore this conversation and continue as normal. I'll open up a separate issue and make any changes after we settle on a solution.

t4c1: What you suggested solves the problem of deciding whether to return `Matrix` or `Array`. However that is not the same as returning an expresion.

andrjohns: But are these methods ever going to be able to return an expression? Don't they need to instantiate some container of ```var``` with the calculated ```vari(s)``` and then return that?

t4c1: I have to say I don't exactely have deep understanding of autodiff, but I believe with some work it is possible.

andrjohns: Given that the ```varis``` are just a contiguous section of memory, if an expression is returned it has to be done to a particular 'shape' and plain type.

It could be something like:
```
Eigen::Map<matrix_vi> VariRef(&baseVari->variRefAplusB_[0], A.rows(), A.cols());

return VariRef.unaryExpr([](vari** x) { return var(x); });
```

But that still requires deducing whether the 'base type' should be ```Matrix``` or ```Array```

stan/math/prim/fun/add.hpp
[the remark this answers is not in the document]
andrjohns: Sounds good to me! I'll drop this comment

stan/math/rev/fun/add.hpp
t4c1: That should work. But I think it would not be any faster than current implementation. I think we can do better. I don't have exact plan how yet. I will start thinking about this after I finish with generalizations of functions.

andrjohns: Given that the benefits of returning expressions are that only the needed parts of a matrix/array get evaluated, I don't know if we'll be able to achieve that since the actual calculations for the values and adjoints are done on the entirety of the input matrices and the results copied into ```varis```. I don't think there's way to propogate expressions that only evaluate the needed part of the matrix into ```vari``` and then return those as ```var```, since the calculations with ```varis``` can't return expressions.

stan/math/rev/fun/subtract.hpp
SteveBronder: One of these days!!!

bbbales2:
@dpsimpson How did this compare in the end to the un-autodiffed version?

dpsimpson:
Sorry - got caught up in a work thing. Will do stuff tonight and tomorrow.

wds15:
This will be in 2.24? Or can this move in today?

SteveBronder:
I think this would be a new feature so it will be in the next release

wds15:
Of course...this is new stuff. The pr comments sounded as if this is about ready. I expect this pr to speed up things quite a bit which is why I was bumping this for some attention.

But let’s wait for the next release.

dpsimpson:
Definitely next release - swamped with exam stuff. Will pick it up ASAP.
Sorry.

SteveBronder:
Closing this for now but feel free to open when you have time to fixup!

spinkney:
@dpsimpson any chance this can get resurrected?

dpsimpson:
I thought it was superseded. But absolutely! I’ll have a look over the
weekend.

Issue #58, opened 2015-07-06
Interpolation methods in Stan?


syclik:
_From @edbaskerville on December 9, 2014 23:33_

It would be nice if Stan had some simple one-dimensional interpolation methods built in. (Not super-important, since this should be doable within a Stan model.)

One place this issue arises when implementing ODE models driven by empirical data sampled at discrete time intervals. Because the time derivative needs to be evaluated at arbitrary time points, the values of the driving variables need to be evaluated between sample times.

Even linear interpolation would be sufficient for many cases: e.g.,

```
data {
  int nTimepoints;
  real[nTimepoints] x;
}
// ...
x_t <- interpolate_linear(x, t);
```

would result in `x_t` equal to a value linearly interpolated from `x[i]`, `x[i + 1]`, where `i <= t <= i + 1`.

_Copied from original issue: stan-dev/stan#1165_

syclik:
_From @bob-carpenter on December 10, 2014 18:12_

Related to #731 (in a good way).

syclik:
_From @bob-carpenter on December 10, 2014 18:13_

We don't really need that in the diff eq solver.  It already uses interpolation internally, with the added bonus of error bounding (in case linear interpolation isn't accurate enough).

But there are certainly other uses.

syclik:
_From @edbaskerville on December 11, 2014 0:25_

Cool, sounds like interpolation based on smooth functions is on the way, glad to hear it.

For ODEs—and perhaps I'm not following—I wasn't talking about the solver per se, but rather the use of an interpolated function inside the time derivative (making it a non-autonomous system). That is dy/dt (say, infection rate) is a function of an interpolated x(t) (which is calculated from an observed, discrete time series). E.g., `dy/dt = alpha * x(t)`, where `alpha` is a parameter and `x(t)` is an interpolated function.

In the meantime, I went ahead and implemented linear interpolation between adjacent points in an array directly in Stan. It's rather horrible, because I had to sneak around the no-real-to-int-conversion restriction by implementing an integer-floor function via bisection.

For my purposes (using an interpolated x(t) in an ODE), I think this still is OK because the partial derivatives of dy/dt with respect to the parameters should treat values of x(t) as constants (or perhaps I'm missing something?)—and the function, though not differentiable, is continuous, and thus can be integrated.

Here it is, in its mostly untested, frightening glory:

```
int intFloor(int leftStart, int rightStart, real iReal)
{
[...]

PR #1814, opened 2020-03-31
Interpolation functions

pgree:
## Summary

I added a function that does interpolation of a set of function values, `(x_i, y_i)`, specified by the user. The algorithm works by first doing a linear interpolation between points and then smoothing that linear interpolation by convolution with a Gaussian. 

In order to enforce that the interpolated function values coincide with the user-inputted function values, the algorithm at each step convolves a Gaussian with a new piecewise linear function in which the function values are shifted by the difference between the most recent interpolation and the desired, original value.

The interpolated function can be evaluated analytically aside from an evaluation of the error function. The derivative of the interpolated function can be evaluated analytically. 

This is a work-in-progress. I have discussed some of the design choices with @bbbales2 but that's it. 

## Tests

There are tests for the "accuracy" of the interpolation and tests for the derivative of the interpolation that compare autodiff to finite differences.

## Side Effects

The variance of the Gaussian in the convolution is currently 1/10 the minimum distance between the x-values specified by the user. This has the potentially to lead to undesired behavior. 

There is a tolerance hardcoded for when to stop the iterative algorithm in terms of the maximum distance between the interpolated values at the original points and the user-specified values at those points. 

## Checklist

- [x] Math issue #58


stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/*
+evaluate the derivative of conv_gaus_line with respect to x
+*/
+double der_conv_gaus_line(double t0, double t1, double a, double b, double x0,
+                          double sig2) {
```
SteveBronder, line 16: Generally we want prim functions to be templated generically to work with any type, then the specialization in `rev` will be for analytic calculations.

stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <stan/math/prim/prob/normal_cdf.hpp>
+#include <cmath>
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/*
+evaluate the derivative of conv_gaus_line with respect to x
+*/
```
SteveBronder, line 14: Need full docs for external facing functions

stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,67 @@
...
+double der_conv_gaus_line(double t0, double t1, double a, double b, double x0,
+                          double sig2) {
+  using stan::math::normal_cdf;
+  double pi = stan::math::pi();
+  using std::exp;
+  using std::pow;
+  using std::sqrt;
```
SteveBronder, line 21: Put `using` aliases together at the top of the function

stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,67 @@
...
+  double sig = sqrt(sig2);
```
SteveBronder, line 22:
```suggestion
  const double sig = sqrt(sig2);
```

stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,67 @@
...
+  double y;
+
+  double alpha = sqrt(2 * pi * sig2);
```
SteveBronder, line 25:
```suggestion
  const double alpha = sqrt(2 * pi * sig2);
```

stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/*
+evaluate the integral
+
+                    | t1
+(2*pi*sig2)**-(1/2) |   (a*t + b) * exp(-(t - x0)^2 / (2*sig2)) dt
+                    | t0
+
+*/
```
SteveBronder, line 43: full docs

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+using namespace std;
+
+namespace stan {
+namespace math {
+
+// set tolerance for how close interpolated values need to be to
+// the specified function values
+const double INTERP_TOL = 1e-8;
+const double SIG2_SCALE = 0.1;
```
SteveBronder, line 17: If this is specific to a function they should be in the function instead of floating in the stan math namespace

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+/*
+given a set of points (x_i, y_i) with x_i's in increasing order, find
+a_i, b_i such that the line between (x_i, y_i) and (x_{i+1}, y_{i+1}) is
+f(t) = a_i*t + b_i, store the a_i's and b_i's in as and bs.
+*/
+void lin_interp_coefs(int n, std::vector<double> xs, std::vector<double> ys,
+                      std::vector<double>& as, std::vector<double>& bs) {
```
SteveBronder, line 25: I'm generally against in/out parameters and I think this function's definition could just be inlined where it's used

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+/*
+linear interpolation at one point, x, given the coefficients
+*/
+double lin_interp_pt(int n, vector<double> xs, vector<double> ys,
+                     vector<double> as, vector<double> bs, double x) {
+  // find interval where x lives
+  if (x <= xs[0])
+    return ys[0];
+  if (x >= xs[n - 1])
+    return ys[n - 1];
```
SteveBronder, line 46:
```suggestion
  if (x <= xs[0]) {
    return ys[0];
  }
  if (x >= xs[n - 1]) {
    return ys[n - 1];
  }
```

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+    as[i] = (ys[i + 1] - ys[i]) / (xs[i + 1] - xs[i]);
+    bs[i] = -xs[i] * as[i] + ys[i];
+  }
+}
+
+/*
+linear interpolation at one point, x, given the coefficients
+*/
+double lin_interp_pt(int n, vector<double> xs, vector<double> ys,
+                     vector<double> as, vector<double> bs, double x) {
```
SteveBronder, line 41: These should all be const& parameters so copies are not made

SteveBronder: This is true for everywhere you are using std::vector, arithmetic types can be passed by value

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+linear interpolation at a vector of points
+*/
+vector<double> lin_interp(int n, std::vector<double> xs, std::vector<double> ys,
+                          int n_new, std::vector<double> xs_new) {
+  // compute coefficients of linear interpolation
+  vector<double> as, bs;
+  lin_interp_coefs(n, xs, ys, as, bs);
+
+  // evaluate at new points
+  vector<double> ys_new;
```
SteveBronder, line 65: Reserve the vector size ahead of the loop to avoid reallocation
```suggestion
  vector<double> ys_new;
  ys_new.reserve(n_new);
```

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  return y;
+}
+
+/*
+find the smallest difference between successive elements in a sorted vector
+*/
+template <typename Tx>
+double min_diff(int n, std::vector<Tx> xs) {
```
SteveBronder, line 111: If this is only used in this file as an internal function then it should be in the internal namespace

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,167 @@
...
+  double dmin = value_of(xs[1]) - value_of(xs[0]);
```
SteveBronder, line 112: idt this will work for `fvar<var>`

pgree:
@SteveBronder thanks for reviewing and for these comments. There is some discussion --  https://github.com/stan-dev/design-docs/pull/18 -- about how and if this should be implemented so I will wait until there is some agreement before iterating further on this.

syclik:
Just changed the title to reflect this is WIP.

SteveBronder:
@pgree are there any blockers for this?

pgree:
@SteveBronder sorry for the delay on this, I got distracted and had some Catalina-related problems. I will get to this asap

SteveBronder:
Cool no worries just wanted to check in!

stan/math/prim/fun/conv_gaus_line.hpp
pgree: moved this function to the rev version of this file

stan/math/prim/fun/gaus_interp.hpp
pgree: It has been inlined

stan/math/prim/fun/gaus_interp.hpp
pgree: I made these changes generally and also I created a new file for lin_interp. It seemed to make sense to separate the linear interpolation from this smooth interpolation.

stan/math/prim/fun/gaus_interp.hpp
pgree: I think this is no longer relevant, because now the xs (and ys) will only be inputted as data

stan/math/prim/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * @param t0 lower integration bound
+ * @param t1 upper integration bound
+ * @param a coefficient of t in line
+ * @param b constant in line
+ * @param x point at which convolution is evaluated
+ * @param sig2 variance of the Gaussian kernel
+ * @return The value of the derivative
+ */
+template <typename Tx>
+double conv_gaus_line(double t0, double t1, double a, double b, Tx const& x,
```
SteveBronder, line 26: Just an fyi we use west const

```suggestion
double conv_gaus_line(double t0, double t1, double a, double b, const Tx& x,
```

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,198 @@
...
+template <typename Tx>
+double min_diff(int n, std::vector<Tx> const& xs) {
+  double dmin = value_of(xs[1]) - value_of(xs[0]);
+  for (int i = 1; i < n - 1; i++) {
+    if (value_of(xs[i + 1]) - value_of(xs[i]) < dmin) {
+      dmin = value_of(xs[i + 1]) - value_of(xs[i]);
+    }
+  }
+  return dmin;
+}
```
SteveBronder, line 35: Do you want the smallest absolute difference (two closest number)?

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,198 @@
...
+  const double NSTDS = 10;
+  int n = xs.size();
+
+  // create copy of xs so that endpoints can be extended
+  std::vector<double> xs2 = xs;
+
+  // extend out first and last lines for convolution
+  double sig = std::sqrt(params.sig2);
+  xs2[0] = xs[0] - NSTDS * sig;
+  xs2[n - 1] = xs[n - 1] + NSTDS * sig;
```
SteveBronder, line 83: Should you also check that xs > 0 for 0 sized vectors?

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,198 @@
...
+    for (int i = 0; i < n; i++) {
+      dd = ys[i] - gaus_interp(xs, y2s, params, xs[i]);
+      y2s[i] += dd;
+      dmax = std::max(std::abs(dd), dmax);
+    }
+    if (dmax < INTERP_TOL)
+      break;
+  }
+
+  return params;
```
SteveBronder, line 164: [optional] personally I prefer constructing this at the end like

```suggestion
  return gaus_interp_params(as, bs, sig2);
```

stan/math/prim/fun/lin_interp.hpp
```diff
@@ -0,0 +1,59 @@
+#ifndef STAN_MATH_PRIM_FUN_LIN_INTERP
+#define STAN_MATH_PRIM_FUN_LIN_INTERP
+
+#include <cmath>
+#include <vector>
+#include <algorithm>
+
+using namespace std;
```
SteveBronder, line 8:
```suggestion
```

stan/math/prim/fun/gaus_interp.hpp
pgree: This will only be used on vectors with increasing elements, so the differences here should all be positive.

stan/math/prim/fun/gaus_interp.hpp
pgree: good point, I added check to make sure that xs is of length at least 2

pgree:
@SteveBronder do you have any idea why the multiple translation unit tests is failing? As far as I can tell `lin_interp` is only defined in one place.

```
test/unit/multiple_translation_units2.o: In function `stan::math::lin_interp(std::vector<double, std::allocator<double> > const&, std::vector<double, std::allocator<double> > const&, double)':


multiple_translation_units2.cpp:(.text+0x710): multiple definition of `stan::math::lin_interp(std::vector<double, std::allocator<double> > const&, std::vector<double, std::allocator<double> > const&, double)'
```

rok-cesnovar:
You need to make sure all the functions are `inline`.

pgree:
Thanks @rok-cesnovar

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+  params.sig2 = square(min_diff(n, xs) * SIG2_SCALE);
+  params.as.resize(n - 1);
+  params.bs.resize(n - 1);
+
+  // copy ys into a new vector that will be changed
+  std::vector<double> y2s = ys;
+
+  // interatively find interpolation that coincides with ys at xs
+  int max_iters = 50;
+  double dmax, dd;
```
SteveBronder, line 151: We normally seperate out declerations
```suggestion
  double dmax
  double dd;
```

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+  for (int j = 0; j < max_iters; j++) {
+    // find slope and intercept of line between each point
+    for (int i = 0; i < n - 1; i++) {
+      params.as[i] = (y2s[i + 1] - y2s[i]) / (xs[i + 1] - xs[i]);
+      params.bs[i] = -xs[i] * params.as[i] + y2s[i];
+    }
+
+    dmax = 0;
```
SteveBronder, line 159: You can define dmax here

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+    for (int i = 0; i < n; i++) {
+      dd = ys[i] - gaus_interp(xs, y2s, params, xs[i]);
+      y2s[i] += dd;
+      dmax = std::max(std::abs(dd), dmax);
+    }
+    if (dmax < INTERP_TOL)
```
SteveBronder, line 165: I would just call `INTERP_TOL` `max_tolerance` or something

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+ * points.
+ *
+ * @param xs vector of independent variable of reference points
+ * @param ys vector of dependent variable of reference points
+ * @param xs_new vector of point at which to evaluate interpolation
+ * @return vector of interpolation values
+ */
+
+template <typename Tx>
+inline std::vector<Tx> gaus_interp_vect(const std::vector<double>& xs,
```
SteveBronder, line 185: You can just call this `gauss_interp` Since it has a different signature the overloading is fine.

Also I remember you mentioned for something about how this was written to work with the old ODE syntax. You may want to check if the new ODE syntax will change your design at all

https://mc-stan.org/users/documentation/case-studies/convert_odes.html

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+ * @param ys vector of dependent variable of reference points
+ * @param params a struct created by gaus_interp_precomp that
+ * @param x the point at which to evaluate the interpolation
+ * @return value of the interpolation at x
+ */
+template <typename Tx>
+inline return_type_t<Tx> gaus_interp(const std::vector<double>& xs,
+                                     const std::vector<double>& ys,
+                                     const gaus_interp_params& params,
+                                     const Tx& x) {
```
SteveBronder, line 66: If a user calls this you'll need a version that takes in params as an `std::vector<double>` since idt we plan to expose the struct `gaus_interp_params` to the language

Also do you want this to work on eigen vectors (Stan's `vector[]` type) as well? We'd need to modify this a bit like 

```suggestion
template <typename VecX, typename VecY, typename Tx, 
  require_all_vector_like_t<VecX, VecY>* = nullptr>
inline return_type_t<VecX, VecY, Tx> gaus_interp(const VecX& xs,
                                     const VecY& ys,
                                     const gaus_interp_params& params,
                                     const Tx& x) {
```

And use some type traits stuff in stan to deal with the two vector types. I can help out with that once the rest of this is tidied up

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+  params.as.resize(n - 1);
+  params.bs.resize(n - 1);
+
+  // copy ys into a new vector that will be changed
+  std::vector<double> y2s = ys;
+
+  // interatively find interpolation that coincides with ys at xs
+  int max_iters = 50;
+  double dmax, dd;
+  for (int j = 0; j < max_iters; j++) {
```
SteveBronder, line 152: [optional] both could be just `size_t`

stan/math/prim/fun/lin_interp.hpp
```diff
@@ -0,0 +1,64 @@
...
+  if (x <= xs[0]) {
+    return ys[0];
+  }
+  if (x >= xs[n - 1]) {
+    return ys[n - 1];
+  }
+
+  // find in between which points the input, x, lives
+  auto ub = upper_bound(xs.begin(), xs.end(), x);
+  int ind = distance(xs.begin(), ub);
```
SteveBronder, line 46: call using `std::`

```suggestion
  auto ub = upper_bound(xs.begin(), xs.end(), x);
  auto ind = distance(xs.begin(), ub);
```

stan/math/prim/fun/lin_interp.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+  // check if the interpolation point falls on a reference point
+  if (x == *(ub - 1))
```
SteveBronder, line 49: Think you can just grab `xs[ind - 1]` here?

SteveBronder: Yeah that for all of these

stan/math/rev/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * @param sig2 variance of the Gaussian kernel
+ * @return The value of the derivative
+ */
+inline double der_conv_gaus_line(double t0, double t1, double a, double b,
+                                 double x0, double sig2) {
+  using stan::math::normal_cdf;
+  using std::exp;
+  using std::pow;
+  using std::sqrt;
+  const double pi = stan::math::pi();
```
SteveBronder, line 34: You should just be able to call `pi()` since we are in stan::math

stan/math/rev/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,81 @@
...
+  const double sig = sqrt(sig2);
+  const double alpha = sqrt(2 * pi * sig2);
+
+  double y = (a * x0 + b) / alpha
+             * (-exp(-pow(t1 - x0, 2) / (2 * sig2))
+                + exp(-pow(t0 - x0, 2) / (2 * sig2)));
+  y += a * (normal_cdf(t1, x0, sig) - normal_cdf(t0, x0, sig));
+  y -= a / alpha
+       * ((t1 - x0) * exp(-pow(t1 - x0, 2) / (2 * sig2))
+          - (t0 - x0) * exp(-pow(t0 - x0, 2) / (2 * sig2)));
```
SteveBronder, line 44: [very optional] Wonder if there is someway to simplify the algebra here? `exp` and `pow` are pretty expensive

stan/math/rev/fun/conv_gaus_line.hpp
```diff
@@ -0,0 +1,81 @@
...
+        t1_(t1),
+        a_(a),
+        b_(b),
+        sig2_(sig2) {}
+
+  void chain() {
+    avi_->adj_
+        += adj_ * der_conv_gaus_line(t0_, t1_, a_, b_, avi_->val_, sig2_);
+  }
+};
```
SteveBronder, line 69: This is going to be pretty expensive. For the vector `xs` you'll be essentially doubling it's size since you copy elements `i` and `i+1` for each cell. I'd check out some of our matrix rev functions to see how you can bring the whole vector here instead of in bits

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+namespace stan {
+namespace math {
+
+// struct for passing precomputation data
+struct gaus_interp_params {
+  std::vector<double> as;
+  std::vector<double> bs;
+  double sig2;
+};
```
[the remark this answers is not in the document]
pgree: I got rid of the gaus_interp_params struct altogether and instead just used a vector<double> to store the parameters. I think we discussed doing this because otherwise we would have had to create a new data type in Stan for this struct.

stan/math/prim/fun/gaus_interp.hpp
pgree: I don't think it was discussed whether this will have to work on Stan's `vector[]` types. I will tidy up the rest and then it'll be good to discuss further.

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+  // no need to convolve far from center of gaussian, so
+  // get lower and upper indexes for integration bounds
+  auto lb = upper_bound(xs.begin(), xs.end(), x - NSTDS * sig);
+  int ind_start = distance(xs.begin(), lb) - 1;
+  ind_start = std::max(0, ind_start);
+
+  auto ub = upper_bound(xs.begin(), xs.end(), x + NSTDS * sig);
+  int ind_end = distance(xs.begin(), ub);
+  ind_end = std::min(n - 1, ind_end);
```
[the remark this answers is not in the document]
pgree: I couldn't use exactly this since lower bound still searches for the first element >= the input. I added in the std::'s though

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+  // sum convolutions over intervals
+  using return_t = return_type_t<Tx>;
+  return_t y = 0;
+  for (int i = ind_start; i < ind_end; i++) {
+    y += conv_gaus_line(xs2[i], xs2[i + 1], params.as[i], params.bs[i], x,
+                        params.sig2);
+  }
+
+  return y;
```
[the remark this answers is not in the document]
pgree: As far as I can tell, I don't think there is a way to avoid the var += here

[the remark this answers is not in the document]
pgree: > would be good to just have a conv_gauss_line that takes in vectors, does the accumulation and returns back a var.

Added this

stan/math/rev/fun/conv_gaus_line.hpp
pgree: @SteveBronder Apologies, I'm not sure that I understand your comment. Only the parameter `x` in 
```
inline return_type_t<Tx> gaus_interp(const std::vector<double>& xs,
                                     const std::vector<double>& ys,
                                     const gaus_interp_params& params,
                                     const Tx& x) {
```
is going to be templated to allow for var types. Ideally, for each evaluation of the gradient, der_conv_gaus_line will be called three times maximum, and  the gradient will be evaluated with a formula. But it sounds like that is not happening currently?

SteveBronder: `der_conv_gaus_line` will be called in the reverse pass `ind_start - ind_end` times (since that's how many of these `conv_gaus_line_vari` are created) . It's fine as is but if we could make a version of this that worked on the whole vector it would save some extra calls. I can try to play around with this during the week but if I don't then you can ignore this

stan/math/prim/fun/gaus_interp.hpp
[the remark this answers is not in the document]
pgree: I ended up changing this back. Do you have a strong opinion about this? I'm not sure there's a way to do it without the code being quite a bit less clear and readable.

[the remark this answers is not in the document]
pgree: Do you have a strong opinion about this, @SteveBronder? My impression is that if we want to use the formula for the derivative, which seems to make sense to me, then making this change will not be worth it. The code will become a lot less clear and concise and won't follow the mathematical logic as nicely.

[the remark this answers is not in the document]
SteveBronder: Nah I don't it's fine as is

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+namespace stan {
+namespace math {
+
+namespace internal {
+/*
+ * find the smallest difference between successive elements in a sorted vector
+ */
+template <typename Tx>
+inline double min_diff(int n, const std::vector<Tx>& xs) {
```
SteveBronder, line 19: Document `n`

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+}  // namespace internal
+
+/**
+ * Given a set of reference points \f$(xs_i, ys_i)\f$, create a mollifier
+ * that intersects the reference points. This function requires as input
+ * a struct created by the function gaus_interp_precomp. The algorithm
+ * used to create the mollifier is an iterative algorithm that works
+ * as follows. First a linear
```
SteveBronder, line 36: Docs need updated

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+  check_greater_or_equal(function, "Interpolation point", x, xs.front());
+  check_ordered(function, "xs", xs);
+  check_not_nan(function, "xs", xs);
+  check_not_nan(function, "ys", ys);
+  check_not_nan(function, "x", x);
+  check_greater(function, "xs", xs.size(), 1);
+
+  // number of standard deviations to extend endpoints for convolution
+  const double NSTDS = 10;
+  int n = xs.size();
```
SteveBronder, line 72: [optional] we usually use small case `n` when we do an iterator in a loop. Better name would be `N` or `size_xs`

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  // extend out first and last lines for convolution
+  double sig = std::sqrt(params[2 * n - 2]);
+  xs2[0] = xs[0] - NSTDS * sig;
+  xs2[n - 1] = xs[n - 1] + NSTDS * sig;
+
+  // no need to convolve far from center of gaussian, so
+  // get lower and upper indexes for integration bounds
+  auto lb = std::lower_bound(xs.begin(), xs.end(), x - NSTDS * sig);
+  int ind_start = &(*lb) - &xs[0] - 1;
```
SteveBronder, line 87: Can you use `std::distance()` here to get the distance from the return of `lower_bound()` to the start of the vector?

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+  ind_start = std::max(0, ind_start);
```
SteveBronder, line 88: Something is wrong if you are getting a negative number here

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  auto ub = std::upper_bound(xs.begin(), xs.end(), x + NSTDS * sig);
+  int ind_end = &(*ub) - &xs[0];
+  ind_end = std::min(n - 1, ind_end);
```
SteveBronder, line 92: Same as above

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  // sum convolutions over intervals
+  return_type_t<Tx> y = 0;
+  for (int i = ind_start; i < ind_end; i++) {
+    y += conv_gaus_line(xs2[i], xs2[i + 1], params[i], params[(n - 1) + i], x,
+                        params[2 * n - 2]);
```
SteveBronder, line 98: Is this packing from stuffing the other parameters at the end?

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+  check_not_nan(function, "xs", xs);
+  check_not_nan(function, "ys", ys);
+  check_ordered(function, "xs", xs);
+  check_greater(function, "xs", xs.size(), 1);
+
+  using internal::min_diff;
+  static const double max_diff = 1e-8;
+  // set Gaussian kernel to sig2_scale times smallest difference between
+  // successive points
+  static const double sig2_scale = 0.1;
```
SteveBronder, line 132: Should these be things users can choose?

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+  int n = xs.size();
+
+  // create the vector to be returned that consists of as, bs, sig2
+  std::vector<double> params;
+  params.resize(2 * n - 1);
+  params[2 * n - 2] = square(min_diff(n, xs) * sig2_scale);
```
SteveBronder, line 138: Resizing just allocates memory but doesn't initialize so calling `params[2 * n - 2] is undefined behavior. To tack these on at the end here you would need to initialize the whole vector like `std::vector<double> params(2 * n - 1, 0);`

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  // copy ys into a new vector that will be changed
+  std::vector<double> y2s = ys;
+
+  // interatively find interpolation that coincides with ys at xs
+  int max_iters = 50;
```
SteveBronder, line 144: Should this be user defined?

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+  double dd;
+  for (int j = 0; j < max_iters; j++) {
+    // find slope and intercept of line between each point
+    for (size_t i = 0; i < n - 1; i++) {
+      params[i] = (y2s[i + 1] - y2s[i]) / (xs[i + 1] - xs[i]);
+      params[(n - 1) + i] = -xs[i] * params[i] + y2s[i];
+    }
```
SteveBronder, line 151: You have to do the thing above with initialization or this is UB

stan/math/prim/fun/gaus_interp.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+    double dmax = 0;
+    for (size_t i = 0; i < n; i++) {
+      dd = ys[i] - gaus_interp(xs, y2s, params, xs[i]);
+      y2s[i] += dd;
+      dmax = std::max(std::abs(dd), dmax);
+    }
+    if (dmax < max_diff)
+      break;
```
SteveBronder, line 160: [optional] I prefer having scopes defined even for one liners

```suggestion
    if (dmax < max_diff) {
      break;
    }
```

stan/math/prim/fun/lin_interp.hpp
```diff
@@ -0,0 +1,91 @@
...
+  if (x <= xs[0]) {
+    return ys[0];
+  }
+  if (x >= xs[n - 1]) {
+    return ys[n - 1];
+  }
+
+  // find in between which points the input, x, lives
+  auto ub = std::upper_bound(xs.begin(), xs.end(), x);
+  int ind = &(*ub) - &(xs[0]);
```
SteveBronder, line 46: same thing with distance

stan/math/rev/fun/conv_gaus_line.hpp
SteveBronder: Instead of doing `exp(-pow(t1 - x0, 2) / (2 * sig2)` multiple times would be good to pull out and do once

SteveBronder (changes requested):
Sorry for the delay, posted some Qs

stan/math/prim/fun/gaus_interp.hpp
pgree: I couldn't remember if we settled on std::distance or pointer arithmetic in our call. Will change.

stan/math/prim/fun/gaus_interp.hpp
pgree: The problem is std::lower_bound finds the first element in a vector that is larger than an inputted value, x. But what we really want is the previous element, the last element that's less than x. So we need to subtract 1. However when the first element in the list is greater than x, that would give us -1, so we need to compensate for that. I couldn't figure out a cleaner way to do it, but let me know if you have a thought on how to do it.

stan/math/prim/fun/gaus_interp.hpp
pgree: Instead of using the struct in the previous iteration of the code, I used a vector, `params`. The first n entries of `params` is one vector of parameters, the next n are another and the last entry is the width of the kernel.

stan/math/prim/fun/gaus_interp.hpp
pgree: I don't think so

rok-cesnovar:
Bump. What is the status of this PR? Waiting on a re-review or something else?

pgree:
@rok-cesnovar I think all the requested changes were made so just waiting for a re-review.

wds15:
Didn't we also have a design doc for this. Not sure what the status is there?

pgree:
> Didn't we also have a design doc for this. Not sure what the status is there?

Yes, exactly. This PR reflects what was discussed there, including your suggestions @wds15

pgree:
@bgoodri - per our discussion today, here's the PR

bgoodri:
Sorry, didn't realize this existed until today. It sounds like it is already been reviewed, in which case we should merge it once the conflict with fun.hpp is fixed. But what were the pros and cons of this interpolation scheme as compared to any of the ones in [Boost](https://www.boost.org/doc/libs/1_77_0/libs/math/doc/html/interpolation.html) that we could wrap?

pgree:
> But what were the pros and cons of this interpolation scheme as compared to any of the ones in Boost that we could wrap?

This interpolation schemes has a few of advantages
- the interpolated function is smooth
- the interpolated function can be evaluated with a formula
- the derivative of the interpolated function can be evaluated with a formula
- the interpolated function coincides with the inputted points

At the time we decided on this interpolation scheme, Boost didn't have anything with these features

bob-carpenter:
It'd be great to add an interpolation function to Stan.  Andrew and others have been asking for something like the interpolation function in BUGS for ages.  

Is the approach unlike any of the Boost implementations? https://www.boost.org/doc/libs/1_77_0/libs/math/doc/html/interpolation.html

bgoodri:
> At the time we decided on this interpolation scheme, Boost didn't have anything with these features

Fair enough, although since we will probably include `pchip` at some point, we should think about having some sort of a common interface for them, at least at the Stan Language level. It would be great it we could soon have `callable f = interpolation(x, y);` in the `transformed data` block.

pgree:
> Is the approach unlike any of the Boost implementations? https://www.boost.org/doc/libs/1_77_0/libs/math/doc/html/interpolation.html

From above: 

This interpolation scheme has a few of advantages

- the interpolated function is smooth
- the interpolated function can be evaluated with a formula
-  the derivative of the interpolated function can be evaluated with a formula
- the interpolated function coincides with the inputted points

Also, you end up with non-oscillatory behavior between reference points.  At the time we decided on this interpolation scheme, Boost didn't have anything with these features

The plots for the Akima spline look nice, but I'm not sure how they would do in the environments where people will use them. I don't know if one of the Boost interpolation schemes would work well enough for users' needs. Maybe they would. 

> we should think about having some sort of a common interface for them, at least at the Stan Language level. It would be great it we could soon have callable f = interpolation(x, y); in the transformed data block.

That's pretty much what we did. I discussed this with (I think) @SteveBronder a while back and we ended up settling on 

`lin_interp(xs, ys, x)`
`gaus_interp(xs, ys, x)`

which are both in this PR

bob-carpenter:
Thanks.  

```
lin_interp(xs, ys, x)
gaus_interp(xs, ys, x)
```

1. We've used "normal" rather than "gauss" everywhere in our math lib.  If you want to call out Gauss, it should at least be spelled out all the way to "gauss".   

2. The BUGS function is `interp.lin`.  It's documented on p. 13 of http://www.mrc-bsu.cam.ac.uk/wp-content/uploads/manual14.pdf.  The advantage of `interp_lin` over `lin_interp` is that it's better for autocomplete and indexing.  

I probably won't be reviewing this and we don't double review, but I strongly urge you not to use `gaus_` whatever other decisions are made.

pgree:
great, will change to `interp_lin`. Since `interp_normal` doesn't sound like a useful name, how about `interp_smooth`?

bob-carpenter:
I like the more descriptive name.  I'm OK with `interp_gauss` if you think `interp_normal` is too vague.  I just really don't like just `gaus`.

bgoodri:
Since all of the interpolation schemes we might implement in Stan in the future will presumably be smooth to some extent, I don't think `interp_smooth` would disambiguate. Maybe `interp_convolve`?

bob-carpenter:
If there aren't likely to be other normal or Gaussian interpolations, I'd prefer

* `interp_normal`, or
* `interp_gauss`.

If there might be other Gaussian interpolations that don't involve convolution, then I'd prefer

* `interp_normal_conv`,
* `interp_gauss_conv`,
* `interp_normal_convolve`, or
* `interp_gauss_convolve`.

syclik:
@pgree, sorry about the delay. I'm closing the PR for the moment. Please reopen if it's still active.

When that happens, I'll clear out the old review and give it a fresh one and turn that around quickly.

Right now, the tests fail, so we'll need to trigger the build again anyway.

PR #1817, opened 2020-04-02
Add optional broadcasting to kernel generator

t4c1:
## Summary
Added optional broadcasting to kernel generator. It broadcasts the argument in specified dimmensionif it can be broadcasted. That is if size in that dimension equals 1. If not nothing is broadcasted.

## Tests
Added tests for optional broadcasting.

## Side Effects
None.

## Release notes
Added optional broadcasting to kernel generator.

## Checklist

- [x] Math issue #1342 


SteveBronder:
I think I'm kind of confused on the usecase of this. Are there places where we won't know until runtime whether something should be broadcast? Feels a little dangerous. Is there an eigen equivalent to this?

t4c1:
There is no EIgen equivalent. But we are doing exactley this in existing GLM kernels for broadcasting of y, as we do not get any compile-time info on whether y is scalar or vector.

t4c1:
Alternatively we could add compile-time dimmensions to `matrix_cl`. Although I am not sure it those would be benefitial in any other case.

SteveBronder:
> Alternatively we could add compile-time dimensions to matrix_cl. Although I am not sure if those would be benefitial in any other case.

Let me think about this for a teeny bit. imo I like that idea and it lines our API up with what Eigen does anyway. We could keep everything backwards compatible by setting the compile time rows to dynamic by default

SteveBronder:
Okay I actually like the row/col declaration as part of the template for matrix_cl. We can make it backwards compatible with

```cpp
template <typename T, size_t RowsAtCompileTime = -1, size_t ColsAtCompileTime = -1, typename = void>
class matrix_cl { //...
```

t4c1:
Fine. I think to actually use this in GLMs we will need also a change in the compiler to create row matrix_cl objects.

t4c1:
Actually we should not need any backward compatibility. If we go into this we must make all GPU code support it and also return appropriate values. This would be a lot of work for little benefit.

SteveBronder:
idt it would be that much would it? I think it would be

1. Add the templates
2. Check the templates match up for the constructor called 
2. Look through the functions and setup the rows/cols at compile time when appropriate

Everything by default can have -1, -1 and wouldn't get an optimization which is fine. Then for things like transpose, mul, the glm methods, etc. we would do the compile time rows and cols stuff

t4c1:
>> Everything by default can have -1, -1 and wouldn't get an optimization which is fine.

Nope. Its not just about optimizations. If we do things like that some things can only be done if a compile time dimension is 1 - like that example of GLMs broadcasting. So all the functions must return correct compile time dimensions.

t4c1:
@SteveBronder Do you still insist on adding compile time dimensions to `matrix_cl`? I would really prefere to avoid modifying all OpenCL code.

SteveBronder:
> @SteveBronder Do you still insist on adding compile time dimensions to matrix_cl? I would really prefere to avoid modifying all OpenCL code.

Let me give this a shot this weekend and if it's way too much then we won't do it

SteveBronder:
I got the basic API up, though it still just assumes everything is -1, -1 defined. Ran into a few issues when working with matrices of different compile time sizes and local scoped data I need to sort out

https://github.com/stan-dev/math/tree/feature/opencl-compilesizes

t4c1:
I quickly looked over what you did. Looks fine, but that is only the first part. All other GPU functioins would need to be modified too. That would also make those functions accept templated matrices. This prevents automatic evaluations of kernel generator expressions. To make this work again we would need to generalize all these functions to accept expressions. This means we need a set of requires for working with `matrix_cl`. 

Can we instead add optional broadcasting? If you get trough all this we can easily remove it later.

SteveBronder:
Okay yeah that's fine I'll review this today

stan/math/opencl/kernel_generator/optional_broadcast.hpp
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Brodcast an expression in specified dimension(s) if the size along that
+ * dimmension equals to 1. In that case further
+ * expressions can use this expression as if had any size in the broadcast
+ * dimension, repeating the values.
+ *
+ * Broadcasting evaluates argument expression multiple times. For performance
+ * reasons don't broadcast slow operations. Instead evaluate them in a separate
+ * kernel.
```
SteveBronder, line 186:
```suggestion
/**
 * Broadcast an expression in specified dimension(s) if the size along that
 * dimension equals 1. In that case further
 * expressions can use this expression as if it had any size in the broadcast
 * dimension, repeating the values.
 *
 * Broadcasting evaluates the expression argument multiple times. For performance
 * reasons don't broadcast slow operations. Instead evaluate them in a separate
 * kernel.
```

rok-cesnovar:
Just posting to make sure this wait for after the feature freeze ends which should be later today.

PR #1827, opened 2020-04-07
Add additional testing for Eigen type traits and new definition of is_eigen_matrix

SteveBronder:
## Summary

This is the first half of #1667 that cleans up and expands the type traits for selecting certain Eigen types. This uses Eigen's inheritance structure to determine for each type trait whether the type we are checking inherits a particular Eigen base class. In order, the hierarchy for types is essentially

- `EigenBase` 
    - `DenseBase`
        - `MatrixBase`
            - `Matrix`
        - `ArrayBase`
            - `Array`
        - `MapBase`
        - `LDLT`
        - `LLT`
    - `SparseMatrixBase`
        - `SparseCompressedMatrixBase` 
            - `SparseMatrix`

This PR just does `EigenBase`, `MatrixBase`.

This also changes the definition of `is_eigen_matrix` to be anything that inherits from `Eigen::MatrixBase` with more than one row and column. There were places where we want to specialize for eigen vectors while having other specialization for general eigen matrices. This meant we had to do something like

```cpp
template <typename EigMat, typename = require_eigen_t<EigMat>,
  typename = require_not_eigen_vector_t<EigMat>>
```

Which is annoying. With this PR we can just write

```cpp
template <typename EigMat, typename = require_eigen_matrix_t<EigMat>>
```

## Tests

Tests are done for generic eigen expressions inheriting from each type above and for all stan scalar types. They can be run with

```cpp
./runTests.py ./test/unit/math/mix/meta/ -f eigen
```

## Side Effects

The new defintion of `is_eigen_matrix` should be noted. When users want an object that inherits from `MatrixBase` such as a function to work with matrices and vectors they should use `is_eigen_matrix_base`.

## Release notes

Adds additional testing for checking type traits of Eigen objects and modifies type traits for Eigen matrices so detecting an Eigen matrix will only be true if neither rows or columns are equal to 1 

## Checklist

- [x] Math issue #1308


test/unit/math/mix/meta/eigen_utils.hpp
```diff
@@ -0,0 +1,298 @@
...
+  EXPECT_FALSE((Checker<int>::value));
+
+  // Sparse
+  EXPECT_TRUE((eigen_base_v == Checker<EigenBase<decltype(C * D)>>::value));
+  EXPECT_TRUE((eigen_sparse_compressed_v
+               == Checker<SparseCompressedBase<decltype(C * D)>>::value));
+  EXPECT_TRUE((eigen_sparse_matrix_v
+               == Checker<SparseMatrixBase<decltype(C * D)>>::value));
+  EXPECT_TRUE((eigen_sparse_matrix_v == Checker<decltype(C * D)>::value));
+  EXPECT_TRUE((eigen_sparse_matrix_v == Checker<decltype(C * D)>::value));
```
t4c1, line 49: Duplicated check.

test/unit/math/mix/meta/eigen_utils.hpp
```diff
@@ -0,0 +1,298 @@
...
+  EXPECT_FALSE((Checker<bool>::value));
+  EXPECT_FALSE((Checker<double>::value));
+  EXPECT_FALSE((Checker<int>::value));
+
+  // Sparse
+  EXPECT_TRUE((eigen_base_v == Checker<EigenBase<decltype(C * D)>>::value));
+  EXPECT_TRUE((eigen_sparse_compressed_v
+               == Checker<SparseCompressedBase<decltype(C * D)>>::value));
+  EXPECT_TRUE((eigen_sparse_matrix_v
+               == Checker<SparseMatrixBase<decltype(C * D)>>::value));
```
t4c1, line 47: Why do we need checking with base classes? They never explicitely instantinated in our code (only in eigen functions taking them as arguments). We are only using derived types - only those need to be checked.

I think we need to check with a plain type (for example `declytpe(C)`) and some expression (for example `decltype(C+D)`). Map is fine too (actually, does map for sparse matrix even make sense?).

test/unit/math/mix/meta/eigen_utils.hpp
```diff
@@ -0,0 +1,298 @@
...
+  // scalars
+  EXPECT_FALSE((Checker<bool>::value));
+  EXPECT_FALSE((Checker<double>::value));
+  EXPECT_FALSE((Checker<int>::value));
+
+  //  Dense Matrix Hierarchy
+  EXPECT_TRUE((eigen_base_v == Checker<EigenBase<test_mat>>::value));
+  EXPECT_TRUE((eigen_base_v == Checker<EigenBase<decltype(A * B)>>::value));
+  EXPECT_TRUE((eigen_dense_v == Checker<DenseBase<decltype(A * B)>>::value));
+  EXPECT_TRUE((eigen_matrix_v == Checker<MatrixBase<decltype(A * B)>>::value));
```
t4c1, line 94: Again checks on base classes. Instead I would suggest adding a check on `decltype((A * B).array())` and on plain `A`.

test/unit/math/prim/meta/is_eigen_test.cpp
```diff
@@ -1,57 +0,0 @@
...
-  EXPECT_TRUE(
-      (is_eigen<Eigen::MatrixBase<Eigen::Matrix<double, -1, -1>>&&>::value));
-
-  Eigen::Matrix<double, -1, -1> a;
-  Eigen::Matrix<double, -1, -1> b;
-  Eigen::VectorXd c;
-
-  EXPECT_TRUE((is_eigen<decltype(a * b)>::value));
-  EXPECT_TRUE((is_eigen<decltype(a * b + a.transpose())>::value));
-  EXPECT_TRUE((is_eigen<decltype(c.segment(0, 1))>::value));
```
t4c1, line 30: This check was added after a bug was found. It must be included in the new testing framework.

test/unit/math/prim/meta/is_eigen_test.cpp
```diff
@@ -1,57 +0,0 @@
...
-  EXPECT_FALSE((is_eigen_matrix<std::vector<double>>::value));
-
-  EXPECT_FALSE((is_eigen_matrix<Eigen::EigenBase<Eigen::MatrixXd>>::value));
-  EXPECT_TRUE((is_eigen_matrix<Eigen::Matrix<double, -1, -1>>::value));
-  Eigen::SparseMatrix<double> sparse_mat;
-  EXPECT_TRUE((is_eigen_matrix<decltype(sparse_mat)>::value));
-  EXPECT_FALSE((is_eigen_matrix<Eigen::MatrixBase<Eigen::MatrixXd>>::value));
-
-  EXPECT_TRUE((is_eigen_matrix<const Eigen::Matrix<double, -1, -1>>::value));
-  EXPECT_TRUE((is_eigen_matrix<Eigen::SparseMatrix<double>&>::value));
```
t4c1, line 48: I would also like to see at least one test with references to remain.

stan/math/prim/meta/is_eigen.hpp
```diff
@@ -2,42 +2,41 @@
...
-struct is_eigen<
-    T, std::enable_if_t<internal::is_eigen_base<std::decay_t<T>>::value>>
+struct is_eigen<T, std::enable_if_t<std::is_base_of<
+                       Eigen::EigenBase<typename std::decay_t<T>::MatrixType>,
+                       typename std::decay_t<T>::MatrixType>::value>>
     : std::true_type {};
 
+template <typename T>
+struct is_eigen<Eigen::EigenBase<T>, void> : std::true_type {};
+
```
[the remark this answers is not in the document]
SteveBronder: The new type traits check that the type inherits from a base class, but `EigenBase` is the base class everything inherits from and it does not inherit from itself. So this particular type trait needs an actual specialization for the base class.

test/unit/math/mix/meta/eigen_utils.hpp
```diff
@@ -0,0 +1,298 @@
...
+template <bool eigen_base_v, bool eigen_dense_v, bool eigen_matrix_v,
+          bool eigen_matrix_expr_v, bool eigen_map_v,
+          template <class...> class Checker, int... EigenDims>
+void all_eigen_dense_matrix() {
+  using stan::math::fvar;
+  using stan::math::var;
+
+  test_eigen_dense<eigen_base_v, eigen_dense_v, eigen_matrix_v,
+                   eigen_matrix_expr_v, eigen_map_v, Checker, Eigen::Matrix,
+                   double, EigenDims...>();
```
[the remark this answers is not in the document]
SteveBronder: Yeah idt we need it, but it's not bad

stan/math/prim/meta/is_eigen_matrix.hpp
```diff
@@ -0,0 +1,59 @@
...
+           Eigen::MatrixBase<typename std::decay_t<T>::PlainObject>,
+           typename std::decay_t<T>::PlainObject>::value>>
+    : bool_constant<internal::is_eigen_matrix_impl<std::decay_t<T>>::value> {};
+
+template <typename T>
+struct is_eigen_matrix<
+    T, std::enable_if_t<std::is_base_of<
+           Eigen::MatrixBase<typename std::decay_t<T>::MatrixType>,
+           typename std::decay_t<T>::MatrixType>::value>>
+    : bool_constant<internal::is_eigen_matrix_impl<std::decay_t<T>>::value> {};
```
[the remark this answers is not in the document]
SteveBronder: This is for LDLT and the other decompositions. They dont have a `PlainObject` and instead have a matrix type. But should the decompositions fail these checks? For instance arithmetic ops don't exist for decompositions against matrix classes

https://godbolt.org/z/vvzH9R

stan/math/prim/meta/is_eigen.hpp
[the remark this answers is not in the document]
SteveBronder: Here's an example

https://godbolt.org/z/O9O4DX

stan/math/prim/meta/is_eigen_matrix_base.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace stan {
+
+/** \addtogroup type_trait
+ *  @{
+ */
+
+/**
+ * Checks whether type T is derived from Eigen::MatrixBase and has columns and
+ * rows not equal to 1. If true this will have a
```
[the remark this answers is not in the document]
SteveBronder: Gah! Fixed good catch

test/unit/math/mix/meta/eigen_utils.hpp
SteveBronder: Changed to an addition

test/unit/math/mix/meta/eigen_utils.hpp
SteveBronder: > Why do we need checking with base classes? They never explicitly instantiated in our code (only in eigen functions taking them as arguments). We are only using derived types - only those need to be checked.

What about something like `(A * B).matrix()`? That get's wrapped in a MatrixBase

https://godbolt.org/z/TFXJ5Q

> I think we need to check with a plain type (for example declytpe(C)) and some expression (for example decltype(C+D)). Map is fine too (actually, does map for sparse matrix even make sense?).

`decltype(C)` would be the same type as `C` no? Which would be covered by the check on `test_mat`

https://godbolt.org/z/jj6Ywc

> Map is fine too (actually, does map for sparse matrix even make sense?).

I think so? SparseMapBase is a thing. I could add a test for just plain `Map`

test/unit/math/mix/meta/eigen_utils.hpp
SteveBronder: See my comment above. I agree with a check on `.array()`, though that's also covered by the `ArrayBase` test

test/unit/math/prim/meta/is_eigen_test.cpp
SteveBronder: Yes good catch. I'll include tests for this and block()

test/unit/math/prim/meta/is_eigen_test.cpp
SteveBronder: Sorry what do you mean by remain?

stan/math/prim/meta/is_eigen.hpp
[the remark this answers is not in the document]
bob-carpenter: Would assignability to `EigenBase` be a way to implemen this?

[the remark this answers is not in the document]
SteveBronder: `EigenBase` is not really something that is assigned to. The base class for Eigen essentially just defines their lowest level API that all their other classes make available.

I thought about defining these by assignability, but it starts getting a bit weird. Like [here](https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/EigenBase.h#L134) where they have assignment for `EigenBase` derived expressions to `DenseBase` objects. The way they use CRTP is to define layers of APIs so for our purposes what we want is to check that an object inherits from a particular class so that we can use that class's API.

https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/EigenBase.h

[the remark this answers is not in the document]
t4c1: > The new type traits check that the type inherits from a base class

Why would we nee to change the definition of `is_eigen` in this way? That example only shows a bug in a halfway changed implementation, not in existing one.

> Would assignability to EigenBase be a way to implemen this?

Yes, but `EigenBase` is a template. To check assignability we would still need to decide on template arguments. It would probably allow to remove one case from the implementation suggested in this PR. But I still did not see any arguments why we need to change it at all.

> I thought about defining these

The issue here is change of `is_eigen`. The suggested implementation is both longer, more convoluted and has some weird issues as seen in the example on https://godbolt.org/z/7xxK9n. Other type traits can be discussed separately.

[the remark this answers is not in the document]
t4c1: Actually ... maybe we could implement all or most type traits by checking for pointer conversions. Here is an example for `is_eigen_matrix_base` : https://godbolt.org/z/or6NPL

stan/math/prim/meta/is_eigen_matrix.hpp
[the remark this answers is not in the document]
t4c1: I think decompositions should definitely fail `is_eigen_matrix` and similar checks.

test/unit/math/mix/meta/eigen_utils.hpp
t4c1: > What about something like (A * B).matrix()?

Nice catch. I did not know about this one. I still think we can't get `EigenBase`. Or can we?

> decltype(C) would be the same type as C no? Which would be covered by the check on test_mat

Right.

> I think so? SparseMapBase is a thing. I could add a test for just plain Map

I will leave that up to you. I am not very familiar with sparse matrices.

test/unit/math/mix/meta/eigen_utils.hpp
t4c1: Where is `ArrayBase` test? I can't find it. If I follow your logic it should be in this function?

test/unit/math/mix/meta/eigen_utils.hpp
```diff
@@ -0,0 +1,298 @@
...
+ *   `Map` expression types.
+ * @tparam Checker A struct taking in an expression template type and returning
+ *  a true if check is satisfied and false otherwise.
+ * @tparam EigenDims template parameter pack of `int` defining matrix
+ *  compile time dimensions.
+ */
+template <bool eigen_base_v, bool eigen_dense_v, bool eigen_array_v,
+          bool eigen_array_expr_v, bool eigen_map_v,
+          template <class...> class Checker, int... EigenDims>
+void all_eigen_dense_array() {
```
t4c1: Can you unify the formatting between this and previous function? Differences in blank lines make it less obvious they are almost the same.

test/unit/math/prim/meta/is_eigen_test.cpp
t4c1: I see no tests on references in the new testing framework. Instead of removing all of them I would like to see at least one test on reference remain in stan math.

test/unit/math/mix/meta/eigen_utils.hpp
t4c1: I see I was looking at subset of commits. This is fine.

test/unit/math/mix/meta/is_eigen_matrix_base_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+  all_eigen_dense_exprs<true, true, true, true, is_eigen_matrix_base, Matrix, 1,
+                        -1>;
+  all_eigen_dense_exprs<true, true, true, true, is_eigen_matrix_base, Matrix,
+                        -1, 1>;
+  all_eigen_dense_exprs<true, true, true, true, is_eigen_matrix_base, Array, -1,
+                        -1>;
+  all_eigen_dense_exprs<true, true, true, true, is_eigen_matrix_base, Array, 1,
+                        -1>;
+  all_eigen_dense_exprs<true, true, true, true, is_eigen_matrix_base, Array, -1,
+                        1>;
```
t4c1, line 57: `is_eigen_matrix_base` should fail for `Array` expressions should it not?

stan/math/prim/meta/is_eigen.hpp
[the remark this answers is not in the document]
SteveBronder: Huh! That's a really neat approach! Let me fiddle around with this. I like it's end result but it's a bit confusing (though tbh so is the current thing I'm proposing)

bob-carpenter:
What's the motivation to switching to pointer conversion?

bob-carpenter:
Sorry, forgot to finish.  Doesn't pointer conversion assume the same binary representation?  Assignability just means a string of appropriate conersions (either built-in or through implicit constructors).

SteveBronder:
> Sorry, forgot to finish. Doesn't pointer conversion assume the same binary representation? Assignability just means a string of appropriate conersions (either built-in or through implicit constructors).

I have to look this up. We can't use `is_assignable` because Arrays are assignable to Matrices so we wouldn't be able to differentiate them

https://godbolt.org/z/DCg9-M

SteveBronder:
So pulling this apart, the docs for [`is_convertible`](https://en.cppreference.com/w/cpp/types/is_convertible) are pretty useful since this does a similar thing but with pointer conversion

> If the imaginary function definition To test() { return std::declval<From>(); } is well-formed, (that is, either std::declval<From>() can be converted to To using implicit conversions, or both From and To are possibly cv-qualified void), provides the member constant value equal to true. Otherwise value is false. For the purposes of this check, the use of std::declval in the return statement is not considered an odr-use.

I'm looking at the docs for [implicit conversion](https://en.cppreference.com/w/cpp/language/implicit_conversion#Pointer_conversions) and it doesn't mention the same binary representation. Is that in the standard somewhere? My understanding was that pointer types follow the same implicit conversion rules

bob-carpenter:
I lost the context.  What is similar to what in `is_convertible`?

I'm probably misunderstanding what is meant by pointer conversion here.  I don't see how either of the second two lines could work unless `bar` and `foo` were identical.  

```
foo* x = ...;
bar* y = static_cast<foo*>(x);
bar* y = x;
```

t4c1:
@bob-carpenter `foo*` is also convertible to `bar*` if `foo` is derived from `bar` or `bar` is derived from `foo`. My suggestion was using the former in attempt to determine if some Eigen base class is a base of certain type.

Why can't we use `std::is_convertible` (or even `std::is_base_of`) in the same manner? These two only check exact types. In our case the base class in our check is actually a class template (for example `EigenBase`) and we want to know if there exists such template parameter (`Derived`) that would make it (`EigenBase<Derived>`) a base class of a certain type we are checking (for example `Eigen::Matrix<var, -1, -1>`).

t4c1:
> What's the motivation to switching to pointer conversion?

The code is simpler and less prone to bugs. I changed the implementation of `is_eigen` to pointer conversions, when I found a bug in previous Implementation - it did not recognize segment type as Eigen type, since it was indirectely derived from EigenBase, not directely as most classes.

bob-carpenter:
@t4c1 Thanks for the clear explanation of how the templating's working here.  It never ceases to amaze me.  I also keep forgetting that C++ supports all these non-type-safe operations that can lead to slicing.

SteveBronder:
Thanks @t4c1! Lemme read this over then I'll ping you for review

SteveBronder:
@t4c1 this is ready for review

test/unit/math/mix/meta/eigen_utils.hpp
```diff
@@ -0,0 +1,329 @@
...
+  using Eigen::DenseBase;
+  using Eigen::EigenBase;
+  using Eigen::Map;
+  using Eigen::MatrixBase;
+  std::remove_reference_t<EigenType> A;
+  std::remove_reference_t<EigenType> B;
+  // scalars
+  EXPECT_FALSE((Checker<bool>::value));
+  EXPECT_FALSE((Checker<double>::value));
+  EXPECT_FALSE((Checker<int>::value));
```
t4c1, line 83: I think checks on scalars should be in one of the `all_*` functions.

PR #1834, opened 2020-04-13
Fix ode gradients with respect to t0 (Issue #1833)

### LINKED ISSUE #1833: ODE gradients with respect to t0 not working correctly
## Description
If t0 is a parameter, the gradients of the solution with respect to t0 aren't working properly.

The gradients currently aren't set (partials3_ isn't set in here: https://github.com/stan-dev/math/blob/develop/stan/math/prim/functor/coupled_ode_observer.hpp#L123)

There are tests, but the test problem gradients are actually zero in some cases and tested against zero in any case so the problem isn't showing up.

## Example

Here's test code that will break the current implementations (this affects the rk45 solver, the bdf, and the adams solvers):
```
struct CosArg1 {
  template <typename T0, typename T1, typename T_Arg>
  inline std::vector<stan::return_type_t<T1, T_Arg>>
  operator()(const T0& t, const std::vector<T1>& y,
             const std::vector<T_Arg>& a,
             const std::vector<double>& x,
             const std::vector<int>& x_int,
             std::ostream* msgs) const {
    return { stan::math::cos(a[0] * t) };
  }
};

TEST(StanMathOde_ode_bdf_tol, t0) {
  using stan::math::var;

  std::vector<double> y0 = { 0.0 };
  double dt = 1e-5;
  double t0d = 0.0;
  var t0 = t0d;
  std::vector<double> ts = { 0.45, 1.1 };

  std::vector<double> a = { 1.5 };
  std::vector<double> x;
  std::vector<int> x_int;

  std::vector<std::vector<var>> output = stan::math::integrate_ode_bdf(CosArg1(), y0, t0, ts, a, x, x_int);

  auto f = [&](double t0) {
    return stan::math::integrate_ode_bdf(CosArg1(), y0, t0, ts, a, x, x_int);
  };

  output[0][0].grad();

  std::cout << "autodiff bdf: " << t0.adj() <<
    ", finite diff bdf: " << (f(t0d + dt)[0][0] - f(t0d - dt)[0][0]) / (2 * dt) << std::endl;

  EXPECT_FLOAT_EQ(output[0][0].val(), 0.4165982112);
  EXPECT_FLOAT_EQ(t0.adj(), -1);

  output[1][0].grad();

  EXPECT_FLOAT_EQ(output[1][0].val(), 0.66457668563);
  EXPECT_FLOAT_EQ(t0.adj(), -1.0);
}

TEST(StanMathOde_ode_rk45_tol, t0) {
  using stan::math::var;

  std::vector<double> y0 = { 0.0 };
  double dt = 1e-5;
  double t0d = 0.0;
  var t0 = t0d;
  std::vector<double> ts = { 0.45, 1.1 };

  std::vector<double> a = { 1.5 };
  std::vector<double> x;
  std::vector<int> x_int;

  std::vector<std::vector<var>> output = stan::math::integrate_ode_rk45(CosArg1(), y0, t0, ts, a, x, x_int);

  auto f = [&](double t0) {
    return stan::math::integrate_ode_rk45(CosArg1(), y0, t0, ts, a, x, x_int);
  };

  output[0][0].grad();

  std::cout << "autodiff rk45: " << t0.adj() <<
    ", finite diff rk45: " << (f(t0d + dt)[0][0] - f(t0d - dt)[0][0]) / (2 * dt) << std::endl;

  EXPECT_FLOAT_EQ(output[0][0].val(), 0.4165982112);
  EXPECT_FLOAT_EQ(t0.adj(), -1);

  output[1][0].grad();

  EXPECT_FLOAT_EQ(output[1][0].val(), 0.66457668563);
  EXPECT_FLOAT_EQ(t0.adj(), -1.0);
}
```

#### Current Version:
v3.1.0

bbbales2:
Fixes #1833 

## Tests

There were tests in place, but either:
1. They were testing against ```forced_harm_osc_ode_fun``` which actually had zero gradients
2. Or they were testing that the gradients were zero

So not setting the gradients was accidentally making these things pass. I changed the tests to use ```harm_osc_ode_fun``` where one of the gradients isn't zero, and I updated the checks to look for that.

## Release notes

Fixed problem where ode solvers were not computing the gradients of the output with respect to the initial integration time. This affected the rk45, adams, and bdf solvers.

## Checklist

- [x] Math issue #1833

### DIFF
--- a/stan/math/prim/functor/coupled_ode_observer.hpp
+++ b/stan/math/prim/functor/coupled_ode_observer.hpp
@@ -44,6 +44,7 @@ struct coupled_ode_observer {
   const std::size_t N_;
   const std::size_t M_;
   const std::size_t index_offset_theta_;
+  std::vector<double> f_y0_t0_;
   int next_ts_index_;
 
   /**
@@ -84,7 +85,12 @@ struct coupled_ode_observer {
         N_(y0.size()),
         M_(theta.size()),
         index_offset_theta_(is_constant_all<T1>::value ? 0 : N_ * N_),
-        next_ts_index_(0) {}
+        next_ts_index_(0) {
+    if (!is_constant_all<T_t0>::value) {
+      f_y0_t0_ = f(value_of(t0), value_of(y0), value_of(theta_), x_, x_int_, msgs_);
+      check_size_match("coupled_ode_observer", "dy_dt", f_y0_t0_.size(), "states", N_);
+    }
+  }
 
   /**
    * Callback function for ODE solvers to record values. The coupled
@@ -131,6 +137,10 @@ struct coupled_ode_observer {
           ops_partials.edge2_.partials_[k]
               = coupled_state[N_ + index_offset_theta_ + N_ * k + j];
         }
+      }
+
+      if (!is_constant_all<T_t0>::value) {
+	ops_partials.edge3_.partials_[0] = -f_y0_t0_[j];
       }
 
       if (!is_constant_all<T_ts>::value) {
--- a/stan/math/prim/functor/coupled_ode_observer.hpp
+++ b/stan/math/prim/functor/coupled_ode_observer.hpp
@@ -44,7 +44,7 @@ struct coupled_ode_observer {
   const std::size_t N_;
   const std::size_t M_;
   const std::size_t index_offset_theta_;
-  std::vector<double> f_y0_t0_;
+  std::vector<double> initial_dy_dt_;
   int next_ts_index_;
 
   /**
@@ -87,9 +87,9 @@ struct coupled_ode_observer {
         index_offset_theta_(is_constant_all<T1>::value ? 0 : N_ * N_),
         next_ts_index_(0) {
     if (!is_constant_all<T_t0>::value) {
-      f_y0_t0_
+      initial_dy_dt_
           = f(value_of(t0), value_of(y0), value_of(theta_), x_, x_int_, msgs_);
-      check_size_match("coupled_ode_observer", "dy_dt", f_y0_t0_.size(),
+      check_size_match("coupled_ode_observer", "dy_dt", initial_dy_dt_.size(),
                        "states", N_);
     }
   }
@@ -142,7 +142,7 @@ struct coupled_ode_observer {
       }
 
       if (!is_constant_all<T_t0>::value) {
-        ops_partials.edge3_.partials_[0] = -f_y0_t0_[j];
+        ops_partials.edge3_.partials_[0] = -initial_dy_dt_[j];
       }
 
       if (!is_constant_all<T_ts>::value) {
--- a/stan/math/prim/functor/coupled_ode_observer.hpp
+++ b/stan/math/prim/functor/coupled_ode_observer.hpp
@@ -87,8 +87,10 @@ struct coupled_ode_observer {
         index_offset_theta_(is_constant_all<T1>::value ? 0 : N_ * N_),
         next_ts_index_(0) {
     if (!is_constant_all<T_t0>::value) {
-      f_y0_t0_ = f(value_of(t0), value_of(y0), value_of(theta_), x_, x_int_, msgs_);
-      check_size_match("coupled_ode_observer", "dy_dt", f_y0_t0_.size(), "states", N_);
+      f_y0_t0_
+          = f(value_of(t0), value_of(y0), value_of(theta_), x_, x_int_, msgs_);
+      check_size_match("coupled_ode_observer", "dy_dt", f_y0_t0_.size(),
+                       "states", N_);
     }
   }
 
@@ -140,7 +142,7 @@ struct coupled_ode_observer {
       }
 
       if (!is_constant_all<T_t0>::value) {
-	ops_partials.edge3_.partials_[0] = -f_y0_t0_[j];
+        ops_partials.edge3_.partials_[0] = -f_y0_t0_[j];
       }
 
       if (!is_constant_all<T_ts>::value) {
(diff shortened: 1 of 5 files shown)


wds15:
Thanks for catching this! I can review this later.

bbbales2:
@wds15, @yizhang-yiz, if the tests don't just pass on the first try, we've probably missed this release, but if you get a chance to review this today that'd be nice.

rok-cesnovar:
This is a bugfix so it doesnt fall under the feature freeze. Still better to have it in the release candidate.

stan/math/prim/functor/coupled_ode_observer.hpp
```diff
@@ -133,6 +141,10 @@ struct coupled_ode_observer {
         }
       }
 
+      if (!is_constant_all<T_t0>::value) {
+        ops_partials.edge3_.partials_[0] = -f_y0_t0_[j];
```
wds15, line 145: Why would this be equal to *minus* f_y0_t0_[j]? As I understand, the initial time-point is no different than the remaining time-points such that the derivative wrt to t0 is just f_y0_t0. I mean, in this version we would have at t0 the gradient equal to -f_y0_t0_[j] and just an epsilon later the sign would flip if the user would request at t0+epsilon a solution of the ODE.

So please change to f_y0_t0_[j] unless I am overlooking something here.

I am also not too happy with the variable name as this is called dy_dt usually. Maybe initial_dy_dt_ ?

test/unit/math/rev/functor/coupled_ode_observer_test.cpp
```diff
@@ -221,7 +221,7 @@ TEST_F(StanRevOde, observe_states_ddvd) {
       EXPECT_FLOAT_EQ(ys_coupled[t][n], y[t][n].val());
     for (size_t n = 0; n < 2; n++) {
       y[t][n].grad();
-      EXPECT_FLOAT_EQ(0.0, t0.adj());
+      EXPECT_FLOAT_EQ(-y0[n], t0.adj());
```
wds15, line 224: here as well, the test expectation is y0[n], I think.

stan/math/prim/functor/coupled_ode_observer.hpp
```diff
@@ -84,7 +85,14 @@ struct coupled_ode_observer {
         N_(y0.size()),
         M_(theta.size()),
         index_offset_theta_(is_constant_all<T1>::value ? 0 : N_ * N_),
-        next_ts_index_(0) {}
+        next_ts_index_(0) {
+    if (!is_constant_all<T_t0>::value) {
+      f_y0_t0_
+          = f(value_of(t0), value_of(y0), value_of(theta_), x_, x_int_, msgs_);
+      check_size_match("coupled_ode_observer", "dy_dt", f_y0_t0_.size(),
```
wds15, line 92: Good catch - we need the check here. Is this check triggered somewhere in the tests?

bbbales2: No, but I can add that.

stan/math/prim/functor/coupled_ode_observer.hpp
bbbales2: > Why would this be equal to minus f_y0_t0_[j]?

Imagine an ode solve with only an initial point and a final point. As an integral, if the final point moves right, the value of the integral gets larger. If the initial point moves right, that integral gets smaller (or the thing here: https://en.wikipedia.org/wiki/Leibniz_integral_rule).

> as this is called dy_dt usually

Yeah I was working through making the odes variadic I never found naming that I was super happy with.

dy_dt is just f, by definition, so it's annoying to type the fully dy_dt. I kinda liked f_y0_t0 cause it's evaluated at the initial points (still really painful to type).

I'm happy to go initial_dy_dt. Doesn't matter that much to me.

wds15: but at t0+epsilon the gradient would flip sign. That is not right. Either we have a bigger problem (and the gradients are wrong for the other time-points as well) or this PR is wrong, but I don't think that both are right.

As you are suggesting that your version is the right one, I would need to to dive into it a bit.

Let's keep this as a bugfix which we can merge during this week.

bbbales2: > but at t0+epsilon the gradient would flip sign.

The derivative with respect to the lower limit of the integral will still have that negative sign. Write out the integral for something like `cos(t)` from `t = l` to `t = r` and then differentiate that integral with respect to `l` and `r` and see the difference.

There's finite difference code here if you want to check with that: https://github.com/stan-dev/math/issues/1833 . I wrote this out to make sure I wasn't too off track.

wds15: Sure... I just need a calm moment tomorrow to write this down. It's against my intuition, so I won't approve for just now.

If someone else is confident this is right, then I withdraw my review.

(I guess you are right, I just want to follow along).

bbbales2: Yeah no it's fine it's confusing lol. Tomorrow is good for this.

wds15: Got it. Confusing, but you are right. The odd thing is that the initial is constant here while the initial time varies.

Can you add the test catching the size mis match thing below? Then this is fine to go in.

stan/math/prim/functor/coupled_ode_observer.hpp
bbbales2: Checks added!

Issue #1829, opened 2020-04-09
Eigen matrix traits for complex numbers


bob-carpenter:
## Description

Add the required traits so that Eigen arithmetic operators and matrix functions work with Stan complex types.

#### Current Version:
v3.1.0

PR #1838, opened 2020-04-13
Feature/1829 complex eigen traits

bob-carpenter:
# Summary

Added `Eigen::ScalarBinaryOpTraits` for forward and reverse-mode autodiff to allow matrix operations on matrices of complex autodiff variables.

Added new functions to demonstrate usage:

* `eigenvalues`
* `eigenvectors`
* `complex_schur_decompose`
* `pseudo_eigenvalues`
* `pseudo_eigenvectors`

## Tests

All new functions are unit tested and the traits are tested in all combinations in `multiply_test.cpp`.

## Side Effects

No.

## Release notes

* extended Eigen scalar traits to complex autodiff types to enable complex autodiff matrix operations
* added eigenvalues(), eigenvectors(), pseudo_eigenvalues(), pseudo_eigenvectors(), and complex_schur_decompose() functions

## Checklist

- [x] Math issue #1829

- [ ] Copyright holder: Simons Institute


stan/math/prim/fun/complex_schur_decompose.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/prim/err.hpp>
+
+namespace stan {
+namespace math {
+
+template <typename T>
+Eigen::Matrix<std::complex<T>, -1, -1> complex_schur_decompose(
+    const Eigen::Matrix<T, -1, -1>& m) {
```
SteveBronder, line 12: Docs?

stan/math/prim/fun/isinf.hpp
```diff
@@ -17,10 +19,9 @@ namespace math {
  * @param[in] v argument
  * @return true if argument is infinite
  */
-template <typename ADType, require_autodiff_t<ADType>...>
-inline bool isinf(ADType&& v) {
-  using std::isinf;
-  return isinf(v.val());
+template <typename T>
+inline bool isinf(const T& v) {
+  return is_inf(v);
```
SteveBronder, line 24: Why the change to is_inf?

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -96,10 +94,7 @@ class precomputed_gradients_vari : public vari {
  * @return An autodiff variable that uses the precomputed
  * gradients provided.
  */
-template <typename Arith, typename VecVar, typename VecArith,
-          require_arithmetic_t<Arith>...,
-          require_vector_like_vt<is_var, VecVar>...,
-          require_vector_like_vt<std::is_arithmetic, VecArith>...>
+template <typename Arith, typename VecVar, typename VecArith>
```
SteveBronder, line 97: Why are these gone? I think we only want to support arithmetic and vector like objects here? Or should this also take in complex types now?

SteveBronder: If it should take in complex as well then maybe we should have a `is_real` type trait to check if a type is arithmetic or complex with an arithmetic base type.

bob-carpenter: They're gone because (a) they're causing circular include problems between core and meta, which I wasn't able to resolve, and (b) they're not needed.  The traits restrictions aren't necessary when there's no ambiguity to resolve in calling them.  If a client tries to call this code and the required concepts for the arguments (like supporting `operator[]`) aren't satisfied, the function won't compile at that juncture, rather than at the undefined-because-of-SFINAE level.

stan/math/prim/fun/isinf.hpp
bob-carpenter: Because (a) I couldn't resolve all the dependencies otherwise, and (b) we don't need two definitions.  So I killed two birds with one stone and delegated.

stan/math/prim/fun/complex_schur_decompose.hpp
bob-carpenter: Thanks---forgot those.

stan/math/fwd/fun/Eigen_NumTraits.hpp
```diff
@@ -53,22 +54,114 @@ struct NumTraits<stan::math::fvar<T>> : GenericNumTraits<stan::math::fvar<T>> {
...
+ * complex autodiff arguments.
+ *
+ * @tparam T value and tangent type of autodiff variable
+ * @tparam BinaryOp type of binary operation for which traits are
+ * defined
+ */
+template <typename T, typename BinaryOp>
+struct ScalarBinaryOpTraits<double, std::complex<stan::math::fvar<T>>,
+                            BinaryOp> {
+  typedef std::complex<stan::math::fvar<T>> ReturnType;
```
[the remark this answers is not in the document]
bob-carpenter: I'll convert if `using` will define `ScalarBinaryOpTraits<...>::ReturnType`.

stan/math/rev/core/precomputed_gradients.hpp
bob-carpenter: And no, I wasn't thinking about `complex`.  

That's another whole layer of metaprogram reworking as you suggest.  I started in the last complex PR by splitting return traits into `real_return_t<T1, ..., TN>` and `complex_return_t<T1, ..., TN>`).  Having rhe `is_XXX` and `require_XXX` split is also going to be useful.

P.S.  Thanks for looking through this!

stan/math/prim/fun/isinf.hpp
bob-carpenter: And if you can figure out how to put the constraints back and still satisfy all the header and unit tests, then that'd also be OK, though I'd really just as soon not rely on something like this.

stan/math/fwd/fun/Eigen_NumTraits.hpp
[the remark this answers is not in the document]
bob-carpenter: Yes, that's fine.

bob-carpenter:
[This isn't urgent compared to the release.]

Does anyone know what this means:

```
make -j4 test/performance/logistic_test failed
```

I don't see any other report for the error.  Nothing in this PR should be affecting a performance test.

@serban-nicusor-toptal  Any idea where I can find more info on what went wrong?

serban-nicusor-toptal:
Hey, looks like the `old-imac` machine went down during the build, no issues on your side.  
Restarted the job.

serban-nicusor-toptal:
I am seeing this error `sh: /usr/local/git/bin/mkdir: No such file or directory`
Can that be from your side ? If not I need to investigate, machine seems stable now but the build fails because of that in [here](https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/downstream_tests/1507/pipeline)

bob-carpenter:
@bbbales2 I think I hit all of your requirements, so this is just waiting until after the release to merge.  This is new functionality, not a bug fix.

bbbales2:
> I think I hit all of your requirements

I better not find any brown M&Ms in here!

Or was this waiting on @SteveBronder 's review?

bob-carpenter:
D'oh.  Yes, it's @SteveBronder.  And just a reminder that this is a new feature, so I don't think it can be merged into develop until after the release is tagged off develop.

stan/math/prim/fun/isinf.hpp
SteveBronder: Looking through the specializations of is_inf I think this is fine

stan/math/rev/core/precomputed_gradients.hpp
SteveBronder: Alright I'm fine with that and np!

stan/math/rev/fun/Eigen_NumTraits.hpp
```diff
@@ -79,24 +80,154 @@ struct NumTraits<stan::math::var> : GenericNumTraits<stan::math::var> {
...
+// +  7. var, complex<double>
+// +  8. var, complex<var>
+//    9. complex<double>, double
+// + 10. complex<double>, var
+//   11. complex<double>, complex<double>
+// + 12. complex<double>, complex<var>
+// + 13. complex<var>, double
+// + 14. complex<var>, var
+// + 15. complex<var>, complex<double>
+// ? 16. complex<var>, complex<var>
```
SteveBronder, line 100: Should this comment be deleted?

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  }
+  for (const auto& x : stan::test::square_test_matrices(1, 3)) {
+    test_complex_schur_decompose_complex<d_t>(x);
+    // TODO(carpenter): // require expanding compound arithmetic
+    //                  // to more general Eigen types for *=, -=, etc.
+    // test_complex_schur_decompose_complex<v_t>(x);
+    // test_complex_schur_decompose_complex<fd_t>(x);
+    // test_complex_schur_decompose_complex<ffd_t>(x);
+    // test_complex_schur_decompose_complex<fv_t>(x);
+    // test_complex_schur_decompose_complex<ffv_t>(x);
```
SteveBronder, line 99: Do we want to wait on schur until this todo is resolved?

stan/math/rev/fun/Eigen_NumTraits.hpp
bob-carpenter: Yes.  Didn't mean to leave that in.  I have to write these out to get the combinatorics down in tests and code!

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
bob-carpenter: I meant to say that `test_schur_decompose` only works for `double`.   As far as I can tell, the Schur decompositions themselves work as intended.

The problem is that we can't put them back together because of missing operations.  It's line 42 that's the problem:

```
 auto X2 = U * T * U.adjoint();
```

I'm very unclear on why this doesn't work given that all the other complex arithemtic works.  I think it has to do with the expression templates not picking up all the right traits.  Which makes me wonder how the actual calls to Schur and eigendecomposition and what-not work.

SteveBronder: Let me take a stab at it this weekend for a little bit. If I can't figure it out I think we should just remove it

SteveBronder:
Does this also need `NumTraits` for `complex<var>` etc. or is that going in a separate PR?

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
SteveBronder: This may be a bigger issue, I'm seeing some things like

lib/eigen_3.3.3/Eigen/src/Core/GeneralProduct.h:282:25: error: no match for ‘operator+=’
 (operand types are ‘Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<
std::complex<stan::math::var>, -1, 1, 0, -1, 1>, 0, Eigen::Stride<0, 0> > >’ 
and ‘std::complex<stan::math::var>’)
  282 |           dest.matrix() += actualAlpha * MappedDest(actualDestPtr, dest.size());
Do you think it's because Eigen can't find the += for complex?

SteveBronder:
I took a stab at cleaning this up to get it to work for schur but had no luck. If the above thing is less of an issue than I'm hoping then let's just remove Schur for now.

In a separate branch I rewrote the `ScalarBinaryOpTraits` stuff to use a macro. imo I found this simpler, just did 

```R
t(combn(c("double", "var", "complex<double>",
 "complex<var>", "fvar<T>", "complex<fvar<T>>"), m = 2))
#       [,1]              [,2]              
#  [1,] "double"          "var"             
#  [2,] "double"          "complex<double>" 
#  [3,] "double"          "complex<var>"    
#  [4,] "double"          "fvar<T>"         
#  [5,] "double"          "complex<fvar<T>>"
#  [6,] "var"             "complex<double>" 
#  [7,] "var"             "complex<var>"    
#  [8,] "var"             "fvar<T>"         
#  [9,] "var"             "complex<fvar<T>>"
# [10,] "complex<double>" "complex<var>"    
# [11,] "complex<double>" "fvar<T>"         
# [12,] "complex<double>" "complex<fvar<T>>"
# [13,] "complex<var>"    "fvar<T>"         
# [14,] "complex<var>"    "complex<fvar<T>>"
# [15,] "fvar<T>"         "complex<fvar<T>>" 
```

and applied those results excluding `double` `complex<double>`.

https://github.com/stan-dev/math/compare/fixup/1829-complex-eigen-macro-traits

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
bob-carpenter: I think it is a bigger issue and I think it's a problem with assignment.  I'm going to try to get to the bottom of it.  I'm surprised all these algorithms that work in the complex domain work.  But then Schur broke with complex inputs, so maybe there are different implementation paths.

bob-carpenter:
> I took a stab at cleaning this up

I find the macros really confusing.  I'm already having trouble debugging programs with all the traits macros because I can't grep for definitions of things I'm seeing in code.

Any lues from trying? 

> Does this also need NumTraits for complex<var> etc. or is that going in a separate PR?

I have been assuming their generic definition for `complex<T>` would work given that we've defined traits for the appropriate `T`. See [line 172 of `NumTraits.h`](https://eigen.tuxfamily.org/dox/NumTraits_8h_source.html), which defines complex `NumTraits` generically:

```
  172 template<typename _Real> struct NumTraits<std::complex<_Real> >
  173   : GenericNumTraits<std::complex<_Real> >
  174 {
  175   typedef _Real Real;
  176   typedef typename NumTraits<_Real>::Literal Literal;
  177   enum {
  178     IsComplex = 1,
  179     RequireInitialization = NumTraits<_Real>::RequireInitialization,
  180     ReadCost = 2 * NumTraits<_Real>::ReadCost,
  181     AddCost = 2 * NumTraits<Real>::AddCost,
  182     MulCost = 4 * NumTraits<Real>::MulCost + 2 * NumTraits<Real>::AddCost
  183   };
  184  
  185   EIGEN_DEVICE_FUNC
  186   static inline Real epsilon() { return NumTraits<Real>::epsilon(); }
  187   EIGEN_DEVICE_FUNC
  188   static inline Real dummy_precision() { return NumTraits<Real>::dummy_precision(); }
  189   EIGEN_DEVICE_FUNC
  190   static inline int digits10() { return NumTraits<Real>::digits10(); }
  191 };
```

bob-carpenter:
> complex<var>"    "fvar<T>"   

This is wrong.  We can't combine `var` and `fvar`.  The only binary arguments that are legal have either arithmetic types or matching autodiff types.  The logic extends to ternary operations.  

In the binary case, we need all combinations of `(A1, A2)`, `(A1, V)`, `(T, A2)` and `(V, V)`, where `A1` and `A2` are arithmetic types and `V` is an autodiff type. Specifically, if there are two autodiff types, they have to be the same.  

We have to be careful to cover everything without  introducing ambiguity.  This is particularly problematic when only `(A, A)` is built-in and not something more generic like `(A1, A2)` that matches without promotion.

bob-carpenter:
The problems are deeper here.  The matched operations all work, but the mixed operations which should be OK (`var` on left) fail to compile due to missing `madd` instances

```
  Eigen::MatrixXd a = Eigen::MatrixXd::Random(2, 2);
  a *= a; // OK

  Eigen::Matrix<stan::math::var, -1, -1> b = a;  // OK
  b *= b; // OK
  b = b * b; // OK

  b = a * b; // FAILS
  b *= a;  // FAILS
```

The errors look like:

```
lib/eigen_3.3.3/Eigen/src/Core/products/GeneralBlockPanelKernel.h:999:13: error: no matching member function for call to 'madd'
            EIGEN_GEBP_ONESTEP(0);
            ^~~~~~~~~~~~~~~~~~~~~
lib/eigen_3.3.3/Eigen/src/Core/products/GeneralBlockPanelKernel.h:990:22: note: expanded from macro 'EIGEN_GEBP_ONESTEP'
              traits.madd(A2, B_0, C10, B_0); \
              ~~~~~~~^~~~
lib/eigen_3.3.3/Eigen/src/Core/products/GeneralBlockPanelKernel.h:435:28: note: candidate template ignored: deduced conflicting types for parameter 'AccPacketType' ('stan::math::var' vs. 'double')
  EIGEN_STRONG_INLINE void madd(const LhsPacketType& a, const RhsPacketType& b, AccPacketType& c, AccPacketType& tmp) const
```

so it's clear it doesn't know how to mix `var` and `double` despite having the relevant ScalarBinaryOp traits in `stan/math/rev/fun/Eigen_NumTraits.hpp`:

```
/**
 * Traits specialization for Eigen binary operations for reverse-mode
 * autodiff and `double` arguments.
 *
 * @tparam BinaryOp type of binary operation for which traits are
 * defined
 */
template <typename BinaryOp>
struct ScalarBinaryOpTraits<stan::math::var, double, BinaryOp> {
  using ReturnType = stan::math::var;
};

/**
 * Traits specialization for Eigen binary operations for `double` and
 * reverse-mode autodiff arguments.
 *
 * @tparam BinaryOp type of binary operation for which traits are
 * defined
 */
template <typename BinaryOp>
struct ScalarBinaryOpTraits<double, stan::math::var, BinaryOp> {
  using ReturnType = stan::math::var;
};

/**
 * Traits specialization for Eigen binary operations for reverse-mode
 autodiff
 * arguments.
 *
 * @tparam BinaryOp type of binary operation for which traits are
 * defined
 */
template <typename BinaryOp>
struct ScalarBinaryOpTraits<stan::math::var, stan::math::var, BinaryOp> {
  using ReturnType = stan::math::var;
};
```

bob-carpenter:
So I guess the question is where this leaves the Eigen complex things.  That means that some of the Eigen functions don't call these mixed operations internally (like eigendecomposition), but some of them do (like complex Schur).  We're not going to be able to use the ones that call mixed operations internally without sorting out the necessary implementations.  This would have to start down at the autodiff type level and then move onto complex, I'd think.  I wouldn't want to do complex without a working real version.

SteveBronder:
> I find the macros really confusing. I'm already having trouble debugging programs with all the traits macros because I can't grep for definitions of things I'm seeing in code.

That's reasonable, probably an easy to write hard to read on my end.

> So I guess the question is where this leaves the Eigen complex things. That means that some of the Eigen functions don't call these mixed operations internally (like eigendecomposition), but some of them do (like complex Schur). We're not going to be able to use the ones that call mixed operations internally without sorting out the necessary implementations. This would have to start down at the autodiff type level and then move onto complex, I'd think. I wouldn't want to do complex without a working real version.

Let me stare at that code for an hour today and see if I can't sort it out. It seems to be failing at some very specific places in their code. I also want to check out what the other autodiff w/ complex eigen folks did in this situation

SteveBronder:
Okay so the issue here is that we need `gevv_vdv_vari` and `gevv_vvd_vari` specializations for double and var, then we need specializations for general_matrix_vector_product etc. For a little bit I went full doomer and thought we would need to write our own pmadd etc. but I think the above is right.

What's happening is that right now we are calling Eigen's general matrix product for var and double stuff. Specifically the lines here are the bads

```cpp
            SAccPacket C0, C1, C2, C3;
            straits.initAcc(C0);
            straits.initAcc(C1);
            straits.initAcc(C2);
            straits.initAcc(C3);

            const Index spk   = (std::max)(1,SwappedTraits::LhsProgress/4);
            const Index endk  = (depth/spk)*spk;
            const Index endk4 = (depth/(spk*4))*(spk*4);

            Index k=0;
            for(; k<endk4; k+=4*spk)
            {
              SLhsPacket A0,A1;
              SRhsPacket B_0,B_1;

              straits.loadLhsUnaligned(blB+0*SwappedTraits::LhsProgress, A0);
              straits.loadLhsUnaligned(blB+1*SwappedTraits::LhsProgress, A1);

              straits.loadRhsQuad(blA+0*spk, B_0);
              straits.loadRhsQuad(blA+1*spk, B_1);
              straits.madd(A0,B_0,C0,B_0);
// ...
```

Where `madd`'s definition is 

```cpp
EIGEN_STRONG_INLINE void madd(const LhsPacketType& a, 
const RhsPacketType& b, AccPacketType& c, AccPacketType& tmp) const
```

If `SRhsPacket` and `SAccPacket` differ in the call to madd then it throws since it wants the same type here. But I think we can get around all of this lower level internal eigen stuff by going one step higher and specializing general product structs.

@t4c1 based on your knowledge of Eigen does that sound like the right path here?

SteveBronder:
I got a dumb version of this up and passing for `var` and `double`. Though I think I need some help on this because `matrix_d * vector_v` is way slower than `matrix_v * vector_d` and it feels like the opposite should be true. The same goes for `matrix_d * matrix_v` being like 3x slower than `matrix_v * matrix_d` so I have to be doing poor access or storage

https://github.com/stan-dev/math/compare/feature/eigen-specializations-products#diff-989932cf1f16ca6deb20162d5b764e7fR178

```bash
git pull
git checkout feature/eigen-specializations-products
./runTests.py ./test/unit/math/rev/fun/multiply_test.cpp
```

t4c1:
I spent some time looking into this, but to be honest I have no idea how to fix this.

stan/math/prim/fun/complex_schur_decompose.hpp
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @tparam value type of matrix
+ * @param m real matrix to decompose
+ * @return complex unitary matrix of the complex Schur decomposition of the
+ * specified matrix
+ * @see complex_schur_decompose_t
+ */
+template <typename T>
+Eigen::Matrix<complex_return_t<T>, -1, -1> complex_schur_decompose_u(
+    const Eigen::Matrix<T, -1, -1>& m) {
```
t4c1, line 29: I am spendina a lot of time generalizing matrix functions signatures (#1470). Can you also make new functions accept general Eigen expressions?

bob-carpenter:
I would really like to get what works here merged so that I don't have to maintain this as Stan quickly evolves around it.  I think what's here is correct.  It can be generalized for more Eigen operations (as @SteveBronder is suggesting) and for more general matrix inputs (as @t4c1 is suggesting), but I'd like to do that on a later branch.  

I'll just remove Schur decomposition altogether and remove the comments and hopefully we can merge.  

If @SteveBronder and @t4c1 would like this to be generalized before merging, I'd like to ask for help maintaining the branch.

bob-carpenter:
The other option here, seriously, is to just cut our losses on complex and give up and rip out all the other stuff.  It's been a huge pain, it's still not done, and I can't see the light at the end of the tunnel for when it will be done if it has to be fully general.  I haven't even tried FFT on it, which is what I really wanted to get into Stan.

t4c1:
> If @SteveBronder and @t4c1 would like this to be generalized before merging

If that is a problem, you (or me or someone else) can do that later in a separate PR later. 

> The other option here, seriously, is to just cut our losses on complex and give up

Complex stuff is mostly your work, so you should decide.

SteveBronder:
> I would really like to get what works here merged so that I don't have to maintain this as Stan quickly evolves around it. I think what's here is correct. It can be generalized for more Eigen operations (as @SteveBronder is suggesting) and for more general matrix inputs (as @t4c1 is suggesting), but I'd like to do that on a later branch.

The multiply stuff is also an issue for var so it should be put in a seperate PR. I'm going to also file an issue with Eigen to see what the best way to resolve this is. I think the better option here may be to just write `schur` as a stan function instead of relying on eigen. Once tests pass for this I'm fine approving it

bob-carpenter:
I fixed the fix by restricting isinf and isnan to autodiff like we did for other cmath functions.  Hopefully this'll get through Windows tests (there seems to be an extra top-level isnan function floating around in a version of clang++ there that doesn't show up elsewhere).

PR #1828, opened 2020-04-09
Generalize */fun starting with e-l

t4c1:
## Summary

Generalizes functions with names starting with letters e-l to accept general Eigen expressions.

## Tests
This is just a refactor. No new tests. Templating of `get_base1` was changed. As a side effect explicit template parameters are not necessary anymore. Tests has been appropriately modified.

## Side Effects
See tests.

## Release notes
Generalized functions with names starting with letters e-l to accept general Eigen expressions.

## Checklist

- [x] Math issue #1470

- [x] Copyright holder: Tadej Ciglarič

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1844, opened 2020-04-17
Generalize signatures for u to z

SteveBronder:
## Summary

This generalizes the signatures for prim/fwd/rev from u to z

## Tests

Refactor so no new tests

## Side Effects

Not yet, but here's a weird one, we don't call the rev specialization of `unit_vector_constrain` and if we do the tests fail. This happens because the header guard in rev is the same as the prim one, which happened about [4 years ago](https://github.com/stan-dev/math/blame/e661af11ff7a8ddfadb54f5b33bdd8b19ebd6e94/stan/math/rev/fun/unit_vector_constrain.hpp). We could fix it or delete it.

## Release notes

Generalize signatures for Eigen functions from U to Z.

## Checklist


bob-carpenter:
If there's a reverse-mode specialization for `unit_vector_constrain` that works with the way unit-vectors currently work, then we should use it.

SteveBronder:
> If there's a reverse-mode specialization for unit_vector_constrain that works with the way unit-vectors currently work, then we should use it.

@bgoodri can you look at the code there to see why that test is failing when the header guard is changed?

bgoodri:
The header guards should definitely be distinct. Which test is failing if we do that?

bob-carpenter:
@bgoodri fixed the transform and it looks like the reverse implementation matches the new definition's intent in that it has the normal density term `lp -= 0.5 * dot_self(y)` in the constraining transform.

By the way, that's better written as `lp += -0.5 * dot_self(y)` because `-0.5` will be interpreted as a `double` so you get an add and multiply instead of the less efificient subtract and multiply.

SteveBronder:
> The header guards should definitely be distinct. Which test is failing if we do that?

Changing the header and adding `unit_vector_constrain` to rev/fun.hpp causes

```
./runTests.py ./test/unit/math/mix/fun/unit_vector_constrain_test
```

to fail

SteveBronder:
@serban-nicusor-toptal do you know whatsup with the error I'm seeing here?

```
[LLVM/Clang] [-ERROR-] Can't determine head commit using 'git rev-parse'. Skipping blame.

[LLVM/Clang] [-ERROR-] hudson.plugins.git.GitException: Command "/usr/bin/git rev-parse HEAD^{commit}" returned status code 128:

stdout: 

stderr: fatal: Not a git repository (or any of the parent directories): .git
```

https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/downstream_tests/1509/pipeline

idt that would be something on my end. Also I think we should throw if the performance tests fail

stan/math/rev/fun/value_of.hpp
```diff
@@ -20,7 +20,16 @@ namespace math {
  * @param v Variable.
  * @return Value of variable.
  */
-inline double value_of(const var& v) { return v.vi_->val_; }
+template <typename Var, require_var_t<Var>* = nullptr>
```
t4c1, line 23: We don't need requires here. Previous implementation was fine and in my opinion easier to read.

stan/math/rev/fun/variance.hpp
```diff
@@ -38,12 +38,14 @@ inline var calc_variance(size_t size, const var* dtrs) {
  * @param[in] v a vector
  * @return sample variance of specified vector
  */
-inline var variance(const std::vector<var>& v) {
+template <typename StdVec, require_std_vector_t<StdVec>* = nullptr,
+          require_vt_var<StdVec>* = nullptr>
+inline auto variance(StdVec&& v) {
```
t4c1, line 43: why auto return type?

test/unit/math/prim/fun/value_of_test.cpp
```diff
@@ -40,30 +40,6 @@ TEST(MathMatrixPrimArr, value_of) {
...
-  EXPECT_FALSE((std::is_same<decltype(stan::math::value_of(a)),
-                             std::vector<double>>::value));
-  EXPECT_FALSE((std::is_same<decltype(stan::math::value_of(a)),
-                             std::vector<double>&>::value));
-  EXPECT_FALSE((std::is_same<decltype(stan::math::value_of(a)),
-                             const std::vector<double>>::value));
-  EXPECT_TRUE((std::is_same<decltype(stan::math::value_of(a)),
-                            const std::vector<double>&>::value));
-}
-
```
t4c1, line 66: You could modify these tests so that they work with new code instead of deleting them. You can check what I did in #1828, where I did similar changes to `value_of`.

test/unit/math/prim/fun/value_of_test.cpp
```diff
@@ -86,63 +62,3 @@ TEST(MathMatrixPrimMat, value_of) {
...
-  Eigen::Matrix<double, 5, 4> a;
-  EXPECT_FALSE((std::is_same<decltype(stan::math::value_of(a)),
-                             Eigen::Matrix<double, 5, 4>>::value));
-  EXPECT_FALSE((std::is_same<decltype(stan::math::value_of(a)),
-                             Eigen::Matrix<double, 5, 4>&>::value));
-  EXPECT_FALSE((std::is_same<decltype(stan::math::value_of(a)),
-                             const Eigen::Matrix<double, 5, 4>>::value));
-  EXPECT_TRUE((std::is_same<decltype(stan::math::value_of(a)),
-                            const Eigen::Matrix<double, 5, 4>&>::value));
-}
```
t4c1, line 148: You could modify these tests so that they work with new code instead of deleting them.

test/unit/math/rev/functor/adj_jac_apply_test.cpp
```diff
@@ -422,17 +422,8 @@ TEST(AgradRev, test_matrix_sin_multiple_jac) {
...
-      std::vector<double>, std::vector<int>, const std::vector<double>&,
-      const std::vector<int>&, Eigen::Matrix<double, Eigen::Dynamic, 1>,
-      Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>,
-      Eigen::Matrix<double, 2, Eigen::Dynamic>, Eigen::Matrix<double, 5, 1>,
-      const Eigen::Matrix<double, Eigen::Dynamic, 1>&,
-      const Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>&,
-      const Eigen::Matrix<double, 2, Eigen::Dynamic>&,
-      const Eigen::Matrix<double, 5, 1>&) {
+  template <size_t size, typename... Args>
+  Eigen::VectorXd operator()(std::array<bool, size> needs_adj, Args&&...) {
```
t4c1, line 426: I think explicitely specified arguments also serve as a kind of documentation, which types are tested.

bob-carpenter:
I agree with @t4c1 and would like to keep the code as simple as possible.  Specifically, 

* Only forward where necessary.  It isn't necessary for our autodiff types, is it?

* Only use requires where we need them.  If the arguments don't match the concept, it won't compile.  In general, I don't like the idea of repalacing a concrete data type with a template parameter plus requires if the set of matches is the same.

t4c1:
> Only use requires where we need them. If the arguments don't match the concept, it won't compile.

I disagree. I prefere requires to a template and a funtion that does not compile with wrong arguments. In such cases error messages usually have little or nothing in common with the fact that a function does not support certain types.

> In general, I don't like the idea of repalacing a concrete data type with a template parameter plus requires if the set of matches is the same.

+1 to that part. We only need to do that with Eigen types to make functions accept general expressions.

t4c1:
> It isn't necessary for our autodiff types, is it?

No, it is not. It is only needed for types that have move costructors/ move assigment operators. Usually those are the types that dynamically allocate memory, such as `std::vector` or `Eigen::Matrix`.

SteveBronder:
Yeah I went a little forward loco here, it was mostly from trying to get  `value_of` working right. I'll remove it where we have no need for forwarding

stan/math/fwd/fun/unit_vector_constrain.hpp
```diff
@@ -15,44 +15,51 @@
 namespace stan {
 namespace math {
 
-template <typename T, int R, int C>
-inline Eigen::Matrix<fvar<T>, R, C> unit_vector_constrain(
-    const Eigen::Matrix<fvar<T>, R, C>& y) {
+template <typename EigMat, require_eigen_vt<is_fvar, EigMat>* = nullptr>
+inline auto unit_vector_constrain(EigMat&& y) {
```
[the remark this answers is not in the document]
SteveBronder: For this one the function below is 

```cpp
template <typename EigMat, require_eigen_vt<is_fvar, EigMat>* = nullptr>
inline auto unit_vector_constrain(EigMat&& y, value_type_t<EigMat>& lp) {
  const value_type_t<EigMat> squared_norm = dot_self(y);
  lp -= 0.5 * squared_norm;
  return unit_vector_constrain(std::forward<EigMat>(y));
}
``

Can forward `y` to `unit_vector_constrain`.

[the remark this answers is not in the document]
Comment (SteveBronder) [reply stan/math/fwd/fun/unit_vector_constrain.hpp:19]: For the other one it can be `const&`

File: make/cpplint [hunk 7]
```diff
@@ -10,4 +10,4 @@ endif
 
 .PHONY: cpplint
 cpplint:
-	@$(PYTHON2) $(CPPLINT)/cpplint.py --output=vs7 --counting=detailed --root=stan --extension=hpp,cpp --filter=-runtime/indentation_namespace,-build/c++11,-readability/namespace,-legal/copyright,-whitespace/indent,-runtime/reference $(shell (find test/unit -name '*.hpp' -o -name '*.cpp') && (find stan -name '*.hpp' -o -name '*.cpp'))
+	@$(PYTHON2) $(CPPLINT)/cpplint.py --output=vs7 --counting=detailed --root=stan --extension=hpp,cpp --filter=-runtime/indentation_namespace,-build/c++11,-readability/namespace,-whitespace/operators,-legal/copyright,-whitespace/indent,-runtime/reference $(shell (find test/unit -name '*.hpp' -o -name '*.cpp') && (find stan -name '*.hpp' -o -name '*.cpp'))
```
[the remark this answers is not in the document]
SteveBronder: The version of cpplint we use doesn't recognize universal references in function signatures but I'm going to remove those anyway

stan/math/fwd/fun/unit_vector_constrain.hpp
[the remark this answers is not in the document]
t4c1: Once this one accepts const& the other one also does not need any forwarding. 

Forwarding an argument to a function only makes sense if the function either eccepts a forwarding reference or has separate overloads for const lvalue ref and rvalue ref.

stan/math/prim/fun/value_of_rec.hpp
```diff
@@ -65,8 +53,10 @@ inline std::complex<double> value_of_rec(const std::complex<T>& x) {
  * @param[in] x std::vector to be converted
  * @return std::vector of values
  **/
-template <typename T>
-inline std::vector<double> value_of_rec(const std::vector<T>& x) {
+template <typename Vec, require_std_vector_t<Vec>* = nullptr,
+          require_not_vt_arithmetic<Vec>* = nullptr,
+          require_not_vt_var<Vec>* = nullptr>
```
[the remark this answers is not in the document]
SteveBronder: I changed the `arithmetic` to `floating_point` but we have a custom specialization for `var` so I left that in

stan/math/prim/prob/neg_binomial_2_log_glm_lpmf.hpp
```diff
@@ -96,15 +96,15 @@ neg_binomial_2_log_glm_lpmf(
   check_positive_finite(function, "Precision parameter", phi);
 
   if (size_zero(y, phi)) {
-    return 0;
+    return 0.0;
```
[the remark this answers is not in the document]
SteveBronder: @bob-carpenter Why is that? `0` is going to be an int type right?

stan/math/rev/fun/trace_quad_form.hpp
```diff
@@ -15,76 +15,79 @@
...
-      : A_(A), B_(B) {}
+  trace_quad_form_vari_alloc(EigMatL&& A, EigMatR&& B)
+      : A_(std::forward<EigMatL>(A)), B_(std::forward<EigMatR>(B)) {}
 
   double compute() { return trace_quad_form(value_of(A_), value_of(B_)); }
 
-  Eigen::Matrix<Ta, Ra, Ca> A_;
-  Eigen::Matrix<Tb, Rb, Cb> B_;
+  typename std::decay_t<EigMatL>::PlainObject A_;
+  typename std::decay_t<EigMatR>::PlainObject B_;
```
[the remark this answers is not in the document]
SteveBronder: These get pumped into `trace_quad_form` during construction which uses B twice. I figure it's probably just better to set them here

stan/math/rev/fun/variance.hpp
SteveBronder: I like using auto but I'll switch these to var

test/unit/math/rev/functor/adj_jac_apply_test.cpp
SteveBronder: Yeah that's fair I switched all these back

stan/math/fwd/fun/unit_vector_constrain.hpp
```diff
@@ -17,34 +17,24 @@ namespace math {
...
-  partial_mat unit_vector_y_val = unit_vector_constrain(y_val);
-  ref_inner unit_vector_y(y.size());
-  unit_vector_y.val() = unit_vector_y_val;
-
-  eig_partial squared_norm = dot_self(y_val);
-  eig_partial norm = sqrt(squared_norm);
-  eig_partial inv_norm = inv(norm);
+  auto squared_norm = dot_self(y_val);
+  auto norm = sqrt(squared_norm);
+  auto inv_norm = inv(norm);
```
t4c1, line 31: Why auto? You can use `eig_partial`

stan/math/prim/fun/trace_gen_quad_form.hpp
```diff
@@ -63,12 +62,12 @@ inline auto trace_gen_quad_form(EigMatD&& D, EigMatA&& A, EigMatB&& B) {
 template <typename EigMatD, typename EigMatA, typename EigMatB,
           require_all_eigen_vt<std::is_arithmetic, EigMatD, EigMatA,
                                EigMatB>* = nullptr>
-inline double trace_gen_quad_form(EigMatD&& D, EigMatA&& A, EigMatB&& B) {
+inline auto trace_gen_quad_form(const EigMatD& D, const EigMatA& A, const EigMatB& B) {
```
t4c1: Why would be auto better than double?

SteveBronder: Mostly overthinking. If the static matrix stuff goes through one day we'll be able to have floats or longs here but yeah I can change it back

t4c1:
@SteveBronder How is this going? If you are too busy with static matrix stuff I can take over.

SteveBronder:
> @SteveBronder How is this going? If you are too busy with static matrix stuff I can take over.

Hey sorry yes got distracted by static stuff. Let me make one more push on this tonight

rok-cesnovar (commented):
I am not reviewing this one, just noting that this should touch https://github.com/stan-dev/math/blob/develop/test/expressions/stan_math_sigs_exceptions.expected to remove the signatures for which expression inputs now work.

SteveBronder:
@t4c1 can you give this a look when you have a minute?

stan/math/fwd/fun/unit_vector_constrain.hpp
```diff
@@ -15,42 +15,37 @@
...
-  T squared_norm = dot_self(y_t);
-  T norm = sqrt(squared_norm);
-  T inv_norm = inv(norm);
-  Matrix<T, Eigen::Dynamic, Eigen::Dynamic> J
-      = divide(tcrossprod(y_t), -norm * squared_norm);
-
-  for (int m = 0; m < y.size(); ++m) {
+template <typename EigMat, require_eigen_vt<is_fvar, EigMat>* = nullptr>
+inline auto unit_vector_constrain(const EigMat& y) {
+  using eig_mat = std::decay_t<EigMat>;
```
t4c1, line 20: This decay is redundant. As the argument is declared `const EigMat& y`, `EigMat` can not be deduced as either const or reference.

stan/math/fwd/fun/unit_vector_constrain.hpp
```diff
@@ -15,42 +15,37 @@
...
+  using eigen_type = typename eig_mat::PlainObject;
```
t4c1, line 21:
```suggestion
  using T_plain = typename eig_mat::PlainObject;
```

I find the name `eigen_type` a bit confusing.

stan/math/fwd/fun/unit_vector_constrain.hpp
```diff
@@ -15,42 +15,37 @@
...
+  using eig_partial = partials_type_t<value_type_t<EigMat>>;
+  using partial_mat = Eigen::Matrix<eig_partial, eig_mat::RowsAtCompileTime,
+                                    eig_mat::ColsAtCompileTime>;
```
t4c1, line 24:
```suggestion
  using partial_mat = promote_scalar_t<eig_partial, EigMat>;
```

stan/math/prim/fun/fill.hpp
```diff
@@ -47,15 +48,15 @@ void fill(T& x, const S& y) {
...
  * @param[in, out] y Value.
  */
-template <typename T, typename S>
-void fill(std::vector<T>& x, const S& y) {
-  for (typename std::vector<T>::size_type i = 0; i < x.size(); ++i) {
-    fill(x[i], y);
+template <typename Vec, typename S, require_std_vector_t<Vec>* = nullptr>
+inline void fill(Vec& x, S&& y) {
+  for (auto& x_val : x) {
+    fill(x_val, std::forward<S>(y));
```
t4c1, line 59: You can only forward a variable once!

stan/math/prim/fun/unit_vector_constrain.hpp
```diff
@@ -17,44 +17,43 @@ namespace math {
...
+          require_not_vt_fvar<EigMat>* = nullptr>
+auto unit_vector_constrain(const EigMat& y) {
   using std::sqrt;
   check_vector("unit_vector_constrain", "y", y);
   check_nonzero_size("unit_vector_constrain", "y", y);
-  T SN = dot_self(y);
+  value_type_t<EigMat> SN = dot_self(y);
   check_positive_finite("unit_vector_constrain", "norm", SN);
-  return y / sqrt(SN);
+  return (y / sqrt(SN)).eval();
```
t4c1, line 33: You will ned to use `to_ref` on `y`.

stan/math/prim/fun/unit_vector_constrain.hpp
```diff
@@ -17,44 +17,43 @@ namespace math {
...
+auto unit_vector_constrain(const EigMat& y, LP& lp) {
   using std::sqrt;
   check_vector("unit_vector_constrain", "y", y);
   check_nonzero_size("unit_vector_constrain", "y", y);
-  T SN = dot_self(y);
+  value_type_t<EigMat> SN = dot_self(y);
   check_positive_finite("unit_vector_constrain", "norm", SN);
   lp -= 0.5 * SN;
-  return y / sqrt(SN);
+  return (y / sqrt(SN)).eval();
```
t4c1, line 56: Same here.

stan/math/prim/fun/unit_vector_free.hpp
```diff
@@ -18,11 +18,10 @@ namespace math {
  * @param x unit vector of dimension K
  * @return Unit vector of dimension K considered "free"
  */
-template <typename T>
-Eigen::Matrix<T, Eigen::Dynamic, 1> unit_vector_free(
-    const Eigen::Matrix<T, Eigen::Dynamic, 1>& x) {
+template <typename EigVec, require_eigen_vector_t<EigVec>* = nullptr>
+decltype(auto) unit_vector_free(EigVec&& x) {
```
t4c1, line 22:
```suggestion
EigVec unit_vector_free(EigVec&& x) {
```
No need for `decltype(auto)` when we know the return type.

stan/math/prim/fun/variance.hpp
```diff
@@ -37,28 +39,25 @@ inline return_type_t<T> variance(const std::vector<T>& v) {
...
  *
  * @param m matrix
  * @return sample variance of coefficients
  * @throw <code>std::invalid_argument</code> if the matrix has size zero
  */
-template <typename T, int R, int C>
-inline return_type_t<T> variance(const Eigen::Matrix<T, R, C>& m) {
+template <typename EigMat, require_eigen_t<EigMat>* = nullptr,
+          require_not_vt_var<EigMat>* = nullptr>
+inline value_type_t<EigMat> variance(EigMat&& m) {
```
t4c1, line 51:
```suggestion
inline value_type_t<EigMat> variance(const EigMat& m) {
```

stan/math/prim/fun/variance.hpp
```diff
@@ -37,28 +39,25 @@ inline return_type_t<T> variance(const std::vector<T>& v) {
...
+  using eig_value = value_type_t<EigMat>;
+  using ref_inner = const typename std::decay_t<EigMat>::PlainObject;
```
t4c1, line 53:
```suggestion
  using T_plain = const plain_type_t<EigMat>;
```
I prefere naming variables by what they are, not how they will be used.

stan/math/prim/fun/variance.hpp
```diff
@@ -13,21 +13,23 @@ namespace math {
...
+ * @tparam StdVec A standard library vector that does not contain a var.
  * @param v specified vector
  * @return sample variance of vector
  * @throw <code>std::invalid_argument</code> if the vector has size zero
  */
-template <typename T>
-inline return_type_t<T> variance(const std::vector<T>& v) {
+template <typename StdVec, require_std_vector_t<StdVec>* = nullptr,
+          require_not_vt_var<StdVec>* = nullptr>
+inline value_type_t<StdVec> variance(StdVec&& v) {
```
t4c1, line 23: Would it not be simpler to just wrap the argument in `Eigen::Map` and call Eigen overload?

stan/math/prim/fun/vec_concat.hpp
```diff
@@ -7,31 +7,55 @@
...
+/**
+ * Get the internal sizes of a pack of vectors.
+ * @tparam Vec Type of first vector to count.
+ * @tparam VecArgs Parameter pack of vector types to accumulate sizes of
+ * @param x vector to start accumulation of sizes with.
+ * @param args pack of vectors to accumulate sizes for.
+ */
+template <typename Vec, typename... VecArgs>
+inline auto sum_vector_sizes(Vec&& x, VecArgs&&... args) {
+  return x.size() + sum_vector_sizes(std::forward<VecArgs>(args)...);
```
t4c1, line 21:
```suggestion
inline auto sum_vector_sizes(const Vec& x, const VecArgs&... args) {
  return x.size() + sum_vector_sizes(args...);
```

No need for forwarding here.

stan/math/prim/fun/vec_concat.hpp
```diff
@@ -7,31 +7,55 @@
...
+ * @param y First vector to insert into the in/out vector.
+ * @param args Pack of other vectors to fill in the in/out vector.
  */
-template <typename T>
-inline const std::vector<T>& vec_concat(const std::vector<T>& v1) {
-  return v1;
+template <typename VecInOut, typename VecIn, typename... VecArgs>
+inline void append_vectors(VecInOut& x, VecIn&& y, VecArgs&&... args) {
+  x.insert(x.end(), y.begin(), y.end());
+  append_vectors(x, std::forward<VecArgs>(args)...);
```
t4c1, line 40:
```suggestion
inline void append_vectors(VecInOut& x, const VecIn& y, const VecArgs&... args) {
  x.insert(x.end(), y.begin(), y.end());
  append_vectors(x, args...);
```

stan/math/prim/fun/vec_concat.hpp
```diff
@@ -7,31 +7,55 @@
...
-inline const std::vector<T> vec_concat(const std::vector<T>& v1,
-                                       const Args... args) {
-  std::vector<T> vec = vec_concat(args...);
-  vec.insert(vec.end(), v1.begin(), v1.end());
+template <typename Vec, typename... Args>
+inline auto vec_concat(Vec&& v1, Args&&... args) {
+  std::vector<value_type_t<Vec>> vec;
+  vec.reserve(internal::sum_vector_sizes(v1, args...));
+  internal::append_vectors(vec, std::forward<Vec>(v1),
+                           std::forward<Args>(args)...);
```
t4c1, line 58: Why do you need to change this implementation? As I see it does the same. If there is a good argument, remove the forwarding here too.

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -47,51 +47,54 @@ class unit_vector_elt_vari : public vari {
...
+ * @tparam EigMat type inheriting from `EigenBase` that has a `var`
+ *  scalar type.
  * @param y vector of K unrestricted variables
  * @return Unit length vector of dimension K
  **/
-template <int R, int C>
-Eigen::Matrix<var, R, C> unit_vector_constrain(
-    const Eigen::Matrix<var, R, C>& y) {
+template <typename EigMat, require_eigen_vt<is_var, EigMat>* = nullptr>
+auto unit_vector_constrain(EigMat&& y) {
```
t4c1, line 56:
```suggestion
auto unit_vector_constrain(const EigMat& y) {
```

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -47,51 +47,54 @@ class unit_vector_elt_vari : public vari {
...
+  using ref_inner = const typename std::decay_t<EigMat>::PlainObject;
```
t4c1, line 57:
```suggestion
  using T_plain = const plain_type_t<EigMat>;
```

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -47,51 +47,54 @@ class unit_vector_elt_vari : public vari {
...
   check_vector("unit_vector", "y", y);
   check_nonzero_size("unit_vector", "y", y);
-  vector_d y_d = y.val();
+  const Eigen::Ref<ref_inner, Eigen::Aligned16, Eigen::Stride<0, 0>>& y_mat = y;
```
t4c1, line 60: How can you be sure about the alignment? Also why do you need to explicitely set stride? Are not 0,0 default values anyway?

Why not just use `ref_type_t` or `to_ref`?

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -47,51 +47,54 @@ class unit_vector_elt_vari : public vari {
...
-
   check_positive_finite("unit_vector", "norm", norm);
   Eigen::Map<vector_d> unit_vecd(unit_vector_y_d_array, y.size());
   unit_vecd = y_d / norm;
 
-  Eigen::Matrix<var, R, C> unit_vector_y(y.size());
-  for (int k = 0; k < y.size(); ++k)
+  Eigen::Matrix<var, std::decay_t<EigMat>::RowsAtCompileTime,
+                std::decay_t<EigMat>::ColsAtCompileTime>
+      unit_vector_y(y.size());
```
t4c1, line 77:
```suggestion
  T_plain unit_vector_y(y.size());
```

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -47,51 +47,54 @@ class unit_vector_elt_vari : public vari {
...
  * @param y vector of K unrestricted variables
  * @return Unit length vector of dimension K
  * @param lp Log probability reference to increment.
  **/
-template <int R, int C>
-Eigen::Matrix<var, R, C> unit_vector_constrain(
-    const Eigen::Matrix<var, R, C>& y, var& lp) {
-  Eigen::Matrix<var, R, C> x = unit_vector_constrain(y);
+template <typename EigMat, require_eigen_vt<is_var, EigMat>* = nullptr>
+auto unit_vector_constrain(EigMat&& y, var& lp) {
```
t4c1, line 96:
```suggestion
auto unit_vector_constrain(const EigMat& y, var& lp) {
```

stan/math/rev/fun/variance.hpp
```diff
@@ -36,34 +36,39 @@ inline var calc_variance(size_t size, const var* dtrs) {
  * Return the sample variance of the specified standard
  * vector.  Raise domain error if size is not greater than zero.
  *
+ * @tparam StdVec A standard vector with a `var` value type
  * @param[in] v a vector
  * @return sample variance of specified vector
  */
-inline var variance(const std::vector<var>& v) {
+template <typename StdVec, require_std_vector_vt<is_var, StdVec>* = nullptr>
+inline var variance(StdVec&& v) {
```
t4c1, line 44:
```suggestion
inline var variance(const std::vector<var>& v) {
```

stan/math/rev/fun/variance.hpp
```diff
@@ -36,34 +36,39 @@ inline var calc_variance(size_t size, const var* dtrs) {
...
- * @tparam C number of columns, can be Eigen::Dynamic
+ * @tparam EigMat type inheriting from `EigenBase` that has a `var`
+ *  scalar type.
  * @param[in] m input matrix
  * @return sample variance of specified matrix
  */
-template <int R, int C>
-var variance(const Eigen::Matrix<var, R, C>& m) {
+template <typename EigMat, require_eigen_vt<is_var, EigMat>* = nullptr>
+var variance(EigMat&& m) {
```
t4c1, line 63:
```suggestion
var variance(const EigMat& m) {
```

stan/math/rev/fun/variance.hpp
```diff
@@ -36,34 +36,39 @@ inline var calc_variance(size_t size, const var* dtrs) {
...
+  using ref_inner = const typename std::decay_t<EigMat>::PlainObject;
```
t4c1, line 64:
```suggestion
  using T_plain = const plain_type_t<EigMat>;
```

stan/math/rev/fun/variance.hpp
```diff
@@ -36,34 +36,39 @@ inline var calc_variance(size_t size, const var* dtrs) {
...
   check_nonzero_size("variance", "m", m);
   if (m.size() == 1) {
-    return 0;
+    return var{0.0};
   }
-  return internal::calc_variance(m.size(), &m(0));
+
+  const Eigen::Ref<ref_inner, Eigen::Aligned16, Eigen::Stride<0, 0>>& mat = m;
```
t4c1, line 70:
```suggestion
  ref_type_t<EigMat> mat = m;
```

t4c1 (changes requested):
Oh, the ancient PR is alive! ;)

All the functions you changed should be tested by expression testing framework.

Can you please stop pushing perfect forwarding everywhere?

stan/math/fwd/fun/unit_vector_constrain.hpp
SteveBronder: We have much better type traits now so I'll clean this whole thing up

stan/math/prim/fun/fill.hpp
SteveBronder: Yikes!

stan/math/prim/fun/unit_vector_free.hpp
SteveBronder: One caveat I wanna check, if we return by value here and assign this to a `const auto&` is that legal? Also, micro-optim so not a great argument, but `decltype(auto)` can save a construction (though it's a move construction so not really a big deal)

https://godbolt.org/z/YWEoEx

stan/math/prim/fun/variance.hpp
SteveBronder: Yes!

stan/math/prim/fun/vec_concat.hpp
SteveBronder: > Why do you need to change this implementation? 

It doesn't need to be changed, but this one will be just a hair better since we allocate all the space up-front instead of using `push_back()` and resizing.

> As I see it does the same. If there is a good argument, remove the forwarding here too.

Yeah I'll change this, I didn't read the docs for `vector.insert()` which doesn't do anything with forwarding references

stan/math/rev/fun/unit_vector_constrain.hpp
SteveBronder: Yes i wrote this before we had `to_ref` (which is soooo much easier)

SteveBronder:
> Oh, the ancient PR is alive! ;)

Lol yes I think at one point I was like, "Let me ping you after I read this over" and then it was lost to the sands of time

> All the functions you changed should be tested by expression testing framework.

Reading that over how do I add internal functions that are not directly callable in stan like `unit_vector_constrain`? I removed the ones from the exceptions list like variance. 

Can you please stop pushing perfect forwarding everywhere?

Yes, now that I've dived harder into Stan's memory pattern they are really not useful for var types (for anyone reading since vars are just pointers to vari copying a var is the same as a move). Though they are useful when working with floating point matrices!

stan/math/prim/fun/unit_vector_free.hpp
t4c1: Yes it is legal. Saving a construction can also make a seemingly innocent code wrong. You actually have an example in the linked godbolt:

```
	return decltype(auto) from rvalue to auto&&:
		Default Constructor
		Destruct
```
There was no copy or move constructor call - The result got destructed!

t4c1 (changes requested):
Did you just generalize whole prim/err? 

> Reading that over how do I add internal functions that are not directly callable in stan like unit_vector_constrain? I removed the ones from the exceptions list like variance.

Those in general don't actually need to accept Eigen matrices. You can either write tests yourself or extend the framework adding a list of non-exposed signatures you want tested.

SteveBronder:
> Did you just generalize whole prim/err?

Yeah sorry is that alright? I was getting errors up in the stan lib that those functions didn't accept `Eigen::Map<>` types since some of them are used in reader and I'd like reader to work over maps

> Those in general don't actually need to accept Eigen matrices. You can either write tests yourself or extend the framework adding a list of non-exposed signatures you want tested.

Yeah I can do that

stan/math/prim/err/is_column_index.hpp
```diff
@@ -11,15 +11,16 @@ namespace math {
...
  * @param y matrix to test
  * @param i column index to check
  * @return <code>true</code> if column index is in bounds
  */
-template <typename EigMat, require_eigen_matrix_t<EigMat>* = nullptr>
+template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
 inline bool is_column_index(const EigMat& y, size_t i) {
   return i >= stan::error_index::value
-         && i < static_cast<size_t>(y.cols()) + stan::error_index::value;
+         && i < static_cast<size_t>(to_ref(y).cols())
```
t4c1, line 22: You are making a copy just to see how many columns it has. Why?

stan/math/prim/err/is_lower_triangular.hpp
```diff
@@ -18,7 +20,8 @@ namespace math {
  */
 template <typename EigMat, require_eigen_matrix_t<EigMat>* = nullptr>
 inline bool is_lower_triangular(const EigMat& y) {
-  return y.unaryExpr([](auto&& x) { return std::isnan(x) ? 1.0 : x; })
+  return to_ref(y)
```
t4c1, line 23: Why to ref? y is used just once.

stan/math/prim/err/is_mat_finite.hpp
```diff
@@ -17,7 +18,7 @@ namespace math {
  **/
 template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
 inline bool is_mat_finite(const EigMat& y) {
-  return y.allFinite();
+  return to_ref(y).allFinite();
```
t4c1, line 21: Just why?

stan/math/prim/err/is_square.hpp
```diff
@@ -17,7 +17,8 @@ namespace math {
  */
 template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
 inline bool is_square(const EigMat& y) {
-  return is_size_match(y.rows(), y.cols());
+  const auto& y_ref = to_ref(y);
+  return is_size_match(y_ref.rows(), y_ref.cols());
```
t4c1, line 21: Please just revert all to_ref calls where the variable is used just once or only its size is needed.

stan/math/prim/fun/eval.hpp
```diff
@@ -17,8 +17,20 @@ namespace math {
...
+template <typename T,
+          require_same_t<std::decay_t<T>, plain_type_t<T>>* = nullptr>
+inline const T& eval(const T& arg) {
+  return arg;
+}
+
+template <typename T,
+          require_same_t<std::decay_t<T>, plain_type_t<T>>* = nullptr>
+inline T& eval(T& arg) {
+  return arg;
```
t4c1, line 33: Why did you feel you need to write these out? You could just replace `decltype(auto)` with `T`.

stan/math/prim/fun/sum.hpp
```diff
@@ -18,7 +18,7 @@ namespace math {
  * @return Same value (the sum of one value).
  */
 template <typename T, require_stan_scalar_t<T>* = nullptr>
-inline decltype(auto) sum(T&& m) {
+inline auto sum(T&& m) {
```
t4c1, line 21:
```suggestion
inline T sum(T&& m) {
```

stan/math/prim/fun/unit_vector_free.hpp
```diff
@@ -20,9 +21,11 @@ namespace math {
  * @return Unit vector of dimension K considered "free"
  */
 template <typename EigVec, require_eigen_vector_t<EigVec>* = nullptr>
-inline decltype(auto) unit_vector_free(EigVec&& x) {
-  check_unit_vector("stan::math::unit_vector_free", "Unit vector variable", x);
-  return std::forward<EigVec>(x);
+inline auto unit_vector_free(EigVec&& x) {
+  const auto& x_ref = to_ref(x);
```
t4c1, line 25:
```suggestion
  auto&& x_ref = to_ref(std::forward<EigVec>(x));
```

stan/math/prim/fun/value_of.hpp
```diff
@@ -18,8 +18,18 @@ namespace math {
...
+}
+
+template <typename T, require_st_arithmetic<T>* = nullptr>
+inline T& value_of(T& x) {
+  return x;
+}
+
+template <typename T, require_st_arithmetic<T>* = nullptr>
+inline const T& value_of(const T& x) {
+  return x;
```
t4c1, line 32: Again no need to write all these out. Replacing `decltype(auto)` with `T` is enough.

test/expressions/generateExpressionTests.py
```diff
@@ -79,6 +79,42 @@ def parse_signature_file(sig_file):
         res.append(signature)
     return res
 
+def add_extra_signatures(res):
+    """
+    """
```
t4c1, line 84: decumentation is missing

test/expressions/generateExpressionTests.py
```diff
@@ -79,6 +79,42 @@ def parse_signature_file(sig_file):
...
+    res.append("int is_matching_dims(vector, row_vector)")
+    res.append("int is_matching_dims(row_vector, vector)")
+    res.append("int is_matching_dims(vector, vector)")
+    res.append("int is_pos_definite(matrix)")
+    res.append("int is_square(matrix)")
+    res.append("int is_square(vector)")
+    res.append("int is_square(row_vector)")
+    res.append("int is_symmetric(matrix)")
+    res.append("int is_unit_vector(vector)")
+    res.append("int is_unit_vector(row_matrix)")
```
t4c1, line 116: I would prefer to add all these in a single call. Something like:
```suggestion
    res.extend([
      "int is_unit_vector(vector)",
      "int is_unit_vector(row_matrix)",
    ...
    ])
```

test/expressions/generateExpressionTests.py
```diff
@@ -100,14 +136,14 @@ def get_signatures():
         universal_newlines=True,
         shell=True,
     )
-
+    #import pdb; pdb.set_trace()
```
t4c1, line 139: remove commented code

test/expressions/generateExpressionTests.py
```diff
@@ -139,10 +175,13 @@ def parse_signature(signature):
     "lkj_corr_lpdf": [1, None],
     "lkj_corr_log": [1, None],
     "log_diff_exp": [3, None],
-    "log_inv_logit_diff": [1.2, 0.4],  
+    "log_inv_logit_diff": [1.2, 0.4],
     "multinomial_log" : [None, 1],
     "multinomial_lpmf" : [None, 1],
     "multinomial_rng" : [1, None, None],
+    "log_inv_logit_diff": [1.2, 0.4],
```
t4c1, line 182: This line is duplicated on line 178.

t4c1 (changes requested):

t4c1:
@SteveBronder Can we try getting this in for the release?

SteveBronder:
Yes def I'l clean this up today!

stan/math/prim/err/is_column_index.hpp
SteveBronder: Ah good catch, I'll give all the similar error functions a look over for this as well

SteveBronder: One thing I'm also thinking about. I'd guess 9/10 times someone passing an expression to the error checking functions they forgot to use a `to_ref()` in the higher level code. We could have a static_assert in these like

```
static_assert(!is_plain_type<T>::value, "Eigen Expression passed to {error_function_name} but eigen expressions must be evaluated at the call site before calling error functions.");

File: stan/math/prim/fun/eval.hpp (hunk 39 above)
Comment (SteveBronder) [reply stan/math/prim/fun/eval.hpp:33]: Ah the copy ellision thing was intuitive to me so I wrote it by hand but I think I got this now so I'll fix these up

https://godbolt.org/z/a3TMPW

File: stan/math/prim/fun/unit_vector_free.hpp (hunk 41 above)
Comment (SteveBronder) [reply stan/math/prim/fun/unit_vector_free.hpp:25]: On a side note, should we be using `auto&&` instead of `const auto&` when we do to_ref()? tmk `auto&&` evaluates to any reference type. Though I'm indifferent about it

https://godbolt.org/z/aYG5na

File: stan/math/prim/err/is_column_index.hpp (hunk 35 above)
Comment (t4c1) [reply stan/math/prim/err/is_column_index.hpp:22]: That should go into its own PR. Also check should accept trivial expressions, not just plain types.

Comment (t4c1) [reply stan/math/prim/err/is_column_index.hpp:22]: This one is not fixed yet.

File: stan/math/prim/fun/unit_vector_free.hpp (hunk 41 above)
Comment (t4c1) [reply stan/math/prim/fun/unit_vector_free.hpp]: Actually it does not matter, unless you need the reference to be non-const. I am just more used to const&.

File: test/expressions/generateExpressionTests.py [hunk 47]
```diff
@@ -76,9 +76,47 @@ def parse_signature_file(sig_file):
         if not signature.endswith(")\n"):
             part_sig = signature
             continue
-        res.append(signature)
+        signature)
```
t4c1, line 79: Typo?

test/expressions/generateExpressionTests.py
```diff
@@ -250,8 +250,8 @@ def handle_function_list(functions_input, signatures):
         if "." in f or "/" in f or "\\" in f:
             functions_input.extend(parse_signature_file(open(f)))
         elif " " in f:
-            function_signatures.append(f)
-            signatures.append(f)
+            function_signatuf)
+            signatuf)
```
t4c1, line 254: ... or runaway replace all?

test/expressions/generateExpressionTests.py
```diff
@@ -198,21 +237,21 @@ def save_tests_in_files(N_files, tests):
...
     function_names = []
     function_signatures = []
     for f in functions_input:
         if "." in f or "/" in f or "\\" in f:
             functions_input.extend(parse_signature_file(open(f)))
         elif " " in f:
-            function_signatures.append(f)
-            signatures.append(f)
+            function_signatuf)
+            signatuf)
```
t4c1, line 254: ... or runaway replace all?

t4c1 (changes requested):

test/expressions/generateExpressionTests.py
SteveBronder: Whups, yeah this is a bad find-replace my b

stan/math/prim/err/is_mat_finite.hpp
```diff
@@ -11,14 +12,13 @@ namespace math {
...
+ * @tparam EigMat A type derived from `EigenBase`
  * @param y Matrix to test
  * @return <code>true</code> if the matrix is finite
  **/
-template <typename T, int R, int C>
-inline bool is_mat_finite(const Eigen::Matrix<T, R, C>& y) {
-  return y.allFinite();
+template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
+inline bool is_mat_finite(const EigMat& y) {
+  return to_ref(y).allFinite();
```
SteveBronder, line 21: Really odd, if i don't put a `to_ref` here the expression test fails?

t4c1: Hmm, I looked over Eigen source and it is implemented as:
```
#if EIGEN_COMP_MSVC || (defined __FAST_MATH__)
  return derived().array().isFinite().all();
#else
  return !((derived()-derived()).hasNaN());
#endif
```
Some more info here: https://eigen.tuxfamily.org/bz/show_bug.cgi?id=1008

I think we should call `y.array().isFinite().all()` directly to avoid multiple evaluations. It seems that in Eigen they rely on the compiler to optimize dual evaluations. That might work with primitive scalars, but operations on vars have side effects, so I'm pretty sure they would not be optimized.

t4c1 (changes requested):
Also don't forget to fix the unresolved thing above.

stan/math/prim/err/is_column_index.hpp
SteveBronder: true my b

stan/math/prim/err/is_mat_finite.hpp
SteveBronder: Reading over that thread I'm not sure if the array method actually stops an evaluation. But I just put it up so we'll find out

t4c1 (approved):
Looks good, just fix the conflict.

t4c1 (approved):

PR #1841, opened 2020-04-15
Eigen member function for reading values and derivatives from autodiff matrices

andrjohns:
## Summary 
Adds a member function for more efficient extraction of values and derivatives from matrices of ```var```/```vari```/```fvar<T>```. The following functions are added:

- ```read_fvar()```
- ```read_vi_val_adj()```
- ```read_vi_val```
- ```read_vi_adj()```
- ```read_val_adj()```

And are used like so:
```
  matrix_vi vis(100, 100);
  MatrixXd vals(100, 100);
  MatrixXd adjs(100, 100);

var_matrix.read_vi_val_adj(vis, vals, adjs);
```
## Tests

Tests are added in ```rev```, ```fwd```, and ```mix``` for matrices, vectors, and row-vectors

## Side Effects

N/A

## Release notes

Added Eigen member function for more efficient extraction of values and derivatives from matrices of autodiff variables

## Checklist

- [x] Math issue #1745 

- [x] Copyright holder: Andrew Johnson

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1151, opened 2019-03-11
Always run full GPU unit tests after a few failures on merge to develop.

seantalts:
## Summary

We've seen now two or three failures on merging PRs to develop even though the PRs passed the PR tests due to interactions with the GPU subsystem. The GPU subsystem adds new overloads to existing functions, so this makes some sense. This PR adds the full GPU tests to the suite that must run on a PR before merge.

Does not require review as it's just a CI/CD configuration change, but will make sure the tests pass first.

- [x] Copyright holder: (fill in copyright holder information)

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses: Columbia University
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #1845, opened 2020-04-17
Generalize */fun from m to p

t4c1:
## Summary

Generalizes function with names starting with m to p to accept general Eigen expressions.

## Tests
Just a refactor. No new tests.

## Side Effects

None.

## Release notes

Generalized function with names starting with m to p to accept geenral Eigen expressions.

## Checklist

- [x] Math issue #1470


serban-nicusor-toptal:
Hey @SteveBronder 
The Jenkins log is very cryptic in this case.

Recently logic has been added to Jenkinsfiles that, if there are no source code changes a pipeline will not run entirely to save time and resources. For stan what is source code is defined [here](https://github.com/stan-dev/stan/blob/develop/Jenkinsfile#L172).

Looking at the logs you can see printed right above the errors `There aren't any differences in the source code, CI/CD will not run.`, the errors are happening because the log parsers don't get any input so they can't generate results in Jenkins. If you go to the stan build [#1151](https://jenkins.mc-stan.org/job/Stan/job/downstream_tests/1511/) relative to this math PR the build is actually a success since source code of stan wasn't touched in the process.

SteveBronder:
Could we have the performance tests always run? 

> Looking at the logs you can see printed right above the errors There aren't any differences in the source code, CI/CD will not run., the errors are happening because the log parsers don't get any input so they can't generate results in Jenkins. If you go to the stan build #1151 relative to this math PR the build is actually a success since source code of stan wasn't touched in the process.

Does that just check the current repo or also the sub repos? I'm pretty sure everything upstream should have gone off for that PR

serban-nicusor-toptal:
Strange, I've just tested a bit and couldn't get any git diff, tried with [--submodule diff](https://github.com/git/git/blob/master/Documentation/RelNotes/2.14.0.txt#L50) but couldn't get any diff.
I've made all downstream_tests run by default.

SteveBronder:
Nice thanks @serban-nicusor-toptal !!

stan/math/prim/fun/mdivide_left_spd.hpp
```diff
@@ -12,35 +12,36 @@ namespace math {
...
-  auto llt = Eigen::Matrix<return_type_t<T1, T2>, R1, C1>(A).llt();
+  auto llt
+      = Eigen::Matrix<return_type_t<EigMat1, EigMat2>,
+                      EigMat1::RowsAtCompileTime, EigMat1::ColsAtCompileTime>(A)
+            .llt();
   check_pos_definite(function, "A", llt);
-  return llt.solve(Eigen::Matrix<return_type_t<T1, T2>, R2, C2>(b));
+  return llt.solve(
+      Eigen::Matrix<return_type_t<EigMat1, EigMat2>, EigMat2::RowsAtCompileTime,
+                    EigMat2::ColsAtCompileTime>(b));
```
SteveBronder, line 44: I like how you used `constexpr` before for the eigen compile time rows and cols. How about this:

```cpp
template <typename EigMat1, typename EigMat2,
          require_all_eigen_t<EigMat1, EigMat2>* = nullptr>
inline auto mdivide_left_spd(const EigMat1& A, const EigMat2& b) {
  static const char* function = "mdivide_left_spd";
  constexpr Eigen::Index R1 = EigMat1::RowsAtCompileTime;
  constexpr Eigen::Index C1 = EigMat1::ColsAtCompileTime;
  constexpr Eigen::Index R2 = EigMat2::RowsAtCompileTime;
  constexpr Eigen::Index C2 = EigMat2::ColsAtCompileTime;
  using ret_value_t = return_type_t<EigMat1, EigMat2>;
  using eigen_return_t = Eigen::Matrix<ret_value_t, R1, C2>;
  check_multiplicable(function, "A", A, "b", b);
  check_symmetric(function, "A", A);
  check_not_nan(function, "A", A);
  if (A.size() == 0) {
    return eigen_return_t{0, b.cols()};
  }
  auto llt = Eigen::Matrix<ret_value_t, R1, C1>(A).llt();
  check_pos_definite(function, "A", llt);
  return llt.solve(Eigen::Matrix<ret_value_t, R1, C2>(b)).eval();
}
```

SteveBronder: If the auto and/or constexpr stuff makes sense to you then it applies to most of the mdivide stuff

stan/math/prim/fun/mean.hpp
```diff
@@ -10,36 +10,18 @@ namespace math {
...
+ * @return Sample mean of container coefficients.
  */
-template <typename T, int R, int C>
-inline return_type_t<T> mean(const Eigen::Matrix<T, R, C>& m) {
+template <typename T, require_container_t<T>* = nullptr>
+inline return_type_t<T> mean(const T& m) {
   check_nonzero_size("mean", "m", m);
-  return m.mean();
+  return apply_vector_unary<T>::reduce(m,
+                                       [](const auto& a) { return a.mean(); });
```
SteveBronder, line 24: [Q] is this not the kind of place we would want to forward along `m`?

stan/math/prim/fun/min.hpp
```diff
@@ -12,53 +12,25 @@ namespace math {
...
- */
-template <typename T, int R, int C>
-inline T min(const Eigen::Matrix<T, R, C>& m) {
-  if (m.size() == 0) {
+template <typename T, require_container_t<T>* = nullptr>
+inline value_type_t<T> min(const T& m) {
+  if (std::is_integral<value_type_t<T>>::value) {
+    check_nonzero_size("min", "int vector", m);
+  } else if (m.size() == 0) {
     return INFTY;
```
SteveBronder, line 30: [side note] Should this be negative infinity?

stan/math/prim/fun/ordered_constrain.hpp
```diff
@@ -14,27 +14,27 @@ namespace math {
...
+    const EigVec& x) {
   using Eigen::Dynamic;
   using Eigen::Matrix;
   using std::exp;
-  using size_type = index_type_t<Matrix<T, Dynamic, 1>>;
+  using size_type = Eigen::Index;
 
   size_type k = x.size();
-  Matrix<T, Dynamic, 1> y(k);
+  Eigen::Matrix<value_type_t<EigVec>, Eigen::Dynamic, 1> y(k);
```
SteveBronder, line 31: Could you just use the plainobject here
```suggestion
    typename EigVec::PlainObject y(k);
```

SteveBronder: Though I only think this matters if we wanted to support fixed size matrices

stan/math/prim/fun/ordered_free.hpp
```diff
@@ -23,23 +23,23 @@ namespace math {
...
   check_ordered("stan::math::ordered_free", "Ordered variable", y);
   using Eigen::Dynamic;
   using Eigen::Matrix;
   using std::log;
-  using size_type = index_type_t<Matrix<T, Dynamic, 1>>;
+  using size_type = Eigen::Index;
 
   size_type k = y.size();
-  Matrix<T, Dynamic, 1> x(k);
+  Eigen::Matrix<value_type_t<EigVec>, Eigen::Dynamic, 1> x(k);
```
SteveBronder, line 36: Same thing here about plain object and auto

stan/math/fwd/fun/mdivide_left_ldlt.hpp
```diff
@@ -15,29 +15,29 @@ namespace math {
...
-  Eigen::Matrix<T2, R2, C2> b_der(b.rows(), b.cols());
+  const Eigen::Ref<const plain_type_t<EigMat>>& b_ref = b;
+  Eigen::Matrix<T, R2, C2> b_val(b.rows(), b.cols());
+  Eigen::Matrix<T, R2, C2> b_der(b.rows(), b.cols());
   for (int j = 0; j < b.cols(); j++) {
     for (int i = 0; i < b.rows(); i++) {
-      b_val(i, j) = b(i, j).val_;
-      b_der(i, j) = b(i, j).d_;
+      b_val.coeffRef(i, j) = b_ref.coeff(i, j).val_;
+      b_der.coeffRef(i, j) = b_ref.coeff(i, j).d_;
```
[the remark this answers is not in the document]
t4c1: What you are suggesting is indeed equivalent code. However that internally runs two loops, wich is less efficient as it loads b_ref into cache twice. The solution to this dillema is comming in PR #1841

stan/math/prim/fun/matrix_power.hpp
```diff
@@ -21,9 +19,14 @@ namespace math {
...
+inline Eigen::Matrix<value_type_t<EigMat>, EigMat::RowsAtCompileTime,
+                     EigMat::ColsAtCompileTime>
+matrix_power(const EigMat& M, const int n) {
+  using T = value_type_t<EigMat>;
+  constexpr int R = EigMat::RowsAtCompileTime;
+  constexpr int C = EigMat::ColsAtCompileTime;
+
   check_square("matrix_power", "M", M);
   if (n < 0)
     invalid_argument("matrix_power", "n", n, "is ", ", but must be >= 0!");
```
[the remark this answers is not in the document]
t4c1: I think that is fine.

stan/math/prim/fun/mdivide_left_ldlt.hpp
```diff
@@ -13,29 +13,31 @@ namespace math {
...
+mdivide_left_ldlt(const LDLT_factor<T, R, C>& A, const EigMat& b) {
   check_multiplicable("mdivide_left_ldlt", "A", A, "b", b);
   if (A.cols() == 0) {
     return {0, b.cols()};
   }
 
-  return A.solve(Eigen::Matrix<return_type_t<T1, T2>, R2, C2>(b));
+  return A.solve(
+      Eigen::Matrix<return_type_t<T, EigMat>, EigMat::RowsAtCompileTime,
+                    EigMat::ColsAtCompileTime>(b));
```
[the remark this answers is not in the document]
t4c1: I dislike using auto for return types, because it hides what the return type is. That I have to search for return statements to determine return type. In this case that is not enough. Next I have to search what `A.solve(ret_mat(b)).eval();` returns or what `ret_mat` is. 

In general code if more often read than changed. That is why I think we should prefere readability to ease of implementation.

About RVO: In this functions we need to construct return matrix anyway. We gain nothing performance-wise if we construct it explicitely and can do RVO. Instead what I wrote does these conversions implicitely.

> Also having a bit of a hard time following the requires for this function.

That last require is needed not to use this implementation for what we have a fwd implementation. It lools confusing because we only have fwd implementation for the case of ` LDLT_factor == double` and `EigMat = fvar`. I checked it again and it is correct. I did simplify it the way you suggested.

stan/math/prim/fun/mdivide_left_spd.hpp
t4c1: I don't feel the need for those variables if they are only used once. See the comment above about auto.

stan/math/prim/fun/mean.hpp
t4c1: No. `apply_vector_unary<T>::reduce()` does not return `m` or a struct/object containing `m`. It also does not make a copy of `m` into local variable. So it can work with const ref as well.

Whenever a quistion is raised about whether perfect forwarding should be used for a certain argument try going trough what happens when a functions accepts the argument as const ref or as a forwarding ref. Do any copy or move constructor needs to be invoked when passed a lvalue/rvalue? If no copying of argument can be avoided by perfect forwarding than it is not needed.

stan/math/prim/fun/min.hpp
t4c1: No.

stan/math/prim/fun/ordered_constrain.hpp
t4c1: It is less code, which is good enough for me.

stan/math/prim/fun/ordered_free.hpp
t4c1: fixed

stan/math/prim/fun/mdivide_left_ldlt.hpp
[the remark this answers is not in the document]
SteveBronder: I'm fine with leaving it as is

> About RVO: In this functions we need to construct return matrix anyway. We gain nothing performance-wise if we construct it explicitely and can do RVO. Instead what I wrote does these conversions implicitely.

This was my bad because I forgot that if you have implicit constructors available and returning temps RVO can still happen. Yes the matrix is constructed in the function either way but RVO let's the memory it's assigned to be at the return callee and not copied (or moved via NRVO) right?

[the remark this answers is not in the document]
SteveBronder: > I dislike using auto for return types

On this though, I prefer them for some functions where we just use eigen functions since a lot of the time we can reuse that code for the sparse matrix stuff I'm doing right now. i.e. for add, mul, etc. those functions can currently take in and return sparse matrices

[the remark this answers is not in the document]
SteveBronder: Though a lot of these wouldn't work for sparse matrices so it's not an issue here

PR #1856, opened 2020-04-30
Allow nested gradients in reverse mode chain (Issue 1855)

### LINKED ISSUE #1855: Calling nested gradient in reverse mode sweep can cause segfaults
## Description

Calling nested gradients in the reverse mode sweep can cause segfaults.

This happens when ```ChainableStack::instance_->var_stack_``` needs to reallocate itself.

You can make the bug happen with this code (put it in test somewhere):
```
#include <gtest/gtest.h>
#include <test/unit/math/rev/fun/util.hpp>
#include <stan/math/rev/fun/sin.hpp>

namespace stan {
namespace math {

class test_vari : public vari {
 public:
  explicit test_vari() : vari(0.0) {
  }

  virtual void chain() {
    stan::math::nested_rev_autodiff nested;

    // Add enough vars to make the the var_stack_ vector reallocate                                                                                                                                                
    int N_new_vars = ChainableStack::instance_->var_stack_.capacity() + 1;

    var total = 0.0;
    for (int i = 0; i < N_new_vars; ++i) {
      total += i;
    }

    total.grad();
  }
};

}
}

TEST(AgradRev, grad_in_reverse_mode) {
  using stan::math::var;

  var total = 0.0;
  for (int i = 0; i < 2; ++i) {
    total += i;
  }

  var test_var(new stan::math::test_vari());

  test_var.grad();
}
```

#### Current Version:
v3.2.0

bbbales2:
Fixes #1855

## Tests

This adds a new test that will fail with the old code.

## Side Effects

Calling nested gradients in reverse mode works now

## Release notes

Fixed issue where doing nested gradients while there is an ongoing chain would cause segfaults.

## Checklist

- [x] Math issue #1855


wds15:
Ui... that's simple, but makes sense. Is there an effect on performance? Let's see (there shouldn't be one).

Do we need to adjust `set_zero_adjoints_*` as well? Maybe there are even other functions which should get this change which avoids invalidated iterators.

bbbales2:
Yeah good point. I'll scan the rest of core looking for iterators but if you think of anything else shout it out.

stan/math/rev/core/grad.hpp
```diff
@@ -28,20 +28,11 @@ namespace math {
...
-  using it_t = std::vector<vari*>::reverse_iterator;
   vi->init_dependent();
-  it_t begin = ChainableStack::instance_->var_stack_.rbegin();
-  it_t end = empty_nested() ? ChainableStack::instance_->var_stack_.rend()
-                            : begin + nested_size();
-  for (it_t it = begin; it < end; ++it) {
-    (*it)->chain();
+  size_t end = ChainableStack::instance_->var_stack_.size();
+  size_t begin = empty_nested() ? 0 : end - nested_size();
+  for (size_t i = end; i > begin; --i) {
```
bob-carpenter, line 34: There's an idiom for these loops counting downward with `size_t`:

```
for (size_t i = end; --i > begin; ) {
  foo(...var_stack[i]...)
}
```

It works in the case that `begin = 0` with `size_t` variables.

bbbales2: I switched back to this (that's how it was originally).

bbbales2:
> Do we need to adjust set_zero_adjoints_* as well?

I went looking around and core and didn't find anything.

Some things use iterators, but inside the loop they're not doing things that would invalidate the vectors.

stan/math/rev/core/grad.hpp
bbbales2: I undid this change.

Correct version was:
```
for (size_t i = end; i-- > begin; ) {
  foo(...var_stack[i]...)
}
```

And I figure if we're making mistakes with the difference between --i and i--, we should favor a different solution.

bob-carpenter:
I couldn't figure out how to respond inline without starting a review.  

> I figure if we're making mistakes with the difference between --i and i--, we should favor a different solution.

Iterating downward with `size_t` is tricky, which is why it's best to stick to the standard idioms.
The correct version posted just repeated the code I used.  Was that meant to be something else?

The key thing to avoid is this:

```
for (size_t i = K; i >= 0; --i)
```

It works with `int i`, but fails with `size_t i` because `i >= 0` always if `i` is of type `size_t`.

bbbales2:
> The correct version posted just repeated the code I used. Was that meant to be something else?

The code in your first comment was comment was `--i` (https://github.com/stan-dev/math/pull/1856#discussion_r418067937):
```
for (size_t i = end; --i > begin; )
```

Which is what was in the reference implementation in the code before these changes: https://github.com/stan-dev/math/pull/1856/files#diff-2a978e44b6171f19523f9c66a496d1bcL35

I guess in retrospect the real problem was I didn't have an explicit unit test for this. I'll add one to make sure grad gets called on every vari.

Doesn't really matter I'll change it back and add a unit test.

bbbales2:
> I couldn't figure out how to respond inline without starting a review.

Yeah Github comments get pretty confusing. Things hide sometimes, but don't hide sometimes, sometimes replies go in a subthread, sometimes not, yada yada.

bbbales2:
This is ready for review.

wds15:
I can do that. I want to check the performance of the sir thing first... I can handle that.

PR #1863, opened 2020-05-04
[WIP] testing speeding up tests

rok-cesnovar:
Testing some ideas to speedup jenkins testing.


SteveBronder:
Nice! This would also let us use gcov for code coverage

wds15:
Quickly looking through this you seem to attempt to munge all tests into very few translational units. The thing we are - I think - bad with is cleanup. So there should be far more recover_memory calls in the test base. This may pose a problem.

rok-cesnovar:
Yes, that is basically what I am doing.  I am trying a few things out and seeing what could work and how much would it help. I will collect more data and make a post on all the ways we could improve our CI.

I should have probably label this [testing] not WIP.

2 ways of speeding up testing are tighter includes in tests or merging multiple tests together. Both have shown to speedup things to up to 8 times locally. 

Obviously none of these will be for free. But we are not deciding yet :)

wds15:
Sure. If this works - great.

It's just that any test using anything with  reverse mode AD being used should probably call at the end of each test a recover_memory - that isn't done at the moment everywhere. This is irrelevant if there is no cleanup, but it matters if all the tests are being combined. So maybe this is worth a consideration (to deal with some more regex magic to fix this or do something more clever).

Let's cross finger some approach here works. Faster testing would be great.

rok-cesnovar:
By merging files in the following folders in jumbo files (in chunks of 15 test files)
```python
    "test/unit/math/prim/fun",
    "test/unit/math/prim/meta",
    "test/unit/math/prim/prob",
    "test/unit/math/rev/fun",
    "test/unit/math/rev/meta",
    "test/unit/math/fwd/core",
    "test/unit/math/fwd/fun",
    "test/unit/math/fwd/functor",
    "test/unit/math/fwd/meta",
    "test/unit/math/fwd/prob",
    "test/unit/math/mix/core",
    "test/unit/math/mix/fun",
    "test/unit/math/mix/prob"
```

It cuts down the execution time of the Full unit from 1hour 10min to 33min. So promising.

SteveBronder:
Neat! Why chunks of 15?

rok-cesnovar:
I tried 30 but that seems to consume too much ram (we run -j16 or -j25 on Jenkins). So 15 was the next I tried.

SteveBronder:
fyi was goofing around with this, and though a bit janky, got it to work for all of prim

https://github.com/stan-dev/math/tree/feature/prim-jumbo-tests

rok-cesnovar:
Yes yes. Prim works as its easy. Mix is where its at :) With all the templating and all.

rok-cesnovar:
Oh you fixed err tests. Didnt understand at first. Thanks! Feel free to merge here. If design docs is approved I will cherry pick from this branch.

PR #1859, opened 2020-04-30
Feature/1854 OpenCL /prim signatures part 1

rok-cesnovar:
## Summary

Add /prim OpenCL support for the following functions (only matrix ones listed, but this also adds the same signatures for row_vector and vector):

```stan
 matrix add(real, matrix)
 matrix add(matrix, real)
 matrix add(matrix, matrix)
 vector col(matrix, int)
 matrix inv(matrix)
 matrix inv_cloglog(matrix)
 matrix inv_logit(matrix)
 matrix inv_sqrt(matrix)
 matrix inv_square(matrix)
 row_vector row(matrix, int)
 int[] dims(matrix)
 int cols(matrix)
 int rows(matrix)
```

This adds no new OpenCL code, just using the generator.

It also adds argument checks to block(), renames `is_valid_expression` to `is_valid_kernel_expression`, `require_all_valid_expressions*` to `require_all_valid_kernel_expressions*`

## Tests

Added tests for all added functions.

## Side Effects

/

## Release notes

Added OpenCL support for prim versions of add, col, inv, inv_cloglog, inv_logit, inv_sqrt, inv_square, row, dims, cols.
Added argument checks to OpenCL version of block() and renamed `is_valid_expression` to `is_valid_kernel_expression` and `require_all_valid_expressions*` to `require_all_valid_kernel_expressions*`.

## Checklist

- [x] Math issue #1854 

- [x] Copyright holder: Rok Češnovar

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

Issue #1775, opened 2020-03-12
Propogating Eigen expression templates with local temporaries


andrjohns:
## Description
As first discussed in [this comment](https://github.com/stan-dev/math/pull/1754#issuecomment-597107299), there is an issue with matrix functions returning expressions when those expressions use local temporary structures.

A simple example is the following:

```
#include <Eigen/Core>
#include <iostream>

template<typename T>
decltype(auto) test_fun(const T& x) {
    const Eigen::Ref<const Eigen::VectorXd>& x_ref = x;
    return (x_ref.array() - 5.0).matrix();
}

int main()
{
    Eigen::MatrixXd test_mat(2,2);
    test_mat << 5, 5, 5, 5;

    Eigen::VectorXd test_vec(2);
    test_vec << 5, 5;

    Eigen::VectorXd out = test_fun(test_mat.diagonal()).eval();
    Eigen::VectorXd out2 = test_fun(test_vec).eval();

    std::cout << "Expression:" << std::endl <<
                 out << std::endl << "Vector" <<
                 std::endl << out2 << std::endl;
}
```

Which returns:
```
Expression:
-5
 0
Vector
 0
 0
```

Because the temporary ```x_ref.array()``` falls out of scope.

A simple fix that seems to work is to explicitly set the inner stride:
```
template<typename T>
decltype(auto) test_fun(const T& x) {
    const Eigen::Ref<const Eigen::VectorXd,0,Eigen::InnerStride<> >& x_ref = x;
    return (x_ref.array() - 5.0).matrix();
}
```

But [Eigen's documentation](https://eigen.tuxfamily.org/dox/classEigen_1_1Ref.html) notes that will disable vectorisation.
#### Current Version:
v3.1.0

bob-carpenter:
The expression for the negation will also go out of scope.

The proposed fix with strides still has expressions that go out of scope.

andrjohns:
That's what I thought, but it works: https://godbolt.org/z/oqraZ3

t4c1:
Running the "simple fix" on my computer crashes. I also don't see why it should work so I guess this example just happens to work with particular compiler.

t4c1:
Some more testing suggests we only get wrong result if the expression needs to be evaluated into `Ref` (maybe this also just randomly works): https://godbolt.org/z/c2nzPn
Evaluating into local variable instead of using `Ref` does not help: https://godbolt.org/z/4q0d0N
[...]

PR #1865, opened 2020-05-05
Add holder operation to kernel generator

t4c1:
## Summary

Adds holder operation to kernel generator. This operation is a no-op, but the operation object also owns pointers to some dynamically allocated objects, used in the argument expression.

This is the solution to problems such as https://github.com/stan-dev/math/pull/1859#discussion_r419303893 or #1775 with kernel generator expressions.

Also creates default `generate` member funtion in `operation_cl`.

## Tests

Holder is tested to keep the owned objects alive.

## Side Effects

None.

## Release notes

Adds holder operation to kernel generator.

## Checklist

- [x] Math issue #1342, #1775 


t4c1:
I named this operation `holder_cl` as it "holds" other expression so they don't go out of scope. However I'm not a fan of this name, so if anyone has a better suggestion ...

SteveBronder:
Is this equivalent to Eigen::Ref<>?

The name is kind of weird, should it just be copy?

t4c1:
> Is this equivalent to Eigen::Ref<>?

Not even similar. This solves a different problem. Right now we can not have functions that use an argument multiple times and return an expression, as they would break for rvalues.

Example:
```
template <typename T>
auto square(T&& x){
    return elementwise_multiply(x, x);
}
```
If x is a rvalue it will go out of scope when the function returns and multiplication will reference garbage memory as its arguments. Using const ref argument won't help. We could forward x as one of the arguments, but than we have the same problem with the other one. Forwarding an argument twice results in same problem.

So this is the solution. There are two interfaces. The first one is more practical for local variables and the second one for arguments, such as in this example.
1:
```
template <typename T>
auto square(T&& x){
    auto* x_heap = new std::remove_reference_t<T>(std::forward<T>(x));
    return stan::math::holder_cl(elementwise_multiply(*x_heap, *x_heap), x_heap);
}
```
2:
template <typename T>
auto square(T&& x){
    return make_holder([](const auto& x) { return elementwise_multiply(x, x); }, std::forward<T>(x));
}

> The name is kind of weird

I agree. I am open to suggestions once you understand what this is about.

> should it just be copy?

No.

SteveBronder:
This feels kind of onerous to put in the user space, should holder_cl be added inside of functions like `elementwise_multiply`?

By user space I mean any functions that are not in the kernel generator. Can the generator have a constructor for matrix_cl rvalues to handle this?

SteveBronder:
Didn't we do something similar to this for `matrix_cl`?

t4c1:
That would be all kernel generator functions than. I don't think we can make it work that way. As I see it, holder must be explicitely constructed to make it work. When one kg operation is being constructed it has no knowledge of whether the same rvalue is also used in another function. 

An alternative would be to use dynamic allocations and `std::shared_ptr` for all expressions everywhere. However I don't like it.

> Didn't we do something similar to this for matrix_cl?

I don't think so. What exactley do you have in mind?

rok-cesnovar:
I think Steve is thinking of this PR: https://github.com/stan-dev/math/pull/1421

t4c1:
In that PR we also moved some expressions to heap to extend their lifetime. But the reason for that was different. In matrix_cl we wanted to do asynchronous copy and we needed to ensure that the matrix we are copying exists untill copying finishes. 

Here we want to pass a rvalue to other function(s) multiple times. Instead we move it to heap and pass lvalue references to it.

SteveBronder:
> That would be all kernel generator functions than. I don't think we can make it work that way. As I see it, holder must be explicitely constructed to make it work. When one kg operation is being constructed it has no knowledge of whether the same rvalue is also used in another function.

This is all very confusing. Correct where I'm wrong below. The problem is with this code

```cpp
template <typename T>
auto square(T&& x){
    return elementwise_multiply(x, x);
}
```
`x` is able to be an rvalue reference to a `matrix_cl` or kernel generator expression. If that is true and `x` is passed as both sides of another kernel generator expression such as `elementwise_multiply` your saying the object `x` is referencing can fall out of scope leading to picking up the address of garbage memory. 

If that's true, `x` is just a reference to something outside this functions scope. If the issue is x falling out of scope in this function won't we need to wrap holder_cl around every function anyway?

Does Eigen not run into this same issue? Could we have each matrix_cl / expression wait for all it's read/write events before calling the destructor?

t4c1:
Exactley.

> If the issue is x falling out of scope in this function won't we need to wrap holder_cl around every function anyway?

If a matrix_cl or expression is only used once, it can be moved into whatever expression it is used in. So the function return value will also contain the input to the function.

> Does Eigen not run into this same issue?

It does. We will need to do something very similar for many functions working with Eigen expressions, before they can return expressions.

> Could we have each matrix_cl / expression wait for all it's read/write events before calling the destructor?

At the time such a function returns no kernel is running yet - function is only constructing part of the expression that will be used in the kernel.

t4c1:
@SteveBronder Is anything still unclear here?

t4c1:
It seems @SteveBronder is too busy. @rok-cesnovar can you review this?

rok-cesnovar:
Sure, though I am bad at naming stuff, so no new ideas on how to name this. Will look at this later today.

SteveBronder:
> It does. We will need to do something very similar for many functions working with Eigen expressions, before they can return expressions.

I think I'm just confused on this. Does Eigen have the same holder type or what do they do for this?

andrjohns:
Eigen doesn't actually handle it, which has caused issues in the past (and kinda still is an issue), there's more information over in #1775

stan/math/opencl/kernel_generator/operation_cl.hpp
```diff
@@ -225,6 +225,21 @@ class operation_cl : public operation_cl_base {
     return res;
   }
 
+  /**
+   * Generates kernel code for this expression.
+   * @param i row index variable name
+   * @param j column index variable name
+   * @param view_handled whether whether caller already handled matrix view
```
rok-cesnovar, line 232: Double whether here.

stan/math/opencl/kernel_generator/holder_cl.hpp
```diff
@@ -0,0 +1,129 @@
...
+ */
+template <typename T, typename... Ptrs>
+auto holder_cl(T&& a, Ptrs*... ptrs) {
+  return holder_cl_<as_operation_cl_t<T>, Ptrs...>(
+      as_operation_cl(std::forward<T>(a)), ptrs...);
+}
+
+namespace internal {
+template <typename T>
+auto holder_cl_handle_element(const T& a, const T*& res) {
```
rok-cesnovar, line 77: missing doxygen

stan/math/opencl/kernel_generator/holder_cl.hpp
```diff
@@ -0,0 +1,129 @@
...
+  res = &a;
+  return std::make_tuple();
+}
+
+template <typename T>
+auto holder_cl_handle_element(std::remove_reference_t<T>&& a, const T*& res) {
```
rok-cesnovar, line 83: missing doxygen

t4c1:
> Does Eigen have the same holder type or what do they do for this?

Element-wise functions they tend to implement elemet wise - similarly as we do in apply_scalar_unary, except they also add vectorized implementations. In some other cases they don't handle it at all - they evaluate the expression before returning.

t4c1:
Also Eigen is written assuming C++03 standard, so they handle nothing about rvalues and moves. Instead they use simple heuristics to choose whether to store argument expressions by reference or value. Matrices and similar are stored by reference, expressions by value.

t4c1:
@andrjohns Do you have any suggestion for a better name for this?

Issue #21, opened 2015-07-06
NaN derivative in square(sqrt(x)) in model local variable


syclik:
_From @bob-carpenter on September 4, 2013 7:7_

Simon Barthelmé reported  via the stan-users list.  I simplified the reported model to this minimal instance:

```
parameters {
  real x;
}
model {
  real y;
  y <- sqrt(square(x-x));

  x ~ normal(0,1);
}
```

As Simon correctly diagnosed, this model causes a problem with autodiff, as can be seen when running from RStan 1.3:

```
TESTING GRADIENT FOR MODEL 'barthelme2' NOW (CHAIN 4).

TEST GRADIENT MODE

 Log probability=-0.0571533

 param idx           value           model     finite diff           error
         0       -0.338093             nan        0.338093             nan
```

First step is to move this into a unit test that displays the behavior.  

_Copied from original issue: stan-dev/stan#203_

syclik:
_From @bob-carpenter on September 23, 2013 18:39_

We can't really fix this problem without a *huge* amount of work moving to expression templates;  with expression templates, we could follow Eigen in carrying out symbolic math at compile time, such as reducing ```sqrt(square(u))``` to ```u```.

I just checked that the boundary conditions in both ```sqrt()``` and ```square()``` are correct, as is the application of the chain rule.  

The problem is that the chain rule is applied literally with floating-point arithmetic, which given

```
sqrt'(u) = 0.5 / sqrt(u)

square'(v) = 2 * v
```

yields

```
d/dx sqrt(square(x-x))

  = sqrt'(square(x-x)) * square'(x - x) * d/dx (x - x)

  = (+ infinity)  * 0 * 0

  = NaN.
```

I am going to add a discussion of this case to the manual, so I'm linking #

The problem is that ```(infinity * 0)``` produces ```NaN``` in floating-point arithmetic.  There's no way given just the infinity and 0 that mulitplying them should produce 1.

syclik:
_From @bob-carpenter on November 20, 2013 5:20_

Implement as Marcus suggested in e-mail:

Putting those two things together, maybe  we can just rewrite `chain()` for `sqrt()` to be:

```
void chain() {
  if (val_ > 1e-300)
    avi_->adj_ += adj_ / (2.0 * val_);
  else
    avi_->adj_ += adj_ / 2.0;
  }
```

I asked:    Will this do the right thing for HMC and BFGS?  And Marcus responded:

[...]

PR #1873, opened 2020-05-08
[WIP] add vari_value and var_value to replace vari and var

SteveBronder:
## Summary

This is the first step from the design document for [static matrices](https://github.com/stan-dev/design-docs/pull/21). This PR

1. Adds three new classes `vari_value<T>`, `var_value<T>`, and pure virtual class `vari_base`. `vari_value<T>` inherits from `vari_base`.
2. `vari_value` has a not type template parameter for specialization using `requires`. This is used to define a `vari_type<T, require_arithmetic_t<T>>` and will be used be used to write out the specialization for container types.
2. `ChainableStack`'s vectors now hold pointer types of `vari_base` so that conversion from `vari_type` and it's derived types is possible without any templates on `ChainableStack`.
3. Adds aliases `var` and `vari` as double templated aliases of `var_value` and `vari_value`
4. Changes the operators defined in `var_value` that operate on `this` from working on `var` types to `var_value`.

## Tests

@bob-carpenter for the tests on this we need to test that var_type<T> can properly handle floating point and arithmetic types, Since everything only works with var_type<double> right now is it okay to keep all the test_ad stuff as is? I've started working on another branch which handles long double, double, and float in test_ad.

## Side Effects

Yes

While `var` and `vari` stay the same, the underlying structure of our autodiff type has changed. 

## Release notes

Replaces Stan's underlying autodiff type so that it can take accept a type `T` instead of only holding a value and adjoint of type `double`

## Checklist

- [x] Math issue [#21](https://github.com/stan-dev/design-docs/pull/21)

### DIFF
--- a/stan/math/rev/core/vari.hpp
+++ b/stan/math/rev/core/vari.hpp
@@ -4,13 +4,42 @@
 #include <stan/math/rev/core/chainable_alloc.hpp>
 #include <stan/math/rev/core/chainablestack.hpp>
 #include <ostream>
+#include <type_traits>
 
 namespace stan {
 namespace math {
 
 // forward declaration of var
-class var;
+template <typename T>
+class var_value;
 
+class vari_base {
+public:
+  /**
+   * Apply the chain rule to this variable based on the variables
+   * on which it depends.  The base implementation in this class
+   * is a no-op.
+   */
+  virtual void chain() = 0;
+
+  /**
+   * Initialize the adjoint for this (dependent) variable to 1.
+   * This operation is applied to the dependent variable before
+   * propagating derivatives, setting the derivative of the
+   * result with respect to itself to be 1.
+   */
+  virtual void init_dependent() = 0;
+
+  /**
+   * Set the adjoint value of this variable to 0.  This is used to
+   * reset adjoints before propagating derivatives again (for
+   * example in a Jacobian calculation).
+   */
+  virtual void set_zero_adjoint() = 0;
+
+  virtual ~vari_base() {}
+
+};
 /**
  * The variable implementation base class.
  *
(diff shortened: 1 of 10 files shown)


SteveBronder:
@serban-nicusor-toptal  it looks like this job has been hanging around for a while?

bob-carpenter:
I don't think this should be merged until (a) the design doc is approved , (b) there's a testing plan for the code.  

I don't think waiting is going to cause these new changes to go stale as they seem pretty orthogonal to the rest of the code base.

And that's just me.  But I'd also like to hear from @syclik and other devs on this one.

wds15:
Yup... changing what `var` is in stan math is a huge change indeed.

This is marked WIP, so no reason to assume this going to be merged. 

We need to make sure it‘s compatible, fast as usual, tested & documented. That‘s nothing special in particular, but given this is var we have to be through, I think.

Having said that, the first benchmarks were really promising and showing that the effort will be worth it. I would love to see an end-to-end benchmark with a Stan model as well, if that is possible without too much pain.

This looks like going to be a really cool thing!

SteveBronder:
Yeah I'm v fine with waiting (marked as WIP) and this is a pretty drastic change that needs a lot of benchmarking (those sir numbers and a few others are worrisome)

This is mostly to review the basic design to see if it makes sense / to do testing

bbbales2:
> I don't think this should be merged until (a) the design doc is approved

I don't think it should be merged until the design-doc is approved either, but I think we need to do a few design iterations over here in code before finishing the design doc too. It's an implicit solve!

stan/math/rev/core/vari.hpp
```diff
@@ -27,21 +64,33 @@ class var;
...
+                           std::decay_t<Val>>;
 
  public:
+  using Scalar = floating_point_promoter<T>;
+  using value_type = Scalar;
   /**
    * The value of this variable.
    */
-  const double val_;
+  const Scalar val_;
```
bbbales2, line 87: For this to not cause memory leaks, `Scalar` needs to be something that does not require its destructor to be call. Somehow that needs documated.

In this theme, should we try to make this templated thing as general as possible? Or should we just implement: `vari_matrix`, `vari_vector`, `vari_row_vector` and use those everywhere? (they could certainly depend on something templated)

SteveBronder: So this specialization is just for arithmetic types

```cpp
class vari_value<T, std::enable_if_t<std::is_arithmetic<T>::value>>
    : public vari_base {
```

It should do the same thing as the previous version. Will having two arithmetic types here leak memory?

SteveBronder: The actual Eigen implementation is here. Do you want me to open up a PR on that?

https://github.com/stan-dev/math/compare/feature/eigen-vari#diff-130c5a75cc427d7d41715e9fca8281f4R176

bbbales2: > std::enable_if_t<std::is_arithmetic<T>::value>

Oooh, got it.

> The actual Eigen implementation is here.

Oooh, I see.

> Do you want me to open up a PR on that?

Yeah, this makes me want to do it all together. I don't think I'm gonna be able to keep all this stuff separate in my head and I'm just gonna keep asking questions that have been answered elsewhere.

SteveBronder: Word! See below

https://github.com/stan-dev/math/pull/1877

Issue #1880, opened 2020-05-13
Implement LambertW


Christopher22:
## Description
Lambert W is a function having a "myriad of applications" (according to the boost authors). Exposing the in great detail specified, performance-optimized, and well-tested function from boost to Stan might be useful in many domains (i.e. https://discourse.mc-stan.org/t/adding-lambert-w-transforms-to-stan/13906/3). I will start doing this.

#### Current Version:
v3.2.0

bob-carpenter:
This'd be great.  The main thing to worry about is that each argument be independently differentiable (or marked as `data` in stanc3 and in the doc).  The autodiff testing framework should make testing of the autodiff versions straightforward.  Also, it's OK to start by just autodiffing the algorithm, though ideally we'd find good iterative algorithms for all the relevant partials.  Let us know if you need more help navigating the Stan code base or practices.

PR #1882, opened 2020-05-13
Implementing Lambert W

Christopher22:
## Summary
This pull request adds primitive versions of both branches of the Lambert W function according to the implementation of Boost.

## Tests
The functions are tested in the new file "lambert_w_test.cpp". Many additionals test regarding the correctness of the results are run by the Boost implementation itself.

## Release notes
Implements both branches of the Lambert W function. 

## Checklist

- [x] Math issue #1880 

- [x] Copyright holder: Christopher Gundler


SteveBronder:
Nice and thanks! A couple things. 

1. You'll want a test at least like the following to make sure this is differentiable

```
// ./test/unit/math/mix/fun/lambertw_test.cpp
#include <stan/math/mix.hpp>
#include <test/unit/math/test_ad.hpp>
#include <gtest/gtest.h>

TEST(mathMixMatFun, lambert_w0) {
  using stan::test::expect_unary_vectorized;
  auto f = [](const auto& x1) {
    using stan::math::lambert_w0;
    return lambert_w0(x1);
  };
  stan::test::expect_ad_vectorized(f, 0.0);
  // seg fault
  //stan::test::expect_ad_vectorized(f, -0.3);
  // stan::test::expect_ad_vectorized(f, -0.1);
  // Wrong answer + seg fault
  stan::test::expect_ad_vectorized(f, 1);
}
```

With your current code I think it's causing an infinite loop to happen because it just runs forever.

2. If you remove the int and double specializations then it runs but seg faults or gives the wrong answer. I tried adding a var specialization to handle reverse mode ([here](https://github.com/stan-dev/math/compare/feature/christopher22/lambertw)*) thinking that using their deriv calculation would be useful though I might be doing that wrong. I think the real issue is that they internally use higher precision types and casting those to double for var is causing issues (idk what causes the seg faults tho you wanna use gdb to figure that out). A while ago I tried writing my own [lambertw function](https://gist.github.com/SteveBronder/f0dc12df45280be4c1320df5a08e7601) based on [Fukushima](https://www.sciencedirect.com/science/article/pii/S0377042712005213) but then stopped at trying to add it to stan because I thought it was too confusing and was going to be weird to get to work for var, fvar, etc with the approximations.

If you don't want it to be differentiable then we'd have to figure out the use-case for it

*there's actually two bugs I had to fix, one for `var` not having constructor for unsigned long long types and another that std::numeric_limits is not fully defined for `var` and `fvar`. Gonna make an issue for that

Christopher22:
Well, I have to thank you for the extend of your improvements! As far as I understand the Boost implementation, they offer sub-implementations for floats, doubles, and higher precision types where the latter makes us the headache. Because we did not specify this otherwise explicitly, they use the most precise one. We can actually [limit](https://www.boost.org/doc/libs/1_73_0/libs/math/doc/html/math_toolkit/pol_ref/precision_pol.html) the desired precision by changing the default Boost policy incorporated from <stan/math/prim/fun/boost_policy.hpp>.

Would it be "philosophically" sound to change from the now incorporated "Boost, do it as precise as possible" to "Boost, do it with 52 bits double precision"?

SteveBronder:
> Would it be "philosophically" sound to change from the now incorporated "Boost, do it as precise as possible" to "Boost, do it with 52 bits double precision

Sure!

SteveBronder:
> We can actually limit the desired precision by changing the default Boost policy incorporated from <stan/math/prim/fun/boost_policy.hpp>.

Instead of changing the default policy though you can just use a particular policy when calling the lambert functions

Christopher22:
I tried to combine the best of both worlds: I added a default compile-time argument to the policy. Than we have a single policy when "communicating" to Boost and yet the freedom to change the precision in case it is needed. Unless explicitly stated, the precision should be exactly the same as before.

Christopher22:
Mmh, even with reduced accuracy the calculation seems still wrong...

@SteveBronder Unfortunately, at least on my Windows 10 x64, it seems that g++ is unable to handle your overload
> *there's actually two bugs I had to fix, one for `var` not having constructor for unsigned long long types and another that std::numeric_limits is not fully defined for `var` and `fvar`. Gonna make an issue for that

I just got a corresponding error during testing, even if Jenkins is able to run the tests
```
./stan/math/rev/core/var.hpp:212:3: error: 'stan::math::var::var(size_t)' cannot be overloaded with 'stan::math::var::var(long long unsigned int)' 
    var(size_t x) : vi_(new vari(static_cast<double>(x), false)) {}  // NOLINT
```

SteveBronder:
Hey sorry this too me so long to look at! Had my head down on another PR. I had time to look at this and figured out that I'm dumb and was doing the autodiff wrong. I have a branch below which fixes up the w0 branch (you'll have to do the same things for w1).

1. Adds an fvar specialization. Not having this was causing the segfaults as the forward mode gradients were being calculated using `boost::math::lambert_w0` directly and it couldn't handle the `fvar<T>` type. 
2. For the W_0'(0) case threw in an if statement to just do `W_0'(0)=1` for reverse mod.
3. Instead of using `lambert_w0_prime` the derivative is pretty simple so we just do it directly for rev and fwd
4. Your original code in `prim` was correct and I goofed that, added back the specialization in prim but templated it to only work for arithmetic types.
5. We didn't need the changes to the boost policy, since now boost only ever sees actual arithmetic values it can do whatever it wants underneath the hood.

The rev and fwd mode still needs to check that the value coming is in a space ([-1/e, inf] for W_0 see the [wiki](https://en.wikipedia.org/wiki/Lambert_W_function))  and tests for that, then this stuff needs to be added for W_1 as well. Sorry if my changes to your branch threw you off the path that's my L

https://github.com/stan-dev/math/compare/feature/lambertw0-fixes

Christopher22:
Amazing. Even as a complete newbie, your code makes it quite easy to get a feeling for the rather complex code base. :)

Regarding the errors: Do we really need to introduce the corresponding "check_" functions to our code? Boost throws std::domain_error automatically, should that not be enough?

Christopher22:
Even the rev tests for the W-1 branch are successfull now.  Just this problem still persists on Windows:

> 
> I just got a corresponding error during testing, even if Jenkins is able to run the tests
> 
> ```
> ./stan/math/rev/core/var.hpp:212:3: error: 'stan::math::var::var(size_t)' cannot be overloaded with 'stan::math::var::var(long long unsigned int)' 
>     var(size_t x) : vi_(new vari(static_cast<double>(x), false)) {}  // NOLINT
> ```

stan/math/rev/fun/lambert_w.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace stan {
+namespace math {
+
+namespace internal {
+class lambertw0_vari : public op_v_vari {
+ public:
+  explicit lambertw0_vari(vari* avi) : op_v_vari(lambert_w0(avi->val_), avi) {}
+  void chain() {
+    if (avi_->val_ == 0.0) {
```
bob-carpenter, line 18: Does the else case not work here?  If it does, then remove the conditional and just use the general case.  Conditionals are very expensive when speculative evaluation fails and the CPU has to backtrack.  It's better to pay a penalty most of the time in doing a little more arithmetic in the mainline case.  It also helps with inlining.

stan/math/prim/fun/lambert_w.hpp
```diff
@@ -0,0 +1,82 @@
...
+  template <typename T>
+  static inline T fun(const T& x) {
+    return lambert_wm1(x);
+  }
+};
+
+/**
+ * Vectorized version of lambert_wm1().
+ *
+ * @tparam T type of container
```
bob-carpenter, line 68: This can be either a real-valued scalar or a container.

stan/math/prim/fun/lambert_w.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Structure to wrap lambert_w0() so it can be vectorized.
+ *
+ * @tparam T type of variable
+ * @param x variable
+ * @return value of the W0 branch of the Lambert W function at x.
+ * @throw std::domain_error if x is smaller than -e^(-1)
+ */
+struct lambert_w0_fun {
```
bob-carpenter, line 29: These utility structs should be in the internal namespace.

stan/math/rev/fun/lambert_w.hpp
Christopher22: As far that I understand @SteveBronder added the clause because it leads to failures otherwise. Maybe he can enlighten us

stan/math/prim/fun/lambert_w.hpp
Christopher22: Sorry, I do not really understand. Are the scalars not covered with the "inline auto lambert_w0(const T& x)" overloads above?

stan/math/rev/fun/lambert_w.hpp
SteveBronder: I can check tmrw but I remember it failing without the else clause

SteveBronder:
> I just got a corresponding error during testing, even if Jenkins is able to run the tests

I think the var type needs a long long int constructor. Or we could take all the integral type constructors and just add a templated constructor that only works with all integral types

stan/math/rev/fun/lambert_w.hpp
SteveBronder: Nope that if can be deleted!

SteveBronder:
Actually for the long long error I think we could just cast the result of the return of the boost lambert functions in prim to double

stan/math/prim/fun/lambert_w.hpp
SteveBronder: Bob the apply_scalar_unary functions for lambert have a `require_not_stan_scalar_t<T>` on them so it can only be a container type. There are specializations for arithmetic, var, and fvar types

stan/math/prim/fun/boost_policy.hpp
```diff
@@ -15,13 +15,19 @@ namespace math {
  * (1) overflow errors return error numbers on error.
  * (2) pole errors return error numbers on error.
  * (3) doubles passed to Boost are not promoted to long double.
+ *
+ * The policy is equipped with an optional generic argument B controlling the
+ * precision in some functions. If set to 0, the maximum precision available
+ * in the type being used is demanded from Boost. Otherwise, it correspond to
+ * the approximately B-bit precision, i.e. for trading speed for accuracy.
  */
+template <int B = 0>
```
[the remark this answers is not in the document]
Christopher22: We do not need it. Nevertheless, I would say we may keep it that way just for making changes easier in future: The boost policy is now everywhere registered as generic struct with default arguments with the `boost_policy<>()` rather than `boost_policy()` syntax. Adding other default template arguments is now far easier than before, as long as the C++ version is not lifted to C++17 where both expressions are equal. But personally, I do not care.

SteveBronder:
Pinging @bob-carpenter about the NA vs throw

stan/math/prim/fun/lambert_w.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace internal {
+
+/**
+ * Structure to wrap lambert_w0() so it can be vectorized.
+ *
+ * @tparam T type of variable
+ * @param x variable
+ * @return value of the W0 branch of the Lambert W function at x.
+ * @throw std::domain_error if x is smaller than -e^(-1)
```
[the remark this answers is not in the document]
bob-carpenter: For use in Stan, we generally want to throw an exception so that users can catch errors early.  That's how all of our other distributions work.  About the only place we don't do this is for built-in C++ functions that return NA rather than throwing (like log(-2) or 0 / 0).

Christopher22:
> 
> 
> Actually for the long long error I think we could just cast the result of the return of the boost lambert functions in prim to double

May you take care of this, @SteveBronder ? I (yet) do not feel comfortable with this fancy amount of (generic) numeric types in the codebase...

SteveBronder:
Oh sorry for the confusion I just meant this

https://github.com/stan-dev/math/commit/c6ed4d35535d7376e6786cabb625fea5c48c4f5b

You can pull in that branch from `feature/lambert-return-double `

SteveBronder:
I'm pretty sure the issue was just that lambertw was returning an unsigned long long, so var had to deal with that and windows had something weird about size_t and unsigned long long

Christopher22:
Amazing! Now it seems to work :)

SteveBronder:
@Christopher22 if you merge to develop this should pass and be good to merge!

Christopher22:
@SteveBronder That should be done! Thank you again for your guideline through the process! Next step would be the exposure in the Stan interface - I'm looking forward to another code base ;)

SteveBronder:
Awesome and no problem!! I filed an issue in the compiler with a link to the file that needs to be modified. Should just be a little two liner!

https://github.com/stan-dev/stanc3/issues/571

We should also think about how to do the stuff from [Georg Goerg](https://www.hindawi.com/journals/tswj/2015/909231/) for either Tukey's H/HH distributions or a general functor like `lambert_transform_h/hh(lpdf_func, lambert_parameters, paramters...)`

SteveBronder:
Also feel free to merge when this passes

Christopher22:
> 
> 
> Also feel free to merge when this passes

May you approve this again? For some reasons, the branch it still blocked and I still can not merge it.

rok-cesnovar:
Congrats on your first merged pull request to Stan Math. :tada:

PR #1884, opened 2020-05-13
[WIP] Reverse Mode For Static Matrix Multiplication

SteveBronder:
## Summary

If all these WIPs are getting annoying I can close all the others since this one has everything in it.

This is adds the [reverse mode matrix multiplication](https://github.com/stan-dev/math/pull/1884/files#diff-4d75dd68c0cf62489e92b7584c06fde3) for static matrices. At uses a trick in the `chain()` method to call either the standard multiplication chain method or the matrix chain method. The `chain_impl()` function has a doc going over how this works It also adds a `multiply_vari` specialization for Arith * eigen_var vs. eigen_var * Arith.

This leaks memory right now as op_vari holds the matrix which is not allocated on our stack. I've been having some trouble writing the code to allocate that memory with `op_vari`, if @t4c1 or @bbbales2 know some tuple magic to write that it would be very appreciated! Another option is just to remove `op_vari` and template `dv_vari`, `vd_vari` etc. to take in and allocate the mem for eigen matrices. It's a code density vs maintanence tradeoff. If we can find a nice solution then `op_vari` is fine, but if it's too confusing then it may be better to go back to the old op code. I have the start of the code for op_vari with stack allocated mem for the eigen matrices [here](https://github.com/stan-dev/math/pull/1884/files#diff-d92597d2e956a8ff897693925a32a249)

## Tests

I just wrote an informal test right now that checks if static vs dynamic matrices return the same adjoint calculations after calling .grad()

```
./runTests.py -j18 ./test/unit/math/rev/core/operator_multiplication_test.cpp
```

## Checklist

- [x] Math issue [#21](https://github.com/stan-dev/design-docs/pull/21)


bob-carpenter:
> If all these WIPs are getting annoying I can close all the others since this one has everything in it.

Yes, please close redundant PRs.

> This leaks memory right now as op_vari holds the matrix which is not allocated on our stack. 

I'm not one of the people you called out, but I see two ways to remediate:

1. put a pointer to the matrix being allocated on the stack to have its destructor called

2. save just the values and use a `Map<Matrix<...>...>`.

> It's a code density vs maintanence tradeoff.

Can you say more about this tradeoff?

bbbales2:
> I've been having some trouble writing the code to allocate that memory with op_vari

I favor an adj_jac_apply solution in these cases so we aren't writing a bunch of different varis for the combinations of var/arithmetic/`var_type<MatrixXd>` arguments coming in.

I'm curious what @t4c1 thinks here.

SteveBronder:
> Yes, please close redundant PRs

Done!

> I'm not one of the people you called out, but I see two ways to remediate:

Just the two I thought off the top of my head but much appreciated! (2) is what I was having trouble with but I'll look into how to do (1)

> Can you say more about this tradeoff?

It's mostly the tradeoff of one class that can do a lot of stuff but has complicated templating / logic vs. multiple specialized classes that are simpler but more code. So either less code but harder to parse vs. more code but easier to read.  `op_vari` replaces all of the specialized classes, which is a good bit of code, but if we want to do the (2) that you mentioned I think we may end up with a complicated class. Maybe there's a clever solution I'm not aware of. I'm fine with either and it sounds like Ben has an idea for something completely different

> I favor an adj_jac_apply solution in these cases so we aren't writing a bunch of different varis for the combinations of var/arithmetic/var_type<MatrixXd> arguments coming in.

I haven't looked into adj_jac_apply that much but can read it over. If you have a different implementation idea that would work I'm open to it!

t4c1:
> adj_jac_apply solution

What is that? Can someone explain or link some explanation?

bbbales2:
It's questionably doc'd outside of the code, but here: https://discourse.mc-stan.org/t/adj-jac-apply/5163/6 (autodiffing `A x = b` solve: https://gist.github.com/bbbales2/523db7848df48117dd39788815dfe1a2)

Here for an add function: https://discourse.mc-stan.org/t/adj-jac-apply/5163/6

and here for simplex_constrain: https://discourse.mc-stan.org/t/adj-jac-apply/5163/15?u=bbbales2

bbbales2:
The biggest annoyance I had with it was how I had to copy values/adjoints from varis into Eigen::VectorXds and stuff to make the interface sane. Now the copying won't be necessary!

bob-carpenter:
> > adj_jac_apply solution

> What is that? Can someone explain or link some explanation?

When lifted up to a multivariate setting, the local chain-rule propagations for a node in an expression down to its operand nodes operates as follows.  For each operand, the adjoint of the result is mulitplied by the Jacobian of the result  and then added to the adjoint of the operand.  In univariate terms, if y = f(x), then the adjoint of y is d.lp/d.y, the adjoint of x is d.lp/d.x and the adjoint of x is incemented by the product of the Jacobian f'(x) and the adjoint of y, so that's

d.lp/d.x += d.lp/d.y * f'(x)

All the adjoints are initialized to 0, and then d.lp/d.lp is initialized to 1 to start the chain-rule propagation.

The final result lp is always univariate.  But now suppose x and y are multivariate so that f' is a Jacobian and the product is whichever vector/matrix/tensor proudct you need to make all the dimensions work out and you're good to go (there may need to be a transpose in there somewhere---I'm always too lazy to work out the dimensions).

bob-carpenter:
> Now the copying won't be necessary!

That is the whole point of this data type.

betanalpha:
> 
> The final result lp is always univariate. But now suppose x and y are multivariate so that f' is a Jacobian and the product is whichever vector/matrix/tensor proudct you need to make all the dimensions work out and you're good to go (there may need to be a transpose in there somewhere---I'm always too lazy to work out the dimensions).

The Jacobian is transposed.  It helps to think of the Jacobian as linear operator
acting on a tangent vector — forward mode is

v_{N} 
= J * v_{1} 
= (J_{N} * J_{N - 1} * … * J_{2} * J_{1}) * v_{0}
= J_{N} * J_{N - 1} * … * J_{2} * (    v_{1}      )
...
= J_{N} * v_{N - 1}

while reverse mode is the transpose operation action on cotangent/adjoint vectors,

alpha_{1} 
= J^{T} * alpha_{N + 1} * 
=   (J_{N} * J_{N - 1} * … * J_{2} * J_{1})^{T} * alpha_{N + 1} *
= J^{T}_{1} * J^{T}_{2} … * J^{T}_{N - 1} * J^{T}_{N} * alpha_{N + 1}
= J^{T}_{1} * J^{T}_{2} … * J^{T}_{N - 1} * (      alpha_{N}        )
…
= J^{T}_{1} * alpha_{2}

Most of our autodiff implementations compute J and then a vector-matrix product,
but in most cases we can compute the action of the vector-matrix product without 
having to construct all of J, saving valuable memory.  This is particularly true for all 
of the matrix operations.

bob-carpenter:
There are two things to consider in terms of sizing.  First is that we can construct the Jacobian in the reverse pass as a `double` matrix a lot more efficiently than we can put the Jacobian on the autodiff stack.  But usually we don't ever need to actually form the Jacobian in memory.

SteveBronder:
Neat! So for this I would need to 

1. Cleanup adj_jac_apply (or specialize?) to work with `var_value<Eigen::Matrix>`
2. Write the multiply functors
3. Have an `operator*` specialized for at least one of the types being `var_value<Eigen>` that calls 

`stan::math::adj_jac_apply<MultFunctor>(x1, x2)`

betanalpha:
> Neat! So for this I would need to
> 
> Cleanup adj_jac_apply (or specialize?) to work with var_value<Eigen::Matrix>
> Write the multiply functors
> Have an operator* specialized for at least one of the types being var_value<Eigen> that calls
> stan::math::adj_jac_apply<MultFunctor>(x1, x2)
> 

I believe so!  Once (1) is set up then all of the magic is in efficient 
Jacobian transpose/adjoint product implementations, especially 
working out how to compute things sequentially to make the linear
algebra as efficiently as possible.

stan/math/rev/core/var.hpp
```diff
@@ -31,30 +31,38 @@ static void grad(vari* vi);
  * like exponentiation and powers are overridden to operate on
  * var values objects.
  */
-class var {
+template <typename T>
+class var_value {
+  template <typename Val>
+  using floating_point_promoter
+      = std::conditional_t<std::is_integral<std::decay_t<Val>>::value, double,
+                           std::decay_t<Val>>;
```
andrjohns, line 39: Can this just be replaced by ```return_type_t```?

stan/math/rev/core/var.hpp
```diff
@@ -31,30 +31,38 @@ static void grad(vari* vi);
...
+
  public:
   // FIXME: doc what this is for
-  using Scalar = double;
-
+  using Scalar = floating_point_promoter<T>;
```
andrjohns, line 43: This seems to work:

```suggestion
  using Scalar = return_type_t<T>;
```

stan/math/rev/core/var.hpp
```diff
@@ -63,185 +71,47 @@ class var {
...
+   * is arithmetic.
    *
    * @param x Value of the variable.
    */
-  var(__float128 x) : vi_(new vari(static_cast<double>(x), false)) {}  // NOLINT
+  template <typename IntegralT, typename T1 = T,
+            require_not_same_t<T1, IntegralT>* = nullptr,
+            require_integral_t<IntegralT>* = nullptr,
+            require_arithmetic_t<T1>* = nullptr>
+  var_value(IntegralT x) : vi_(new vari_value<T>(x, false)) {}  // NOLINT
```
andrjohns, line 95:
```suggestion
  template <typename IntegralT, require_integral_t<IntegralT>* = nullptr>
  var_value(IntegralT x) : vi_(new vari_value<return_type_t<IntegralT>>(x, false)) {}  // NOLINT
```

stan/math/rev/core/vari.hpp
```diff
@@ -55,11 +104,11 @@ class vari {
    *
    * @param x Value of the constructed variable.
    */
-  explicit vari(double x) : val_(x), adj_(0.0) {
+  explicit vari_value(T x) : val_(x), adj_(0.0) {
```
andrjohns, line 107: Could combine these constructors with a default argument?:
```
  explicit vari_value(T x, bool stacked = 1) : val_(x), adj_(0.0) {
    if (stacked) {
    } else {
      ChainableStack::instance_->var_nochain_stack_.push_back(this);
    }
  }
```

stan/math/rev/core/vari.hpp
```diff
@@ -68,37 +117,156 @@ class vari {
...
+        cols_(x.cols()),
+        size_(x.size()) {
+    val_ = x;
+    adj_.setZero();
+    if (stacked) {
+      ChainableStack::instance_->var_stack_.push_back(this);
+    } else {
+      ChainableStack::instance_->var_nochain_stack_.push_back(this);
+    }
+  }
```
andrjohns, line 250: Could do something similar here with the default argument:

```suggestion
  explicit vari_value(const T& x, bool stacked = 1)
      : val_mem_(ChainableStack::instance_->memalloc_.alloc_array<eigen_scalar>(
            x.size())),
        adj_mem_(ChainableStack::instance_->memalloc_.alloc_array<eigen_scalar>(
            x.size())),
        val_(val_mem_, x.rows(), x.cols()),
        adj_(adj_mem_, x.rows(), x.cols()),
        rows_(x.rows()),
        cols_(x.cols()),
        size_(x.size()) {
    val_ = x;
    adj_.setZero();
    if (stacked) {
      ChainableStack::instance_->var_stack_.push_back(this);
    } else {
      ChainableStack::instance_->var_nochain_stack_.push_back(this);
    }
  }
```

bbbales2:
I just pushed some changes straight to your branch. There were a couple problems with how `std::vector` was working.

Autodiff fails on the Eigen types now. I think the first problem is that the copy operators need to create ops on the autodiff stack.

So the code here: https://github.com/stan-dev/math/blob/822df6d9a9a4db327763ab57c0810b18241e61df/stan/math/rev/core/var.hpp#L304

That copies the adjoints on the forward pass which is not correct. There needs to be something during chain that copies the adjoints from the `Eigen::Matrix<var, R, C>` to the adjoints of the `var_type<MatrixXd` on the backward pass. So basically that line needs to happen in reverse mode via some operation.

And same for going in the other direction (constructing a `var_type<MatrixXd` from a `Eigen::Matrix<var, R, C>`).

SteveBronder:
Thanks!

> That copies the adjoints on the forward pass which is not correct. There needs to be something during chain that copies the adjoints from the Eigen::Matrix<var, R, C> to the adjoints of the var_type<MatrixXd on the backward pass. So basically that line needs to happen in reverse mode via some operation.

I suspected it was something there, though I'm not really sure how to go about the above. Do you have any ideas?

bbbales2:
It'll need to be a vari of some sort.

And like the chain has references to the varis of the `Matrix<var, R, C>` thing and the vari of the `var_value<MatrixXd>` thing.

Psuedo-coding it out, something like:

```
class copy_to_individual_varis_vari {
  int N;
  varis** output;
  vari_value<Eigen::MatrixXd>* input;

  ...

  void chain() {
    for(size_t n = 0; n < N; ++n) {
      input->adj_[n] += output[n]->adj_;
    }
  }
};
```

bbbales2:
(This would be one of those varis with no value thing, that just sits on the stack to make sure things get chained)

SteveBronder:
Actually idt the conversion operator is the issue. I added

```cpp
  puts("\nThis was called\n");
```

to the converter and then just made a debug test for one of the tests that's failing to get the adjoints and the converter is never called.

```bash
$ ./runTests.py ./test/unit/math/rev/functor/adj_jac_apply_debug_test
[==========] Running 1 test from 1 test case.
[----------] 1 test from AgradRev
[ RUN      ] AgradRev.test_vector_sin_multiple_jac
test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp:55: Failure
  x1(0).adj()
    Which is: 0
  0.5403023058681398
    Which is: 0.54030228
test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp:63: Failure
  x2(0).adj()
    Which is: 0
  -0.4161468365471424
    Which is: -0.41614684
test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp:64: Failure
  x2(1).adj()
    Which is: 0
  0.5403023058681398
    Which is: 0.54030228
test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp:72: Failure
  x2(0).adj()
    Which is: 0
  1.73 * -0.4161468365471424
    Which is: -0.71993405
test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp:73: Failure
  x2(1).adj()
    Which is: 0
  1.57 * 0.5403023058681398
    Which is: 0.84827465
[  FAILED  ] AgradRev.test_vector_sin_multiple_jac (0 ms)
[----------] 1 test from AgradRev (0 ms total)

[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] AgradRev.test_vector_sin_multiple_jac
```

bbbales2:
I thought I did the same? But maybe not.

Anyway you can print the adjoints coming in to the chain operator and see if they're non-zero (that's easy cause the chain operator is user-defined). If they aren't, then either the adjoints didn't get passed to chain correctly, or chain didn't have the correct adjoints to begin with.

Anyhow the output of adj_jac_apply will be a `var_value<MatrixXd>` right? And then the y variable in the test I was working with ([this one](https://github.com/stan-dev/math/blob/d7984206b89fcd458398a507c29ef53aa6458e2d/test/unit/math/rev/functor/adj_jac_apply_test.cpp#L143)) was a `Matrix<var, R, C>` so there has to be a conversion at some point (assuming the output is indeed a `var_value`).

SteveBronder:
Oh yes I modified that test so that y1/y2 were var_value<eigen>. Though we should have tests for both

SteveBronder:
The conversion operator itself may still need something like that

SteveBronder:
debug test is here

https://github.com/stan-dev/math/pull/1884/files#diff-f5562026e9eb4369703cfff6b2296b47

bbbales2:
Add a `std::cout << adj(n) << std::endl;` [here](https://github.com/stan-dev/math/blob/d7984206b89fcd458398a507c29ef53aa6458e2d/test/unit/math/rev/functor/adj_jac_apply_test.cpp#L111).

If they're all zeros, then either the adjoints aren't getting passed in or maybe they aren't being set right.

What does your grad call look like now? Originally it would set the adjoint of one of the varis to 1 and then chain. Now we need to set one element of the vari_value adjoint to 1 and chain.

bbbales2:
Can you link to the test file + line number?

SteveBronder:
That link should have gone to it (it takes a minute to load after the page opens) but the tests are in `test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp` here's a link directly to the file

https://github.com/stan-dev/math/blob/d38c1557ca81c7e92e2367a56a7bd463314dc501/test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp

bbbales2:
I'm running that code and neither test passes?

This needs to change anyway: https://github.com/stan-dev/math/blob/d38c1557ca81c7e92e2367a56a7bd463314dc501/test/unit/math/rev/functor/adj_jac_apply_debug_test.cpp#L53

We should make the vi argument to grad optional [here](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/grad.hpp)

Something like:
```
static void grad(vari* vi = NULL) {
  if(vi) {
    vi->init_dependent();
  }
  std::vector<vari*>& var_stack = ChainableStack::instance_->var_stack_;
  size_t end = var_stack.size();
  size_t beginning = empty_nested() ? 0 : end - nested_size();
  for (size_t i = end; i-- > beginning;) {
    var_stack[i]->chain();
  }
}
```

And then for the static matrices `y2(1).grad();` can be written:

```
y2->vi_.adj_[0] = 0.0;
y2->vi_.adj_[1] = 1.0;
grad()
```

Or something like that.

SteveBronder:
> I'm running that code and neither test passes?

Yes there are 3 tests in that file now. The one at the very bottom that uses all static matrices passes and the other two fail.

The rest of your comment is unclear to me. Is that so we can take grads wrt just parts of the matrix? I think there are some intricacies to that which we shouldn't think about right now

bbbales2:
Oh okay well that makes sense then. The adjoints won't ever move all the way in the other two tests if there are not varis handling the `var_value<MatrixXd>` and `Matrix<var, R, C>` conversions.

I can wiggle around the other thing. I'm changing it for the ode stuff anyway.

SteveBronder:
Word! I put that std::cout in the `multiply_adjoint_jacobian` and it looks like the adjs_ are being set to 1 correctly, so it must be something in accumulate_adjoints_in_varis

```
[ RUN      ] AgradRev.test_vector_sin_multiple_jac_static_return
Begin MulAdjJac: # For y2
cos(x_mem_[0]): -0.416147  adj(0): 0
cos(x_mem_[1]): 0.540302  adj(1): 0
End MulAdjJac:
Begin MulAdjJac: # for y1
cos(x_mem_[0]): 0.540302  adj(0): 1
End MulAdjJac:
```

bbbales2:
Oh yeah, there is no conversion in the first one, d'oh. Yeah. It should be in accumulate.

SteveBronder:
Hmm, so I think what needs to happen is something like the pseudocode below

```cpp
// takes an input var_value<Matrix> and propogates values to a Matrix<var>
class copy_matrix_varis {
  int N;
  varis**& output_; // Dumb Q: Can these be references instead of storing them?
  vari_value<Eigen::MatrixXd>*& input_;
  copy_matrix_varis(T1& input, T2& output) : input_(input.vi_) {
    for (int i = 0; i < output.size(); ++i) {
      output_[i] = output[i].vi_;
    }
  }
  void chain() {
    for(size_t n = 0; n < N; ++n) {
      // We want to assign grad of var_value<Eigen> to Eigen<var> right?
      output[n]->adj_ += input->adj_[n]; 
    }
  }
};
```

```cpp
class adj_jac_apply {
// stuff here
//...
copy_matrix_varis** mat_copies_;
// ...
class  adj_jac_vari()
      : vari(NOT_A_NUMBER),  // The val_ in this vari is unused
        x_vis_(),
        y_vi_(nullptr),
        mat_copies_(mem_alloc<copy_matrix_varis*>(num_of_mat_vars<Targs...>::value) {}
// ...
  inline auto operator()(const Targs&... args) {
    fill_adj_jac(x_vis_, args...);
    auto func_result = f_(is_var_, value_of(args)...);
    auto res = build_return_varis_and_vars(func_result);
    // add N copy_matrix_varis to propogate res adj_ to Matrix<var> adjs_
    propogate_matrix_adjs(res, args...);
    return res;
  }
// For each Matrix<var>, add a chain() to propogate the value up
template <typename T1, typename EigVar, typename... Args,
  require_eigen_vt<is_var, EigVar>* = nullptr>
  inline void propogate_matrix_adj(const T1& res, const EigVar& x, const Args&... args) {
    // make t from clever indexing
    mat_copies[t] = new copy_matrix_varis(res, x);
    propogate_matrix_adj(res, args...);
  }
```
So the `copy_matrix_varis` will go on the stack, then adj_jac_vari will go on the stack. So when we call the stack in reverse order we do the grad calc for the output, then propogate that output to the matrix<var> values. Does that make sense?

bbbales2:
That'll need to happen anytime we have a conversion -- it's not just an adj_jac_apply thing.

The basic test for this is something like:
```
Eigen::MatrixXd v = Eigen::Matrix::Ones(1, 1);

var_value<Eigen::MatrixXd> x1(v);
Eigen::Matrix<var_value<double>, R, C> x2(x1);

x2(0, 0).grad();
std::cout << x1.adj()(0, 0) << std::endl; // This should print 1.0
```

(And then the other direction as well).

We can use `vari*&` instead of `vari**` as long as we don't need to change the pointer once it's initialized but I don't think it makes anything clearer.

And from the code snippets, probably `intput->adj_[n] += output[n]->adj_;` instead of `output[n]->adj_ += input->adj_[n];`.

SteveBronder:
For this

```cpp
#include <gtest/gtest.h>

TEST(RevCore, static_to_dynamic) {
  using stan::math::var_value;
  Eigen::MatrixXd v = Eigen::MatrixXd::Ones(3, 3);

  var_value<Eigen::MatrixXd> x1(v);
  Eigen::Matrix<var_value<double>, -1, -1> x2 = x1;
  x2(0, 0).grad();
  std::cout << x1.adj() << std::endl;
}
``` 

Do we want `x1.adj()` to be
```
1 0 0
0 0 0
0 0 0
```

I think as long as the `copy_matrix_varis` is used in the conversion operator that should work. To be able to do `Eigen::Matrix<var, -1, -1> x2(x1);` I think we'll have to add a constructor to the eigen_plugins. For the opposite `var_value<Matrix> x1(Matrix<var>)` I think we can add a constructor that does the same thing in reverse

SteveBronder:
> We can use vari*& instead of vari** as long as we don't need to change the pointer once it's initialized but I don't think it makes anything clearer.

idt it will be clearer at all, in fact it will be weird. But I think saving the reference and the size of the dynamic matrix will save a big allocation and we can just iterate from the start to start+size to do the chain()

bbbales2:
> Do we want x1.adj() to be

Yeah

> But I think saving the reference and the size of the dynamic matrix will save a big allocation and we can just iterate from the start to start+size to do the chain()

Oooh, we can't avoid the allocate and copy. The issue is that anything we use in the forward pass might go out of scope for the reverse pass so we have to save a copy of everything on the stack.

SteveBronder:
> Oooh, we can't avoid the allocate and copy. The issue is that anything we use in the forward pass might go out of scope for the reverse pass so we have to save a copy of everything on the stack.

I think we can! I have two weird vari's called `dynamic_to_static_vari` and `static_to_dynamic_vari` that almost does this. It works for vectors but is failing on of the matrix tests right now. But it just saves the pointer to where the vari* is on the stack and the size of the matrix and iterates though allocating the adjoint values. Those won't get erased since all of the vari are on the stack and won't get erased till we call "recover_memory" right?

bbbales2:
Hmm, I think you're right to say I'm wrong about memory, but I still don't think this works in general.

For instance:

```
Eigen::Matrix<var, -1, 1> myvar = rep(var(1.0), 3);
```

So there were `vari*` s to myvar varis presumably allocated in a sequence (these are saved as the output adjoints we'll pull from in chain).

`myvar(1) = var(2.0)`

Now there is still a pointer to the three output varis pointers, but `myvar(1).vi_`, `myvar(2).vi_`, `myvar(3).vi_` is a different sequence of varis.

SteveBronder:
Gah! Yes that's totally correct, the thing I wrote assumed the memory would be contiguous on the stack

SteveBronder:
Hang on, it looks like our current stuff also does not propogate like the above? For the test

```cpp
// dynamic -> static
TEST(RevCore, dynamic_to_static) {
  using stan::math::var_value;
  Eigen::MatrixXd v = Eigen::MatrixXd::Ones(3, 3);
  Eigen::Matrix<var_value<double>, -1, -1> x1(v);
  var_value<Eigen::MatrixXd> x2(x1);
  x1(1, 1) = var_value<double>(10);
  x2.grad();
  puts("X2: ");
  std::cout << x2.adj() << std::endl;
  puts("X1: ");
  std::cout << x1.adj() << std::endl;
}

// dynamic -> dynamic
TEST(RevCore, dynamic_to_dynamic) {
  using stan::math::var_value;
  Eigen::MatrixXd v = Eigen::MatrixXd::Ones(3, 3);
  Eigen::Matrix<var_value<double>, -1, -1> x1(v);
  Eigen::Matrix<var_value<double>, -1, -1> x2 = x1;
  x1(1, 1) = var_value<double>(10);
  for (int i = 0; i < x2.size(); ++i) {
    x2(i).grad();
  }
  puts("X2: ");
  std::cout << x2.adj() << std::endl;
  puts("X1: ");
  std::cout << x1.adj() << std::endl;
}
```

The results are

```
[       OK ] RevCore.static_to_dynamic_multi_grad_clear (0 ms)
[ RUN      ] RevCore.dynamic_to_static
X2: 
1 1 1
1 1 1
1 1 1
X1: 
1 1 1
1 0 1
1 1 1
[       OK ] RevCore.dynamic_to_static (0 ms)
[ RUN      ] RevCore.dynamic_to_dynamic
X2: 
1 1 1
1 1 1
1 1 1
X1: 
1 1 1
1 0 1
1 1 1
```

Am I not thinking about this right or should they have the same behavior?

bbbales2:
```c++
Eigen::Matrix<var_value<double>, -1, -1> x2 = x1;
x1(1, 1) = var_value<double>(10);
```

x1 is changing after the copy to x2 which is where this is coming from.

bbbales2:
Just this is enough to make not-copying the `vari*`s a thing.

```
Eigen::Matrix<var_value<double>, -1, -1> x1(v);
x1(1, 1) = var_value<double>(10);
```

So if something uses x1 after the `(1, 1)` assignment, it needs to copy the `vari*`s. If something uses x1 before the `(1, 1)` assignment, conceivably it could use the old `vari**` and not reallocate, but there's no easy way to tell that an assignment hasn't happened.

SteveBronder:
I went with the copy version, good news adj_jac_apply and the functions that use adj_jac_apply all pass! Bad news is that a bunch of the prob tests fail. Oddly some stuff is already trying to use this and I need to figure out how to make sure that's not a thing. I'm not sure if the better path here is to start workingon the sub-PRs for this or to fix everything in this PR. Worried I'll end up hacking things together here

SteveBronder:
@serban-nicusor-toptal the windows box upstream is showing

```
cc1plus.exe: out of memory allocating 65536 bytes

cc1plus.exe: out of memory allocating 65536 bytes

cc1plus.exe: out of memory allocating 65536 bytes
```

Is that a this PR thing or is that a windows thing?

rok-cesnovar:
There are two separate issues:
https://github.com/stan-dev/stan/pull/2902
which is waiting on
https://github.com/stan-dev/cmdstan/pull/879

SteveBronder:
Oh shoot! Yes I'll look those over in the morning

SteveBronder:
@serban-nicusor-toptal do the cmdstan custom performance tests work again? I remember a while back we had an issue with one of them goofing up. I wanted to run a custom job to look at how this runs on windows and mac

https://jenkins.mc-stan.org/job/CmdStan%20Performance%20Tests/job/Custom/

SteveBronder:
On my machine (ubuntu w/ gcc-9 and amd threadripper) I ran the base tests for 30 runs and didn't see any changes, which is good! A bit slower on compilation but that's expected

```
SUCCESS: Gold golds/stat_comp_benchmarks_benchmarks_arK_arK.gold passed.
+ ./comparePerformance.py develop_performance.csv performance.csv markdown
---RESULTS---
| stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan | 2.38 | 2.34 | 1.02 | 1.49% faster |
| stat_comp_benchmarks/benchmarks/low_dim_corr_gauss/low_dim_corr_gauss.stan | 0.01 | 0.01 | 1.0 | 0.24% faster |
| stat_comp_benchmarks/benchmarks/eight_schools/eight_schools.stan | 0.05 | 0.05 | 0.99 | -0.51% slower |
| stat_comp_benchmarks/benchmarks/arK/arK.stan | 1.5 | 1.28 | 1.17 | 14.41% faster |
| stat_comp_benchmarks/benchmarks/irt_2pl/irt_2pl.stan | 4.35 | 4.32 | 1.01 | 0.63% faster |
| performance.compilation | 74.43 | 79.37 | 0.94 | -6.64% slower |
| stat_comp_benchmarks/benchmarks/low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan | 5.58 | 5.41 | 1.03 | 3.03% faster |
| stat_comp_benchmarks/benchmarks/pkpd/one_comp_mm_elim_abs.stan | 13.27 | 13.5 | 0.98 | -1.73% slower |
| stat_comp_benchmarks/benchmarks/sir/sir.stan | 71.61 | 85.69 | 0.84 | -19.66% slower |
| stat_comp_benchmarks/benchmarks/pkpd/sim_one_comp_mm_elim_abs.stan | 0.22 | 0.23 | 0.99 | -0.95% slower |
| stat_comp_benchmarks/benchmarks/low_dim_gauss_mix/low_dim_gauss_mix.stan | 2.08 | 1.97 | 1.05 | 4.99% faster |
| stat_comp_benchmarks/benchmarks/gp_regr/gp_regr.stan | 0.1 | 0.1 | 1.02 | 2.09% faster |
| stat_comp_benchmarks/benchmarks/arma/arma.stan | 0.35 | 0.35 | 1.01 | 1.05% faster |
| stat_comp_benchmarks/benchmarks/gp_regr/gen_gp_data.stan | 0.03 | 0.03 | 1.02 | 2.17% faster |
| stat_comp_benchmarks/benchmarks/garch/garch.stan | 0.34 | 0.34 | 1.01 | 0.68% faster |
Mean result: 1.00526503967
```

serban-nicusor-toptal:
Hey Steve, they should work just fine. I'll keep an eye on the [job](https://jenkins.mc-stan.org/job/CmdStan%20Performance%20Tests/job/Custom/158/).

SteveBronder:
Nice I'll keep an eye on it!

Also though now I'm eyeballing the SIR model and it's V slow. I think at this point it's worth just breaking this up into smaller PRs to diagnose what's going on in the performance tests

bbbales2:
@SteveBronder mind if I play around in here? I want to merge https://github.com/stan-dev/math/pull/1915 with this and then point this pull there (so then we see the matrix diff).

And then I want to try implementing and benchmarking a few more matrix ops.

SteveBronder:
Yeah go for it!

bbbales2:
@SteveBronder Yo my hope by merging `feature/vari-base-templates` into this and pointing the pull there would be that it would simplify the diff. There's still 78 things there, including: https://github.com/stan-dev/math/pull/1884/files#diff-320e0518dfbfe51abbc870b3ae852b13 which doesn't look like part of either pull.

Any advice on what I should be doing to getting these things more in sync?

bbbales2:
Hmm, and now there are all these merge conflicts! Not sure I'm doing my gits correctly.

SteveBronder:
tbh this is kind of why I wanted to do this in pieces. There's enough very dramatic changes here that when a bunch of stuff at one level changes it causes a bunch of conflicts in the larger branch. I think we should focus efforts on #1915  so then we can start the Eigen var PR and then the adj_jac_apply PR.

bbbales2:
Well I just know I'm not gonna understand #1915 until I know how it filters up to matrices and stuff.

You think there might be a difference in rebase and merge here?

bbbales2:
Eh I'll just try it lol.

SteveBronder:
Lemme look at this rq

SteveBronder:
Huh seems fine, tbh just pulled it down and `git merge feature/vari-base-templats`

bbbales2:
> Huh seems fine, tbh just pulled it down and git merge feature/vari-base-templats

I wonder if I was merging in my local version of your branch or something? But it seems like that wouldn't have given me conflicts twice. Oh well.

SteveBronder:
¯\\\_(ツ)\_/¯

SteveBronder:
@bbbales2 moving this convo over from #1915 the convo [here](https://github.com/stan-dev/math/pull/1915#discussion_r436405088) I think we could do something like the below where (T1 and T2 are `var_value`s). But the problem is that `op_vari` needs to know what the `adj_` type should be so we need to deduce that which requires a lot of weird looking duplicated code. imo it's more confusing than having the boilerplate in the function.

```cpp

/**
 * Deduces the return type for matrix multiplication of two types
 */
template <typename T1, typename T2, typename = void>
struct mat_mul_return_type {};

// arithmetic is just double
struct mat_mul_return_type<T1, T2, require_all_arithmetic_t<T1, T2>> {
  using type = double;
};

struct mat_mul_return_type<T1, T2, require_any_eigen_t<T1, T2>> {
  using type = decltype((std::declval<T1>() * std::declval<T2>()).eval());
};

// helper alias
using mat_mul_return_t = typename mat_mul_return_type<T1, T2>::type;

class multiply_vari<T1, T2, require_all_var_t<T1, T2>>
    final : public op_vari<mat_mul_return_t<value_type_t<T1>, value_type_t<T2>>, vari_type_t<T2>*, vari_type_t<T2>*> {
  using op_vari<mat_mul_return_t<value_type_t<T1>, value_type_t<T2>>, vari_type_t<T2>*, vari_type_t<T2>*>::avi;
  using op_vari<mat_mul_return_t<value_type_t<T1>, value_type_t<T2>>, vari_type_t<T2>*, vari_type_t<T2>*>::bvi;
  using lhs_type = vari_type_t<T1>;
  using rhs_type = vari_type_t<T2>;
  using return_t = mat_mul_return_t<value_type_t<T1>, value_type_t<T2>>;

 public:
  multiply_vari(lhs_type* avi, rhs_type* bvi)
      : op_vari<return_t, lhs_type*, rhs_type*>(avi->val_ * bvi->val_, avi, bvi) {}

  template <typename TT1 = T1, typename TT2 = T2,
            require_all_var_vt<std::is_arithmetic, TT1, TT2>* = nullptr>
  inline void chain_impl() {
    avi()->adj_ += bvi()->val_ * this->adj_;
    bvi()->adj_ += avi()->val_ * this->adj_;
  }

  template <typename TT1 = T1, typename TT2 = T2,
            require_all_var_vt<is_eigen, TT1, TT2>* = nullptr>
  inline void chain_impl() {
    avi()->adj_ += this->adj_ * bvi()->val_.transpose();
    bvi()->adj_ += avi()->val_.transpose() * this->adj_;
  }

  void chain() {
    if (unlikely(is_any_nan(avi()->val_, bvi()->val_))) {
      fill(avi()->adj_, NOT_A_NUMBER);
      fill(bvi()->adj_, NOT_A_NUMBER);
    } else {
      chain_impl();
    }
  }
};

template <typename T1, typename T2, require_all_var_t<T1, T2>* = nullptr>
inline auto operator*(const T1& a, const T2& b) {
  using multiply_type = internal::multiply_vari<T1, T2>;
  // store the return type multiply_vari
  using mat_return = typename multiply_type::return_t;
  return var_value<mat_return>(new multiply_type(a.vi_, b.vi_)};
}
```

We could try to simplify this with something like a default template with a `VarOpTraits` class that stores the return type, value_type, and vari_type of everything but I'm really not sure if that's going to clean things up in a meaningful way

SteveBronder:
From @bbbales2 comment [here](https://github.com/stan-dev/math/pull/1915#discussion_r436405799) wrt to

```cpp
template <typename T1, typename T2, require_all_var_t<T1, T2>* = nullptr>
inline auto operator*(const T1& a, const T2& b) {
  using multiply_type = internal::multiply_vari<T1, T2>;
  // store the return type multiply_vari
  using mat_return = typename multiply_type::return_type;
  return var_value<mat_return>(new multiply_type(a.vi_, b.vi_)};
}
```

> ```cpp
>    var_value<mat_return>
>    internal::multiply_vari<T1, T2>
>```
> Both of these are constructors so you gotta pass the template arguments.

We can deduce the constructor's parameters from T1 and T2

> It'd be nice to not do that, cause the types should be deducible in the arguments. Presumably we have to define this somewhere, but doing it in every function would not be great.

Yeah we need the boilerplate somewhere, imo it's simpler to have in the function in than in the class (though is annoying). One thing we could do is put the type traits stuff into op_vari. Then multiply etc. could look like

```cpp
  class multiply_vari<T1, T2, require_all_var_t<T1, T2>>
      final : public op_vari<mat_mul_return_t<T1, T2>, T1, T2> {
    using return_t = mat_mul_return_t<T1, T2>;
    using op_vari<return_t, T1, T2>::avi;
    using op_vari<return_t, T1, T2>::bvi;
    using lhs_vari = vari_type_t<T1>;
    using rhs_vari = vari_type_t<T2>;
   public:
    multiply_vari(lhs_vari* avi, rhs_vari* bvi)
        : op_vari<return_t, T1, T2>(avi->val_ * bvi->val_, avi, bvi) {}
// yada yada
```

Then in `op_vari` do all the deduction stuff

> This is connected to the multiple definitions of operator* and the multiple multiply_vari types.

> I at least want less usings here.

Yeah it would be good to have less, the Q is just where do they go

SteveBronder:
clicked close by accident

bbbales2:
> Yeah it would be good to have less, the Q is just where do they go

Yes, and how much this will need repeated in other places.

SteveBronder:
I got something almost working here.

https://github.com/stan-dev/math/tree/feature/eigen-vari-ops-vari-deduction

I'm doing something dumb with inheritance though and getting an error that it doesn't understand it can use the `var_value(vari_value<T>* vi)` constructor for `multiply_vari`

SteveBronder:
^That works now. Some parts are a little icky but I'd clean that up when we are at this stage

PR #1889, opened 2020-05-15
Clean up adj_jac_apply (static matrix design doc)

bbbales2:
`count_vars` and `save_varis` dropped in directly. The `accumulate_adjoints` here was doing something different from the `accumulate_adjoints` we have.

In the one we have, we're taking adjoints from varis and saving them in an array of doubles.

In the one here, we're taking adjoints from data structures that use doubles and accumulating them in an array of varis.

## Checklist

- [x] Copyright holder: Columbia Unversity

### DIFF
--- a/stan/math/rev/functor/adj_jac_apply.hpp
+++ b/stan/math/rev/functor/adj_jac_apply.hpp
@@ -4,6 +4,8 @@
 #include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/rev/meta.hpp>
 #include <stan/math/rev/fun/value_of.hpp>
+#include <stan/math/rev/core/count_vars.hpp>
+#include <stan/math/rev/core/save_varis.hpp>
 #include <stan/math/prim/functor/apply.hpp>
 #include <stan/math/prim/fun/value_of.hpp>
 #include <stan/math/prim/fun/constants.hpp>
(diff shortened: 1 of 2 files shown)


bbbales2:
I think we should compute an actual type here: https://github.com/stan-dev/math/blob/8bc54219fe019dfd52011cbc0a7d3469271c7da7/stan/math/rev/functor/adj_jac_apply.hpp#L355

```
auto y_adj_jacs = f_.multiply_adjoint_jacobian(is_var_, y_adj);
```

If we just leave that as auto, then it will be hard for people to debug if they return the wrong types.

If we passed the return data structures as arguments this might make the errors clearer too (it will be a `multiply_adjoint_jacobian` not found, rather than a `cannot convert X to Y` error).

I dunno about the `accumulate_adjoints_in_varis` naming. Maybe `distribute_adjoints` just to be something different? I'm not sure this function generalizes outside of adj_jac_apply.

SteveBronder:
> If we just leave that as auto, then it will be hard for people to debug if they return the wrong types.

Would that mean parsing out the return tuple?

> If we passed the return data structures as arguments this might make the errors clearer too (it will be a multiply_adjoint_jacobian not found, rather than a cannot convert X to Y error).

idk, personally I don't like in/out parameters. I think failing a conversion is a more clear error than a signature failing (it could be any of the other arguments)

bbbales2:
> Would that mean parsing out the return tuple?

Well like instead of auto:

```
std::tuple<Eigen::MatrixXd, Eigen::VectorXd, double> out = ...
```

That way if someone returns the tuple of the wrong type, they'll get an interpretable error rather than the code just failing to compile or failing somewhere else.

> idk, personally I don't like in/out parameters

Sure we can stay with tuple.

SteveBronder:
Yeah but out can be a lot of different types right? That function returns back an `std::make_tuple(user_return_args)` So wouldn't you have to deduce those returns?

bbbales2:
It has to match `std::tuple<decltypes(value_of(Fargs))...>` I think.

They've gotta be the same arguments as `operator()` takes, but with doubles.

SteveBronder:
Oh that's weird, this saw I merged this branch into `feature/eigen-vari-ops` and closed it? That's super odd. Can we re-open it to merge to develop?

bbbales2:
Oh lol, weird. Eh, this will come through with the new stuff anyway. I don't see any need to merge it into develop independently.

Issue #1893, opened 2020-05-19
lognormal_lpdf returning wrong values with integer sigma


andrjohns:
## Description
As discovered by a user [on the forums](https://discourse.mc-stan.org/t/strange-behavior-with-lognormal-sampling-statement/15229/9), the ```lognormal``` distribution samples terribly when given an integer value for sigma, e.g.:
```
theta ~ lognormal(2.0,2);
```
This is because an integer sigma results in an integer division when calculating the inverse of sigma:
```
inv_sigma[n] = 1 / value_of(sigma_vec[n]);
```

#### Current Version:
v3.2.0

PR #1894, opened 2020-05-19
Bugfix: lognormal - wrong lpdf with integer sigma

andrjohns:
## Summary

As discovered by a user on the forums, the lognormal distribution samples terribly when given an integer value for sigma, e.g.:
```
theta ~ lognormal(2.0,2);
```
This is because an integer sigma results in an integer division when calculating the inverse of sigma:
```
inv_sigma[n] = 1 / value_of(sigma_vec[n]);
```
## Tests

Added test for lognormal distribution to check equality of lpdf with ```double``` and ```int``` sigma

## Side Effects

N/A

## Release notes
Fixed bug causing wrong log-probability values to be returned when using the lognormal distribution with an integer argument for sigma.

## Checklist

- [x] Math issue #1893 

### DIFF
--- a/stan/math/prim/prob/lognormal_lpdf.hpp
+++ b/stan/math/prim/prob/lognormal_lpdf.hpp
@@ -4,6 +4,7 @@
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/constants.hpp>
+#include <stan/math/prim/fun/inv.hpp>
 #include <stan/math/prim/fun/log.hpp>
 #include <stan/math/prim/fun/max_size.hpp>
 #include <stan/math/prim/fun/size.hpp>
@@ -63,7 +64,7 @@ return_type_t<T_y, T_loc, T_scale> lognormal_lpdf(const T_y& y, const T_loc& mu,
       inv_sigma_sq(size(sigma));
   if (include_summand<propto, T_y, T_loc, T_scale>::value) {
     for (size_t n = 0; n < stan::math::size(sigma); n++) {
-      inv_sigma[n] = 1 / value_of(sigma_vec[n]);
+      inv_sigma[n] = inv(value_of(sigma_vec[n]));
     }
   }
   if (include_summand<propto, T_y, T_loc, T_scale>::value) {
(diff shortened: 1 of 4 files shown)


t4c1:
`value_of_rec` converts `int`s to `double`s. So the glm is not affected. I think even in `lognormal_lpdf` we should instead replace `value_of` with `value_of_rec`, just in case someone replaced `1.0` with `1`. I know @bob-carpenter was promoting not having `.0` in code at one time.

stan/math/prim/prob/lognormal_lpdf.hpp
```diff
@@ -85,7 +87,7 @@ return_type_t<T_y, T_loc, T_scale> lognormal_lpdf(const T_y& y, const T_loc& mu,
       stan::math::size(y));
   if (!is_constant_all<T_y>::value) {
     for (size_t n = 0; n < stan::math::size(y); n++) {
-      inv_y[n] = 1 / value_of(y_vec[n]);
+      inv_y[n] = 1 / value_of_rec(y_vec[n]);
```
wds15, line 90: isn't it better to use "1.0 / ..." ? This way the expression is forced to be double.

Can you remind me of the difference of value_of and value_of_rec ?

andrjohns: ```value_of_rec``` will convert the int to a double, whereas ```value_of```` will just return the int. Not too sure about other differences beyond that, to be honest

andrjohns: I'm not particularly married to either approach, so happy to change either way

mcol: What about using `inv` (as you included the header anyway)?

mcol: Also `value_of_rec` will not do what you expect for "nested" types like `fvar<var<double>`, as it would return `double`, while `value_of` would return `var<double>`. I haven't checked, but that may be at the bottom of the failure in the distribution tests.

andrjohns: Ah good call, I'll make that change

bob-carpenter:
I was promoting not having 0.0 in code because we should be handling integers appropriately.  We don't want a situation where 0.0 works but 0 doesn't.

inv(x) will be better than 1 / x if x is an autodiff variable.

andrjohns:
This is ready for review

PR #1902, opened 2020-05-20
[Draft] Feature/profiling

rok-cesnovar:
I have been circling around this idea of profiling the auto diff tree for a while and finally made an implementation that compiles and at least seems to return the correct results.

The idea is to profile the autodiff this way
```cpp
// profiling info map
stan::math::profiles profiles;

// start profiling for ID = 0
start_profiling(0, profiles);
matrix_v L = cholesky_decompose(Y) + Y;
stop_profiling(0, profiles);

// start profiling for ID = 1
start_profiling(1, profiles);
matrix_v B = L * A + L * A;
matrix_v C = B * transpose(B);
stop_profiling(1, profiles);
C(0,0).grad();

stan::math::print_profiling(profiles);
```

Just throwing up a draft if anyone has time/energy to comment if it makes any sense or am I striking out :). This does waste var's even if what it is profiling is /prim functions, but not sure if

### DIFF
--- a/stan/math/prim/fun/typedefs.hpp
+++ b/stan/math/prim/fun/typedefs.hpp
@@ -28,6 +28,13 @@ using vector_d = Eigen::Matrix<double, Eigen::Dynamic, 1>;
  */
 using row_vector_d = Eigen::Matrix<double, 1, Eigen::Dynamic>;
 
+typedef struct {
+    double fwd_pass_time;
+    double bckwd_pass_time;
+} profile;
+
+using profiles = std::map<int, profile>;
+
 }  // namespace math
 }  // namespace stan
 
--- a/stan/math/prim/fun/typedefs.hpp
+++ b/stan/math/prim/fun/typedefs.hpp
@@ -29,8 +29,10 @@ using vector_d = Eigen::Matrix<double, Eigen::Dynamic, 1>;
 using row_vector_d = Eigen::Matrix<double, 1, Eigen::Dynamic>;
 
 typedef struct {
-    double fwd_pass_time;
-    double bckwd_pass_time;
+    std::chrono::time_point<std::chrono::steady_clock> fwd_pass_time_start;
+    std::chrono::time_point<std::chrono::steady_clock> fwd_pass_time_stop;
+    std::chrono::time_point<std::chrono::steady_clock> bkcwd_pass_time_start;
+    std::chrono::time_point<std::chrono::steady_clock> bkcwd_pass_time_stop;
 } profile;
 
 using profiles = std::map<int, profile>;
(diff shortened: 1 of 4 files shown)


wds15:
I would have expected that a call to grad is inside the profiling thing...no?

syclik:
Mind posting a more complete example? I'm having trouble following what the different things are in the snippet.

It definitely solves a problem that I couldn't wrap my head around: how to time the forward and backward pass for the same expression (or group of expressions). But if I understand correctly, this will handle it? Now that I think about it, it makes sense... you should be able to tell where on the stack you are in the backwards sweep if you assume things get called in the same order.

rok-cesnovar:
> I would have expected that a call to grad is inside the profiling thing...no?

If I am looking to profile in the log_prob evaluation in a model I think this might be easier to handle.

> But if I understand correctly, this will handle it?

That is what I am targeting yes.

> Mind posting a more complete example?

This is a complete example at the C++ level:

```cpp
#include <stan/math/rev.hpp>
#include <boost/random/mersenne_twister.hpp>
#include <cstdlib>


int main(int argc, char** argv) {
    using stan::math::matrix_d;
    using stan::math::matrix_v;
    using stan::math::var;
    using stan::math::wishart_rng;
    using stan::math::cholesky_decompose;
    using stan::math::start_profiling;
    using stan::math::stop_profiling;
    
    int size = 300;
    if (argc > 1) {
        size = atoi(argv[1]);
    }

    // init input
    boost::random::mt19937 rng;
    matrix_d I(size, size);
    I.setZero();
    I.diagonal().setOnes();
    matrix_v Y = wishart_rng(5000, I, rng);
    matrix_v A = Eigen::VectorXd::Ones(size);

    // profiling info map
    stan::math::profiles profiles;

    // start profiling for ID = 0
    start_profiling(0, profiles);
    matrix_v L = cholesky_decompose(Y) + Y;
    stop_profiling(0, profiles);

    // start profiling for ID = 1
    start_profiling(1, profiles);
    matrix_v B = L * A + L * A;
    matrix_v C = B * transpose(B);
    stop_profiling(1, profiles);
    C(0,0).grad();

    stan::math::print_profiling(profiles);
}
```

rok-cesnovar:
As we discussed at the math meeting this could also profile other stuff like the amount of stack used etc.

rok-cesnovar:
This looks promising enough that I am going to make a post on how this could be used through to the Stan level. If that gets any traction then its real work time 🎉

rok-cesnovar:
See https://discourse.mc-stan.org/t/proposal-for-profiling-stan-models/15322/2

rok-cesnovar:
Will redo with suggestions on the forums and then reopen.

Issue #1890, opened 2020-05-18
Binary vectorisation framework


andrjohns:
## Description

We currently have a framework for applying unary scalar functions to containers, and containers of containers (```apply_scalar_unary```), but we don't have a similar implementation for binary scalar functions.

I've worked up an ```apply_scalar_binary``` framework which will work with combinations of scalars, eigen objects/expressions, ```std::vectors```, and ```std::vectors``` of these. The code is up on [this branch](https://github.com/stan-dev/math/compare/develop...andrjohns:feature/scalar_binary).

For example, given the following code:
```
template <typename T1, typename T2>
inline auto binary_log_loss(T1 y, const T2& y_hat) {
  return apply_scalar_binary<T1, T2>::apply(
      y, y_hat, [&](const auto& v, const auto& v_hat) {
        return v ? -log(v_hat) : -log1m(v_hat);
      });
}
```
The following signatures are available
```
binary_log_loss(int, double)
binary_log_loss(int, VectorXd)
binary_log_loss(VectorXi, VectorXd)
binary_log_loss(VectorXi, double)
binary_log_loss(VectorXi, MatrixXd.diagonal()); // Or other expression/type
binary_log_loss(std::vector<int>, double)
binary_log_loss(int, std::vector<double>)
binary_log_loss(std::vector<int>, std::vector<double>)
```
The containers can also be nested, so ```std::vector<VectorXd>``` or ```std::vector<std::vector<double>>```inputs can also be also be used

I need to add documentation to the framework, and look at working out a binary equivalent of the ```expect_unary_vectorized``` autodiff testing framework, then this will be ready for more eyes. 

#### Current Version:
v3.2.0

andrjohns:
The unary framework actually already works with expression templates, since [unaryExpr is used](https://github.com/stan-dev/math/blob/19f5f25bf9de1d267c92522d68eccfd0d73e12d0/stan/math/prim/meta/apply_scalar_unary.hpp#L63) and the input templating will accept any eigen types

andrjohns:
Oh, you meant the testing framework, not apply_scalar_unary

andrjohns:
@bob-carpenter @bbbales2 Can I get an opinion on something with the autodiff testing framework?

I'm working up the methods for testing the vectorised binary functions, but they're taking oddly long to run. At the moment the tests basically comprise:
```
  expect_ad(tols, f, VectorXd, VectorXd);
  expect_ad(tols, f, VectorXd, double);
[...]

PR #1907, opened 2020-05-24
Issue 1890 - Binary vectorization framework

andrjohns:
## Summary

This pull introduces a framework for the simple vectorisation/broadcasting of binary scalar functions, as well as frameworks for testing that the values and gradients are consistent between scalar and vector inputs.

To use the framework with a given function, the existing specialisations are restricted to only take in scalar inputs (using either the ```require_``` generics or explicit specialisations). A call to ```apply_scalar_binary``` is then added to the ```prim``` header for that function, which is used when any of the inputs are containers. Using ```beta(a, b)``` as an example, this looks like:
```
inline auto beta(const T1& a, const T2& b) {
  return apply_scalar_binary<T1, T2>::apply(
      a, b, [&](const auto& c, const auto& d) { return beta(c, d); });
}
```
This call will delegate to the appropriate definition of ```beta```, depending on whether the container holds arithmetic, ```var```, or ```fvar<T>``` types. 
 
## Tests
 
Two testing frameworks are added, one for arithmetic types and one for autodiff types.

### Arithmetic Testing

The ```binary_scalar_tester``` framework has been added, which tests that the vectorised version of a given binary function returns the same values as looping over the inputs and iteratively applying the function. The framework is called using a functor that implements the function to be called, and two Eigen column vectors. These column vectors are tested, and then transformed to Eigen row vectors, Eigen matrices, Eigen arrays, Eigen expressions, and std::vectors (as well as std::vectors of each of these) to also be tested.

Using ```beta(a, b)``` as an example, the framework is called like so:
```
  auto f = [](const auto& x1, const auto& x2) {
    return stan::math::beta(x1, x2);
  };

  Eigen::VectorXd in1 = Eigen::VectorXd::Random(6);
  Eigen::VectorXd in2 = Eigen::VectorXd::Random(6);

  stan::test::binary_scalar_tester(f, in1, in2);
```

### Autodiff Testing
The ```expect_ad_vectorized_binary``` framework has been added, which tests that the values and gradients are consistent with finite differences for Eigen vector and std::vector inputs (and std::vectors of these).

Using ```beta(a, b)``` as an example, the framework is called like so:
```
  auto f = [](const auto& x1, const auto& x2) {
    using stan::math::beta;
    return beta(x1, x2);
  };

  Eigen::VectorXd in1 = Eigen::VectorXd::Random(3);
  Eigen::VectorXd in2 = Eigen::VectorXd::Random(3);
  stan::test::expect_ad_vectorized_binary(f, in1, in2);
```

## Side Effects

N/A

## Release notes

Added framework for vectorizing/broadcasting binary scalar functions, and utilities for testing the values and gradients of these

## Checklist

- [x] Math issue #1890 


andrjohns:
The distribution test failures are caused by #1908, so will need to wait on that before this can be merged

andrjohns:
This is ready for review

stan/math/prim/meta/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+/**
+ * Specialisation for use with one Eigen input and one scalar. Eigen's
+ * unaryExpr framework is used for more efficient indexing of both row-
+ * and column-major inputs. The unaryExpr framework also allows for the
+ * scalar to be captured and applied to each element in the Eigen object.
+ */
+template <typename T1, typename T2>
+struct apply_scalar_binary<T1, T2, require_any_eigen_t<T1, T2>,
+                           require_any_stan_scalar_t<T1, T2>> {
```
t4c1, line 82: How did you decide when to speciallize the `apply_scalar_binary` struct and when to overload the `apply` function in the struct? I would like one of those options used consistently. 

Actually, since you use requires to select the overload, we don't need a struct at all. You could just overload a free function. That would also let you save a few lines on docs.

stan/math/prim/meta/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,326 @@
...
+            require_std_vector_vt<is_container, T1_2>* = nullptr,
+            require_stan_scalar_t<T2_2>* = nullptr>
+  static inline auto apply(const T1& x, const T2& y, const F& f) {
+    using T1_vt = value_type_t<T1>;
+    using T_return
+        = decltype(apply_scalar_binary<T1_vt, T2>::apply(x[0], y, f));
+    size_t x_size = x.size();
+    std::vector<T_return> result(x_size);
+    for (size_t i = 0; i < x_size; ++i)
+      result[i] = apply_scalar_binary<T1_vt, T2>::apply(x[i], y, f);
```
t4c1, line 293: Please use braces consistently, even if a body of a for/if is just one line.

test/unit/math/prim/fun/binary_scalar_tester.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * @tparam T2 Type of second vector.
+ * @param x First vector input to which operation is applied.
+ * @param y Second vector input to which operation is applied.
+ * @param f functor to apply to inputs.
+ */
+template <typename F, typename T1, typename T2,
+          require_all_std_vector_t<T1, T2>* = nullptr>
+void binary_scalar_tester_impl(const F& f, const T1& x, const T2& y) {
+  auto vec_vec = f(x, y);
+  auto vec_scal = f(x, y[1]);
```
t4c1, line 79: Why dou you use the second element for the scalar? Would it not make more sence tu use the first one, so this can also be called with a vector of length 1?

stan/math/prim/meta/apply_scalar_binary.hpp
andrjohns: Good idea, will do

stan/math/prim/meta/apply_scalar_binary.hpp
andrjohns: Fixed

test/unit/math/prim/fun/binary_scalar_tester.hpp
andrjohns: This was just a side-effect of spending too much time in R for work, now fixed.

stan/math/prim/meta/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,280 @@
...
+ * @param x First std::vector input to which operation is applied.
+ * @param y Second std::vector input to which operation is applied.
+ * @param f functor to apply to std::vector inputs.
+ * @return std::vector with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
+          require_all_std_vector_vt<is_container, T1, T2>* = nullptr>
+inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+  using T1_vt = value_type_t<T1>;
+  using T2_vt = value_type_t<T2>;
```
t4c1, line 212: These two seem to be unused.

stan/math/prim/meta/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,280 @@
...
+ * @param x std::vector input to which operation is applied.
+ * @param y Scalar input to which operation is applied.
+ * @param f functor to apply to inputs.
+ * @return std::vector with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
+          require_std_vector_vt<is_container, T1>* = nullptr,
+          require_stan_scalar_t<T2>* = nullptr>
+inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+  using T1_vt = value_type_t<T1>;
```
t4c1, line 240: Another unused alias.

stan/math/prim/meta/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,280 @@
...
+ * @param x Scalar input to which operation is applied.
+ * @param y std::vector input to which operation is applied.
+ * @param f functor to apply to inputs.
+ * @return std::vector with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
+          require_stan_scalar_t<T1>* = nullptr,
+          require_std_vector_vt<is_container, T2>* = nullptr>
+inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+  using T2_vt = value_type_t<T2>;
```
t4c1, line 268: another one

bob-carpenter:
Sorry for the spam, but I just had to jump in and say this is awesome!  I've wanted to do this for years and found the project too daunting.

Where we really need it pushed through the language is in `pow(a, b)` and the syntactic sugar `a^b`.  

Will this also work for complex inputs for things like `pow`?  That'd be really amazing.

andrjohns:
Thanks! I just tried to add this to ```pow```, but I'm running into a weirder problem where ```pow(var, var)``` returns ```-nan``` for no reason that I can see.

I added the following to the mix tests for ```pow``` (on develop):
```
TEST(mathMixScalFun, pow_comp) {
  stan::math::var in_var1(-0.4);
  stan::math::var in_var2(0.5);

  std::cout << "var1 = " << in_var1.val() << ", var2 = " << in_var2.val() << std::endl
            << "pow = " << stan::math::pow(in_var1, in_var2).val() << std::endl;
}
```
And the output is:
```
var1 = -0.4, var2 = 0.5
pow = -nan
```

Any idea where this might be coming from?

andrjohns:
Oh never mind, ```std:pow``` returns ```-nan``` for fractional exponents

andrjohns:
Fractional exponents with a negative base, I mean

andrjohns:
The vectorisation framework is working without issue for the mix tests of ```var/fvar```, but I can't test with vectorised complex inputs properly since the testing framework doesn't play nicely with containers of complex types (the deserializer is trying to a copy an ```std::complex<double>``` to a container of ```double```.

I've done some basic tests for ```std:complex``` types of ```var```,```fvar<double>```,```fvar<fvar<double>>```, and ```fvar<fvar<var>>```, and all of the values and derivatives are coming out equivalent. So I would say that the vectorisation framework is (most probably) working without a hitch for ```pow``` and ```std::complex<T>``` types.

bob-carpenter:
I should've left breadcrumbs.  I went through the same line of "uh oh" and then "d'oh" as I remembered how `pow(negative, fraction)` fails.  

And doubly cool it works for complex types.  I think it has a lot to do with all the cleaned up metaprograms.

Issue #1908, opened 2020-05-25
Fwd definition of beta function not called by distributions


andrjohns:
## Description
Currently, the ```beta``` function in the distribution tests only works with ```fvar<T>``` inputs through autodiff-ing the ```prim``` definition, not through calling the ```fwd``` definition.

## Example
This can be verified by restricting the ```prim``` inputs to be arithmetic:
```
template <typename T1, typename T2,
          require_all_arithmetic_t<T1, T2>* = nullptr>
inline return_type_t<T1, T2> beta(const T1 a, const T2 b) {
  using std::exp;
  return exp(lgamma(a) + lgamma(b) - lgamma(a + b));
}
```

And calling a distribution test which uses the ```beta``` function:
```
make -j16 test/prob/neg_binomial_2/neg_binomial_2_cdf_log_00000_generated_ffv_test
```
Which returns:
```
./stan/math/prim/prob/beta_lcdf.hpp:97:27: error: no matching function for call to ‘beta(const T_partials_return&, const T_partials_return&)’
   97 |         = stan::math::beta(alpha_dbl, beta_dbl);
```
When the input types are:
```
T_location = stan::math::fvar<stan::math::fvar<stan::math::var> >; T_precision = stan::math::fvar<stan::math::fvar<stan::math::var> >;
```

#### Current Version:
v3.2.0

andrjohns:
The failing code belongs to the ```beta_lcdf``` header, and this section specifically:
```
    const T_partials_return betafunc_dbl
        = stan::math::beta(alpha_dbl, beta_dbl);
```
The namespace qualifier on beta is needed here as one of the input parameters is also called ```beta```. However, it looks like this is inhibiting the proper lookup of the ```fwd``` definition. If I rename the ```beta``` parameter to ```beta_param``` and remove the namespace qualifier from ```beta```, as  in:
```
    const T_partials_return betafunc_dbl = beta(alpha_dbl, beta_dbl);
```
Then the distribution tests compile without issue.

bob-carpenter:
Thanks for the explanation---now I see why you called it `_param`---to separate from the function name.

PR #1909, opened 2020-05-25
Issue 1908 - Fwd definition of beta function not always called

andrjohns:
## Summary

This PR fixes the issue where distributions that call the beta function with a namespace qualifier (i.e., ```stan::math::beta(a,b)```) don't delegate to the ```fwd``` definition for ```fvar``` inputs, rather relying on autodiff-ing the ```prim``` function.

By removing the namespace qualifier from the call to beta, the correct (```fwd```) definition is called.

## Tests
No tests added, as this PR fixes existing failures.

As shown in #1908, current tests will fail when the ```prim``` definition of ```beta``` is restricted to only take arithmetic inputs. This PR introduces that restriction, and tests pass with the added changes.

## Side Effects

N/A

## Release notes

Fixed a bug where the appropriate specialization for the beta function was not called for forward-mode autodiff variables

## Checklist

- [x] Math issue #1908

### DIFF
--- a/stan/math/prim/fun/beta.hpp
+++ b/stan/math/prim/fun/beta.hpp
@@ -48,7 +48,8 @@ namespace math {
  * @param b Second value
  * @return Beta function applied to the two values.
  */
-template <typename T1, typename T2>
+template <typename T1, typename T2,
+          require_all_arithmetic_t<T1, T2>* = nullptr>
 inline return_type_t<T1, T2> beta(const T1 a, const T2 b) {
   using std::exp;
   return exp(lgamma(a) + lgamma(b) - lgamma(a + b));
--- a/stan/math/prim/fun/beta.hpp
+++ b/stan/math/prim/fun/beta.hpp
@@ -48,8 +48,7 @@ namespace math {
  * @param b Second value
  * @return Beta function applied to the two values.
  */
-template <typename T1, typename T2,
-          require_all_arithmetic_t<T1, T2>* = nullptr>
+template <typename T1, typename T2, require_all_arithmetic_t<T1, T2>* = nullptr>
 inline return_type_t<T1, T2> beta(const T1 a, const T2 b) {
   using std::exp;
   return exp(lgamma(a) + lgamma(b) - lgamma(a + b));
--- a/stan/math/prim/prob/beta_lccdf.hpp
+++ b/stan/math/prim/prob/beta_lccdf.hpp
@@ -29,7 +29,7 @@ namespace math {
  * @tparam T_scale_fail type of failure parameter
  * @param y (Sequence of) scalar(s) between zero and one
  * @param alpha (Sequence of) success parameter(s)
- * @param beta (Sequence of) failure parameter(s)
+ * @param beta_param (Sequence of) failure parameter(s)
  * @return log probability or sum of log of probabilities
  * @throw std::domain_error if alpha or beta is negative
  * @throw std::domain_error if y is not a valid probability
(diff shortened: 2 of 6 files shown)


stan/math/prim/fun/beta.hpp
```diff
@@ -48,7 +48,7 @@ namespace math {
  * @param b Second value
  * @return Beta function applied to the two values.
  */
-template <typename T1, typename T2>
+template <typename T1, typename T2, require_all_arithmetic_t<T1, T2>* = nullptr>
```
bob-carpenter, line 51: [question]
I'm confused as to why we have seemingly three different ways to do this:

1.  `typename = require_all_arithmetic_t<...>`
2. `require_all_arithmetic_t<...>...`
3. `require_all_arithmetic_t<T1, T2>* = nullptr`

I'd rather just settle on one unless there's an important difference.  @SteveBronder?

stan/math/prim/prob/beta_lccdf.hpp
```diff
@@ -29,43 +29,43 @@ namespace math {
...
+ * @param beta_param (Sequence of) failure parameter(s)
  * @return log probability or sum of log of probabilities
  * @throw std::domain_error if alpha or beta is negative
  * @throw std::domain_error if y is not a valid probability
  * @throw std::invalid_argument if container sizes mismatch
  */
 template <typename T_y, typename T_scale_succ, typename T_scale_fail>
 return_type_t<T_y, T_scale_succ, T_scale_fail> beta_lccdf(
-    const T_y& y, const T_scale_succ& alpha, const T_scale_fail& beta) {
+    const T_y& y, const T_scale_succ& alpha, const T_scale_fail& beta_param) {
```
bob-carpenter, line 40: `alpha` and `beta` are the same type of argument here, so either they should both get decorated with `_param` or neither should.  I'd prefer not to decorate conventional names like this, but I won't complain if you decide to add `_param` to `alpha`, too.  Shouldn't `y` then also be decorated with `y_outcome` or something like that?

stan/math/prim/prob/neg_binomial_lccdf.hpp
```diff
@@ -93,7 +93,7 @@ return_type_t<T_shape, T_inv_scale> neg_binomial_lccdf(
     const T_partials_return p_dbl = beta_dbl * inv_beta_p1;
     const T_partials_return d_dbl = square(inv_beta_p1);
     const T_partials_return Pi = 1.0 - inc_beta(alpha_dbl, n_dbl + 1.0, p_dbl);
-    const T_partials_return beta_func = stan::math::beta(n_dbl + 1, alpha_dbl);
+    const T_partials_return beta_func = beta(n_dbl + 1, alpha_dbl);
```
bob-carpenter, line 96: [question]
Any idea why the right version wasn't getting picked up with the qualifier?  It should have all the candidates in scope as long as they've been included and if they haven't been included, I don't see how the unqualified version could work.

bob-carpenter:
Nothing stopping the merge here, so I approved.  Feel free to merge when tests pass.

stan/math/prim/fun/beta.hpp
SteveBronder: Let's stick with `require_all_arithmetic_t<T1, T2>* = nullptr`

bob-carpenter: I'm OK with any of (1) -- (3).  Should we always use the `ptr = nullptr` idiom?  I was wondering if there were differences in the approaches.

t4c1: 1. `typename = ` cannot be used for function overloads that opnly differ in requires. 
2. `...` can not be used on classes that need more than one require. Although I don't think we use requires for classes much.
3. `* = nullptr` has AFAIK no such restrictions. 

So either 2. or 3. is fine. 2. is the shortest, but so far we have used 3. the most.

PR #1914, opened 2020-05-29
Add holder

t4c1:
## Summary
Adds a holder operation for Eigen expressions. This is a no-op operation that owns some pointers to heap and deletes them when it is destroyed. It can be used in functions returning expressions to make sure matrices these expressions reference (local variables or rvalue arguments) do not go out of scope.

Existing functions returning expressions (`value_of`, `value_of_rec`, `as_column_vector_or_scalar` and `as_array_or_scalar`) now use holder. That makes them work correctly with rvalue inputs.

## Tests
Holder is tested. Added tests for modified functions that check that they work with rvalues. `as_column_vector_or_scalar` and `as_array_or_scalar` previously had no tests so they were added.

## Side Effects
None.

## Release notes

Added holder operation for Eigen expressions. `value_of`, `value_of_rec`, `as_column_vector_or_scalar` and `as_array_or_scalar` now work with rvalue inputs.

## Checklist

- [x] Math issue #1775


andrjohns:
@tadej, I'll review this one later this arvo (aus time)

andrjohns:
So at the moment (as I understand), each individual function gets wrapped in a call to ```make_holder```, where the function itself is defined by a lambda. I think that approach might not be the most readable, and is also bit repetitive in the coding. One approach that could be bit simpler would be to have single ```make_holder``` call that captures all definitions for a given function (across ```prim```/```rev```/```fwd```).

Using ```as_column_vector_or_scalar``` as an example, this could look like:
```
template <typename T, require_stan_scalar_t<T>* = nullptr>
inline T as_column_vector_or_scalar(const T& a) {
  return a;
}

template <typename T, require_t<is_eigen_col_vector<T>>* = nullptr>
inline T&& as_column_vector_or_scalar(T&& a) {
  return std::forward<T>(a);
}

template <typename T, require_t<is_eigen_row_vector<T>>* = nullptr>
inline auto as_column_vector_or_scalar_impl(T&& a) {
  return a.transpose();
}

inline auto as_column_vector_or_scalar_impl(T&& a) {
  using plain_vector = Eigen::Matrix<value_type_t<T>, Eigen::Dynamic, 1>;
  using optionally_const_vector
      = std::conditional_t<std::is_const<std::remove_reference_t<T>>::value,
                           const plain_vector, plain_vector>;
  return Eigen::Map<optionally_const_vector>(a.data(), a.size());
}

template <typename T,
          require_not_t<disjunction<is_eigen_col_vector<T>,
                                    is_stan_scalar<T>>>* = nullptr>
inline auto as_column_vector_or_scalar(T&& a) {
  return make_holder([](auto& x) { return as_column_vector_or_scalar_impl(x); },
                     std::forward<T>(a));
}
```

This way, any existing definitions that need to be ```Holder```-ed just need an ```_impl``` appended, and a single ```Holder``` definition can capture all of them. What do you think?

t4c1:
My way has to duplicate `return make_holder([](auto& x) { ` and `; }, std::forward<T>(a)` for each overload of a function. It is not that much code.

Four your suggestion we have to add an extra function. This might be less code if the number of overloads for a function is high enough. We could make a macro for that, but ... well ... macros. I don't think we have that many overloads that would need holder for most functions or that many functions that would warrant a macro.

If someone searches for implementation of certain function he has to go over one more function signature. Honestly I don't think it makes that much of a difference. Your suggestion is fine, but I still slightly prefere my way as I find it a bit more readable.

t4c1:
Actually your suggestion has another downside. If we have a function that must use holder to store both a rvalue argument and a local variable we would have to split its implementation in two parts, which again reduces readability.

andrjohns:
Yeah that's a good point, hadn't thought of the rvalue complexities, sounds good to me

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+template <class ArgType, typename... Ptrs>
+struct traits<stan::math::Holder<ArgType, Ptrs...>> {
+  typedef typename ArgType::StorageKind StorageKind;
+  typedef typename traits<ArgType>::XprKind XprKind;
+  typedef typename ArgType::StorageIndex StorageIndex;
+  typedef typename ArgType::Scalar Scalar;
+  enum {
+    Flags = (ArgType::Flags
+             & (RowMajorBit | LvalueBit | LinearAccessBit | DirectAccessBit
```
andrjohns, line 30: Isn't ```RowMajorBit``` redundant given that you've also specified ```NoPreferredStorageOrderBit```?

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+ * @tparam T type of the result expression
+ * @tparam Is index sequence for `ptrs`
+ * @tparam Args types of pointes to heap
+ * @param expr result expression
+ * @param ptrs pointers to heap that need to be released when the expression is
+ * destructed
+ * @return `holder` referencing given expression
+ */
+template <typename T, std::size_t... Is, typename... Args>
+auto make_holder_impl2(T&& expr, std::index_sequence<Is...>,
```
andrjohns, line 225: I'd probably go with a more descriptive name here, if possible (even if just ```step_2``` rather than ```2```

stan/math/prim/meta/holder.hpp
t4c1: They are two separate bits that are possible in flags. `NoPreferredStorageOrderBit`'s doc says it means transpositions are never needed when working with this expression or something like this.

bbbales2:
I had a look at this, not being familiar with Eigen I also am nervous about doing a super formal review.

So I'll try to provide comments on the tests.

For this code:
```
const auto& tmp = value_of(std::move(a));
a.setZero();
```

Why does `a.setZero()` work? Didn't it give it's memory to the `r_value`? If not, then wouldn't this code work with the previous implementation too?

Also, if the `value_of` implementation switched to returning `MatrixXd` and not expressions, I think this code would still work.

Given that at this point we expect expressions, I think this test should fail if they aren't there.

I tried writing things in a way they would fail, but I didn't actually check that this fails in the current develop (I think it should before it's really the right test):

```
TEST(AgradMatrixRev, value_of_matrix_rvalue2) {
  using Eigen::Matrix;
  using Eigen::MatrixXd;
  using stan::math::value_of;
  using stan::math::var;

  Matrix<var, -1, -1> *a = new Matrix<var, -1, -1>(MatrixXd::Random(7, 4));

  MatrixXd correct = value_of(*a);
  auto expr = value_of(std::move(*a));

  delete a;

  MatrixXd res = expr;

  // Is there an explicit way to check that something is an Eigen expression?                                                                                                
  EXPECT_FALSE((std::is_same<std::decay_t<decltype(expr)>, std::decay_t<decltype(res)>>::value));

  // I don't know how to do this other than access the memory. But it's not                                                                                                  
  //  guaranteed there will be an error if we do this. I can allocate another                                                                                                
  //  matrix before doing the access, but I don't think this guarantees                                                                                                      
  //  anything.                                                                                                                                                              
  MatrixXd overwrite = MatrixXd::Random(7, 4);
  EXPECT_MATRIX_NEAR(res, correct, 1e-10);
}
```

t4c1:
> Why does a.setZero() work? Didn't it give it's memory to the r_value?

If everything is working as intended a does not own any memory. However, after being moved from it must be left in some unknown, but valid state (for Eigen matrices I think it only makes sense to make it into 0x0 matrix). So method calls on it must work.

> Didn't it give it's memory to the r_value?

That is exactly what this test is checking. If it did not and the expression tmp is still referencing `a`, the result `res` will be wrong after a is set to zero.

> Also, if the value_of implementation switched to returning MatrixXd and not expressions, I think this code would still work.

You are right here. This test just checks that it works as expected when given a rvalue input.

> // Is there an explicit way to check that something is an Eigen expression?

We can check that it is an eigen type, but not a plain object. But we don't have an `is_expression` trait and I don't think we need it, expect maybe in some tests.

I will add a test for checking expressions, but I think I can make it a bit simpler than what you suggested - I think that dynamic allocation is not necessary.

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+};
+
+}  // namespace math
+}  // namespace stan
+
+namespace Eigen {
+namespace internal {
+
+template <typename ArgType, typename... Ptrs>
+struct evaluator<stan::math::Holder<ArgType, Ptrs...>>
```
SteveBronder, line 102: I'm looking at the core evaluators in Eigen to check this. Is there a different evaluator I should be referencing here?

SteveBronder: Is this based off of unaryevaluator?

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+    : evaluator_base<stan::math::Holder<ArgType, Ptrs...>> {
+  typedef stan::math::Holder<ArgType, Ptrs...> XprType;
+  typedef typename remove_all<typename nested_eval<ArgType, 1>::type>::type
+      ArgTypeNestedCleaned;
```
SteveBronder, line 106: Is there a reason not to have 

```
typedef typename internal::nested_eval<ArgType,Factor>::type ArgTypeNested;
```

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+  template <int StoreMode, typename PacketType>
+  EIGEN_STRONG_INLINE void writePacket(Index row, Index col,
+                                       const PacketType& x) {
+    return m_argImpl.template writePacket(row, col, x);
+  }
+  template <int StoreMode, typename PacketType>
+  EIGEN_STRONG_INLINE void writePacket(Index index, const PacketType& x) {
+    return m_argImpl.template writePacket(index, x);
+  }
```
SteveBronder, line 152: I'm confused on why `packet` needs to pass along the templates but `writepacket` does not. Looking at some of the stuff based on `unary_evaluator` they pass writePacket at least the `StoreMode`

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+ * @return holder operation
+ */
+template <typename T, typename... Ptrs,
+          std::enable_if_t<sizeof...(Ptrs) >= 1>* = nullptr>
+Holder<T, Ptrs...> holder(T&& arg, Ptrs*... pointers) {
+  return Holder<T, Ptrs...>(std::forward<T>(arg), pointers...);
+}
+// trivial case with no pointers constructs no holder object
+template <typename T>
+T holder(T&& arg) {
```
SteveBronder, line 178: Won't `T` as the return make a copy here?

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+ * @tparam T type of the element
+ * @param a element to handle
+ * @param res resulting pointer to element
+ * @return tuple of pointer allocated on heap (containing single pointer).
+ */
+template <typename T>
+auto holder_handle_element(std::remove_reference_t<T>&& a, T*& res) {
+  res = new T(std::move(a));
+  return std::make_tuple(res);
+}
```
SteveBronder, line 212: I'm not following why the `std::remove_reference_t` here?

SteveBronder: Also when does the `res` pointer get cleaned up?

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+template <typename T, std::size_t... Is, typename... Args>
+auto make_holder_impl_step1(const T& func, std::index_sequence<Is...>,
+                            Args&&... args) {
+  std::tuple<std::remove_reference_t<Args>*...> res;
+  auto ptrs = std::tuple_cat(
+      holder_handle_element(std::forward<Args>(args), std::get<Is>(res))...);
+  return make_holder_impl_step2(
+      func(*std::get<Is>(res)...),
+      std::make_index_sequence<std::tuple_size<decltype(ptrs)>::value>(), ptrs);
+}
```
SteveBronder, line 248: This needs to be more clear than first and second step. It looks like the first step here is to evaluate the function on the args, where args are first checked for rvalueness and if so copied onto the heap. `eval_and_store_holder_impl`?`

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+ * @param expr result expression
+ * @param ptrs pointers to heap that need to be released when the expression is
+ * destructed
+ * @return `holder` referencing given expression
+ */
+template <typename T, std::size_t... Is, typename... Args>
+auto make_holder_impl_step2(T&& expr, std::index_sequence<Is...>,
+                            const std::tuple<Args*...>& ptrs) {
+  return holder(std::forward<T>(expr), std::get<Is>(ptrs)...);
+}
```
SteveBronder, line 228: Is there a standard place to put boilerplate stuff like index_sequence? I always get confused where to put them and do either the front or end

stan/math/prim/meta/holder.hpp
SteveBronder: If you just use `T&&` here could you just use an `enable_if_t<is_rvalue<>>` instead of this?

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+      std::make_index_sequence<std::tuple_size<decltype(ptrs)>::value>(), ptrs);
+}
+
+}  // namespace internal
+
+/**
+ * Constructs an expression from given arguments using given functor.
+ * This is similar to calling the functor with given arguments. Except that any
+ * rvalue argument will be moved to heap first. The arguments moved to heap are
+ * deleted once the expression is destructed.
```
SteveBronder, line 256: I'm getting confused on who is in charge of deleting this

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+ * @tparam T type of functor
+ * @tparam Args types of arguments
+ * @param func the functor
+ * @param args arguments for the functor
+ */
+template <typename T, typename... Args,
```
SteveBronder, line 262: Think `T` should be `F` here so it's more clear this is a function template parameter

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,274 @@
...
+          require_eigen_t<
+              decltype(std::declval<T>()(std::declval<Args&>()...))>* = nullptr>
```
SteveBronder, line 264: Is this doing the same thing as `is_invocable` in C++17? Not necessary but could be good to pull out as it's own backport so this is cleaned up and we can use it elsewhere

stan/math/prim/meta/holder.hpp
t4c1: Tbh I have no idea. I was just following Eigen's tutorial on adding new expressions: https://eigen.tuxfamily.org/dox/TopicNewExpressionType.html

stan/math/prim/meta/holder.hpp
t4c1: It was that way in the tutorial I followed. I removed it since it was only used once. What is the reason you would prefere it that way?

stan/math/prim/meta/holder.hpp
t4c1: It probably has some defaults. Those probably should be passed along.

stan/math/prim/meta/holder.hpp
t4c1: For lvalue input T will be lvalue ref, so no. For rvalue input yes, but only of the compiler does not do RVO.

stan/math/prim/meta/holder.hpp
t4c1: No idea.

stan/math/prim/meta/holder.hpp
t4c1: Holder object is.

stan/math/prim/meta/holder.hpp
t4c1: Yep, it is the same.

stan/math/prim/meta/holder.hpp
t4c1: I guess at the end makes most sense.

t4c1:
Here is benchmark result for value_of using holder vs eval:
```
-------------------------------------------------------------------
Benchmark                            Time           CPU Iterations
-------------------------------------------------------------------
value_of_bench/1000000        35656319 ns   34226190 ns         21
value_of_eval_bench/1000000   46436053 ns   44791667 ns         15
```
Benchmark code is here: https://gist.github.com/t4c1/d891190ca084368d705d634aa343f868
This includes time of constructing two matrices of var, so actual speedup of `value_of` is even larger.

stan/math/prim/meta/holder.hpp
SteveBronder: Yes good call!

https://godbolt.org/z/nzkf2m

SteveBronder:
I tried taking yours and writing it out for vectors and matrices of different sizes but it throws a segfault for `matrix_value_of_bench`, am I doing something wrong here? It's a bit of a mess but I have the gdb backtrace [here](https://gist.github.com/SteveBronder/a2f04c2fd85151fb5d44ebf7c29317a3)

https://gist.github.com/SteveBronder/4229073f6b12c564d2c0a6bd17cf80f2

stan/math/prim/meta/holder.hpp
SteveBronder: Where I'm getting confused in here

```cpp
template <typename T, require_t<std::is_rvalue_reference<T&&>>* = nullptr>
auto holder_handle_element(T&& a, T*& res) {
  res = new T(std::move(a));
  return std::make_tuple(res);
}
```

Where is res deleted at? Is it handled somewhere in Eigen?

t4c1: First this is only used in holder (and its opencl version). So any pointer allocated here ends passed into holder. Holder constructs `std::unique_ptr` around it (see line 65) and `std::unique_ptr` deletes it when its destructor is called.

t4c1:
`Eigen::Matrix<var>::Random` played wierdly with holder. I simplified how holder works with flags, which seemed to fix the problem. Anyway you shouldn't use `Eigen::Matrix<var>::Random` - it makes varis for operations within RNG. Use instead `Eigen::Matrix<double>::Random` and than convert to `var`s. Also including time for RNG in benchmark of a fast function, such as `value_of`, is a bad idea - large part of time will be spent in RNG.

SteveBronder:
> Eigen::Matrix<var>::Random played wierdly with holder. I simplified how holder works with flags, which seemed to fix the problem. Anyway you shouldn't use Eigen::Matrix<var>::Random. Use instead Eigen::Matrix<double>::Random and than convert to vars

ftr we want to still be able to use Eigen::<var>::Random post this PR.

>  Also including time for RNG in benchmark of a fast function, such as value_of, is a bad idea - large part of time will be spent in RNG.

Yes good call I moved these but now I'm getting an error from `XprHelper`

```
matrix_value_of_bench/128/manual_time           511206 ns      2157330 ns         1379
matrix_value_of_bench/256/manual_time          2048423 ns      8486119 ns          347
matrix_value_of_bench/512/manual_time          8220623 ns     33569091 ns           88
matrix_value_of_bench/1024/manual_time        32625450 ns    134413560 ns           22
matrix_value_of_bench/2048/manual_time       144199522 ns    572894706 ns            5
matrix_value_of_bench/4096/manual_time       611693881 ns   2400861336 ns            1
matrix_value_of_bench/8192/manual_time      2674345112 ns   9673294759 ns            1
holder_bench: math/lib/eigen_3.3.7/Eigen/src/Core/util/XprHelper.h:110: Eigen::internal::variable_if_dynamic<T, Value>::variable_if_dynamic(T) [with T = long int; int Value = 1]: Assertion `v == T(Value)' failed.
Aborted (core dumped)
```

SteveBronder:
https://gist.github.com/SteveBronder/4229073f6b12c564d2c0a6bd17cf80f2

t4c1:
> ftr we want to still be able to use Eigen::<var>::Random post this PR.

As I explained you really should not use that neither before nor post this PR, due to it being inefficient. If you are attached to the function name, I think we could speciallize `::Random` for `var` to make it efficient.

> but now I'm getting an error from XprHelper

That has nothing to do with holder. You cant call two argument version `::Random` on a vector (unless one of those is set to 1).

Also you are never calling `value_of_eval` (always `value_of`) in the new benchmark.

stan/math/prim/meta/holder.hpp
SteveBronder: Aight cool!

SteveBronder:
Aight word! Fixing up my benchmark the results look good!

[bench + local results](https://gist.github.com/SteveBronder/4229073f6b12c564d2c0a6bd17cf80f2)

Let me re-read the Eigen adding expression stuff and give this another look over today. The only non-code related thing here is that if we are going to be using this in a lot of places we should have it doc'd pretty well so that people have a reference to read on why we do this. Though I'm not sure where to put it. Maybe in a new doc module or in the contributors guide?

Also how do you feel about us filing an issue in the eigen gitlab repo to tell them about this thing? If it works rather generally then this may be a thing they want to give to the rest of their userbase (plus would be nice to have Eigen people review this)

t4c1:
> Though I'm not sure where to put it. Maybe in a new doc module or in the contributors guide?

I'm always confused aboout where to put extended doc. So I make a new group in doxygen, or ... ?

> Also how do you feel about us filing an issue in the eigen gitlab repo to tell them about this thing?

Eigen still lives in C++03 world, where rvalue references do not exist yet. Also I'm pretty sure one of Eigen's main authors has seen my question and answer on SO about this and did not express any interest in it.

stan/math/prim/meta/holder.hpp
```diff
@@ -0,0 +1,319 @@
...
+ * scope. So a function returning an expression referencing local matrices or
+ * matrices that were rvalue reference arguments to the function will not work.
+ *
+ * A workarount to this issue is allocating and constructing or moving such
+ * objects to heap. `Holder` object is a no-op operation that can also take
+ * pointers to such objects and release them when it goes out of scope. It can
+ * be created either by directly supplying pointers to such objects to `holder`
+ * function or by forwarding function arguments and moving local variables to
+ * `make_holder`, which will move any rvalues to heap first.
+ */
```
SteveBronder, line 38: I took this and included an example so we could describe what causes the failure in the example and how you would use `make_holder` to fix the code. What causes the need for holder seems like a weird issue most people would not have seen before so I want to make sure new developers can see this explainer and get what's happening. How do you feel about the below? Not sure if I got the explanation right but feel free to modify whatever

https://gist.github.com/SteveBronder/84c710c3f355eb53aa946983e48289a9

stan/math/prim/meta/holder.hpp
```diff
@@ -26,7 +26,29 @@
...
+ * Eigen::MatrixXd test_mat(2,2);
+ * test_mat << 5, 5, 5, 5;
+ * VectorXd X  = f(test_mat.diagonal());
+ * ```
+ * This function will return back a `CwiseBinaryOp` Eigen expression, which is
+ * then evaluated out of the function scope when assigned to `X`. The expression
+ * references `x_ref`, which was created withing function and destroyed, when
+ * the function returned. The returned expression is evaluated later than the
+ * function returned, so its evaluation references a matrix that was already
+ * deleted. In other words returned expression contains a dangling reference.
```
SteveBronder, line 49:
```suggestion
 * references `x_ref`, which was created within the function and destroyed via RAII when
 * the function returned. The returned expression is evaluated after the
 * function returned the expression, so its evaluation references a matrix that was already
 * deleted. In other words the returned expression contains a dangling reference.
```

Small grammatical changes

t4c1: Fixed grammar. I don't think other two suggestions add anything.

PR #1917, opened 2020-06-03
Extend binary vectorisation to complex inputs

andrjohns:
## Summary

This pull extends ```apply_scalar_binary``` to work with complex variables as inputs, and some minor fixes to the ```expect_ad``` testing framework to test these inputs.

```pow()``` is used as an example implementation.

## Tests

Additional ```prim``` tests for ```pow``` have been added to test that  primitive inputs are consistent across vectorised and scalar inputs. Additional ```mix``` tests have been added for both ```std::complex``` and autodiff inputs.

## Side Effects

Complex types are added as a ```stan_scalar``` type to keep the templating simple. This isn't required, and I work around it if this isn't preferred.

## Release notes

Vectorized binary functions extended to take complex variables as inputs

## Checklist

- [x] Math issue #1890 

### CODE REVIEW

**Comment by andrjohns** [comment]
Would someone mind giving this a quick review when they've got a minute?

File: stan/math/prim/meta/apply_scalar_binary.hpp [hunk 1]
```diff
@@ -51,6 +51,12 @@ inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
 template <typename T1, typename T2, typename F,
           require_all_eigen_t<T1, T2>* = nullptr>
 inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+  if ((x.rows() != y.rows()) || (x.cols() != y.cols())) {
+    std::ostringstream msg;
+    msg << "Inputs to vectorized binary function must match in"
+        << " size if one is not a scalar";
+    throw std::domain_error(msg.str());
+  }
```
t4c1, line 59: Don't we have some check function to turn this into a single line?

test/unit/math/serializer.hpp
```diff
@@ -92,10 +93,30 @@ struct deserializer {
    * @param x pattern argument to determine result shape and size
    * @return deserialized value with shape and size matching argument
    */
-  template <typename U>
-  typename stan::math::promote_scalar_type<T, std::vector<U>>::type read(
-      const std::vector<U>& x) {
-    typename stan::math::promote_scalar_type<T, std::vector<U>>::type y;
+  template <typename U, require_std_vector_t<U>* = nullptr,
+            require_not_std_vector_st<is_complex, U>* = nullptr>
```
t4c1, line 97:
```suggestion
            require_not_st_complex<U>* = nullptr>
```

test/unit/math/serializer.hpp
```diff
@@ -60,7 +60,8 @@ struct deserializer {
    * @param x pattern argument to determine result shape and size
    * @return deserialized value with shape and size matching argument
    */
-  template <typename U>
+  template <typename U, require_stan_scalar_t<U>* = nullptr,
+            require_not_complex_t<U>* = nullptr>
   T read(const U& x) {
     return vals_[position_++];
   }
```
t4c1, line 67: I don't see any overload for complex scalars. Is it intentionally missing?

test/unit/math/serializer.hpp
```diff
@@ -92,10 +93,30 @@ struct deserializer {
...
-      const std::vector<U>& x) {
-    typename stan::math::promote_scalar_type<T, std::vector<U>>::type y;
+  template <typename U, require_std_vector_t<U>* = nullptr,
+            require_not_std_vector_st<is_complex, U>* = nullptr>
+  typename stan::math::promote_scalar_type<T, U>::type read(const U& x) {
+    typename stan::math::promote_scalar_type<T, U>::type y;
+    y.reserve(x.size());
+    for (size_t i = 0; i < x.size(); ++i)
+      y.push_back(read(x[i]));
+    return y;
```
t4c1, line 103: Why do we even need separate overloads for complex/non-complex? Code seems exactly the same.

stan/math/prim/meta/apply_scalar_binary.hpp
andrjohns: We do, but all the check functions include ```prim/meta.hpp```, which includes this header, and spirals into circular dependencies

test/unit/math/serializer.hpp
andrjohns: That's because there's already a complex scalar overload on L80, but because I added complex types to ```is_stan_scalar```, this templating became ambiguous

test/unit/math/serializer.hpp
andrjohns: It's mostly so that the return type gets instantiated correctly. The current state of the AD testing framework has all of the templating referring to the type within the complex variable (```std::complex<T>```), so these templates need to do the same (even if the function calls themselves are near-identical). Otherwise I'd need to update the std::complex templating throughout the rest of the framework (so that ```T``` refers to ```std::complex<double>``` rather than just ```double```) to collapse these two definitions to one.

stan/math/prim/meta/apply_scalar_binary.hpp
t4c1: Hmm, do we need this to be in meta? I think it would belong into functor.

test/unit/math/serializer.hpp
t4c1: I think second option would lead to much clearer code. Unless that is too much work for this PR?

stan/math/prim/meta/apply_scalar_binary.hpp
andrjohns: I had it in meta to be consistent with ```apply_scalar_unary```. So if I move this one I'd have to move the other two, and I'm a bit hesitant to be moving existing headers after all of the issues that moving the ```Eigen.hpp``` headers has caused with RStan

test/unit/math/serializer.hpp
andrjohns: Yeah it would be a bit much just for the result to be collapsing these two definitions, the new definitions are doc-ed as being for complex inputs, so I figure that's explanatory enough

stan/math/prim/meta/apply_scalar_binary.hpp
t4c1: I think apply_scalar_unary should also eventually be moved (no need to do it in this PR). I think reducing code duplication should be a priority compared to being consistent with what turns out to be a poor design choice.

I am not saying you must change this, but I'd prefere if you did.

andrjohns: Works for me, I can move those files in this pull

andrjohns: I've moved the ```apply_*``` headers to the functor folder and updated the other functions that use them. This is ready for another review

andrjohns:
I can test this locally and work out a fix

PR #1916, opened 2020-06-02
Generalize bernoulli distribution functions

t4c1:
## Summary
Generalizes functions related to Bernoulli distribution to accept Eigen expressions. Some unnecesarry checks (check_finite together with check_bounded) were also removed from these functions. 

## Tests

No new tests. This is just a refactor.

## Side Effects

None.

## Release notes

Generalized functions related to Bernoulli distribution to accept Eigen expressions.

## Checklist

- [x] Math issue #1470

- [x] Copyright holder: Tadej Ciglarič

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #1920, opened 2020-06-04
Revert "Generalize bernoulli distribution functions"

SteveBronder:
## Summary

Reverts #1916


SteveBronder:
@t4c1 if you can get a patch in tmrw lmk else we can do the revert and revert-revert

rok-cesnovar:
Its today CET time :) If a fix is simple and we can get it in by the end of Friday CET we will do that, else I will approve/merge this.

wds15:
As I understand our performance tests actually catch this stuff... which is really cool.

PR #1928, opened 2020-06-08
[WIP] Reworked the multiplication operator to use adj_jac_apply

bbbales2:
Alright here's a version of the multiplication that I like better.

After writing this I'm not convinced we'd actually do multiplication with it. There might be some overheads especially with scalar * scalar multiplication that means we should stick with the previous varis, but it works for illustration and will be appropriate for a lot of the functions we do have to tackle (https://gist.github.com/bbbales2/a1b2bcc1f616820ce5db59b9be46c49c)

In this implementation there are `adj_jac_apply` functions to handle the implementations of the multiplication operator between different combinations of scalar and matrix types.

Type traits are used to make sure `adj_jac_apply` gets called for the right combination of scalar/matrix arguments. There are four combinations we gotta worry about:

```
operator*(scalar, scalar)
operator*(matrix, scalar)
operator*(scalar, matrix)
operator*(matrix, matrix)
```

The outside-facing function ended up looking like:
```
	  require_scalar_t<T1>...,
	  require_scalar_t<T2>...,
	  require_any_var_value_t<T1, T2>...>
inline auto operator*(const T1& a, const T2& b) {
  return adj_jac_apply<OpMultiplyScalarScalar>(a, b);
}
```

The type traits I needed were:
* ```require_scalar_t``` which accepts only mathematical scalar objects,
* ```require_matrix_t``` which only accepts objects that are matrices (this includes vectors/row_vectors) and then
* ```require_any_var_value_t``` to make sure at least one thing was a var (and avoid conflicting with prim implementations)

Implementations of ```require_scalar_t``` and ```require_matrix_t``` are [here](https://github.com/stan-dev/math/pull/1928/files#diff-4d75dd68c0cf62489e92b7584c06fde3R159)

So for the most part now, ```adj_jac_apply``` is doing all the juggling with autodiff types and whatnot, and so for implementing functions I don't really need to deal with lots of type_traits.

I'm not sure how much this would balloon if it supported things with complex number specializations or what those would look like. It's probably worth thinking about though.

The implementation starts around like [159](https://github.com/stan-dev/math/blob/4ab15812179e40fe6c26c25e5552d26e649430b6/stan/math/rev/core/operator_multiplication.hpp)

## Checklist

### DIFF
--- a/stan/math/rev/functor/adj_jac_apply.hpp
+++ b/stan/math/rev/functor/adj_jac_apply.hpp
@@ -45,37 +45,6 @@ public:
 namespace internal {
 
 /**
- * Store the adjoint in y_vi[0] in y_adj
- *
- * @tparam size dimensionality of M
- * @param[in] y_vi pointer to pointer to vari
- * @param[in] M shape of y_adj
- * @param[out] y_adj reference to variable where adjoint is to be stored
- */
-template <typename T>
-inline auto build_y_adj(vari_base** y_vi, int M) {
-  return dynamic_cast<T>(y_vi[0])->adj_;
-}
-
-/**
- * Store the adjoints from y_vi in y_adj
- *
- * @tparam size dimensionality of M
- * @param[in] y_vi pointer to pointers to varis
- * @param[in] M shape of y_adj
- * @param[out] y_adj reference to std::vector where adjoints are to be stored
- */
-template <>
-inline auto build_y_adj<vari**>(vari_base** y_vi, int M) {
-  std::vector<double> y_adj;
-  y_adj.reserve(M);
-  for (size_t m = 0; m < M; ++m) {
-    y_adj.emplace_back(dynamic_cast<vari*>(y_vi[m])->adj_);
-  }
-  return y_adj;
-}
-
-/**
  * Compute the dimensionality of the given template argument. The
  * definition of dimensionality is deferred to specializations. By
  * default don't have a value (fail to compile)
@@ -150,7 +119,7 @@ struct adj_jac_vari : public vari {
   static constexpr std::array<bool, sizeof...(Targs)> is_var_{
     {is_var_value<scalar_type_t<Targs>>::value...}};
   using FReturnType =
-    std::decay_t<std::result_of_t<F(decltype(is_var_), decltype(value_of(std::declval<Targs>()))...)>>;
+    std::decay_t<plain_type_t<std::result_of_t<F(decltype(is_var_), decltype(value_of(std::declval<Targs>()))...)>>>;
 
   F f_;  // Function to be invoked
 
@@ -162,30 +131,28 @@ struct adj_jac_vari : public vari {
   using y_vari_type = std::conditional_t<is_std_vector<FReturnType>::value,
 					 vari**,
 					 vari_value<FReturnType>*>;
-  vari_base** y_vi_;  // vari pointer for output.
-
-  void fill_adj_jac() {}
+  y_vari_type y_vi_;  // vari pointer for output.
+  
+  inline void fill_adj_jac() {}
 
   template <typename Arith, typename... Pargs,
             require_st_arithmetic<Arith>* = nullptr>
-  void fill_adj_jac(Arith&& x, Pargs&&... args) {
+  inline void fill_adj_jac(const Arith& x, const Pargs&... args) {
     static constexpr int t = sizeof...(Targs) - sizeof...(Pargs) - 1;
     std::get<t>(x_vis_) = nullptr;
     fill_adj_jac(args...);
   }
 
   template <typename T, typename... Pargs>
-  void fill_adj_jac(const var_value<T>& x, Pargs&&... args) {
+  inline void fill_adj_jac(const var_value<T>& x, const Pargs&... args) {
     static constexpr int t = sizeof...(Targs) - sizeof...(Pargs) - 1;
-    std::get<t>(x_vis_)
-      = ChainableStack::instance_->memalloc_.alloc_array<vari_value<T>>(1);
     std::get<t>(x_vis_) = x.vi_;
     fill_adj_jac(args...);
   }
 
   // std::vector<var_value>
   template <typename... Pargs>
-  void fill_adj_jac(const std::vector<var>& x, Pargs&&... args) {
+  inline void fill_adj_jac(const std::vector<var>& x, const Pargs&... args) {
     static constexpr int t = sizeof...(Targs) - sizeof...(Pargs) - 1;
 
     std::get<t>(x_vis_)
@@ -198,10 +165,9 @@ struct adj_jac_vari : public vari {
     fill_adj_jac(args...);
   }
 
-  // Eigen
   template <int R, int C, typename... Pargs>
-  void fill_adj_jac(const Eigen::Matrix<var, R, C>& x,
-                    Pargs&&... args) {
+  inline void fill_adj_jac(const Eigen::Matrix<var, R, C>& x,
+			   const Pargs&... args) {
     static constexpr int t = sizeof...(Targs) - sizeof...(Pargs) - 1;
 
     std::get<t>(x_vis_)
@@ -217,17 +183,20 @@ struct adj_jac_vari : public vari {
     fill_adj_jac(args...);
   }
 
-  /**
-   * Return a var with a new vari holding the given value
-   *
-   * @param val_y output of F::operator()
-   * @return var
-   */
-  template <typename T>
-  inline var_value<T> build_return_varis_and_vars(const T& val_y) {
-    y_vi_ = ChainableStack::instance_->memalloc_.alloc_array<vari_base*>(1);
-    y_vi_[0] = new vari_value<T>(val_y);
-    return dynamic_cast<vari_value<T>*>(y_vi_[0]);
+  inline var build_return_varis_and_vars(double val_y) {
+    const_cast<double&>(val_) = val_y;
+    y_vi_ = static_cast<vari*>(this);//new vari(val_y, false);
+    return this;
+  }
+
+  template <typename Derived,
+	    typename T = Eigen::Matrix<double,
+				       Eigen::MatrixBase<Derived>::RowsAtCompileTime,
+				       Eigen::MatrixBase<Derived>::ColsAtCompileTime>>
+  inline var_value<T>
+  build_return_varis_and_vars(const Eigen::MatrixBase<Derived>& val_y) {
+    y_vi_ = new vari_value<T>(val_y, false);
+    return y_vi_;
   }
 
   /**
@@ -246,7 +215,7 @@ struct adj_jac_vari : public vari {
       (val_y.size());
     for (size_t m = 0; m < val_y.size(); ++m) {
       y_vi_[m] = new vari_value<T>(val_y[m], false);
-      var_y.emplace_back(dynamic_cast<vari_value<T>*>(y_vi_[m]));
+      var_y.emplace_back(static_cast<vari_value<T>*>(y_vi_[m]));
     }
     return var_y;
   }
@@ -269,19 +238,17 @@ struct adj_jac_vari : public vari {
    * @return Output of f_ as vars
    */
   inline auto operator()(const Targs&... args) {
-    auto func_ret = f_(is_var_, value_of(args)...);
-    auto return_obj = build_return_varis_and_vars(func_ret);
-    return return_obj;
+    return build_return_varis_and_vars(f_(is_var_, value_of(args)...));
   }
 
   template <typename T>
   inline void accumulate_adjoints_in_varis_impl(std::nullptr_t, T&&) {}
 
-  template <int R1, int C1, int R2, int C2>
+  template <int R1, int C1, typename Derived>
   inline void accumulate_adjoints_in_varis_impl
-  (vari_value<Eigen::Matrix<double, R1, C1>> *vi, const Eigen::Matrix<double, R2, C2>& y_adj_jac) {
-    if(y_adj_jac.size() > 0)
-      vi->adj_ += y_adj_jac;
+  (vari_value<Eigen::Matrix<double, R1, C1>> *vi, const Eigen::MatrixBase<Derived>& y_adj_jac) {
+    for (int n = 0; n < y_adj_jac.size(); ++n)
+      vi->adj_.coeffRef(n) += y_adj_jac.coeff(n);
   }
 
   template <int R, int C>
@@ -299,7 +266,7 @@ struct adj_jac_vari : public vari {
     }
   }
 
-  inline void accumulate_adjoints_in_varis_impl(vari *vi, const double& y_adj_jac) {
+  inline void accumulate_adjoints_in_varis_impl(vari *vi, double y_adj_jac) {
     vi->adj_ += y_adj_jac;
   }
 
@@ -307,12 +274,46 @@ struct adj_jac_vari : public vari {
 
   template <typename T,
 	    typename... Pargs>
-  inline void accumulate_adjoints_in_varis(const T& y_adj_jac, const Pargs&... args) {
+  inline void accumulate_adjoints_in_varis(const T& y_adj_jac,
+					   const Pargs&... args) {
     static constexpr size_t t = sizeof...(Targs) - sizeof...(Pargs) - 1;
-    if(std::get<t>(x_vis_)) {
+    if(is_var_[t])
       accumulate_adjoints_in_varis_impl(std::get<t>(x_vis_), y_adj_jac);
-    }
     accumulate_adjoints_in_varis(args...);
+  }
+
+  /**
+   * Store the adjoint in y_vi[0] in y_adj
+   *
+   * @tparam size dimensionality of M
+   * @param[in] y_vi pointer to pointer to vari
+   * @param[in] M shape of y_adj
+   * @param[out] y_adj reference to variable where adjoint is to be stored
+   */
+  template <typename T>
+  inline auto build_y_adj(vari_value<T>* y_vi, int M) {
+    return y_vi->adj_;
+  }
+
+  inline auto build_y_adj(vari* y_vi, int M) {
+    return adj_;
+  }
+
+  /**
+   * Store the adjoints from y_vi in y_adj
+   *
+   * @tparam size dimensionality of M
+   * @param[in] y_vi pointer to pointers to varis
+   * @param[in] M shape of y_adj
+   * @param[out] y_adj reference to std::vector where adjoints are to be stored
+   */
+  inline auto build_y_adj(vari** y_vi, int M) {
+    std::vector<double> y_adj;
+    y_adj.reserve(M);
+    for (size_t m = 0; m < M; ++m) {
+      y_adj.emplace_back(static_cast<vari*>(y_vi[m])->adj_);
+    }
+    return y_adj;
   }
 
   /**
@@ -327,13 +328,11 @@ struct adj_jac_vari : public vari {
    * This operation may be called multiple times during the life of the vari
    */
   inline void chain() {
-    auto y_adj_jacs = f_.multiply_adjoint_jacobian(is_var_, internal::build_y_adj<y_vari_type>(y_vi_, M_));
-
     apply(
         [&, this](auto&&... args) {
           this->accumulate_adjoints_in_varis(args...);
         },
-        y_adj_jacs);
+        f_.multiply_adjoint_jacobian(is_var_, build_y_adj(y_vi_, M_)));
   }
 };
 
--- a/stan/math/rev/functor/adj_jac_apply.hpp
+++ b/stan/math/rev/functor/adj_jac_apply.hpp
@@ -16,6 +16,32 @@
 
 namespace stan {
 namespace math {
+
+class AdjJacOp {
+public:
+  template <typename Derived>
+  double* allocate_and_save(const Eigen::EigenBase<Derived>& x) const {
+    double* x_mem_
+      = stan::math::ChainableStack::instance_->memalloc_.alloc_array<double>(x.size());
+
+    Eigen::Map<Eigen::MatrixXd>(x_mem_, x.rows(), x.cols()) = x;
+
+    return x_mem_;
+  }
+
+  Eigen::Map<Eigen::MatrixXd> map_matrix(double* mem, size_t rows, size_t cols) {
+    return Eigen::Map<Eigen::MatrixXd>(mem, rows, cols);
+  }
+
+  Eigen::Map<Eigen::VectorXd> map_vector(double* mem, size_t rows) {
+    return Eigen::Map<Eigen::VectorXd>(mem, rows);
+  }
+
+  Eigen::Map<Eigen::RowVectorXd> map_row_vector(double* mem, size_t cols) {
+    return Eigen::Map<Eigen::RowVectorXd>(mem, cols);
+  }
+};
+  
 namespace internal {
 
 /**
@@ -122,9 +148,9 @@ template <typename F, typename... Targs>
 struct adj_jac_vari : public vari {
   // holds whether each input type held a vari_value
   static constexpr std::array<bool, sizeof...(Targs)> is_var_{
-      {is_var_value<scalar_type_t<Targs>>::value...}};
+    {is_var_value<scalar_type_t<Targs>>::value...}};
   using FReturnType =
-    std::decay_t<std::result_of_t<F(decltype(is_var_), decltype(value_of(Targs()))...)>>;
+    std::decay_t<std::result_of_t<F(decltype(is_var_), decltype(value_of(std::declval<Targs>()))...)>>;
 
   F f_;  // Function to be invoked
 
@@ -389,14 +415,18 @@ constexpr std::array<bool, sizeof...(Targs)> adj_jac_vari<F, Targs...>::is_var_;
  * @param args input to the functor
  * @return the result of the specified operation wrapped up in vars
  */
-template <typename T>
+template <typename T,
+	  require_not_eigen_vt<is_var, T>* = nullptr>
 const T& convert_to_whole_matrix(const T& arg) {
   return arg;
 }
 
-template <int R, int C>
-var_value<Eigen::Matrix<double, R, C>> convert_to_whole_matrix(
-    const Eigen::Matrix<var, R, C>& arg) {
+template <typename T,
+	  require_eigen_vt<is_var, T>* = nullptr>
+var_value<Eigen::Matrix<double,
+			T::RowsAtCompileTime,
+			T::ColsAtCompileTime>>
+convert_to_whole_matrix(const T& arg) {
   return arg;
 }
 
@@ -409,7 +439,7 @@ inline auto adj_jac_apply_impl(const Targs&... args) {
 
 template <typename F, typename... Targs>
 inline auto adj_jac_apply(const Targs&... args) {
-  return adj_jac_apply_impl<F>(convert_to_whole_matrix(args)...);
+  return adj_jac_apply_impl<F>(args...);
 }
 
 }  // namespace math
--- a/stan/math/rev/functor/adj_jac_apply.hpp
+++ b/stan/math/rev/functor/adj_jac_apply.hpp
@@ -19,53 +19,35 @@ namespace math {
 namespace internal {
 
 /**
- * Place the data for the y_adj_ into the appropriate container.
- */
-template <typename T, typename = void>
-struct build_y_adj;
-
-/**
- * Specialization returns the real adjoint value for arithmetic and eigen
- * types. Since `vari_value<Eigen>`` holds the dimensions we don't need M.
+ * Store the adjoint in y_vi[0] in y_adj
+ *
+ * @tparam size dimensionality of M
+ * @param[in] y_vi pointer to pointer to vari
+ * @param[in] M shape of y_adj
+ * @param[out] y_adj reference to variable where adjoint is to be stored
  */
 template <typename T>
-struct build_y_adj<T,
-                   require_t<disjunction<std::is_arithmetic<T>, is_eigen<T>>>> {
-  /**
-   * Retreive the adjoints for y_adj_
-   *
-   * @tparam size dimensionality of M
-   * @param[in] y_vi pointer to pointer to vari
-   * @param[in] M shape of y_adj (not used in this specialization)
-   */
-  template <typename VariType, size_t size>
-  static inline auto& apply(VariType&& y_vi, const std::array<int, size>& M) {
-    return y_vi->adj_;
-  }
-};
+inline auto build_y_adj(vari_base** y_vi, int M) {
+  return dynamic_cast<T>(y_vi[0])->adj_;
+}
 
 /**
- * Specialization returns the real adjoint value for std vector return types.
+ * Store the adjoints from y_vi in y_adj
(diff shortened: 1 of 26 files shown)


bbbales2:
@SteveBronder see whatcha think of this organization. Ended up able to use at the high level type traits for mathematical things (scalars, matrices) and keep the autodiff type traits hidden in adj_jac_apply (which can sorta be as obtuse as it needs to be). Dunno if this generalizes to the rest of the functions we gotta do but that's the hope.

SteveBronder:
I really like it! For a few major reasons

1. It looks good and follows a standard pattern
2. We can ditch all the op_vari code so that saves a ton of effort
3. If the type traits as they are work fine then we can also not worry about that.

Can you do a benchmark to see what scalar * scalar + grad looks like with this vs develop? If it works fine for scalar * scalar then I think we can run with what you have here. Else we could also keep the vari_op for scalars and have an adj_jac method for anything involving matrices.

bbbales2:
> Can you do a benchmark to see what scalar * scalar + grad looks like with this vs develop?

I am scared we'll lose in a pretty significant way, but this is probably a good way to tighten up the performance on `adj_jac_apply` as much as possible.

SteveBronder:
If that's the case then why don't we keep the current method, then in the `operator*` use SFINAE to have one that only accepts scalars and one that accepts matrices? imo that might not be a bad tradeoff. The version I wrote was very dense where more code here is not bad if it makes it easier to understand

bbbales2:
This is more about how everything else gets written than how `operator*` gets written. I'm fine with going back to what was there for `operator*`, but I will do the perf tests anyway.

stan/math/rev/functor/adj_jac_apply.hpp
```diff
@@ -19,53 +19,35 @@ namespace math {
...
-    for (size_t m = 0; m < M[0]; ++m) {
-      y_adj.emplace_back(y_vi[m]->adj_);
-    }
-    return y_adj;
+template <>
+inline auto build_y_adj<vari**>(vari_base** y_vi, int M) {
+  std::vector<double> y_adj;
+  y_adj.reserve(M);
+  for (size_t m = 0; m < M; ++m) {
+    y_adj.emplace_back(dynamic_cast<vari*>(y_vi[m])->adj_);
```
SteveBronder, line 47: Why did you change this? idt this will work in general?

bbbales2: The `adj_jac_apply` tests weren't working (some sort of out of bounds indexing) so I just desperately rewrote the varis logic to use fixed length tuples that hold `std::nullptr_t` s where non-autodiff types were.

This was the end of a long series of changes lol.

There are two hidden assumptions in this code regarding matrix types, but:

1. `adj_jac_apply` only deals on input with `var_value<MatrixXd>` matrices (they're converted into this format before they go in). This could be relaxed.

2. `adj_jac_apply` only deals with `var_value<MatrixXd>` matrices on the output. I don't think we should relax this.

3. `adj_jac_apply` only deals with arrays of scalars.

By assumption 2 & 3, if the output is of type `vari**` it is a `std::vector<var>`.

bbbales2: > There are two hidden assumptions

Edit: At least two lol

bbbales2: We should also relax assumption 3, eventually (I think 1 comes first though).

stan/math/rev/functor/adj_jac_apply.hpp
```diff
@@ -19,53 +19,35 @@ namespace math {
...
-   * @tparam size dimensionality of M
-   * @param[in] y_vi pointer to pointer to vari
-   * @param[in] M shape of y_adj (not used in this specialization)
-   */
-  template <typename VariType, size_t size>
-  static inline auto& apply(VariType&& y_vi, const std::array<int, size>& M) {
-    return y_vi->adj_;
-  }
-};
+inline auto build_y_adj(vari_base** y_vi, int M) {
```
SteveBronder, line 30: I'd recommend we don't use `vari_base` outside of vari.hpp and `chainablestack`. I might make it in namespace internal to emphasize that. I think we can actually know all of the `var_value<T>` types that come into `adj_jac`

bbbales2: Cool, yeah, this part I put together in a bit of a panic at the end (just wanted something to work). I'm happy getting rid of the `dynamic_cast` s too.

SteveBronder: Word!

bbbales2:
I worked on some performance stuff with this.

The short version is I'm pretty happy putting `adj_jac_apply` anywhere now and would be happy replacing already-written autodiff varis with it.

I wouldn't replace the scalar ops, though it's possible to make `adj_jac_apply` implementations competitive there.

Also I think `adj_jac_apply` is always going to return `var_value<Eigen::Matrix<double, R, C>>` types, which means if we are assigning things to `Eigen::Matrix<var, R, C>`, there's an extra copy that isn't included in the calculations here (also casts from `Eigen::Matrix<var, R, C>` aren't accounted for).

I was trying to make this loop fast:

```
var_value<Eigen::MatrixXd> r = Eigen::MatrixXd::Ones(state.range(0), 1);
for(size_t i = 0; i < 1000; ++i) {
  r = r * 0.99;
}
var_value<double> lp = stan::math::sum(r);
lp.grad()
```

That's all scalar * vector multiplies, so this is something where the `adj_jac_apply` overhead will be significant for small vectors.

Anyway I got it to a point where the break-even with the old-style `Eigen::Matrix<var, R, C>` autodiff was a vector of length 4.

I timed the performance against develop and `feature/eigen-vari-ops` commit `bcacdb` and plotted the speedup of both vs. develop:

perf

The comparison I'm missing is if I tried hard to make `bcacdb` fast here. I think I should do that just to verify it is the speed limit for large vectors.

For the small vector lengths you gotta be careful to not malloc things.

I also did a benchmark computing the first P terms of a geometric series:
```
var r = 1.0;
var sum = 0.0;
for(size_t i = 0; i < state.range(0); ++i) {
  sum += r;
  r = r * 0.99;
}
sum.grad();
```

perf_scalar

There are a bunch of little tricks (including a not-so-nice const-cast of a `vari->val_`), but it's nearly as efficient as the base thing. The message I'm taking away from this is that the compiler isn't having much trouble optimizing around all the weird template stuff we're doing in `adj_jac_apply`. I still think we write custom ops for our scalars.

bbbales2:
Just did the vector * scalar tests where I cast `Eigen::Matrix<var, R, C>` values to `var_value<Eigen::MatrixXd>` before calling multiply through `adj_jac_apply` and then casting them back afterwards (to simulate what would happen if we used the `adj_jac_apply` functions on code that wasn't using `var_values`).

It took 10x as long for the smallest vector and 1.5x as long for the largest vector. That's pretty bad. 

I'd like to say on input if `adj_jac_apply` doesn't get any `var_value` matrices, and the output type is a matrix, `adj_jac_apply` should probably return an `Eigen::Matrix<var, R, C>` to avoid these potential copies.

But if we're assigning the output to a `var_value` then this just created a bunch of unnecessary vars, so that isn't great.

It will be possible for `adj_jac_apply` to accept `Eigen::Matrix<var, R, C>` types, but we won't be able to get around this return ambiguity without Stan compiler stuff.

bbbales2:
Here's a gist with all the benchmarks: https://gist.github.com/bbbales2/dd93b94202b9bd2fcc79df7300d5185d

bob-carpenter:
> I'm pretty happy putting adj_jac_apply anywhere now and would be happy replacing already-written autodiff varis with it

That's really great news!  But then the next message seems to take it back:

> It took 10x as long for the smallest vector and 1.5x as long for the largest vector. 

Is the point that we need to only wrap things, not copy inside the adjoint-Jacobian method?  

What's `var_value<Eigen::MatrixXd>`?   If it were `Matrix<var, R, C>` would it just be a view onto the values of the nested `vari*`?

I don't see where the return ambiguity is.  Why is there a return?  Shouldn't adj_jac_apply just add the adjoint-Jacobian product to the operands' adjoints?

bbbales2:
`var_value<Eigen::Matrix<double, R, C>>` is the new `whole matrix`/`state matrix` autodiff type. `Eigen::Matrix<var, R, C>` are the old types.

> But then the next message seems to take it back:

The issue is when we write the `adj_jac_apply` operators, we write them in non-autodiff terms.

So if someone's function returns a matrix, should `adj_jac_apply` wrap it up as a `Eigen::Matrix<var, R, C>` or a `var_value<Eigen::Matrix<double, R, C>>`? It can do either.

If `func` is something that was defined with `adj_jac_apply`, then if `adj_jac_apply` always returns `Eigen::Matrix<var, R, C>`s, then this thing has an unnecessary copy from `var_value<Eigen::Matrix<double, R, C>>` to `Eigen::Matrix<var, R, C>`:

```
matrix[N, N] out = func(A);
```

This is because `adj_jac_apply` doesn't know about the return type.

Maybe there's a way to do this `adj_jac_apply` using expression types like Eigen and doing something special with the `assign` operator code gen'd in stanc3?

What if there was an expression type, `Expression<adj_jac_apply>` or something, and when the code:

`assign(Eigen::Matrix<var, R, C>&, Expression<adj_jac_apply>)` gets called, then the `Eigen::Matrix<var, R, C>` version of `adj_jac_apply` gets invoked and otherwise there's a `var_value<Eigen::Matrix<double, R, C>>` constructor from an `Expression<adj_jac_apply>` type (so C++ will just automatically do that every other time it gets an expression where it could use a matrix type).

We could actually disallow assigning/constructing `Eigen::Matrix<var, R, C>` from `var_value<Eigen::Matrix<double, R, C>>` except through the assign operator.

I don't know how well this would work.

Anyway I've seen enough to go back through and clean up this pull req and properly test it. It's in too fragile a state right now to do much more optimizing on.

SteveBronder:
Once I finish cleaning up the other PR I'll look at this, but I bet there is a way where, if any of the input matrices are `Matrix<var>` we can promote the return type to `Matrix<var>` instead.

I _think_ we can still keep the `x_vis_` as a `var<Eigen>` since that is copied already and wont need to be coped another time

bbbales2:
> I think we can still keep the x_vis_ as a var<Eigen> since

If the argument came in as an `Eigen<var>`, then I think we shouldn't do the copy. At some point the pointer-chasing happens, might as well be inside `adj_jac_apply`.

bbbales2:
> if any of the input matrices are `Matrix<var>` we can promote the return type to `Matrix<var>` instead.

Oh this is definitely true, but also something we wouldn't want to do. We should stay in `var_value` land as long as we can. I think the only time we convert out is if someone explicitly needs to work elementwise with something (so they assign to a type that can be indexed, or they index an intermediate calculation that could otherwise have been left as a `var_value<Matrix>`)

SteveBronder:
> Oh this is definitely true, but also something we wouldn't want to do. We should stay in var_value land as long as we can. I think the only time we convert out is if someone explicitly needs to work elementwise with something (so they assign to a type that can be indexed, or they index an intermediate calculation that could otherwise have been left as a var_value<Matrix>)

Yeah but isn't the issue that if someone does want that dynamic access they pay that 10x cost for the conversion? I think it's reasonable to think that if none of the input types are a dynamic matrix then it's safe to use `var_value<Eig>` instead

bbbales2:
@SteveBronder I'm getting errors like:

```
lib/eigen_3.3.7/Eigen/src/Core/DenseBase.h:289:14: note: candidate template ignored: could not match 'EigenBase' against 'var_value'
    Derived& operator+=(const EigenBase<OtherDerived> &other);
             ^
lib/eigen_3.3.7/Eigen/src/Core/MatrixBase.h:158:14: note: candidate template ignored: could not match 'MatrixBase' against 'var_value'
    Derived& operator+=(const MatrixBase<OtherDerived>& other);
             ^
lib/eigen_3.3.7/Eigen/src/Core/MatrixBase.h:475:46: note: candidate template ignored: could not match 'ArrayBase' against 'var_value'
    template<typename OtherDerived> Derived& operator+=(const ArrayBase<OtherDerived>& )
```

which I believe come from an `Eigen::Matrix<var, R, C>` trying to add a `var_value<Eigen::Matrix<double, R, C>>` to itself. What are my options for making this okay (because it seems like it should be possible)?

bbbales2:
@SteveBronder I converted a few more functions and did benchmarking. I'm not so sure about `operator*` anymore lol. I'll add a couple things at the end on that.

The functions I'm benchmark either take one or two arguments.

In either case what I am benchmarking is the evaluation of that function with one autodiff argument (the other, if there is one, set to double) and one backwards sweep (a grad).

Since the functions have vector output what I do is:

```
sum(f(x)).grad()
```

and

```
sum(f(x, y)).grad()
```

I am benchmarking the new functions written with `adj_jac_apply` and `var_value<Eigen>` types vs. the autodiff in develop.

One slight weirdness is because the sum is in there, I am always benchmarking the function *and* the sum, so there's that caveat. `adj_jac_apply` always returns a `var_value<Eigen>` type for matrix output, which will have a faster sum than sum `Matrix<var>` (at least it should lol).

When the input arguments are `var_value<Eigen>` types for the new code and `Matrix<var>` types for the old code, the speedups look like:

speedup_new

When the input arguments are `Matrix<var>` types to both implementations, the speedups look like:

speedup

The big speedups for tcrossprod and `multiply_lower_tri_self_transpose` (which is what `multiply_lower` is) are because the new custom autodiff is written differently than the old stuff was.

Back to `operator*`, I was have trouble working with that cause it's just hard. Also `row_vector * vector` in Stan is a `real`. In Eigen it is a 1x1 matrix (which means the `prim` `operator*` outputs a 1x1 matrix in the `row_vector * vector` case, which means `rev` `operator*` needs to be too, which means `operator*` is different from `multiply`). Anyway we can come back at the end and decide what we want to do for all that.

To get this moving I hacked together a questionable test framework. I've also just been throwing up whatever type traits I need. I'm gonna keep working through the functions and then come back at the end and clean this stuff up. Too hard to think through what would be useful beforehand.

bbbales2:
The method to picking these functions was, I just picked the first off this list: https://gist.github.com/bbbales2/a1b2bcc1f616820ce5db59b9be46c49c (`columns_dot_product`) and then re-implemented stuff till things compiled again, which meant I needed to convert a few other functions on the list.

SteveBronder:
Hey sorry I've had my head down while I'm trying to sort out the performance regression from adding the template to var.

I'll try to look at those eigen errors tonight/tmrw. I think the speed tests make sense (it's nice some of these go much faster even for smallish N!) Kind of curious that dot_prod and column dot prod would be the same speed

bob-carpenter:
This looks really great.  Is the memory usage similar?

bbbales2:
Oh another thing, for this to be convenient to code, we'd really want `var_value<Eigen>` to behave like a `const MatrixBase` thing.

I think there's too much there to overload to actually make it a matrix base, but so far I've hit situations where it would have been convenient to have `.col(i)`, `.row(j)`, `.transpose()`, and `.reshaped(N, M)`. The `col` and `row` accessors should be fine for reading.

Just leaving a note here for future us. I'll work around it for now (by providing overloads for the `transpose` function for `var_value<Eigen>`).

bbbales2:
> This looks really great. Is the memory usage similar?

For the ops implemented it should be. I'll try to remember to check on the next set of benchmarks.

When I was working on this I used some workspace memory to avoid malloc/free in the multiply_adjoint_jacobian reverse pass ([here](https://github.com/stan-dev/math/blob/f1f32af911b15e3ca4f991b85f03caba9d93caf5/stan/math/rev/core/operator_multiplication.hpp#L219)). This trick might be handy in some places, and it does up the memory usage.

I didn't end up using it in the ops here. Returning expressions was faster ([here](https://github.com/stan-dev/math/blob/f1f32af911b15e3ca4f991b85f03caba9d93caf5/stan/math/rev/fun/columns_dot_product.hpp#L55)) or it didn't seem to matter so I just allocated a temporary ([here](https://github.com/stan-dev/math/blob/f1f32af911b15e3ca4f991b85f03caba9d93caf5/stan/math/rev/fun/tcrossprod.hpp#L41)).

SteveBronder:
Ben do you mind if I make some changes to this? Right now this doesn't support Mat * Vec and it also loses type information when doing the multiplies (i.e. it return back a `Matrix<double, -1, -1>` for a `mat * vec` which should be a `Matrix<double, -1, 1>`).

Also I think the construction of the structs like `OpMultiplyMatrixMatrix` should separate out construction from `operator()`.  We should have some little wrapper thing that the inputs go into when dealing with doubles/mats etc that stores the info we needed. Something like

```cpp
class adj_op;

template <typename T>
class adj_op<T, is_eigen<T>> {
   const int rows;
   const int cols;
   static constexpr int RowsAtCompileTime{T::RowsAtCompileTime};
   static constexpr int ColsAtCompileTime{T::ColsAtCompileTime};
   T::Scalar* mem_
   Eigen::Map<T> map_
    // constructor stuff that only allocates mem when needs_adj is true
};
//...
struct OpMultiplyMatrixMatrix {
  adj_op<T1> A_;
  adj_op<T2> B_;
  explicit OpMultiplyMatrixMatrix(const std::array<bool, size>& needs_adj, 
    const T1& A, const T2& B) : A_(A, needs_adj[0]), B_(B, needs_adj[1]) {}
  auto operator()(const T1& A, const T2& B) {
    return A * B;
  }
  template <std::size_t size, typename Derived>
  auto multiply_adjoint_jacobian(const std::array<bool, size>& needs_adj,
                                 const Eigen::MatrixBase<Derived>& adj) {
    Eigen::MatrixXd adjA;
    Eigen::MatrixXd adjB;

    if(needs_adj[0]) {
      adjA = adj * B_.transpose();
    }
    if(needs_adj[1]) {
      adjB = A_.transpose() * adj;
    }
    return std::forward_as_tuple(adjA, adjB);
  }
};
```

SteveBronder:
ftr I'm trying to run this benchmark with this branch and `feature/eigen-vari-ops`

```cpp
#include <benchmark/benchmark.h>
#include <stan/math.hpp>
#include <utility>

// Just to kick off the stack allocation
static void mat_vec_mul(benchmark::State& state) {
  using stan::math::var_value;
  using vec_mat = Eigen::Matrix<double, -1, -1>;
  for (auto _ : state) {
    Eigen::MatrixXd x = Eigen::MatrixXd::Random(state.range(0), 10);
    var_value<vec_mat> b = vec_mat::Random(10, 1);
    benchmark::DoNotOptimize(x.data());
    benchmark::DoNotOptimize(b.val().data());
    benchmark::DoNotOptimize(b.adj().data());
    auto start = std::chrono::high_resolution_clock::now();
    var_value<vec_mat> result = x * b;
    result.grad();
    auto end = std::chrono::high_resolution_clock::now();
    benchmark::DoNotOptimize(result.val().data());
    benchmark::DoNotOptimize(result.adj().data());
    benchmark::ClobberMemory();
    auto elapsed_seconds =
      std::chrono::duration_cast<std::chrono::duration<double>>(end - start);
    state.SetIterationTime(elapsed_seconds.count());
    stan::math::recover_memory();
    benchmark::ClobberMemory();
  }
}

static void mat_vec_mul_reverse(benchmark::State& state) {
  using stan::math::var_value;
  using vec_mat = Eigen::Matrix<double, -1, -1>;
  Eigen::MatrixXd x = Eigen::MatrixXd::Random(state.range(0), 10);
  var_value<vec_mat> b = vec_mat::Random(10, 1);
  benchmark::DoNotOptimize(x.data());
  benchmark::DoNotOptimize(b.val().data());
  benchmark::DoNotOptimize(b.adj().data());
  var_value<vec_mat> result = x * b;
  for (auto _ : state) {
    auto start = std::chrono::high_resolution_clock::now();
    result.grad();
    auto end = std::chrono::high_resolution_clock::now();
    benchmark::DoNotOptimize(result.val().data());
    benchmark::DoNotOptimize(result.adj().data());
    benchmark::ClobberMemory();
    auto elapsed_seconds =
      std::chrono::duration_cast<std::chrono::duration<double>>(end - start);
    state.SetIterationTime(elapsed_seconds.count());
  }
}

// The start and ending sizes for the benchmark
int start_val = 12;
int end_val = 4096;
BENCHMARK(mat_vec_mul)->RangeMultiplier(2)->Range(start_val, end_val)->UseManualTime();
BENCHMARK(mat_vec_mul_reverse)->RangeMultiplier(2)->Range(start_val, end_val)->UseManualTime();
BENCHMARK_MAIN();
```

bbbales2:
> Ben do you mind if I make some changes to this? Right now this doesn't support Mat * Vec

Yeah sure, I agree with what you're saying.

> I think the construction of the structs like OpMultiplyMatrixMatrix should separate out construction from operator()

This would make respecting types of things cleaner. I think I avoided that just to have free type deduction, but we can specify all the types we need here. Go ahead, if you decide it makes sense.

I also wanted to make `needs_adj` template parameters. Like you'd make an op with something like:

```
template <bool needs_adj_T1, bool needs_adj_T2, typename T1, typename T2>
class adj_op {
   const int rows;
   const int cols;
```

so that we could add conditionals for `needs_adj_T1` and never have to worry about performance stuff. I guess there could also be template specializations that could be handy.

Right now I think I have `operator*` defined like `multiply` in Stan. That needs changed to be like Eigen I think. `multiply` in Stan is `row_vector * vector -> real` in Stan, and `operator*` in Eigen is `row_vector * vector -> matrix[1, 1]`.

As a side note, `operator*` was hard for me to work with as a start (it's so complicated). I had better luck thinking about the other [reverse mode ops](https://gist.github.com/bbbales2/a1b2bcc1f616820ce5db59b9be46c49c).

bbbales2:
> I guess there could also be template specializations that could be handy.

What I mean by this is when I was returning expression types from `multiply_adjoint_jacobian`, it was difficult to work in the runtime `needs_adj` logic because the return types would be different for different specifications of `needs_adj` (if you don't need the adjoint, return nothing, if you need the adjoint, return an expression, etc.).

This isn't great though cause it's getting back to the old-style tons-of-specializations form. I was able to get around this issue using zero sized operations [here](https://github.com/stan-dev/math/blob/f1f32af911b15e3ca4f991b85f03caba9d93caf5/stan/math/rev/fun/columns_dot_product.hpp#L52).

Anyway this might not make sense.

bbbales2:
Oh yeah for benching do a simpler op than `operator*`. Especially since that's not an op we'll be calling in Stan code, I don't wanna revolve all effort around it. `multiply` would probably be better.

Maybe do `columns_dot_product` as implemented here cause it's simple and difficult to do as well as the old code.

SteveBronder:
> Especially since that's not an op we'll be calling in Stan code, I don't wanna revolve all effort around it. multiply would probably be better.

For `multiply()` I was just going to do `A * B` on the inside for static matrices

> I also wanted to make `needs_adj` template parameters. Like you'd make an op with something like:

```
template <bool needs_adj_T1, bool needs_adj_T2, typename T1, typename T2>
class adj_op {
   const int rows;
   const int cols;
```

> so that we could add conditionals for `needs_adj_T1` and never have to worry about performance stuff. I guess there could also be template specializations that could be handy.

I have to check godbolt whether, since `needs_adj` is `static constexpr`, the compiler is smart enough to get rid of those if statements though dead code elimination. It would be nice to use the `adj_op` and `if`'s that contain compile time info to just have one `AdjMultiply` struct. If we had C++17 we could just use `if constexpr` but we have to compile at O3 here and cross our fingers a bit that the compiler is smart enough to figure these things out.

> Right now I think I have `operator*` defined like `multiply` in Stan. That needs changed to be like Eigen I think. `multiply` in Stan is `row_vector * vector -> real` in Stan, and `operator*` in Eigen is `row_vector * vector -> matrix[1, 1]`.

Yeah another place `if constexpr` would be handy. We can throw all of that in specialized functions in the struct though it's a bit annoying

SteveBronder:
Also I've been thinking about the return type of `vari<Eigen>` vs `Eigen<var>` ops. We could add one more rule

> For A dynamic matrix to be converted to a static matrix any operation on the matrix must also return a static matrix.

This makes the compiler side more difficult because we have to track the returns of each matrix op, but it makes the math side a lot easier in that now we just have the hard rule that if we see a dynamic matrix we can assume the result is also going to be a dynamic matrix

bbbales2:
> For A dynamic matrix to be converted to a static matrix any operation on the matrix must also return a static matrix.

> if we see a dynamic matrix we can assume the result is also going to be a dynamic matrix

I want to aggressively convert to `var_value<Eigen>` types, though I guess it's hard to evaluate this outside of the context of specific models. I think we could do some sort of expression thing in the future (the output of an `adj_jac_apply` remains an expression until consumed -- the constructor of an `Eigen<var>` would do things the old way and any other operator would evaluate it to a `var_value<Eigen>` type.

For hierarchical regression, we'll have things like:

```
y ~ normal(X * b + alpha1[group1], sigma);
```

It's entirely possible `b` and `alpha1` are `Eigen<var>` types. The troublesome thing here is that `y` is often much larger than `b` or `alpha1`. So you might have 50 groups (if they're states), but then 30000 responses (y values). In this case we definitely want `var_value<Eigen>` types fed into the `normal` even though everything might start as `Eigen<var>` types.

bbbales2:
> For multiply() I was just going to do A * B on the inside for static matrices

Just be aware the definitions are different for `row_vector * vector`.

Also `multiply` is easier since we never have to deal with Eigen having defined something internally.

> If we had C++17 we could just use

If C++17 makes it easier to return different types based on conditionals (the use case I know of for this), scrap it and do whatever is cleanest to program, we're rapidly moving towards C++17 anyway right (wasn't it R that was holding it up but now it's upgraded)? No need to design around something that might just get fixed by a compiler upgrade that might happen within a year.

bob-carpenter:
I think C++17 is a ways away.  We also have Python and just because R itself may now make it possible, it doesn't mean everyone else has upgraded.  But that's a discussion for a different thread.

> I want to aggressively convert to var_value<Eigen> types, though I guess it's hard to evaluate this outside of the context of specific models.

Will those still be assignable to Matrix<var, ...> types where necessary?

How does this interact with the parser and type checker?

SteveBronder:
Yeah idt we should depend on bumping up to C++17

> Will those still be assignable to Matrix<var, ...> types where necessary?

`var<Matrix>` will be implicitly convertible to `Matrix<var>`, the Q is how expensive that is. I think it will end up being pricey since it's a full copy

bbbales2:
> the Q is how expensive that is. I think it will end up being pricey since it's a full copy

The expensive part is all the pointer chasing with the varis in chain. The implementation is [here](https://github.com/stan-dev/math/blob/f1f32af911b15e3ca4f991b85f03caba9d93caf5/stan/math/rev/core/propogate_static_matrix.hpp#L13).

> I think C++17 is a ways away.

Bummer.

SteveBronder:
Alright so I started reworking this here but want to chat for a bit before I move forward with this.

There's a few big changes

- I'm trying to put everything together into [`OpMultiply`](https://github.com/stan-dev/math/compare/feature/eigen-vari-ops-adj-jac-apply2...feature/eigen-vari-adj-jac-apply-proto#diff-4d75dd68c0cf62489e92b7584c06fde3R343) and just have [specializations for the reverse pass.](https://github.com/stan-dev/math/compare/feature/eigen-vari-ops-adj-jac-apply2...feature/eigen-vari-adj-jac-apply-proto#diff-4d75dd68c0cf62489e92b7584c06fde3R363)
1. Allocating data is left to the constructor for each functor that goes into `adj_jac_apply`. We'll have to call this constructor when we call `adj_jac_apply`'s constructor
2. There's a new helper class called [`adj_op`](https://github.com/stan-dev/math/compare/feature/eigen-vari-ops-adj-jac-apply2...feature/eigen-vari-adj-jac-apply-proto#diff-0dc9dae1b5c3968c72caf717596dfb2fR55) that handles that allocation. If it sees a var or eigen var it does an allocation, otherwise it's empty.
3. The annoying thing here is that inner struct for calling the right thing in the `reverse_pass`. But there's so many specializations for reverse passes of different structures those specializations will exist somewhere. My thinking was that by just having the specializations in one class we'd cut out the most boilerplate. However this also leads us to template madness a bit.
4. `adj_jac` no longer needs `is_var_` People can either do the specialization thing I'm doing for matmul or they can use the `constexpr` `is_var` available in `adj_op` just like `needs_adj` via.

```cpp
    if(A_.is_var_) {
      Eigen::Map<Eigen::MatrixXd> B(B_mem_, N2_, M2_);
      adjA = adj * B.transpose();
    }

    if(B_.is_var_) {
      Eigen::Map<Eigen::MatrixXd> A(A_mem_, N1_, M1_);
      adjB = A.transpose() * adj;
    }
```
I'm free tmrw if you wanna chat and have some time to look this over

bbbales2:
Had a look at the example op (apologies for delay).

1. I like automatically handling op memory. Especially if we have ops taking weird things like `std::vector<Eigen::Matrix>` types. Seems handy to do that automatically, given we can do regular copies n' such since things need to live on the Stan stack.

2. I don't mind the different specializations though yeah I agree they are verbose. Those will make returning expressions easier, and it's also an optimization that doesn't have to be everywhere (though I'd call the functions `multiply_adjoint_jacobian_impl` just to emphasize they're only there as an implementation of `multiply_adjoint_jacobian`).

3. I don't like that var types have leaked in, both with the `value_of` calls in the constructor and as template types in the `multiply_adjoint_jacobian` calls. Part of `adj_jac_apply` is you can work with simpler types (non-autodiff types, less templates, more explicit things).

Can we achieve this with extra template arguments to `multiply_adjoint_jacobian`?

```
template <typename S, bool needs_adj_0, bool needs_adj_1, ...>
auto multiply_adjoint_jacobian(const S& adj) {
  return multiply_adjoint_jacobian_impl<needs_adj_0, needs_adj_1, ...>(adj); // and maybe _impl is a functor
}
```

4. I think `make_tuple` of an expression should be faster than `forward_tuple` of an eval'd thing. Since we're mapping on the stack alloc, we're always copying and we can't just move pointers, so avoiding the intermediate evals is good.

bob-carpenter:
Would it be more efficient to replace

```
if(A_.is_var_) {
      Eigen::Map<Eigen::MatrixXd> B(B_mem_, N2_, M2_);
      adjA = adj * B.transpose();
}
```

with

```
if(A_.is_var_) {
      Eigen::Map<Eigen::MatrixXd> B(B_mem_, N2_, M2_);
      adjA = B * adj.transpose();
}
```

so that there's a vector transpose rather than a matrix transpose?  Either way, we need to traverse `B` in row major order, so maybe it doesn't matter.  We don't need a final transpose because Eigen lets you assign a row vector to a column vector and vice-versa.

bbbales2:
> Would it be more efficient to replace

Could be. I didn't try this. I'm just finger-crossing that Eigen handles this internally.

bbbales2:
@SteveBronder how are the adj_jac_apply modifications coming? Variadic ODEs is finishing up so I wanted to come back here.

SteveBronder:
Hey I've been reviewing a ton of code and trying to figure out how to get wine to work with R to test out the flto stuff. I'm going to look at it today

bbbales2:
Gimme gimme I wanna get back to rewriting the matrix functions

SteveBronder:
Apologies on the delay working on this rn. Just cleaning up adj_jac_apply like we talked about

SteveBronder:
Closing this because we went in a different direction

Issue #1935, opened 2020-06-11
Full shape match check for nested containers


nhuurre:
## Description
The Stan math library has several functions for checking that two containers have the same size.
* `check_matching_dims`
* `check_matching_sizes`
* `check_consistent_sizes`
* `check_consistent_sizes_mvt`

But these are limited. The first one only operates on matrices. The other three check only the outermost dimension.
Sometimes it's necessary to validate that _all_ dimensions match. For example
```cpp
std::vector<Eigen::VectorXd> a(2, Eigen::VectorXd::Constant(2, 1.0));
std::vector<Eigen::VectorXd> b(N, Eigen::VectorXd::Constant(K, 2.0));
stan::math::check_consistent_sizes("vectorized_add", "A", a, "B", b);
stan::math::apply_scalar_binary(a, b, [=](double x, double y) { return x + y; });
```
* `N=2,K=2`: ok
* `N=3,K=2`: throws exception `"A has size = 2, but B has size 3; and they must be the same size."` (as expected)
* `N=2,K=3`: aborted: assertion failed (very bad)

I need something that can check arbitrary deep dimensions for https://github.com/stan-dev/stanc3/pull/552

#### Current Version:
v3.2.0

PR #1936, opened 2020-06-11
more general check_matching_dims

nhuurre:
## Summary

Generalize `check_matching_dims` so it can handle arbitrary containers and not just matrices. All dimensions must match for deep containers.

## Tests

Expanded `check_matching_dims` unit tests to cover more cases.

## Side Effects

None

## Release notes

Extended check_matching_dims to handle arbitrary containers.

## Checklist

- [X] Math issue #1935

- [X] Copyright holder: Niko Huurre


stan/math/prim/err/check_matching_dims.hpp
```diff
@@ -70,6 +70,43 @@ inline void check_matching_dims(const char* function, const char* name1,
...
+ * @param name2 variable name for the second container (for error messages)
+ * @param y2 second container to test
+ * @throw <code>std::invalid_argument</code> if the dimensions of the
+ *    containers do not match
+ */
+template <typename T1, typename T2>
+inline void check_matching_dims(const char* function, const char* name1,
+                                const std::vector<T1>& y1, const char* name2,
+                                const std::vector<T2>& y2) {
+  check_size_match(function, name1, y1.size(), name2, y2.size());
```
rok-cesnovar, line 105: You could simplify this by using `dims()`: https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/dims.hpp

No need for the scalar overload then, just call dims and iterate over the resulting vector.

I know there is not much code here, so not a huge deal, but still makes sense not to duplicate.

nhuurre: You're right. I wonder why I didn't do that... I think I wanted to preserve `rows of`/`columns of` in the error message when the base type is a matrix. But the error messages have are half-baked anyway...

rok-cesnovar: I think with calling dims the error message might actually be better as you could mention on which dimension sizes differ.

rok-cesnovar:
There is an issue with overloaded dims functions that clashes with the overload in /opencl. Yack. I will push a fix here if you dont mind and then let @t4c1 review since this PR will include my changes also.

Issue #1861, opened 2020-05-02
Probability test framework didn't catch bug with vectorization


bbbales2:
## Description

This pull request went in: https://github.com/stan-dev/math/pull/1331

And later was reverted: https://github.com/stan-dev/math/pull/1530

@mcol had a look why the testing framework did not catch the problem: https://github.com/stan-dev/math/pull/1752#issuecomment-596207740

It looks like the code [here](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L577) tests in psuedo-code:
```
N * logp(a) = logp(rep(a, N))
```

We should also test the more general:
```
logp(a) + logp(b) + ... = logp({a, b, ...})
```

This issue is a two parter:
1. Confirm that such a test would actually have caught https://github.com/stan-dev/math/pull/1331
2. If it would have, then confirm this issue and make a test. If it doesn't, delete this issue.

#### Current Version:
v3.2.0

bbbales2:
This was a duplicate but it seems to reference a different issue: https://github.com/stan-dev/math/issues/1978

PR #1944, opened 2020-06-18
Generalize cauchy

t4c1:
## Summary

Generalizes Cauchy distribution related functions to accept general Eigen expressions.

## Tests
Tested with expression testing framework.

## Side Effects
None

## Release notes
Generalized Cauchy distribution related functions to accept general Eigen expressions.

## Checklist

- [x] Math issue #1470 


SteveBronder:
Before we rewrite all of these do we want to wait for the var_value stuff? It feels like if we don't we'll end up rewriting these distributions multiple times

t4c1:
Some functions I have rewritten using Eigen expressions instead of `scalar_seq_view` (for example `cauchy_lpdf`). I think these willl need no changes for `var_value`. Once `operands_and_partials` and `value_of` support `var_value` these should just work.

Other functions still use `scalar_seq_view` (for example `cauchy_cdf`). Some of these would be slower using Eigen stuff (due to accessing memory multiple times). I think we can eventually do something for Eigen expressions that will work similarly to multi result kernel in kernel generator to fix that. Other function use some other functions, for which we do not have vectorized versions. I think we will eventually have all probability functions use expressions. 

Whatever the reason, these will probably need some changes. I am not sure we can support `var_value` in `scalar_seq_view` in efficient fashion. However the changes I am doing now to these functions are pretty much trivial. So it should not matter too much they will need another change later.

stan/math/prim/prob/cauchy_lpdf.hpp
```diff
@@ -37,13 +38,17 @@ template <bool propto, typename T_y, typename T_loc, typename T_scale>
 return_type_t<T_y, T_loc, T_scale> cauchy_lpdf(const T_y& y, const T_loc& mu,
                                                const T_scale& sigma) {
   using T_partials_return = partials_return_t<T_y, T_loc, T_scale>;
+  using T_partials_array = Eigen::Array<T_partials_return, Eigen::Dynamic, 1>;
   using std::log;
+  using T_y_ref = ref_type_if_t<!is_constant<T_y>::value, T_y>;
```
bbbales2, line 43: What is the logic here? This is the type that will be responsible for .eval() ing the input right?

stan/math/prim/prob/cauchy_lpdf.hpp
```diff
@@ -53,57 +58,76 @@ return_type_t<T_y, T_loc, T_scale> cauchy_lpdf(const T_y& y, const T_loc& mu,
...
-      log_sigma(size(sigma));
-  for (size_t i = 0; i < stan::math::size(sigma); i++) {
-    const T_partials_return sigma_dbl = value_of(sigma_vec[i]);
-    inv_sigma[i] = 1.0 / sigma_dbl;
-    sigma_squared[i] = sigma_dbl * sigma_dbl;
-    if (include_summand<propto, T_scale>::value) {
-      log_sigma[i] = log(sigma_dbl);
-    }
-  }
+  const auto& y_arr = as_array_or_scalar(y_col);
```
bbbales2, line 68: Why does this need to go `as_column_vector_or_scalar` to `as_array_or_scalar`?

Can it not go directly `as_array_or_scalar`?

stan/math/prim/prob/cauchy_lpdf.hpp
```diff
@@ -53,57 +58,76 @@ return_type_t<T_y, T_loc, T_scale> cauchy_lpdf(const T_y& y, const T_loc& mu,
...
+  const auto& mu_arr = as_array_or_scalar(mu_col);
+  const auto& sigma_arr = as_array_or_scalar(sigma_col);
 
-  for (size_t n = 0; n < N; n++) {
-    const T_partials_return y_dbl = value_of(y_vec[n]);
-    const T_partials_return mu_dbl = value_of(mu_vec[n]);
+  ref_type_t<decltype(value_of(y_arr))> y_val = value_of(y_arr);
```
bbbales2, line 72: Why worry about ref types here? We already evaluated everything.

stan/math/prim/prob/cauchy_lpdf.hpp
```diff
@@ -53,57 +58,76 @@ return_type_t<T_y, T_loc, T_scale> cauchy_lpdf(const T_y& y, const T_loc& mu,
...
 
-    if (include_summand<propto>::value) {
-      logp -= LOG_PI;
-    }
-    if (include_summand<propto, T_scale>::value) {
-      logp -= log_sigma[n];
-    }
-    logp -= log1p(y_minus_mu_over_sigma_squared);
+  const auto& inv_sigma
+      = to_ref_if<!is_constant_all<T_scale>::value>(inv(sigma_val));
```
bbbales2, line 82: Why are `to_ref` s needed at this point?

stan/math/prim/prob/cauchy_lpdf.hpp
```diff
@@ -53,57 +58,76 @@ return_type_t<T_y, T_loc, T_scale> cauchy_lpdf(const T_y& y, const T_loc& mu,
...
+        = to_ref_if<!is_constant_all<T_scale>::value>(square(y_minus_mu));
+    if (!is_constant_all<T_y, T_loc>::value) {
+      const auto& mu_deriv
+          = to_ref_if < !is_constant_all<T_y>::value
+            && !is_constant_all<T_loc>::value
+                   > (2 * y_minus_mu / (sigma_squared + y_minus_mu_squared));
+      if (!is_constant_all<T_y>::value) {
+        if (is_vector<T_y>::value) {
+          ops_partials.edge1_.partials_
+              = forward_as<T_partials_array>(-mu_deriv);
```
bbbales2, line 107: What does forward_as do? Why this instead of just forward? Is it casting things?

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: Right, we are only evaling this if it is not constant. The reason being is that `y_ref` seems to be used twice. Once in construction of `operands_and_partials` and once in creation of `y_col`. However, prim implementation of `operands_and_partials` does not actually do anything with the arguemnts. So if y is double it is actually only used once and we don't need to eval it.

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: Distributions can accept combination of row and column vector inputs. To make dimensions match when doing math we are transposing any row vectors.

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: `value_of(y_arr)` is returning an expression if y contains `var`s or `fvar`s.

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: This particular one is here to future-proof the code for when `inv()` will be returning expressions.

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: I prefere the old name of `forward_as`, which was `assume_type`. It assumes that its input is of given type (eigen version also accepts compatible expressions). If the argument type is different it has the assumed type for return type, but it throws. This way we can within ifs with type dependant condition assume of what type a particular variable will be.

It exists just to make the code compile. Otherwise we could not assign a scalar `mu_deriv` to `ops_partials.edge1_.partials_`. But since this assignment is executed within `if (is_vector<T_y>::value) {`, we know `mu_deriv` can not be a scalar.

bbbales2: Oh so it's a runtime type check + placeholder sorta thing

stan/math/prim/prob/cauchy_lpdf.hpp
bbbales2: Right, but that's what `as_column_vector_or_scalar` does.

So we could just do this and not worry about the arrays:
```cpp
const auto& y_col = as_column_vector_or_scalar(y_ref);
const auto& mu_col = as_column_vector_or_scalar(mu_ref);
const auto& sigma_col = as_column_vector_or_scalar(sigma_ref);

ref_type_t<decltype(value_of(y_arr))> y_val = value_of(y_col);
ref_type_t<decltype(value_of(mu_arr))> mu_val = value_of(mu_col);
ref_type_t<decltype(value_of(sigma_arr))> sigma_val = value_of(sigma_col);
```

stan/math/prim/prob/cauchy_lpdf.hpp
bbbales2: Hmm okay and so it's important to use this pattern if we might use the values multiple times (which might vary from function to function but if we always do it we're safe).

stan/math/prim/prob/cauchy_lpdf.hpp
bbbales2: Since it will get used at least once, it would always be safe to eval it though. So we can get rid of the conditional and it'll be the same.

t4c1: Yeah it is safe. However I prefere not to for performance reasons. If we avaluate multiple operations in same expression we need to load the data into cache just once. If we evaluate each operation separately we need to load them every time.

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: Arrays are necesary to do things such as addition of an array and a scalar. Such operation is not defined for matrices. 

It is true, however that these two functions are usually called together. So we could make a single function that does both of these.

stan/math/prim/prob/cauchy_lpdf.hpp
t4c1: Yep. Sooner or later majority of our funcitons will return expressions.

stan/math/prim/prob/cauchy_lpdf.hpp
bbbales2: Oh that's array in the Eigen sense, not the Stan sense. Given that I definitely think it makes sense that `as_column_vector_or_scalar` and `as_array_or_scalar` should take in the same types (since the difference in a column vector an an array is a `.array()`), but that can be a different pull or whatever. This is good.

stan/math/prim/prob/cauchy_lpdf.hpp
bbbales2: I was looking at `ref_type_if_t` to see what it did and it looks like it returns a `plain_type` if it doesn't do a reference type.

That will induce a copy anyway: https://mc-stan.org/math/d2/d08/ref__type_8hpp_source.html

t4c1: Assuming `T_y` is `MatrixXd` this is `ref_type_if_t<false, MatrixXd>`, so its results is set to `T_optionally_ref`, which is `const MatrixXd&`. No copy as I see it.

bbbales2: Right but that will still cause an eval. In which case we're evaling this regardless of it is constant or not and so:

```
using T_y_ref = ref_type_t<T_y>;
```

will achieve the same thing in the end as:

```
using T_y_ref = ref_type_if_t<!is_constant<T_y>::value, T_y>;
```

t4c1: @bbbales2 I hope this is clear now. Or is it ... ?

bbbales2: Yeah it's clear. I delayed cause this (and the other lpdf pulls) are major changes to lpdf and I got nervous and wanted at least https://github.com/stan-dev/math/pull/1989 to get in.

Until that is in there's at least a couple weaknesses in the testing framework that let bugs through before (though only one of those is with pmfs/pdfs - https://github.com/stan-dev/math/issues/1861).

I didn't get around to breaking that pull up Friday.

Issue #1875, opened 2020-05-10
Meta Issue for Static Matrices


SteveBronder:
## Description

This is a meta issue to list out the steps needed for static matrices (list is in no particular order). If I'm missing anything please make a comment and I'll update

- [x] A. Code reviews for the base implementation (I think it could be made a lot better)
- [ ] B. Write a script that replaces vari and var with templated versions in the math library.
    - I think a clever R or Python script could read through the library and add those templates as necessary.
- [ ] C. Write out the derivatives for the operators in core for mixes of matrices, vectors, scalars, etc.   
    - This is just going through Giles and using matrixcalculus.org to template out matrix specializations
- [ ] D. Figure out what stuff can be reused in rev   
    - I'm not totally sure how much can be reused. Things like rev inverse etc probably need specializations, but it looks like a lot of the unary functions can just reuse the same code with some templating. I'd like to support everything that a matrix can currently do and I need some help from people smarter than I to figure out what can be reused with some clever refactoring. 
- [ ] E. Have distributions take in var types with inner Eigen types   
    - Writing out specializations for operands_and_partials, scalar_seq_view, and VectorBuilder. 
    - This can only be done once the underlying functions distributions depend on accept static matrices
    - We will also need to add static matrices to the distribution test suite 
- [ ] F. Refactor the AD test suite to work on static matrices    - There's a lot of places in the AD test suite that assumes the var's inner type is double. The steps for this could be to implement the derivatives of the operators first, then once those are in refactor the AD test suite / add a specialization for matrix var types.

#### Current Version:
v3.2.0

bbbales2:
> C. Write a script that replaces vari and var with templated versions in the math library.
> D. Figure out what stuff can be reused in rev

I think this is where my head is at on this stuff. I don't think I'll know how to judge progress until we've identified all the things that can be converted. Which is a matter of going through rev/fun. Maybe I'll do it later today, but if someone else beats me to it, I'm not complaining.

> F. Refactor the AD test suite to work on static matrices

I think we could just go through each argument and copy all the `var<Eigen::Matrix<double, R, C>>` s to `Eigen::Matrix<var, R, C>` s and that would work as a first hack.

I don't know how this works with higher order autodiff. I guess the same way as anything does, so I guess we just gotta try it and see what blows up.

bbbales2:
So I went through and looked through the functions in prim and rev (manually, so error prone in that way).

[...]

PR #1952, opened 2020-06-28
Add Eigen Dense and Sparse vari_value types

SteveBronder:
## Summary

Adds a dense and sparse matrix specialization to `vari_value`. The dense and sparse `vari_value` matrix types use an `Eigen::Map` to hold stack allocated memory for the `val_` and `adj_`.

## Tests

This needs more testing, but I could use some help figuring out what needs done. Right now I'm testing that 

1. Eigen matrix expressions are allowed to be passed into the constructor
2. Standard `var_value(var_value())` and `var_value(vari_value())` construction works
3. That member vals are the same as the input vals for the above cases.

## Side Effects

Yes, we need to be careful when constructing `vari_value` and `var_value` types that the template type is a plain eigen type (aka not an Expression just an `Matrix`, `Array`, or `SparseMatrix`, etc.). I think I'll add an `is_plain_type` type trait to check that templates for `vari_value` are actually plain types.

## Release notes

Allows `vari` to hold an Eigen type

## Checklist

- [x] Math issue #1875 


SteveBronder:
@dpsimpson @rtrangucci would yinz mind having a look at the sparse matrix `vari_value` specialization? It looks a bit weird because unlike dense matrices you can't do assignment from a sparse matrix to a sparse map. That's why that `make_adj` and `make_val` functions just iterate over the mem of the sparse matrix

https://godbolt.org/z/Yd9Xeg

dpsimpson:
In a moment of probable stupidity, I can't work out why 
```  template <typename S>
  inline eigen_map make_val(eigen_index*& outer_mem, eigen_index*& inner_mem,
                            eigen_scalar*& val_mem, S&& x) {
    for (int k = 0; k < x.innerSize(); ++k) {
      val_mem[k] = x.valuePtr()[k];
      inner_mem[k] = x.innerIndexPtr()[k];
    }
    for (int k = 0; k < x.outerSize(); ++k) {
      outer_mem[k] = x.outerIndexPtr()[k];
    }
    return eigen_map(x.rows(), x.cols(), x.nonZeros(), outer_mem, inner_mem,
                     val_mem);
}
```
isn't just 

```
template <typename S>
  inline eigen_map make_val(eigen_index*& outer_mem, eigen_index*& inner_mem,
                            eigen_scalar*& val_mem, S&& x) {
    return eigen_map(x.rows(), x.cols(), x.nonZeros(), x.outerIndexPtr(), x.innerIndexPtr(),
                     x.valuePtr());
}
```
(ie why do you need to do the copy?)

I'm also not sure where the problem is here? Can you point to a bit of code that is different for matrices compared to sparse matrices?

SteveBronder:
Yes I was super surprised and bummed when just returning an eigen map didn't work! There's a few things

1. We want a copy so that the sparse vari's mem sits on our stack allocator and is not allocated by Eigen. I thought we didn't need to do this allocation at first but running valgrind / heaptrack I found that the memory eigen created in the vari was leaking out / not being cleaned up.

2. There's some really weird things about sparse maps I couldn't wrap my head around. If the input sparse matrix is const then the map's type has to be const as well i.e. the map has to be `Map<const SparseMatrix>`. But Eigen's sparse matrix impl won't let you assign a const sparse map to a sparse map! Here's an example

https://godbolt.org/z/BSyVZB

You can see in the specialization of matrix vari this works perfectly fine for dense types!

```cpp
template <typename S>
inline eigen_map make_val(eigen_scalar*& mem, S&& x) {
  eigen_map(mem, x.rows(), x.cols()) = std::forward<S>(x);
  return eigen_map(mem, x.rows(), x.cols());
}
```

It seems that Map sort of means two different things for dense and sparse matrices! In the dense case it copies the data from the input matrices pointers to the map's pointers. I think in the sparse case Map is actually copying over Eigen's pointers instead of the data the pointer points to. 

So instead of dealing with Map and yadayada I'm like fine fine we know what we want copied over so we'll just do it manually

I was very confused by (2) and filed an issue with Eigen asking about it (as well as why Eigen won't let you assign a sparse matrix to a map!)
https://gitlab.com/libeigen/eigen/-/issues/1929#note_369540956

bob-carpenter:
There's a third stack to push on classes whose destructors need to be called.  Or you can allocate using the allocator and wrap with an `Eigen::Map`.  I wanted to move fully to the latter forms and remove the third autodiff stack, but others objected.

How does assigning a sparse matrix to a map make sense?  They'd have to have the same sparsity structure, right?

dpsimpson:
For anything that works with vars the lhs and rhs of any sparse matrix
assignment must have the same sparsity pattern. Otherwise we will be adding
and dropping vars and we will break all of our algorithms.

SteveBronder:
> There's a third stack to push on classes whose destructors need to be called. Or you can allocate using the allocator and wrap with an Eigen::Map. I wanted to move fully to the latter forms and remove the third autodiff stack, but others objected.

Yeah so after doing more deep diving I was trying to use the latter, but sparse matrices can realloc or alloc mem for inplace assignments like

```cpp
  using sparse_mat = Eigen::SparseMatrix<double>;
  sparse_mat x = make_sparse_matrix_random(10, 10); // makes a 10x10 sparse matrix
  x_map = Eigen::Map<sparse_mat>(x.rows(), x.cols(),
     x.nonZeros(),x.outerIndexPtr(), x.innerIndexPtr(),
     x.valuePtr());
  x_map += x; // fails
```

https://godbolt.org/z/-Wi4RE

We need to be able to do the `+=` for the adjoint accumulation right?

> How does assigning a sparse matrix to a map make sense? 

I forgot about the above with `+=` now idt it does make sense to use a map.

> They'd have to have the same sparsity structure, right?
> For anything that works with vars the lhs and rhs of any sparse matrix
assignment must have the same sparsity pattern. Otherwise we will be adding
and dropping vars and we will break all of our algorithms.

As long as the sparsity pattern for each iteration of the model is the same that's fine right? Because doing something like the above with `+=` will change the sparsity pattern as the `.chain()` methods are called for the grad calls. It can change across iterations if someone used a conditional statement at the language level, but I think that convo is better discussed when we expose these at the language level

dpsimpson:
> As long as the sparsity pattern for each iteration of the model is the same that's fine right? Because doing something like the above with += will change the sparsity pattern as the .chain() methods are called for the grad calls. It can change across iterations if someone used a conditional statement at the language level, but I think that convo is better discussed when we expose these at the language level

If `A` is a sparse matrix, then if, for examplle, the entry `A[1,2]` is non-zero for *any* step of the algorithm, it must be non-zero for all steps of the algorithm. Because we will spit out a whole chain of values for `A[1,2]` and none of our algorithms will work if it is sometimes identically zero (that would be the same as having a discrete variable indicating which entries are non-zero at each step of the algorithm, and we cannot do that).

So we can't allow `+=` or `.chain()` to change the sparsity pattern.  (Actually simple arithmetic operations can probably be dealt with in the language, but something more complex like `.chain()` wouldn't be able to be.)

It's worth noting that all of the linear algebra things do respect the sparsity structure, so I don't think `.chain()` should ever actually work on anything other than the values of the sparse matrix (ie they should not touch the inner or outer indices)

SteveBronder:
> If A is a sparse matrix, then if, for example, the entry A[1,2] is non-zero for any step of the algorithm, it must be non-zero for all steps of the algorithm. Because we will spit out a whole chain of values for A[1,2] and none of our algorithms will work if it is sometimes identically zero (that would be the same as having a discrete variable indicating which entries are non-zero at each step of the algorithm, and we cannot do that).

> So we can't allow += or .chain() to change the sparsity pattern. (Actually simple arithmetic operations can probably be dealt with in the language, but something more complex like .chain() wouldn't be able to be.)


So I'm trying to think here and it feels like I'm missing something. I'm looking at the code for `grad()` and we iterate over all of the vari in the stack on each grad call

```cpp
template <typename Vari>
static void grad(Vari* vi) {
  vi->init_dependent();
  size_t end = ChainableStack::instance_->var_stack_.size();
  size_t beginning = empty_nested() ? 0 : end - nested_size();
  for (size_t i = end; i-- > beginning;) {
    ChainableStack::instance_->var_stack_[i]->chain();
  }
}
```

Since it goes over the whole stack once we run `grad()` the sparsity structure won't change once we run grad the first time. Is that enough to be okay or do we ever look at the the adjoints before calling `grad()`?  At the algorithm level I get pretty confused. 

Though if someone ever calls `prune()` on one of these adjoints everything goes to doomtown.


> It's worth noting that all of the linear algebra things do respect the sparsity structure, so I don't think .chain() should ever actually work on anything other than the values of the sparse matrix (ie they should not touch the inner or outer indices)

Can you explain this more? I'm not sure why the sparsity pattern wouldn't change for the lin alg but would change for simpler ops?

dpsimpson:
> Though if someone ever calls prune() on one of these adjoints everything goes to doomtown.
Yep. We could probably write a github action / linter check that checks if a PR uses a bad sparse matrix function like `prune` and yells loudly. It's a pretty finite list.

> Can you explain this more? I'm not sure why the sparsity pattern wouldn't change for the lin alg but would change for simpler ops?

Just like you can think of a `matrix[m,n]` as a `m * n` array of vars that only know about their mathematical structure when you call `.chain()`, the same happens with `sparse_matrix[n, m, nnz]`, which is just a `nnz`-dimensional array of vars. And just like an ordinary matrix doesn't "create" new vars when you call `.chain()` neither does the sparse matrix. 

Mathematically, the only adjoints that are defined are the ones for the non-zero elements of the sparse matrix. Because the zero elements are constantly zero.  It means you've got to do all of the formulas carefully and properly for computing the adjoints, but it turns out you just drop the zero elements everywhere. It's neat!

**So I think I understand your design??** 
You're working on the assumption that the sparsity pattern will stay fixed and it can be pulled out and fed to the reader etc when the gradients and values are computed. That is *cool*!

But I've got two questions: 
1) Is `stan::io::reader` smart enough to deduce the sparsity pattern / number of non-zeros in the sparse matrix? 
2) Will the stanc3 compiler be smart enough to notice bad things like a conditional where the condition is not data or transformed data that will change the sparsity of a matrix? 

If those to things work and if we can verify that this code does what it should 
```
double zero = 0.0;
double not_zero = 1.1;
Eigen::SparseMatrix<double> A = make_sparse_matrix_random(10, 10);
Eigen::SparseMatrix<double> B = make_sparse_matrix_random(10, 10);
Eigen::SparseMatrix<double> C = A + zero * B;
Eigen::SparseMatrix<double> D = A + not_zero * B;

// Check that C and D have the same sparsity pattern!
```

then we are safe! Basically we just need eigen not to be smart with multiplication by zero!

t4c1:
If there as still open quastions regarding sparse support, I suggest you split dense/sparse into separate PRs. I would like to start working on `var<matrix_cl>`, but I think it would be good to have `var<dense matrix>` in first.

Also is this PR intentionally not merging into develop?

SteveBronder:
Thanks @dpsimpson that makes sense and I think we are aligned now!

> Yep. We could probably write a github action / linter check that checks if a PR uses a bad sparse matrix function like prune and yells loudly. It's a pretty finite list.

Yes thats a very good idea!

> But I've got two questions:

> 1. Is stan::io::reader smart enough to deduce the sparsity pattern / number of non-zeros in the sparse matrix?
> 2. Will the stanc3 compiler be smart enough to notice bad things like a conditional where the condition is not data or transformed data that will change the sparsity of a matrix?

@rybern came up with a good idea for this [here](https://github.com/stan-dev/stanc3/pull/491#issuecomment-620134464) but tl;dr yes! Sparse matrices will have a nice feature that you can define the sparsity pattern using sparse matrices defined in the transformed data block. Since the user defines the sparsity pattern either in data, transformed data, or sparse matrices defined in transformed data the model can know/verify the sparsity pattern for each matrix is correct.

> If those to things work and if we can verify that this code does what it should

I will check this out!

> If there as still open quastions regarding sparse support, I suggest you split dense/sparse into separate PRs. I would like to start working on var<matrix_cl>, but I think it would be good to have var<dense matrix> in first.

I think the sparse stuff is resolved

> Also is this PR intentionally not merging into develop?

Yes, so we are waiting till after the coming release (in like 3 weeks or so) to merge in `vari-base-templates` so we have a whole release cycle to make sure flto etc works upstream. Until then the plan is to treat `feature/vari-base-templates` as the base branch that vari matrix stuff will be merged into. So if you would like to work on matrix_cl var stuff (I'm very excited about that!) you can make the branch either directly off of `vari-base-templates` or this one if we need to swap back and forth between `vari<Eigen>` and `vari<matrix_cl>` . I think it would be good to have the base `vari<matrix_cl>` stuff in here now so we can think about them as we are building out the matrix ops

SteveBronder:
@bbbales2 can you take a look at this? Once it looks good we can merge it into `feature/vari-base-templates` so we don't have to maintain a bunch of branches

bbbales2:
@SteveBronder thanks for the ping, yeah I'll have a look through this later today.

SteveBronder:
Word ty! I'm looking over the adj_jac_apply stuff rn

bbbales2:
> I'm looking over the adj_jac_apply stuff rn

Nice!

stan/math/prim/meta/is_eigen_sparse_base.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/prim/meta/bool_constant.hpp>
+#include <stan/math/prim/meta/is_base_pointer_convertible.hpp>
+#include <stan/math/prim/meta/require_helpers.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Checks whether type T is derived from Eigen::MatrixBase.
```
bbbales2, line 13: Eigen::SparseMatrixBas

stan/math/rev/core/var.hpp
```diff
@@ -34,11 +34,22 @@ template <typename T, typename = void>
...
 template <typename T>
-class var_value<T, require_floating_point_t<T>> {
+class var_value<T, require_vt_floating_point<T>> {
+  /**
+   * Checks whether the template type `T` is an assignable type.
+   */
+  template <typename S>
+  using check_plain_type
+      = std::is_same<std::decay_t<S>, std::decay_t<plain_type_t<S>>>;
+  static_assert(
```
bbbales2, line 44: Nice I want to use these a bunch in `adj_jac_apply` so it's easier to work with.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+  using PlainObject = std::decay_t<plain_type_t<T>>;
+  using eigen_scalar = value_type_t<PlainObject>;  // A floating point type
+  eigen_scalar* val_mem_;  // Pointer to memory allocated on the stack for val_
+  eigen_scalar* adj_mem_;  // Pointer to memory allocated on the stack for adj_
+  using Scalar = PlainObject;  // The underlying type for this class
+  using value_type = Scalar;   // The underlying type for this class
+  using eigen_map = Eigen::Map<PlainObject>;  // Maps for adj_ and val_
+  const Eigen::Index rows_;                   // Number of rows
+  const Eigen::Index cols_;                   // Number of cols
+  const Eigen::Index size_;                   // Size of rows * cols
```
bbbales2, line 216: Can these all be functions that access the maps? Seems like they take up unnecessary memory.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+  /**
+   * Number of rows known at compile time
+   */
+  static constexpr Eigen::Index RowsAtCompileTime = Scalar::RowsAtCompileTime;
+  /**
+   * Number of columns known at compile time
+   */
+  static constexpr Eigen::Index ColsAtCompileTime = Scalar::ColsAtCompileTime;
```
bbbales2, line 224: These don't actually take memory right?

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+  const Eigen::Index size() const { return size_; }
+
+  virtual void chain() {}
+  /**
+   * Initialize the adjoint for this (dependent) variable to 1.
+   * This operation is applied to the dependent variable before
+   * propagating derivatives, setting the derivative of the
+   * result with respect to itself to be 1.
+   */
+  void init_dependent() { adj_.setOnes(); }
```
bbbales2, line 319: There's no obvious default for this. I am confused as to what to do here. I think maybe we should just turn off `.grad()` and `.init_dependent()` for these types.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   */
+  const Eigen::Index size() const { return size_; }
+  void chain() {}
+  /**
+   * Initialize the adjoint for this (dependent) variable to 1.
+   * This operation is applied to the dependent variable before
+   * propagating derivatives, setting the derivative of the
+   * result with respect to itself to be 1.
+   */
+  inline void init_dependent() {
```
bbbales2, line 516: Same as above. I think we should just not allow `.grad()` and `.init_dependent()` for these types.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+    : public vari_base {
+ public:
+  /**
+   * `PlainObject` represents a user constructible type such as Matrix or Array
+   */
+  using PlainObject = std::decay_t<plain_type_t<T>>;
+  using eigen_scalar = value_type_t<PlainObject>;  // A floating point type
+  eigen_scalar* val_mem_;  // Pointer to memory allocated on the stack for val_
+  eigen_scalar* adj_mem_;  // Pointer to memory allocated on the stack for adj_
+  using Scalar = PlainObject;  // The underlying type for this class
```
bbbales2, line 211: This isn't a scalar in the math sense.

bbbales2:
Will `is_convertible` handle Eigen expressions correctly? And it's slightly weird that we'd do `is_convertible` against type `value_type` in the dense stuff -- technically it all needs convertible to an Eigen Map? Would that ever cause problems?

dpsimpson:
> I don't have experience with Eigen sparse stuff so hard to say. There was something about Eigen maps not working right for sparse types? Does that mean we get unnecessary copying here?

I’m somewhat familiar with the sparse stuff but I’m not knowledgeable this deep into the Stan infrastructure. But if someone wants to go line by line on a video call I’d be happy to. (I hope there’s someone better to review, but if there isn’t let’s do the slow team-y thing!) it’d have to be this week because I am on a stop-work holiday starting 3 days ago and am enforcing that by not having home internet from Monday)

t4c1:
> Until then the plan is to treat feature/vari-base-templates as the base branch that vari matrix stuff will be merged into.

So you plan to merge everything into one branch and than open one huge PR to merge this into develop? I'm not a fan of huge PRs and I'd like to review this.

> So if you would like to work on matrix_cl var stuff (I'm very excited about that!) you can make the branch either directly off of vari-base-templates or this one if we need to swap back and forth between vari<Eigen> and vari<matrix_cl> .

I will try to minimize the amount of code duplication for scalar/Eigen/OpenCL versions of var, which probably involves some changes to what you have here. If I do that while you are still changing those, it will lead to lots of painful merge conflicts. So I'd prefere to wait until you are mostly done with those.

rok-cesnovar:
> So you plan to merge everything into one branch and then open one huge PR to merge this into develop?

It's not ideal but if we are strict we can probably make it work.

The base PR has to be approved and once that is done all subsequent commits must only be merges of PRs to that branch. In which case the final huge PR does not have to be reviewed again in my opinion.

t4c1:
Maybe we should setup another develop-like branch for this. Start it as a copy of develop, than require any merges towards it to be tested by Jenkins and reviewed. Than the final merge into develop really need not be reviewed again.

rok-cesnovar:
Not sure how is that different from the `feature/vari-base-templates` branch (base branch & PR). These subsequent PRs are all tested on Jenkins as well.

t4c1:
Oh, right they are. The base was reviewed and assuming we are all adults, we don't need the base branch to be protected. Than I will just wait untill this PR is merged, before I start with the matrix_cl stuff.

rok-cesnovar:
Oh yeah, I see what you mean. Yes, no one should merge PR to feature/vari-base-templates without review/passing tests. If we respect that we are in the clear.

stan/math/rev/core/var.hpp
```diff
@@ -77,14 +88,13 @@ class var_value<T, require_floating_point_t<T>> {
    * @param x Value of the variable.
    */
   template <typename S, require_convertible_t<S&, value_type>* = nullptr>
-  var_value(const S& x) : vi_(new vari_type(x, false)) {}  // NOLINT
+  var_value(S&& x) : vi_(new vari_type(std::forward<S>(x), false)) {}  // NOLINT
```
t4c1, line 91: We don't need perfect forwarding here if vari is making a copy anyway. When I suggested perfect forwarding, I assumed vari will be holding `Eigen::Matrix` so moving would be possible.

stan/math/rev/core/vari.hpp
t4c1: Nope.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   * Insertion operator for vari. Prints the current value and
+   * the adjoint value.
+   *
+   * @param os [in, out] ostream to modify
+   * @param v [in] vari object to print.
+   *
+   * @return The modified ostream.
+   */
+  friend std::ostream& operator<<(std::ostream& os, const vari_value<T>* v) {
+    return os << v->val_ << ":" << v->adj_;
```
t4c1, line 338: To make printing those readable we need (at least one) newline between val and adjoint.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+  friend class var_value;
+  /**
+   * Create the map to the val_ stack allocated memory for an Eigen input.
+   * @tparam S an Eigen type.
+   * @param mem A pointer to stack allocated memory.
+   * @param x The Eigen type whose values will be assigned to `val_`
+   */
+  template <typename S>
+  inline eigen_map make_val(eigen_scalar*& mem, S&& x) {
+    eigen_map(mem, x.rows(), x.cols()) = std::forward<S>(x);
```
t4c1, line 380: No point in forwarding if the types can not be the same.

stan/math/rev/core/vari.hpp
```diff
@@ -184,6 +186,400 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+  template <typename, typename>
+  friend class var_value;
+  /**
+   * Create the map to the val_ stack allocated memory for an Eigen input.
+   * @tparam S an Eigen type.
+   * @param mem A pointer to stack allocated memory.
+   * @param x The Eigen type whose values will be assigned to `val_`
+   */
+  template <typename S>
+  inline eigen_map make_val(eigen_scalar*& mem, S&& x) {
```
t4c1, line 379: The first argument is not needed. Member function can access member variable.

stan/math/rev/core/vari.hpp
SteveBronder: Good call!

stan/math/rev/core/vari.hpp
SteveBronder: Confused by what you mean here. Why wouldn't we set the matrix to ones?

SteveBronder: By the above I mean to ask why would `.grad()` and `init_dependent()` not make sense?

stan/math/rev/core/vari.hpp
SteveBronder: Yeah should I just remove it and only have `value_type`? Or shoudl scalar return back the type in the eigen matrix?

SteveBronder:
> I think for the dense types we should support read accessors, both element-wise and the eigen block operations: https://eigen.tuxfamily.org/dox/group__TutorialBlockOperations.html, but that can wait until there's a definite need (which will probably come up sometime in converting functions).

Yes! I think we can make add a constructor that takes in slices of the adjoint and value matrices and doesn't allocate any mem.

SteveBronder:
> Will is_convertible handle Eigen expressions correctly? And it's slightly weird that we'd do is_convertible against type value_type in the dense stuff -- technically it all needs convertible to an Eigen Map? Would that ever cause problems?

I have tests it's passing for but can add some tests to verify that!

> I’m somewhat familiar with the sparse stuff but I’m not knowledgeable this deep into the Stan infrastructure. But if someone wants to go line by line on a video call I’d be happy to. (I hope there’s someone better to review, but if there isn’t let’s do the slow team-y thing!) it’d have to be this week because I am on a stop-work holiday starting 3 days ago and am enforcing that by not having home internet from Monday)

Would love your help though I think you should respect your stop-work holiday! If you want to get on a quick call I'm free pretty much all week besides Thursday after 6:00 PM

stan/math/prim/meta/is_eigen_dense_base.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/prim/meta/bool_constant.hpp>
+#include <stan/math/prim/meta/is_base_pointer_convertible.hpp>
+#include <stan/math/prim/meta/require_helpers.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Checks whether type T is derived from Eigen::MatrixBase.
```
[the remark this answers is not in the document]
SteveBronder: I need to add tests for these as well

stan/math/rev/core/var.hpp
```diff
@@ -34,11 +34,22 @@ template <typename T, typename = void>
...
+      = std::is_same<std::decay_t<S>, std::decay_t<plain_type_t<S>>>;
+  static_assert(
+      check_plain_type<T>::value,
+      "The template for this var is an"
+      " expression but a var_value's inner type must be assignable such as"
+      " a double, Eigen::Matrix, or Eigen::Array");
+
  public:
-  using value_type = std::decay_t<T>;        // Numeric type in vari_value.
+  using value_type = std::decay_t<std::decay_t<T>>;  // type in vari_value.
```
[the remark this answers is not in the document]
SteveBronder: None! Good catch

stan/math/rev/core/var.hpp
SteveBronder: Yeah I'll move this into it's own thing

stan/math/rev/core/var.hpp
SteveBronder: Yeah my b. If we used `chainable_alloc` we could we could have Eigen matrices here. Though idk of the cost/benefit of that my priors feel like it would be slower

SteveBronder: Yes agree. Should we also have two separate constructors here, one for eigen and one for everything else, that wraps `x` into a `ref_type_t` so that if an expression comes in here it's only evaluated once?

stan/math/rev/core/vari.hpp
bbbales2: As far as I know, the adjoints in a reverse mode pass always represent gradients of one value. When we do:

```
var a;
var b = a * a;
b.grad()
```

then the adjoint of `b` is `db/db` and the adjoint of `a` is `db/da`.

If we had something like:
```
var_value<Matrix> a;
var_value<Matrix> b = a * a;
b.grad();
```

What do the adjoints of `b` or the adjoints of `a` mean on their own?

Initializing the adjoints of `b` to all 1s is like implicitly defining `b.grad();` as equivalent to `sum(b).grad()`.

I think it's entirely reasonably you'd want to initialize the adjoints of `b` to something and then call `grad(b.vi_)`. That's how  you'd compute `row_vector * Jacobian(f)` products, but then you'd want to initialize the adjoints to some specific values and call grad manually, not work through the `b.grad()` + `init_dependent` interface.

I don't think I did a terribly good job of making myself clear there. @bob-carpenter and @betanalpha might have suggestions here on the interface here.

SteveBronder:
> I think for the dense types we should support read accessors, both element-wise and the eigen block operations: https://eigen.tuxfamily.org/dox/group__TutorialBlockOperations.html, but that can wait until there's a definite need (which will probably come up sometime in converting functions).

For this do you think we would need to make copies?

bbbales2:
> I have tests it's passing for but can add some tests to verify that!

I think this probably isn't necessary and I was worrying too much when I saw that. The map/matrix compatibility assumption is probably a fine enough. Let it breaks when it breaks, otherwise we're just guessing tests to add and I don't think a random walk in test space is going to be that useful.

stan/math/rev/core/var.hpp
t4c1: No, one constructor is fine. Wraping in `ref_type_t` would just make a redundant copy.

>> my priors feel like it would be slower

I think as soon as a model can use move instead of copy here on one big matrix it will outweight several destructor calls.

t4c1:
> Yes! I think we can make add a constructor that takes in slices of the adjoint and value matrices and doesn't allocate any mem.

Maybe we could allow `var_value<Eigen::Block>` instead of another constructor?

> Will is_convertible handle Eigen expressions correctly? And it's slightly weird that we'd do is_convertible against type value_type in the dense stuff -- technically it all needs convertible to an Eigen Map? Would that ever cause problems?

There should not be any problems. Map is essentisally a matrix with externally allocated memory.

SteveBronder:
I got the slices (`block`, `head`, `tail`, `segment`) working with var and vari maps in the branch below. But it's a lot of changes so I think it may be good to have in a separate PR. Unless @bbbales2 and @t4c1 this it's fine to merge into this branch

https://github.com/stan-dev/math/tree/vari-base-eigen-slices

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +177,393 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   */
+  template <typename S, require_convertible_t<S&, value_type>* = nullptr>
+  explicit vari_value(S&& x)
+      : val_mem_(ChainableStack::instance_->memalloc_.alloc_array<eigen_scalar>(
+            x.size())),
+        adj_mem_(ChainableStack::instance_->memalloc_.alloc_array<eigen_scalar>(
+            x.size())),
+        adj_(make_adj(x)),
+        val_(make_val(x)) {
+    ChainableStack::instance_->var_stack_.push_back(this);
```
t4c1, line 263: How about we completely eliminate these private functions?

```suggestion
        adj_(adj_mem_, x.rows(), x.cols()),
        val_(val_mem_, x.rows(), x.cols()) {
    adj_.setZero();
    val_ = x;
    ChainableStack::instance_->var_stack_.push_back(this);
```

t4c1:
Lets go with a separate PR.

stan/math/rev/core/vari.hpp
SteveBronder: Yes I figured out how to do this over the weekend!

SteveBronder:
@t4c1 @bbbales2 have either of yinz seen or know how to fix this error?

https://gist.github.com/SteveBronder/3b45d456af88cc859168a57e712d2308

It's only happening in the distribution tests on the jenkins linux machine and I can't replicate it locally

SteveBronder:
The tldr is this

```
In file included from test/prob/double_exponential/double_exponential_cdf_00000_generated_ffd_test.cpp:3:
In file included from ./test/prob/test_fixture_distr.hpp:4:
In file included from ./stan/math/mix.hpp:4:
In file included from ./stan/math/mix/meta.hpp:10:
In file included from ./stan/math/rev/core.hpp:13:
In file included from ./stan/math/rev/core/init_chainablestack.hpp:6:
In file included from lib/tbb_2019_U8/include/tbb/task_scheduler_observer.h:20:
In file included from lib/tbb_2019_U8/include/tbb/atomic.h:28:
In file included from lib/tbb_2019_U8/include/tbb/tbb_machine.h:245:
In file included from lib/tbb_2019_U8/include/tbb/machine/linux_common.h:44:
In file included from /usr/include/linux/futex.h:5:
/usr/include/linux/types.h:27:9: error: unknown type name '__u16'
typedef __u16 __bitwise __le16;
```

I can't figure out the include order error that's happening here that for some reason would cause this. @serban-nicusor-toptal have you seen this before? It looks like the tbb is looking at the wrong header for some reason?

rok-cesnovar:
This is not related to this PR, the gelman group linux is not being too friendly. As if it knows its release time :)

Nic fixed this fotmr now by excluding that machine from distr. tests.

rok-cesnovar:
We still dont know what changed on the machine though.

SteveBronder:
LoL I need to ask these questions earlier spent too much time hacking away at this.

Yeah it seems like a really weird clang/linux/tbb mix of a bug

SteveBronder:
@rok-cesnovar @serban-nicusor-toptal does the current error on Jenkins look like the error we were getting from the other box?

```
In file included from <built-in>:1:
In file included from ./stan/math/rev/functor/reduce_sum.hpp:5:
In file included from ./stan/math/prim/functor.hpp:25:
In file included from ./stan/math/prim/functor/reduce_sum.hpp:7:
In file included from lib/tbb_2019_U8/include/tbb/task_arena.h:20:
In file included from lib/tbb_2019_U8/include/tbb/task.h:21:
In file included from lib/tbb_2019_U8/include/tbb/tbb_machine.h:245:
In file included from lib/tbb_2019_U8/include/tbb/machine/linux_common.h:44:
In file included from /usr/include/linux/futex.h:5:
/usr/include/linux/types.h:27:9: error: unknown type name '__u16'
typedef __u16 __bitwise __le16;
        ^
/usr/include/linux/types.h:28:9: error: unknown type name '__u16'
typedef __u16 __bitwise __be16;
```

rok-cesnovar:
Yep. It probably ran on gelman-group-linux. I am AFK right now, can check later.

serban-nicusor-toptal:
Hey, the gelman-linux has this error. It was happening in distribution tests so I've changed the machine for those but didn't knew about the headers check.
Looking into it now.

SteveBronder:
Thanks Serban and Rok!

rok-cesnovar:
@SteveBronder can you merge in develop? It currently runs the async tests that fail due to gelman-group-linux issues. Develop commented those out until we figure this out.

SteveBronder:
Done!

SteveBronder:
@bbbales2 when you get a chance can you take another look over this so we can merge it to the base branch? I think the `matrix_cl` and `ops_and_partials` branches are getting close as well and we can merge all this down to one branch

bbbales2:
@SteveBronder can do. I want to give the tests in a base branch a lookover too.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+  }
+
+  /**
+   * Construct an dense Eigen variable implementation from a value. The
+   *  adjoint is initialized to zero and if `stacked` is `false` this vari
+   *  will be not be put on the var_stack. Instead it will only be put on
+   *  a stack to keep track of whether the adjoint needs to be set to zero.
+   *
+   * All constructed variables are added to a stack.  Variables
```
bbbales2, line 261: This paragraph seems outdated considering the one above it.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   *  to insure proper partial derivative propagation.  During
+   *  derivative propagation, the chain() method of each variable
+   *  will be called in the reverse order of construction.
+   *
+   * @tparam S A dense Eigen type that is convertible to `value_type`
+   * @param x Value of the constructed variable.
+   * @param stacked If false will put this this vari on the nochain stack so
+   * that its `chain()` method is not called.
+   */
+  template <typename S, require_convertible_t<S&, value_type>* = nullptr>
```
bbbales2, line 272: I prefer `T` to `value_type`.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   * Return the number of columns for this class's `val_` member
+   */
+  const Eigen::Index cols() const { return val_.rows(); }
+  /**
+   * Return the size of this class's `val_` member
+   */
+  const Eigen::Index size() const { return val_.size(); }
+
+  virtual void chain() {}
+  /**
```
bbbales2, line 301: I don't think init_dependent makes sense here (there's no single obvious thing to initialize it to -- reverse mode differentiation gets us gradients of one thing -- not gradients of a bunch of things [every element in a matrix]). I'd like to just get rid of it if possible.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   * Insertion operator for vari. Prints the current value and
+   * the adjoint value.
+   *
+   * @param os [in, out] ostream to modify
+   * @param v [in] vari object to print.
+   *
+   * @return The modified ostream.
+   */
+  friend std::ostream& operator<<(std::ostream& os, const vari_value<T>* v) {
+    return os << "val: \n" << v->val_ << " \nadj: \n" << v->adj_;
```
bbbales2, line 326: I'm not gonna complain if we take the `vari_value` s in by reference here and in the `double` implementation.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+    this->set_zero_adjoint();
+    ChainableStack::instance_->var_stack_.push_back(this);
+  }
+  /**
+   * Construct an sparse Eigen variable implementation from a value. The
+   *  adjoint is initialized to zero and if `stacked` is `false` this vari
+   *  will be not be put on the var_stack. Instead it will only be put on
+   *  a stack to keep track of whether the adjoint needs to be set to zero.
+   *
+   * All constructed variables are added to a stack.  Variables
```
bbbales2, line 427: Same comment as previous.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+    }
+  }
+
+  /**
+   * Set the adjoint value of this variable to 0.  This is used to
+   * reset adjoints before propagating derivatives again (for
+   * example in a Jacobian calculation).
+   */
+  inline void set_zero_adjoint() noexcept final {
+    for (int k = 0; k < adj_.outerSize(); ++k) {
```
bbbales2, line 482: Is this the only difference in dense and sparse matrices? The rest looks really similar.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+ * A variable implementation is constructed with a constant
+ * value. It also stores the adjoint for storing the partial
+ * derivative with respect to the root of the derivative tree.
+ *
+ */
+template <typename T>
+class vari_value<T, std::enable_if_t<is_eigen_sparse_base<T>::value>>
+    : public vari_base, chainable_alloc {
+ public:
+  using PlainObject
```
bbbales2, line 379: Do we need to assert things aren't expressions or does the enable_if handle that?

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+ * This class is complete (not abstract) and may be used for
+ * constants.
+ *
+ * A variable implementation is constructed with a constant
+ * value. It also stores the adjoint for storing the partial
+ * derivative with respect to the root of the derivative tree.
+ *
+ */
+template <typename T>
+class vari_value<T, std::enable_if_t<is_eigen_sparse_base<T>::value>>
```
bbbales2, line 376: Is there a `require` for this?

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+   * @return Pointer to allocated bytes.
+   */
+  static inline void* operator new(size_t nbytes) {
+    return ChainableStack::instance_->memalloc_.alloc(nbytes);
+  }
+
+  /**
+   * Delete a pointer from the underlying memory pool.
+   *
+   * This no-op implementation enables a subclass to throw
```
bbbales2, line 519: Are these gonna get marked final?

test/unit/math/mix/meta/is_eigen_sparse_base_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+  using stan::math::test::all_eigen_sparse;
+  all_eigen_sparse<false, true, true, true, is_eigen_sparse_base>();
+}
+
+TEST(MathMetaPrim, is_eigen_sparse_base_decomp_tests) {
+  using Eigen::Array;
+  using Eigen::Matrix;
+  using stan::is_eigen_sparse_base;
+  using stan::math::test::all_eigen_dense_decomp;
+  all_eigen_dense_decomp<false, is_eigen_sparse_base>();
```
bbbales2, line 38: I don't get what this test is doing.

test/unit/math/rev/core/var_test.cpp
```diff
@@ -39,26 +42,117 @@ void ctor_overloads_impl() {
...
+}  // namespace test
+}  // namespace stan
+TEST_F(AgradRev, ctorfloatOverloads) {
+  stan::test::ctor_overloads_float<float>();
+  stan::test::ctor_overloads_float<double>();
+  stan::test::ctor_overloads_float<long double>();
+}
+
+TEST_F(AgradRev, ctormatrixOverloads) {
+  using dense_mat = Eigen::Matrix<double, -1, -1>;
```
bbbales2, line 151: VectorXd and RowVectorXd

stan/math/prim/meta/is_plain_type.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <type_traits>
+
+namespace stan {
+/** \ingroup type_trait
+ * Checks whether the template type `T` is an assignable type. This is used
+ * to detect whether a type is an Eigen matrix expression.
+ */
+template <typename S>
+using is_plain_type
+    = std::is_same<std::decay_t<S>, std::decay_t<plain_type_t<S>>>;
```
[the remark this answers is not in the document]
t4c1: `plain_type_t` is never const or reference so that decay is reduundant.

stan/math/rev/core/var.hpp
```diff
@@ -30,15 +30,21 @@ static void grad(Vari* vi);
...
-class var_value {};
-
 template <typename T>
-class var_value<T, require_floating_point_t<T>> {
+class var_value {
+  static_assert(
+      is_plain_type<T>::value,
+      "The template for this var is an"
+      " expression but a var_value's inner type must be assignable such as"
+      " a double, Eigen::Matrix, or Eigen::Array");
```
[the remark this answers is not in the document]
SteveBronder: I think it should be fine to allow for arrays. I can see some cases where we would want an array for instance if we wanted to do element-wise ops.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
 };
 
 // For backwards compatability the default is double
 using vari = vari_value<double>;
 
+/**
+ * The variable implementation for Eigen dense matrix types.
+ *
+ * This class is complete (not abstract) and may be used for
+ * constants.
```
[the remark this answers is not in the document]
SteveBronder: Idk, tbh I quoted the original docs. @bob-carpenter do you know what constants means in this case?

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+ * derivative with respect to the root of the derivative tree.
+ *
+ */
+template <typename T>
+class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
+ public:
+  static_assert(
+      is_plain_type<T>::value,
+      "The template for this var is an"
+      " expression but a var_value's inner type must be assignable such as"
```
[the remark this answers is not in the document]
SteveBronder: What part of this are you referencing? From the specialization with `require_eigen_dense_base_t` we know that `T` is an dense eigen type. If it's a dense eigen type it's either going to be an expression, matrix, or array.

stan/math/rev/core/vari.hpp
SteveBronder: I think both paragraphs can be combined to remove duplicate info

stan/math/rev/core/vari.hpp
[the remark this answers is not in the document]
bbbales2: "The template for this var is an expression" is what I was referencing, but I also think I was wrong and you're right.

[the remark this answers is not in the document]
SteveBronder: With the requires `require_eigen_dense_base_t` we know it's an eigen type. From there it can only be an expression, matrix, or array.

stan/math/rev/core/vari.hpp
```diff
@@ -179,11 +165,375 @@ class vari_value<T, std::enable_if_t<std::is_floating_point<T>::value>>
...
+      " a double, Eigen::Matrix, or Eigen::Array");
+
+  /**
+   * `PlainObject` represents a user constructible type such as Matrix or Array
+   */
+  using PlainObject = std::decay_t<plain_type_t<T>>;
+  using Scalar = PlainObject;  // The underlying type for this class
+  using value_type = Scalar;   // The underlying type for this class
+  using eigen_scalar = value_type_t<PlainObject>;  // A floating point type
```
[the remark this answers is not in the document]
SteveBronder: I'm fine with making it an actual scalar type. @t4c1 made `scalar_type_t` return a `var<double>` for `var<matrix>` so I'm fine with making this a `var<double>`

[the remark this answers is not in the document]
SteveBronder: Or should it be a `double`?

stan/math/rev/core/vari.hpp
SteveBronder: I'm not sure I'm following here. In `grad()` we call `vi->init_dependent();` to set the adjoint to ones. Would should this start with instead?

stan/math/rev/core/vari.hpp
SteveBronder: You mean using a `const vari_value<T>&` instead of a `const vari_value<T>*`?

stan/math/rev/core/vari.hpp
SteveBronder: The sparse impl also inherits from `chainable_alloc`

stan/math/rev/core/vari.hpp
SteveBronder: The sparse impl also inherits from `chainable_alloc`

stan/math/rev/core/vari.hpp
SteveBronder: The `enable_if` just checks if the type `T` inherits from `Eigen::SparseMatrixBase`. We still want to do an assertion that the template type is a plain eigen object

stan/math/rev/core/vari.hpp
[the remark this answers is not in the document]
bbbales2: I think what @t4c1 did makes sense if we want this to be a scalar_type (which maybe we do so some metaprograms work?).

I could also see a matrix-something making sense if we wanted to rename this to something else.

This is one of those issues I'm not gonna be able to understand fully until I write code (to know which thing is useful).

stan/math/rev/core/vari.hpp
SteveBronder: Yep!

stan/math/rev/core/vari.hpp
SteveBronder: `final` is just for virtual functions. Since these are not virtual they can't be marked final

test/unit/math/mix/meta/is_eigen_sparse_base_test.cpp
SteveBronder: It's a bit of a silly check but it just makes sure that `is_eigen_sparse_base` returns `false` for the Eigen dense decomposition types (like `LLT` etc)

stan/math/rev/core/vari.hpp
bbbales2: Here's an example: https://github.com/stan-dev/math/blob/7811368763e51bdc65a263a186e41656a63e6cec/stan/math/rev/functor/cvodes_integrator.hpp#L276

There's a for loop that fills up a bunch of adjoints and then I call `grad()` to do the reverse mode autodiff. `grad()` has been modified to not require a vari, and with no vari it does no `init_dependent` call.

This is a situation where the `init_dependent` isn't obvious, so we don't have it.

stan/math/rev/core/vari.hpp
bbbales2: Yah yah, so the way to print a vari would be:

```
vari a;
std::cout << a << std::endl;
```

instead of:
```
vari a;
std::cout << &a << std::endl;
```

stan/math/rev/core/vari.hpp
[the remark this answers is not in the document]
SteveBronder: Yeah it's a really confusing area because this class can be both a `Scalar` type on it's own or a container. I think I like `Scalar = double`. If you want to query whether it's a container / it's container type you can look at the `value_type` but if you just want to know "Is this a `double, float, long double` floating point type" you can query the `Scalar`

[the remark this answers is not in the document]
bbbales2: Well I suspect the uses of a scalar_type here would be to determine if things are autodiff or not or build a return type. In that case I want the var-ish-ness to come through.

So I'm also a little shaky nailing it down until we have the code that uses this.

stan/math/rev/core/vari.hpp
SteveBronder: > Would should this start with instead?

So for your example you have what the adjoint is being filled in with which is what I'm asking here. 

>  reverse mode differentiation gets us gradients of one thing -- not gradients of a bunch of things [every element in a matrix]). I'd like to just get rid of it if possible.

Isn't the point of the static stuff to take the gradients of every element of a matrix (aka the whole matrix) at once?

stan/math/rev/core/vari.hpp
SteveBronder: Oh, yeah sure!

SteveBronder: Oh ya know what tho. I think we would normally call this like
```
var a(10);
std::cout << a.vi_ << std::endl;
```

So if we make it a ref then we'd have to do
```
var a(10);
std::cout << *a.vi_ << std::endl;
```

stan/math/rev/core/vari.hpp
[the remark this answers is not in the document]
SteveBronder: Should I just get rid of it till we want it?

stan/math/rev/core/vari.hpp
bbbales2: > Isn't the point of the static stuff to take the gradients of every element of a matrix (aka the whole matrix) at once?

Reverse mode autodiff is effectively giving us gradients of a function with one output.

When we have a function that outputs a vector, when we set init_dependent to all ones, we are effectively taking the gradient of that vector function dotted with a vector of ones. It's useful to be able to dot the function with all sorts of vectors -- there's nothing special about the ones (and no reason to use that as the default).

Something like that at least.

bob-carpenter:
> When we have a function that outputs a vector, when we set init_dependent to all ones, we are effectively taking the gradient of that vector function dotted with a vector of ones. It's useful to be able to dot the function with all sorts of vectors -- there's nothing special about the ones (and no reason to use that as the default).

This is much more efficient with forward mode, which you can use to compute a directional derivative (gradient dotted with arbitrary vector).  It's implemented as a Stan functional.  But I still think the solvers are missing forward-mode autodiff.

stan/math/rev/core/vari.hpp
SteveBronder: Okay I think I'm seeing what you mean now. What if we just had a grad function that we passed in a lambda to like

```cpp
template <typename Vari, typename Init>
static void grad(Vari* vi, Init&& init_func) {
  init_func(vi);
  size_t end = ChainableStack::instance_->var_stack_.size();
  size_t beginning = empty_nested() ? 0 : end - nested_size();
  for (size_t i = end; i-- > beginning;) {
    ChainableStack::instance_->var_stack_[i]->chain();
  }
}
```

bbbales2: I am fond of:

```
static void grad(Vari* vi = nullptr) {
  if(vi) {
    vi->init_dependent(); // just make init_dependent throw for vari_value<Eigen> types
  }
  
  ...
}
```

SteveBronder: Why does it need to throw? I'm not a big fan of that style personally I'd rather keep initialization as part of taking the gradient (aka passing the initiliazer as a lambda). Otherwise sometimes the gradient is ready to go here and other times you need to read the code around the call site to get the context on what this function is actually going to do.

Also again what would we do for default initlaization for `vari<Eigen>`. Would each one need it's own initializer? I understand it doesn't need to be set to ones, but it does need to be set to not zero values right?

SteveBronder: I think this is the only thing we are waiting on? imo this sounds like it's optional and we can have a separate PR for initializing with whatever we want in the future

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,204 @@
...
+   * adjoint is initialized to zero.
+   *
+   * All constructed variables are added to the stack. Variables
+   * should be constructed before variables on which they depend
+   * to insure proper partial derivative propagation.  During
+   * derivative propagation, the chain() method of each variable
+   * will be called in the reverse order of construction.
+   *
+   * @tparam S A `matrix_cl` or kernel generator expression type that is
+   * convertible to `value_type`
```
bbbales2, line 48: value_type isn't here anymore

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,204 @@
...
+   *  a stack to keep track of whether the adjoint needs to be set to zero.
+   *
+   * All constructed variables are added to a stack.  Variables
+   *  should be constructed before variables on which they depend
+   *  to insure proper partial derivative propagation.  During
+   *  derivative propagation, the chain() method of each variable
+   *  will be called in the reverse order of construction.
+   *
+   * @tparam S A `matrix_cl` or kernel generator expression type that is
+   * convertible to `value_type`
```
bbbales2, line 92: value_type isn't here anymore

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,204 @@
...
+    ChainableStack::instance_->var_stack_.push_back(this);
+  }
+
+  /**
+   * Construct an matrix_cl variable implementation from a value. The
+   *  adjoint is initialized to zero and if `stacked` is `false` this vari
+   *  will be not be put on the var_stack. Instead it will only be put on
+   *  a stack to keep track of whether the adjoint needs to be set to zero.
+   *
+   * All constructed variables are added to a stack.  Variables
```
bbbales2, line 85: This sentence isn't true if stacked == false.

SteveBronder: I think the specific wording `a stack` here makes this sentence true. Things are either added to the `var_stack_` or `var_nochain_stack_`

SteveBronder:
@bbbales2 as we discussed I added a `grad()` method that just starts at the bottom of the stack and calls each chain without setting the adjoints to anything first. Also added the stuff so that the `grad()` method in `var_value` only exist if the inner type is a scalar

@t4c1 do you want to get the operands_and_partials branch approved/merged in here first before approving/merging this branch to develop? I'm fine either way

SteveBronder:
@serban-nicusor-toptal getting a Jenkins error I haven't seen before

```
+ git pull
Already up to date.
+ git checkout develop
error: pathspec 'develop' did not match any file(s) known to git
script returned exit code 1
```

Do you know what's up with that?

serban-nicusor-toptal:
Hey @SteveBronder everything is back to normal, sorry for the trouble!
Jenkins was only fetching this PR from the origin so I had to [make sure](https://github.com/stan-dev/jenkins-shared-libraries/commit/49a91285cf915a9de692416717cb27f20c6f1738) it checks out all of them.

stan/math/rev/core/grad.hpp
```diff
@@ -37,6 +37,26 @@ static void grad(Vari* vi) {
...
+ * rule is applied working down the stack from the last vari created on the
+ * AD tape and then calling each vari's `chain()` method in turn.
+ *
+ * <p>This function computes a nested gradient only going back as far
+ * as the last nesting.
+ *
+ * <p>This function does not recover any memory from the computation.
+ *
+ */
+static void grad() {
```
bbbales2, line 52: The duplicate code here can be eliminated pretty easily.

Either call this grad inside the other grad or do a default nullptr argument in the other thing:

```cpp
template <typename Vari>
static void grad(Vari* vi) {
  vi->init_dependent();
  grad();
}
```

or

```cpp
template <typename Vari>
static void grad(Vari* vi = nullptr) {
  if(!vi)
    vi->init_dependent();
  size_t end = ChainableStack::instance_->var_stack_.size();
  size_t beginning = empty_nested() ? 0 : end - nested_size();
  for (size_t i = end; i-- > beginning;) {
    ChainableStack::instance_->var_stack_[i]->chain();
  }
}
```

and delete the second definition.

SteveBronder: Done deal!

SteveBronder:
@serban-nicusor-toptal much appreciated!

t4c1:
> @t4c1 do you want to get the operands_and_partials branch approved/merged in here first before approving/merging this branch to develop? I'm fine either way

I think you don't need to wait for me. If this gets ready before ops_partials go ahead and merge it.

stan/math/rev/core/var.hpp
```diff
@@ -76,14 +86,13 @@ class var_value<T, require_floating_point_t<T>> {
    * @param x Value of the variable.
    */
   template <typename S, require_convertible_t<S&, value_type>* = nullptr>
-  var_value(const S& x) : vi_(new vari_type(x, false)) {}  // NOLINT
+  var_value(S&& x) : vi_(new vari_type(x, false)) {}  // NOLINT
```
t4c1, line 89:
```suggestion
  var_value(const S& x) : vi_(new vari_type(x, false)) {}  // NOLINT
```

stan/math/rev/core/vari.hpp
```diff
@@ -198,41 +183,48 @@ using vari = vari_value<double>;
...
    */
-  static constexpr Eigen::Index RowsAtCompileTime = Scalar::RowsAtCompileTime;
+  static constexpr Eigen::Index RowsAtCompileTime
+      = PlainObject::RowsAtCompileTime;
   /**
    * Number of columns known at compile time
    */
-  static constexpr Eigen::Index ColsAtCompileTime = Scalar::ColsAtCompileTime;
+  static constexpr Eigen::Index ColsAtCompileTime
+      = PlainObject::ColsAtCompileTime;
```
t4c1: Eigen uses enums to store these. I like our approach with constexprs better, but I think the type should be int.

stan/math/rev/core/vari.hpp
```diff
@@ -450,14 +407,8 @@ class vari_value<T, std::enable_if_t<is_eigen_sparse_base<T>::value>>
...
-  template <typename S, require_convertible_t<S&, value_type>* = nullptr>
-  explicit vari_value(S&& x)
-      : rows_(x.rows()),
-        cols_(x.cols()),
-        size_(x.size()),
-        val_(x),
-        adj_(x),
-        chainable_alloc() {
+  template <typename S, require_convertible_t<S&, T>* = nullptr>
+  explicit vari_value(S&& x) : val_(x), adj_(x), chainable_alloc() {
```
t4c1, line 411:
```suggestion
  explicit vari_value(const S& x) : val_(x), adj_(x), chainable_alloc() {
```

stan/math/rev/core/vari.hpp
```diff
@@ -478,14 +429,8 @@ class vari_value<T, std::enable_if_t<is_eigen_sparse_base<T>::value>>
...
-  template <typename S, require_convertible_t<S&, value_type>* = nullptr>
-  vari_value(S&& x, bool stacked)
-      : rows_(x.rows()),
-        cols_(x.cols()),
-        size_(x.size()),
-        val_(x),
-        adj_(x),
-        chainable_alloc() {
+  template <typename S, require_convertible_t<S&, T>* = nullptr>
+  vari_value(S&& x, bool stacked) : val_(x), adj_(x), chainable_alloc() {
```
t4c1, line 433:
```suggestion
  vari_value(const S& x, bool stacked) : val_(x), adj_(x), chainable_alloc() {
```

stan/math/rev/core/vari.hpp
andrjohns: I thought that enums were preferred since the ```static constexpr``` doesn't always operate as purely compile-time (based on [this writeup](https://www.informit.com/articles/article.aspx?p=2832416&seqNum=6)):
> Static constant members are lvalues (see Appendix B). So, if we have a declaration such as
```void foo(int const&);```
and we pass it the result of a metaprogram:
```foo(Pow3<7>::value);```
a compiler must pass the address of ```Pow3<7>::value```, and that forces the compiler to instantiate and allocate the definition for the static member. As a result, the computation is no longer limited to a pure “compile-time” effect.
Enumeration values aren’t lvalues (i.e., they don’t have an address). So, when we pass them by reference, no static memory is used. It’s almost exactly as if you passed the computed value as a literal. The first edition of this book therefore preferred the use of enumerator constants for this kind of applications.
C++11, however, introduced constexpr static data members, and those are not limited to integral types. They do not solve the address issue raised above, but in spite of that shortcoming they are now a common way to produce results of metaprograms.

t4c1: I did not know that and I have to say I don't fully understand it. How can enums be passed around without taking memory?

Anyway I am fine with changing this to an enum if that is prefered.

andrjohns: I definitely don't know enough to have preference here, just something I stumbled on recently that I thought was interesting.

SteveBronder: Reading the section I think the only concern is when the `static constexpr` is used by a function taking a reference? imo I like the static constexpr version and idt the performance difference here would be drastic with either/or

stan/math/rev/core/vari.hpp
SteveBronder: Actually should these be forwarding because sparse holds an actual sparse matrix while the dense matrix one should use `const S&` since it always allocated memory?

SteveBronder: I changed them all to be `const S&`

t4c1: I thought sparse vari used maps too. Than you are right and forwarding can be taken advantage of. In that case you also need to call forward on the assignment.

SteveBronder: Done deal!

SteveBronder:
Cool cool so I think this is ready to go!

PR #1966, opened 2020-07-06
Allow binary vectorisation to take combinations of integer vectors

andrjohns:
## Summary

At the moment, if a vectorised binary function is called with a container of ```int``` and container of ```double``` (e.g., ```pow(VectorXd, VectorXi)```), then Eigen will throw an error about mixing different numeric types. To avoid this, we need to add a specialisation to ```ScalarBinaryOpTraits``` for each combination of ```int``` and ```double```/```var```/```fvar```.

This pull adds those specialisations, and vectorises ```bessel_first_kind``` (integer first argument) and ```falling_factorial``` (integer second argument) as test cases.

## Tests

Vectorised testing in both ```prim``` and ```mix``` has been added for the ```bessel_first_kind``` and ```falling_factorial``` functions

## Side Effects

N/A

## Release notes

Extends binary vectorisation framework to take combinations of ```real``` and ```int``` containers as inputs

## Checklist

- [x] Math issue #1890 


test/unit/math/test_ad.hpp
```diff
@@ -1334,6 +1393,27 @@ void expect_ad_vectorized_binary(const ad_tolerances& tols, const F& f,
...
+ * @tparam F type of polymorphic, vectorized functor to test
+ * @tparam T1 type of first argument
+ * @tparam T1 type of second argument
+ * @param tols tolerances for test
+ * @param f functor to test
+ * @param x value to test
+ * @param y value to test
+ */
+template <typename F, typename T1, typename T2,
+          require_any_std_vector_t<T1, T2>* = nullptr>
```
bbbales2, line 1411: Do any of these things need to be std::vectors?

The two things exposed here are:
```
template <typename F, typename T1, typename T2,
          require_all_eigen_col_vector_t<T1, T2>* = nullptr>
void expect_ad_vectorized_binary(const ad_tolerances& tols, const F& f,
                                 const T1& x, const T2& y)
```

```
template <typename F, typename T1, typename T2,
          require_any_std_vector_t<T1, T2>* = nullptr>
void expect_ad_vectorized_binary(const ad_tolerances& tols, const F& f,
                                 const T1& x, const T2& y)
```

Should two row_vectors work?

Edit: I guess this goes to the other question. Do we intend for this interface to work with `row_vectors` at the Stan level? Not saying it has too -- just trying to understand the extent of the vectorization here.

test/unit/math/mix/fun/bessel_first_kind_test.cpp
```diff
@@ -13,3 +13,19 @@ TEST(mathMixScalFun, besselFirstKind) {
...
+TEST(mathMixScalFun, besselFirstKind_vec) {
+  auto f = [](const auto& x1, const auto& x2) {
+    using stan::math::bessel_first_kind;
+    return bessel_first_kind(x1, x2);
+  };
+
+  std::vector<int> std_in1{3, 1};
+  Eigen::VectorXd in2(2);
+  in2 << 0.5, 3.4;
+  stan::test::expect_ad_vectorized_binary(f, std_in1, in2);
```
bbbales2, line 26: Does this also test if in2 is an Eigen row vector or a std::vector type?

A second question is, what signatures are we exposing at the Stan level?

Is it just:

```
vector bessel_first_kind(int [] i, real a);
vector bessel_first_kind(int [] i, vector a);
vector bessel_first_kind(int i, vector a);
```

Or does it also have row vector types?
```
vector bessel_first_kind(int i, row_vector a);
```

Or something.

test/unit/math/test_ad.hpp
andrjohns: > Should two row_vectors work?


Yep, my original goal was to have the same testing that was used in the ```prim``` vectorisation tests, which test:
```
f(vector, vector)
f(row_vector, row_vector)
f(matrix, matrix)
f(array, array)    // Eigen::Array
f(expr, expr)
f(std::vector, std::vector)
```

Then within each of those combinations, it tests having nested arrays of the inputs, and reals with the inputs. But with the ```expect_ad``` framework, the time for testing got pretty long so I settled for testing:
```
f(vector, vector)
f(std::vector, std::vector)
```

And then combinations of nested arrays and reals with those inputs.

> Should two row_vectors work?

Yep, works like a charm

test/unit/math/mix/fun/bessel_first_kind_test.cpp
andrjohns: > Does this also test if in2 is an Eigen row vector or a std::vector type?

Not at the moment, should I add it?

> A second question is, what signatures are we exposing at the Stan level?

I was supposed to update the top-level description, sorry. This will expose the following signatures:
```
real bessel_first_kind(int v, real x)
real[] bessel_first_kind(int[] v, real[] x)
real[] bessel_first_kind(int v, real[] x)
real[] bessel_first_kind(int[] v, real x)
vector bessel_first_kind(int[] v, vector x)
vector bessel_first_kind(int v, vector x)
row_vector bessel_first_kind(int[] v, row_vector x)
row_vector bessel_first_kind(int v, row_vector x)
matrix bessel_first_kind(int[,] v, matrix x)
matrix bessel_first_kind(int v, matrix x)
```

These can also be arbitrarily nested (depending the level of signature we want to expose), as long as both arguments have the same level of nesting or one is a scalar:
```
real[,] bessel_first_kind(int[,] v, real[,] x)
real[,] bessel_first_kind(int v, real[,] x)
real[,] bessel_first_kind(int[,] v, real x)
vector[] bessel_first_kind(int[,] v, vector[] x)
vector[] bessel_first_kind(int v, vector[] x)
row_vector[] bessel_first_kind(int[,] v, row_vector[] x)
row_vector[] bessel_first_kind(int v, row_vector[] x)
matrix[] bessel_first_kind(int[,,] v, matrix[] x)
matrix[] bessel_first_kind(int v, matrix[] x)

```

bbbales2: Oh okay so it was prohibitively slow to have all the combos in the test framework. That's less than ideal but makes sense.

I feel like we should have at least one test per signature we're exposing at the Stan level. Can you do this with prim for these functions? Like you can make these tests pretty easily. Just define an `Eigen::VectorXd`, an `Eigen::RowVectorXd`, a `std::vector` and fill them with the same values and call them with the arguments in the forward order + reversed and just check the values are always the same as some reference (use `EXPECT_MATRIX_FLOAT_EQ`).

Or I guess that could go in the prim binary testing bit? However is fine. I think we should at least have code in there checking values + checking that these things compile.

I'm a little hesitant about going crazy with the nesting -- only need so many signatures here -- I'd be happy with 1D things. This sounds like an extension of some vectorization that's already in. Presumably just match whatever you did for that.

bbbales2: > Yep, my original goal was to have the same testing that was used in the prim vectorisation tests, which test:

Oh wait I just read your comments again. Are these being tested with all these different types in prim already?

andrjohns: Yep that's being tested in ```prim```, but I realised I need to expand the testing with ```std::vector<int>``` inputs, so now it's testing:
```
bessel_first_kind(int, real)
bessel_first_kind(int, vector)
bessel_first_kind(int, row_vector)
bessel_first_kind(int, matrix)
bessel_first_kind(int, vector[])
bessel_first_kind(int, row_vector[])
bessel_first_kind(int, matrix[])
bessel_first_kind(int[], real)
bessel_first_kind(int[], vector)
bessel_first_kind(int[], row_vector)
bessel_first_kind(int[,], matrix)
```

bob-carpenter:
In terms of things being prohibitively slow, if there are vector inputs, keep sizes small, and if there are vector outputs, they can be reduced using something like `sum()`.  That may not be enough---just something to keep in mind.

PR #1967, opened 2020-07-06
Add vari_base<matrix_cl>

t4c1:
## Summary
Adds `matrix_cl` support to `vari_base`. `var_base` containing Eigen matrices can be copied to opencl device resulting in `var<matrix_cl>`. `var<matrix_cl>` can be copied back to host resulting in `var<Eigen::Matrix>`. Adjoints can be propagated over these copies. 

To do that I modified a few template metaprograms and added a simple kg expression for constant.

## Tests
Added tests for constant, vari<matrix_cl> and copiing of vars to and from OpenCL device.

## Side Effects
None.

## Release notes

vari_base can now be constructed with matrix_cl value and adjoint. Copying between host and device is possible for vars containing Eigen::Matrix/matrix_cl. Adjoints are propagated over these copies.

## Checklist


stan/math/opencl/rev/copy.hpp
```diff
@@ -16,6 +16,25 @@
...
+    a_.adj()
+        += from_matrix_cl<T::RowsAtCompileTime, T::ColsAtCompileTime>(adj());
+  }
+};
+
+template <typename T>
+inline var_value<matrix_cl> to_matrix_cl(const var_value<T>& a) {
+  return op_copy_to_cl_vari<>(
+      a.vi_);  // Is this corret type? Only if T is actual type, or ... ?
+}
```
SteveBronder, line 36: Like this right?

```suggestion
template <typename T, require_eigen_t<T>* = nullptr>
inline auto to_matrix_cl(const var_value<T>& a) {
  return var_value<matrix_cl<value_type_t<T>>>(new op_copy_to_cl_vari<T>(
      a.vi_));
}
```

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,184 @@
...
+   * @param adj Adjoint of the constructed variable.
+   */
+  template <typename R, typename S, require_convertible_t<R&, T>* = nullptr,
+            require_convertible_t<S&, T>* = nullptr>
+  vari_value(R&& val, S&& adj)
+      : chainable_alloc(),
+        adj_(std::forward<S>(adj)),
+        val_(std::forward<R>(val)) {
+    ChainableStack::instance_->var_stack_.push_back(this);
+  }
```
SteveBronder, line 79: Should this be in protected or private?

stan/math/opencl/rev/copy.hpp
```diff
@@ -16,6 +16,25 @@
 namespace stan {
 namespace math {
 
+template <typename T, require_vari_vt<is_eigen, T>* = nullptr>
+class op_copy_to_cl_vari : public vari<matrix_cl<value_type_t<T::value_type>>> {
+  T& a_;
+
+ public:
+  op_to_cl_copy_vari(T& a) : vari(to_matrix_cl(a.val())), a_(a) {}
```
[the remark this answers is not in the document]
t4c1: Whoopsie, I meant to put this in next Pr. It is only halfway done. I'll fix it tomorrow.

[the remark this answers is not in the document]
t4c1: Since part of this is already in this PR, I am just putting everything here.

stan/math/opencl/rev/vari.hpp
t4c1: I'll make it protected for now.

SteveBronder:
I'll review this tmrw!

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,190 @@
...
+ * constants.
+ *
+ * A variable implementation is constructed with a constant
+ * value. It also stores the adjoint for storing the partial
+ * derivative with respect to the root of the derivative tree.
+ *
+ */
+template <typename T>
+class vari_value<T, require_kernel_expression_lhs_t<T>>
+    : public vari_base, public chainable_alloc {
```
SteveBronder, line 24: Listing this as very optional, but we could do a map/unmap and use the stack allocators memory like something from here. Though idt we need to worry about that right now

https://software.intel.com/content/www/us/en/develop/articles/getting-the-most-from-opencl-12-how-to-increase-performance-by-minimizing-buffer-copies-on-intel-processor-graphics.html

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,190 @@
...
+  template <typename R, typename S, require_convertible_t<R&, T>* = nullptr,
+            require_convertible_t<S&, T>* = nullptr>
+  vari_value(R&& val, S&& adj)
+      : chainable_alloc(),
+        adj_(std::forward<S>(adj)),
+        val_(std::forward<R>(val)) {
+    ChainableStack::instance_->var_stack_.push_back(this);
+  }
+
+ public:
```
SteveBronder, line 87: Is tehre a reason to flip between public and protected here?

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,190 @@
...
+   *  will be not be put on the var_stack. Instead it will only be put on
+   *  a stack to keep track of whether the adjoint needs to be set to zero.
+   *
+   * All constructed variables are added to a stack.  Variables
+   *  should be constructed before variables on which they depend
+   *  to insure proper partial derivative propagation.  During
+   *  derivative propagation, the chain() method of each variable
+   *  will be called in the reverse order of construction.
+   *
+   * @tparam S A dense Eigen type that is convertible to `value_type`
```
SteveBronder, line 100: This has to be a `matrix_cl` type right? Idt we have methods to implicitly convert from `Eigen` to `matrix_cl`

SteveBronder: I think all the `dense Eigen` just needs to be replaced by `matrix_cl`

We could also have a constructor that takes in an eigen type and does `to_matrix_cl`

stan/math/opencl/rev/copy.hpp
```diff
@@ -16,6 +16,74 @@
 namespace stan {
 namespace math {
 
+namespace internal {
+template <typename T_arg_adj, require_eigen_t<T_arg_adj>* = nullptr>
+class op_copy_to_cl_vari final
+    : public vari_value<matrix_cl<value_type_t<T_arg_adj>>> {
+  T_arg_adj arg_adj_;
```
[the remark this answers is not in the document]
t4c1: Yep, reference is fine. We can't use pass by value here, even if we wanted to, as `Eigen::Map`'s copy constructor copies values, not the object itself.

stan/math/opencl/rev/vari.hpp
t4c1: That is benefitial only if running OpenCL kernels on CPU or integrated GPUs that use host memory anyway. For discrete GPUs copies would happen anyway during the kernel runtime (and they could possibly be finer grained making them slower).

stan/math/opencl/rev/vari.hpp
t4c1: I used to have everything public, but than you wanted to have this constructor protected or privat. I don't understand what is the issue here.

stan/math/opencl/rev/vari.hpp
t4c1: fixed

t4c1:
> we should get rid of matrix_cl<var> at some point in time because I think that will cause confusion between it and this class

I agree. It is actually used in one place, so I'll put it in a separate PR.

stan/math/opencl/rev/vari.hpp
SteveBronder: It's not an issue codewise, it's that style wise flipping between access specifiers (public to protected and then back to public) looks a little funny

t4c1: I don't know in what way do you want me to structure this. Back in kernel generator I groupped things so that everything public, private, etc. was together and you wanted me to group things by what is used together. Following that rule I guessed it made sense to put constructors together and that is not ok either. So what rule should I follow?

stan/math/opencl/rev/vari.hpp
SteveBronder: Yes thats why I'm commenting it as very optional. Could be neat! But not necessary at all

t4c1: Than we would have to detect whether we are running on discrete GPU or not and manage memory differently. This is definitely possible, but I am not sure if it is necessary. I doubt many peopre will run OpenCL on integrated GPUs or CPUs. My guess is that only people who have discrete GPU and want to use it even try using OpenCL. Anyway if we decide to do it, it will definitely happen in its own PR.

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,209 @@
...
+ * This class is complete (not abstract) and may be used for
+ * constants.
+ *
+ * A variable implementation is constructed with a constant
+ * value. It also stores the adjoint for storing the partial
+ * derivative with respect to the root of the derivative tree.
+ *
+ */
+template <typename T>
+class vari_value<T, require_kernel_expression_lhs_t<T>>
```
SteveBronder, line 23: Could make this final but nbd

stan/math/opencl/rev/vari.hpp
SteveBronder: Apologies if I'm being confusing. It's a non-issue but all I mean is right now you have

```cpp
class foo {
public:
 // public stuff
protected:
 // protected stuff
public:
 // more public stuff
}
```

Is there a reason this can't be just one public like this?

```cpp
class foo {
public:
 // public stuff
protected:
 // protected stuff
}
```

Has this shown up previously and I said to do something like multiple access specifiers? My b if so. I'm pretty fine with how it is now but I think it's easier to read if there is just one of each access specifier.

stan/math/opencl/rev/vari.hpp
```diff
@@ -0,0 +1,209 @@
...
+   * @param rows number of rows in the block
+   * @param cols number of columns in the block
+   * @return block
+   */
+  auto block(int row, int col, int rows, int cols) {
+    auto&& val_block = stan::math::block(val_, row, col, rows, cols);
+    auto&& adj_block = stan::math::block(adj_, row, col, rows, cols);
+    return vari_value<std::decay_t<decltype(val_block)>>(std::move(val_block),
+                                                         std::move(adj_block));
+  }
```
SteveBronder, line 148: [Q] This is just something that I'm not sure about

When we do views like blocks etc do they need to be put on the `var_stack_`? idt we will ever do `.chain()` on this vari will we? Since this makes a copy of the data we want to be able to zero it's adjoints, but I think that means this can just go on the `var_nochain_stack_`. So we could use something like 

```cpp
 // false to put on nochain stack
vari_value<std::decay_t<decltype(val_block)>>(std::move(val_block),
                                                         std::move(adj_block), false);
```

stan/math/opencl/rev/vari.hpp
t4c1: Fine, I will change it.

> Has this shown up previously and I said to do something like multiple access specifiers?
https://github.com/stan-dev/math/pull/1343#discussion_r331448932

stan/math/opencl/rev/vari.hpp
t4c1: Good point. However we also don't need to ever zero its adjoints. They will be zeroed when the matrix, this is view into, is zeroed.

t4c1:
@SteveBronder Do you want to check this again? I pushed another commit, but for some reason the review was not dismissed.

SteveBronder:
Let me have one more quick peek at this today

stan/math/opencl/rev/copy.hpp
```diff
@@ -16,6 +16,74 @@
...
+    : public vari_value<Eigen::Matrix<value_type_t<T>, Rows, Cols>> {
+  vari_value<T>& a_;
+
+ public:
+  explicit op_copy_from_cl_vari(vari_value<T>& a)
+      : vari_value<Eigen::Matrix<value_type_t<T>, Rows, Cols>>(
+            from_matrix_cl<Rows, Cols>(a.val_)),
+        a_(a) {}
+
+  virtual void chain() { a_.adj_ = a_.adj_ + to_matrix_cl(this->adj_); }
```
SteveBronder, line 66: Do we not have a `+=` operator for `matrix_cl`?

stan/math/opencl/rev/copy.hpp
```diff
@@ -26,19 +94,17 @@ namespace math {
...
-  dst.val() = to_matrix_cl(src.val().eval());
-  dst.adj() = to_matrix_cl(src.adj().eval());
-  return dst;
-} catch (const cl::Error& e) {
-  check_opencl_error("copy var Eigen->(OpenCL)", e);
-  matrix_cl<T> dst(src.rows(), src.cols());
-  return dst;
+template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+inline var_value<matrix_cl<value_type_t<value_type_t<T>>>> to_matrix_cl(
+    T& src) {
```
SteveBronder, line 99: Is this supposed to be a `const T&`?

stan/math/opencl/rev/copy.hpp
```diff
@@ -26,19 +94,17 @@ namespace math {
...
+  // the matrix can go out of scope before chain() is called. So we store a map
+  // to the data
+  var* src_array
+      = ChainableStack::instance_->memalloc_.alloc_array<var>(src.size());
```
SteveBronder, line 103: var's require having their destructor/delete called right? Should this be a `vari**`?

stan/math/opencl/rev/copy.hpp
```diff
@@ -26,19 +94,17 @@ namespace math {
...
+  Eigen::Map<plain_type_t<T>> src_stacked(src_array, src.rows(), src.cols());
+  src_stacked = src;
+  return new internal::op_copy_to_cl_vari<decltype(src_stacked.adj())>(
+      src_stacked.val(), src_stacked.adj());
```
SteveBronder, line 107: Is a pointer to a `vari_value` inherited type implicitly convertible to a `var_value`? Or should this be like below

```suggestion
  return {new internal::op_copy_to_cl_vari<decltype(src_stacked.adj())>(
      src_stacked.val(), src_stacked.adj())};
```

stan/math/opencl/rev/copy.hpp
t4c1: nope

stan/math/opencl/rev/copy.hpp
t4c1: No. AFAIK only vari_value<matrix_cl> needs destructor called.

stan/math/opencl/rev/copy.hpp
t4c1: Yep, it is implicitly convertible. We have an implicit constructor.

stan/math/opencl/rev/copy.hpp
t4c1: I mean it could be vari**, but i see no need for that. Var is essentially of the same size and we can do math on it, which is very conveniant.

stan/math/opencl/rev/copy.hpp
t4c1: We need t be able to update adjoint of `src`. I can see a use case of this accepting rvalues, but than we would need to const cast them to update adjoint. Maybe forwarding reference? Hmm, its getting a bit late here, I will think on this some more on monday.

bob-carpenter:
> `var`'s require having their destructor/delete called right?

No!  A `var` doesn't control any resources, so it's OK to skip the no-op destructor calls.

stan/math/opencl/rev/copy.hpp
t4c1: Added const. I forgot we can go around adj() and directly access .vi_.adj_ to avoid getting const on the adjoint.

PR #1970, opened 2020-07-09
Add support for var_value in operands_and_partials

t4c1:
## Summary
`operands_and_partials` (and `precomputed_gradients_vari`) now supports `var_value` operands.

`scalar_type` now also works with `var_value`. `scalar_type_t<var_value<T>>` is now `var_value<scalar_type_t<T>>`.

Added versions of `std::vector` (by using custom allocator) and `Eigen::Matrix` (by subclassing `Eigen::Map`) that hold dynamically allocated memory on AD stack.

## Tests
Changes in `operands_and_partials` and `precomputed_gradients_vari` are tested.

## Side Effects
`std::vector`s with custom allocators are not supported in many functions in stan math. Signatures of these functions will have to be adapted to accept those if/when we need them. So far I only modified `dims`.

## Release notes
`operands_and_partials` now supports `var_value` operands.

## Checklist

- [x] Math issue #1875 


t4c1:
@SteveBronder this is ready for review.

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -2,11 +2,13 @@
 #define STAN_MATH_REV_CORE_PRECOMPUTED_GRADIENTS_HPP
 
 #include <stan/math/prim/err/check_consistent_sizes.hpp>
+#include <stan/math/prim/err/check_matching_sizes.hpp>
 // #include <stan/math/rev/meta.hpp>
```
SteveBronder, line 6:
```suggestion
```

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +19,119 @@ namespace math {
...
    * @param[in] size Size of operands and gradients
    * @param[in] varis Operand implementations.
    * @param[in] gradients Gradients with respect to operands.
+   * @param container_operands any container operands
+   * @param container_gradients any container gradients
    */
-  precomputed_gradients_vari(double val, size_t size, vari** varis,
-                             double* gradients)
-      : vari(val), size_(size), varis_(varis), gradients_(gradients) {}
+  template <typename COps = std::tuple<>, typename CGrads = std::tuple<>>
```
SteveBronder, line 79: I'd rather call these `ContainerOps` and `ContainerGrads`

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -87,18 +166,33 @@ class precomputed_gradients_vari : public vari {
...
+          typename ContainerGradinets = std::tuple<>>
+inline var precomputed_gradients(Arith value, const VecVar& operands,
+                                 const VecArith& gradients,
+                                 ContainerOperands&& container_operands
+                                 = std::tuple<>(),
+                                 ContainerGradinets&& container_gradients
+                                 = std::tuple<>()) {
+  return {
+      new precomputed_gradients_vari_template<std::decay_t<ContainerOperands>,
+                                              std::decay_t<ContainerGradinets>>(
```
SteveBronder, line 192:
```suggestion
      new precomputed_gradients_vari_template<std::decay_t<ContainerOperands>,
                                              std::decay_t<ContainerGradients>>(
```

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +19,119 @@ namespace math {
...
+   * @param container_gradients any container gradients
    * @throws std::invalid_argument if the sizes of the vectors
    * don't match.
    */
-  template <typename Arith, typename VecVar, typename VecArith>
-  precomputed_gradients_vari(Arith val, VecVar&& vars, VecArith&& gradients)
+  template <typename Arith, typename VecVar, typename VecArith,
+            typename COps = std::tuple<>, typename CGrads = std::tuple<>,
+            require_all_vector_t<VecVar, VecArith>* = nullptr>
+  precomputed_gradients_vari_template(Arith val, const VecVar& vars,
```
SteveBronder, line 118: Should we just call this class `precomputed_gradients_impl`?

stan/math/rev/core/var.hpp
```diff
@@ -295,5 +295,18 @@ class var_value {
...
+ * Template specialization defining the scalar type of
+ * values stored in var_value.
+ *
+ * @tparam T type to check.
+ * @ingroup type_trait
+ */
+template <typename T>
+struct scalar_type<math::var_value<T>> {
+  using type = math::var_value<scalar_type_t<T>>;
+};
```
SteveBronder, line 309: So this is kind of confusing to me. What's the logic on returning back `var_value<double>` for a `var_value<MatrixXd>`?

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +19,119 @@ namespace math {
...
+        (check_size_match("precomputed_gradients_vari", "rows of operands",
+                          std::get<Is>(container_operands_).vi_->rows(),
+                          "rows of gradients",
+                          std::get<Is>(container_gradients_).rows()),
+         check_size_match("precomputed_gradients_vari", "cols of operands",
+                          std::get<Is>(container_operands_).vi_->cols(),
+                          "cols of gradients",
+                          std::get<Is>(container_gradients_).cols()),
+         0)...});
+  }
```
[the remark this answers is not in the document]
t4c1: Its a standard way of doing things in C++. I don't think we should avoid that, just because you are unfamilliar with it. That said, I don't mind your `for_each` approach, so we can replace this, once `for_each` is in.

stan/math/rev/core/var.hpp
t4c1: Well, `scalar_type` always was one element from the container. To capture all the data from present at particular index of `var_value<MatrixXd>` we need to have both value and adjoint. The scalar type containing that is `var_value<double>`. 

For example I would expect indexing of `var_value<MatrixXd>` to return `var_value<double>` once we implement that. Returning just a value or just adjoint would be the confusing thing.

stan/math/rev/core/precomputed_gradients.hpp
t4c1: `_impl`s are usually things, hidden in internal namespace that noone, except implementor needs to know about. This class could be used directly. And it actually differs from `precomputed_gradients_vari` by the fact of being a template.

t4c1:
> Should we have an op_partials_edge for std::vector<var_value<Op>>?

Yeah, definitely!

>I think we should rewrite these a bit. It's weird we have a mix of partial specialization and SFINAE. Though we don't need to do that in this PR

What is wrong with mixing requires in specializations? We do it in many places and it tends to be the easiest way to implement some things.

stan/math/rev/core/precomputed_gradients.hpp
t4c1: fixed

stan/math/rev/core/precomputed_gradients.hpp
t4c1: done

stan/math/rev/core/precomputed_gradients.hpp
t4c1: fixed

SteveBronder:
> What is wrong with mixing requires in specializations? We do it in many places and it tends to be the easiest way to implement some things.

There's nothing wrong with it in terms of code correctness. It just means having to remember more rules when reading different specializations. If we can do something using one rule then that feels simpler to me and should be preferred

t4c1:
When working with classes or structs we can't use overloads, so we have to use specializations. In this case I see no way to avoid requires, as some specializations are dealing with Eigen types.

SteveBronder:
> When working with classes or structs we can't use overloads, so we have to use specializations. In this case I see no way to avoid requires, as some specializations are dealing with Eigen types.

Yes I think we should use all requires and not partial specializations. But I also don't think that needs to happen in this PR

stan/math/rev/fun/dims.hpp
```diff
@@ -0,0 +1,25 @@
...
+  dims(*x.vi_, result);
+}
+template <typename T, typename = require_eigen_t<T>, typename = void>
+inline void dims(const vari_value<T>& x, std::vector<int>& result) {
+  result.push_back(x.rows());
+  result.push_back(x.cols());
+}
+
+}  // namespace math
+}  // namespace stan
```
SteveBronder, line 23: Are these used somewhere?

stan/math/prim/err/check_matching_dims.hpp
```diff
@@ -120,6 +61,41 @@ inline void check_matching_dims(const char* function, const char* name1,
...
+ * @param name2 variable name for the second matrix (for error messages)
+ * @param y2 second matrix to test
+ * @throw <code>std::invalid_argument</code> if the dimensions of the matrices
+ *    do not match
+ */
+template <bool check_compile, typename Mat1, typename Mat2,
+          typename = require_all_eigen_t<Mat1, Mat2>>
+inline void check_matching_dims(const char* function, const char* name1,
+                                const Mat1& y1, const char* name2,
+                                const Mat2& y2) {
```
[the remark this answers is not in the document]
t4c1: A quick search says no. Btw I did not change anything in this overload. It is just moved so it can see and call the one I changed.

stan/math/rev/fun/dims.hpp
t4c1: Yes, in `check_matching_dims`, which is in turn used in `precomputed_gradients_vari_template`.

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +17,118 @@ namespace math {
...
  * directly.
+ *
+ * @tparam ContainerOperands tuple of any container operands (var_value
+ * containing Eigen types)
+ * @tparam ContainerGradients tupleof any container gradients (Eigen types)
  */
-class precomputed_gradients_vari : public vari {
+template <typename ContainerOperands = std::tuple<>,
+          typename ContainerGradients = std::tuple<>>
+class precomputed_gradients_vari_template : public vari {
```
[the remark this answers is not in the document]
t4c1: `precomputed_gradients` is only used where return type is a scalar. So the answer to both of those questions is probably no, unless you intend to extend this to be able to return matrices.

stan/math/prim/err/check_matching_dims.hpp
```diff
@@ -120,6 +61,41 @@ inline void check_matching_dims(const char* function, const char* name1,
...
+  if (check_compile
+      && (static_cast<int>(Mat1::RowsAtCompileTime)
+              != static_cast<int>(Mat2::RowsAtCompileTime)
+          || static_cast<int>(Mat1::ColsAtCompileTime)
+                 != static_cast<int>(Mat2::ColsAtCompileTime))) {
```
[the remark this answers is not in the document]
t4c1: This is again something I did not change.

stan/math/rev/fun/dims.hpp
SteveBronder: Cool should they be put into internal? Else they need doc'd

t4c1: Added docs.

t4c1:
@wds15 I have a threading test failure here. I think this PR changes nothing that would affect reduce_sum. Actually the last commit changed only doxygen (and everything passed before). I don't know the code, but given it is threading that just randomly failed it could be there are some data races. Since you were working on threading a lot, I think you may want to check this out.

wds15:
The failing test complains that it did only run in a single thread, but the expectation was that multiple threads are supposed to get used. Smells for a Jenkins / machine issue to me.

t4c1:
Ok, I will just restart and hope it does not repeat.

SteveBronder:
That Jenkins error might be some weird thing where `-std=c++1y` for `mpicxx.openmpi` doesn't fully implement C++14? We could just put in the for_each thing I have above which should fix that error (at least when I used it in the base branch it passed fine)

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +17,118 @@ namespace math {
...
    * @param[in] gradients Gradients with respect to operands.
+   * @param container_operands any container operands
+   * @param container_gradients any container gradients
    */
-  precomputed_gradients_vari(double val, size_t size, vari** varis,
-                             double* gradients)
-      : vari(val), size_(size), varis_(varis), gradients_(gradients) {}
+  template <typename ContainerOps = std::tuple<>,
+            typename ContainerGrads = std::tuple<>>
+  precomputed_gradients_vari_template(double val, size_t size, vari** varis,
```
bbbales2, line 74: I think the way to go on this is instead of separating things between `vari**`/`double*` arguments and `std::tuples` in one signature, we do one signature that is only `vari**`/`double` (this one can be for all the precomputed gradient vari stuff we have), and then one that does the tuple thing (this can handle all the operands and partials stuff).

If this thing is named something else and is only used inside operands and partials that's fine.

So the signature here would be:
```cpp
template <typename... T>
precomputed_gradients_vari_template(double val, size_t size,
                    std::tuple<vari_value<T>*...> varis,
                    std::tuple<T*...> gradients) {
  ...
}
```

I guess this is unrelated but related, but all the logic we built up for the variadic stuff is gonna have to graduate to tuples as well for this ([save_varis](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/save_varis.hpp), [accumulate_adjoints](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/accumulate_adjoints.hpp) in particular). I don't think we'll want to do a mixed `vari**`/`std::tuple<vari_value<T>*...>` solutions if we can just go all `std::tuple<vari_value<T>*...>`.

t4c1: What you are suggesting would prevent us from using `var_value<Matrix>` and `Matrix<var>` in the same distribution (same `operands_and_partials` instance. 

> . I don't think we'll want to do a mixed vari**/std::tuple<vari_value<T>*...> solutions if we can just go all std::tuple<vari_value<T>*...>

The issue is we can't in general. If we have Eigen or std vector of `var`s, its length is dynamic, so we can't convert it into a tuple (which must have static size).

stan/math/rev/functor/operands_and_partials.hpp
```diff
@@ -130,8 +133,18 @@ class operands_and_partials<Op1, Op2, Op3, Op4, Op5, var> {
     edge5_.dump_operands(&varis[idx += edge4_.size()]);
     edge5_.dump_partials(&partials[idx]);
 
+    auto container_operands = std::tuple_cat(
+        edge1_.container_operands(), edge2_.container_operands(),
+        edge3_.container_operands(), edge4_.container_operands());
+    auto container_partials = std::tuple_cat(
+        edge1_.container_partials(), edge2_.container_partials(),
+        edge3_.container_partials(), edge4_.container_partials());
```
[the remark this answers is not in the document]
t4c1: Ugh, 1-based indexing got me. Fixed

stan/math/rev/core/precomputed_gradients.hpp
bbbales2: Right right, my bad, but we do this in one of the adj_jac_apply variants and it works now (it's [here](https://github.com/stan-dev/math/blob/f1f32af911b15e3ca4f991b85f03caba9d93caf5/stan/math/rev/functor/adj_jac_apply.hpp#L133)).

So input args of like `var, Eigen<var>, var_value<Eigen>` would make a tuple like `vari*, vari**, vari_value<Eigen>*` or something. So we're still using the `vari**`s, they're just buried in the tuple so it's all the same code. And then everything is packed into the the tuples which is nice.

I guess we can't make operands_and_partials variadic without a bunch of other changes because the code that uses this all makes `.edge1_`, `.edge2_`, etc. accesses.

t4c1: That would be doable. I'm not sure we really need to do it. It gets us a bit simpler interface, but probably a bit less simple logic to handle it. In some extreme cases it might even be a bit slower imo.

I was also thinking about redesingning operands_and_partials from scratch, but it is used in more places than I would like to change.

SteveBronder: > I was also thinking about redesingning operands_and_partials from scratch, but it is used in more places than I would like to change.

I've been in this same boat. I've thought about semi-meaningful API changes that would be nice but it would be a good bit of work and without performance speedups I found it hard to justify

bbbales2: Jeez this was 3 days ago apologies for the delay.

Alright I see the argument let me go back through things again.

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +17,118 @@ namespace math {
...
   const size_t size_;
   vari** varis_;
   double* gradients_;
+  static_assert(std::tuple_size<ContainerOperands>::value
+                    == std::tuple_size<ContainerGradients>::value,
+                "precomputed_gradients_vari: ContainerOperands and "
+                "ContainerGradients should have same size!");
+  static constexpr size_t N_containers
+      = std::tuple_size<ContainerOperands>::value;
+  ContainerOperands container_operands_;
```
bbbales2, line 38: If I'm reading [this](https://github.com/stan-dev/math/blob/5060b683bc031f102ae8ade4a451013a5e567f59/stan/math/rev/functor/operands_and_partials.hpp#L221) correctly, then I think this container will end up containing Eigen dynamic types, which is a memory leak (the destructor never gets called on varis).

stan/math/rev/functor/operands_and_partials.hpp
```diff
@@ -130,8 +133,20 @@ class operands_and_partials<Op1, Op2, Op3, Op4, Op5, var> {
     edge5_.dump_operands(&varis[idx += edge4_.size()]);
     edge5_.dump_partials(&partials[idx]);
 
+    auto container_operands = std::tuple_cat(
```
bbbales2, line 136: The empty tuples have been bothering me too. If there's a memory leak, then I suspect this code will get touched in the rewrite.

In that case I like a function `collect_container_operands` that goes and looks in each edge and pulls out the right bits (so all the edges don't need extra functions).

Also I'm Pacific time zone and down to pair program this via Hangouts or whatever. I tried something myself and faceplanted.

t4c1: Hmm, there is 9-hour difference between pacific time and CEST. I guess I can take time for this one afternoon. Btw I prefere Skype if you are using it (I have some issues with VoiPs within browsers on PC, but I can use laptop).

stan/math/rev/core/precomputed_gradients.hpp
t4c1: Right, I need use maps for these.

SteveBronder:
@t4c1 fyi I was worried the merge would be weird since I squashed the `vari_base` branch so I did the merge here after merging the eigen branch this is based off of. Turned out to be fine but didn't want you to have to muck through it if it was bad

stan/math/rev/core/precomputed_gradients.hpp
t4c1: This turned out to be a bit complicated to fix. Using pointers and maps for elements of these tuples would be a bit ugly. Instead I decided to add separate types for eigen matrices and std vectors that have dynamically allocated memory on AD stack, so they are trivially destructible. Std vectors required just required custom allocator. For Eigen matrix I subclassed `Eigen::Map` and added memory management.

I believe these will also be useful in other places.

stan/math/rev/functor/ad_stack_matrix.hpp
```diff
@@ -0,0 +1,111 @@
...
+#define STAN_MATH_REV_FUNCTOR_AD_STACK_MATRIX_HPP
+
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/rev/core.hpp>
+
+namespace stan {
+namespace math {
+
+template <typename MatrixType>
+class AD_stack_matrix : public Eigen::Map<MatrixType> {
```
bbbales2, line 11: Need docs

stan/math/rev/functor/ad_stack_matrix.hpp
```diff
@@ -0,0 +1,111 @@
...
+ public:
+  using Scalar = value_type_t<MatrixType>;
```
bbbales2, line 13: `scalar_type_t`?

stan/math/rev/functor/ad_stack_matrix.hpp
bbbales2: I could also just go for `stack_matrix`.

stan/math/rev/functor/ad_stack_matrix.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Copy constructor.
+   * @param other matrix to copy from
+   */
+  AD_stack_matrix(const AD_stack_matrix<MatrixType>& other)
+      : Eigen::Map<MatrixType>::Map(const_cast<Scalar*>(other.data()),
+                                    other.rows(), other.cols()) {}
+
+  using Eigen::Map<MatrixType>::operator=;
```
bbbales2, line 68: Is this not inherited?

stan/math/rev/functor/ad_stack_matrix.hpp
```diff
@@ -0,0 +1,111 @@
...
+            other.rows(), other.cols()) {
+    *this = other;
+  }
+
+  /**
+   * Copy constructor.
+   * @param other matrix to copy from
+   */
+  AD_stack_matrix(const AD_stack_matrix<MatrixType>& other)
+      : Eigen::Map<MatrixType>::Map(const_cast<Scalar*>(other.data()),
```
bbbales2, line 65: Does MatrixType here need to have a const?

stan/math/rev/functor/ad_stack_matrix.hpp
```diff
@@ -0,0 +1,111 @@
...
+                                      other.rows(), other.cols());
+    return *this;
+  }
+
+  /**
+   * Resizes this object to given number of rows and columns.
+   * @param rows number of rows
+   * @param cols number of columns
+   */
+  void resize(Eigen::Index rows, Eigen::Index cols) {
```
bbbales2, line 87: This won't work for const things

stan/math/rev/functor/operands_and_partials.hpp
```diff
@@ -194,6 +234,35 @@ class ops_partials_edge<double, Op, require_eigen_st<is_var, Op>> {
...
+  const var_value<Op>& operands_;
+
+  void dump_operands(vari** varis) {}
+  void dump_partials(double* partials) {}
+  int size() { return 0; }
+  std::tuple<const var_value<Op>&> container_operands() {
+    return std::forward_as_tuple(operands_);
+  }
+  std::tuple<partials_t&&> container_partials() {
+    return std::forward_as_tuple(std::move(partials_));
```
bbbales2, line 264: Could you walk me through the forwarding/moving here?

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -87,18 +190,29 @@ class precomputed_gradients_vari : public vari {
...
-  return {new precomputed_gradients_vari(value, std::forward<VecVar>(operands),
-                                         std::forward<VecArith>(gradients))};
+template <typename Arith, typename VecVar, typename VecArith,
+          typename... ContainerOperands, typename... ContainerGradients>
+inline var precomputed_gradients(
+    Arith value, const VecVar& operands, const VecArith& gradients,
+    const std::tuple<ContainerOperands...>& container_operands = std::tuple<>(),
+    const std::tuple<ContainerGradients...>& container_gradients
+    = std::tuple<>()) {
+  return {new precomputed_gradients_vari_template<
```
bbbales2, line 212: So everything will be copied to the stack on construction here?

stan/math/rev/core/ad_allocator.hpp
```diff
@@ -0,0 +1,48 @@
...
+namespace math {
+
+/**
+ * std library compatible allocator that uses AD stack.
+ * @tparam T type of scalar
+ */
+template <typename T>
+struct AD_allocator {
+  using value_type = T;
+
```
bbbales2, line 16: According to: https://en.cppreference.com/w/cpp/memory/allocator

It looks like there are some other boilerplate types to include.

test/unit/math/rev/functor/ad_stack_matrix_test.cpp
```diff
@@ -0,0 +1,75 @@
+#include <stan/math.hpp>
```
bbbales2, line 1: Missing resize tests. Also missing some sorta test for `AD_allocator`

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -17,51 +18,115 @@ namespace math {
...
+      double val, size_t size, vari** varis, double* gradients,
+      const std::tuple<ContainerOps...>& container_operands = std::tuple<>(),
+      const std::tuple<ContainerGrads...>& container_gradients = std::tuple<>())
+      : vari(val), size_(size), varis_(varis), gradients_(gradients) {
+    index_apply<N_containers>([&, this](auto... Is) {
+      this->container_operands_
+          = std::make_tuple(to_AD_stack(std::get<Is>(container_operands))...);
+      this->container_gradients_
+          = std::make_tuple(to_AD_stack(std::get<Is>(container_gradients))...);
+    });
```
SteveBronder, line 84: [optional] if we add `for_each` this can be

```suggestion
      : vari(val), size_(size), varis_(varis), gradients_(gradients), 
        container_operands_(for_each([](const auto& x) { return to_AD_stack(x);}, container_operands)),
        container_operands_(for_each([](const auto& x) { return to_AD_stack(x);}, container_gradients)),
```

stan/math/rev/fun/to_ad_stack.hpp
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * For scalars this is a no-op.
+ * @tparam type of scalar
+ * @param a argument
+ * @return argument
+ */
+template <typename T, require_not_container_t<T>* = nullptr>
+T to_AD_stack(T&& a) {
+  return std::forward<T>(a);
+}
```
SteveBronder, line 23: You need `decltype(auto)` for forwarding here right?

```suggestion
template <typename T, require_not_container_t<T>* = nullptr>
decltype(auto) to_AD_stack(T&& a) {
  return std::forward<T>(a);
}
```

stan/math/rev/fun/to_ad_stack.hpp
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Converts eigen types to `AD_stack_matrix`.
+ * @tparam type of argument
+ * @param a argument
+ * @return argument copied/evaluated on AD stack
+ */
+template <typename T, require_eigen_t<T>* = nullptr>
+AD_stack_t<T> to_AD_stack(const T& a) {
+  return AD_stack_t<T>(a);
+}
```
SteveBronder, line 37: [Optional ]

```suggestion
template <typename T, require_eigen_t<T>* = nullptr>
AD_stack_t<T> to_AD_stack(const T& a) {
  return {a};
}
```

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: Does not matter as we never nest containers in Eigen matrices. If we did `value_type_t` would be the correct choice.

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: It is. However, without this compiler preferes to use combination of implicit constructor and copy assignment operator to just assignment operator inherited from `Eigen::Map` when assigned an expression.

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: Memory is always allocated on AD stack, never in adtual static memory, so we can afford to do that. If we did not want to const cast, we could only copy the data. rev types are always passed around as const refs anyway and than used as not const (adjoint needs to be modified in reverse pass).

There might be a better option how to do this, but I don't see it.

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: It does not need to. Const objects are const and cant be resized.

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: Done. Imo just stack_matrix could be mistaken as having something to do with `std::stack` or some other stack.

stan/math/rev/functor/operands_and_partials.hpp
t4c1: I am modifying this a bit. Now partials in edges are already allocated on AD stack, so there is no need to do a deep copy when putting them on AD stack in `precomputed_gradients`. This call to `std::move` here is not necessary anymore so I removed it. 

For both operands and partials `forward_as_tuple` constructs a tuple containing a lvalue reference. Than nothing interesting happens until the constructor of `precomputed_gradients`. In the constructor a call to `to_AD_stack` copies operands to stack. Partials are already on stack so `to_AD_stack` just makes a shallow copy (equivalent to move, except the original object is unchanged).

stan/math/rev/core/precomputed_gradients.hpp
t4c1: I am modifying this a bit. Now it is just everything not already on stack.

stan/math/rev/core/ad_allocator.hpp
t4c1: Your link points to docs for `std::allocator`. Actual requirements for allocaters are here: https://en.cppreference.com/w/cpp/named_req/Allocator
They list everything else as optional.

test/unit/math/rev/functor/ad_stack_matrix_test.cpp
t4c1: Added tests

stan/math/rev/core/precomputed_gradients.hpp
t4c1: Not a major difference in my opinion. Fell free to change it if you decide to add it.

stan/math/rev/fun/to_ad_stack.hpp
t4c1: Nope. That is needed only if you need to compiler to deduce return type. I know my return type is T (which can be a value or a lvalue reference).

I modified this not to use perfect forwarding anyway.

stan/math/rev/fun/to_ad_stack.hpp
t4c1: done

stan/math/rev/fun/to_ad_stack.hpp
SteveBronder: Cool cool, for clarification I thought if you wanted to return an rvalue from a forwarded input you had to use `decltype(auto)` like the below? It looks like `T` makes a copy (though this function will almost surely be inlined anyway)

https://godbolt.org/z/hnTGe8

t4c1: Yeah it is a copy until RVO comes into play (and I think `decltype(auto)` works the same). There is also an option of `T&&`, but that has some other issues.

SteveBronder: icic I figured it might be copy ellision

stan/math/rev/functor/ad_stack_matrix.hpp
bbbales2: Can it be `value_type` or something like that on the left then?

stan/math/rev/functor/ad_stack_matrix.hpp
bbbales2: Oh, so if this wasn't here we'd actually do an unnecessary allocate?

Add a comment so this doesn't get deleted in a cleanup (what you said above is probably fine).

stan/math/rev/functor/ad_stack_matrix.hpp
bbbales2: Oh resize reallocates too. I don't see where I would resize anything though. Am I missing an obvious use here?

Like `stack_matrix` just kinda wraps stuff that's already been allocated and math has been done on it and all.

stan/math/rev/functor/ad_stack_matrix.hpp
SteveBronder: What about instead of using the word `stack` we use the word `arena`? Tho' we use the word `stack` all the time

t4c1: Arena would be fine if it was used anywhere. Is anything wrong with AD_stack in name? I think it is not too long and the meaning is quite clear .

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: Nope. This is also used by Eigen when this type is used in expressions. Every Eigen type has a Scalar typedef.

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: Added a comment. Also tests fail if this gets deleted.

stan/math/rev/functor/ad_stack_matrix.hpp
t4c1: Resize is used by Eigen when a default constructed object is assigned an expression. However, it does not call the derived version, so I had to override the assignment operator to resize before delegating to base operator. So this is not actually needed. Deleted.

stan/math/rev/core/ad_allocator.hpp
```diff
@@ -0,0 +1,48 @@
...
+namespace stan {
+namespace math {
+
+/**
+ * std library compatible allocator that uses AD stack.
+ * @tparam T type of scalar
+ */
+template <typename T>
+struct AD_allocator {
```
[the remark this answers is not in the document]
t4c1: Yeah, it is a fine idea to be consistent with names. I will just go with AD_stack_* everywhere.

stan/math/rev/fun/to_ad_stack.hpp
```diff
@@ -0,0 +1,112 @@
+#ifndef STAN_MATH_REF_FUN_TO_AD_STACK_HPP
+#define STAN_MATH_REF_FUN_TO_AD_STACK_HPP
+
+#include <stan/math/rev/meta.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <vector>
+#include <cstring>
+
+namespace stan {
+namespace math {
```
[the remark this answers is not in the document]
t4c1: added tests

stan/math/rev/fun/to_ad_stack.hpp
```diff
@@ -0,0 +1,112 @@
...
+std::vector<T, AD_allocator<T>> to_AD_stack(
+    const std::vector<T, AD_allocator<T>>& a) {
+  // What we want to do here is the same as moving input into output, except
+  // that we want input to be left unchanged. With any normal allocator that
+  // lead to deallocating memory twice (probably segfaulting). However,
+  // dealocation with `AD_allocator` is a no-op, so we can do that.
+  std::vector<T, AD_allocator<T>> res;
+  std::memcpy(static_cast<void*>(&res), static_cast<const void*>(&a),
+              sizeof(std::vector<T, AD_allocator<T>>));
+  return res;
```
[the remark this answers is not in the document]
t4c1: Exactly. There is no interface in std vector to do this, so I had to do it in a hacky way.

stan/math/rev/functor/ad_stack_matrix.hpp
bbbales2: Then this should be `scalar_type_t` then. I don't think there are plans to nest containers in Eigen matrices.

stan/math/rev/functor/ad_stack_matrix.hpp
bbbales2: The `AD` bothers me a bit. Like it seems like something we'd do with `CamelCase` names.

So seeing the `AD_stack_matrix` kinda makes me want `ad_stack_matrix`. Then the issue is basically everything in `stan::math` could be prefixed with `ad_` (it's an autodiff library, after all) and so that just leaves me with `stack_matrix`.

I'm happy switching to `arena` instead of `stack` to avoid confusion with whatever other stacks (the CPU stack/`std::stack` s, etc.).

In that case we'd have as a collection of things: `arena_matrix`, `to_arena`, `arena_t` and whatever else.

stan/math/rev/meta/ad_stack_type.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace math {
+// forward declaration
+template <typename MatrixType>
+class AD_stack_matrix;
+}  // namespace math
+
+namespace internal {
+template <typename T, typename = void>
+struct AD_stack_type_impl {
+  using type = T;
```
bbbales2, line 18: Should this be an error if it's not a specific type? Or does this handle all types?

test/unit/math/rev/core/ad_stack_allocator_test.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <gtest/gtest.h>
+#include <vector>
+
+void ad_stack_allocator_test() {
+  std::vector<int, stan::math::AD_stack_allocator<int>> v;
+  v.reserve(3);
+  for (int i = 0; i < 5; i++) {
+    v.push_back(1);
+  }
+
```
bbbales2, line 11: Can you sprinkle around in appropriate places tests like:

```EXPECT_TRUE(stan::math::ChainableStack::instance_->memalloc_.in_stack(v.data()));```

Or use these: https://github.com/stan-dev/math/blob/develop/test/unit/math/rev/util.hpp

That'll just make sure we're allocating things in the right place.

**Comment by wds15** [comment]
This is off topic to some extent... still the SIR example runs now a lot slower due to the new templated var_value thing. This is fixed by switching on lto with the compilers as I understood, but apparently this is not turned on by default if I see this right. Can lto be turned on in the makefiles permanently to ensure good performance or is there a reason against that?

If this bears more discussion, I am happy to file this as an issue or kick off a discourse discussion, whatever is more appropriate.

**Comment by t4c1** [comment]
This is a question for @SteveBronder. Also this PR does not touch templating itself. Templating was already merged in https://github.com/stan-dev/math/pull/1952.

File: stan/math/rev/functor/ad_stack_matrix.hpp (hunk 16 above)
Comment (t4c1) [reply stan/math/rev/functor/ad_stack_matrix.hpp:20]: Why? `value_type_t` is both simpler and the one that will still work if we decide to extend this. I see no reason to use `scalar_type_t`.

File: stan/math/rev/functor/ad_stack_matrix.hpp (hunk 15 above)
Comment (t4c1) [reply stan/math/rev/functor/ad_stack_matrix.hpp:18]: OK. I renamed everything to arena_*.

File: stan/math/rev/meta/ad_stack_type.hpp (hunk 30 above)
Comment (t4c1) [reply stan/math/rev/meta/ad_stack_type.hpp:18]: Yeah, I made it an arror to use this on an unsupported type.

File: test/unit/math/rev/core/ad_stack_allocator_test.cpp (hunk 31 above)
Comment (t4c1) [reply test/unit/math/rev/core/ad_stack_allocator_test.cpp:11]: Done. I can't directly use the utils function, as it expect that scalar is var.

**Comment by SteveBronder** [comment]
> This is off topic to some extent... still the SIR example runs now a lot slower due to the new templated var_value thing. This is fixed by switching on lto with the compilers as I understood, but apparently this is not turned on by default if I see this right. Can lto be turned on in the makefiles permanently to ensure good performance or is there a reason against that?

> If this bears more discussion, I am happy to file this as an issue or kick off a discourse discussion, whatever is more appropriate.

The issue on our Jenkins computer is that it uses clang-6, which we can use with `flto` but have to specify the `llvm-ar` in the build. Specifying `gcc-ar` and `llvm-ar` is necessary for `gcc-4.9.3` and `clang-6` but not for `gcc >= 5` and `clang >= 7`. I've tried this and got `flto` working when I log into the mac and linux computers, but I tried setting this up for Jenkins and it couldn't recognize when I setup the `AR` makefile var. 

It may be good to make a separate issue. I think we can support clang-6, but test on clang-7 so that we can just turn on the `-flto` flag on Jenkins and for the perf tests. If we suggest a bunch of non-standard compiler flags it would be good to test with all of them turned on.

clang-6 was released on March 8th 2018 and clang-7 was released 6 months later on September 19th 2018 so I think it's fine/safe to test on 7

**Comment by bbbales2** [comment]
@wds15 not really off topic, but a separate issue is probably good so that @SteveBronder doesn't have to repeat himself more than necessary. Keep harassing us until there is a solution there.

File: stan/math/rev/functor/ad_stack_matrix.hpp (hunk 16 above)
Comment (bbbales2) [reply stan/math/rev/functor/ad_stack_matrix.hpp]: > Why?

I just want scalars to be scalars to be scalars :(, but okay, we'll go with this.

**Comment by wds15** [comment]
Ok...the makefile should ideally switch lto on when running with recent enough compilers and leave it off otherwise ... in full automation. I file an issue for that. Sure.

File: stan/math/rev/fun/to_arena.hpp [hunk 32]
```diff
@@ -0,0 +1,131 @@
...
+ * @param a argument
+ * @return argument
+ */
+template <typename T>
+inline std::vector<T, arena_allocator<T>> to_arena(
+    const std::vector<T, arena_allocator<T>>& a) {
+  // What we want to do here is the same as moving input into output, except
+  // that we want input to be left unchanged. With any normal allocator that
+  // lead to deallocating memory twice (probably segfaulting). However,
+  // dealocation with `AD_stack_allocator` is a no-op, so we can do that.
```
bbbales2, line 46: arena_allocator

stan/math/rev/meta/arena_type.hpp
```diff
@@ -0,0 +1,48 @@
+#ifndef STAN_MATH_REV_META_AD_STACK_TYPE_HPP
```
bbbales2, line 1: ARENA_TYPE

SteveBronder:
> Ok...the makefile should ideally switch lto on when running with recent enough compilers and leave it off otherwise ... in full automation. I file an issue for that. Sure.

I had a bit of trouble with this for clang specifically. `clang-6 -dumpversion` gives 4.2.1, but so does `clang-7`! With clang-8 it returns 8, same with clang-9. But I'm not sure how to tell the diff between 6 and 7 across systems.

t4c1:
@SteveBronder @bbbales2 Can we merge this or do you need to go over the code one more time?

Issue #2013, opened 2020-08-13
Use var_value to make all arguments differentiable in GPU GLMs


t4c1:
## Description
We can resolve the dillemma that GLM arguments could be either on GPU or differentiable. By accepting arguments of type `var_value<matrix_cl>` all arguments can be both differentiable and hold data on GPU. This reduces copying between host and GPU memory and adds GPU GLM signatures that can differentiate `x` and `y` arguments.

To reduce the amount of duplicated code we should remove all GPU GLM signatures that accept either Eigen types or `std::vector`. New implementations will accept only scalars, `matrix_cl`, `var_value<matrix_cl>` and kernel generator expressions. This adds requirement that changes to GLM implementations must be merged together with changes to stanc3 that will copy all arguments to GPU when calling GPU GLMs.

## Example
```
var y;
var_value<matrix_cl> x;
double alpha;
var_value<matrix_cl> beta;
matrix_cl sigma;
var lp = normal_id_glm_lpdf(y, x, alpha, beta, sigma);

```

## Expected Output
GLMs accept `var_value` arguments.

#### Current Version:
v3.3.0

PR #2015, opened 2020-08-14
Use var_value<matrix_cl> in normal_id_glm_lpdf

t4c1:
## Summary

Changes GPU implementation `normal_id_glm_lpdf` to accept `var_value` inputs, allowing diferentiation of `x` and `y`.
New interface still requires `x` and `beta` to be `matrix_cl`s. All other arguments can be either `matrix_cl`s or scalars. (Anything that can be `matrix_cl` can also be `var_value<matrix_cl>`.)

This includs a number of smaller changes:
- Added support for `matrix_cl` to various functions and metaprograms: `is_constant`, `size`, `sum`, `arena_t` (returning a class derived from `matrix_cl` and `chainable_alloc`), `operands_and_partials`, `to_arena`, `dims`
- Reworked `is_matrix_cl` to also return `true` for classes derived from `matrix_cl`
- Some kernel generator expressions now also accept scalars (checks, colwise reductions) and references to scalars (`operator*`)
- A number of files now only include headers they need instead of more general ones
- `elt_divide` now also accepts two scalar arguments
- bugfixed `is_stan_scalar` to return `false` for `var_value<matrix>`
- replaced `+=` in precomputed_gradients with `=` and `+` to make it work with `matrix_cl`
- added `.rows()` and `.cols()` to `var_value` 
- `is_matrix_cl` is moved to prim/meta

## Tests
Modified the tests to test the new interface of `normal_id_glm_lpdf`.

## Side Effects
Interface of GPU `normal_id_glm_lpdf` is completly changed - previous signatures (accepting EIgen types and std vectors) are not accepted anymore.

## Release notes
GPU implementation `normal_id_glm_lpdf` can also be used when derivatives of `x` or `y` are needed.

## Checklist

- [x] Math issue #2013


rok-cesnovar:
Ok, just to clarify. Given that the supported signatures in Stan are

```
real normal_id_glm_lpdf(real, matrix, real, vector, vector)
real normal_id_glm_lpdf(real, matrix, vector, vector, vector)
real normal_id_glm_lpdf(vector, row_vector, real, vector, vector)
real normal_id_glm_lpdf(vector, row_vector, vector, vector, vector)
real normal_id_glm_lpdf(vector, matrix, real, vector, real)
real normal_id_glm_lpdf(vector, matrix, vector, vector, real)
```
this means that any of these signatures can now use the matrix_cl version, regardless whether they are data or parameters? 
x and beta are not scalars in any of these.

rok-cesnovar:
And a follow-up, scalars do not have to be copied with `to_matrix_cl()` right? They are handled by the kernel generator if I recall correctly.

t4c1:
Right. Just if any argument is `matrix_cl`/`var_value<matrix_cl>` all must be scalar/`matrix_cl`/`var_value<matrix_cl>` (no Eigen or std vector arguments)

t4c1:
Also row_vector x is still not supported on GPU.

rok-cesnovar:
So the rule is, 
If x is not a row_vector:
- scalar arg -> leave as is
- else -> to_matrix_cl-it (data is moved in the model constructor, params just before the call)

t4c1:
Correct.

stan/math/opencl/prim/size.hpp
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Returns the size (number of the elements) of a `matrix_cl`.
+ * @param a `matric_cl` to determine size of
+ * @return number of elements in a
+ */
+template <typename T,
+          require_all_kernel_expressions_and_none_scalar_t<T>* = nullptr>
+int size(const T& a) {
+  return a.rows() * a.cols();
+}
```
SteveBronder, line 20: Do we not have a `.size()` member function for kernel expressions? (I'm assuming that's why you are doing `rows() * cols()`

stan/math/opencl/prim/sum.hpp
```diff
@@ -0,0 +1,37 @@
...
+  if (m.rows() == 1) {
+    // using rowwise_sum would run just 1 thread
+    res = colwise_sum(transpose(m));
+  } else if (m.cols() == 1) {
+    res = colwise_sum(m);
+  } else {
+    res = colwise_sum(rowwise_sum(m));
+  }
+  return sum(from_matrix_cl<Eigen::Dynamic, 1>(res));
+}
```
SteveBronder, line 31: [optional] It would be a bit harder to read but rvo can be used here if instead of using `res` there are multiple returns like below

https://godbolt.org/z/4q65zf

SteveBronder: Also is there ever an instance we'd want to keep this single value over on the gpu (i.e. summing stuff on the gpu)? If so then maybe this function should return back a `matrix_cl`. Then at the call site we would do we can do the `from_matrix_cl`

stan/math/opencl/rev/arena_matrix_cl.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace stan {
+namespace math {
+
+/**
+ * A variant of `matrix_cl` that schedules its destructor to be called, so it
+ * can be used on the AD stack.
+ */
+template <typename T>
+class arena_matrix_cl : public chainable_alloc, public matrix_cl<T> {
```
SteveBronder, line 18: Is there a reason this is it's own class and not a specialization of `arena_matrix_cl`? My though process was that if we can have reverse mode functions that work with `arena_matrix` then they could also accept `matrix_cl`. Is there a reason we would want to have explicit versions that need to call this version of `arena_matrix_cl`?

stan/math/opencl/rev/to_arena.hpp
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @tparam T type of scalar
+ * @param a argument
+ * @return argument
+ */
+template <typename T>
+arena_t<matrix_cl<T>> to_arena(const matrix_cl<T>& a) {
+  return arena_t<matrix_cl<T>>(const_cast<cl::Buffer&>(a.buffer()), a.rows(),
+                               a.cols(), a.view());
+}
```
SteveBronder, line 28: The buffer in a matrix_cl is not const so I'm not following why we need a const_cast here. Why not just have a case that makes a copy of the buffer for actual constant `matrix_cl`'s?

stan/math/prim/fun/log.hpp
```diff
@@ -43,7 +43,9 @@ struct log_fun {
  * @return Elementwise application of natural log to the argument.
  */
 template <typename Container,
-          require_not_container_st<std::is_arithmetic, Container>* = nullptr>
+          require_not_container_st<std::is_arithmetic, Container>* = nullptr,
+          require_any_t<is_stan_scalar<Container>,
+                        is_container<Container>>* = nullptr>
```
SteveBronder, line 48: idk if this is a side note or not but should we think of `matrix_cl`'s as containers? Cuz if not then the definition of the `is_container` seems weird

stan/math/prim/meta/is_matrix_cl.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+/** \ingroup type_traits
+ * Checks if the decayed type of T is a matrix_cl.
+ */
+template <typename T>
+struct is_matrix_cl
+    : public std::is_base_of<math::matrix_cl_base, std::decay_t<T>> {};
+
+STAN_ADD_REQUIRE_UNARY(matrix_cl, is_matrix_cl, matrix_cl_group);
+STAN_ADD_REQUIRE_CONTAINER(matrix_cl, is_matrix_cl, matrix_cl_group);
```
SteveBronder, line 26: We should have an `require_opencl_types` group and add it to the meta.hpp like

/**
 * \ingroup require_meta
 * \defgroup require_opencl_types OpenCL
 * `require` type traits to detect types used with OpenCL.
 */

stan/math/prim/meta/is_matrix_cl.hpp
```diff
@@ -0,0 +1,29 @@
...
+class matrix_cl_base {};
+
+}  // namespace math
+
+/** \ingroup type_traits
+ * Checks if the decayed type of T is a matrix_cl.
+ */
+template <typename T>
+struct is_matrix_cl
+    : public std::is_base_of<math::matrix_cl_base, std::decay_t<T>> {};
```
SteveBronder, line 23: What's the logic here on doing a base class versus an impl method like we do for `is_std_vector`?

stan/math/rev/core/precomputed_gradients.hpp
```diff
@@ -165,7 +166,7 @@ class precomputed_gradients_vari_template : public vari {
   template <typename Op, typename Grad,
             require_all_not_std_vector_t<Op, Grad>* = nullptr>
   void chain_one(Op& op, const Grad& grad) {
-    op.vi_->adj_ += this->adj_ * grad;
+    op.vi_->adj_ = op.vi_->adj_ + this->adj_ * grad;
```
SteveBronder, line 169: Is this because we don't have a += operator for matrix_cl? Should we?

stan/math/rev/core/var.hpp
```diff
@@ -306,6 +306,24 @@ class var_value {
     }
     return os << v.val();
   }
+
+  /**
+   * Returns number of rows. Only available if `T` is a matrix.
+   * @return number of rows.
+   */
+  template <typename U = T>
```
SteveBronder, line 314: Do you need a require here?

stan/math/opencl/prim/size.hpp
t4c1: Nope we don't.

stan/math/opencl/prim/sum.hpp
t4c1: Res is not what is returned - it is copied to host and summed before returning.

> Also is there ever an instance we'd want to keep this single value over on the gpu

There probably exist use cases where this would be the optimal thing to do. However, I think what is in this PR is not significantly slower. Also if we want to do the optimal thing it is probably not to sum in a separate kernel, but instead to directly use `res` in the next kernel an sum it as a part of that kernel. That is getting a bit complicated and probably not necessary. So what we have here should be good enough.

stan/math/opencl/rev/arena_matrix_cl.hpp
t4c1: > Is there a reason this is it's own class and not a specialization of arena_matrix_cl?

I don't understand Are you asking why `arena_matrix_cl` is not specialization of `arena_matrix_cl`?

> we can have reverse mode functions that work with arena_matrix then they could also accept matrix_cl

It might be possible to do that for some functions, but `arena_matrix_cl` and `arena_matrix` do not have exactly the same interface.

> Is there a reason we would want to have explicit versions that need to call this version of arena_matrix_cl?

What do you mean by explicit versions and this version of `arena_matrix_cl`? There is no other version of `arena_matrix_cl`.

stan/math/opencl/rev/to_arena.hpp
t4c1: We are accepting a const reference here. So the buffer in it is const. (Accepting a forwarding reference would also accept rvalues without adding the const, but it would not help if the argument this function is called with is actually const).

> Why not just have a case that makes a copy of the buffer for actual constant matrix_cl's?

Because we don't need to and it is inefficient to copy the buffer.

stan/math/prim/fun/log.hpp
t4c1: Probably. Not as a part of this PR tho. I can change this to `require_not_matrix_cl_t`.

stan/math/prim/meta/is_matrix_cl.hpp
t4c1: Added that.

stan/math/prim/meta/is_matrix_cl.hpp
t4c1: This also works for classes derived from `matrix_cl` (`arena_matrix_cl`). I could also add a new case to previous implementation, but this is less code.

stan/math/rev/core/precomputed_gradients.hpp
t4c1: Yes and probably. When adding that we should probably also add it to kg expressions that can function as lvalues.

stan/math/rev/core/var.hpp
t4c1: Not really, but I can add one for clarity.

rok-cesnovar:
Just an FYI: the stanc3 update for GPU GLMs will be done at once for all GLMs so we either merge this to develop and once all are in also update stanc3 or we merge all of it together. Stanc3 side of this is mostly ready so just let me know how you want to do this @SteveBronder @t4c1

stan/math/opencl/prim/normal_id_glm_lpdf.hpp
```diff
@@ -47,97 +52,83 @@ namespace math {
...
-  auto log_sigma_sum_expr = colwise_sum(log(sigma_bc_expr));
+  auto log_sigma_sum_expr = colwise_sum(log(sigma_val));
 
   const int wgs = y_scaled_sq_sum_expr.rows();
 
-  constexpr bool need_mu_derivative = !is_constant_all<T_beta, T_alpha>::value;
+  constexpr bool need_mu_derivative
+      = !is_constant<T_beta>::value
+        || (!is_constant<T_alpha>::value && is_alpha_vector)
+        || (!is_constant<T_y>::value && is_y_vector);
```
SteveBronder, line 124: [side note] It takes up too much brain ram for me when I see all the `!value`'s chained together. We should add something like `is_differentiable` or something like that

stan/math/opencl/prim/normal_id_glm_lpdf.hpp
```diff
@@ -152,41 +143,63 @@ return_type_t<T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
...
-    ops_partials.edge1_.partials_[0]
-        = sum(from_matrix_cl(mu_derivative_sum_cl));
+    mu_derivative_sum = sum(from_matrix_cl(mu_derivative_sum_cl));
   }
-  if (!is_constant_all<T_beta>::value) {
-    const matrix_cl<double> mu_derivative_transpose_cl(
-        mu_derivative_cl.buffer(), 1,
-        mu_derivative_cl.rows());  // transposition of a vector can be done
-    // without copying
+  if (!is_constant_all<T_y>::value) {
```
SteveBronder, line 152:
```suggestion
  if (!is_constant<T_y>::value) {
```

stan/math/opencl/prim/normal_id_glm_lpdf.hpp
```diff
@@ -152,41 +143,63 @@ return_type_t<T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
...
+        = forward_as<matrix_cl<double>&>(ops_partials.edge4_.partials_);
+    matrix_cl<double> edge4_partials_transpose_cl
+        = mu_derivative_transpose_cl * x_val;
+    edge4_partials = matrix_cl<double>(edge4_partials_transpose_cl.buffer(),
+                                       edge4_partials_transpose_cl.cols(), 1);
+    if (beta_val.rows() != 0) {
+      edge4_partials.add_write_event(
+          edge4_partials_transpose_cl.write_events().back());
+    }
+  }
```
SteveBronder, line 188: very spicy

stan/math/opencl/prim/sum.hpp
SteveBronder: imo it's not about efficiency, it's for API consistency since for all the other ops we stay on the gpu. I understand that this would get wrapped in `scalar_<>` anyway, but it feels odd to have one function that returns something not on the gpu

stan/math/opencl/rev/arena_matrix_cl.hpp
SteveBronder: > I don't understand Are you asking why arena_matrix_cl is not specialization of arena_matrix_cl?

Ack sorry massive typo I meant `arena_matrix`. I also misread earlier and see now this is being used in `arena_type_impl`

> It might be possible to do that for some functions, but arena_matrix_cl and arena_matrix do not have exactly the same interface.

I'll resolve this comment for now as that's what I want to look at but is separate.

stan/math/opencl/rev/to_arena.hpp
SteveBronder: Are there times where we would modify this? I'm mostly just worried about UB if we ever write to this buffer and it's actually const (if I'm reading [this](https://stackoverflow.com/a/54504302/2269255) correctly)

stan/math/rev/core/precomputed_gradients.hpp
SteveBronder: Yeah let's do that so that we can keep using the `+=` stuff

stan/math/opencl/prim/sum.hpp
t4c1: I could also argue that the way I implemented it is also for consistency. Namely between overloads of the `sum` function. They all return a scalar. This is actually needed in https://github.com/bstatcomp/math/blob/8116932357a8492b754f91436429d978a6982379/stan/math/prim/functor/broadcast_array.hpp#L32 (this is used in `operands_and_partials7 and can be called with a `matrix_cl`).

stan/math/opencl/rev/to_arena.hpp
t4c1: I don't think we are ever declaring `matrix_cl`s `const`. Also the buffer object (and all other OpenCL objects) only contains an identifier (imagine int or a pointer) of an object allocated by the OpenCL implementation. Constness of the identifier does not affect constness of the actual memory (it can instead be set by flags, such as `CL_MEM_READ_ONLY` ...) So I am pretty sure it would just work even if we did what C++ consideres UB.

stan/math/rev/core/precomputed_gradients.hpp
t4c1: Lets make it a separate PR.

stan/math/opencl/prim/normal_id_glm_lpdf.hpp
t4c1: Just be ready when adding that - this is used in almost every distribution at least 5 times :)

stan/math/opencl/prim/normal_id_glm_lpdf.hpp
t4c1: Done. I don't know why all the distributions are using the `_all` variant for single argument ...

stan/math/opencl/rev/to_arena.hpp
SteveBronder: Why don't we just make the buffer `mutable`? Then it doesn't need to be const cast.

stan/math/opencl/prim/sum.hpp
SteveBronder: lol this function didn't exist when that code went in tho'! For the rest of the API we make it clear when we cross the boundary from OpenCL and back. I'd rather have an overload for that operator when passing stuff back from the GPU. Are there other places we would want to be swapping back and forth implicitly from a `matrix_cl` to an Eigen matrix (or any boundary crossing)?

stan/math/opencl/prim/normal_id_glm_lpdf.hpp
SteveBronder: lol yeah I'll put it on my to do eventually list

test/unit/math/opencl/rev/to_arena_test.cpp
```diff
@@ -0,0 +1,65 @@
+#ifdef STAN_OPENCL
+#include <stan/math/opencl/rev/opencl.hpp>
+#include <test/unit/util.hpp>
+#include <gtest/gtest.h>
+
+TEST(AgradRev, to_arena_matrix_cl_test) {
```
SteveBronder, line 6: Should we have tests that check rvalues and scoped values?

rok-cesnovar:
> Does this have to wait for the changes to the stanc3 compiler?

I would prefer to have all the GLMs modified before opening the stanc3 PR because it will be very annoying to special-case just a few GLMs to var_value. So if we dont want to have a develop where GPU GLMs dont work, we need to have all math PRs done and merge everything together.

t4c1:
I need to move the source branch to stan_dev repo (now it is on fork) so that jenkins will run on other PRs pointing into it.

Issue #2018, opened 2020-08-18
Simplifying current reverse mode functions


bbbales2:
## Description

This pull introduced a new way to write reverse mode autodiff: https://github.com/stan-dev/math/pull/2011

Because this is simple to use, it should be a good way to program all the upcoming `var_value<MatrixXd>` implementations.

Given it'll be used there, it would also be cool to convert all the current reverse mode stuff. This should cut a ton of code. This issue is for tracking efforts to convert the current autodiff (and make sure the new stuff is general enough and fast enough).

#### Current Version:
v3.3.0

PR #2007, opened 2020-08-08
Feature/elementwise checks revert revert

bbbales2:
## Summary

This is the revert of the revert to put it back in the elementwise checks from @peterwicksstringfield from https://github.com/stan-dev/math/pull/1798

We reverted that pull before the release cause of performance things (https://discourse.mc-stan.org/t/cmdstan-2-24-release-candidate-now-available/16818/39).

The trick was the optimization discussed here: https://github.com/stan-dev/math/pull/1798#discussion_r397367183

Eigen must do some tricks to make their `.isFinite()` check fast. For eigen types, we run the fast checks first to see if something isn't finite before running the slower checks that produce the good error messages.

## Tests

There are new tests in here: https://github.com/stan-dev/math/pull/1798, but none on top of that

## Side Effects

Just the side effects of https://github.com/stan-dev/math/pull/1798 (a bugfix for sparse stuff)

## Release notes

New functions is_nonnegative and is_positive_finite to parallel check_nonnegative and check_positive_finite. They signal failure by returning false instead of by throwing std::domain_error.

Functions check_not_nan, check_nonnegative, check_positive, check_finite, check_positive_finite, is_not_nan, is_nonnegative, is_positive, is_scal_finite, and is_positive_finite now operate on nested containers.

Clearer error messages when csr_u_to_z is called with out of range indices.

check_positive now throws domain error when given an unsigned 0.

(Copied those from #1798)

## Checklist

- [x] Math issue #1635 (don't close)

- [x] Copyright holder: Peter Wicks Stringfield

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #2019, opened 2020-08-18
Simplifying existing autodiff to prepare for adding `var_value<Matrix<T, R, C>>` support to rev/fun

bbbales2:
## Summary

This is an attempt to reimplement reverse mode autodiff using `reverse_pass_callback` in preparation for adding `var_value<Matrix<T, R, C>>` to rev/fun.

## Side Effects

It's possible this is slower than the old stuff. Dunno till it's benchmarked.

## Release notes

Re-implemented reverse mode autodiff using `reverse_pass_callback` and new arena tools

## Checklist

- [x] Math issue #2018


stan/math/rev/fun/divide.hpp
```diff
@@ -10,171 +10,41 @@
...
+  double invc = 1.0 / value_of(c);
+
+  using Mat_v
+      = Eigen::Matrix<var, Mat::RowsAtCompileTime, Mat::ColsAtCompileTime>;
+
+  arena_matrix<Mat_v> res = invc * value_of(m);
+
+  reverse_pass_callback([=]() mutable {
+    if (!is_constant<Mat>::value) {
+      accumulate_adjoints(arena_m, invc * res.adj());
```
bbbales2, line 39: arena_m might be an Eigen::MatrixXd or it might be an `arena_matrix<Eigen::Matrix<var, R, C>>` so I gotta have a function here that does different things for the different types.

The way you currently have dot_product is different, and that again is different from the way you originally had dot_product (with values and vars separate and explicitly typed so this wouldn't be necessary).

This code doesn't seem great to me but I dunno which way we should go. Thoughts?

stan/math/rev/fun/tcrossprod.hpp
```diff
@@ -37,18 +37,8 @@ tcrossprod(const T& M) {
...
-  int pos = 0;
   for (int m = 0; m < M.rows(); ++m) {
-    for (int n = 0; n < M.cols(); ++n) {
-      vs[pos++] = M(m, n).vi_;
-    }
-  }
-  for (int m = 0; m < M.rows(); ++m) {
-    MMt.coeffRef(m, m)
-        = var(new internal::dot_self_vari(vs + m * M.cols(), M.cols()));
+    MMt.coeffRef(m, m) = dot_self(M.row(m));
```
bbbales2, line 41: I'm a bit concerned about these functions that use dot_product internally. I think we should just rewrite them to not lean on other functions.

I think with the rewrite we're saving to the autodiff stack O(N^3) varis when really we only need to save O(N^2) (if the input is an N*N matrix).

t4c1: > I think with the rewrite we're saving to the autodiff stack O(N^3) varis when really we only need to save O(N^2) (if the input is an N*N matrix).

That is a good point. Although I think we can still use prim dot_product.

stan/math/rev/fun/divide.hpp
```diff
@@ -10,171 +10,41 @@
...
-      m.rows(), m.cols());
-  result.vi()
-      = Eigen::Map<matrix_vi>(baseVari->adjResultRef_, m.rows(), m.cols());
-  return result;
+template <typename Mat, typename Scal, require_eigen_t<Mat>* = nullptr,
+          require_stan_scalar_t<Scal>* = nullptr,
+          require_any_st_var<Mat, Scal>* = nullptr>
+inline Eigen::Matrix<var, Mat::RowsAtCompileTime, Mat::ColsAtCompileTime>
+divide(const Mat& m, Scal c) {
+  auto arena_m = to_arena_if<!is_constant<Mat>::value>(m);
```
t4c1, line 28: I just figured this could cause a memory leak. If `m` contains `double`s, `to_arena_if` returns  a reference to `m`. `arena_m` is captured by the lambda making a copy (it is used in lambda, so it is captured even if that if is not executed. It might be optimized by compiler, but I would not rely on compiler to eliminate potential memory leak). That last copy's destructor will never be called.

I guess that was poor design on my part, as I did the same thing in `dot_product`.

stan/math/rev/fun/divide.hpp
t4c1: I am solving these kind of problems using `forward_as`. It allows you to specify which type you are expecting in a particular if with a type-dependent condition. Its return type will be what you specified, so the code can always compile. It should only be used in such ifs, so that it is actually called only with a compatible type.

```suggestion
      forward_as<Mat_v>(arena_m).adj() += invc * res.adj();
```

stan/math/rev/fun/gp_exp_quad_cov.hpp
```diff
@@ -208,82 +27,83 @@ class gp_exp_quad_cov_vari<T_x, double, T_l> : public vari {
...
-inline Eigen::Matrix<var, -1, -1> gp_exp_quad_cov(const std::vector<T_x> &x,
-                                                  const var &sigma,
-                                                  const var &length_scale) {
-  check_positive("gp_exp_quad_cov", "sigma", sigma);
-  check_positive("gp_exp_quad_cov", "length_scale", length_scale);
-  size_t x_size = x.size();
-  for (size_t i = 0; i < x_size; ++i) {
+template <typename T_x, typename T_sigma, typename T_l,
+          require_stan_scalar_t<T_sigma>* = nullptr,
+          require_stan_scalar_t<T_l>* = nullptr,
```
t4c1, line 32:
```suggestion
          require_all_stan_scalar_t<T_x, T_sigma, T_l>* = nullptr,
```

stan/math/rev/fun/gp_exp_quad_cov.hpp
```diff
@@ -208,82 +27,83 @@ class gp_exp_quad_cov_vari<T_x, double, T_l> : public vari {
...
+          require_any_st_var<T_x, T_sigma, T_l>* = nullptr>
+inline Eigen::Matrix<var, -1, -1> gp_exp_quad_cov(const std::vector<T_x>& x,
+                                                  T_sigma sigma,
+                                                  T_l length_scale) {
+  check_positive("gp_exp_quad_cov", "marginal standard deviation", sigma);
+  check_positive("gp_exp_quad_cov", "length scale", length_scale);
+  for (size_t i = 0; i < x.size(); ++i) {
     check_not_nan("gp_exp_quad_cov", "x", x[i]);
   }
```
t4c1, line 41:
```suggestion
  check_not_nan("gp_exp_quad_cov", "x", x);
```

stan/math/rev/fun/gp_exp_quad_cov.hpp
```diff
@@ -208,82 +27,83 @@ class gp_exp_quad_cov_vari<T_x, double, T_l> : public vari {
...
-    return cov;
+  double inv_half_sq_l_d
+      = 0.5 / (value_of(length_scale) * value_of(length_scale));
+  for (size_t j = 0; j < x.size(); ++j) {
+    for (size_t i = 0; i < j; ++i) {
+      double dist_sq = squared_distance(value_of(x[i]), value_of(x[j]));
+      dist(i, j) = dist_sq;
+      res_val(i, j) = sigma_sq_d * std::exp(-dist_sq * inv_half_sq_l_d);
+      res_val(j, i) = res_val(i, j);
+      dist(j, i) = dist(i, j);
```
t4c1, line 64:
```suggestion
      dist.coeffRef(j, i) = dist.coeffRef(i, j) = dist_sq;
      res_val.coeffRef(j, i) = res_val.coeffRef(i, j) = sigma_sq_d * std::exp(-dist_sq * inv_half_sq_l_d);
```

stan/math/rev/fun/divide.hpp
t4c1: I think we need another variant of `to_arena_if` that will return some trivially destructible type (for example empty Map) when the condition is false. Not sure how to name them not to cause confusion, as we also need this version in some cases.

stan/math/rev/fun/tcrossprod.hpp
bbbales2: I will try to remember to just rewrite the autodiff for these. It shouldn't be hard to just do the whole thing.

stan/math/rev/fun/gp_exp_quad_cov.hpp
bbbales2: T_x can be an Eigen type, so I'm just changing it for the other two.

stan/math/rev/fun/gp_exp_quad_cov.hpp
bbbales2: This pull got reverted. It's waiting over here to get back in: https://github.com/stan-dev/math/pull/2007

stan/math/rev/core/accumulate_adjoints.hpp
```diff
@@ -147,6 +147,20 @@ inline double* accumulate_adjoints(double* dest, Arith&& x, Pargs&&... args) {
...
+}
+
+template <typename T1, typename T2, require_st_var<T1>* = nullptr>
+inline void accumulate_adjoints(T1& sum, const T2& summand) {
+  sum.adj() += summand;
+}
+
+template <typename T1, typename T2, require_not_st_var<T1>* = nullptr>
+inline void accumulate_adjoints(T1& sum, const T2& summand) {}
+
```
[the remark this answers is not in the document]
bbbales2: Removed. forward_as is good I think.

stan/math/rev/fun/divide.hpp
bbbales2: Oh yeah, so things copied into the lambda would never have their destructors called? Like the destructor on the vari that holds the lambda is never called -- what would happen if I allocated a std::vector inside a lambda passed to a callback function? Is this what you're getting at?

stan/math/prim/meta/forward_as.hpp
```diff
@@ -31,7 +31,8 @@ constexpr bool eigen_static_size_match(T1 desired, T2 actual) {
  */
 template <typename T_desired, typename T_actual,
           typename
-          = std::enable_if_t<std::is_convertible<T_actual, T_desired>::value
+          = std::enable_if_t<std::is_same<std::decay_t<T_actual>,
```
bbbales2, line 34: To make this work I had to change the is_convertible logic to is_same:

```
double x = 1.0;
forward_as<var>(x).adj() += 2.0;
```

double can convert to var, so in the old logic `decltype(forward_as<var>(1.0))` is a double. Also I didn't see tests for `forward_as`. There should be tests, but I can add those after we decide what `forward_as` should do for sure.

stan/math/rev/core/var.hpp
```diff
@@ -102,24 +102,14 @@ class var_value {
...
-  /**
-   * Return a reference to the derivative of the root expression with
-   * respect to this expression.  This method only works
-   * after one of the `grad()` methods has been
-   * called.
-   *
-   * @return Adjoint for this variable.
-   */
-  inline auto& adj() { return vi_->adj_; }
+  inline auto& adj() const { return vi_->adj_; }
```
bbbales2, line 112: We never really respected const for the adjoint before.

I needed to make this change so something like this would work:
```cpp
myfunc(const var& a) {
  ...
  reverse_pass_callback([=]() {
    ...
    a.adj() += whatever;
  });
}
}
```

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -19,12 +20,49 @@ template <typename Mat1, typename Mat2,
...
+  check_matching_sizes("columns_dot_product", "v1", v1, "v2", v2);
+
+  ref_type_t<Mat1> v1_ref = v1;
+  ref_type_t<Mat2> v2_ref = v2;
+  
+  using Mat1_double = plain_type_t<decltype(value_of(v1_ref))>;
+  using Mat1_var = promote_scalar_t<var, Mat1_double>;
+
+  using Mat2_double = plain_type_t<decltype(value_of(v2_ref))>;
+  using Mat2_var = promote_scalar_t<var, Mat2_double>;
```
bbbales2, line 32: Please suggest simplifications to all this type logic lol.

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -19,12 +20,49 @@ template <typename Mat1, typename Mat2,
...
-  return ret;
+
+  Eigen::RowVectorXd out_val(arena_v1_val.cols());
+  for(size_t m = 0; m < arena_v1_val.cols(); ++m)
+    out_val.coeffRef(m) = arena_v1_val.col(m).dot(arena_v2_val.col(m));
+
+  arena_matrix<Eigen::Matrix<var, 1, Eigen::Dynamic>> out = out_val;
+
+  reverse_pass_callback([out, arena_v1, arena_v2,
+			 arena_v1_val, arena_v2_val]() mutable {
```
bbbales2: Maybe as a project policy we should list the variables by name we're copying in. It will be too easy to miss stuff otherwise (and copy something inappropriate).

stan/math/rev/fun/gp_exp_quad_cov.hpp
```diff
@@ -208,82 +29,72 @@ class gp_exp_quad_cov_vari<T_x, double, T_l> : public vari {
...
+  arena_matrix<Eigen::MatrixXd> dist(x.size(), x.size());
+  arena_matrix<Eigen::MatrixXd> res_val(x.size(), x.size());
 
-  size_t pos = 0;
-  for (size_t j = 0; j < x_size - 1; ++j) {
-    for (size_t i = (j + 1); i < x_size; ++i) {
-      cov.coeffRef(i, j).vi_ = baseVari->cov_lower_[pos];
-      cov.coeffRef(j, i).vi_ = cov.coeffRef(i, j).vi_;
-      ++pos;
+  auto arena_x = to_arena_if<!is_constant<T_x>::value>(x);
```
bbbales2, line 55: I now think this might be a bug. This is what you were saying in another comment right?

If this returns false, arena_x is simply x. arena_x gets used in the reverse pass callback (though maybe that branch of code doesn't execute) and therefore might get copied in.

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -19,12 +20,49 @@ template <typename Mat1, typename Mat2,
...
+  ref_type_t<Mat1> v1_ref = v1;
+  ref_type_t<Mat2> v2_ref = v2;
+  
+  using Mat1_double = plain_type_t<decltype(value_of(v1_ref))>;
+  using Mat1_var = promote_scalar_t<var, Mat1_double>;
+
+  using Mat2_double = plain_type_t<decltype(value_of(v2_ref))>;
+  using Mat2_var = promote_scalar_t<var, Mat2_double>;
+
+  arena_matrix<Mat1_double> arena_v1_val = value_of(v1_ref);
```
bbbales2, line 34: I'm still not totally happy with these value_ofs. There will be one unnecessary copy if one input is autodiff and one is double.

stan/math/rev/fun/divide.hpp
t4c1: >Oh yeah, so things copied into the lambda would never have their destructors called? Like the destructor on the vari that holds the lambda is never called

Yeah.

> what would happen if I allocated a std::vector inside a lambda passed to a callback function?

Allocating stuff in the lambda is fine. Just capturing things that allocate stuff is a problem.

stan/math/rev/fun/gp_exp_quad_cov.hpp
t4c1: > I now think this might be a bug. This is what you were saying in another comment right?

Yeah, that is the same thing.

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -19,12 +20,49 @@ template <typename Mat1, typename Mat2,
...
-  check_matching_sizes("dot_product", "v1", v1, "v2", v2);
-  Eigen::Matrix<var, 1, Mat1::ColsAtCompileTime> ret(1, v1.cols());
-  for (size_type j = 0; j < v1.cols(); ++j) {
-    ret.coeffRef(j) = dot_product(v1.col(j), v2.col(j));
+  check_matching_sizes("columns_dot_product", "v1", v1, "v2", v2);
+
+  ref_type_t<Mat1> v1_ref = v1;
+  ref_type_t<Mat2> v2_ref = v2;
+
+  using Mat1_double = plain_type_t<decltype(value_of(v1_ref))>;
```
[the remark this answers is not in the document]
bbbales2: Maybe we should rename it to `set_scalar_t` or `swap_scalar_t` or something.

stan/math/rev/fun/columns_dot_product.hpp
bbbales2: I don't want to depend on value_of being an expression though. If someone changed the definition of value_of to be not-an-expression this would break.

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -19,12 +20,49 @@ template <typename Mat1, typename Mat2,
...
-  return ret;
+
+  Eigen::RowVectorXd out_val(arena_v1_val.cols());
+  for (size_t m = 0; m < arena_v1_val.cols(); ++m)
+    out_val.coeffRef(m) = arena_v1_val.col(m).dot(arena_v2_val.col(m));
+
+  arena_matrix<Eigen::Matrix<var, 1, Eigen::Dynamic>> out = out_val;
+
+  reverse_pass_callback(
+      [out, arena_v1, arena_v2, arena_v1_val, arena_v2_val]() mutable {
```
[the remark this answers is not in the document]
bbbales2: I'll start doing this then.

stan/math/rev/fun/tcrossprod.hpp
```diff
@@ -27,35 +27,20 @@ tcrossprod(const T& M) {
...
+
+  arena_matrix<promote_scalar_t<var, T>> arena_M = M;
+  arena_matrix<promote_scalar_t<double, T>> arena_M_val = value_of(arena_M);
+
+  Eigen::MatrixXd res_val(M.rows(), M.rows());
+  arena_matrix<promote_scalar_t<var, T>> res
+      = arena_M_val * arena_M_val.transpose();
+
+  reverse_pass_callback([res, arena_M, arena_M_val]() mutable {
+    ref_type_t<decltype(res.adj())> adj = res.adj();
```
bbbales2, line 39: How should we be future-proofing these things for `var_value<MatrixXd>` (when getting the adjoint out of a var/matrix thing is easy)? Does this look right?

stan/math/rev/fun/inverse.hpp
```diff
@@ -75,16 +21,23 @@ class inverse_vari : public vari {
...
-  res.vi() = Eigen::Map<matrix_vi>(baseVari->vari_ref_A_inv_, res.rows(),
-                                   res.cols());
+  arena_matrix<Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>> arena_m = m;
+  arena_matrix<Eigen::MatrixXd> res_val = value_of(arena_m).inverse();
+  arena_matrix<Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>> res
+      = res_val;
+
+  reverse_pass_callback([res, res_val, arena_m]() mutable {
+    Eigen::MatrixXd res_adj = res.adj();
+    arena_m.adj() -= res_val.transpose() * res_adj * res_val.transpose();
```
bbbales2, line 38: This code didn't compile and I don't know why:

```
arena_m.adj() -= res_val.transpose() * res.adj() * res_val.transpose();
```

andrjohns: When calling ```.val()```/```.adj()``` on a non-const Eigen matrix the returned type is a ```CWiseUnaryView```, and there's currently a bug with multiplying two ```CWiseUnaryView``` objects ([reported here](https://gitlab.com/libeigen/eigen/-/issues/1730)). I think that might be the source of your problems

stan/math/rev/core/var.hpp
bbbales2: @SteveBronder so yah know

stan/math/rev/fun/gp_periodic_cov.hpp
```diff
@@ -300,95 +38,106 @@ class gp_periodic_cov_vari<T_x, double, T_l, T_p> : public vari {
...
+  double pi_over_p = pi() / p_d;
+  double negative_two_over_l_squared = -2.0 / (l_d * l_d);
+
+  arena_matrix<Eigen::MatrixXd> arena_dist(x.size(), x.size());
+  arena_matrix<Eigen::MatrixXd> arena_sin_squared(x.size(), x.size());
+  arena_matrix<Eigen::MatrixXd> arena_sin_squared_derivative(x.size(),
+                                                             x.size());
+  arena_matrix<Eigen::MatrixXd> res_val(x.size(), x.size());
+
+  auto arena_x = to_arena_if<!is_constant<T_x>::value>(x);
```
bbbales2, line 74: Another bug here.

stan/math/rev/fun/inverse.hpp
bbbales2: Cool beans, I'll just leave it as is then, thanks

stan/math/rev/fun/divide.hpp
bbbales2: I made an issue for this: https://github.com/stan-dev/math/issues/2025 .

stan/math/rev/fun/gp_periodic_cov.hpp
SteveBronder: Should we put these bug fixes in another PR?

stan/math/rev/fun/divide.hpp
SteveBronder: Wonder if it's worth me spending a day figuring out how to use the clang address sanitizer stuff to see if we can catch these things in tests

stan/math/rev/fun/gp_periodic_cov.hpp
bbbales2: Yeah, there's an issue for it here: https://github.com/stan-dev/math/issues/2025

I just made a comment here to remind me to fix this in the future (presumably whatever happens for #2025 will be suitable for this, or at least reveal a solution).

stan/math/rev/core/var.hpp
SteveBronder: That feels weird to me. It's happening because you are taking in a `const &` so it calls the `const` version that gives back... a `const &`! imo the answer is to not use `const var&` when you need to modify the adjoint like this. Or maybe we should make `adj_` `mutable` in `vari_value? We do that for the read/write event stacks in `matrix_cl<>`.

stan/math/rev/fun/columns_dot_product.hpp
t4c1: That is a good point. Also it makes `to_arena_if` useless as it is implemented now. I think we can change it so it returns something that never allocates dynmaic memory, for example an empty `Map` if the condition is false. I will open a PR with this.

stan/math/rev/fun/columns_dot_product.hpp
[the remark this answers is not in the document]
t4c1: I like `set_scalar_t` better. Or maybe `change_scalar_t`.

stan/math/rev/fun/tcrossprod.hpp
t4c1: I would only do that if `adj` is used more than once. Otherwise leaving it an expression by making its type `const auto&` or even not making it a separate variable is preferable.

Also, when you do want a ref this is simpler: `const auto& adj = to_ref(res.adj());`

stan/math/rev/fun/columns_dot_product.hpp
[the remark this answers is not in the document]
bbbales2: This should go in a separate issue/pull, but `set_scalar_t` seems fine to me. It looks like this isn't used in many places yet so it probably won't be that hard to change it.

stan/math/rev/core/var.hpp
bbbales2: > is to not use const var& when you need to modify the adjoint like this

Yeah that's why I was giving it a shoutout cause it's going to look weird on review, but I think we practically make this assumption everywhere.

How this worked with the old stuff is that we just copied vari pointers out of everything. Those pointers themselves aren't constant so we were free to modify the stuff they contain.

bbbales2:
@SteveBronder the reference timing for the 1024x1024 matrix in this case is about 16ms for me.

The stuff from a couple days ago I think you would have been working off of (https://github.com/stan-dev/math/pull/2019/commits/912d437ea158ec731400671aaa604f0af93847d2) was about 19ms for the same calculation for me.

I pushed some more changes to how the calculations are done here: https://github.com/stan-dev/math/pull/2019/commits/1e09d8b30a52d2e6d5c7e0e8bfe6a5d2ea59f1cc

With that in place things go down to 17ms.

To get the last speed bump (and match the last release at about 16ms) we'd need to revert this: https://github.com/stan-dev/math/commit/75bbe3f979b1c2e7b33f000b4a8fe14d8391f00a which are the elementwise checks that I thought I fixed in https://github.com/stan-dev/math/pull/2007.

I'm not keen to revert the elementwise checks again but apparently they need work.

SteveBronder:
Aight ill try to take a look at the element wise checks again to see if we can do anything there.  If its just shaving off a few ms there could be a few things here and there we can get

bbbales2:
I'm wondering if it's the checks in here: https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/squared_distance.hpp

Since we use squared distance internally in that function we'll be checking stuff too much (O(N^2) checks instead of O(N)). But it was fast before with these in place so presumably it can be fast again.

bbbales2:
Here's with the updated gps:

bench_aug23

SteveBronder:
Looking at the two graphs it may be useful to us `--benchmark-repetitions=30` or so (for instance divide looks v different for both graphs but that's probably just in some bounds

https://github.com/google/benchmark#statistics-reporting-the-mean-median-and-standard-deviation-of-repeated-benchmarks

SteveBronder:
Though periodic I think those graphs are fine (it's essentially 1 to 1 after 16 N) but gp_exp_quad_cov. Worst case we could have a specialization for that one case that uses the old version

t4c1:
I agree with `--benchmark-repetitions=30`. also it would be better to benchmark this against develop, not last release.

bbbales2:
I'll try adding repetitions and see if it gets rid of the wiggles. I think they're pretty stable though. Is there a way to get the standard deviations for 1 repetition (cuz the number reported is the mean of something anyway)?

> also it would be better to benchmark this against develop, not last release

Benching against develop tells us how fast `var_value<T>` is against 3.3 vars, which is the thing I want to know. I'm not too concerned with overhead in the `arena`/`callback` stuff at this point (which I think is what comparing against develop would tell us).

@t4c1 can you do the `arena` + `var_value<Eigen>` stuff or do you want me to take a swing at it here (or is it already there and I didn't notice)? For the functions here I want to get that in so I can add a second line to this graph (current `var_value<Eigen>` vs. 3.3.0 `Matrix<var, R, C>`)

bbbales2:
> I think they're pretty stable though

I guess I said that with too much confidence -- what I meant was the gp stuff seemed pretty stable, so I'm not too worried about the noise. You are right though, the divide is wiggling quite a bit.

t4c1:
> @t4c1 can you do the arena + var_value<Eigen> stuff

What exactly do you mean? Benchmarking?

bbbales2:
> What exactly do you mean? Benchmarking?

I imagine if I plugged `var_value<MatrixXd>` in for `m` here:

https://github.com/stan-dev/math/blob/97fedc314c026653a84705996f5ce2a7195ba2af/stan/math/rev/fun/determinant.hpp#L16

that I would not quite get what I expected (for instance this might create a lot of little vars: https://github.com/stan-dev/math/blob/97fedc314c026653a84705996f5ce2a7195ba2af/stan/math/rev/fun/determinant.hpp#L24, or maybe the template specialization wouldn't work, etc.).

t4c1:
Yeah, all functions need to be modified to accept var_value. Anyway I would prefere if you started with this. For the ones I reimplemented with `reverse_pass_callback` the changes will probably be minimal, so I would suggest starting with those. I might help with some, however I am already working on a few thigs in parallel.

bbbales2:
Sure thing can do

SteveBronder:
> I agree with --benchmark_repetitions=30. also it would be better to benchmark this against develop, not last release.

Yes I also like to use `--benchmark_report_aggregates_only=false` then you can pull out the times and look at the 95% CI in R

SteveBronder:
> Benching against develop tells us how fast var_value<T> is against 3.3 vars, which is the thing I want to know. I'm not too concerned with overhead in the arena/callback stuff at this point (which I think is what comparing against develop would tell us).

Did you mean "Benchmarking this branch against 3.3"? idt we are going to revert the `var_value` changes so it would probs be better to bench against develop. I'm going to put up the cmdstan PR to add the compiler optims this week so idt we should worry too much about 3.3

bbbales2:
> changes so it would probs be better to bench against develop

I can just benchmark against develop as well.

stan/math/rev/fun/rows_dot_product.hpp
```diff
@@ -17,12 +16,43 @@ template <typename Mat1, typename Mat2,
...
+  arena_matrix<promote_scalar_t<var, Mat1>> arena_v1;
+  arena_matrix<promote_scalar_t<var, Mat2>> arena_v2;
+
+  if (!is_constant<Mat1>::value) {
+    arena_v1 = v1_ref;
+  }
+
+  if (!is_constant<Mat2>::value) {
+    arena_v2 = v2_ref;
   }
```
SteveBronder, line 36: How would yinz feel about something like 
```suggestion
  using arena_v1_type = conditional_arena_matrix<promote_scalar_t<var, Mat1>, !is_constant<Mat1>::value>;
  arena_v1_type arena_v1(v1_ref);
  using arena_v2_type = conditional_arena_matrix<promote_scalar_t<var, Mat2>, !is_constant<Mat2>::value>;
  arena_v2_type arena_v2(v2_ref);
```

So in the case of the RHS being `false` the `conditional_arena_matrix` is an `Eigen::Map<promote_scalar_t<var, Mat2>>(nullptr, 0, 0)`

stan/math/rev/fun/inverse.hpp
```diff
@@ -75,16 +21,23 @@ class inverse_vari : public vari {
...
   check_square("inverse", "m", m);
   if (m.size() == 0) {
     return {};
   }
 
-  matrix_v res(m.rows(), m.cols());
-  internal::inverse_vari *baseVari = new internal::inverse_vari(m);
-  res.vi() = Eigen::Map<matrix_vi>(baseVari->vari_ref_A_inv_, res.rows(),
-                                   res.cols());
+  arena_matrix<Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>> arena_m = m;
```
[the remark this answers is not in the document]
SteveBronder: Mostly asking because it looks like in most places we than access these things by `.val()` or `.adj()`. I think it's just the cost calculation of whether it's better to allocate N vars and have non-contiguous access on the `val()` and `adj()` or make N*2 double allocations + 1 vari allocation and have contiguous access on the `val()` and `adj()`

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,34 @@
...
+
+  if (L.rows() == 0)
+    return T_var();
+
+  arena_matrix<T_var> arena_L = L;
+  arena_matrix<T_double> arena_L_val
+      = value_of(arena_L).template triangularView<Eigen::Lower>();
+
+  arena_matrix<T_var> res = arena_L_val.template triangularView<Eigen::Lower>()
+                            * arena_L_val.transpose();
```
SteveBronder, line 28: Another place I think var_value<Matrix> could be used. We would have something like 
```suggestion
  // Should arena_matrix with var<matrix> work?
  arena_matrix<var_matrix_converter_t<T_var>> arena_L = L;
  arena_matrix<T_var> res = arena_L.val().template triangularView<Eigen::Lower>() * arena_L.val().transpose();
```

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,34 @@
...
+
+  reverse_pass_callback([res, arena_L, arena_L_val]() mutable {
+    ref_type_t<decltype(res.adj())> adj = res.adj();
+    Eigen::MatrixXd adjL = (adj.transpose() + adj) * arena_L_val;
+
+    for (size_t j = 1; j < adjL.cols(); ++j)
+      for (size_t i = 0; i < std::min(static_cast<size_t>(adjL.rows()), j); ++i)
+        adjL(i, j) = 0.0;
```
SteveBronder, line 36:
```suggestion
  adjL.lowerTriangular<Eigen::Lower>().setZero()
```

Is this zeroing out the lower triangular? If you are worried about non-square matrices Eigen can handle that

https://godbolt.org/z/bc6P4Y

stan/math/rev/fun/rows_dot_product.hpp
```diff
@@ -17,12 +16,43 @@ template <typename Mat1, typename Mat2,
...
-  return ret;
+
+  Eigen::VectorXd out_val(arena_v1_val.rows());
+  for (size_t m = 0; m < arena_v1_val.rows(); ++m)
+    out_val.coeffRef(m) = arena_v1_val.row(m).dot(arena_v2_val.row(m));
```
SteveBronder, line 40: From [this](https://stackoverflow.com/a/53720369/2269255) stackoverflow it looks like the below is a hair faster

```suggestion
  Eigen::VectorXd out_val =  (arena_v1_val.cwiseProduct(arena_v2_val)).rowwise().sum();
```

Seems to have same result
https://godbolt.org/z/cqnvnv

stan/math/rev/fun/tcrossprod.hpp
SteveBronder: For `var_value<MatrixXd>` the `ref_type_t<decltype(res.adj())>` should just be the same as `adj_`'s type (`Eigen::Map<double, R, C>`) right? If so idt this would need any changes

stan/math/rev/core/var.hpp
SteveBronder: Why don't we just get rid of the `const` version of `adj()`?

SteveBronder: Or make `adj_` in the vari implementation `mutable` so even if we have a `const var` or whatever the `adj_` is still mutable

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -19,12 +20,43 @@ template <typename Mat1, typename Mat2,
...
+  }
+
+  if (!is_constant<Mat2>::value) {
+    arena_v2 = v2_ref;
   }
-  return ret;
+
+  Eigen::RowVectorXd out_val(arena_v1_val.cols());
+  for (size_t m = 0; m < arena_v1_val.cols(); ++m)
+    out_val.coeffRef(m) = arena_v1_val.col(m).dot(arena_v2_val.col(m));
```
SteveBronder, line 44: See the other comment that talks about the row version of this for the splainer'
```suggestion
Eigen::VectorXd out_val = (arena_v1_val.cwiseProduct(arena_v2_val)).colwise().sum();
```

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,34 @@
...
-      LLt.coeffRef(m, n) = LLt.coeffRef(n, m)
-          = dot_product(L.row(m).head((J < (n + 1)) ? J : (n + 1)),
-                        L.row(n).head((J < (n + 1)) ? J : (n + 1)));
-    }
-  }
-  return LLt;
+template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+inline Eigen::Matrix<var, T::RowsAtCompileTime, T::RowsAtCompileTime>
+multiply_lower_tri_self_transpose(const T& L) {
+  using T_double = promote_scalar_t<double, T>;
```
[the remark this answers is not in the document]
bbbales2: Talked to @t4c1 and we thought `set_scalar_t` would be good (https://github.com/stan-dev/math/pull/2019#discussion_r472824795). It just needs a separate issue and someone to do it.

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
bbbales2: > // Should arena_matrix with var<matrix> work?

That hasn't been done yet. Tadej said he's busy right now. If you want to open the pull, I'm happy to review. I very much want to see these benchmarks with `var_value<Eigen>` types.

SteveBronder: Cool cool I'll work on that

SteveBronder:
Alright I started fooling around with arena matrix stuff for `var<mat>` on [this](https://github.com/stan-dev/math/tree/feature/new-adj-tadj-var_value) branch. Have some benchmarks I haven't prettied up yet [here](https://gist.github.com/SteveBronder/79f42ae8addd7ccaeedf5e648b62e6c3) but essentially after 32x32 sizes `var<mat>` is about twice as fast and then blasts off pretty exponentially.

I'm having some goofs with getting `var<mat>` types to play nicely with the dynamic matrices and `arena_matrix` (at one point I had code that compiled on clang but not gcc lol). My guess is that if we do dynamic matrices and under the hood convert them to var<mat> it will be a bit of a cost for small matrices but larger matrices should see a nice benefit.

stan/math/rev/fun/inverse.hpp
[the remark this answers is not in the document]
t4c1: Well at the end we have to update adjoints of the `m`. I think right now we do not have an operation that would propagate adjoints form `var_value<matrix>` to `Matrix<var>`. Even if we did it would do exactly what is done here. So doing it directly is preferable.

stan/math/rev/fun/rows_dot_product.hpp
t4c1: That is more or less how the new `to_arena_if` in https://github.com/stan-dev/math/pull/2026 works. After that is merged this can be:
```suggestion
  arena_matrix<promote_scalar_t<var, Mat1>> arena_v1 = to_arena_if<!is_constant<Mat1>::value>(v1_ref);
  arena_matrix<promote_scalar_t<var, Mat2>> arena_v2 = to_arena_if<!is_constant<Mat2>::value>(v2_ref);
```

stan/math/rev/fun/inverse.hpp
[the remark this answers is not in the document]
t4c1: Once we start supporting `var_value` we need something here that will support both option, such as
```suggestion
  arena_t<T> arena_m = m;
```

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
t4c1: > Tadej said he's busy right now.
I'm not too busy for reviewing or answering questions. I am just not starting another PR like this on my own.

> Should arena_matrix with var<matrix> work?
No. `arena_matrix` is intended for eigen types. `var_value` has allocations on stack anyway, so it can replace `arena_matrix<Eigen something>`. Or you can use `arena_t<>` on either Eigen type or `var_value` and you will get something that can be captured in lambda for `reverse_pass_callback`.

t4c1:
> Alright I started fooling around with arena matrix stuff for var<mat> on this branch.

As explained [in this comment](https://github.com/stan-dev/math/pull/2019#discussion_r477056600) we do not need that at all. For transparent handling of both matrix stypes you can use `arena_t` to determine type to use that will store tha data in arena.

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
bbbales2: > var_value has allocations on stack anyway

Oh right. It's just a pointer.

> I'm not too busy for reviewing or answering questions.

Right right, just letting Steve know that nobody else was working on it.

stan/math/rev/fun/inverse.hpp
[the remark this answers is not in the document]
bbbales2: I'm not sure we need it. If the input is a `Matrix<var, R, C>` type, copying to a `var_value<Matrix>` type only saves us time if the adjoints and vals are accessed multiple times. Adjoints are basically only accessed once ever in a custom reverse pass implementation. If values are used more than once, they can be copied out separately.

stan/math/rev/fun/rows_dot_product.hpp
SteveBronder: Nice! Yeah I was doing this on the other branch I was playing around in

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
SteveBronder: I'll open up a draft PR for this tmrw

> Or you can use arena_t<> on either Eigen type or var_value and you will get something that can be captured in lambda for reverse_pass_callback.

I'll write more when I open the draft, I need to play with it more, but I just wrote a specialiation for `arena_matrix` that just inherits from a `var_value<>` eigen. We might want that style if we want to be able to reuse the same functions for both var<mat> and mat<var> and their combinations

stan/math/rev/fun/rows_dot_product.hpp
SteveBronder: Odd, if I move that expression directly into the assignment like

```cpp
arena_matrix<Eigen::Matrix<var, Eigen::Dynamic, 1>> out = (arena_v1_val.cwiseProduct(arena_v2_val)).rowwise().sum();
``
it throws an error from eigen

File: stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp (hunk 21 above)
Comment (bbbales2) [reply stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp:28]: > I'll open up a draft PR for this tmrw

Cool. Feel free to take any of the functions here and use in that pull, or softmax is already in: https://github.com/stan-dev/math/blob/develop/stan/math/rev/fun/softmax.hpp

File: stan/math/rev/fun/rows_dot_product.hpp (hunk 23 above)
Comment (t4c1) [reply stan/math/rev/fun/rows_dot_product.hpp:40]: I think `.sum()` is missing.

**Comment by bbbales2** [comment]
I put up new benchmarks in a shiny app: https://bbbales2.shinyapps.io/math-shiny-perf/

I'll try to update that as I push stuff here. I'm putting the benchmark code in this pull request for now: https://github.com/bbbales2/perf-math/pull/1 .

The plan is to benchmark every function we change this release since we're doing so much new stuff and the benchmarks are relatively easy to add. You might notice some misspellings. That's cuz the script I have processing the output removes "ns" strings.

**Comment by SteveBronder** [comment]
Just a note: Once #2050 goes in we should take the function from here we are 100% are more efficient and make a PR out of them that also has the `var<mat>` versions

**Comment by t4c1** [comment]
> I put up new benchmarks in a shiny app

Wow, it looks like you put quite some work in these benchmarks. Given the number of functions to benchmark it might make sense to automatically generate them for all functions from signatures in stanc3, like I did with expressions tests. I could do that.

**Comment by bbbales2** [comment]
Updated: https://bbbales2.shinyapps.io/math-shiny-perf/

I think these uncertanties are super underestimated. Look at dot product. It is unchanged in this pull. Unfortunately when I upload a new version of the site it nukes the old version. I think I need to figure out how to turn off cpu scaling on my computer and run these things when I'm not using it at all. I dunno how to randomize the execution order.

Check out `exponential_lccdf_data`. I wonder what happened lol? Is that an artifact in the benchmarking?

I will add some tools to dig around in the benchmarks more quickly (sort by largest performance loss -- largest average performance loss, etc.)

> automatically generate them

Maybe eventually. Not yet. Right now it takes less time to add the benchmark than convert the functions to the new reverse mode thing, so not so bad. Mostly copy-paste. The list I'm worrying about are functions with custom autodiff ([here](https://gist.github.com/bbbales2/a1b2bcc1f616820ce5db59b9be46c49c) and distributions), and only certain signatures.

I would like to get these running regularly/updating regularly. It'd be good to save the old experiments too. But I also don't wanna spend too much time on infrastructure before finishing the first pass conversion.

**Comment by SteveBronder** [comment]
If your on ubuntu you can turn of cpu scaling with 
```
sudo cpupower frequency-set --governor performance
```

Besides that you could also slim down the benchmarks by getting rid of anything that's unnecessary (i.e. the cost of making the tuples your using and the call to `apply()`. You'll have to repeat code in the benches but it'll get rid of a lot of stuff that can be giving you variance

**Comment by bbbales2** [comment]
> sudo cpupower frequency-set --governor performance

Thanks. Next time I run I'll do this.

I think this is keeping the variable initialization outside of the timing zone right: https://github.com/bbbales2/perf-math/pull/1/files#diff-d212271de28991ee7a73465df51423f3R19 ?

**Comment by SteveBronder** [comment]
Oh yeah, you still pay the `apply()` cost though that's probably low overhead. Also what compiler flags are you using? Would be good to have on `-march=native -mtune=native`. I'll put up the PR for cmdstan to use the extra optim flags today

**Comment by bbbales2** [comment]
> Also what compiler flags are you using?

Just defaults. I'll add `-march=native` and `-mtune=native`

**Comment by SteveBronder** [comment]
Do you have the code up somewhere to run those benchmarks and make the shiny app? I'd like to test those out with the new compiler flags as well (they are also available [here](https://github.com/stan-dev/cmdstan/blob/d8bb3d28950ac4239ae2db949edccf80ff566a91/makefile#L62) if you want to try them out

**Comment by bbbales2** [comment]
This is the shiny code: https://github.com/bbbales2/shiny-math-perf

The benchmark code is here: https://github.com/bbbales2/perf-math/pull/1

Inside the the perf-math repo I run this script:

```
for branch in feature/convert_to_tadj1 fdf5db8 develop
do
    cd math
    git pull
    git checkout $branch
    make clean-all
    make -f make/standalone math-libs -j8
    cd ../
    make -f callback.make clean_callback
    make -f callback.make build -j8
    make -f callback.make callback/benchmark.csv
    mv callback/benchmark.csv `basename $branch.csv`
done
```

It's rather fragile lol. If you feel like changing any bit of it, feel free, with the added caveat that it might all get changed again quickly. It's all experimental to see what is useful.

File: stan/math/rev/fun/mdivide_left.hpp [hunk 26]
```diff
@@ -12,238 +12,48 @@
...
-inline Eigen::Matrix<var, T1::RowsAtCompileTime, T2::ColsAtCompileTime>
-mdivide_left(const T1 &A, const T2 &b) {
-  check_square("mdivide_left", "A", A);
-  check_multiplicable("mdivide_left", "A", A, "b", b);
-  if (A.size() == 0) {
-    return {0, b.cols()};
-  }
+  reverse_pass_callback([arena_A, arena_B, arena_A_val, res]() mutable {
+    Eigen::Matrix<double, T1::RowsAtCompileTime, T2::ColsAtCompileTime> adjB
+        = arena_A_val.transpose().colPivHouseholderQr().solve(res.adj());
```
bbbales2, line 48: Is there a way to store things like the output of a `covPivHouseholderQr` in the arena (or the output of llt, ldlt, etc.)? In this function we recompute the decomposition in the callback cause it's not easy to save stuff.

stan/math/rev/fun/mdivide_left_tri.hpp
```diff
@@ -285,11 +285,11 @@ class mdivide_left_tri_vd_vari : public vari {
...
 #endif
       adjA.noalias()
-          = -Map<Matrix<double, R1, C1> >(A_, M_, M_)
+          = -Map<Matrix<double, R1, C1>>(A_, M_, M_)
                  .template triangularView<TriView>()
                  .transpose()
                  .solve(adjC
-                        * Map<Matrix<double, R1, C2> >(C_, M_, N_).transpose());
+                        * Map<Matrix<double, R1, C2>>(C_, M_, N_).transpose());
 #ifdef STAN_OPENCL
```
bbbales2, line 293: I may have broken OpenCL stuff here. Not sure. I want to delete these varis though.

stan/math/rev/fun/mdivide_left.hpp
SteveBronder: No idt so, looking at the code for [`ColPivHouseholderQR`](https://eigen.tuxfamily.org/dox/ColPivHouseholderQR_8h_source.html) it won't work with `Map`. But we could write a more generic `arena_object<>` that would then inherit from `chainable_alloc` and just hold an object so it's delete get's called.

SteveBronder:
Nice! No worries it's fine if it's a little jank. Is there an updated makefile for the multiple run version?

bbbales2:
> Is there an updated makefile for the multiple run version?

The only thing I have for the multiple runs is the bash script here: https://github.com/stan-dev/math/pull/2019#issuecomment-684165872

I'm not sure about using Make for anything more than I did lol. It was already kinda painful. I think only build the binaries with make. The actually running of the binaries and such should probably be done with something else.

Updated from yesterday: https://bbbales2.shinyapps.io/math-shiny-perf/ . I think our old matrix power function was leaking memory, fwiw.

stan/math/rev/fun/mdivide_left.hpp
bbbales2: > But we could write a more generic arena_object<> that would then inherit from chainable_alloc

Hmm that does sound interesting.

Could we implemented something like a `shared_ptr` for our arena? It would track things allocated on the heap that will be destructed with `recover_memory`/`recover_memory_nested`. The shared pointers themselves would be lightweight so they could be copied around easily.

So the code here would become:

```cpp
auto hhqr = stan::math::make_shared(arena_A_val.transpose().colPivHouseholderQr());
arena_matrix<Eigen::Matrix<var, T1::RowsAtCompileTime, T2::ColsAtCompileTime>>
      res = hhqr->solve(value_of(B_ref));

reverse_pass_callback([arena_A, arena_B, arena_A_val, res, hhqr]() mutable {
    Eigen::Matrix<double, T1::RowsAtCompileTime, T2::ColsAtCompileTime> adjB
        = hhqr->solve(res.adj());
  ...
```

SteveBronder:
Some of these look good! I think we need to be careful with the distribution changes tho' some of those seem a bit worrisome. Though for other things like `gp_cov_exp_quad_likely` the answer there is just to either dig in an figure out where the diff comes from or just write a overload for that case that uses the old version. Then we get the fancy fast cases and keep the old speed case as well

bbbales2:
> gp_cov_exp_quad_likely

This function is basically always attached to a cholesky decomposition too. I wanna see those numbers before I worry. This is O(N^2) and Cholesky is O(N^3) so it might just be fast enough already. I expect most models are in the N = 10-200 range.

> distribution changes tho' some of those seem a bit worrisome

Not all of them have been merged yet. I merged cholesky and skew normal. There were only major changes in the lpdfs for those. I dunno why the exponential distribution is up to 10% slower in develop now vs the last release for instance. There's not a lot of polymorphic chain calls there.

bbbales2:
The reason I added exponential is cause @t4c1 has a pull request here: https://github.com/stan-dev/math/pull/2028 . So this was in preparation for that.

stan/math/rev/fun/mdivide_left.hpp
t4c1: You may want to check out how I implemented arena version of `matrix_cl`.

bbbales2:
@t4c1 I converted the cpu part of cholesky_decompose over to the new thing here: https://github.com/stan-dev/math/blob/843ca080a37320a144d7ac74373f7d3d0dca7c1f/stan/math/rev/fun/cholesky_decompose.hpp

Do you wanna take these changes and do the rest of the OpenCL conversion in a different pull? Or do you want me to copy-paste in the old OpenCL code here?

I also made the reverse mode tests run whether or not OpenCL is enabled (so grab the test file if you take the CPU stuff too). Without those tests we weren't actually testing the Cholesky factorization for large matrices.

t4c1:
> Do you wanna take these changes and do the rest of the OpenCL conversion in a different pull? Or do you want me to copy-paste in the old OpenCL code here?

I guess you can just leave the old code in for now. I will eventually rewrite it, but I guess you can merge cpu changes first.

> I also made the reverse mode tests run whether or not OpenCL is enabled (so grab the test file if you take the CPU stuff too). Without those tests we weren't actually testing the Cholesky factorization for large matrices.

I guess it is tested in mix tests well enough and we don't actually need to test cpu code on large matrices. We only need to do it for OpenCL to check partitioning into smaller ones (that only happens for large matrices) and multiple work groups work too.

bbbales2:
@SteveBronder, @t4c1 once https://github.com/stan-dev/math/pull/2054 and https://github.com/stan-dev/math/pull/2050 are in, this pull request should only function rewrites.

This rewrite was done to make it easy to add `var<mat>` support to all of these functions, but I would like to get these changes merged before we go through and make all of these function support `var<mat>`.

I'm gonna work on getting the tests to pass. The performance plots are up to date (https://bbbales2.shinyapps.io/math-shiny-perf/) but I haven't gone through yet and checked everything. My hope is that we can merge this into develop without any performance slowdowns vs develop, but this might not be possible.

Anyway don't need to review this until the tests pass, just updating the plan here.

bbbales2:
@t4c1 do you know what the error is here: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2019/62/pipeline ? It's something with opencl. It's in gp_periodic_cov (which I changed in this pull), but I don't see opencl stuff in the develop implementation of gp_periodic_cov.

t4c1:
I looks like something related to includes. Possibly some are missing or more likely there are some circular includes.

SteveBronder:
@bbbales2 it was a header issue I fixed up, not related to an opencl thing. Also had a go at making gp_exp_quad_cov faster in the more likely case though doing some simple benches idt it helped out much

bbbales2:
@SteveBronder nice thanks!

bbbales2:
Noooo, we're green checked but merge conflict.

I'm gonna fix this merge conflict but this is green check central so ready for big ol' review.

bbbales2:
Oh I forgot, this isn't really ready to go until we work through the benchmarks.

I'm running updated benchmarks for this. I changed the default comparison to point to the stuff here since this will take the most work. The right side selection box should be 'convert_to_tadj1_0aaea95d30_matvar' and the branch on the left is a branch right before `var_value<T>` got merged in: https://bbbales2.shinyapps.io/math-shiny-perf/

Sort the table by minimum speedup. Once you find a function that you want to look at more closely, you can type the name in the box below for the plots.

stan/math/rev/fun/trace_quad_form.hpp
```diff
@@ -14,83 +14,51 @@
...
-namespace internal {
-template <typename Ta, int Ra, int Ca, typename Tb, int Rb, int Cb>
-class trace_quad_form_vari_alloc : public chainable_alloc {
- public:
-  trace_quad_form_vari_alloc(const Eigen::Matrix<Ta, Ra, Ca>& A,
-                             const Eigen::Matrix<Tb, Rb, Cb>& B)
-      : A_(A), B_(B) {}
 
-  double compute() { return trace_quad_form(value_of(A_), value_of(B_)); }
+template <typename EigMat1, typename EigMat2,
```
bbbales2, line 18: Even though this has two arguments and could be broken out into three implementations (code [here](https://github.com/stan-dev/math/blob/4f3705c1386017faec72ebf852f0d1e754c3d863/stan/math/rev/fun/trace_quad_form.hpp)), I went with one.

Here are benchmarks broken into three different var/double input implementations (I don't know why the first is slow -- probably something that could be fixed):
```
----------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
----------------------------------------------------------------------------------
toss_me                                   10067059 ns     10057209 ns           70
trace_quad_form/2/manual_time                  419 ns          724 ns      1663771
trace_quad_form/4/manual_time                  673 ns         1331 ns      1019910
trace_quad_form/8/manual_time                 1434 ns         3460 ns       491865
trace_quad_form/16/manual_time                5120 ns        13416 ns       134934
trace_quad_form/32/manual_time               30203 ns        68312 ns        23236
trace_quad_form/64/manual_time              213271 ns       381398 ns         3592
trace_quad_form/128/manual_time            1647797 ns      2342545 ns          376
trace_quad_form/256/manual_time           12940950 ns     16308948 ns           57
trace_quad_form/512/manual_time           91161131 ns    105007828 ns            6
trace_quad_form/1024/manual_time         815430043 ns    886466576 ns            1
trace_quad_form_data_A/2/manual_time           337 ns          620 ns      2154932
trace_quad_form_data_A/4/manual_time           688 ns         1455 ns      1156487
trace_quad_form_data_A/8/manual_time          1133 ns         2989 ns       540222
trace_quad_form_data_A/16/manual_time         3549 ns         9960 ns       153987
trace_quad_form_data_A/32/manual_time        19209 ns        46700 ns        37916
trace_quad_form_data_A/64/manual_time       128111 ns       240987 ns         5633
trace_quad_form_data_A/128/manual_time      948054 ns      1413701 ns          739
trace_quad_form_data_A/256/manual_time     8214071 ns     10240882 ns           86
trace_quad_form_data_A/512/manual_time    70229497 ns     78740468 ns           11
trace_quad_form_data_A/1024/manual_time  462998300 ns    510277851 ns            2
trace_quad_form_data_B/2/manual_time           353 ns          651 ns      2144927
trace_quad_form_data_B/4/manual_time           556 ns         1193 ns      1241467
trace_quad_form_data_B/8/manual_time          1021 ns         2807 ns       685590
trace_quad_form_data_B/16/manual_time         3629 ns        10400 ns       197400
trace_quad_form_data_B/32/manual_time        20944 ns        50909 ns        29866
trace_quad_form_data_B/64/manual_time       122996 ns       231928 ns         5894
trace_quad_form_data_B/128/manual_time      836762 ns      1260478 ns          792
trace_quad_form_data_B/256/manual_time     7263060 ns      9077717 ns           92
trace_quad_form_data_B/512/manual_time    55530521 ns     63071147 ns           12
trace_quad_form_data_B/1024/manual_time  424628951 ns    467905904 ns            2
```

Here are benchmarks for the single implementation:
```
----------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
----------------------------------------------------------------------------------
toss_me                                   10728582 ns     10738756 ns           50
trace_quad_form/2/manual_time                  498 ns          869 ns      1720492
trace_quad_form/4/manual_time                  739 ns         1481 ns       916035
trace_quad_form/8/manual_time                 1634 ns         3890 ns       425756
trace_quad_form/16/manual_time                5417 ns        14142 ns       133363
trace_quad_form/32/manual_time               28244 ns        64834 ns        24729
trace_quad_form/64/manual_time              177849 ns       320618 ns         3782
trace_quad_form/128/manual_time            1303900 ns      1881009 ns          527
trace_quad_form/256/manual_time           11417925 ns     14418493 ns           60
trace_quad_form/512/manual_time           85913338 ns     99149686 ns            7
trace_quad_form/1024/manual_time         691839532 ns    761973120 ns            1
trace_quad_form_data_A/2/manual_time           306 ns          565 ns      2316540
trace_quad_form_data_A/4/manual_time           504 ns         1051 ns      1349713
trace_quad_form_data_A/8/manual_time          1008 ns         2676 ns       695297
trace_quad_form_data_A/16/manual_time         3438 ns         9724 ns       203602
trace_quad_form_data_A/32/manual_time        18199 ns        44801 ns        38201
trace_quad_form_data_A/64/manual_time       118339 ns       224733 ns         5928
trace_quad_form_data_A/128/manual_time      855046 ns      1283016 ns          815
trace_quad_form_data_A/256/manual_time     8964877 ns     11102297 ns           91
trace_quad_form_data_A/512/manual_time    65373202 ns     73202014 ns           10
trace_quad_form_data_A/1024/manual_time  547440607 ns    591318001 ns            1
trace_quad_form_data_B/2/manual_time           329 ns          605 ns      2215157
trace_quad_form_data_B/4/manual_time           502 ns         1062 ns      1365079
trace_quad_form_data_B/8/manual_time           931 ns         2611 ns       751464
trace_quad_form_data_B/16/manual_time         3356 ns         9701 ns       207811
trace_quad_form_data_B/32/manual_time        17583 ns        43897 ns        39979
trace_quad_form_data_B/64/manual_time       117260 ns       223372 ns         5960
trace_quad_form_data_B/128/manual_time      944731 ns      1409704 ns          739
trace_quad_form_data_B/256/manual_time     7431989 ns      9315718 ns           87
trace_quad_form_data_B/512/manual_time    56982713 ns     64847214 ns           12
trace_quad_form_data_B/1024/manual_time  444922367 ns    490421169 ns            2
```

stan/math/rev/fun/determinant.hpp
```diff
@@ -3,56 +3,35 @@
...
   check_square("determinant", "m", m);
+
   if (m.size() == 0) {
     return 1;
   }
 
-  return {new internal::determinant_vari<T::RowsAtCompileTime,
-                                         T::ColsAtCompileTime>(m)};
+  auto arena_m = to_arena(m);
+  auto m_lu = arena_m.val().partialPivLu();
```
bbbales2, line 24: This is the same algorithm that was being used previously. The implementation just looks different, see https://github.com/stan-dev/math/issues/2120

bbbales2:
@SteveBronder this is ready for review. I'm gonna work on the red light, but I tried to go through everything and update the code style to be like what we're converging on.

I also benchmarked and optimized everything. It should be good to go. I'm going to update the benchmarks overnight. You can browse them as of last night here (they're pretty okay): https://bbbales2.shinyapps.io/math-shiny-perf/

I wrote these things to be as fast as possible for `mat<var>`. We might have to branch some of them for `var<mar>`, but probably a lot should just be a matter of changing the return type calculations and adding tests. I did a lot of testing breaking stuff out into different variations for different template argument types. If it wasn't faster, I went back to the single if-statement version.

SteveBronder:
Running through these now. Because it's a larger PR it's a bit easier for me to make another branch with suggested changes I'm working on / benching over here

https://github.com/stan-dev/math/compare/feature/convert_to_tadj1...feature/convert-to-tadj1-changes

The tl;dr so far is that there's a good few functions here we know for a fact are much better and should port them into a separate PR so they can go in this release. But there's a handful of functions that I'm wishy washy on (gp_exp_quad_cov and the like among them) and need more work.

Why don't we cherry pick the ones that are much faster and are probably used a lot like

- `unit_vector_constrain`
- `tcrossprod`
- `simplex_constrain`
- `log_softmax`
-  `mdivide_left_tri`
- `ordered_constrain`
- `divide`

bbbales2:
Yo you around for a call? I don't like breaking this up.

The point was to convert stuff over. If there is something that is too slow, let's fix it now while we're going through it.

I see `mat<var>` stuff showing up here too which was not the intention of this pull either, though I'm not opposed to working on a mega-pull with that stuff too.

bbbales2:
I checked out the benchmarks for this. The comparison we want I think is this branch vs. develop.

So on the benchmark page (https://bbbales2.shinyapps.io/math-shiny-perf/)

Set the left dropdown to `develop_50d56cc7ce_matvar` and the right dropdown to `convert_to_tadj1_654aa5119a_matvar`

With that I think the minimum list of functions we want to move are:

```
matrix_power
tcrossprod (have to look at graphs for this -- the table is messed up)
log_softmax
multiply_lower_self_tri_etc_etc
rows_dot_product
log_determinant
mdivide_left
mdivide_left_tri
columns_dot_product
```

I would also like ```determinant```, ```trace_quad_form``` and ```quad_form```, but at this point the benchmarks get fishy.

It says this branch vs. development that the minimum speedup for dot product is 0.9 and the maximum speedup is 1.5. Similarly for matrix multiply. I don't believe it since neither of those things change, so something is going on. It might be that we need to randomize the order of the test repetition or running two at a time is bad.

I only want to do `mat<var>`. We can pick up `var<mat>` after Monday.

bbbales2:
> There's a handful of  functions like gp_exp_quad_cov for the general case that are much much slower than what's in develop right now (I ran the gp benchmark example with that branch and it's 13,000% slower!)

Was that a 13% (makes sense) or 13000% (I would not expect this and this would bear further investigation)?

SteveBronder:
Ah that came from the cmdstan performance tests repo

You can also see it here when the performance tests ran on this PR

https://github.com/stan-dev/math/pull/2019#issuecomment-702455669

bbbales2:
OOOo, nice, I was not paying attention to those numbers! Thanks

bbbales2:
I think we've done the majority of the conversions here, so I'm gonna close this. The issue itself should stay open cause there's still a lot of simplification/cleanup we can do.

PR #2020, opened 2020-08-18
Adds flags for compiler optimizations

### LINKED ISSUE #2008: Performance regression due to absence of lto
## Description
The freshly merged static matrix refactor things do slow down stan programs significantly whenever lto is not turned on.

So as of now in the static matrix world we absolutely need the lto optimisation for good performance. Thus, our makefiles should ideally switch on lto whenever the compiler is capable enough of handling it. The user should not be required to add `-flto` to the `make/local` to get good performance

## Example

See https://github.com/stan-dev/math/pull/2007#issuecomment-672028909

where it is demonstrated that turning on lto gives back the old performance numbers.

## Expected Output

Results should be produced as fast as in the pre static matrices.

#### Current Version:
v3.3.0

SteveBronder:
## Summary

After adding a template parameter to `vari` we saw a 20% regression in the SIR model. This PR includes some extra compiler flag options so if a user has either `clang >= 8` or `gcc >= 5`. flto and other compiler optimization flags will turn on. flto will only turn on for `clang >= 8` because clang-6 and clang-7's `-dumpversion` give `4.2.1` so we cannot tell them apart. `clang-6` requires the use of either the `ar` gold linker plugin or `llvm-ar`. Since the gold plugin is a non-standard extension for some versions of `ar` to have installed we turn off flto by default for both clang 6 and 7

## Tests

make feature so no new tests, though because our benchmarks use clang-6 they won't give the new benchmark values so I will paste them here. These tests were run on Ubuntu 20.04 with AMD Threadripper 1950x along with DDR4 ram. Tests for each compiler were run 10 times then averaged together

#### Clang-10 

The clang-10 results were run with 5000 warmup and 5000 samples, though overall I didn't see the increase in iteration numbers help produce better reliability in the test numbers.

| gp_pois_regr/gp_pois_regr.stan | 7.1 | 6.71 | 1.06 | 5.49% faster |
| low_dim_corr_gauss/low_dim_corr_gauss.stan | 0.01 | 0.02 | 0.88 | -13.0% slower |
| eight_schools/eight_schools.stan | 0.18 | 0.17 | 1.03 | 3.28% faster |
| arK/arK.stan | 5.27 | 5.28 | 1.0 | -0.16% slower |
| irt_2pl/irt_2pl.stan | 15.57 | 15.87 | 0.98 | -1.88% slower |
| performance.compilation | 48.23 | 58.49 | 0.82 | -21.28% slower |
| low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan | 23.93 | 22.65 | 1.06 | 5.38% faster |
| pkpd/one_comp_mm_elim_abs.stan | 59.22 | 56.66 | 1.05 | 4.31% faster |
| sir/sir.stan | 224.85 | 190.03 | 1.18 | 15.48% faster |
| pkpd/sim_one_comp_mm_elim_abs.stan | 0.25 | 0.25 | 1.01 | 0.75% faster |
| low_dim_gauss_mix/low_dim_gauss_mix.stan | 6.93 | 6.61 | 1.05 | 4.66% faster |
| gp_regr/gp_regr.stan | 0.33 | 0.32 | 1.03 | 2.54% faster |
| arma/arma.stan | 0.76 | 0.75 | 1.02 | 1.96% faster |
| gp_regr/gen_gp_data.stan | 0.03 | 0.03 | 1.0 | 0.24% faster |
| garch/garch.stan | 1.97 | 1.75 | 1.13 | 11.35% faster |
Mean result: 1.0199644747

#### GCC-10

Because I didn't see any increased reliability from running more than the standard number of iterations the gcc results were run with 1000 warmup and 1000 samples. 

| gp_pois_regr/gp_pois_regr.stan | 2.34 | 1.98 | 1.18 | 15.29% faster |
| low_dim_corr_gauss/low_dim_corr_gauss.stan | 0.01 | 0.01 | 1.05 | 4.75% faster |
| eight_schools/eight_schools.stan | 0.05 | 0.05 | 0.99 | -0.91% slower |
| arK/arK.stan | 1.24 | 1.25 | 0.99 | -0.95% slower |
| irt_2pl/irt_2pl.stan | 4.22 | 4.03 | 1.05 | 4.43% faster |
| performance.compilation | 86.27 | 96.83 | 0.89 | -12.23% slower |
| low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan | 4.59 | 4.7 | 0.98 | -2.53% slower |
| pkpd/one_comp_mm_elim_abs.stan | 14.17 | 14.03 | 1.01 | 0.95% faster |
| sir/sir.stan | 87.9 | 70.69 | 1.24 | 19.57% faster |
| pkpd/sim_one_comp_mm_elim_abs.stan | 0.23 | 0.23 | 1.0 | 0.11% faster |
| low_dim_gauss_mix/low_dim_gauss_mix.stan | 1.68 | 1.7 | 0.99 | -1.08% slower |
| gp_regr/gp_regr.stan | 0.09 | 0.1 | 0.93 | -8.02% slower |
| arma/arma.stan | 0.33 | 0.34 | 0.98 | -1.56% slower |
| gp_regr/gen_gp_data.stan | 0.02 | 0.02 | 0.99 | -1.37% slower |
| garch/garch.stan | 0.34 | 0.37 | 0.91 | -9.7% slower |
Mean result: 1.01176292109

## Side Effects

There's are some compiler flags which require the gcc graphite compiler extension, though tmk this comes with most standard builds of gcc unless you build gcc from source in which case I believe you need to specify whether the graphite extension should be compiled or not. If a user does not have graphite setup a warning is given and the program continues to run.

One of the interesting thing I found out is that several of these tests have a pretty wide variance. For instance here is the results of running develop against develop 10 times

---RESULTS---
| gp_pois_regr/gp_pois_regr.stan | 2.34 | 2.29 | 1.02 | 2.37% faster |
| low_dim_corr_gauss/low_dim_corr_gauss.stan | 0.01 | 0.01 | 0.92 | -8.15% slower |
| eight_schools/eight_schools.stan | 0.05 | 0.05 | 1.01 | 0.92% faster |
| arK/arK.stan | 1.27 | 1.3 | 0.98 | -1.86% slower |
| irt_2pl/irt_2pl.stan | 4.23 | 4.18 | 1.01 | 1.24% faster |
| performance.compilation | 60.81 | 60.2 | 1.01 | 1.0% faster |
| low_dim_gauss_mix_collapse/low_dim_gauss_mix_collapse.stan | 4.63 | 4.79 | 0.97 | -3.4% slower |
| pkpd/one_comp_mm_elim_abs.stan | 14.07 | 14.15 | 0.99 | -0.57% slower |
| sir/sir.stan | 82.19 | 81.9 | 1.0 | 0.36% faster |
| pkpd/sim_one_comp_mm_elim_abs.stan | 0.25 | 0.25 | 1.03 | 2.95% faster |
| low_dim_gauss_mix/low_dim_gauss_mix.stan | 1.71 | 1.73 | 0.99 | -1.28% slower |
| gp_regr/gp_regr.stan | 0.1 | 0.1 | 1.01 | 1.36% faster |
| arma/arma.stan | 0.35 | 0.34 | 1.02 | 2.22% faster |
| gp_regr/gen_gp_data.stan | 0.03 | 0.03 | 1.05 | 5.15% faster |
| garch/garch.stan | 0.35 | 0.34 | 1.01 | 1.03% faster |

So `low_dim_corr_gauss` and `gen_gp_data` can have pretty large variance with even a decent number of runs

## Release notes

Add compiler flags to give better optimizations

## Checklist

- [x] Closes #2008 


make/compiler_flags
```diff
@@ -62,6 +62,23 @@ CXX_MAJOR := $(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f1)
...
+	ifeq ($(shell expr $(CXX_MAJOR) \>= 5), 1)
+		CXXFLAGS_FLTO ?= -flto=full -fwhole-program-vtables -fstrict-vtable-pointers -fforce-emit-vtables
+		AR := llvm-ar-$(CXX_MAJOR)
+	endif
+endif
+ifneq (,$(findstring mingw32-g,$(CXX_TYPE)))
+else ifneq (,$(findstring gcc,$(CXX_TYPE)))
+	CXXFLAGS_OPTIM ?= -fweb -fivopts -ftree-loop-linear -fstrict-aliasing -funroll-loops -floop-strip-mine -floop-block -floop-nest-optimize -floop-unroll-and-jam -fsplit-loops -funswitch-loops -ftree-vectorize -fvect-cost-model='unlimited' -fivopts -ftree-loop-linear 	 -fvisibility=hidden -fvisibility-inlines-hidden
+	SUNDIALS_OPTIM ?= -O3 -fweb -fivopts -ftree-loop-linear
+	ifeq ($(shell expr $(CXX_MAJOR) \>= 5), 1)
```
rok-cesnovar, line 76: `expr` is available with RTools on Windows, so this is ok Windows-wise, but I think a safer option for this logic would be to check that the CXX_MAJOR is not 4 (we dont support anything below 4 anyways), which removes the need for expr here.

We already rely on on the cut gnu tool for getting CXX_MAJOR (`$(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f1)`) so take this comment as optional.

make/compiler_flags
```diff
@@ -62,6 +62,23 @@ CXX_MAJOR := $(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f1)
 CXX_MINOR := $(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f2)
 
 
+ifneq (,$(findstring clang,$(CXX_TYPE)))
+	CXXFLAGS_OPTIM ?= -fvectorize -ftree-vectorize -fslp-vectorize -ftree-slp-vectorize -fno-standalone-debug -fstrict-return -ftrigraphs -fvisibility=hidden -fvisibility-inlines-hidden
+	ifeq ($(shell expr $(CXX_MAJOR) \>= 5), 1)
```
rok-cesnovar, line 67: This logic is for clang but you then check CXX_MAJOR is >=5 ? Can you elaborate on this?

make/compiler_flags
```diff
@@ -62,6 +62,23 @@ CXX_MAJOR := $(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f1)
...
+	endif
+endif
+ifneq (,$(findstring mingw32-g,$(CXX_TYPE)))
+else ifneq (,$(findstring gcc,$(CXX_TYPE)))
+	CXXFLAGS_OPTIM ?= -fweb -fivopts -ftree-loop-linear -fstrict-aliasing -funroll-loops -floop-strip-mine -floop-block -floop-nest-optimize -floop-unroll-and-jam -fsplit-loops -funswitch-loops -ftree-vectorize -fvect-cost-model='unlimited' -fivopts -ftree-loop-linear 	 -fvisibility=hidden -fvisibility-inlines-hidden
+	SUNDIALS_OPTIM ?= -O3 -fweb -fivopts -ftree-loop-linear
+	ifeq ($(shell expr $(CXX_MAJOR) \>= 5), 1)
+		CXXFLAGS_FLTO?=-flto=jobserver -fuse-linker-plugin
+  endif
+	ifeq ($(shell expr $(CXX_MAJOR) \> 9), 1)
```
rok-cesnovar, line 79: This can be removed I guess?

make/compiler_flags
```diff
@@ -62,6 +62,23 @@ CXX_MAJOR := $(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f1)
 CXX_MINOR := $(shell $(CXX) -dumpversion 2>&1 | cut -d'.' -f2)
 
 
+ifneq (,$(findstring clang,$(CXX_TYPE)))
```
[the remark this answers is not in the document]
rok-cesnovar: And if a user overwrites the CXX_TYPE to something unknown we should not set any optimization flags in that case.

rok-cesnovar:
We also need a quick summary to understand in which cases there is a regression. If I understand correctly then the logic is:
```
if g++ version < 5 or clang version < 8
  - some regression is expected for models where memory allocation is a big part of the model. Which I guess is ODE models? Any others?
else
  - no changes expected?
```

wds15:
Just as FYI: I did find performance regression for SIR, yes, but also for a mixture logistic regression model!

Thanks for putting this together, it looks like a lot of work! I expected that this would be quick and easy, but it‘s a lot of work as it looks.

rok-cesnovar:
> Just as FYI: I did find performance regression for SIR, yes, but also for a mixture logistic regression model!

Oh, thanks for the info. Then it will be a bit more challenging to get an intuition on.

IMHO g++ < 5 and clang < 8 is not going to be a lot our user base. Its mostly going to be RTools 3.5 Windows users, which hopefully will upgrade sooner rather than later. Anyone that cares much about performance and execution times has probably already upgraded to RTools 4.0. Last 3 Ubuntu LTS come with a compiler version >= 5.

I am not so sure on Macs, but if I am reading [this](https://trac.macports.org/wiki/XcodeVersionInfo) right, the supported clang versions in XCode for anything above and include Sierra is fine for us. I dont think there are many users still left on El Capitan and older.

stan/math/rev/fun/log.hpp
```diff
@@ -17,7 +17,7 @@ namespace stan {
 namespace math {
 
 namespace internal {
-class log_vari : public op_v_vari {
+class log_vari final : public op_v_vari {
```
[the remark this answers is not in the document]
SteveBronder: Worst case this won't hurt, but giving a class that inherits from another the `final` attribute let's the compiler know that no other classes will inherit from this class. This helps the compiler when it's trying to devirtualize member functions

stan/math/rev/core/vari.hpp
```diff
@@ -272,9 +269,9 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
         val_(eigen_map(val_mem_, x.rows(), x.cols()) = x),
         adj_(eigen_map(adj_mem_, x.rows(), x.cols()).setZero()) {
     if (stacked) {
-      ChainableStack::instance_->var_stack_.emplace_back(this);
+      ChainableStack::instance_->var_stack_.push_back(this);
     } else {
-      ChainableStack::instance_->var_nochain_stack_.emplace_back(this);
+      ChainableStack::instance_->var_nochain_stack_.push_back(this);
```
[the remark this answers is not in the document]
SteveBronder: It's not really an optimization, it's just that this was previously an over-optimization on my part. `emplace_back` is used when you want to add an object to a vector and allow the compiler to construct the element in place. `push_back` always copies the object you create into the element in the vector. Here we are just pushing pointers onto the stack which is always a copy so `emplace_back` doesn't do anything useful here. So I think it's more clear to have `push_back`

stan/math/rev/core/vari.hpp
```diff
@@ -30,7 +30,6 @@ class vari_base {
    */
   virtual void chain() = 0;
   virtual void set_zero_adjoint() = 0;
-  virtual ~vari_base() noexcept {}
```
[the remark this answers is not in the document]
SteveBronder: Also a previous goof I'm fixing in this PR. A virtual destructor allows the destructor of inherited classes to be called when an object of an inherited class is assigned to an object with the type of the base class. But the `vari_base` class never has it's destructor called since it only ever sits in our stack allocator. So this just added a pointer to a vtable we didn't need

make/compiler_flags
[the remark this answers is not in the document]
SteveBronder: Yes good call! I was just copy pasting from above but what you suggest is much more clear.

I also added a comment describing what this is doing and what flags it sets

> And if a user overwrites the CXX_TYPE to something unknown we should not set any optimization flags in that case.

Yes, looking over those lines that's what this does already right?

make/compiler_flags
SteveBronder: If it's fine I'd like to leave it because we also use one compiler flag if the gcc version is greater than 7 via

```make
ifeq ($(shell expr $(CXX_MAJOR) \>= 7), 1)
```

Tho idt it's a big deal to remove that one flag and do the `eq` and `neq` if you think this will cause issues in the future

rok-cesnovar: We can keep it. Not an issue in general. In the past we were a bit too loose with the use of gnu tools, its why I have some ptsd when I see shell :). But both Rtools 35 as 40 have expr and Rtools is our suggested “toolchain” so this is cool with me.

make/compiler_flags
SteveBronder: So this is a clang oddity. `clang-6` and `clang-7` both give `4.2.1` for `-dumpversion`. using `-v` instead of `-dumpversion` can give some weird results across operating systems and requires some extra trickery and gotchas to figure out. So while we could support `flto` by default with clang-7, clang-6 needs the `llvm-ar`*. And since we can't tell them apart I think it's just easier to only support flto by default for systems with clang > 7.

\* I thought this was true for all versions of clang, but for clang > 6 you can just use the system `ar`. I just changed the logic here so that we don't set the `AR` for clang when compiling with flto.

make/compiler_flags
SteveBronder: Yes!

SteveBronder:
> Thanks! The performance tests look good.

np!

> I am a bit confused as to which clang versions allow these flags. In the Summary you mention clang 8 but in the side effects you write: "the most important one being that if a user uses a clang-[7-10] they must also have llvm-ar-[7-10] hooked into their path as well for flto.". I think you mention that there is some confusion with clang and versions anyways?

Yes sorry I submitted this before bed. This was a previous goof. If the user is using clang > 7 then we can use the system `ar`. I'll copy/paste what I wrote above

So this is a clang oddity. clang-6 and clang-7 both give 4.2.1 for -dumpversion. using -v instead of -dumpversion can give some weird results across operating systems and requires some extra trickery and gotchas to figure out. So while we could support flto by default with clang-7, clang-6 needs the llvm-ar. And since we can't tell them apart I think it's just easier to only support flto by default for systems with clang > 7.

> Another question is regarding "they must also have llvm-ar-[7-10] hooked into their path". What does this mean for a typical Stan user that is not an expert on C++ toolchains. Is this installed for Mac users with the Xcode toolchain and works out of the box? Does one have to do anything special on Ubuntu apart from "sudo apt install clang".

My b you can ignore that

> I am not as worried for anyone that compiles gcc on their own. Those that do, do not require much assistance wrt to setting up a working toolchain anyways.


That's what I figure as well

>> One of the interesting thing I found out is that several of these tests have a pretty wide variance.

> Yep, when doing some performance benchmarks in the past, I had to run like 40000 iterations for garch, gp_regr, eight_schools, and other quick models to get consistent timings.

Ugh yeah it makes me question whether we should really use those for benchmarks. Maybe we need to change up how we do the performance tests to something like how google perf does them where they run N times till they get a stable estimate.

SteveBronder:
> Just as FYI: I did find performance regression for SIR, yes, but also for a mixture logistic regression model!

> Thanks for putting this together, it looks like a lot of work! I expected that this would be quick and easy, but it‘s a lot of work as it looks.

Yes these flags should fix up both of those cases. I can look at that logistic regression you posted as well. May be a good idea to add it to the perf tests.

> IMHO g++ < 5 and clang < 8 is not going to be a lot our user base. Its mostly going to be RTools 3.5 Windows users, which hopefully will upgrade sooner rather than later. Anyone that cares much about performance and execution times has probably already upgraded to RTools 4.0. Last 3 Ubuntu LTS come with a compiler version >= 5.

Yep!

> I am not so sure on Macs, but if I am reading this right, the supported clang versions in XCode for anything above and include Sierra is fine for us. I dont think there are many users still left on El Capitan and older.

Is our testing mac on El Capitan or Sierra? I know it has clang-7 as the default compiler so it's either an updated El Capitan or an outdated Sierra

SteveBronder:
Oh I just realized this is PR #2020 

Let's quadruple check this one so it doesn't blow up and ruin everything

rok-cesnovar:
> Oh I just realized this is PR #2020
   Let's quadruple check this one so it doesn't blow up and ruin everything

Or let's just merge and get over it ASAP :)

Joking aside, this looks good now, there just seems to be a Windows related issue at the [Cmdstan level](https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan/detail/downstream_tests/1925/pipeline). Will see if I can debug it a bit more.

SteveBronder:
Ah, mingw's gcc ran into [this](https://jenkins.mc-stan.org/blue/organizations/jenkins/CmdStan/detail/downstream_tests/1925/pipeline) compiler bug. This was fixed in gcc-8, just to be safe I'm not turning on flto for anything less than gcc-8 because I couldn't tell from the discussion whether this patch was backported to earlier gcc versions

rok-cesnovar:
Not sure I understand. This Windows machine uses g++ 8.3.0. How does setting the limit to 7 help here?

SteveBronder:
Oh shoot I thought that part of the testing used Rtools 3.5. I'll have to dive into that more to figure out what's going on then

SteveBronder:
Hmm okay [this](https://github.com/RIOT-OS/RIOT/issues/5775) hints that the issue is when a linked library is not compiled with flto but other parts are. I'll have to play around with this

rok-cesnovar:
All Jenkins agents are now running RTools 4.0. The RTools 3.5 tests were moved to Github Actions. There will be an overview doc of all of this once the CI refactor is done.

rok-cesnovar:
Hm, ok, then it might be the boost program options we build in cmdstan and statically link. Not sure where this pops up. Will take a look on my Windows machine.

rok-cesnovar:
Can confirm the suspicion. You can build the model, the problem is `make build` that also builds bin/stansummary, which requires Boost program options. Stansummary is statically linked in. I presume that lib needs `-flto` and the rest of flags?

I am guessing you need to add the flags here: https://github.com/stan-dev/cmdstan/blob/develop/make/command
Let me know if I can help/test anything.

SteveBronder:
@rok-cesnovar what do you think about just adding the makevar variables in this PR then in cmdstan adding these optimization flags? I think doing these flags upstream gives us a larger view on what we want to use flto on. I also don't totally trust windows and would bet a dollar we'll have to do something particular for rtools4.0 and pystan

rok-cesnovar:
Yes, that seems like a good plan. Have them defined in math and use thevin cmdstan seems fine to me.

rok-cesnovar:
Cool. The only thing I am a bit worried about is math performance tests won't have these flags now that they will live in Cmdstan. But I guess that can be handled separately?

SteveBronder:
I thought the performance tests ran using the develop version of cmdstan, is that false?

rok-cesnovar:
I mean if someone will do performance tests with Math-only (with Google benchmark for example), that won't be representative of how it will actually run in the context of Cmdstan. 

rstan/pystan do not use math makefiles anyways so its a different story there.

SteveBronder:
Yeah thats true tho imo I think thats fine. Still technically a regression in the math lib but upstream performance improvement

rok-cesnovar (approved):
LGTM

PR #2024, opened 2020-08-20
Adds views for var_value<Eigen::Matrix> class

SteveBronder:
## Summary

This adds the views `block()`, `row()`, `col()`, `operator()()`, `coeff()` `segment()`, `head()`, `tail()` to `var_value` and `vari_value`.

Some things to note

1. These views are read only and so the returning type and member functions are `const`. 
2. Blocks of eigen matrices need different `Strides` so this adds a new template parameter to `vari_value` that allows the underlying `Eigen::Map` to have an `InnerStride<>` or `OuterStride<>`.
3. Because the mapped type is `const` and we need to specify a stride for the `vari_type` inside of `var_value` there's an additional template `VariType` for `var_value` that allows for the underlying vari_type to be overridden. So for example in the `block()` function we have

```cpp
  inline const auto block(Eigen::Index i, Eigen::Index j, Eigen::Index p, Eigen::Index q) const {
    using vari_sub = decltype(vi_->block(1, 1, 3, 3));
    using var_sub = var_value<const typename vari_sub::PlainObject, const typename vari_sub::vari_type>;
    return var_sub(new vari_sub(vi_->block(i, j, p, q)));
  }
```

where `vari_sub::vari_type` is an `vari_value<Eigen::MatrixXd, Eigen::OuterStride<>>`

I'd really like to get rid of the extra templates here but I'm not sure how else to handle the `InnerStride<>` and `OuterStride<>` for the underlying `Eigen::Map` without them.

## Tests

Tests were added to `var_test` and `vari_test` to check whether subsets are taken correctly.

## Side Effects

Not a side effect but some todo's include

1. Add `var_value<Matrix>` overrides for the functions `head()` etc for the stan language to access
2. Add these subsetting ops to `var_value<SparseMatrix>`. I wanted to wait on that to make sure the pattern here looked okay before I wrote it out for sparse as well.

## Release notes

## Checklist

- [x] Math issue #1805 


stan/math/rev/core/var.hpp
```diff
@@ -290,7 +290,108 @@ class var_value {
...
+  inline var_value<T, VariType>& operator/=(T b);
+
+  /**
+   * A block view of the underlying Eigen matrices.
+   * @param i Starting row of block.
+   * @param j Starting columns of block.
+   * @param p Number of rows to return.
+   * @param q Number of columns to return.
+   */
+  inline const auto block(Eigen::Index i, Eigen::Index j, Eigen::Index p,
```
t4c1, line 302: [optional] These view functions are very similar. I am starting to think about macros..

stan/math/rev/core/var.hpp
```diff
@@ -290,7 +290,108 @@ class var_value {
...
+                              const typename vari_sub::vari_type>;
+    return var_sub(new vari_sub(vi_->col(i)));
+  }
+
+  /**
+   * View element of eigen matrices
+   * @param i Element to access
+   */
+  inline const auto operator()(Eigen::Index i) const {
+    using vari_coeff = decltype((*vi_)(2));
```
t4c1, line 371: [optional] I would prefere if you used actual arguments in these decltypes instead of some random values.
```suggestion
    using vari_coeff = decltype((*vi_)(i));
```

stan/math/rev/core/vari.hpp
```diff
@@ -182,8 +193,9 @@ using vari = vari_value<double>;
  * derivative with respect to the root of the derivative tree.
  *
  */
-template <typename T>
-class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
+template <typename T, typename StrideType>
+class vari_value<T, StrideType, require_eigen_dense_base_t<T>>
```
t4c1, line 197: I don't think adding a stride template is the best option here. I would prefere to have `T` be the actual expression in the `vari_value`. You can use some metaprogramming to convert plain types to maps. This would also allow adding triangular views and transpositions.

stan/math/rev/core/vari.hpp
```diff
@@ -300,12 +314,115 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
...
+  template <typename Checker = T, require_t<std::is_const<Checker>>* = nullptr>
+  inline void set_zero_adjoint_impl() {}
+
   /**
    * Set the adjoint value of this variable to 0.  This is used to
    * reset adjoints before propagating derivatives again (for
    * example in a Jacobian calculation).
    */
-  inline void set_zero_adjoint() final { adj_.setZero(); }
+  inline void set_zero_adjoint() final { set_zero_adjoint_impl(); }
```
t4c1, line 336:
```suggestion
  inline void set_zero_adjoint() final { 
    if(!is_constant<T>::value){
      adj_.setZero();
    }
  }
```
We don't need those impls.

stan/math/rev/core/vari.hpp
```diff
@@ -300,12 +314,115 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
...
+
+  /**
+   * A block view of the underlying Eigen matrices.
+   * @param i Starting row of block.
+   * @param j Starting columns of block.
+   * @param p Number of rows to return.
+   * @param q Number of columns to return.
+   */
+  inline const auto block(Eigen::Index i, Eigen::Index j, Eigen::Index p,
+                          Eigen::Index q) const {
```
t4c1, line 346:
```suggestion
  inline const auto block(Eigen::Index row, Eigen::Index col, Eigen::Index rows,
                          Eigen::Index cols) const {
```

Can you please not use single-letter variable names? I don't mind `i` for loop variables, or `x` and `y` for coordinates or `n` for number, but those `p` and `q` are really not meaningful.

t4c1 (changes requested):

stan/math/rev/core/vari.hpp
```diff
@@ -300,12 +314,115 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
...
+        val_col, adj_col, val_col.outerStride(), adj_col.outerStride());
+  }
+
+  /**
+   * Get coefficient of eigen matrices
+   * @param i Row index
+   * @param j Column index
+   */
+  inline const auto coeff(Eigen::Index i, Eigen::Index j) const {
+    return vari_value<double>(val_(i, j), adj_(i, j));
```
t4c1, line 407: Making a copy of the value and adjoint is not the correct thing to do here, as setting adjoint of the copy will not propagate it back to the matrix.

stan/math/rev/core/vari.hpp
SteveBronder: Yeah I don't like them either but was just following the Eigen names on these. Agree on changing the names tho'

stan/math/rev/core/vari.hpp
SteveBronder: Yes that's intented since you can't assign to subsets of a `var_value<>`. That's also why these are all returning const types

stan/math/rev/core/vari.hpp
```diff
@@ -300,12 +314,115 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
...
+  inline const auto head(Eigen::Index n) const {
+    return vari_value<const PlainObject>(val_.head(n), adj_.head(n));
+  }
+
+  /**
+   * View of the tail of the Eigen vector types.
+   * @param n Number of elements to return from bottom of vector.
+   */
+  inline const auto tail(Eigen::Index n) const {
+    return vari_value<const PlainObject>(val_.tail(n), adj_.tail(n));
```
andrjohns, line 366: Possibly a dumb question (warning).

If I'm understanding the implementation, these views initialise a new ```vari_value``` - which allocates new memory for the values and adjoints to be 'viewed' and copies the values/adjoints into that. Have I got that right?

If so, given that the ```vari_values``` contain pointers to memory for the values/adjoints which is then ```Eigen:Map```-ed, would it be possible to instead create a ```vari_value``` specialisation that is instantiated with pointers to the memory needed? That way the internal ```Eigen::Map``` is directly mapping the memory from the Matrix being 'viewed', rather than allocating new memory and copying into it?

Does that make sense? I haven't entirely grok-ed the static matrix stuff, so I might be off here

stan/math/rev/core/vari.hpp
SteveBronder: Yeah I agree. I'll play around with this today

stan/math/rev/core/var.hpp
SteveBronder: I thought about that but I feel comfortable cargo-culting no macros here. The verbosity is pretty fine and idt this will be repeated in many other places

stan/math/rev/core/vari.hpp
SteveBronder: I think the major bit to note here is the protected constructor for making these views

```cpp
 protected:
  template <typename S, typename K, typename StrideS = Eigen::Stride<0, 0>,
            typename StrideK = Eigen::Stride<0, 0>,
            require_convertible_t<S&, value_type>* = nullptr,
            require_convertible_t<K&, value_type>* = nullptr>
  explicit vari_value(S&& val, K&& adj,
                      StrideS val_stride = Eigen::Stride<0, 0>(),
                      StrideK adj_stride = Eigen::Stride<0, 0>())
      : val_(val.data(), val.rows(), val.cols(), val_stride),
        adj_(adj.data(), adj.rows(), adj.cols(), adj_stride) {}
```

That doesn't copy or allocate any memory. Instead it just constructs a map that is given a copy of the pointer (8 bytes), the dimensions of the array, and the stride (the blocks ops are essentially just maps with a specific stride). Does that make sense? It's a bit weird for sure but I think I nipped this right so we don't make copies or alloc new mem

SteveBronder: Though I think you are correct that we may want another class just to handle the views

stan/math/rev/core/vari.hpp
SteveBronder: How do you feel about this version with a [`vari_view`](https://github.com/stan-dev/math/compare/develop...feature/vari-view-eigen#diff-130c5a75cc427d7d41715e9fca8281f4R177) that's used in the `var_value`. Though I'm having a major problem where in the [var_value test](https://github.com/stan-dev/math/compare/develop...feature/vari-view-eigen#diff-2b0d4604da2786aba9769c587db655c5R169) When I print out the result of `A.block()` and it's rows and columns it matches up to `A_var.block()`. But when I check these with the macro `EXPECT_MATRIX_FLOAT_EQ` it says that the returned `val_` has 0 rows and 0 columns?

stan/math/rev/core/vari.hpp
SteveBronder: This actually still fails. If it was `if constexpr (!is_const<T>::value)` this would compile. We could also just make the adjoint part `mutable` so constness is ignored

stan/math/rev/core/vari.hpp
andrjohns: Ah missed that part, that would do it! I'll pipe down now

SteveBronder:
Ugh I think we updated doxygen and it's catching a new doc error. Gonna look into it

SteveBronder:
Ack this might be a doxygen bug. @serban-nicusor-toptal do you know what version of doxygen Jenkin's uses? If it's from 1.8.14-1.8.17 it might be the bug below. 

https://github.com/doxygen/doxygen/issues/6754

SteveBronder:
It looks like the amici folks were able to work around this and I just asked them what the problem was for them so we could potentially fix it here. 

https://github.com/AMICI-dev/AMICI/issues/507

I'm not sure updating doxygen would work. I'm going to set doxygen warnings to not error out for a bit. I'm on a debian buster setup right now that only has access to doxygen 1.8.13 so can't replicate the error but I think I can replicate this on my home desktop

rok-cesnovar:
I upgraded locally to 1.9.0 and still see the exact same error, so an upgrade will definitely not help.

t4c1:
I think I encountered this doxygen error before. I don't remember exactly what it was, but it was related to different overloads of same function having different names for same argument.

rok-cesnovar:
It does look something like that. Removing ` * @param x Container` from one of the comments removes the warning/error, but that seems bizarre.

```cpp
/**
 * Returns the elementwise `acos()` of the input,
 * which may be a scalar or any Stan container of numeric scalars.
 *
 * @tparam Container type of container
 * @param x Container
 */
template <typename Container,
          require_not_container_st<std::is_arithmetic, Container>* = nullptr>
inline auto acos(const Container& x) {
  return apply_scalar_unary<acos_fun, Container>::apply(x);
}

/**
 * Version of `acos()` that accepts std::vectors, Eigen Matrix/Array objects
 *  or expressions, and containers of these.
 *
 * @tparam Container Type of x
 * @param x Container
 */
template <typename Container,
          require_container_st<std::is_arithmetic, Container>* = nullptr>
inline auto acos(const Container& x) {
  return apply_vector_unary<Container>::apply(
      x, [](const auto& v) { return v.array().acos(); });
}
```

rok-cesnovar:
@serban-nicusor-toptal maybe we should rather downgrade to 1.8.13 as that seems to work fine and this looks like a doxygen bug to me. And we dont need to waste too much time here finding a workaround for us.

Just take https://github.com/doxygen/doxygen/releases/tag/Release_1_8_13 

and run

```
cd doxygen
mkdir build
cd build
cmake -G "Unix Makefiles" ..
make
```
and then `sudo make install`

serban-nicusor-toptal:
Hey @rok-cesnovar good afternoon, I've downgraded the version. Thanks for the quick snippet!
```
:~/doxygen-Release_1_8_13/build$ doxygen --version
1.8.13
```

In case anyone needs to downgrade on ubuntu, turned out I also needed:
```
sudo apt-get install flex
sudo apt-get install bison
```

SteveBronder:
Thanks @rok-cesnovar and @serban-nicusor-toptal!

> I think I encountered this doxygen error before. I don't remember exactly what it was, but it was related to different overloads of same function having different names for same argument.

That's what I thought it was too but I tried matching them all up but was no dice :-/

SteveBronder:
Did we upgrade gtest? It looks like the dist.log throws up a ton of warnings that we are using deprecated functions from gtest (then clang crashed for some reason). If so I can make a PR trying to clean that up

rok-cesnovar:
Yeah we did when adding google benchmark.

rok-cesnovar:
@serban-nicusor-toptal this is also the answer to your question earlier. This ton of warnings seems to creat a LOT of logs and that clogs up the disk space. 

@SteveBronder I can take a look now if you are busy with this PR. Lmk, don't want to duplicate.

SteveBronder:
No worries I'll take a look!

stan/math/rev/core/vari.hpp
SteveBronder: Yes so I agree with this, what I'm trying to figure out right now is that for the var to decide if it holds a view or an actual `vari_value`  I'm doing 

```
  using vari_type = std::conditional_t<is_plain_type<value_type>::value,
                                       vari_value<value_type>, vari_view<T>>;
```

So this is fine and good for `Block` types and other views, but for views into coefficients `value_type`/`T` will be a `double` which is the plain type so this will become a `vari_value<double>`. We don't want to explicitly have a second template parameter to set the `vari_type` to, can you think of another way to handle this?

t4c1: I would make a separate class `var_view` (not `vari`). This also avoids two pointer dereferences (one to get vari and second to get value/adjoint) )when operationg on such a scalar. Also since this class is only used for scalars we don't need to template it.

We could also make this object be implicitly convertible to `var_value<double>`, registering the callback to propagate adjoint on the stack. Although for performance reason we could also not (or at least make it explicit) and explicitly use `var_view` wherever this is used.

stan/math/prim/meta/is_eigen_dense_plain_type.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <stan/math/prim/meta/bool_constant.hpp>
+#include <stan/math/prim/meta/require_helpers.hpp>
+#include <stan/math/prim/meta/is_eigen_dense_base.hpp>
+#include <stan/math/prim/meta/is_plain_type.hpp>
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Checks whether type T is derived from Eigen::DenseBase.
```
t4c1, line 14:
```suggestion
 * Checks whether type T is a plain type derived from Eigen::DenseBase.
```

stan/math/prim/meta/plain_type.hpp
```diff
@@ -40,13 +42,21 @@ struct eval_return_type {
...
+}  // namespace internal
+
 /**
  * Determines plain (non expression) type associated with \c T. For \c Eigen
  * expression it is a type the expression can be evaluated into.
  * @tparam T type to determine plain type of
  */
 template <typename T>
-struct plain_type<T, require_eigen_t<T>> {
+struct plain_type<T, require_t<is_detected<T, internal::plain_type_check_t>>> {
```
t4c1, line 59: What is the reason for this change?

stan/math/prim/meta/is_eigen_dense_plain_type.hpp
t4c1: Also, is this used anywhere?

stan/math/prim/meta/plain_type.hpp
SteveBronder: `EigenBase` types do not have a `PlainObject` so doing something like `plain_type_t<EigenBase<Whatever>>` will fail

stan/math/rev/core/vari.hpp
SteveBronder: Interesting! Okay I'll work on this today

stan/math/prim/meta/plain_type.hpp
t4c1: Where do you need that? Also your code looks to me like it would fail too. Or am I wrong?

SteveBronder: It's needed for the tests with checking `is_plain_type`

Actually tho' for `Eigen::EigenBase<Derived>` this would return back the `Eigen::EigenBase<Derived>`, but I think what we want is a specialization for Eigen types without `PlainType` that calls `plain_type_t` on the derived type

stan/math/prim/meta/is_eigen_dense_plain_type.hpp
SteveBronder: I think I used this in another version and can remove it now

stan/math/rev/core/vari.hpp
SteveBronder: So I had a bit of a hard time getting what you suggested to work. In `var<mat>.coeff` I added a callback to propogate the adjoint of the `var_value<double>` back to the original `var<mat>`. I think this makes since assuming that the two cases this will be used 99% of the time are

1. assigning a cell of a static matrix to a var

```stan
real a = -B[i];
```

2. iterating over a static matrix to assign to the cells in a dynamic matrix

```stan
matrix[N, M] A; // assume static
matrix[N, M] B; // assume dynamic
for (i in 1:N) {
  B[i] = -A[i]; // very inefficient but a thing a user might want to do
}
```

so for both of these we will need to make a var anyway so having a `var_view` would not help. Are there normal cases I'm missing that a `var_view` would be more efficient? If you have any time to take a crack at this I'd be interested in seeing what you have in mind!

stan/math/prim/meta/plain_type.hpp
SteveBronder: I added a test to show the usecase I'm talking about. Though it is a weird edge case

stan/math/prim/meta/plain_type.hpp
```diff
@@ -70,7 +70,7 @@ struct plain_type<T, require_t<is_detected<T, internal::plain_type_check_t>>> {
  */
 template <typename T>
 struct plain_type<T,
-                  require_t<bool_constant<
```
SteveBronder: @t4c1 I had to add this so that it would use the default for kernel expressions. Or should we add a `PlainObject` to kernel expressions?

SteveBronder:
@serban-nicusor-toptal it looks like the distrubtion tests are failing because the jenkins machine ran out of disk storage, do you know how to fix that?

stan/math/rev/core/var.hpp
```diff
@@ -365,35 +369,41 @@ class var_value {
...
-  inline const auto coeff(Eigen::Index i) const {
-    using vari_coeff = decltype((*vi_)(i));
-    using var_coeff = var_value<double>;
-    return var_coeff(new vari_coeff((*vi_)(i)));
+  inline auto coeff(Eigen::Index i) const {
+    using vari_coeff_type = decltype(vi_->coeff(i));
+    auto* vari_coeff = new vari_coeff_type(vi_->coeff(i));
+    reverse_pass_callback([vari_coeff, this, i]() mutable {
+      this->vi_->adj_(i) += vari_coeff->adj_;
+    });
```
t4c1, line 377: I don't like having to do a copy for the reverse pass. I guess it works, so we can leave it for now, but I sure am revisiting this soon.

stan/math/prim/meta/plain_type.hpp
t4c1: This is not an actual use case. We are never using `EigenBase` explicitly, so we can safely ignore that. That class is only used internally in Eigen (probably only becouse they don't have our requires).  Even if we did use it in the way it is used in Eigen, we would always have derived type available to use any metaprograms on.

stan/math/prim/meta/plain_type.hpp
t4c1: > Or should we add a PlainObject to kernel expressions?

No. Adding a require for eigen types here is fine. If/when we do need a `plain_type` for kernel expressions we should add a specialization, not `PlainObject`.

stan/math/rev/core/var.hpp
SteveBronder: Yeah it's a bummer, but I think given the use case it make sense no? Like if most of the time we are using `.coeff()` to assign to a var then I think we need to do something like this right? But I'm fine with it for now so I can go start the plumbing up in the stan repo so we can kick of work on the compiler stuff

SteveBronder:
@t4c1 when you get a min can you take a look at this?

stan/math/prim/meta/plain_type.hpp
```diff
@@ -2,6 +2,8 @@
 #define STAN_MATH_PRIM_META_PLAIN_TYPE_HPP
 
 #include <stan/math/prim/meta/plain_type.hpp>
+#include <stan/math/prim/meta/bool_constant.hpp>
+#include <stan/math/prim/meta/is_detected.hpp>
```
t4c1, line 6: Now that you reverted the rest of the file you also don't need new includes.

stan/math/rev/core/var.hpp
```diff
@@ -35,19 +36,22 @@ static void grad(Vari* vi);
...
-      is_plain_type<T>::value,
-      "The template for this var is an"
-      " expression but a var_value's inner type must be assignable such as"
-      " a double, Eigen::Matrix, or Eigen::Array");
+  /*  static_assert(
+        is_plain_type<T>::value,
+        "The template for this var is an"
+        " expression but a var_value's inner type must be assignable such as"
+        " a double, Eigen::Matrix, or Eigen::Array");
+  */
```
t4c1, line 44: Remove commented code.

stan/math/rev/core/vari.hpp
```diff
@@ -182,27 +220,21 @@ using vari = vari_value<double>;
...
   using eigen_scalar = value_type_t<PlainObject>;  // A floating point type
-  /**
-   * Maps for adj_ and val_
-   */
-  using eigen_map = Eigen::Map<PlainObject, Eigen::Aligned8>;
-  using vari_type = vari_value<T, require_eigen_dense_base_t<T>>;
-  eigen_scalar* val_mem_;  // Pointer to memory allocated on the stack for val_
-  eigen_scalar* adj_mem_;  // Pointer to memory allocated on the stack for adj_
+  //  using check_const = std::conditional_t<std::is_const<T>::value, const
+  //  PlainObject, PlainObject>;
```
t4c1, line 234: remove commmented code

t4c1 (changes requested):
Looks good, there is just some cleanup left to do.

stan/math/rev/core/var.hpp
t4c1: Can you wait a few days with changes in stan repo and compiler? I have an idea I would like to try on this views. I think we can make them assignable, so there is no need for thius type of matrices to be static. This would allow us to eventuallly completely replace `matrix<var>` with `var<matrix>`.

t4c1 (approved):

Issue #2038, opened 2020-08-26
High order functions not being tested in `test_repeat_as_vector`


bbbales2:
## Description

[This](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L640) is calling the first order gradients three times in a row:
```cpp
calculate_gradients_1storder(multiple_gradients1, multiple_lp, x1);
calculate_gradients_1storder(multiple_gradients2, multiple_lp, x1);
calculate_gradients_1storder(multiple_gradients3, multiple_lp, x1);
```

It is comparing against [this](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L616):
```cpp
double single_lp = calculate_gradients_1storder(single_gradients1, logprob, s1);
calculate_gradients_2ndorder(single_gradients2, logprob, s2);
calculate_gradients_3rdorder(single_gradients3, logprob, s3);
```

#### Current Version:
v3.3.0

bbbales2:
Fixed in #2042

Issue #2037, opened 2020-08-26
Probability test framework not computing some gradients correctly


bbbales2:
## Description

There are a bunch of functions that look like ([here](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L320)):
```cpp
double calculate_gradients_1storder(vector<double>& grad, var& logprob,
                                      vector<var>& x) {
    logprob.grad(x, grad);
    stan::math::recover_memory();
    return logprob.val();
  }
```

The gradients are stored in grad.

In some of them the gradients are being stored in the wrong place ([here](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L336)):
```cpp
double calculate_gradients_1storder(vector<double>& grad,
                                      fvar<double>& logprob, vector<var>& x) {
    x.push_back(logprob.d_);
    return logprob.val();
  }
```

This happens the tests for the lpmfs/lpmds, cdfs, lcdfs and lccdfs.

#### Current Version:
v3.3.0

bbbales2:
This was fixed in #2042

PR #2041, opened 2020-08-26
Fix probability framework allocations (Issue #2036)

bbbales2:
Fixes #2036

## Release notes

Fixes some memory allocation problems with test framework

## Checklist

- [x] Math issue #2036 

- [x] Copyright holder: (fill in copyright holder information)

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [ ] the new changes are tested

Issue #2006, opened 2020-08-06
gamma_p and gamma_q gradient problems


bbbales2:
## Description

gamma_p doesn't work great with vars and gamma_q doesn't work great with `fvar<double>`

Here is test code:
```
#include <test/unit/math/test_ad.hpp>
#include <limits>

TEST(mathMixScalFun, gamma_q) {
  using stan::math::var;
  using stan::math::fvar;
  using stan::math::gamma_q;
  using stan::math::gamma_p;
  using stan::math::gamma_lccdf;
  using stan::math::log;
  using stan::math::log1m;

  var yv = 12;
  var alphav = 3;
  var betav = 3;

  fvar<double> yf(12.0, 1.0);
  fvar<double> alphaf(3.0, 1.0);
  fvar<double> betaf(3.0, 1.0);

  var lp1 = log1m(gamma_p(alphav, betav * yv));
  fvar<double> lp2 = log(gamma_q(alphaf, betaf * yf));

  lp1.grad();

  std::cout << "Computed: " << std::endl;
  std::cout << "lp1.val(): " << lp1.val() << std::endl;
  std::cout << "lp2.val(): " << lp2.val() << std::endl;
  std::cout << "y.adj(): " << yv.adj() << std::endl;
  std::cout << "alpha.adj(): " << alphav.adj() << std::endl;
  std::cout << "beta.adj(): " << betav.adj() << std::endl;
  std::cout << "lp2.d_: " << lp2.d_ << std::endl;
}

TEST(mathMixScalFun, gamma_q_ref) {
  using stan::math::var;
  using stan::math::fvar;
  using stan::math::gamma_q;
  using stan::math::gamma_p;
  using stan::math::gamma_lccdf;
  using stan::math::log;
  using stan::math::log1m;

  var yv = 12;
  var alphav = 3;
  var betav = 3;

  fvar<double> yf(12.0, 1.0);
  fvar<double> alphaf(3.0, 1.0);
  fvar<double> betaf(3.0, 1.0);

  var lp1 = gamma_lccdf(yv, alphav, betav);
  fvar<double> lp2 = gamma_lccdf(yf, alphaf, betaf);

  lp1.grad();

  std::cout << "Reference: " << std::endl;
  std::cout << "lp1.val(): " << lp1.val() << std::endl;
  std::cout << "lp2.val(): " << lp2.val() << std::endl;
  std::cout << "y.adj(): " << yv.adj() << std::endl;
[...]

PR #2042, opened 2020-08-26
Fix problems with higher order gradients in probability test framework

### LINKED ISSUE #2036: Probability test framework doesn't clean up memory properly
## Description

Various calculate gradients functions get called in a row, for instance [here](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L540). The problem is that some of these functions call `stan::recover_memory` and so are clearing off the autodiff stack that will get used later ([here](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L369) and [here](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_distr.hpp#L375)).

For instance, when testing `fvar<var>` values, the sequence here recovers memory on the first call and then calls grad in the subsequent two calls:

```
calculate_gradients_1storder(expected_gradients1, logprob_funct, x1);
calculate_gradients_1storder(gradients1, logprob, y1);
calculate_gradients_2ndorder(expected_gradients2, logprob_funct, x2);
```

#### Current Version:
v3.3.0

bbbales2:
Fixes #2036 and #2037 and #2038 

## Release notes

Fix problems with higher order gradients in probability test framework

## Checklist

- [x] Math issue #2036, #2037, #2038


stan/math/prim/prob/frechet_lpdf.hpp
```diff
@@ -81,7 +81,7 @@ return_type_t<T_y, T_shape, T_scale> frechet_lpdf(const T_y& y,
                 T_partials_return, T_y>
       inv_y(size(y));
   for (size_t i = 0; i < stan::math::size(y); i++) {
-    inv_y[i] = 1.0 / value_of(y_vec[i]);
```
bbbales2, line 84: The problem here was this second value_of (inv_y has already had one value_of).

test/prob/binomial/binomial_cdf_test.hpp
```diff
@@ -52,10 +52,10 @@ class AgradCdfBinomial : public AgradCdfTest {
     using std::exp;
     using std::log;
 
-    stan::return_type_t<T_prob> cdf(1);
+    stan::return_type_t<T_prob> cdf(0);
 
     for (int i = 0; i <= n; i++) {
-      cdf *= binomial_coefficient<double>(N, i)
+      cdf += binomial_coefficient<double>(N, i)
```
bbbales2, line 58: This code is the same as the log cdf code without the log: https://github.com/stan-dev/math/blob/develop/test/prob/binomial/binomial_cdf_log_test.hpp

test/prob/gamma/gamma_ccdf_log_test.hpp
```diff
@@ -32,11 +32,11 @@ class AgradCcdfLogGamma : public AgradCcdfLogTest {
     ccdf_log.push_back(
         std::log(1.0 - 0.6321205588285576659757));  // expected ccdf_log
 
-    param[0] = 16.0;  // y
-    param[1] = 3.0;   // alpha
-    param[2] = 3.0;   // beta
+    param[0] = 9.0;  // y
```
bbbales2, line 35: I made the test case easier because the test implementations were struggling numerically. See: https://github.com/stan-dev/math/issues/2006

The Stan implementations of the ccdf didn't seem to have this problem.

test/prob/gamma/gamma_ccdf_log_test.hpp
```diff
@@ -90,10 +90,10 @@ class AgradCcdfLogGamma : public AgradCcdfLogTest {
...
       const T4&, const T5&) {
-    using boost::math::gamma_q;
-    using stan::math::gamma_q;
-    using std::log;
+    using boost::math::gamma_p;
+    using stan::math::gamma_p;
+    using stan::math::log1m;
 
-    return log(gamma_q(alpha, beta * y));
+    return log1m(gamma_p(alpha, beta * y));
```
bbbales2, line 97: Doing log1m(cdf) worked better than log(ccdf) for forward mode. Check here for the lcdf: https://github.com/stan-dev/math/blob/develop/test/prob/gamma/gamma_cdf_log_test.hpp#L84

test/prob/gumbel/gumbel_test.hpp
```diff
@@ -88,6 +88,6 @@ class AgradDistributionGumbel : public AgradDistributionTest {
   stan::return_type_t<T_y, T_loc, T_scale> log_prob_function(
       const T_y& y, const T_loc& mu, const T_scale& beta, const T3&, const T4&,
       const T5&) {
-    return -log(beta) - (y - mu) / beta + exp((mu - y) / beta);
+    return -log(beta) - (y - mu) / beta - exp((mu - y) / beta);
```
bbbales2, line 91: Gumbel pdf here: https://en.wikipedia.org/wiki/Gumbel_distribution

test/prob/inv_chi_square/inv_chi_square_ccdf_log_test.hpp
```diff
@@ -56,6 +56,6 @@ class AgradCcdfLogInvChiSquare : public AgradCcdfLogTest {
                                                     const T5&) {
     using stan::math::gamma_p;
 
-    return log(1.0 - gamma_p(0.5 * nu, 0.5 / y));
+    return log(gamma_p(0.5 * nu, 0.5 / y));
```
bbbales2, line 59: Here's the math version: https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/inv_chi_square_lccdf.hpp#L92

(this is checked with finite differences so I assume it's the correct one)

test/prob/normal/normal_cdf_log_test.hpp
```diff
@@ -24,11 +24,11 @@ class AgradCdfLogNormal : public AgradCdfLogTest {
     parameters.push_back(param);
     cdf_log.push_back(-0.1727537790234499326392);  // expected cdf_log
 
-    param[0] = -2;  // y
-    param[1] = 0;   // mu
-    param[2] = 1;   // sigma
+    param[0] = -2;   // y
```
bbbales2, line 27: I changed this test cause we test the gradients with absolute error, and because the gradients here were like 17.1234, an error on the 4th digit past the decimal meant that I would have to have lowered the tolerance to 1e-3 when this has more precision than that.

This error happened with `fvar<var>` variables. Anyway we should be using relative tolerances where it makes sense and autodiff-mode aware tolerances, but I'm trying to skip that to keep it simple.

Anyway I changed the parameters here and the function works better again.

bbbales2:
@serban-nicusor-toptal had an out of memory disk space error here on the distribution tests: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2042/15/pipeline/40/

Does that mean anything specific or does it just look like a random failiure?

serban-nicusor-toptal:
@bbbales2 Thanks a lot of letting me know, looks like some machines had a different physical disk name under `fdisk -l` ( /dev/xvdb vs /dev/nvme1n1 ) so the machine it failed on didn't mount the disk ( 500 GB ) at all and it used the base one ( 20 GB )
I've now fixed the provisioning script to include both cases and it should be fine from now on.

bbbales2:
@serban-nicusor-toptal I'm glad to here it wasn't just ghosts! Thanks!

bbbales2:
@t4c1 This is ready to review. It's the bugfix part of https://github.com/stan-dev/math/pull/1989 . I'm also putting the allocation stuff from https://github.com/stan-dev/math/pull/2041 here since that separate branch wasn't passing tests (I don't know why).

There are bunch of little small weird changes. I tried to add explanations inline. Ask if it's not clear why something is okay/what happened.

test/prob/gamma/gamma_ccdf_log_test.hpp
```diff
@@ -32,11 +32,11 @@ class AgradCcdfLogGamma : public AgradCcdfLogTest {
...
 
-    param[0] = 16.0;  // y
-    param[1] = 3.0;   // alpha
-    param[2] = 3.0;   // beta
+    param[0] = 9.0;  // y
+    param[1] = 1.2;  // alpha
+    param[2] = 1.2;  // beta
     parameters.push_back(param);
-    ccdf_log.push_back(std::log(1.7116220633718619e-18));  // expected ccdf_log
+    ccdf_log.push_back(-10.221534317077268);  // expected ccdf_log
```
t4c1, line 39: I did not check this number. Where did you get it from?

test/prob/normal/normal_cdf_log_test.hpp
```diff
@@ -24,11 +24,11 @@ class AgradCdfLogNormal : public AgradCdfLogTest {
...
+    param[1] = 0;    // mu
+    param[2] = 1.5;  // sigma
     parameters.push_back(param);
-    cdf_log.push_back(-3.78318433368203166367);  // expected cdf_log
+    cdf_log.push_back(-2.3945773661586438052);  // expected cdf_log
```
t4c1, line 31: I did not check this number. Where did you get it from?

test/prob/test_fixture_cdf.hpp
```diff
@@ -485,18 +490,16 @@ class AgradCdfTestFixture : public ::testing::Test {
                                      const size_t N_REPEAT) {
     if (is_vec) {
       for (size_t i = 0; i < N_REPEAT; i++) {
-        EXPECT_FLOAT_EQ(
-            single_gradients[pos_single] * pow(single_cdf, N_REPEAT - 1),
-            multiple_gradients[pos_multiple])
+        EXPECT_NEAR(single_gradients[pos_single] / N_REPEAT,
+                    multiple_gradients[pos_multiple], 1e-7)
```
t4c1, line 494: I am confused by this change.

test/prob/test_fixture_cdf.hpp
```diff
@@ -485,18 +490,16 @@ class AgradCdfTestFixture : public ::testing::Test {
...
                "failed";
         pos_multiple++;
       }
       pos_single++;
     } else {
-      EXPECT_FLOAT_EQ(N_REPEAT * single_gradients[pos_single]
-                          * pow(single_cdf, N_REPEAT - 1),
-                      multiple_gradients[pos_multiple])
+      EXPECT_NEAR(single_gradients[pos_single],
+                  multiple_gradients[pos_multiple], 1e-7)
```
t4c1, line 502: And this one

test/prob/test_fixture_cdf.hpp
```diff
@@ -533,9 +539,10 @@ class AgradCdfTestFixture : public ::testing::Test {
       add_vars(s2, p0_, p1_, p2_, p3_, p4_, p5_);
       add_vars(s3, p0_, p1_, p2_, p3_, p4_, p5_);
 
-      T_return_type cdf
-          = TestClass.template cdf<Scalar0, Scalar1, Scalar2, Scalar3, Scalar4,
-                                   Scalar5>(p0_, p1_, p2_, p3_, p4_, p5_);
+      T_return_type cdf = pow(
+          TestClass.template cdf<Scalar0, Scalar1, Scalar2, Scalar3, Scalar4,
+                                 Scalar5>(p0_, p1_, p2_, p3_, p4_, p5_),
+          N_REPEAT);
```
t4c1, line 545: And this one.

test/prob/utility.hpp
```diff
@@ -54,13 +54,69 @@ std::ostream& operator<<(std::ostream& os, const vector<var>& param) {
...
+// ------------------------------------------------------------
+
 // default template handles Eigen::Matrix
 template <typename T>
 T get_params(const vector<vector<double>>& parameters, const size_t p) {
   T param(parameters.size());
   for (size_t n = 0; n < parameters.size(); n++)
     if (p < parameters[0].size())
-      param(n) = parameters[n][p];
+      param(n) = get_param<stan::scalar_type_t<T>>(parameters[n], p);
```
t4c1, line 119: Why/where do you need this change?

t4c1 (commented):
I don't see any obvious bugs. However I am hesitant to say this is ok, as I don't know the code. I am also not a mathematician, so there is come confusion regarding math changes. 

Maybe it would be better if some more knowledge about prob testing and math reviewed this? Or I can persist, but there will need to be quite some explaining.

stan/math/prim/prob/frechet_lpdf.hpp
bbbales2: Yeah whichever goes first is fine with me.

test/prob/gamma/gamma_ccdf_log_test.hpp
bbbales2: It's the gamma log ccdf. log ccdf is `log(1.0 - cdf)`.

So in r we can do:
```
log(1.0 - pgamma(9.0, shape = 1.2, rate = 1.2))
```

`beta` in Stan is inverse scale which is the `rate` parameter in R's `pgamma` (https://stat.ethz.ch/R-manual/R-devel/library/stats/html/GammaDist.html). `pgamma` is the CDF of the Gamma pdf in R.

test/prob/normal/normal_cdf_log_test.hpp
bbbales2: This r code:

```
pnorm(-2.0, mean = 0.0, sd = 1.5, log.p = TRUE)
```

test/prob/test_fixture_cdf.hpp
bbbales2: Yeah yeah, this is connected to https://github.com/stan-dev/math/pull/2042/files#r483523418 and https://github.com/stan-dev/math/pull/2042/files#r483523944.

This is comparing the results of a vectorized cdf to the equivalent collection of individual cdf calls. It compares something like:

```
cdf([a, a, a, a])
```

To:
```
cdf(a) * cdf(a) * cdf(a) * cdf(a)
```

So the second thing is equivalent to `pow(cdf(a), N)` where `N` is the length of the vector of `a` values in the vectorized call.

What we need to do is verify that `cdf([a, a, a, a])` computes the same thing as `pow(cdf(a), N)`, which is what happens now.

Previously the pow was done separately and so the gradients tried to handle this with some chain rule stuff. I think this was just wrong and wasn't gonna work with higher order stuff.

test/prob/test_fixture_cdf.hpp
bbbales2: First read: https://github.com/stan-dev/math/pull/2042/files#r485173044

That explains the pow. The other thing is the `/ N_REPEAT`. This has to do with the way `[a, a, a, a]` is constructed. As it is now, all those `a` s are actually different autodiff variables, and the `a` in `pow(cdf(a), N)` is the same, so each `a` in the vectorized case gets a 1/Nth of the adjoint value if that makes sense.

It's effectively done like this in the other files.

test/prob/utility.hpp
bbbales2: This is important for forward mode autodiff when we initialize a vector vs. when we initialize scalars.

If the test scalar type is `fvar<var>`, then the scalars get initialized with `get_param`.

This looks like:

```
param = params[n];
param.d_ = 1.0;
```

Previously, when initializing vectors, the initialization looked like:

```
param(n) = parameters[n][p];
```

I think that leaves the `.d_` initialized to zero, which means forward mode autodiff tests fail. Anyway, now we just use the same code to initialize scalars as to initialize each element of a vector so we don't have these problems.

t4c1: Hmm, do you know why were forward tests passing than before this PR? Actually this PR seems like it fixes quite a number of serious bugs. Any idea why nothing failed before?

stan/math/prim/prob/frechet_lpdf.hpp
t4c1: Now its in and you have a merge conflict.

test/prob/utility.hpp
bbbales2: I'm not sure, and give the seriousness of the bugs I kinda worry there's more.

bbbales2: Well, I know there's more problems lol. There will be a follow up pull request to this that tests that: `cdf([a, b, c, d])` evaluates to the same thing as `cdf(a) * cdf(b) * cdf(c) * cdf(d)`. Lack of those tests caused a bug previously. I also know in the other branch where I put this all together there were problems with discrete_range which didn't show up here. Not sure what the deal is. Maybe adding that other test will uncover them again so I can put the fixes back in?

t4c1 (approved):
This looks ok.

PR #2043, opened 2020-08-26
Fix possible test memory leaks

SteveBronder:
## Summary

For gtest, when we run tests that use var we should make sure that memory is deallocated after each test. We have a little fixture that handles this already

```cpp
struct AgradRev : public testing::Test {
  void SetUp() {
    // make sure memory's clean before starting each test
    stan::math::recover_memory();
  }
};
```

But to have the [fixture](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests) clean up memory we need to use `TEST_F(AgradRev, *` instead of just `TEST`. So this PR just moves `AgradRev` to a util file, adds a `TearDown()` method that recover's memory after each test, then I did a `grep` for any `TEST(AgradRev*,` and replaced them with `TEST_F(AgradRev, *)`

All the tests in rev/mix that use `TEST` should probably use this, but for now I just got the low hanging fruit that I could do with a find/replace

## Tests

None

## Side Effects

Tests will take a little longer since we clean up after but idt much.

## Checklist


bbbales2:
> For gtest, when we run tests that use var we should make sure that memory is deallocated after each test

Recover memory doesn't actually deallocate if that technicality matters: https://github.com/stan-dev/math/blob/develop/stan/math/memory/stack_alloc.hpp#L194

You could also look for `TEST(` (whether they say Agrad or not) in rev and convert them all.

t4c1:
> Side effects
Tests will take a little longer since we clean up after but idt much.

I think the tests will actually be a bit faster, as we reuse memory instead of allocating more and more. Although the difference is probably negligible.

bbbales2:
> bbbba23

When I first saw that I thought that was my name or something. Anyway four bs in a row nice

bbbales2:
@SteveBronder  This has merge conflicts

rok-cesnovar:
@bbbales2 did #2146 make this sort of redundant?

SteveBronder:
Eh sort of. It at least made it not as important. I'm going to close this now until I have time to come back to it

Issue #2045, opened 2020-08-27
Create var_value conversion functions and helper metaprograms for stanc3


rok-cesnovar:
## Description

We discussed `var_value` stanc3 stuff this morning with @t4c1 and a couple of stuff popped up that would need to be done to simplify the transition to var_value in stanc3.

- [ ] conversion function between `var_value<Matrix<double>>` and `Eigen<var>`
- [ ] UDF return type helper metaprogram. Example:

```cpp
template <typename T0__, typename T1__, typename T2__>
std::conditional_t<
  is_any_var_vt<is_eigen, T0__, T1__, T2__>::value, 
  var_value<Eigen::Matrix<double, -1, -1>>
  Eigen::Matrix<stan::promote_args_t<T0__, T1__, T2__>, -1, -1>
>
foo(const T0__& a,
    const T1__& b,
    const T2__& c, std::ostream* pstream__) {
```
meaning that if any input is var_value the return will be a var_value (if matrix/vector/row_vector).

#### Current Version:
v3.3.0

rok-cesnovar:
Do we have a metaprogram that does something like:

If T is double (or not-var essentially, but always a "scalar type"), then set the type as Eigen<T>, otherwise, set it to var_value<Eigen<double>>

t4c1:
This is for local variables in log_prob for the case it is called with non-var arguments, right? I think we can go without such a metaprogram by having compiler generate `using var_value_vector = std::conditional_t<is_var<Argument>::value, var_value<Eigen::VectorXd>>, Eigen::VectorXd>;` (and same for matrix and row vector) and than use this when it determines `var_value` should be used.

rok-cesnovar:
> This is for local variables in log_prob for the case it is called with non-var arguments, right?

Yeah, any non-data variable in a model. Those are all defined in log_prob yes. 

Your idea sounds good to me. Will close and add your comment to the stanc3 issue.

PR #2047, opened 2020-08-27
Add var_value utils

t4c1:
## Summary

Adds some `var_value` related utilities. These include functions for conversion between `var_value` and Eigen matrix of vars and a metaprogram that determines return type of a function that accepts a mixture of `var_value`s and Eigen matrices and propagates `var_value` if present.

## Tests

New functions and metaprogram are tested.

## Side Effects

None.

## Release notes
Added functions for conversion between `var_value` and Eigen matrix of vars and a metaprogram that determines return type of a function that accepts a mixture of `var_value`s and Eigen matrices and propagates `var_value` if present.

## Checklist

- [x] Math issue #2045 


stan/math/rev/fun/to_var_value.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/rev/functor/arena_matrix.hpp>
+#include <stan/math/rev/functor/reverse_pass_callback.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Converts an Eigen matrix (or vector or row_vector) or expression of `var`s
+ * into `var_value`. Adjoint is propagated back to arguument in the reverse
```
SteveBronder, line 14:
```suggestion
 * into `var_value`. Adjoint is propagated back to argument in the reverse
```

stan/math/rev/fun/to_var_value.hpp
```diff
@@ -0,0 +1,34 @@
...
+template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+var_value<Eigen::Matrix<double, T::RowsAtCompileTime, T::ColsAtCompileTime>>
+to_var_value(const T& a) {
+  arena_matrix<plain_type_t<T>> a_arena = a;
+  var_value<Eigen::Matrix<double, T::RowsAtCompileTime, T::ColsAtCompileTime>>
+      res(a_arena.val());
+  reverse_pass_callback(
+      [res, a_arena]() mutable { a_arena.adj() += res.adj(); });
+  return res;
+}
```
SteveBronder, line 29: We've mostly been working with the Matrix type but do you think it would make sense to have this work with `promote_scalar_t` so it can take in `Array` types as well like

```suggestion
var_value<promote_scalar_t<double, T>>
to_var_value(const T& a) {
  arena_matrix<plain_type_t<T>> a_arena = a;
  var_value<promote_scalar_t<double, T>> res(a_arena.val());
  reverse_pass_callback(
      [res, a_arena]() mutable { a_arena.adj() += res.adj(); });
  return res;
}
```

stan/math/rev/fun/from_var_value.hpp
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Converts `var_value` into an Eigen Matrix. Adjoint is propagated back to
+ * arguument in the reverse pass.
+ *
+ * @tparam T type of the input
+ * @param a matrix to convert
+ */
+template <typename T, require_var_vt<is_eigen, T>* = nullptr>
+Eigen::Matrix<var, value_type_t<T>::RowsAtCompileTime,
+              value_type_t<T>::ColsAtCompileTime>
```
SteveBronder, line 21: `Rows/ColsAtCompileTime` are available in `vari_value`. Let's add them to `var_value` so then we can remove the value_type_t here.

I'm pretty sure all we need to do is add a `static constexpr int Rows/ColsAtCompileTime{0}` to the `vari_value<double>` class then the `var_value` can just use

```cpp
static constexpr int RowsAtCompileTime{vari_type::RowsAtCompileTime};
static constexpr int ColsAtCompileTime{vari_type::ColsAtCompileTime};
```

stan/math/rev/fun/from_var_value.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/rev/functor/arena_matrix.hpp>
+#include <stan/math/rev/functor/reverse_pass_callback.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Converts `var_value` into an Eigen Matrix. Adjoint is propagated back to
+ * arguument in the reverse pass.
```
SteveBronder, line 14:
```suggestion
 * argument in the reverse pass.
```

stan/math/rev/meta/rev_matrix_type.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace stan {
+
+/**
+ * Determines a return type for a function that addepts given inputs and wants
+ * to return a matrix(or vector or row vector) with given compile time number of
+ * rows and columns.
+ *
+ * Returns `var_value` if any of the inputs is `var_value`, otherwise a matrix
+ * of `var` or `double`.
```
SteveBronder, line 19:
```suggestion
 * Determines a return type for a function that accepts given inputs and wants
 * to return a matrix (or vector or row vector) with given compile time number of
 * rows and columns.
 *
 * Returns `var_value` if any of the inputs are `var_value`, otherwise returns an Eigen matrix
 * of `var` or `double`.
```

test/unit/math/rev/fun/to_var_value_test.cpp
```diff
@@ -0,0 +1,45 @@
+#include <stan/math/rev.hpp>
+#include <test/unit/util.hpp>
+#include <gtest/gtest.h>
+
+TEST(AgradRevMatrix, to_var_value_matrix_test) {
+  Eigen::MatrixXd val(2, 3);
+  val << 1, 2, 3, 4, 5, 6;
+  Eigen::MatrixXd adj(2, 3);
+  val << 4, 5, 6, 7, 8, 9;
+  Eigen::Matrix<stan::math::var, Eigen::Dynamic, Eigen::Dynamic> var = val;
```
SteveBronder, line 10: [nbd] I'd make this var_mat instead of just `var`

stan/math/rev/fun/to_var_value.hpp
t4c1: fixed

stan/math/rev/fun/to_var_value.hpp
t4c1: That makes sense. Changed.

stan/math/rev/fun/from_var_value.hpp
t4c1: Done.

stan/math/rev/fun/from_var_value.hpp
t4c1: fixed

stan/math/rev/meta/rev_matrix_type.hpp
t4c1: I am typing too fast ... Fixed.

test/unit/math/rev/fun/to_var_value_test.cpp
t4c1: fixed

t4c1:
> We could do this here or in another PR, but I think we should have a var_value constructor for var_value<Matrix>(Matrix<var>). 

Now that we have a conversion function I am not sure we need it.

> I'm 50/50 on whether we should have an operator Matrix<var> for implicitly converting from var_value<mat> to mat<var>

I am against that. There is a reason why implicit conversions are in general undesirable. We are only using them where removing them would bring significant downsides.

stan/math/rev/fun/from_var_value.hpp
```diff
@@ -0,0 +1,31 @@
...
+ */
+template <typename T, require_var_vt<is_eigen, T>* = nullptr>
+Eigen::Matrix<var, T::RowsAtCompileTime, T::ColsAtCompileTime> from_var_value(
+    const T& a) {
+  arena_matrix<Eigen::Matrix<var, T::RowsAtCompileTime, T::ColsAtCompileTime>>
+      res(a.val());
+  reverse_pass_callback([res, a]() mutable { a.vi_->adj_ += res.adj(); });
+  return res;
+}
+
```
SteveBronder, line 27: [optional/questionable] Should we have a version that's a no-op for a `Matrix<var>`?

stan/math/rev/fun/to_var_value.hpp
```diff
@@ -0,0 +1,33 @@
...
+template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+var_value<Eigen::Matrix<double, T::RowsAtCompileTime, T::ColsAtCompileTime>>
+to_var_value(const T& a) {
+  arena_matrix<plain_type_t<T>> a_arena = a;
+  var_value<promote_scalar_t<double, T>> res(a_arena.val());
+  reverse_pass_callback(
+      [res, a_arena]() mutable { a_arena.adj() += res.adj(); });
+  return res;
+}
+
```
SteveBronder, line 29: Same as above about no-op version. In the same vein though we could have a signature with `Matrix<double>` that just constructs a `var<Mat>`. I'm not really sure what the compiler will need

stan/math/rev/meta/rev_matrix_type.hpp
```diff
@@ -0,0 +1,38 @@
...
+struct rev_matrix_type {
+  using type = std::conditional_t<
+      math::disjunction<math::conjunction<
+          is_var<Inputs>, is_eigen<value_type_t<Inputs>>>...>::value,
+      math::var_value<Eigen::Matrix<double, Rows, Cols>>,
+      Eigen::Matrix<return_type_t<Inputs...>, Rows, Cols>>;
+};
+
+template <int Rows, int Cols, typename... Inputs>
+using rev_matrix_t = typename rev_matrix_type<Rows, Cols, Inputs...>::type;
```
SteveBronder, line 34: Why not `var_matrix_t<>`?

SteveBronder: In #2050 I have a `is_var_matrix<>` that checks whether a type is either `var<Mat>` or `Mat<var>` so it feels like this is in the same family

stan/math/rev/meta/rev_matrix_type.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * @tparam Inputs types of arguments the function accepts
+ */
+template <int Rows, int Cols, typename... Inputs>
+struct rev_matrix_type {
+  using type = std::conditional_t<
+      math::disjunction<math::conjunction<
+          is_var<Inputs>, is_eigen<value_type_t<Inputs>>>...>::value,
+      math::var_value<Eigen::Matrix<double, Rows, Cols>>,
+      Eigen::Matrix<return_type_t<Inputs...>, Rows, Cols>>;
+};
```
SteveBronder, line 31: So it looks like it's decision time on whether functions that can operator on mixes of `var<mat>` and `mat<var>` should return a `var<mat>` or `mat<var>`. 

I agree with the more aggressive version here where `var<mat>` is returned by default. For most problems it feels like having the reverse mode pass using `var<mat>` would help offset the copy cost when we need to go back and forth between the two for some functions. @bbbales2 @rok-cesnovar do yinz have opinions on this as well?

bbbales2: I am very for returning `var_value<Eigen>` as much as possible.

If `matrix[N, M]` in the language stays corresponding to an `Eigen::Matrix<var, R, C>` with no compiler fanciness, then we'll induce in reverse mode an expensive copy every time we copy a `var_value` to one of these.

What makes this expensive is that to do propagate the adjoints back we need to:
1. Copy out the adjoints of the `Eigen::Matrix<var, R, C>` into the adjoint of the `var_value<Matrix>`
2. Read the adjoint of the `var_value<Matrix>` and propagate stuff back from there

This is compared to the cheap operation:
1. Read the adjoints of `Eigen::Matrix<var, R, C>` and propagate stuff back from there.

Saw this amusing thing the other day -- I think it applies: [cigarettes and cocaine](http://observationalepidemiology.blogspot.com/2020/08/a-cigarettes-and-cocaine-argument.html). We're cutting the cigarettes with this optimization -- the pointer chasing in reading the adjoints of `Eigen::Matrix<var, R, C>` is the expensive thing and we're doing it either way.

And the upside here is that returning `var_value<Matrix>` variables, we'll have more of them floating around and can get efficiencies in basically any situation other than assigning to `Eigen::Matrix<var, R, C>` types.

bbbales2:
> [Steve] I'm 50/50 on whether we should have an operator Matrix for implicitly converting from var_value to mat

> [Tadej] I am against that. There is a reason why implicit conversions are in general undesirable.

If we can get away without them, I agree that is cool. I guess that has to happen at the language level somehow?

stan/math/rev/meta/rev_matrix_type.hpp
SteveBronder: Word then I think this code is A-OK!

stan/math/rev/fun/from_var_value.hpp
t4c1: I don't see the need right now. If we ever do need it, it is simple to add.

stan/math/rev/meta/rev_matrix_type.hpp
t4c1: I could do that, but `var_matrix` sounds to me like it should be `var_value<matrix>`. Maybe you can change your PR instead to be consistent?

t4c1:
> I guess that has to happen at the language level somehow?

Not at language level, but the compiler will track which types are used where and insert conversion functions where needed.

stan/math/rev/meta/rev_matrix_type.hpp
SteveBronder: Yeah I had trouble naming it. `is_rev_matrix<>`, I think that can make sense I'll change mine

PR #2034, opened 2020-08-25
Var value normal id glm

t4c1:
## Summary

Changes GPU implementation `normal_id_glm_lpdf` to accept `var_value` inputs, allowing diferentiation of `x` and `y`.
New interface still requires `x` and `beta` to be `matrix_cl`s. All other arguments can be either `matrix_cl`s or scalars. (Anything that can be `matrix_cl` can also be `var_value<matrix_cl>`.)

This includs a number of smaller changes:
- Added support for `matrix_cl` to various functions and metaprograms: `is_constant`, `size`, `sum`, `arena_t` (returning a class derived from `matrix_cl` and `chainable_alloc`), `operands_and_partials`, `to_arena`, `dims`
- Reworked `is_matrix_cl` to also return `true` for classes derived from `matrix_cl`
- Some kernel generator expressions now also accept scalars (checks, colwise reductions) and references to scalars (`operator*`)
- A number of files now only include headers they need instead of more general ones
- `elt_divide` now also accepts two scalar arguments
- bugfixed `is_stan_scalar` to return `false` for `var_value<matrix>`
- replaced `+=` in precomputed_gradients with `=` and `+` to make it work with `matrix_cl`
- added `.rows()` and `.cols()` to `var_value` 
- `is_matrix_cl` is moved to prim/meta

## Tests
Modified the tests to test the new interface of `normal_id_glm_lpdf`.

## Side Effects
Interface of GPU `normal_id_glm_lpdf` is completly changed - previous signatures (accepting EIgen types and std vectors) are not accepted anymore.

## Release notes
GPU implementation `normal_id_glm_lpdf` can also be used when derivatives of `x` or `y` are needed.

## Checklist

- [x] Math issue #2013

- [x] Copyright holder: Tadej Ciglarič

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

[...]

PR #2050, opened 2020-08-28
Add var<mat> softmax and function for testing mat<var> and var<mat> methods

### LINKED ISSUE #2048: softmax setting adjoint instead of incrementing adjoint
## Description

@SteveBronder found that softmax isn't quite working right. The reverse mode softmax code here sets the adjoint instead of incrementing it: https://github.com/stan-dev/math/blob/develop/stan/math/rev/fun/softmax.hpp#L39

Changing the `=` to `+=` fixes the problem.

I think this could be tested in the autodiff test framework. The test would be set the adjoints of the inputs to something non-zero, do the reverse mode pass, and then check the adjoints have been incremented by what you'd expect if the adjoints had been zero.

If this is fixed without the more general fix to the autodiff tester, make a new issue for that.

#### Current Version:
v3.3.0

SteveBronder:
## Summary

This pull request adds a `var_value<Matrix>` overload for `softmax()`, though tbh it's more about how to test that a `Matrix<var>` and `var<Matrix>` return the same values.

Some other smaller things

- Fixes #2048 by adding += to reverse mode accumulation on adjoint
- makes `adj_` mutable in `vari_value<Matrix>` so that `var_value<Matrix>.adj()` can be assigned to even if the object is passed to by a const&.
- Fixes `softmax()` so instead of throwing for zero sized matrices it just passes back a zero sized matrix (I believe this is our intended behavior)
- Adds `rows()`, `cols()` and `size()` methods to var_value to work with simple loops that go over the inner container.
- Adds a `sum()` method to work with `var_value<Matrix>` types
-

## Tests

Since all we want to do is check that both functions inputs/outputs have the same value and adjoint, I wrote a little function called `expect_ad_matvar`. All it does is run `f(var_value<Matrix>)` and `f(Matrix<var>)`, then sums the results, and takes the gradient of the sum. Then we just check that the return object and input object's values and adjoints are equal for both types. If anyone has better testing suggestions I'm def all ears

## Side Effects

None

## Release notes

Adds softmax function for static matrices and testing scheme for checking array of structs vs struct of arrays matrix var implimentation.

## Checklist

- [x] Math issue #1805 


stan/math/rev/fun/softmax.hpp
```diff
@@ -24,24 +24,46 @@ namespace math {
...
+ * @param alpha Unconstrained input vector.
+ * @return Softmax of the input.
+ * @throw std::domain_error If the input vector is size 0.
+ */
+template <typename Mat, require_var_vt<is_eigen, Mat>* = nullptr>
+inline auto softmax(const Mat& alpha) {
+  if (alpha.size() == 0) {
+    return alpha;
+  }
+  Mat res(softmax(value_of(alpha)));
```
bbbales2, line 59: Looks good. I think this should use the same code as the other version though.

This is the other version:
```cpp
arena_matrix<Eigen::VectorXd> res_val = softmax(value_of(alpha));
arena_matrix<Eigen::Matrix<var, Eigen::Dynamic, 1>> res = res_val;
arena_matrix<Eigen::Matrix<var, Eigen::Dynamic, 1>> alpha_arena = alpha;
```

If the input argument is type `const T&`, we can do something like:
```cpp
const arena_matrix<Eigen::VectorXd> res_val = softmax(value_of(alpha));
arena_t<T> res = res_val;
arena_t<T> alpha_arena = alpha;
```

And for the `var_value<Eigen>` version `arena_t` just returns `var_value` types. The extra trick to avoid copying is that:

```cpp
var_value<Eigen::VectorXd> res = res_val;
```

should recognize that `res_val` is a const arena_matrix and so it's safe to just take the value pointer in. We're using that assumption in this constructor: https://github.com/stan-dev/math/blob/develop/stan/math/rev/functor/arena_matrix.hpp#L71 for instance

stan/math/rev/core/vari.hpp
```diff
@@ -200,7 +209,8 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
   using eigen_map = Eigen::Map<PlainObject, Eigen::Aligned8>;
   using vari_type = vari_value<T, require_eigen_dense_base_t<T>>;
   eigen_scalar* val_mem_;  // Pointer to memory allocated on the stack for val_
-  eigen_scalar* adj_mem_;  // Pointer to memory allocated on the stack for adj_
+  mutable eigen_scalar*
+      adj_mem_;  // Pointer to memory allocated on the stack for adj_
```
t4c1, line 213: Do we ever need to change a value of a vari?

stan/math/rev/core/vari.hpp
```diff
@@ -88,6 +88,15 @@ class vari_value<T, require_floating_point_t<T>> : public vari_base {
...
   value_type adj_;
 
+  /**
+   * Rows at compile time
+   */
+  static constexpr int RowsAtCompileTime{0};
+  /**
+   * Columns at compile time
+   */
+  static constexpr int ColsAtCompileTime{0};
```
t4c1, line 98: Value of 1 would make more sense for scalars.

t4c1 (changes requested):

stan/math/rev/fun/softmax.hpp
SteveBronder: > And for the var_value<Eigen> version arena_t just returns var_value types. The extra trick to avoid copying is that:

Just to make sure we are on the same page, are you saying that we should have an `arena_matrix<Eigen::Matrix<double>>` constructor for `var_value` and `vari_value` that just takes in the `arena_matrix<Eigen::Matrix<double>>` and assigns the memory it's pointing to the `vari_value`'s `val_mem`? I think that makes sense to me

stan/math/rev/core/vari.hpp
SteveBronder: Good catch! No idt we will ever need to change the pointer directly here just the underlying values

stan/math/rev/fun/softmax.hpp
bbbales2: I think it would be a `const arena_matrix<Eigen::Matrix<double>>&` constructor, but yeah, it can use the same memory since it's constant.

stan/math/rev/fun/sum.hpp
```diff
@@ -98,6 +98,22 @@ inline var sum(const EigMat& m) {
...
+ * Returns the sum of the coefficients of the specified
+ * `var_value<matrix/column_vector/row vector>`.
+ *
+ * @tparam T type of the matrix of vector (Must be derived from \c
+ * Eigen::MatrixBase and contain \c var scalars)
+ * @param x Specified var_value containing a matrix or vector.
+ * @return Sum of coefficients of matrix.
+ */
+template <typename T, require_eigen_t<T>* = nullptr>
+inline var sum(const var_value<T>& x) {
```
bbbales2, line 111: Can you make this work for `Eigen::Matrix<var, R, C>` too? Also convert the `std::vector` specialization?

SteveBronder: I couldn't figure out the how to get this to work for std::vector as well since this can also take in a regular matrix. Though I shrunk it down to work on `var<mat>` and `mat<var>`

bbbales2: The trick is `as_array_or_scalar`. Check out the variance implementation [here](https://github.com/stan-dev/math/blob/62dfc61752298a5ebc2a5b3a7e74e6eecb40f086/stan/math/rev/fun/variance.hpp#L24).

It uses this: https://mc-stan.org/math/dc/d25/as__array__or__scalar_8hpp_source.html

bbbales2: I guess `var<mat>` might not work with `as_array_or_scalar`/`as_column_vector_or_scalar` yet. But those should be updated to support `var<mat>` at some point cuz they're super handy functions.

SteveBronder: Let's do that in another PR. I'd like to have this up since once we have at least a function that uses `var<mat>` and #2024 goes in I can add the indexing stuff for `var<mat>` and we can start doing the work to have `var<mat>` used in the compiler

bbbales2: Ofc, ofc, makes sense to do that separately

stan/math/rev/core/vari.hpp
```diff
@@ -221,7 +232,7 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
    * The adjoint of this variable, which is the partial derivative
    * of this variable with respect to the root variable.
    */
-  eigen_map adj_;
+  mutable eigen_map adj_;
```
t4c1, line 235: Where do you need to change adjoint map?

SteveBronder: Oh before when I wasn't using `arena_matrix` just doing `alpha.adj() += ...` was giving me an eigen error about mutating a `const` type. But putting everything in one and using an `arena_matrix` I don't see that anymore so I'll remove this

stan/math/rev/core/var.hpp
```diff
@@ -313,6 +316,16 @@ class var_value {
...
 
+template <typename MatrixType>
+class arena_matrix;
+
+template <typename MatrixType>
+class arena_matrix<var_value<MatrixType>> : public var_value<MatrixType> {
+ public:
+  template <typename T>
+  arena_matrix(T&& x) : var_value<MatrixType>(std::forward<T>(x)) {}  // NOLINT
+};
```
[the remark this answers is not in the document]
SteveBronder: So without this constructor I'm getting a conversion error since `arena_matrix` doesn't understand how to convert between a `var<matrix>` and `arena_matrix<var<matrix>>`.

```cpp
./stan/math/rev/fun/softmax.hpp:33:27: error: conversion from ‘const stan::math::var_value<Eigen::Matrix<double, -1, 1> >’ to non-scalar type ‘stan::math::arena_matrix<stan::math::var_value<Eigen::Matrix<double, -1, 1> > >’ requested
   arena_matrix<mat_plain> alpha_arena = alpha;
                           ^~~~~~~~~~~
./stan/math/rev/fun/softmax.hpp:35:27: error: conversion from ‘arena_matrix<Eigen::Matrix<double, -1, 1>>’ to non-scalar type ‘arena_matrix<stan::math::var_value<Eigen::Matrix<double, -1, 1> >>’ requested
   arena_matrix<mat_plain> res = res_val;
```

Since this adds the implicit `operator=` for converting a `var<matrix>` to an `arena_matrix<var<matrix>>` and everything is actually handled by `var_value<Matrix>`. `var<matrix>` also has constructors to take in `arena_matrix<var<matrix>>` so we don't copy mem

SteveBronder:
@bbbales2 can you take a look at this?

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: You should only use arena matrix for Eigen types. `var<matrix>` is not an Eigen type so you should use `arena_t` metaprogram to determine appropriate type for arena storage (which turns out to be `var<matrix>` itself).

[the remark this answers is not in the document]
SteveBronder: Oh that's a much nicer way to do it!

bbbales2:
Sure, I'll try to do it tomorrow.

stan/math/rev/fun/softmax.hpp
```diff
@@ -24,19 +24,20 @@ namespace math {
...
-
-  reverse_pass_callback([=]() mutable {
+template <typename Mat, require_rev_matrix_t<Mat>* = nullptr>
+inline plain_type_t<Mat> softmax(const Mat& alpha) {
+  using mat_plain = plain_type_t<Mat>;
+  if (alpha.size() == 0) {
+    return alpha;
+  }
+  arena_t<mat_plain> alpha_arena = alpha;
+  arena_t<Eigen::VectorXd> res_val = softmax(value_of(alpha_arena));
```
bbbales2, line 34: `Eigen::VectorXd` -> `mat_plain`

test/unit/math/test_ad.hpp
```diff
@@ -1432,6 +1433,54 @@ void expect_ad_vectorized_binary(const F& f, const T1& x, const T2& y) {
   expect_ad_vectorized_binary(tols, f, x, y);
 }
 
+template <typename T1, typename T2,
+          require_all_stan_scalar_t<T1, T2>* = nullptr>
+void expect_float_equal(const T1& x, const T2& y) {
+  EXPECT_FLOAT_EQ(x.val(), y.val());
```
bbbales2, line 1439: This needs to be more than `EXPECT_FLOAT_EQ`. Check here: https://github.com/stan-dev/math/blob/develop/test/unit/math/test_ad.hpp#L122

test/unit/math/test_ad.hpp
```diff
@@ -1432,6 +1433,54 @@ void expect_ad_vectorized_binary(const F& f, const T1& x, const T2& y) {
...
+ *
+ * @tparam F A lambda or functor type that calls the unary function.
+ * @tparam T An Eigen matrix type
+ * @param f a lambda
+ * @param x An Eigen matrix.
+ *
+ */
+template <typename F, typename T>
+void expect_ad_matvar(F&& f, const T& x) {
+  using stan::plain_type_t;
```
bbbales2, line 1464: This checks that `var<matrix>` and `matrix<var>` return the same values and give the same gradients.

I think it should also check that it has the same throws as the other stuff.

In the future we'll need multiple argument testing as well (which means some args are autodiff and some are primitive). Are you going to do that in another pull? We should probably test mixes of `var<mat>` `mat<var>` in that case too.

Right now I think it makes sense that the entry-point to this testing is separate from the rest of the testing framework, but the Spirit of the Testing Framework is that all tests run automatically, so I think this should become part of the regular `expect_ad` function eventually.

stan/math/rev/core/vari.hpp
```diff
@@ -247,6 +259,17 @@ class vari_value<T, require_eigen_dense_base_t<T>> : public vari_base {
     ChainableStack::instance_->var_stack_.push_back(this);
   }
 
+  template <typename S, require_convertible_t<S&, T>* = nullptr,
+            require_arena_matrix_vt<std::is_arithmetic, S>* = nullptr>
+  explicit vari_value(const S& x)
```
bbbales2, line 264: S should be a const arena_matrix for this to happen, otherwise the memory might accidentally get changed after the function is called.

It makes sense to modify some `arena_matrix` variables after construction and in an elementwise way, so we should only take the memory if it is const.

Do we need `is_arithmetic` here?

stan/math/rev/fun/sum.hpp
```diff
@@ -1,11 +1,14 @@
 #ifndef STAN_MATH_REV_FUN_SUM_HPP
 #define STAN_MATH_REV_FUN_SUM_HPP
 
+#include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/rev/core.hpp>
 #include <stan/math/rev/meta.hpp>
+#include <stan/math/rev/functor/arena_matrix.hpp>
```
bbbales2, line 7: This should be in `core` not `functor`. Mind making the change here? Otherwise make an issue and we'll do it somewhere else.

bbbales2 (changes requested):
Review!

stan/math/rev/fun/softmax.hpp
t4c1: This is a value, `mat_plain` contains vars ir is var_value.

stan/math/rev/fun/sum.hpp
t4c1: Should it? It is not a feature, it is impossible to make rev implementations without. I guess it is more like a useful helper, like `reverse_pass_callback`.

bbbales2: Oh. Good point. I'd always assumed what went in the 'functor' folder was anything that took in another function. But C++ functors are just objects that act like functors, and I don't really know what the math definition of functor means, so now I'm confused.

I just put things in core things that we use to make the rest of the autodiff work. This is one of those things. I'd probably put reverse_pass_callback in there too. If it seems confusing, let's just leave it where it is and bring this up at the next math meeting if anyone remembers.

t4c1: Hmm, I guess I just assumed everything that is not directly a function and not essential for the rest to go into functor (Arena matrices are not really functors. Same for many other things in functor). It might be a good idea to discuss folder structure in general on the meeting.

SteveBronder: imo I would like this to be in core and can move it over in this PR. Mostly on the basis that we will be using it in a lot of places and so conceptually it's sort of a core part of the autodiff

test/unit/math/test_ad.hpp
SteveBronder: > This checks that var<matrix> and matrix<var> return the same values and give the same gradients.

> I think it should also check that it has the same throws as the other stuff.

Agree adding that now!

> In the future we'll need multiple argument testing as well (which means some args are autodiff and some are primitive). Are you going to do that in another pull? We should probably test mixes of var<mat> mat<var> in that case too.

Yeah I'm building these as we need them. I don't want to try to future proof now without examples

> Right now I think it makes sense that the entry-point to this testing is separate from the rest of the testing framework, but the Spirit of the Testing Framework is that all tests run automatically, so I think this should become part of the regular expect_ad function eventually.

Once we have a binary op in I'm going to start looking at how to do this. I wasn't sure how to automatically do this at first but I have an idea now. We can just detect whether the signature `f(var<mat>)` exists and only run these tests in that instance

stan/math/rev/core/vari.hpp
SteveBronder: > S should be a const arena_matrix for this to happen, otherwise the memory might accidentally get changed after the function is called.

> It makes sense to modify some arena_matrix variables after construction and in an elementwise way, so we should only take the memory if it is const.

Do you mean the arena matrix could point to new memory as in the pointer to memory will change in the `arena_matrix`? tbh I'm not sure if that's a big thing to worry about since this is a copy op. 

> Do we need is_arithmetic here?

Yes since we treat the arithmetic data as the values

```cpp
      : val_mem_(x.data()),
        adj_mem_(ChainableStack::instance_->memalloc_.alloc_array<eigen_scalar>(
            x.size())),
```

We would need a different specialization for `arena_matrix<Matrix<var>>` that would require hard copies of both the `val_` and `adj_` mem

stan/math/rev/fun/sum.hpp
SteveBronder: tbh I'd rather actually move `reverse_pass_callback` to core as well. It's a weird semantic but I think of `core` as like "This will be used everywhere to implement reverse mode" where `functor` is more for operating over those implementations. `reverse_pass_callback` also only relies on core

bbbales2:
@SteveBronder Just looking at this pull: https://github.com/stan-dev/math/pull/2034

```
bugfixed is_stan_scalar to return false for var_value<matrix>
...
added .rows() and .cols() to var_value
```

Those might end up conflicting with stuff over here, I forget.

SteveBronder:
Yeah #2034 is waiting on some stuff from the compiler tho' so I'm guessing this PR will go in first and we the merge conflicts should be pretty simple

test/unit/math/test_ad.hpp
SteveBronder: > We can detect whether the signature f(var<mat>) exists and only run these tests in that instance

Ack, actually I'm not sure how to do this. The issue is that we to have a function that only calls the tests when the function has an overload that accepts a `var<matrix>`

```cpp
// if F(var<matrix>) is a valid expression do the tests
template <typename F, typename T, require_t<is_overloaded<F, var<matrix>>>* = nullptr>
void expect_ad_matvar(F&& f, const T& y) {
  // do the tests
}

// if F(var<matrix>) is a not valid expression don't do the tests
template <typename F, typename T, require_t<is_not_overloaded<F, var<matrix>>>* = nullptr>
void expect_ad_matvar(F&& f, const T& y) {}
```

But in our test suite `F` is always a lambda with `auto` as the argument type like

```cpp
auto test_lambda_F = [](const auto& x) { return actual_func(x);};
```

 so `F(var<matrix>)` is a valid expression and passes SFINAE here because of the `auto`. But when the compiler tries to do resolution it finds that it can't compile the body of the lambda and errors.

There's an example here where I'm just doing this with `double` (equivalent of `Matrix<var>`) and tries to go to the failure overload for an input of `std::string`

https://godbolt.org/z/o7EW8E

bbbales2: > so F(var<matrix>) is a valid expression and passes SFINAE here because of the auto.

Ugh. Not obvious to me how to get around this. I guess it'll just be our responsibility to turn on `var_value` testing for functions where it is implemented.

Since we aren't doing explicit casts at some point we'll have to be tracking what functions support `var<mat>` types and what functions can only do `mat<var>`, so there is already a problem at least as complicated as this one in existence.

SteveBronder: Yeah let's re-approach this once we get closer to full coverage

stan/math/rev/core/vari.hpp
bbbales2: > Do you mean the arena matrix could point to new memory

I mean:

```cpp
arena_matrix<MatrixXd> memory(100, 100);
arena_matrix<var_value<MatrixXd>> result = memory;

memory = 1.0; // result changes too!
```

I think it should only steal the pointer to the values if memory is `const`.

test/unit/math/test_ad.hpp
andrjohns: You can do this by having the third parameter of the function be ill-formed if the input/s don't meet particular constraints. 
I've modified your code to show this in action: https://godbolt.org/z/ejP9Mn

Where the ```whatsup``` functions are given a third parameter that applies a function that would only be appropriate for the specified type, otherwise its ill-formed.

From templates to parameters, it's SFINAE all the way down!

bbbales2: I think the requirement is that the function `F` take in a certain type, not that the `y` is of a certain type. This is different from the thing you posted right?

test/unit/math/test_ad.hpp
```diff
@@ -1432,6 +1433,78 @@ void expect_ad_vectorized_binary(const F& f, const T1& x, const T2& y) {
...
+    }
+  }
+  var b_mv = sum(A_mv_f);
+  var b_vm = sum(A_vm_f);
+  var final_var = b_mv + b_vm;
+  stan::math::grad(final_var.vi_);
+  if (!is_finite(x) || !stan::math::isfinite(final_var)) {
+    return;
+  }
+  EXPECT_MATRIX_FLOAT_EQ(A_vm.val(), A_mv.val());
```
bbbales2, line 1503: This should really be using `expect_near_rel`.

test/unit/math/test_ad.hpp
```diff
@@ -1432,6 +1433,78 @@ void expect_ad_vectorized_binary(const F& f, const T1& x, const T2& y) {
...
+ * gradient of the sum which will propogate the adjoints upwards.
+ *
+ * @tparam F A lambda or functor type that calls the unary function.
+ * @tparam T An Eigen matrix type
+ * @param f a lambda
+ * @param x An Eigen matrix.
+ *
+ */
+template <typename F, typename T>
+void expect_ad_matvar(F&& f, const T& x) {
```
bbbales2, line 1463: This should take in a tolerances argument to support `expect_near_rel`: https://github.com/stan-dev/math/blob/0f9b5f031e2171994a49d9dce95f6c50e857066a/test/unit/math/test_ad.hpp#L1123

test/unit/math/test_ad.hpp
```diff
@@ -1432,6 +1433,78 @@ void expect_ad_vectorized_binary(const F& f, const T1& x, const T2& y) {
...
+    try {
+      f(A_mv);
+      FAIL() << "var<matrix> throws, expect matrix<var> version to throw";
+    } catch (...) {
+      SUCCEED();
+    }
+  }
+  var b_mv = sum(A_mv_f);
+  var b_vm = sum(A_vm_f);
+  var final_var = b_mv + b_vm;
```
bbbales2, line 1498: These should be checking the full Jacobian of the function, not the Jacobian of the sum of the function.

With the old style vars, this means doing a loop over the output, doing `.grad()` of each.

In the new code, this means doing a loop over the output elements, setting the appropriate element of the output adjoint variable to 1, zeroing all the others, calling `grad()`, and then checking the results against the `mat<var>` version.

bbbales2: > not the Jacobian of the sum of the function.

We could do different random weighted sums as a way of checking these things, but we've just done the brute force thing everywhere else.

test/unit/math/test_ad.hpp
bbbales2: Still needs changed to use `expect_near_rel` or removed. `expect_near_rel` is set up to handle values near zero better than `EXPECT_FLOAT_EQ`. If `EXPECT_FLOAT_EQ` worked, then `expect_near_rel` probably will. It's a better way of doing these comparisons so edge cases don't trip you up.

stan/math/rev/core/vari.hpp
bbbales2: I still think const goes here.

bbbales2 (changes requested):
More review. I'm around to hangouts if you want to talk about the `const` thing or the ad test stuff.

test/unit/math/test_ad.hpp
SteveBronder: I was trying something similar to that before, the issue is that the lambda always passes SFINAE since it's argument is `auto`. The compiler fails after SFINAE while processing the body of the function. 

https://godbolt.org/z/nj3rn7

The only scheme I've been able to work out is to actually pass the function as well so that type deduction can be done directly on the function we want to test `var<mat>` vs. `mat<var>` for

https://godbolt.org/z/brGofz

stan/math/rev/fun/sum.hpp
bbbales2: Oh yeah I agree with `reverse_pass_callback`. Can move here or later or whenever.

test/unit/math/test_ad.hpp
andrjohns: Oh got it, got it. After some digging looks like an easy fix is to have a trailing return type for the lambda:
```
int main() {
  whatsup([](auto x) -> decltype(foo(x)) { foo(x);}, 2.0);
  whatsup([](auto x) -> decltype(foo(x)) { foo(x);}, std::string("UUggghhh!"));
  return 0;
}
```

Working here: https://godbolt.org/z/TeaMde

bbbales2: This seems like a good solution to me given the tests won't compile without this. I would prefer to do the unified `mat<var>` `var<mat>` test work in another pull though to not stall the other stuff in this branch.

test/unit/math/test_ad.hpp
SteveBronder: Is there a reason we would want to use `expect_near_rel` instead of an equality? They should both compute the same thing right?

test/unit/math/test_ad.hpp
SteveBronder: Are you thinking of something like this?

```cpp
  for (Eigen::Index i = 0; i < A_vm.size(); ++i) {
    A_vm_f.adj()(i) = 1;
    A_mv_f.adj()(i) = 1;
    stan::math::grad();
    EXPECT_FLOAT_EQ(A_vm.adj()(i), A_mv.adj()(i));
    EXPECT_FLOAT_EQ(A_vm_f.adj()(i), A_mv_f.adj()(i));
  }
```

stan/math/rev/core/vari.hpp
SteveBronder: > I think it should only steal the pointer to the values if memory is const.

If it's const idt we can take the memory because you can't have a non-const reference to a const object

```
const double a = 5;
double& b = a; // fails
```

I think stealing that memory makes sense. Like when we copy a var we just copy over the `*vi_` pointer so the same thing happens like

```cpp
  using stan::math::var;
  var A(5);
  var B = A;
  B.vi_->adj_ = 4;
  std::cout << "Adjoint of A: " << A.vi_->adj_;
  // Adjoint of A: 4
```

SteveBronder:
Yeah @bbbales2 after the meeting tmrw let's jump on a call rq

stan/math/rev/core/vari.hpp
t4c1: This would be a nasty bug if it actually happens anywhere. @bbbales2 Can you think of a realistic example where arena_matrix is used to construct a var_value and than changed? Or another way - where a var_value is changed and the arena_matrix it is constructed from should not? After some thinking about this I can't think of any such example.

bbbales2: > After some thinking about this I can't think of any such example.

@t4c1 yeah I agree with what you're saying. Had a talk with @SteveBronder this morning too and he convinced me that we shouldn't worry about this.

SteveBronder:
@bbbales2 I went through and updated the tests like we talked about. 

I left the sum of the jacobian checks in along with the check of the full jacobian. We could remove it but idk I kind of like testing some arbitrary compound expression to make sure everything checks out

test/unit/math/test_ad.hpp
```diff
@@ -1493,24 +1596,11 @@ void expect_ad_matvar(F&& f, const T& x) {
...
-    EXPECT_FLOAT_EQ(A_vm_f.adj()(i), A_mv_f.adj()(i));
-    stan::math::set_zero_all_adjoints();
-  }
-  var b_mv = sum(A_mv_f);
-  var b_vm = sum(A_vm_f);
-  var final_var = b_mv + b_vm;
-  stan::math::grad(final_var.vi_);
-  if (!is_finite(x) || !stan::math::isfinite(final_var)) {
+  test_matvar_gradient(A_mv, A_vm, A_mv_f, A_vm_f);
+  if (!x.allFinite()) {
```
bbbales2, line 1600: Why here? I'm surprised you'd need this. If `test_matvar_gradient` passes I'm surprised `test_matvar_sum_gradient` wouldn't.

test/unit/math/test_ad.hpp
```diff
@@ -1447,6 +1447,109 @@ void expect_float_equal(const T1& x, const T2& y) {
...
+  if (!stan::math::isfinite(final_var)) {
+    return;
+  }
+  expect_near_rel("var<Matrix> vs Matrix<var> sum intermediate value",
+                  b_vm.val(), b_mv.val(), 1e-12);
+  expect_near_rel("var<Matrix> vs Matrix<var> sum intermediate adjoints",
+                  b_vm.adj(), b_mv.adj(), 1e-12);
+  expect_near_rel("var<Matrix> vs Matrix<var> sum result value", A_vm.val(),
+                  A_mv.val(), 1e-12);
+  expect_near_rel("var<Matrix> vs Matrix<var> sum result adjoints", A_vm.adj(),
```
bbbales2, line 1546: Don't check results adjoints

test/unit/math/test_ad.hpp
```diff
@@ -1447,6 +1447,109 @@ void expect_float_equal(const T1& x, const T2& y) {
...
+                  A_mv.adj(), 1e-12);
+  expect_near_rel("var<Matrix> vs Matrix<var> sum input value", A_vm_f.val(),
```
bbbales2, line 1548: Don't check input vals

test/unit/math/test_ad.hpp
```diff
@@ -1447,6 +1447,109 @@ void expect_float_equal(const T1& x, const T2& y) {
...
+          typename ResultVarMat, require_eigen_t<ResultMatVar>* = nullptr>
+inline void test_matvar_gradient(MatVar&& A_mv, VarMat&& A_vm,
+                                 ResultMatVar&& A_mv_f, ResultVarMat&& A_vm_f) {
+  for (Eigen::Index i = 0; i < A_vm.size(); ++i) {
+    A_vm_f.adj()(i) = 1;
+    A_mv_f.adj()(i) = 1;
+    stan::math::grad();
+    expect_near_rel("var<Matrix> vs Matrix<var> result value", A_vm.val(),
+                    A_mv.val(), 1e-12);
+    expect_near_rel("var<Matrix> vs Matrix<var> result adjoints", A_vm.adj(),
```
bbbales2, line 1504: Don't check results adjoints

test/unit/math/test_ad.hpp
```diff
@@ -1447,6 +1447,109 @@ void expect_float_equal(const T1& x, const T2& y) {
...
+                    A_mv.adj(), 1e-12);
+    expect_near_rel("var<Matrix> vs Matrix<var> input value", A_vm_f.val(),
```
bbbales2, line 1506: Don't check input vals

test/unit/math/test_ad.hpp
```diff
@@ -1447,6 +1447,109 @@ void expect_float_equal(const T1& x, const T2& y) {
...
+template <typename MatVar, typename VarMat, typename ResultMatVar,
+          typename ResultVarMat, require_var_t<ResultMatVar>* = nullptr>
+inline void test_matvar_gradient(MatVar&& A_mv, VarMat&& A_vm,
+                                 ResultMatVar&& A_mv_f, ResultVarMat&& A_vm_f) {
+  A_vm_f.adj() = 1;
+  A_mv_f.adj() = 1;
+  stan::math::grad();
+  expect_near_rel("var<Matrix> vs Matrix<var> result value", A_vm.val(),
+                  A_mv.val(), 1e-12);
+  expect_near_rel("var<Matrix> vs Matrix<var> result adjoints", A_vm.adj(),
```
bbbales2, line 1472: Same comment with results adjoints/input vals.

bbbales2 (changes requested):
No need to check adjoints of output (you're setting those) or values of inputs (those are given).

I'd like to handle the default tolerances differently but let's get that in the next pull request.

test/unit/math/test_ad.hpp
SteveBronder: Oh actually this should be above `test_matvar_gradient()` like in `test_gradient()`

test/unit/math/test_ad.hpp
SteveBronder: The actual final value is `var final_var = b_mv + b_vm;` and we run everything up from it `stan::math::grad(final_var.vi_);` so wouldn't it make sense to check propagates right?

SteveBronder:
Aight this should be good!

test/unit/math/test_ad.hpp
bbbales2: Oh, I guess here that's fine. Certainly not going to hurt.

bbbales2:
@serban-nicusor-toptal Did something weird happen with the tests here or should I just restart them?

serban-nicusor-toptal:
A few jobs got stuck on a windows machine and I had to stop them, sorry for the inconvenience.
I've restarted the above one from the downstream tests [here](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2050/25/pipeline)

SteveBronder:
@serban-nicusor-toptal it looks like the windows machine is down?

SteveBronder:
@serban-nicusor-catena it looks the windows server went down again? Should I just restart the job or is there something on the backend that needs updated?

rok-cesnovar:
I think the only thing you can do is restart. Jenkins machine have been quite busy these last few days, will make sure to check in the morning CET time and restart if necessary.

serban-nicusor-toptal:
Hey @SteveBronder sorry for the trouble, one Linux and windows machine we have flipped on/off  past few days and may have slowed down our ci a little or 'break' some already running jobs. We also have in parallel AWS machines which took care of the load so everything should be fine.

bbbales2 (approved):

bbbales2:
I approve. We need a follow up issue made for generalizing the testing stuff introduced in this pull request (I'm leaving this to @SteveBronder):

To generalize it should include at least:
1. Add `ad_tolerance` arguments to the test functions (expect_ad has a default if none is passed in)

2. If I recall correctly, we decided somewhere to not construct `mat<var>` s from `var<mat>` s, and that's why we wanted to only test functions with `var<mat>` if they're actually supported. If we can do this detection automatically it would be nice (https://github.com/stan-dev/math/pull/2050#discussion_r486030621) or if the tests are enabled manually per-function that's probably okay since we'll have to track this somewhere anyway.

3. Multiple argument testing. If a function supports `var<mat>` and `mat<var>`, then it should be tested with all combinations of those (and prim matrices any other arguments it takes)

SteveBronder:
@t4c1 you approve?

t4c1 (approved):

rok-cesnovar:
Now that this is in we can start experimenting with the stanc3 implementation for var_value. If anyone has a semi-realistic model that uses softmax that we can also test performance with, that would be great.

SteveBronder:
Do we also need the sunsetting stuff? I'm almost done with that up in Stan but it's rough around the edges rn

bbbales2:
> Now that this is in we can start experimenting with the stanc3 implementation for var_value

@andrewgelman that's your signal to put together some test models. I can make up some pretend regressions otherwise, but it wouldn't be bad if they made sense.

> that uses softmax that we can also test performance with, that would be great.

I'll start working on converting the rest of the functions to `var<mat>` and adding benchmarks in the shiny thing.

rok-cesnovar:
> Do we also need the sunsetting stuff?

I am not exactly sure what you mean by that.

SteveBronder:
Sorry typo meant subsetting like `A[, 5]` on the static matrix

bbbales2:
@rok-cesnovar The benchmarks for softmax are:

```
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
toss_me                            5655812 ns      5652462 ns          125
softmax/2/manual_time                  168 ns          240 ns      4191861
softmax/4/manual_time                  183 ns          279 ns      3843557
softmax/8/manual_time                  226 ns          361 ns      3100167
softmax/16/manual_time                 307 ns          529 ns      2268886
softmax/32/manual_time                 489 ns          881 ns      1431106
softmax/64/manual_time                 852 ns         1581 ns       822573
softmax/128/manual_time               1615 ns         3024 ns       433020
softmax/256/manual_time               3242 ns         6025 ns       215452
softmax/512/manual_time               6380 ns        11769 ns       109682
softmax/1024/manual_time             12457 ns        23184 ns        56164
softmax_varmat/2/manual_time           127 ns          194 ns      5535520
softmax_varmat/4/manual_time           142 ns          234 ns      4945153
softmax_varmat/8/manual_time           172 ns          291 ns      4054579
softmax_varmat/16/manual_time          220 ns          398 ns      3189740
softmax_varmat/32/manual_time          321 ns          616 ns      2181287
softmax_varmat/64/manual_time          522 ns         1052 ns      1340087
softmax_varmat/128/manual_time         957 ns         1962 ns       732250
softmax_varmat/256/manual_time        1883 ns         3881 ns       371860
softmax_varmat/512/manual_time        3595 ns         7487 ns       194910
softmax_varmat/1024/manual_time       7125 ns        14873 ns        98541
```

So doing `var<mat>` here is a little less than twice as fast.

PR #2064, opened 2020-09-07
Allow the var views to compound multiple expressions and remove var.coeff() callbacks

SteveBronder:
## Summary

This PR allows for `var<matrix>` types to chain together multiple slicing methods such as 

```
auto A_flip = A_v.rowwise().reverse().colwise().reverse();
```

It also has a specialization of `vari_value<T>` when `T` is a reference to a scalar type. This specialization removes the callbacks that were needed in the last implementation of the `var_value.coeff()` methods.

While I was working on the `var<Matrix>` indexing methods in this stan [branch](https://github.com/stan-dev/stan/compare/feature/eigen-slices) I realized I needed to be able to write compound expressions such as `a_var_matrix.rowwise().reverse()`. `reverse()` was needed in particular because stan's indexing method for code such as

```stan
matrix[K, K] A = Other[5:K, 5:K]
```

uses an `index_min_max(5, K)` when performing indexing, but this name is dubious! The stan language will use `index_min_max(5, K)` even if we don't know whether `K` is greater than the starting index! In the case of a decreasing index we need to be able to call eigen's `.reverse()` function so that the `min_max` index can grab the appropriate slice and then reverse the order.

This PR adds those methods and allows them to be chained together. To do that I had to also change `is_eigen` and `plain_type_t`. Currently `is_eigen` checks to see whether or not a pointer to a type `T` is convertible to `Eigen::EigenBase<T>*`. But some Eigen expressions such as [`VectorWiseOp`](https://eigen.tuxfamily.org/dox/VectorwiseOp_8h_source.html) don't actually inherit from `EigenBase` so `is_eigen<VectorWiseOp<T>>` fails (see line 156 of the `VectorWiseOp` code link). `VectorWiseOp` also does not have a `PlainObject` trait so `plain_type_t` fails for these types. This means I had to change both of these definitions a bit where `is_eigen` now checks whether the type `T` is pointer convertible to `EigenBase<T>*` or if `T` has a trait named `ExpressionTypeNestedCleaned`. A similar thing is used for `plain_type_t` where we detect whether the class has either a `PlainObject` trait or a `ExpressionTypeNestedCleaned` trait and report back either. (@t4c1 this feels a bit ad-hoc so if you have a nicer way to deduce these I'm all ears!)

Another notable change in this PR is that there is now a `vari_value<T>` specialization for when `T` is a reference type. Having this specialization allows us to remove the callbacks that were needed when slicing a `vari<matrix>` type to get back a single coefficient.

## Tests

This adds additional tests in `var_test` to check if compound slicing expressions work and if their values and adjoints match up to the original values.

## Side Effects

Yes, I worry that the `vari_value<T>` specialization that accepts references is not defensive enough. Right now I have the specialization marked as `final` which should stop anyone from deriving a class from it. This class should really only be used in `var_value`, is there some way to allow that?

## Release notes

Adds views to `vari_value<Matrix>` types that allow for compound slicing expressions.

## Checklist

- [x] Math issue #1805 


SteveBronder:
Marking this as WIP because I also need to sort out how to assign a `var_value` that has an expression to a `var_value` with a plain type

t4c1:
> is_eigen<VectorWiseOp<T>> fails

In my opinion that is fine. We don't want to (and can't) support `VectorWiseOp` everywhere we support all other eigen expressions. So if we do need to support it somewhere we should add some special handling (even if that is just an extended require).

> Another notable change in this PR is that there is now a vari_value<T> specialization for when T is a reference type.

Nice. However I believe a number of open questions will appear regarding this implementation.

SteveBronder:
> In my opinion that is fine. We don't want to (and can't) support VectorWiseOp everywhere we support all other eigen expressions. So if we do need to support it somewhere we should add some special handling (even if that is just an extended require).

I agree with the `is_eigen` stuff and can revert that. Though I'd like to leave the `plain_type_t` because I think that does make sense

>> Another notable change in this PR is that there is now a vari_value specialization for when T is a reference type.

> Nice. However I believe a number of open questions will appear regarding this implementation.

Yeah. I think one solution could be to make all of it's constructors private. Then since `var` is a private friend it should be the only one able to access them (I think?) if not I can take another crack at the `var_view` type

t4c1:
> Though I'd like to leave the plain_type_t because I think that does make sense

I am a bit hesitant, but let's say that is fine. To avoid the adhoc method, I think it might be possible to do something similar we have in `is_eigen` - using overload resolution.

> Yeah. I think one solution could be to make all of it's constructors private. Then since var is a private friend it should be the only one able to access them (I think?) if not I can take another crack at the var_view type

I don't see making things hidden enough a problem. We can always declare member functions private or move stuff in internal namespace. 

What I mean is things like: How do we make the interface with varis containing references? There are two main groups of accesses to vari (that I see):

a) trough the var (calls to .adj() and .val(), var constructors and assignment operators)
b) from other varis representing operaions that use the particular vari (chain function updating arguments' adjoints)

In both of this cases the accessing class either needs to restricts itself to only accessing virtual functions or knowing exact type in the vari (is it `double` or `double&` / plain matrix or a block or a rowwise reversed view ...). 

For the Eigen operations we already have rev implementations templated with argument types. I think we need to do the same for scalar ones in order to use the references.

I am not so sure about accesses trough var. It already knows exact type of the vari, but ... does it need to? Right now that makes calls to `adj()` and `val()` efficient, but assignments of `var<double&>` to `var<double>` need to do a conversion including copying data and setting up a reverse pass callback.

Instead we could have assignments just copy the vari pointer, but `adj()` and `val()` would have to be implemented through a call to a virtual function on the vari.

SteveBronder:
Had to read your comment a few times to get a full grasp of the issue and those are good points.

> For the Eigen operations we already have rev implementations templated with argument types. I think we need to do the same for scalar ones in order to use the references.

I think this is v possible. By the Eigen operations do you mean the math ops that just take in an eigen type? Like how we went from `Matrix<T, R, C>` to just a template parameter `EigMat`.

> Instead we could have assignments just copy the vari pointer, but adj() and val() would have to be implemented through a call to a virtual function on the vari. Although this probably complicates the implementation of templated operations.

This one kind of worries me. I'd have to sit down and play around with this to get a better grasp of it.

For coefficient access, If we want to think about this more why don't we just do the hard copy for now and get rid of the `var_value<double&>` specialization. Then when we can write some tests and have a better understanding of the problem we can come back to this. If we go forward with the templated argument types I think we would want to do that first before we have `var_value<double&>` types floating around

For the `vari_views`   I'm not super worried atm. My main thing with those was just to get them up and running so we could write the indexing stuff the compiler needs and eod they will end up copied into a `var_value<plain_type_t<T>>`.

SteveBronder:
>  If we go forward with the templated argument types I think we would want to do that first before we have var_value<double&> types floating around

Oh actually saying that out loud I think we essentially need to do that before we have a `vari_value<double&>` specialization or something like `A[1, 1] * B[2, 2]` in a stan program would fail to compile

t4c1:
> By the Eigen operations do you mean the math ops that just take in an eigen type? 

Yeah, since we are talking reverse, either the scalar is var or the type is var_value.

> This one kind of worries me. I'd have to sit down and play around with this to get a better grasp of it.

I think we can start with templating, before trying more virtual calls. Than we can see what needs to be made virtual for this to work and benchmark to see if virtual version is faster or slower.

> For coefficient access, If we want to think about this more why don't we just do the hard copy for now and get rid of the var_value<double&> specialization. Then when we can write some tests and have a better understanding of the problem we can come back to this. 

Well, just using tem in expression `var_value<double&>` will sure be faster. If you want to leave this change for another PR or for me, that is fine.

> If we go forward with the templated argument types I think we would want to do that first before we have var_value<double&> types floating around

I think it is not that large change so both can be done in the same PR. If you end up doing it, you can structure PRs as you want.

> Oh actually saying that out loud I think we essentially need to do that before we have a vari_value<double&> specialization or something like A[1, 1] * B[2, 2] in a stan program would fail to compile if A and B were var_value<Matrix> types

Yes.

SteveBronder:
@t4c1 Currently fiddling with the block assignement test [here](https://github.com/stan-dev/math/pull/2064/files#diff-2b0d4604da2786aba9769c587db655c5R237). Should Matrix<var> and var<matrix> do the same thing here? I'm trying to investigate why it's not propogating the block's values to the final A_v adjoints

```cpp
  // setup
  Eigen::MatrixXd A(4, 4);
  A << 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15;
  // construct
  var_value<Eigen::MatrixXd> A_v(A);
  Eigen::Matrix<stan::math::var, -1, -1> A_alt = A;
  // blocks
  A_v.block(0, 0, 3, 3) = A_v.block(1, 1, 3, 3);
  A_alt.block(0, 0, 3, 3) = A_alt.block(1, 1, 3, 3);
  // row slices
  var_value<Eigen::RowVectorXd> A_v_row = A_v.row(3);
  auto A_alt_row = A_alt.row(3).eval();

  stan::math::var blah1 = stan::math::sum(A_alt_row);
  stan::math::var blah = stan::math::sum(A_v_row);

  blah1.vi_->adj_ = 1;
  blah.vi_->adj_ = 1;
  stan::math::grad();
```

result:
```
A_v: 
0 0 0 0
0 0 0 0
0 0 0 0
1 1 1 1
A_alt: 
0 0 0 0
0 0 0 0
1 1 1 0
1 1 1 1
A_v_row: 
1 1 1 1
A_alt_row: 
1 1 1 1
```

SteveBronder:
Ahh okay I had to get a pen and piece of paper for this.

So what I was confused by was the output where for `Matrix<var>` doing 

```cpp
  A_v.block(0, 0, 3, 3) = A_v.block(1, 1, 3, 3);
```

Actually puts the vari pointers on the rhs into the var's on the lhs, but then the original vari pointers from the lhs still exist for any `chain()` calls that need them. But with a `var<matrix>` we need to rewind the adjoints to before that block op happened so that chain calls that need the original values have them.

Also updated eigen's gdb pretty printers to work with maps which folks may find useful.

https://gist.github.com/SteveBronder/d95bdf4bbb8ab0fa195a7ed9e3d0a140

SteveBronder:
@t4c1 can you take a look at this?

stan/math/rev/core/var.hpp
```diff
@@ -385,44 +629,91 @@ class var_value {
...
+    using vari_sub = decltype(vi_->colwise());
+    using var_sub = var_value<typename vari_sub::value_type>;
+    return var_sub(new vari_sub(vi_->colwise()));
+  }
+
+  /**
+   * Return an expression an expression to reverse the order of the coefficients
+   * inside of a `vari` matrix
+   */
+  inline auto reverse() {
```
t4c1, line 711: Should this require a vector or can we do it on a matrix?

stan/math/rev/core/var.hpp
```diff
@@ -467,7 +758,7 @@ class var_value {
    */
   template <typename S, require_convertible_t<S&, value_type>* = nullptr,
             require_all_plain_type_t<T, S>* = nullptr>
-  var_value<T> operator=(const var_value<S>& other) {
+  inline var_value<T>& operator=(const var_value<S>& other) {
```
t4c1, line 761: Nice catch!

stan/math/rev/core/var.hpp
```diff
@@ -502,6 +793,20 @@ class var_value {
     return *this;
   }
 
+  // No-op to match with Eigen methods which call eval
+  template <typename T_ = T, require_plain_type_t<T_>* = nullptr>
+  inline auto& eval() noexcept {
```
t4c1, line 798: This is missing doxygen.

stan/math/rev/core/var.hpp
```diff
@@ -514,9 +819,7 @@ class var_value {
    * @param other the value to assing
    * @return this
    */
-  var_value<T> operator=(std::conditional_t<is_eigen<value_type>::value,
-                                            const var_value<T>&, not_var_value>
-                             other) {
+  var_value<T>& operator=(const var_value<T>& other) {
```
t4c1, line 822: That if in the doxygen is now redundant. Remove it.

stan/math/rev/core/var.hpp
```diff
@@ -502,6 +793,20 @@ class var_value {
...
+   * For non-plain types evaluate to the plain type
+   */
+  template <typename T_ = T, require_not_plain_type_t<T_>* = nullptr>
+  inline auto eval() noexcept {
+    return var_value<plain_type_t<T>>(*this);
+  }
+
   // this is a bit ugly workarount to allow var_value<double> to have default
   // assignment operator, which avoids warnings when used in Eigen matrices.
  private:
```
t4c1, line 812: This is not needed anymore - remove.

stan/math/rev/core/vari.hpp
```diff
@@ -78,27 +78,18 @@ class vari_base {
...
 template <typename T>
-class vari_value<T, require_floating_point_t<T>> : public vari_base {
+class vari_value<T, require_t<std::is_floating_point<T>>> : public vari_base {
  public:
   using value_type = std::decay_t<T>;
   /**
    * The value of this variable.
    */
-  const value_type val_;
+  const value_type val_{0.0};
```
t4c1, line 87: I think all constructors set the val, so we do not need the default. Also I don't think it is very meaningful.

stan/math/rev/core/vari.hpp
```diff
@@ -190,9 +181,7 @@ template <typename T, typename = void>
 class vari_view;
 
 template <typename T>
-class vari_view<T, require_all_t<bool_constant<!is_plain_type<T>::value>,
-                                 is_eigen_dense_base<T>>>
-    final : public vari_base {
+class vari_view<T, require_not_plain_type_t<T>> final : public vari_base {
```
t4c1, line 184: Why the change of the require? Does this now also work on some non-eigen types? Which ones?

stan/math/rev/core/vari.hpp
```diff
@@ -210,9 +199,134 @@ class vari_view<T, require_all_t<bool_constant<!is_plain_type<T>::value>,
   template <typename S, typename K,
             require_convertible_t<S&, value_type>* = nullptr,
             require_convertible_t<K&, value_type>* = nullptr>
-  vari_view(const S& val, const K& adj) : val_(val), adj_(adj) {}
+  vari_view(S&& val, K&& adj) noexcept : val_(val), adj_(adj) {}
```
t4c1, line 202: `&&` should be paired with `std::forward` in 99% of the use cases. Also I am not sure we need perfect forwarding. This accepts Eigen expressions and not plain types. I think these do not use any dynamically allocated memory, so moving is no better than copying.

stan/math/rev/core/vari.hpp
```diff
@@ -210,9 +199,134 @@ class vari_view<T, require_all_t<bool_constant<!is_plain_type<T>::value>,
...
+   */
+  Eigen::Index rows() const { return val_.rows(); }
+  /**
+   * Return the number of columns for this class's `val_` member
+   */
+  Eigen::Index cols() const { return val_.cols(); }
+  /**
+   * Return the size of this class's `val_` member
+   */
+  Eigen::Index size() const { return val_.size(); }
```
t4c1, line 329: I think we could use a common base class for vari_view and Eigen vari_value to avoid duplicating this code.

stan/math/rev/core/vari.hpp
```diff
@@ -236,7 +350,7 @@ class vari_value<T, require_all_t<is_plain_type<T>, is_eigen_dense_base<T>>>
   using PlainObject = plain_type_t<T>;
   using value_type = PlainObject;  // The underlying type for this class
   using eigen_scalar = value_type_t<PlainObject>;  // A floating point type
-  using eigen_map = Eigen::Map<PlainObject, Eigen::Aligned8,
+  using eigen_map = Eigen::Map<PlainObject, Eigen::Unaligned,
```
t4c1, line 353: Does the arena allocator not guarantee some alignment? Hmm a comment says so, but I can't find any code that would ensure that. Anyway if you are using default alignment and stride you don't need to specify them.

stan/math/rev/core/var.hpp
```diff
@@ -177,12 +149,284 @@ class var_value {
...
+   * @return Reference to the specified output stream.
+   */
+  friend std::ostream& operator<<(std::ostream& os, const var_value<T>& v) {
+    if (v.vi_ == nullptr) {
+      return os << "uninitialized";
+    }
+    return os << v.val();
+  }
+
+  inline auto& eval() noexcept { return *this; }
```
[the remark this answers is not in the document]
SteveBronder: Nah idt so

stan/math/rev/core/var.hpp
```diff
@@ -177,12 +149,284 @@ class var_value {
...
+ * other scalar. Arithmetical functions like negation, addition,
+ * and subtraction, as well as a range of mathematical functions
+ * like exponentiation and powers are overridden to operate on
+ * var values objects.
+ * @tparam T An Floating point type.
+ */
+template <typename T>
+class var_value<
+    T, require_t<bool_constant<is_eigen<T>::value || !is_plain_type<T>::value
+                               || is_matrix_cl<T>::value>>> {
```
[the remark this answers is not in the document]
SteveBronder: Because some Eigen types like the return from `rowwise()` don't actually inherit from `EigenBase`. Should we have some other detection like `is_eigen_expr`?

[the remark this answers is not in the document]
SteveBronder: Also I need the rowwise thing so we can efficiently do `reverse()` when users ask for `min_max_index` subsetting like `A[4:1,]`

stan/math/rev/core/var.hpp
```diff
@@ -177,12 +149,284 @@ class var_value {
...
+   * Return a reference to the derivative of the root expression with
+   * respect to this expression.  This method only works
+   * after one of the `grad()` methods has been
+   * called.
+   *
+   * @return Adjoint for this variable.
+   */
+  inline auto& adj() { return vi_->adj_; }
+  inline auto& adj() const { return vi_->adj_; }
+  inline auto& adj_op() { return vi_->adj_; }
```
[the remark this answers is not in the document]
SteveBronder: to avoid an unneeded eval in the var matrix multiply I has to add [`adj_op`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/eigen_plugins.h#L167) to the eigen plugins to return back a non-const `CwiseUnaryOp`. This just is so that `var<matrix>` and `matrix<var>` can use the same code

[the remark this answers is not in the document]
SteveBronder: See discussion below

https://github.com/stan-dev/math/pull/2091#discussion_r493966109

stan/math/rev/core/vari.hpp
SteveBronder: `rowwise()` returns a [`VectorWiseOp`](https://eigen.tuxfamily.org/dox/classEigen_1_1VectorwiseOp.html) which does not inherit from `DenseBase`. If we wanted to avoid this I think `reverse()` always returns a dense expression so for these cases we could have functions instead like `rowwise_reverse()`

stan/math/rev/core/vari.hpp
SteveBronder: Yeah I think this is just a lazy case from me which is why there's no forward here either.

SteveBronder: If they did hold dynamic memory I'm pretty sure that would be real bad for this design because that would mean it needs to inherit from chainable_alloc(). ftr this view type is assuming the expressions here don't alloc any dynamic memory

stan/math/rev/core/vari.hpp
SteveBronder: Yeah the allocator guarantees 8 byte alignment I can fix this up

It looks like Map uses `Unaligned` by default. One thing I was confused by here, does telling Map the alignment tell it "This is _at least_ 8 byte aligned" or that it is exactly 8 byte aligned?

https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html

stan/math/rev/core/vari.hpp
SteveBronder: Yes! I wrote up a neat lil' crtp to do this I'll post up now

stan/math/rev/core/var.hpp
```diff
@@ -385,44 +629,91 @@ class var_value {
...
-    auto* vari_coeff = new vari_coeff_type(vi_->coeff(i));
-    reverse_pass_callback([vari_coeff, this, i]() mutable {
-      this->vi_->adj_(i) += vari_coeff->adj_;
+  inline auto coeff(Eigen::Index i) {
+    using vari_sub = decltype(vi_->coeff(i));
+    vari_sub* vari_coeff = new vari_sub(vi_->coeff(i));
+    reverse_pass_callback([this_vi = this->vi_, vari_coeff, i]() {
+      auto tmp_adj = this_vi->adj_.coeffRef(i);
+      this_vi->adj_.coeffRef(i) += vari_coeff->adj_;
+      vari_coeff->adj_ += tmp_adj;
```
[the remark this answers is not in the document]
SteveBronder: I was really confused here and agree it seems like nonsense to change the result. I was trying to get these to work like the other views where the coefficient would be the same as the coefficient in the matrices. But until we can return back references to doubles I'll just have this slice off a new var that updates to the matrix.

stan/math/rev/core/var.hpp
SteveBronder: We can do it on matrices

stan/math/rev/core/var.hpp
```diff
@@ -322,7 +566,7 @@ class var_value {
    * @param num_cols Number of columns to return.
    */
   inline auto block(Eigen::Index start_row, Eigen::Index start_col,
-                    Eigen::Index num_rows, Eigen::Index num_cols) const {
+                    Eigen::Index num_rows, Eigen::Index num_cols) {
```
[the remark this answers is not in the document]
SteveBronder: Should they be const? Like if we do `A.block(0, 0, 2, 2)` that is returning back a new type, but that type will modify the underlying `vari*` that was returned from the const method which idt is supposed to happen right?

stan/math/rev/core/vari.hpp
```diff
@@ -189,10 +180,184 @@ using vari = vari_value<double>;
...
 
+/**
+ * This struct is follows the CRTP for methods common to `vari_view<>` and
+ * `vari_value<Matrix>`.
+ */
 template <typename T>
-class vari_view<T, require_all_t<bool_constant<!is_plain_type<T>::value>,
-                                 is_eigen_dense_base<T>>>
-    final : public vari_base {
+struct vari_view_crtp {
```
SteveBronder, line 188: @t4c1 if we can define these for `matrix_cl` can just plug this in! Then I think all the `rvalue()` subsetting functions I'm doing should also work for `matrix_cl` as well!

stan/math/rev/core/vari.hpp
t4c1: Why does it matter? If some object is aligned to more than 8byte it is at the same time 8byte aligned.

Also if that is so, probably the arena matrix also should tell its maps the data is aligned.

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: Yeah `!is_plain_type<T>::value` is a tiny bit confusing. It is fine if you want to leave it as it is. I don't think adding `is_eigen_expr` is necessary (and if it would not be a negation of `is_plain_type` it would really get confusing). Instead I would suggest adding `is_eigen_vectorwise` that would detect these rowwise and colwise expressions.

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: Hmm, would not making .adj() non-const solve the problem?

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: So far in the whole math library const var or const matrix of vars means only that the var or matrix of var is const, but not the underlying data. Whole library is modifying adjoints of vars and matrices it got in as const arguments.

The problem will be that most of times matrices are passed around by const references. So if this function is not const, a function accepting a matrix will not be able to get a block of its argument.

Alternatively you will need to duplicate all these methods to have const aversion returning a const object and a non-const version returning a non-const object.

stan/math/rev/core/var.hpp
t4c1: I see how it can be done on rowwise or colwise view into matrix. But what does it do on a plain matrix?

stan/math/rev/core/var.hpp
t4c1: Also I believe this function should be const as well, for the same reason as above. That it also need to return a const reference (which matches the Eigen implementation)

stan/math/rev/core/vari.hpp
t4c1: Actually I would prefere functions, such as `rowwise_reverse()`. Not a mandatory change if you disagree.

stan/math/rev/core/vari.hpp
```diff
@@ -180,29 +180,41 @@ using vari = vari_value<double>;
...
+  }
+
+  /**
+   * Helper function to return a reference to the derived type's adjoint.
+   */
+  inline auto& derived_adj() { return static_cast<T&>(*this).adj_; }
+  /**
+   * Helper function to return a reference to the derived type's adjoint.
+   */
+  inline auto& derived_adj() const { return static_cast<T&>(*this).adj_; }
```
t4c1, line 216: I don't think we need these as separate functions. We can always use `derived().val_`.

stan/math/rev/core/vari.hpp
t4c1: That is fine.

stan/math/rev/core/vari.hpp
t4c1: That should be possible.

stan/math/rev/core/vari.hpp
```diff
@@ -180,29 +180,41 @@ using vari = vari_value<double>;
...
+/**
+ * This struct is follows the CRTP for methods common to `vari_view<>` and
+ * `vari_value<Matrix>`.
+ */
 template <typename T>
-class vari_view<T, require_not_plain_type_t<T>> final : public vari_base {
- public:
-  using PlainObject = plain_type_t<T>;
-  using value_type = std::decay_t<T>;  // The underlying type for this class
+struct vari_view_crtp {
```
t4c1, line 188: I would prefere some other name for this class. Perhaps `vari_dense_base`? Also you can make it inherit from `vari_base` so the classes derived from this one don't need to.

stan/math/rev/core/vari.hpp
```diff
@@ -747,9 +676,6 @@ class vari_value<T, require_eigen_sparse_base_t<T>> : public vari_base,
   friend class var_value;
 };
 
-// For backwards compatability the default is double
-using vari = vari_value<double>;
```
t4c1, line 615: Why is this deleted?

stan/math/rev/meta/is_vari.hpp
```diff
@@ -13,15 +13,22 @@ struct is_vari_impl : std::false_type {};
...
  * vari_value.
  */
 template <typename T>
-struct is_vari<T,
-               std::enable_if_t<internal::is_vari_impl<std::decay_t<T>>::value>>
+struct is_vari<T, require_t<internal::is_vari_impl<std::decay_t<T>>>>
     : std::true_type {};
 
+template <typename T>
+struct value_type<T, require_t<is_vari<T>>> {
```
t4c1, line 30: `value_type` should be defined in value_type.hpp header!

t4c1 (changes requested):

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: Tagging @andrjohns as I think he has a better view on this. My answer is idt so

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: Aight yeah so with the `rowwise_reverse()` stuff we can just move this back to `is_eigen<T> || is_matrix_cl<T>` since those return objects the inherit from `DenseBase`

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: I ended up going with two versions (const and non-const). The code duplication is a bummer but when I tried making all of these just const I ended up with some const-correctness errors from the compiler that were kind of complicated and I couldn't sort out. I also tried a scheme like in the stackoverflow post below but couldn't sort it out because we have multiple levels of objects and we'd need to do some scheme where we make the inner types from the const version also non-const

https://stackoverflow.com/a/47369227/2269255

stan/math/rev/core/var.hpp
SteveBronder: It's pretty much like running `rowwise().reverse().colwise().reverse()`

https://godbolt.org/z/deeo8d

stan/math/rev/core/vari.hpp
SteveBronder: No I like it. It's only really used for `reverse()` anyway so I think it makes sense.

stan/math/rev/core/vari.hpp
SteveBronder: > Why does it matter? If some object is aligned to more than 8byte it is at the same time 8byte aligned.

I didn't think it did, but I've never looked under the hood at Eigen here to check if they do something in particular on particular alignments that needed a certain alignment length.

> Also if that is so, probably the arena matrix also should tell its maps the data is aligned.

Yeah in the multiply PR I changed that since we have the same guarantee

stan/math/rev/core/vari.hpp
SteveBronder: Yeah fair

stan/math/rev/core/vari.hpp
SteveBronder: It's moved up to be defined right after the scalar vari_value on line 172

stan/math/rev/meta/is_vari.hpp
SteveBronder: So for specializations over certain types we tend to define the `type_trait` that let's us identify a certain type, then write the `value_type` overload. For instance in `is_eigen.hpp` we have. 

```
template <typename T>
struct value_type<T, std::enable_if_t<is_eigen<T>::value>> {
  using type = typename std::decay_t<T>::Scalar;
};
```

So I think this should be moved over to `is_vari.hpp` in prim.

stan/math/rev/core/vari.hpp
SteveBronder: I'd like to keep this seperate since it doesn't actually own or make anything. And mixing this with the virtual methods feels weird. 

Yeah I had a hard time naming this. The issue with `vari_dense_base` is that sparse matrices can also use this. I picked this name because it is for creating the views and is just a crtp pattern for calling the right things.

https://eigen.tuxfamily.org/dox/group__TutorialSparse.html#title7

SteveBronder: by "feels weird" I just had a hard time wrapping my head around the mix of crtp and virtual stuff

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: If you have a way to do this without doing the duplication I'm v open to it! I just couldn't figure it out lol

stan/math/rev/core/vari.hpp
SteveBronder: I ended up having to move these and arena_matrix back to the default `Unaligned` because of an error upstream in the stan tests where I'm getting 

```cpp
rvalue_test: lib/stan_math/lib/eigen_3.3.7/Eigen/src/Core/MapBase.h:194: void Eigen::MapBase<Derived, 0>::checkSanity(typename Eigen::internal::enable_if<(Eigen::internal::traits<OtherDerived>::Alignment > 0), void*>::type) const [with T = Eigen::Map<Eigen::Matrix<double, -1, 1>, 8, Eigen::Stride<0, 0> >; Derived = Eigen::Map<Eigen::Matrix<double, -1, 1>, 8, Eigen::Stride<0, 0> >; typename Eigen::internal::enable_if<(Eigen::internal::traits<OtherDerived>::Alignment > 0), void*>::type = void*]: Assertion `( ((internal::UIntPtr(m_data) % internal::traits<Derived>::Alignment) == 0) || (cols() * rows() * minInnerStride * sizeof(Scalar)) < internal::traits<Derived>::Alignment ) && "data is not aligned"' failed.
Aborted (core dumped)
```

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: From the top of my head no, but I might try to deduplicate it next time I touch this code.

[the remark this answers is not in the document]
t4c1: Actually it looks like const and non const overloads have exactly the same code. Which makes the non-const overloads redundant.

stan/math/rev/core/vari.hpp
t4c1: Lol, the doc on the allocator needs some fixing. Please revert the changes to arena_matrix_as well.

stan/math/rev/core/vari.hpp
t4c1: Fine, than leave the inheritance as it is. How about `vari_base_eigen` than?

stan/math/rev/meta/is_vari.hpp
t4c1: Which I think is messy and makes the code hard to find. We should fix that. The fix might require some forward declarations, but that the specializations will not be in random files.

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: You can leave both overloads of the `eval()`, as the return type actually differs.

stan/math/rev/meta/is_vari.hpp
SteveBronder: I'm v cool with fixing it but let's do all of them in a different PR

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: I think they need to be separate because of the eigen types that are returned from the const and non-const versions. Like if I leave only the const versions then I'm seeing errors where the const returned eigen type can't be cast to the non-const eigen type (which makes sense).

```
./stan/math/rev/core/var.hpp: In instantiation of ‘auto stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::block(Eigen::Index, Eigen::Index, Eigen::Index, Eigen::Index) const [with T = Eigen::Matrix<double, -1, -1>; typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type = void; Eigen::Index = long int]’:
test/unit/math/rev/core/var_test.cpp:181:38:   required from here
./stan/math/rev/core/var.hpp:565:12: error: no matching function for call to ‘stan::math::var_value<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>::var_value(vari_sub*)’
  565 |     return var_sub(
      |            ^~~~~~~~
  566 |         new vari_sub(vi_->block(start_row, start_col, num_rows, num_cols)));
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./stan/math/rev/core/var.hpp:397:3: note: candidate: ‘stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::var_value(stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::vari_type*) [with T = Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>; typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type = void; stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::vari_type = stan::math::vari_view<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>]’
  397 |   var_value(vari_type* vi) : vi_(vi) {}  // NOLINT
      |   ^~~~~~~~~
./stan/math/rev/core/var.hpp:397:24: note:   no known conversion for argument 1 from ‘vari_sub*’ {aka ‘stan::math::vari_view<const Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>*’} to ‘stan::math::var_value<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>::vari_type*’ {aka ‘stan::math::vari_view<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>*’}
  397 |   var_value(vari_type* vi) : vi_(vi) {}  // NOLINT
      |             ~~~~~~~~~~~^~
```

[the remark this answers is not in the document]
SteveBronder: I think they need to be separate because of the eigen types that are returned from the const and non-const versions. Like if I leave only the const versions then I'm seeing errors where the const returned eigen type can't be cast to the non-const eigen type (which makes sense).

```
./stan/math/rev/core/var.hpp: In instantiation of ‘auto stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::block(Eigen::Index, Eigen::Index, Eigen::Index, Eigen::Index) const [with T = Eigen::Matrix<double, -1, -1>; typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type = void; Eigen::Index = long int]’:
test/unit/math/rev/core/var_test.cpp:181:38:   required from here
./stan/math/rev/core/var.hpp:565:12: error: no matching function for call to ‘stan::math::var_value<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>::var_value(vari_sub*)’
  565 |     return var_sub(
      |            ^~~~~~~~
  566 |         new vari_sub(vi_->block(start_row, start_col, num_rows, num_cols)));
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./stan/math/rev/core/var.hpp:397:3: note: candidate: ‘stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::var_value(stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::vari_type*) [with T = Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>; typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type = void; stan::math::var_value<T, typename std::enable_if<std::integral_constant<bool, (stan::is_eigen<T>::value || stan::is_matrix_cl<T>::value)>::value, void>::type>::vari_type = stan::math::vari_view<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>]’
  397 |   var_value(vari_type* vi) : vi_(vi) {}  // NOLINT
      |   ^~~~~~~~~
./stan/math/rev/core/var.hpp:397:24: note:   no known conversion for argument 1 from ‘vari_sub*’ {aka ‘stan::math::vari_view<const Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>*’} to ‘stan::math::var_value<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>::vari_type*’ {aka ‘stan::math::vari_view<Eigen::Block<const Eigen::Map<Eigen::Matrix<double, -1, -1> >, -1, -1, false>, void>*’}
  397 |   var_value(vari_type* vi) : vi_(vi) {}  // NOLINT
      |             ~~~~~~~~~~~^~
```

SteveBronder:
@t4c1 I think this is ready for another look over

stan/math/opencl/rev/vari.hpp
```diff
@@ -23,6 +23,7 @@ template <typename T>
 class vari_value<T, require_kernel_expression_lhs_t<T>>
     : public vari_base, public chainable_alloc {
  public:
+  using value_type = T;
```
t4c1, line 26: Why do you need this?

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
t4c1: @andrjohns Did you miss this? I would really prefere if we could have only 1 adj function.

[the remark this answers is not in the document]
t4c1: @andrjohns Did you miss this? I would really prefere if we could have only one adj function.

t4c1 (changes requested):
Beside the comments from this reviewm there is still a number of open things from previous reviews. Just scroll up and check the unresolved conversations.

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: I think he recently talked about it over here

https://github.com/stan-dev/math/pull/2106#discussion_r497631555

stan/math/rev/core/vari.hpp
SteveBronder: Sure that works for me

stan/math/prim/meta/plain_type.hpp
```diff
@@ -40,16 +41,56 @@ struct eval_return_type {
...
+struct plain_type<T, require_t<is_detected<T, internal::has_plain_object_t>>> {
   using type = typename std::decay_t<T>::PlainObject;
 };
 
+template <typename T>
+struct plain_type<
+    T, require_t<is_detected<T, internal::has_nested_expression_t>>> {
+  using type = typename std::decay_t<T>::ExpressionTypeNestedCleaned;
+};
+
```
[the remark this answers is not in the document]
SteveBronder: Yes! Can rip these out

stan/math/rev/core/var.hpp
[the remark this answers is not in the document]
SteveBronder: In var? For both I'm seeing the const and non-const version of each

SteveBronder:
@t4c1 so I think I got everything from the previous comments as well now. If you don't mind I did the `arena_matrix` in `vari_value<matrix>` thing here because it's a pretty simple change and imo looks v nice!

t4c1:
This looks mostly good now. Although I think we should handle the arena matrix by adding `Eigen::internal::traits` specialization for it instead of the changes to ref_type_t.

You can derive it or use values from traits specialization for the `Eigen::Map` used in arena matrix. Maybe you can even typedef it to be the same type as the traits specialization for the map.

SteveBronder:
> This looks mostly good now. Although I think we should handle the arena matrix by adding Eigen::internal::traits specialization for it instead of the changes to ref_type_t.

So I had a go at this before I did the `ref_type` specialization and ended up going down quite a rabbit hole!. Once I defined `traits` for `arena_matrix` I then had to add a bunch of other "trait" like specializations for things like `unary_evaluator` and that started making things super complicated. Since `arena_matrix` just holds a map to data allocated in our arena I prefer this approach for now. I know you know more about Eigen's internals, would you have time to do this in another PR? Though imo I think just saying "`arena_matrix` types can be treated like they are plain objects" is fine atm since we don't have arena's holding expressions.

t4c1:
If that is the case, no need to specialize all the EIgen internal structures. However the current implementation could be buggy in some edge cases. Can you instead make a `ref_type` and `ref_type_for_opencl` specializations that will just be equivalent to specialzations for the `map`, `arena_matrix` is derived from?

stan/math/prim/meta/is_eigen_map_base.hpp
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Checks whether type T is derived from Eigen::MapBase.
+ * If true this will have a static member function named value with a type
+ * of true, else value is false.
+ * @tparam T Type to check if it is derived from `MapBase`
+ * @tparam Enable used for SFINAE deduction.
+ * @ingroup type_trait
+ */
+template <typename T>
+struct is_eigen_map_base
```
t4c1, line 31: This is not used anywhere. Remove it.

SteveBronder: Gah! lol I was doing something else and slammed this in there

t4c1 (approved):

PR #2065, opened 2020-09-08
make var_views assignable

t4c1:
## Summary
Makes it possible to assign to blocks, and other views into `var_value` (does not include assignments of single elements). 

## Tests
Tested assignment of one block to another that contains aliasing.

## Side Effects
With this there is no reason anymore to call `var_value` a static matrix :)

## Release notes

Blocks, and other views into `var_value` can now be assigned to.

## Checklist

- [x] Math issue #1805 


stan/math/rev/core/var.hpp
```diff
@@ -415,6 +416,51 @@ class var_value {
...
+        [this_vi = this->vi_, other_vi = other.vi_, prev_val]() mutable {
+          this_vi->val_ = prev_val;
+
+          // we have no way of detecting aliasing between this->vi_->adj_ and
+          // other.vi_->adj_, so we must copy adjoint before reseting to zero
+
+          // we can reuse prev_val instead of allocating a new matrix
+          prev_val = this_vi->adj_;
+          this_vi->adj_.setZero();
+          other_vi->adj_ += prev_val;
```
SteveBronder, line 451: Should we reset `this_vi->adj_` to `prev_val` here?

stan/math/rev/core/var.hpp
```diff
@@ -415,6 +416,51 @@ class var_value {
...
+  template <typename S, require_convertible_t<S&, value_type>* = nullptr,
+            require_all_plain_type_t<T, S>* = nullptr>
+  var_value<T> operator=(const var_value<S>& other) {
+    vi_ = other.vi_;
+    return *this;
+  }
+
+  template <typename S, require_convertible_t<S&, value_type>* = nullptr,
+            require_any_not_plain_type_t<T, S>* = nullptr>
+  var_value<T> operator=(const var_value<S>& other) {
```
SteveBronder, line 435: docs

test/unit/math/rev/core/var_test.cpp
```diff
@@ -220,6 +220,17 @@ TEST_F(AgradRev, var_vector_views) {
...
+TEST_F(AgradRev, var_matrix_view_assignment) {
+  Eigen::MatrixXd A(4, 4);
+  A << 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15;
+  stan::math::var_value<Eigen::MatrixXd> A_v(A);
+  A_v.block(0, 0, 3, 3) = A_v.block(1, 1, 3, 3);
+  stan::math::sum(stan::math::from_var_value(A_v)).grad();
+  Eigen::MatrixXd deriv(4, 4);
+  deriv << 0, 0, 0, 1, 0, 1, 1, 2, 0, 1, 1, 2, 1, 2, 2, 2;
+  EXPECT_MATRIX_FLOAT_EQ(A_v.adj(), deriv);
+}
```
SteveBronder, line 232: Adding `stan::math::var_value<Eigen::MatrixXd> A_block = A_v.block(0, 0, 3, 3);` fails with
```
test/unit/math/rev/core/var_test.cpp: In member function ‘virtual void AgradRev_var_matrix_view_assignment_Test::TestBody()’:
test/unit/math/rev/core/var_test.cpp:228:61: error: conversion from ‘var_value<Eigen::Block<Eigen::Map<Eigen::Matrix<double, -1, -1>, 8, Eigen::Stride<0, 0> >, -1, -1, false>>’ to non-scalar type ‘var_value<Eigen::Matrix<double, -1, -1>>’ requested
  228 |   stan::math::var_value<Eigen::MatrixXd> A_block = A_v.block(0, 0, 3, 3);
```

I think we just need a little constructor to do that

```cpp
  template <typename S, require_convertible_t<S&, value_type>* = nullptr,
            require_any_not_plain_type_t<T, S>* = nullptr>
  var_value(const var_value<S>& other) : vi_(new vari_type(other.vi_->val_)) {
    reverse_pass_callback(
        [this_vi = this->vi_, other_vi = other.vi_]() mutable {
          this_vi->adj_ += other_vi->adj_;
        });
  }
```

SteveBronder (changes requested):
Looks like I went a little coo-coo trying to be const correct here when it was not needed lol. Just a note to add a construct so 

```cpp
var_value<Eigen::MatrixXd> A_block = A_v.block(0, 0, 3, 3);
```

works and then this looks good! I think we will need that for the compiler unless we can get away with using auto in the generated code

it would also be good to add more tests, though I can add a bunch of stress tests in #2064

SteveBronder:
Yo fyi I merged this into #2064 because I would need it there anyway. We can either do one branch at a time or do just #2064 your call

stan/math/rev/core/var.hpp
t4c1: No. First at this point in code `prev_val` contains the value of `this_vi->adj_`. From this reverse pass callback on `this_vi->adj_` should contain adjoint associated with the values in this block before they were overwritten. Since they were overwritten, they were not used in any other operations and their adjoint is 0, which is set on line 450.

stan/math/rev/core/var.hpp
t4c1: done

test/unit/math/rev/core/var_test.cpp
t4c1: Fixed and added a test.

t4c1:
> it would also be good to add more tests, though I can add a bunch of stress tests in #2064

Yeah I tend to be a bit lazy with tests. I think this is enough to show it works. If you want to add more, feel free. Or if you want me to add something specific, feel free to ask.

I also intend to add something similar for scalars. However, I wan to avoid merge conflicts, so I will wait until #2064 is merged. Also it contains some changes I'd need to do anyway.

t4c1:
@SteveBronder This is waiting for next review.

stan/math/rev/core/var.hpp
```diff
@@ -415,6 +433,68 @@ class var_value {
...
+  /**
+   * Copy assignment operator delegates to general assignment operator if the
+   * var_value contains eigen type.
+   * @param other the value to assing
+   * @return this
+   */
+  var_value<T> operator=(std::conditional_t<is_eigen<value_type>::value,
+                                            const var_value<T>&, not_var_value>
+                             other) {
+    return operator=<T>(other);
```
SteveBronder, line 496: Looking over all this I think we should go back to having another template parameter for `var_value` to do SFINAE on. eod we have a lot of stuff here that is just for eigen types and some stuff that's just for doubles so I think it would make more sense. There would be some code duplication but I think things will look a lot cleaner

```
class var_value;

template <typename T>
class var_value<T, require_floating_point_t<T>> {
  // double stuff  
}

template <typename T>
class var_value<T, require_eigen_t<T>> {
  // eigen stuff  
}
```

stan/math/rev/core/var.hpp
```diff
@@ -94,6 +95,23 @@ class var_value {
...
+  template <typename S, typename T_ = T,
+            require_convertible_t<S&, value_type>* = nullptr,
+            require_plain_type_t<T_>* = nullptr,
+            require_not_same_t<T, S>* = nullptr>
+  var_value(const var_value<S>& other) : vi_(new vari_type(other.vi_->val_)) {
+    reverse_pass_callback(
+        [this_vi = this->vi_, other_vi = other.vi_]() mutable {
+          other_vi->adj_ += this_vi->adj_;
+        });
+  }
```
SteveBronder, line 113: Should we have each `operator=` have a constructor as well. If feels like it would be good for the API to have the same ones.

SteveBronder (approved):
So I think this looks good. Splitting out `var_value` can either be done in this PR or #2064

stan/math/rev/core/var.hpp
t4c1: I agree. Separate PR though.

stan/math/rev/core/var.hpp
t4c1: `require_all_plain_type_t<T, S>* = nullptr` case is handled by defailt copy constructor (T and S can in fact only be same type). We can not define copy assignment operator to do the same as that would also take non-plain types.

`require_any_not_plain_type_t<T, S>* = nullptr` case is handled by this constructor.

We need to explicitely define copy constructor just to delegate to one of the upper two cases, as copy constructor can not be templated and wiould not pich one of those two cases. So I think this is fine.

PR #1989, opened 2020-07-28
Add probability tests to check vectorized/scalar lpmfs/lpdfs are the same

bbbales2:
## Summary

This addresses the missing tests from #1861 

Edit: Breaking up this pull into pieces. I'm gonna leave this here until it's done.

So far https://github.com/stan-dev/math/pull/2039, https://github.com/stan-dev/math/pull/2041, and https://github.com/stan-dev/math/pull/2042

## Release notes

Added extra tests to check lpdfs/lpmfs evaluated with vectors produce the same results as with scalars

## Checklist

- [x] Math issue #1861 

- [x] Copyright holder: Columbia University

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [ ] the new changes are tested

Issue #1978, opened 2020-07-19
vector distribution tests should test with vectors of not all the same values


bbbales2:
## Description

There was a bug in an lcdf distribution that would have been caught if the vectorized versions of the CDFs were tested with vectors of different values (https://github.com/stan-dev/math/issues/1662).

Right now the vectorized tests build vectors of the same value (see the repeat_as_vector [test](https://github.com/stan-dev/math/blob/develop/test/prob/test_fixture_cdf.hpp#L507)).

This would have been caught if there was a test that tested vectorized vs. non-vectorized versions of the functions where it's not just a repeat of the same value (see discussion [here](https://github.com/stan-dev/math/pull/1752#issuecomment-596207740)).

#### Current Version:
v3.2.0

bbbales2:
This is a duplicate of: https://github.com/stan-dev/math/issues/1861

bbbales2:
Whoops, not quite a duplicate. This is for cdfs.

PR #2082, opened 2020-09-11
Generalize normal

t4c1:
## Summary

Generalizes normal distribution related fucntions to accept Eigen expressions.

## Tests

Tested using expression testing framework.

## Side Effects
None.

## Release notes
Generalized normal distribution related fucntions to accept Eigen expressions.

## Checklist

- [ ] Math issue #1470 

- [ ] Copyright holder: Tadej Ciglarič

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #2085, opened 2020-09-15
Add probability tests to check vectorized/scalar lpmfs/lpdfs are the same

### LINKED ISSUE #1861: Probability test framework didn't catch bug with vectorization

bbbales2:
This is the replacement for https://github.com/stan-dev/math/pull/1989

Fixes #1861 and #1978 

This also adds similar tests to prevent problems with the log ccdf and cdf functions.

## Release notes

Added more tests for vectorized probability functions

## Checklist

- [x] Math issue #1861, #1978


test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -700,6 +700,249 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
     }
   }
 
+  void test_as_scalars_vs_as_vector() {
```
t4c1, line 703: Add doxygen.

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -700,6 +700,249 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+    T_return_type single_ccdf_log
+        = TestClass.template ccdf_log<Scalar0, Scalar1, Scalar2, Scalar3,
+                                      Scalar4, Scalar5>(p0s.back(), p1s.back(),
+                                                        p2s.back(), p3s.back(),
+                                                        p4s.back(), p5s.back());
+
+    for (size_t n = 1; n < parameters.size(); n++) {
+      p0s.push_back((is_vector<T0>::value)
+                        ? get_param<Scalar0>(parameters[n], 0)
+                        : p0s.front());
```
t4c1, line 745: Why do we need different handling of scalars and vectors?

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -700,6 +700,249 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
     }
   }
 
+  void test_as_scalars_vs_as_vector() {
+    if (!any_vector<T0, T1, T2, T3, T4, T5>::value) {
```
t4c1, line 704: Why/how do all types being scalars signal there is nothing to test?

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -700,6 +700,249 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+                        : p3s.front());
+      p4s.push_back((is_vector<T4>::value)
+                        ? get_param<Scalar4>(parameters[n], 4)
+                        : p4s.front());
+      p5s.push_back((is_vector<T5>::value)
+                        ? get_param<Scalar5>(parameters[n], 5)
+                        : p5s.front());
+
+      single_ccdf_log += TestClass.template ccdf_log<Scalar0, Scalar1, Scalar2,
+                                                     Scalar3, Scalar4, Scalar5>(
```
t4c1, line 763: Why templating with scalars, not complete types of the arguments?

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -700,6 +700,249 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+
+    if (stan::math::is_inf(stan::math::value_of_rec(single_ccdf_log))
+        && stan::math::value_of_rec(single_ccdf_log)
+               == stan::math::value_of_rec(multiple_ccdf_log)) {
+      return;
+    }
+
+    EXPECT_NEAR(stan::math::value_of_rec(single_ccdf_log),
+                stan::math::value_of_rec(multiple_ccdf_log), 1e-7)
+        << "cdf_log evaluated in loop should match vectorized equivalent";
```
t4c1, line 832:
```suggestion
        << "sum of scalar cdf_logs should match vectorized equivalent";
```

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -700,6 +700,249 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+
+    size_t pos_single = 0;
+    size_t pos_multiple = 0;
+    if (!is_constant_all<T0>::value && !is_empty<T0>::value
+        && !std::is_same<Scalar0, fvar<double>>::value
+        && !std::is_same<Scalar0, fvar<fvar<double>>>::value)
```
t4c1, line 838: Why not testing the gradients of fvars?

t4c1 (changes requested):
For now I just reviewed the first file. Others look similar and I have a number of questions. So I will review the rest after this makes sense to me.

test/prob/test_fixture_ccdf_log.hpp
bbbales2: If there are no vector types tested, then this is not a test of a vectorization.

bbbales2: (And the test here is checking that vectorization works)

test/prob/test_fixture_ccdf_log.hpp
bbbales2: I'm not sure why they're templated at all. Just copy pasting.

test/prob/test_fixture_ccdf_log.hpp
bbbales2: I changed the code so your question would be easier to answer lol.

pX are the arguments that we pass to the vectorized version of the function.

pXs are the `std::vector` s of arguments that we passed to the non-vectorized version of the function.

They should be the same length.

So if argument X is a scalar, pX is a thing of length 1 and pXs is a `std::vector` of length 1.

If argument X is a vector, and the thing we're testing is written:

```
ccdf_log([a, b, c, d]) == ccdf_log(a) + ccdf_log(b) + ccdf_log(c) + ccdf_log(d)
```

Then pX is that vector `[a, b, c, d]` and pXs is a `std::vector` `[a, b, c, d]` (the concatenation of all the arguments on the right).

test/prob/test_fixture_ccdf_log.hpp
bbbales2: Good point. I got rid of these and added `fvar<fvar<double>>` to the 2nd order gradient tests.

It makes me nervous that this test framework is not tested.

test/prob/test_fixture_ccdf_log.hpp
bbbales2: I got rid of the template types. Not sure why they were there. Just copy-paste.

bbbales2:
@t4c1 ready for review again, though we might want to wait for tests to pass again since we're checking more things now that I enabled the `fvar` stuff you pointed out.

t4c1:
@bbbales2 You have some failed tests here

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -582,124 +563,170 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+          << "scalar and vectorized results should produce the same number of "
+             "third order gradients";
+
+      for (size_t i = 0; i < single_gradients1.size(); ++i)
+        EXPECT_NEAR(single_gradients1[i], multiple_gradients1[i], 1e-7)
+            << "scalar and vectorized results should have the same first order "
+               "gradients";
+      for (size_t i = 0; i < single_gradients2.size(); ++i)
+        EXPECT_NEAR(single_gradients2[i], multiple_gradients2[i], 1e-7)
+            << "scalar and vectorized results should have the same first order "
```
t4c1, line 587:
```suggestion
            << "scalar and vectorized results should have the same second order "
```

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -582,124 +563,170 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+               "gradients";
+      for (size_t i = 0; i < single_gradients3.size(); ++i)
+        EXPECT_NEAR(single_gradients3[i], multiple_gradients3[i], 1e-7)
+            << "scalar and vectorized results should have the same first order "
```
t4c1, line 591:
```suggestion
            << "scalar and vectorized results should have the same third order "
```

test/prob/test_fixture_ccdf_log.hpp
t4c1: While this simplifies the code, the error message loses the information which variable has wrong gradient.

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -582,124 +563,170 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+               "gradients";
     }
   }
 
+  /**
+   * Test that the vectorized functions work as expected when the elements
+   * of the vector are different
```
t4c1, line 598: Same as above. Explain "work as expected"

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -828,119 +700,31 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+        << "scalar and vectorized results should produce the same number of "
+           "third order gradients";
+
+    for (size_t i = 0; i < single_gradients1.size(); ++i)
+      EXPECT_NEAR(single_gradients1[i], multiple_gradients1[i], 1e-7)
+          << "scalar and vectorized results should have the same first order "
+             "gradients";
+    for (size_t i = 0; i < single_gradients2.size(); ++i)
+      EXPECT_NEAR(single_gradients2[i], multiple_gradients2[i], 1e-7)
+          << "scalar and vectorized results should have the same first order "
```
t4c1, line 722:
```suggestion
          << "scalar and vectorized results should have the same second order "
```

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -828,119 +700,31 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
+             "gradients";
+    for (size_t i = 0; i < single_gradients3.size(); ++i)
+      EXPECT_NEAR(single_gradients3[i], multiple_gradients3[i], 1e-7)
+          << "scalar and vectorized results should have the same first order "
```
t4c1, line 726:
```suggestion
          << "scalar and vectorized results should have the same third order "
```

t4c1 (changes requested):
There were a number of changes since the last review, so I am again stopping at first file.

stan/math/prim/prob/skew_normal_lpdf.hpp
```diff
@@ -78,13 +78,17 @@ return_type_t<T_y, T_loc, T_scale, T_shape> skew_normal_lpdf(
...
+    logp += log_erfc_alpha_z;
 
     T_partials_return deriv_logerf
         = TWO_OVER_SQRT_PI
-          * exp(-alpha_dbl * y_minus_mu_over_sigma / SQRT_TWO * alpha_dbl
-                * y_minus_mu_over_sigma / SQRT_TWO)
-          / (1 + erf(alpha_dbl * y_minus_mu_over_sigma / SQRT_TWO));
+          * exp(-square(alpha_dbl * y_minus_mu_over_sigma / SQRT_TWO)
+                - log_erfc_alpha_z);
+
```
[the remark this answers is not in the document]
bbbales2: What was happening originally was the denominator of this:
```
exp(-alpha_dbl * y_minus_mu_over_sigma / SQRT_TWO * alpha_dbl
                * y_minus_mu_over_sigma / SQRT_TWO)
          / (1 + erf(alpha_dbl * y_minus_mu_over_sigma / SQRT_TWO))
```

was evaluating to zero and so blowing up.

We have these two properties though:
```
1 - erf(x) = erfc(x)
erf(-x) = -erf(x)
```

and so can rewrite the denominator:
```
1 + erf(x)
1 - erf(-x)
erfc(-x)
```

That gets rid of the `1 + x` underflow situation, and we already had the value of `erfc(-x)` from the logp increment above. The rest is mostly decorative.

test/prob/discrete_range/discrete_range_ccdf_log_test.hpp
```diff
@@ -56,7 +56,11 @@ class AgradCcdfLogDiscreteRange : public AgradCcdfLogTest {
   stan::return_type_t<T_y, T_lower, T_upper> ccdf_log_function(
       const T_y& y, const T_lower& lower, const T_upper& upper, const T3&,
       const T4&, const T5&) {
-    if (y < lower || y > upper) {
+    if (y < lower) {
+      return 1.0;
```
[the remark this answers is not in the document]
bbbales2: Yes, good catch.

test/prob/gumbel/gumbel_cdf_test.hpp
```diff
@@ -34,13 +34,11 @@ class AgradCdfGumbel : public AgradCdfTest {
...
 
-    param[0] = -3.5;  // y
+    param[0] = -1.5;  // y
     param[1] = 1.9;   // mu
     param[2] = 7.2;   // beta
     parameters.push_back(param);
-    cdf.push_back(
-        0.1203922620798295861862650786832089422663975274508450);  // expected
-                                                                  // cdf
+    cdf.push_back(0.20118031381610754216);  // expected cdf
```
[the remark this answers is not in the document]
bbbales2: Should be able to use:

```
mu = 1.9
beta = 7.2
cdf = function(x) { exp(-exp(-(x - mu) / beta)) }
cdf(-1.5)
```

The Gumbel parameterization on Wikipedia is the same one we use (https://en.wikipedia.org/wiki/Gumbel_distribution)

test/prob/neg_binomial/neg_binomial_test.hpp
```diff
@@ -75,8 +69,6 @@ class AgradDistributionsNegBinomial : public AgradDistributionTest {
     using stan::math::multiply_log;
     using std::log;
 
-    if (alpha > 1e10)
-      return -lgamma(n + 1.0) + multiply_log(n, alpha / beta) - alpha / beta;
```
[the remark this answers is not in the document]
bbbales2: I deleted the test values above because they were causing tests to fail with `as_scalars_vs_as_vector` (the new tests).

I'm pretty sure the reason they were failing is because:

1. The test is testing crazy large lpmf values that are at the limit of the numerics.

2. The new `as_scalars_vs_as_vector` stuff mixes and matches parameters from different sets of parameters.

For instance, if parameter 1 is a vector it will get all possible values of parameter 1, but if parameter 2 is a scalar, it will only get one value (and so different parameter 1s and parameter 2s get evaluated together). This is harder numerically than the hard numerical example, and things break.

----

I removed the branching logic in the tests cause that was only for very large parameters. This sort of branching used to exist in the regular implementations but @martinmodrak found a way to fix the numerics and avoid this . I think I did the `neg_binomial` changes for this: https://github.com/stan-dev/math/commit/56988042310b7d28eaddbf2c4fde2799884a3339#diff-43a81d5aa08039bf4bdfa75818ff97b8

I can add gradient checks at those large parameter values easily enough. 

----

I can also roll back `as_scalars_vs_as_vectors` I think. The changes I made to how the gradients are compared is what made the parameter shuffling happen that causes this.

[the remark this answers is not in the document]
bbbales2: And if I roll back the checks then this test can stay.

test/prob/test_fixture_ccdf_log.hpp
```diff
@@ -582,124 +563,170 @@ class AgradCcdfLogTestFixture : public ::testing::Test {
...
-            is_vector<T4>::value, single_ccdf_log, single_gradients3,
-            pos_single, multiple_gradients3, pos_multiple, N_REPEAT);
-      if (!is_constant_all<T5>::value && !is_empty<T5>::value
-          && std::is_same<Scalar5, fvar<fvar<var>>>::value)
-        test_multiple_gradient_values(
-            is_vector<T5>::value, single_ccdf_log, single_gradients3,
-            pos_single, multiple_gradients3, pos_multiple, N_REPEAT);
+      EXPECT_EQ(single_gradients1.size(), multiple_gradients1.size())
+          << "scalar and vectorized results should produce the same number of "
+             "first order gradients";
```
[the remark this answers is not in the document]
bbbales2: Hmm, explain more. I'm pretty sure this is making sure the relationship:

```
lccdf(a) + lccdf(b) + lccdf(c) + ... = lccdf([a, b, c, ...])
```

holds. What do you mean by repeat scalar gradients?

> Also I think this can't fail due to a bug in the tested distribution

Well it definitely failed some of the tests to get this greenlit lol. I do need to check though that this actually would have prevented the original problem.

test/prob/test_fixture_ccdf_log.hpp
bbbales2: This error information wasn't printed anywhere anyway.

test/prob/neg_binomial/neg_binomial_test.hpp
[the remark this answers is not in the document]
bbbales2: Nevermind I think this stays like this.

If we're testing mixes of scalar/vector types, we're going to end up in a situation like this where two sets of parameters just shuffled together and we get weird unexpected tests.

The way to get around this is only test repeated things, which is what was already being tested in the test framework, but was not sufficient to catch the original bugs.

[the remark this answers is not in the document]
t4c1: If this used to be tested on two parameter sets, I would like it to stay that way. Can you not find a second parameter set that mixes with the first one without numerical issues?

test/prob/test_fixture_ccdf_log.hpp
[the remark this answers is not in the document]
t4c1: If I understand correctly what you are doing in this test is (d means derivative and vec variables in the following are scalars, this just means they are used in vector implementation):
`lccdf(a_scalar) + lccdf(b_scalar) + lccdf(c_scalar)`
and
`lccdf([a_vec, b_vec, c_vec, ...])`

And this expect is checking that:
`{d_a_scalar, d_b_scalar, d_c_scalar}.size() == {d_a_vec, d_b_vec, d_c_vec}.size()`, which is obviously true, regardless of what `lccdf()` does.

test/prob/test_fixture_ccdf_log.hpp
t4c1: Oh, right I added it in https://github.com/stan-dev/math/pull/2082, but it is not merged yet. Could you print it? That is very useful info for debugging. (since it is not in yet lets make this optional)

martinmodrak: Just a minor note: you might want to use `expect_near_rel` for inexact comparisons - we thought a bit about good logic for this type of comparisons and also it works out of the box for vector/matrix arguments (see e.g. https://github.com/stan-dev/math/issues/1656)

test/prob/test_fixture_ccdf_log.hpp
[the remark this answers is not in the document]
bbbales2: Oh right. I guess I just added these checks to sanity check the function is actually doing the right thing. I wouldn't leave them if this test framework was tested but since it isn't I kinda want to just leave them as a sanity thing.

test/prob/neg_binomial/neg_binomial_test.hpp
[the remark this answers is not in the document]
bbbales2: I switched to expect_near_rel and was able to add this back.

test/prob/test_fixture_ccdf_log.hpp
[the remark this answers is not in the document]
bbbales2: Don't need this check with expect_near_real.

test/prob/test_fixture_ccdf_log.hpp
bbbales2: @martinmodrak switched to expect_near_rel and life did get easier, thanks for rec.

test/prob/logistic/logistic_test.hpp
```diff
@@ -19,9 +19,9 @@ class AgradDistributionsLogistic : public AgradDistributionTest {
 
     param[0] = -1.0;  // y
     param[1] = 0.2;   // mu
-    param[2] = 0.25;  // sigma
+    param[2] = 0.5;   // sigma
     parameters.push_back(param);
-    log_prob.push_back(-3.430097773556644469295);  // expected log_prob
+    log_prob.push_back(-1.8805251237479538862);  // expected log_prob
```
bbbales2, line 24: I tried to made this easier to avoid a precision problem.

Here's how you verify the new value:
```
options(digits = 20)
dlogis(-1, location = 0.2, scale = 0.5, log = TRUE)
```

The error message was something like:
```
The difference between x1 and x2 is 1.1990408665951691e-14, which exceeds tol_val, where
x1 evaluates to 8.2156503822261584e-15,
x2 evaluates to -3.7747582837255322e-15, and
tol_val evaluates to 1e-14.
expect_near_rel_finite in: expect_near_rel; requite items x1[i] = x2[i]: scalar and vectorized results should have the same second order gradients
```

I just didn't want to turn the tolerances in one place down for this.

bbbales2 (commented):
Comment

test/prob/neg_binomial/neg_binomial_test.hpp
```diff
@@ -24,11 +24,11 @@ class AgradDistributionsNegBinomial : public AgradDistributionTest {
...
 
-    param[0] = 13;
-    param[1] = 1e11;
-    param[2] = 1e10;
+    param[0] = 10;
+    param[1] = 1e6;
+    param[2] = 1e5;
     parameters.push_back(param);
-    log_prob.push_back(-2.6185576442208003);  // expected log_prob
+    log_prob.push_back(-2.0785666431081630812);  // expected log_prob
```
bbbales2, line 31: I made these things easier so the test tolerances could stay high.

This is code to verify:
```
options(digits = 20)
x = 10
alpha = 1e6
beta = 1e5
n = alpha
p = beta / (beta + 1)
dnbinom(x, n, p, log = TRUE)
```

I added 1e11/1e10 tests to unit/math/mix since these high values were clearly here for a reason. This will do low precision checks at the high parameter values.

bbbales2 (commented):
Comment

bbbales2:
Green check green check green check! Man that feels good

t4c1 (approved):

PR #2087, opened 2020-09-18
Generalize ordered

t4c1:
## Summary

Generalizes ordered distributions related functions to accept Eigen expressions.

## Tests

Tested using expression testing framework.

## Side Effects
None.

## Release notes

Generalized ordered distributions related functions to accept Eigen expressions.

## Checklist

- [x] Math issue #1470 


test/unit/math/rev/fun/from_var_value_test.cpp
```diff
@@ -6,7 +6,7 @@ TEST(AgradRevMatrix, from_var_value_matrix_test) {
   Eigen::MatrixXd val(2, 3);
   val << 1, 2, 3, 4, 5, 6;
   Eigen::MatrixXd adj(2, 3);
-  val << 4, 5, 6, 7, 8, 9;
+  adj << 4, 5, 6, 7, 8, 9;
```
bbbales2, line 9: Good catch

stan/math/fwd/meta/value_type.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include <stan/math/prim/meta/is_fvar.hpp>
+#include <stan/math/fwd/core/fvar.hpp>
+
+#include <type_traits>
+
+namespace stan {
+
+template <typename T>
+struct value_type<T, require_fvar_t<T>> {
+  using type = typename std::decay_t<T>::Scalar;
```
bbbales2, line 14: So now a `value_type` of an `fvar<T>` is `T` if I'm reading this correctly. Is this different than it was before?

bbbales2: I'm assuming not because then presumably lots of tests would break. Did we just not use `value_type` on `fvar` s?

I guess also what is this change for. Just so I understand

bbbales2 (changes requested):
Looks good. I left a question on the value_type

stan/math/fwd/meta/value_type.hpp
t4c1: > Did we just not use value_type on fvar s?

So it seems (or we did not notice we do not get back what we would expect).

stan/math/prim/prob/ordered_logistic_lpmf.hpp
```diff
@@ -69,17 +69,21 @@ template <bool propto, typename T_y, typename T_loc, typename T_cut>
...
-  static const char* function = "ordered_logistic";
-
   using T_partials_return = partials_return_t<T_loc, T_cut>;
-  using T_partials_vec = typename Eigen::Matrix<T_partials_return, -1, 1>;
-
-  scalar_seq_view<T_loc> lam_vec(lambda);
-  scalar_seq_view<T_y> y_vec(y);
-  vector_seq_view<T_cut> c_vec(c);
+  using T_partials_array = Eigen::Array<T_partials_return, -1, 1>;
+  using T_cuts_val = value_type_t<scalar_type_t<T_cut>>;
```
t4c1, line 74: `value_type_t` on a potentially `fvar` input is used here for example.

bbbales2: Isn't something like `partials_return_t` the normal way of doing this?

t4c1: Hmm, might be. I am not that familiar with metaprograms having partials or return type in the name. Those names all seem confusing to me.

stan/math/fwd/meta/value_type.hpp
bbbales2: I think delete the value_type thing now that it's unused. Don't want to accidentally make a decision we don't need in these weird template type things.

bbbales2: Can we you delete the value_type specialization for fvar since it isn't used now? That's all I'm waiting on here.

bbbales2 (approved):
Good!

PR #2090, opened 2020-09-18
add a list of functions with no math implementations

### LINKED ISSUE #2079: multiply_log deprecated for lmultiply, but lmultiply is not defined in Math
## Description
The documentation for ```multiply_log``` states:
> real multiply_log(real x, real y)
Warning: This function is deprecated and should be replaced with lmultiply

But there's no ```lmultiply``` header in Math, only ```multiply_log```. Additionally, Stanc3 transpiles any ```lmultiply``` calls to ```multiply_log```, which makes it seem like ```lmultiply``` is the deprecated function.

Should we add an ```lmultiply``` header and just have ```multiply_log``` call that? Like we do with the ```_log``` and ```_lpdf``` functions.

#### Current Version:
v3.3.0

rok-cesnovar:
Fixes #2079 

## Summary

This PR fixes expression tests for the `lmultiply` function. The issue is that stanc3 supports lmultiply but there is no Math implementation of it as stanc3 transpiles lmultiply as multiply_log. We need to cover these exceptions in the expression tests.

## Tests

This is fixing tests.

## Side Effects

/

## Checklist

- [x] Math issue #2079

### DIFF
--- a/test/expressions/generateExpressionTests.py
+++ b/test/expressions/generateExpressionTests.py
@@ -11,9 +11,7 @@ else:
 
 src_folder = "./test/expressions/"
 build_folder = "./test/expressions/"
-exceptions_list_location = (
-    "./test/expressions/stan_math_sigs_exceptions.expected"
-)
+exceptions_list_location = "./test/expressions/stan_math_sigs_exceptions.expected"
 
 eigen_types = ["matrix", "vector", "row_vector"]
 arg_types = {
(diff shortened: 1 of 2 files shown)


rok-cesnovar:
This is currently breaking develop so if @andrjohns @t4c1 or @bbbales2  or anyone else have a minute to review this that would be great. Thanks!

test/expressions/generateExpressionTests.py
```diff
@@ -14,6 +14,10 @@
 exceptions_list_location = (
     "./test/expressions/stan_math_sigs_exceptions.expected"
```
bbbales2, line 15: I think add these function exceptions to `./test/expressions/stan_math_sigs_exceptions.expected`

No need to make a separate list.

bbbales2 (changes requested):
Changes requested! (though I posted the thing as a separate comment)

test/expressions/generateExpressionTests.py
rok-cesnovar: I actually asked @t4c1 about this one and he suggested to go this route. Mainly for two reasons: 
- the exception lists carries the list of sigs that dont support expressions. lmultiply does support it, its just different in that is has no direct Math implementation.Analysis_and_optimization-
- the plan is to eventually get rid of the file by supporting expressions everywhere

rok-cesnovar: I could still be persuaded otherwise. Just think those are both valid points.

bbbales2: I still think stick with the single file for exceptions.

I think there are already functions not in stan math that are in there: https://github.com/stan-dev/math/blob/develop/test/expressions/stan_math_sigs_exceptions.expected#L10

> the plan is to eventually get rid of the file by supporting expressions everywhere

I think we just change the meaning of this file to be all the signatures we can't test in Math because either they are compatible (and this meaning will go away) or they don't exist (cause they're Stan things)

rok-cesnovar: Ok, will add the sigs to the file and we can change the meaning once we clean out the non-compatible ones.

The sigs you linked are actually a bug and should not be listed: https://github.com/stan-dev/stanc3/issues/693

bbbales2:
Looks good. I'll approve if the tests pass (and ping me if they pass and I haven't)

andrjohns:
I discovered last night theres the same issue with the ```lchoose``` function - defined in stance but not Math

rok-cesnovar:
> I discovered last night theres the same issue with the `lchoose` function - defined in stance but not Math

That is correct, `lmultiply` and `lchoose` are handled inside the compiler. But lhooce does not accept any Eigen vectors or matrices (only `real lchoose(real, real)`).

@bbbales2 the expression tests are now passing so feel free to approve.

bbbales2 (approved):

PR #2092, opened 2020-09-21
Optimize distribution tests

rok-cesnovar:
## Summary

Currently, the N_TESTS on Jenkins is set to 300. By changing that to 1500 we can reduce the number of generated files. 

Side note: the name N_TESTS is a bit misleading, it serves more as a N_TESTS_PER_FILE or batch size.

Lets first see how much we would gain and if the tests pass.

## Tests

Yes, this only touches tests.

## Side Effects

None, if N_TESTS does what I think it does.

## Release notes

## Checklist


rok-cesnovar:
This mostly led to nowhere so I am closing this PR. Posting some info on what I found out.

With larger test files we burn more RAM and swap and we run out of disk space. We can decrease the parallel setting (-j) but then that defeats the purpose of speeding this up. So its not really usable and increasing the AWS instances resources is not something we want to do imo. So the non-row-vector tests running for 1h45min and develop tests running for 6-7 hours is what we will just have to deal with for now.

The only way of decreasing the testing burden would maybe be to only test for either Eigen vector or std::vectors (not both) on PRs but I dont have enough confidence in saying how that would play out. Would that cause a lot errors that would only get caught on develop tests? If someone has a better idea on this, that would be one way of moving forward.

Issue #2095, opened 2020-09-23
quad_form_sym: reverse specialization does not return symmetric matrix


syclik:
## Description

`quad_form_sym` should return a symmetric matrix. See https://mc-stan.org/docs/2_24/functions-reference/dot-products-and-specialized-products.html

For reverse mode, there are instances where it can return a non-symmetric matrix. The primitive version for the same values of matrices results in a symmetric matrix.

## Example

pseudocode, but pulled from C++ example
```
A =  1.25882993696386514 -0.03325949401909023
-0.03325949605426004  1.85523447220884385
B =  2.55474619740069508  0.66362927717720988
-1.92917223922387349  2.01853255256721731

M = quad_form_sym(A, B);
M(1, 0) - M(0, 1) = 1.31006e-08 // this should be 0 and is 0 for primitive

```

## Expected Output

```
M(1, 0) - M(0, 1) = 0
```

#### Current Version:
v3.3.0

syclik:
I'm working on branch `bugfix/issue-2095-quad_form_sym` and something is off.

In the primitive version, it's not picking up that the input matrix is not symmetric, but it is for the reverse version. I'm going to dig into `check_symmetric`.

syclik:
thanks, @kmckee90, for providing an example that triggered this behavior!

PR #2096, opened 2020-09-23
Bugfix/issue 2095 quad form sym

syclik:
## Summary

Updates the reverse mode implementation of `quad_form_sym` so that it returns a symmetric matrix. The current code does not do that.

## Tests

There is a new test for `quad_form_sym`. It has a numerical example that fails to return a symmetric matrix in reverse mode.

Unit tests were added to `prim` and `mix` to check the 2x2 example is symmetric. The test fails on develop for the mix test, but passes on this branch.

## Side Effects

None.

## Release notes

- fixes numerical issue with `quad_form_sym`.

## Checklist

- [x] Math issue #2095


syclik:
Anyone available to take a look? I kept the diff down. For reference, see the `prim` version of `quad_form_sym(...)` here: https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/quad_form_sym.hpp#L36

rok-cesnovar:
I can take a look later today.

rok-cesnovar (approved):
Thanks!

test/unit/math/mix/fun/quad_form_sym_test.cpp
```diff
@@ -64,3 +64,40 @@ TEST(MathMixMatFun, quadFormSym) {
...
+  EXPECT_FLOAT_EQ(0, cv(1, 0).val() - cv(0, 1).val());
+
+  stan::math::recover_memory();
+
+  //--------------------
+
+  Eigen::MatrixXd ad(2, 2);
+  Eigen::MatrixXd bd(2, 2);
+
+  ad << 1.25882993696386514, -0.03325949401909023, -0.03325949605426004,
```
bbbales2, line 88: This matrix isn't symmetric.

test/unit/math/mix/fun/quad_form_sym_test.cpp
```diff
@@ -64,3 +64,40 @@ TEST(MathMixMatFun, quadFormSym) {
...
+
+  bv << 2.55474619740069508, 0.66362927717720988, -1.92917223922387349,
+      2.01853255256721731;
+
+  Eigen::Matrix<stan::math::var, -1, -1> cv = stan::math::quad_form_sym(av, bv);
+  EXPECT_FLOAT_EQ(0, cv(1, 0).val() - cv(0, 1).val());
+
+  stan::math::recover_memory();
+
+  //--------------------
```
bbbales2, line 83: We don't do separators like this in other places so unless you're particularly attached to it, delete.

stan/math/rev/fun/quad_form.hpp
```diff
@@ -20,13 +20,13 @@ class quad_form_vari_alloc : public chainable_alloc {
...
+      Cd = M;
+    }
     for (int j = 0; j < C_.cols(); j++) {
       for (int i = 0; i < C_.rows(); i++) {
-        if (sym_) {
-          C_(i, j) = var(new vari(0.5 * (Cd(i, j) + Cd(j, i)), false));
-        } else {
-          C_(i, j) = var(new vari(Cd(i, j), false));
-        }
+        C_(i, j) = var(new vari(Cd(i, j), false));
```
bbbales2, line 29: Why does this work and the previous thing didn't? They look equivalent to me.

stan/math/rev/fun/quad_form.hpp
```diff
@@ -20,13 +20,13 @@ class quad_form_vari_alloc : public chainable_alloc {
   inline void compute(const Eigen::Matrix<double, Ra, Ca>& A,
                       const Eigen::Matrix<double, Rb, Cb>& B) {
     matrix_d Cd = B.transpose() * A * B;
+    if (sym_) {
+      matrix_d M = 0.5 * (Cd + Cd.transpose());
+      Cd = M;
```
bbbales2, line 25:
```
Cd = 0.5 * (Cd + Cd.transpose()).eval();
```

bbbales2 (commented):
Oops we were reviewing in parallel it looks like. I have questions anyway.

test/unit/math/mix/fun/quad_form_sym_test.cpp
bbbales2: Ignore this now.

test/unit/math/mix/fun/quad_form_sym_test.cpp
rok-cesnovar: Line 99 makes it symmetric.

stan/math/rev/fun/quad_form.hpp
syclik: Ah... I missed the `.eval()`; I did that first (without the `.eval()`) and got non-sense, so I just assigned it to a temporary and moved on. 

I can open a new PR to change that.

stan/math/rev/fun/quad_form.hpp
syclik: This part of the code is a refactor; there should be nothing different there other than removing an if statement within the loops. 

The part that’s different is in construction of the `Cd` matrix. There was a flag that was unused. Instead of adding a new vari subclass, I just used the one that was already written and it matches the prim implementation.

test/unit/math/mix/fun/quad_form_sym_test.cpp
syclik: This was an actual matrix in the wild that triggered the bug. I noticed that non-symmetry too, but it’s within tolerance, so it’s ok. This matrix gets used directly in `prim` and gets checked for symmetry.

stan/math/rev/fun/quad_form.hpp
bbbales2: It's fine now, this does the same thing.

test/unit/math/mix/fun/quad_form_sym_test.cpp
bbbales2: Oh I didn't realize symmetric was only symmetric to an absolute tolerance of 1e-8.

I wonder if this is the source of the sporadic symmetric errors that show up in models. From here: https://discourse.mc-stan.org/t/is-it-possible-to-disable-symmetry-checks/6544

Stuff like:
```
Chain 13: Exception: Exception: quad_form_sym: A is not symmetric. A[56,74] = -4.43735e+07, but A[74,56] = -4.43735e+07  (in 'modelb0b43b4c742c_model6_marginalize_interp' at line 178)
```

There are some other symmetry checks in that thread that I am surprised are failing though, so maybe it's not the whole story.

I suspect if we switched to using the tolerance stuff from the test framework this would get rid of most of these sporadic problems: https://github.com/stan-dev/math/blob/develop/test/unit/math/expect_near_rel.hpp .

syclik: > Oh I didn't realize symmetric was only symmetric to an absolute tolerance of 1e-8.

Yup. We can't really do `==` on floating point numbers.

> There are some other symmetry checks in that thread that I am surprised are failing though, so maybe it's not the whole story.

I'd guess there's more to it than just the symmetry checks.

> I suspect if we switched to using the tolerance stuff from the test framework this would get rid of most of these sporadic problems: https://github.com/stan-dev/math/blob/develop/test/unit/math/expect_near_rel.hpp .

+1! That'd be great. It really should be a relative tolerance instead of an absolute tolerance. But this brings up a practical issue. If functions assume symmetry, but use both halves of the matrix, then the errors could accumulate. There are different approaches we can take to clean this up...

bbbales2: > There are different approaches we can take to clean this up

Expand on this.

I want to open an issue for making check_symmetric use `expect_near_real` but before I do that I wanna know about these other things.

PR #2099, opened 2020-09-24
Add OpenCL rev (var_value) implementations of matrix multiply and sum

t4c1:
## Summary

Adds OpenCL reverse mode implementations of matrix multiply and sum (sum is required for easy testing a function that returns non-scalar value).

This also includes a small bugfix for kernel generator transpose.

## Tests
Both multiply and sum are tested.

## Side Effects
Comparing to existing rev matrix multiply this implementation leaves data in the OpenCL device between forward and reverse pass. This can be faster as it avoid copying, but a model that also uses other OpenCL functions could run out of device memory at smaller problem sizes.

## Release notes
Added OpenCL reverse mode implementations of matrix multiply and sum.

## Checklist


test/unit/math/opencl/rev/multiply_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+TEST(OpenCLMatrixMultiply, prim_rev_values_small) {
+  int N = 2;
+  int M = 3;
+  int K = 4;
+
+  Eigen::MatrixXd a(N, M);
+  a << 1, 2, 3, 4, 5, 6;
+  Eigen::MatrixXd b(M, K);
+  b << 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1;
+  stan::math::test::compare_cpu_opencl_prim_rev(matrix_multiply_functor, a, b);
```
rok-cesnovar, line 19: As far as I understand, the tests only tests multiply(matrix_cl, matrix_cl) and multply(var_value, var_value). We need to also test multply(matrix_cl, var_value) multply(var_value, matrix_cl).

rok-cesnovar (changes requested):
Looks great. These OpenCL rev implementations are so clean.

I just have a few comments on testing.

stan/math/opencl/rev/multiply.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * @tparam T_b type of second expression
+ * @param a first expression
+ * @param b second expression
+ * @return Matrix product of given arguments
+ */
+template <
+    typename T_a, typename T_b,
+    require_all_nonscalar_prim_or_rev_kernel_expression_t<T_a, T_b>* = nullptr,
+    require_any_var_t<T_a, T_b>* = nullptr>
+inline auto operator*(const T_a& a, const T_b& b) {
```
SteveBronder, line 29: I think the compiler needs access to the `multiply()` function not `*`

SteveBronder (commented):
Would it be possible to add the opencl version to the test suite? If we want the opencl version to be swappable with the eigen version I think it would be better/safer if they all had the same tests

stan/math/opencl/rev/multiply.hpp
rok-cesnovar: Oh yeah, we need both of these.

t4c1:
Every test is a part of a test suite. Do you mean could they use the ad testing framework? Hmm, that would require some extensions to the framework, as the OpenCL implementations do not have fwd implementations. Also they should be checked that they match the Eigen results. This would be a significant amount of work, but it might be worth doing. Not a part of this PR.

SteveBronder:
Yes what I mean is [here](https://github.com/stan-dev/math/blob/develop/test/unit/math/test_ad.hpp#L1810) we could add a `test_matvar_opencl()`. We could either enforce that every function that accepts `var<Eigen>` also accepts `var<matrix_cl>` or have a bool template parameter that says whether or not a `var<matrix_cl>` version is available. I personally prefer the first choice as I think it will be worth it in the long run

SteveBronder:
Those functions in the ad test suite just check that `var<matrix>` and `matrix<var>` return the same results so we don't need an fvar extension

t4c1:
>I personally prefer the first choice as I think it will be worth it in the long run

I think that would force anyone who implements a var_value overload to also implement a GPU overload. We have a number of non-GPU devs and some functions that can not be easily implemented using kernel generator and would need new kernels. I think that would be a bad idea.

I am not that familiar with the ad testing framework. If you have an idea how it can be done it might be better if you do it.

SteveBronder:
> I think that would force anyone who implements a var_value overload to also implement a GPU overload. We have a number of non-GPU devs and some functions that can not be easily implemented using kernel generator and would need new kernels. I think that would be a bad idea.

Yeah that's fair

> I am not that familiar with the ad testing framework. If you have an idea how it can be done it might be better if you do it.

Yeah I can take a crack at it I'll make a branch off of this one to work on it.

One Q, for `multiply(var_value<matrix_cl<double>> a, var_value<Eigen::MatrixXd>)` should the second argument get promoted / moved over to a `var_value<matrix_cl<double>>`?

t4c1:
Nope. Overloads with mix of OpenCL types and Eigen/std vector types are not supported. It is up to compiler to move all the arguments to either OpenCL device or host in such case.

rok-cesnovar:
Will approve once the tests pass. 

Gregor is preparing me a model that uses a poission GLM and a multiplication for the X argument, where X is a parameter so we can then see test the speedup when a successive multiply and GLM are both run on the GPU without transfers back in between.

t4c1:
Extending the testing to all combinations of prim and rev arguments uncovered a number of bugs in GLMs, including one in Eigen implementation.

rok-cesnovar (approved):
This looks cool beans to me. Would also like  @SteveBronder to approve this one, given that you had some comments. Thanks!

SteveBronder (approved):
lgtm!

Issue #2101, opened 2020-09-24
Make functions with custom autodiff `var<mat>` friendly


bbbales2:
## Description

This issue is part of https://github.com/stan-dev/math/issues/1805.

The goal is to convert every function with custom reverse mode autodiff to accept `var_value<Eigen::Matrix<T, R, C>>` types as arguments. The implementations should probably be based off https://github.com/stan-dev/math/issues/2018 (it's fine if they beat that in, I can just remove them from #2018).

Edit this list and check off functions as they're implemented in develop:
- [x] softmax
- [x] sum
- [x] multiply
- [ ] dot_self
- [ ] columns_dot_self
- [ ] determinant
- [ ] inverse
- [ ] log_determinant
- [ ] multiply_lower_tri_self_transpose
- [ ] tcrossprod
- [ ] matrix_power
- [ ] log_softmax
- [ ] log_sum_exp
- [ ] log_determinant_spd
- [ ] ordered_constrain
- [ ] columns_dot_product
- [ ] cov_exp_quad
- [ ] divide
- [ ] dot_product
- [ ] gp_exp_quad_cov
- [ ] gp_periodic_cov
- [ ] mdivide_left
- [ ] mdivide_left_spd
- [ ] rows_dot_prod
- [ ] mdivide_left_tri
- [ ] quad_form
- [ ] trace_quad_form
- [ ] trace_gen_quad_form
- [ ] mdivide_left_ldlt
- [ ] trace_gen_inv_quad_form_ldlt
- [ ] trace_inv_quad_form_ldlt
- [ ] positive_ordered_constrain
- [ ] simplex_constrain
- [ ] unit_vector_constrain
- [ ] sd
- [ ] variance
- [ ] squared_distance
- [ ] log_determinant_ldlt
- [ ] log_softmax
- [ ] log_sum_exp

#### Current Version:
v3.3.0

bbbales2:
I updated the benchmark thing, now with some more tools for figuring out where slowdowns are: https://bbbales2.shinyapps.io/math-shiny-perf/

At the top you pick your branch. It goes [branch_name]_[git_commit]_[whether it is `var<mat>` or `mat<var>` based]

Then there's a big table showing you, across all `n`, what is the minimum and maximum speedups for a function so you can sort them (it also shows percent test with a positive speedup, so 100% means all cases are faster, 50% means half the cases are faster).

[...]

PR #2105, opened 2020-09-25
var<matrix> Cholesky decompose

SteveBronder:
## Summary

Uses `reverse_pass_callback` for cholesky_decompose and adds a `var<mat>` specialization. 

This also as a sidenote adds two little slice of life features in the testing framework. I found it frustrating when I had a matrix fail and the tests just spit out the values without telling me which. So I added cell row/col numbers to the test failure message and added a macro `STAN_TEST_PRINT_MATRIX_FAILURE` where if the user has it defined it will print out the full matrices if the test has a failure.

Note also I didn't touch the OpenCL version. I think for that version we want to make a copy and then pass over the packed values. Or would it be more efficient to just reuse those values like the other methods and do a full copy? If so I can have the `matrix_cl` version also use a `reverse_pass_callback`.

## Tests

So we weren't actually testing the blocked version of cholesky. In the testing code we use this lambda to generate the cholesky test

```cpp
auto f(int dof) {
  return [=](const auto& x) {
     Eigen::MatrixXd y = stan::math::cov_matrix_constrain(x, dof);
    return stan::math::cholesky_decompose(y);
  };
}
```

but that was failing because of the wrong sizes being sent to `cov_matrix_constrain`. Since the tests pass if all scalar types fail on the test this always failed and therefore always passed. I modified it to

```cpp
auto f(int dof) {
  return [=](const auto& x) {
    stan::math::promote_scalar_t<stan::value_type_t<decltype(x)>, Eigen::Matrix<double, -1, -1>> y;
    try {
      y = stan::math::cov_matrix_constrain(x, dof);
    } catch (...) {
      ADD_FAILURE() << "FAILED AT COV_MATRIX_CONSTRAIN";
      throw;
    }
    return stan::math::cholesky_decompose(y);
  };
}
```

Which will always add a failure to the tests for `cov_matrix_constrain` failing. Fixing that the tests pass but need to have a lower tolerance for `hessian hessian` and `hessian fvar hessian`. I think this may be due to not having enough "wiggle" on the diagonal for the test matrices. But I think it's fine to just increase the tolerance.

Besides that the standard `expect_ad_matvar` checks were added to the cholesky tests in mix

## Side Effects

None

## Release notes

Add's cholesky decomposition for var<matrix> types.

## Checklist

- [x] Math issue #2101 


SteveBronder:
Also here are a few benchmarks. Though tbh so much of the time is dominated by the actual algorithm idt we get much of a speedup here even with var<matrix>. Like the relative performance looks very nice for `var<matrix>` but in terms of real wall time it's not a "wow"

## `matrix<var>`

image

image

## `var<matrix>`

image

image

SteveBronder:
Also is it possible to write the unblocked version in column major order? I think that would help a lot with small N because right now we jump around a lot

https://github.com/stan-dev/math/pull/2105/files#diff-31afaf3649128692cf55ac42a8fba5efR231

bbbales2:
Not totally sure. When I implemented this in the other pull I just kept the row-major thing: https://github.com/stan-dev/math/blob/d309342549c525a561c18f8e214144c92f954183/stan/math/rev/fun/cholesky_decompose.hpp#L556 but what I wrote is slower than develop.

For the big Cholesky I just hijacked the OpenCL tests in Rev: https://github.com/stan-dev/math/pull/2019/files#diff-730d40a53ae490720ac615195d3f4833 . It'd probably be better if we broke these two functions out into inner and just did all the same tests with them in mix.

Btw, can I get help with the OpenCL build over here https://github.com/stan-dev/math/pull/2019 ? I tried installing OpenCL on my computer but it's saying zero devices with clinfo and I was too lazy to figure out any further than that.

bbbales2:
Actually could we go good access patterns if we just incremented adjAtranspose in a separate variable and then just copied in to adjA?

SteveBronder:
Yeah sure I'll take a look at 2019. Oh yeah the transpose is a good idea! We can just do a transposeinplace and I think that would be fine. I'll play around with that

SteveBronder:
Hmm I'm actually not sure how much paying for the transpose is worth it, cuz this only works on max 35x35 matrices so worst case we have 9.8kb of mem which will be all in cache anyway.

bbbales2 (changes requested):
Questions but this looks good if all the questions have answers.

stan/math/prim/fun/cov_matrix_constrain.hpp
```diff
@@ -35,9 +35,9 @@ cov_matrix_constrain(const T& x, Eigen::Index K) {
   using std::exp;
 
   Matrix<value_type_t<T>, Dynamic, Dynamic> L(K, K);
-  check_size_match("cov_matrix_constrain", "x.size()", x.size(),
-                   "K + (K choose 2)", (K * (K + 1)) / 2);
   const Eigen::Ref<const plain_type_t<T>>& x_ref = x;
+  check_size_match("cov_matrix_constrain", "x.size()", x_ref.size(),
```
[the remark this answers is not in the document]
SteveBronder: When I use a `*_ref` I like to use it for the entire function. There shouldn't be a difference between the two though. I just think it looks more consistent

stan/math/prim/fun/cov_matrix_free.hpp
```diff
@@ -35,17 +36,17 @@ namespace math {
...
-  check_nonzero_size("cov_matrix_free", "y", y);
+  const auto& y_ref = to_ref(y);
+  check_square("cov_matrix_free", "y", y_ref);
+  check_nonzero_size("cov_matrix_free", "y", y_ref);
 
   using std::log;
-  int K = y.rows();
-  const Eigen::Ref<const plain_type_t<T>>& y_ref = y;
+  int K = y_ref.rows();
   check_positive("cov_matrix_free", "y", y_ref.diagonal());
```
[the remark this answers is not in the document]
SteveBronder: Is there some performance loss for doing the `to_ref()` sooner?

stan/math/rev/fun/cholesky_decompose.hpp
```diff
@@ -394,43 +213,146 @@ cholesky_decompose(const T& A) {
...
-      accum += j;
-      accum_i = accum;
     }
+  }
+  if (L_A.rows() <= 35) {
+    reverse_pass_callback([L_A, L, A_ref]() mutable {
+      size_t N = A_ref.rows();
+      int pos = (N * (N + 1) / 2) - 1;
+      auto adjL = L.adj().eval();
+      // TODO(Steve): Write this in column major order
```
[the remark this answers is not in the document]
SteveBronder: Yeah that's fair. I was hoping I was smart enough to figure this out but that was a no go lol

stan/math/prim/fun/cov_matrix_free.hpp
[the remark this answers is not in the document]
bbbales2: check_square and check_nonzero_size don't need y dereferenced to fail.

Practically I don't think we'd see any performance difference (if something is failing size checks in Stan, it's probably gonna do it 100 times and error out pretty quickly).

stan/math/prim/fun/cov_matrix_constrain.hpp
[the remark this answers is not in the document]
bbbales2: Same comment as below. I could go either way on this. @t4c1 can weigh in.

stan/math/rev/fun/cholesky_decompose.hpp
```diff
@@ -394,43 +213,146 @@ cholesky_decompose(const T& A) {
...
+          }
+          for (int k = j - 1; k >= 0; --k) {
+            adjL.coeffRef(i, k) -= A.adj().coeff(i, j) * L.val().coeffRef(j, k);
+            adjL.coeffRef(j, k) -= A.adj().coeff(i, j) * L.val().coeffRef(i, k);
+          }
+        }
+      }
+    });
+  } else {
+    reverse_pass_callback([L, A]() mutable {
```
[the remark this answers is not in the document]
SteveBronder: Good catch yes!

SteveBronder:
@rok-cesnovar I was reading over the cholesky autodiff paper's from giles and mxnet, do we need to make a copy of L's adjoints or can we work on them inplace? it feels like the answer is a strong "yes we need to make a copy" but if not that would be nice

rok-cesnovar:
> it feels like the answer is a strong "yes we need to make a copy"

I am afraid I am +1 on that. I do think we need a copy..

SteveBronder:
Cool cool I was pretty sure as well

test/unit/math/mix/fun/cholesky_decompose_test.cpp
```diff
@@ -8,59 +8,156 @@ namespace cholesky_decompose_test {
...
+
 void expect_cholesky(const Eigen::MatrixXd& Sigma) {
   Eigen::VectorXd yy = stan::math::cov_matrix_free(Sigma);
-  stan::test::expect_ad(f(10), yy);
+  // lazy, solving for x in x = (N * (N + 1)) / 2
+  int dof = .5 * (std::sqrt(8 * yy.size() + 1) - 1);
+  stan::test::expect_ad(f(dof), yy);
+}
+template <typename F, typename T>
+void expect_cholesky_var(stan::test::ad_tolerances& tols, const F& f,
```
SteveBronder, line 33: @bbbales2 do you know anyway we can speed up this test? Right now I'm just checking the lower triangular of the output and it's still taking 5 ever

bbbales2: How big are the matrices here?

SteveBronder: 96 x 96 and 160 x 160 lol

SteveBronder: The issue is that it requires `25600` calls to grad and cholesky decompose for the full matrix. Or am I reading this wrong?

bbbales2: Woof, yeah for reverse mode. Lol I didn't think about this but this is gonna be a bit of a trick.

Let me write up some code in R to do this real quick.

bbbales2: Alright what I think we're going to have to do is instead of computing the Jacobian of this (which will be (NxN)x(NxN)), we gotta do the trick where we element-wise multiply the output of the function with some random thing and take the sum to reduce it to a scalar and look at the Jacobian there (which is just NxN).

It's not checking the full Jacobian but with the randomness we're still doing a good check (it'll be really hard to be accidentally right, at least):

So in pseudo-code the calculation is:
```
L = chol(A)
out = sum(L * matrix(runif(N * N), nrow = N))
out.grad()
// Check adjoints of A
```

There's a short analytic form for the adjoints we need in eq. 6 and 9 here: https://homepages.inf.ed.ac.uk/imurray2/pub/16choldiff/choldiff.pdf

Here is R code for that:
```
lchol = function(A) {
  t(chol(A))
}

Sigma = function(A) {
  out = matrix(0, nrow = nrow(A), ncol = ncol(A))
  for(i in 1:nrow(A)) {
    for(j in 1:i) {
      if(j == i)
        out[i, j] = A[i, j] / 2.0
      else
        out[i, j] = A[i, j]
    }
  }
  return(out)
}

N = 3
A = matrix(runif(N * N), nrow = N)
A = A + t(A)
A = A + 2.1 * N * diag(N) # 2 * N should be safe, 2.1 * N to be super safe

L = lchol(A)

# Use sigma here just to make lower triangular random matrix
adjL = Sigma(matrix(runif(N * N), nrow = N))

SigmaLinv = t(solve(t(L), t(Sigma(t(L) %*% adjL))))
S = solve(t(L), SigmaLinv)

# I did this a little differently so our adjoints would have an upper triangle of zero which
# I think matches current Stan behavior. Check this.
adjA = Sigma(S + t(S))
```

And here is finite difference code you can use to convince yourself that this R code is doing the right thing:

```
#### Compute a reference using finite difference (not needed for C++)
fd_adjA = matrix(0, nrow = N, ncol = N)
for(i in 1:N) {
  for(j in 1:i) {
    dx = 1e-3
    Ap = A
    Am = A
    Ap[i, j] = Ap[i, j] + dx
    Am[i, j] = Am[i, j] - dx
    Ap[j, i] = Ap[i, j]
    Am[j, i] = Am[i, j]
    
    adj = (sum(adjL * (lchol(Ap))) - sum(adjL * lchol(Am))) / (2 * dx)
    
    fd_adjA[i, j] = adj
  }
}

adjA
fd_adjA
```

bbbales2: Oh. Whoops. That thing I'm calling a Sigma is a Phi in the paper lol. My bad lol.

rok-cesnovar:
@SteveBronder this job gets stuck in the MathMixMatFun.choleskyDecomposeGeneralBig test and does not stop. 

Its not a Jenkins one-off error: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2105/16/pipeline/122
https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2105/15/pipeline/122
and there are a few more. All these jobs ran for 20 hours+

@serban-nicusor-toptal I keep replaying this job so it unstucks the Jenkins pipeline. You probably have the permission to just stop this job so that it doesnt even restart?

SteveBronder:
Yeah sorry I'm afk atm can you just kill it. We need to sort out a way to test medium sized cholesky

rok-cesnovar:
No problem, just wrote so you know why I keep restarting this :) I am sure Nic will kill it once he is up.

test/unit/math/mix/fun/cholesky_decompose_test.cpp
SteveBronder: Actually Ben could we just bring back a bunch of the old rev tests? It looks like they are doing this?

https://github.com/stan-dev/math/blob/feature/flto/test/unit/math/rev/mat/fun/cholesky_decompose_test.cpp

SteveBronder: (That's just a random branch that has the old tests in it)

bbbales2: Yeah I'm fine with whatever gets it tested. Just as long as we're vaguely testing gradients of cholesky for matrices larger than 35x35 or whatever the cutoff is.

SteveBronder: Woof @bbbales2 excellent catch on this, was very wrong! Now correct!

bbbales2:
@SteveBronder looks like this needs bumped so the tests restart.

rok-cesnovar:
This expression test failures here are not false positives it seems.

SteveBronder:
Whups! Yes Rok good catch thanks!

stan/math/rev/fun/cholesky_decompose.hpp
```diff
@@ -358,76 +183,100 @@ class cholesky_opencl : public vari {
...
+ * Internally calls Eigen::LLT rather than using
+ * stan::math::cholesky_decompose in order to use an inplace decomposition.
+ *
+ * Note chainable stack varis are created below in Matrix<var, -1, -1>
+ * @tparam T A `var_value` holding an inner eigen type.
+ * @param A A square positive definite matrix with no nan values.
+ * @return L Cholesky factor of A
+ */
+template <typename T, require_var_matrix_t<T>* = nullptr>
+inline auto cholesky_decompose(const T& A) {
```
bbbales2, line 272: There should probably be an issue for `var<mat>` cholesky decompose not supporting OpenCL.

stan/math/rev/fun/cholesky_decompose.hpp
```diff
@@ -1,330 +1,151 @@
...
+ *
+ * Note algorithm  as laid out in Giles is row-major, so Eigen::Matrices
+ * are explicitly storage order RowMajor, whereas Eigen defaults to
+ * ColumnMajor. Also note algorithm starts by calculating the adjoint for
+ * A(M_ - 1, M_ - 1), hence pos on line 94 is decremented to start at pos
+ * = M_ * (M_ + 1) / 2.
+ */
+template <typename T1, typename T2, typename T3>
+inline auto unblocked_cholesky_lambda(T1& L_A, T2& L, T3& A) {
+  return [=]() mutable {
```
bbbales2, line 54: Name the captures

stan/math/rev/fun/cholesky_decompose.hpp
```diff
@@ -1,330 +1,151 @@
...
-    using Eigen::Block;
+/**
+ * Reverse mode differentiation algorithm reference:
+ *
+ * Iain Murray: Differentiation of the Cholesky decomposition, 2016.
+ *
+ */
+template <typename T1, typename T2, typename T3>
+inline auto cholesky_lambda(T1& L_A, T2& L, T3& A) {
+  return [=]() mutable {
```
bbbales2, line 85: Name captures

stan/math/rev/fun/cholesky_decompose.hpp
```diff
@@ -1,330 +1,151 @@
...
+/**
+ * Reverse mode differentiation for Cholesky using OpenCL
+ * Reverse mode differentiation algorithm reference:
+ *
+ * Iain Murray: Differentiation of the Cholesky decomposition, 2016.
+ *
+ */
+template <typename AMat, typename LVari>
+inline auto opencl_cholesky_lambda(AMat& arena_A, LVari& vari_L) {
+  return [=]() mutable {
```
bbbales2, line 136: Name captures

stan/math/rev/fun/cholesky_decompose.hpp
rok-cesnovar: It does exist: https://github.com/stan-dev/math/pull/2117#issue-comment-box

OpenCL version will no longer be ifdefed and will be completely separate in /opencl.

The ifdefed versions are going away once we support this in stanc3.

bbbales2 (changes requested):
Pedantic change, but capture things by name on the lambdas. I guess it's unlikely that we capture some sort of global, and presumably the memory checks in test will tell us if it matters, but if the capture is manual then it's easy to know what's going on.

Since this doesn't have the `var<mat>` OpenCL implementation, make an issue for that. Otherwise it all looks good to me.

stan/math/rev/fun/cholesky_decompose.hpp
bbbales2: Oh okay cool

SteveBronder:
Word thanks for fixing that!

bbbales2 (approved):

Issue #2083, opened 2020-09-11
Building Google benchmark library in math


bbbales2:
## Description

The benchmark library has to be built before it can be used.

Assuming BENCHMARK points to the benchmark folder (lib/benchmarks_1.5.1 as of writing), the simplest build directions are:

```
cd $(BENCHMARK)
mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=RELEASE
make
```

I'm not sure if the separate build directory is required or not -- just stealing these instructions from the perf-math readme.

With that build the required flags are:
```
includes: -I $(BENCHMARK)/include
library paths: -L $(BENCHMARK)/build/src
library flag: -lbenchmark
```

I don't think this needs built and included by default.

I assume this should be built as part of `make/math-standalone` though.

#### Current Version:
v3.3.0

PR #2106, opened 2020-09-25
Added a few varmat functions (Issue #2101)

bbbales2:
## Summary

This adds `var_value<Eigen::Matrix<T, R, C>>` implementations for columns_dot_self, determinant, dot_self, inverse, log_determinant, matrix_power, multiply_lower_tri_self_transpose, and tcrossprod. (part of #2101)

## Side Effects

Should be none

## Release notes

Added `var_value<Eigen::Matrix<T, R, C>>` specializations for columns_dot_self, determinant, dot_self, inverse, log_determinant, matrix_power, multiply_lower_tri_self_transpose, and tcrossprod.

## Checklist

- [x] Math issue #2101


stan/math/rev/fun/tcrossprod.hpp
```diff
@@ -21,41 +21,26 @@ namespace math {
...
+
+  if (M.size() == 0) {
+    return M;
   }
-  return MMt;
+
+  arena_t<plain_type_t<T>> arena_M = M;
+  arena_t<Eigen::MatrixXd> arena_M_val = value_of(arena_M);
+
+  arena_t<T_var> res = arena_M_val * arena_M_val.transpose();
```
bbbales2, line 36: This is a better way to handle vars for the matvar case: https://github.com/stan-dev/math/blob/0aaea95d30d6701a2ac70ca9a7aed1af01383aa4/stan/math/rev/fun/tcrossprod.hpp#L39

Should we just be in the habit of doing both implementations? The reverse pass callback code is different. They're really kinda different things.

bbbales2 (commented):
Questions @SteveBronder

stan/math/rev/fun/tcrossprod.hpp
t4c1: I think we can handle both matvar and varmat in the same overload if we are smart about it.

stan/math/rev/fun/tcrossprod.hpp
```diff
@@ -21,41 +21,26 @@ namespace math {
...
+  using T_double = Eigen::MatrixXd;
+  using T_var = plain_type_t<T>;
+
+  if (M.size() == 0) {
+    return M;
   }
-  return MMt;
+
+  arena_t<plain_type_t<T>> arena_M = M;
+  arena_t<Eigen::MatrixXd> arena_M_val = value_of(arena_M);
```
t4c1, line 34: Hmm, I think we need to do something here to avoid the copy in the varmat case. The easiest option would be to change the var<matrix> to use the arena matrix internally instead of maps, so the arena_matrix constructor here will be able to do a shallow copy.

stan/math/rev/fun/tcrossprod.hpp
```diff
@@ -21,41 +21,26 @@ namespace math {
...
+
+  arena_t<T_var> res = arena_M_val * arena_M_val.transpose();
+
+  reverse_pass_callback([res, arena_M, arena_M_val]() mutable {
+    Eigen::MatrixXd adj = res.adj();
```
t4c1, line 39:
```suggestion
    const auto& adj = to_ref(res.adj());
```
This way we can avoid the copy in the varmat case.

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,33 @@
...
+  arena_t<T_var> arena_L = L;
+  arena_t<T_double> arena_L_val
+      = value_of(arena_L).template triangularView<Eigen::Lower>();
+
+  arena_t<T_var> res = arena_L_val.template triangularView<Eigen::Lower>()
+                       * arena_L_val.transpose();
+
+  reverse_pass_callback([res, arena_L, arena_L_val]() mutable {
+    const auto& adj = to_ref(res.adj());
+    Eigen::MatrixXd adjL = (adj.transpose() + adj) * arena_L_val;
```
t4c1, line 31:
```suggestion
    Eigen::MatrixXd adjL = (adj.transpose() + adj) * arena_L_val.template triangularView<Eigen::Lower>();
```
This way you don't need the view when copying on the arena.

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,33 @@
...
+
+    for (size_t j = 1; j < adjL.cols(); ++j)
+      for (size_t i = 0; i < std::min(static_cast<size_t>(adjL.rows()), j); ++i)
+        adjL(i, j) = 0.0;
+
+    arena_L.adj() += adjL;
```
t4c1, line 37:
```suggestion
    arena_L.adj() += adjL.template triangularView<Eigen::Lower>();
```

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,50 @@ class matrix_product_vari : public vari {
...
+  arena_t<plain_type_t<T>> res = powers[powers.size() - 1];
+
+  if(M.size() > 0) {
+  reverse_pass_callback([arena_M, n, N, res, powers]() mutable {
+    const auto& M_val = powers[1];
+    Eigen::MatrixXd adj_C = res.adj();
+    Eigen::MatrixXd adj_M = Eigen::MatrixXd::Zero(N, N);
+    for (size_t i = n; i > 1; --i) {
+      adj_M += adj_C * powers[i - 1].transpose();
+      adj_C = (M_val.transpose() * adj_C).eval();
```
t4c1:
```suggestion
      adj_C = M_val.transpose() * adj_C;
```
Eval is unnecessary.

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,50 @@ class matrix_product_vari : public vari {
...
+    }
+    adj_M += adj_C;
+    arena_M.adj() += adj_M;
```
t4c1:
```suggestion
    arena_M.adj() += adj_M + adj_C;
```

stan/math/rev/fun/log_determinant.hpp
```diff
@@ -4,33 +4,39 @@
...
+  check_square("determinant", "m", m);
 
-  math::check_square("log_determinant", "m", m);
+  if (m.size() == 0) {
+    return 0.0;
+  }
 
-  Eigen::FullPivHouseholderQR<promote_scalar_t<double, EigMat>> hh
-      = m.val().fullPivHouseholderQr();
+  const auto& m_ref = to_ref(m);
```
t4c1, line 24: This variable is unused.

stan/math/rev/fun/inverse.hpp
```diff
@@ -75,16 +22,22 @@ class inverse_vari : public vari {
...
-  internal::inverse_vari *baseVari = new internal::inverse_vari(m);
-  res.vi() = Eigen::Map<matrix_vi>(baseVari->vari_ref_A_inv_, res.rows(),
-                                   res.cols());
+  arena_t<plain_type_t<T>> arena_m = m;
+  arena_t<Eigen::MatrixXd> res_val = value_of(arena_m).inverse();
+  arena_t<plain_type_t<T>> res = res_val;
+
+  reverse_pass_callback([res, res_val, arena_m]() mutable {
+    Eigen::MatrixXd res_adj = res.adj();
+    arena_m.adj() -= res_val.transpose() * res_adj * res_val.transpose();
```
t4c1, line 38:
```suggestion
    arena_m.adj() -= res_val.transpose() * res.adj() * res_val.transpose();
```

stan/math/rev/fun/dot_self.hpp
```diff
@@ -45,10 +20,18 @@ class dot_self_vari : public vari {
  * @param[in] v Vector.
  * @return Dot product of the vector with itself.
  */
-template <typename T, require_eigen_vector_vt<is_var, T>* = nullptr>
+template <typename T, require_rev_matrix_t<T>* = nullptr>
 inline var dot_self(const T& v) {
-  const Eigen::Ref<const plain_type_t<T>>& v_ref = v;
-  return {new internal::dot_self_vari(v_ref)};
+  arena_t<plain_type_t<decltype(value_of(v))>> v_val = value_of(v);
```
t4c1, line 25:
```suggestion
  arena_t<decltype(value_of(v))> v_val = value_of(v);
```
No need to use `plain_type_t` in `arena_t`. `arena_t<plain_type_t<T>> = arena_t<T>`

stan/math/rev/fun/determinant.hpp
```diff
@@ -3,56 +3,37 @@
...
     return 1;
   }
 
-  return {new internal::determinant_vari<T::RowsAtCompileTime,
-                                         T::ColsAtCompileTime>(m)};
+  const auto& m_ref = to_ref(m);
+  const auto& m_val = value_of(m_ref);
+
+  double det_val = m_val.determinant();
+  arena_t<plain_type_t<T>> arena_m = m_ref;
```
t4c1, line 27:
```suggestion
  arena_t<plain_type_t<T>> arena_m = m;
```
No need for a separate ref. Just use `arena_m` instead.

t4c1 (changes requested):

stan/math/rev/fun/inverse.hpp
bbbales2: Whenever I try to do this I get errors in the matrix multiplication that come back to:
```
./stan/math/rev/fun/inverse.hpp:38:21: note: in instantiation of function template specialization
      'Eigen::MatrixBase<Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1>, 8, Eigen::Stride<0, 0> >
      >::adj_Op, Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1>, 8, Eigen::Stride<0, 0> > >
      >::operator-=<Eigen::Product<Eigen::Product<Eigen::Transpose<Eigen::Map<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 8, Eigen::Stride<0, 0> > >,
      Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1>, 8, Eigen::Stride<0, 0> > >::adj_Op,
      Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1>, 8, Eigen::Stride<0, 0> > >, 0>, Eigen::Transpose<Eigen::Map<Eigen::Matrix<double, -1, -1,
      0, -1, -1>, 8, Eigen::Stride<0, 0> > >, 0> >' requested here
      arena_m.adj() -= res_val.transpose() * res.adj() * res_val.transpose();
```

The error itself is:
```
lib/eigen_3.3.7/Eigen/src/Core/products/GeneralMatrixMatrix.h:228:22: error: no matching member function for call to 'coeffRef'
              &m_rhs.coeffRef(0,col), m_rhs.outerStride()
```

Any ideas?

stan/math/rev/fun/tcrossprod.hpp
bbbales2: Well the big difference is that if we know the output is symmetric, then we only need to build new varis on the diagonal and one half of the other side of the matrix.

The issue then is that the reverse mode code is different from how we'd handle the varmats. The off diagonal adjoints of the matvar have been incremented twice, whereas the off diagonal adjoints of the mat var (of which there are two) have each been incremented once.

stan/math/rev/fun/inverse.hpp
andrjohns: It's the bug where multiplying two ```CWiseUnaryViews``` fails. This was why Steve was introducing the ```adj_op()``` member to always return a ```CWiseUnaryOp``` regardless of const-ness.

In this case you'll either have to eval the ```res.adj()``` or add const-ness to ```res``` before calling the adj member. Maybe something like:
```
const_cast<const decltype(res)>(res).adj()
```
?

bbbales2: @andrjohns thanks. I just switched to:

```
arena_m.adj() -= res_val.transpose() * res.adj_op() * res_val.transpose();
```

and it worked. Is `adj_op` supposed to be the same as `adj`?

stan/math/rev/fun/dot_self.hpp
```diff
@@ -45,10 +20,19 @@ class dot_self_vari : public vari {
...
  * @return Dot product of the vector with itself.
  */
-template <typename T, require_eigen_vector_vt<is_var, T>* = nullptr>
+template <typename T, require_rev_matrix_t<T>* = nullptr>
 inline var dot_self(const T& v) {
-  const Eigen::Ref<const plain_type_t<T>>& v_ref = v;
-  return {new internal::dot_self_vari(v_ref)};
+  arena_t<plain_type_t<T>> arena_v = v;
+
+  arena_t<decltype(value_of(arena_v))> v_val = value_of(arena_v);
```
SteveBronder, line 27:
```suggestion
  arena_t<decltype(arena_v.val())> v_val = arena_v.val();
```

stan/math/rev/fun/inverse.hpp
```diff
@@ -75,16 +22,21 @@ class inverse_vari : public vari {
...
-  internal::inverse_vari *baseVari = new internal::inverse_vari(m);
-  res.vi() = Eigen::Map<matrix_vi>(baseVari->vari_ref_A_inv_, res.rows(),
-                                   res.cols());
+  arena_t<plain_type_t<T>> arena_m = m;
+  arena_t<Eigen::MatrixXd> res_val = value_of(arena_m).inverse();
+  arena_t<plain_type_t<T>> res = res_val;
+
+  reverse_pass_callback([res, res_val, arena_m]() mutable {
+    arena_m.adj() -= res_val.transpose() * res.adj_op() * res_val.transpose();
+  });
```
SteveBronder, line 38: [optional]

I wonder if this is one of the places where casting `m` to a `var<matrix>` is always better. We pull out the values and read through them multiple times, then also have to read through the adjoints of `arena_m`. If we made this always use `var<matrix>` for arena_m we'd have to pay the cost of writing the adjoints from the `var<matrix>` back to the `matrix<var>`, but we would get nice access in the `reverse_pass_callback()` here

stan/math/rev/fun/log_determinant.hpp
```diff
@@ -4,33 +4,37 @@
...
-      = ChainableStack::instance_->memalloc_.alloc_array<vari*>(m.size());
-  Eigen::Map<matrix_vi>(varis, m.rows(), m.cols()) = m.vi();
+  Eigen::FullPivHouseholderQR<Eigen::MatrixXd> hh
+      = arena_m.val().fullPivHouseholderQr();
 
-  double* gradients
-      = ChainableStack::instance_->memalloc_.alloc_array<double>(m.size());
-  Eigen::Map<matrix_d>(gradients, m.rows(), m.cols())
-      = hh.inverse().transpose();
+  arena_t<Eigen::MatrixXd> arena_hh_inv_t = hh.inverse().transpose();
```
SteveBronder, line 29: Get rid of `_t`

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,49 @@ class matrix_product_vari : public vari {
...
-  check_square("matrix_power", "M", M);
-  if (n < 0)
-    invalid_argument("matrix_power", "n", n, "is ", ", but must be >= 0!");
-  if (M.rows() == 0)
-    invalid_argument("matrix_power", "M.rows()", M.rows(), "is ",
-                     ", but must be > 0!");
+template <typename T, require_rev_matrix_t<T>* = nullptr>
+inline plain_type_t<T> matrix_power(const T& M, const int n) {
   const auto& M_ref = to_ref(M);
+  check_square("matrix_power", "M", M);
```
SteveBronder, line 31:
```suggestion
  check_square("matrix_power", "M", M_ref);
```

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,50 @@ class matrix_product_vari : public vari {
...
+    return Eigen::MatrixXd::Identity(N, N);
+
+  if (n == 1)
+    return M_ref;
+
+  arena_t<std::vector<Eigen::MatrixXd>> powers(n + 1);
+  arena_t<plain_type_t<T>> arena_M = M_ref;
+
+  powers[0] = Eigen::MatrixXd::Identity(N, N);
+  powers[1] = value_of(M_ref).eval();
```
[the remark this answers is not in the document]
SteveBronder: I think you need to give them a default size when you make the vector like 

```
arena_t<std::vector<Eigen::MatrixXd>> powers(n + 1, Eigen::MatrixXd(N, N));
```

Otherwise the vector holds a bunch of default initialized matrices

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,49 @@ class matrix_product_vari : public vari {
...
+  arena_t<plain_type_t<T>> res = powers[powers.size() - 1];
+
+  if (M.size() > 0) {
+    reverse_pass_callback([arena_M, n, N, res, powers]() mutable {
+      const auto& M_val = powers[1];
+      Eigen::MatrixXd adj_C = res.adj();
+      Eigen::MatrixXd adj_M = Eigen::MatrixXd::Zero(N, N);
+      for (size_t i = n; i > 1; --i) {
+        adj_M += adj_C * powers[i - 1].transpose();
+        adj_C = M_val.transpose() * adj_C;
```
SteveBronder, line 64: Isn't there an aliasing issue with using the `adj_C` on both sides?

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,31 @@
...
+
+  arena_t<T_var> arena_L = L;
+  arena_t<T_double> arena_L_val
+      = value_of(arena_L).template triangularView<Eigen::Lower>();
+
+  arena_t<T_var> res = arena_L_val.template triangularView<Eigen::Lower>()
+                       * arena_L_val.transpose();
+
+  reverse_pass_callback([res, arena_L, arena_L_val]() mutable {
+    const auto& adj = to_ref(res.adj());
```
SteveBronder, line 30: This returns an expression type for `matrix<var>`. I think we need that `eval()` function

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,31 @@
...
+    Eigen::MatrixXd adjL
+        = (adj.transpose() + adj)
+          * arena_L_val.template triangularView<Eigen::Lower>();
+
+    arena_L.adj() += adjL.template triangularView<Eigen::Lower>();
```
SteveBronder, line 35: I think you can do this

```suggestion
    auto adjL
        = (adj.transpose() + adj)
          * arena_L_val.template triangularView<Eigen::Lower>();

    arena_L.adj() += adjL.template triangularView<Eigen::Lower>();
```

stan/math/rev/fun/tcrossprod.hpp
SteveBronder: ^ I was thinking something similar yesterday. idt it would add any unreasonable bulk and would make stuff flow a lot more naturally. I want to get in all the big changes from #2064 then we can do this in a separate PR

stan/math/rev/fun/tcrossprod.hpp
SteveBronder: I think it's totally fine to have separate functions when stuff is different

test/unit/math/mix/fun/inverse_test.cpp
```diff
@@ -5,17 +5,21 @@ TEST(mathMixMatFun, inverse) {
   auto f = [](const auto& x) { return stan::math::inverse(x); };
 
   Eigen::MatrixXd t(0, 0);
-  stan::test::expect_ad(f, t);
+  stan::test::expect_ad_matvar(f, t);
 
+  stan::test::expect_ad(f, t);
```
SteveBronder, line 10: I'd just put these back in place

SteveBronder (changes requested):
Bunch o' comments!

stan/math/rev/fun/tcrossprod.hpp
SteveBronder: Actually this was pretty easy-peasy!

https://github.com/stan-dev/math/compare/feature/multi-expression-vari-views...feature/vari_value-arena-matrix

stan/math/rev/fun/inverse.hpp
andrjohns: Didn't realise that was in yet, neat! Yeah, so the issue with ```adj``` is that will return a different Eigen expression depending on whether the input matrix is const or not:
```
Eigen::Matrix<var, -1, -1> non_const(R, C);
non_const.adj()  // Returns a CWiseUnaryView, which can be used as either an l-value or an r-value

const Eigen::Matrix<var, -1, -1> const(R, C);
const.adj()  // Returns a CWiseUnaryOp, which can only be used as an r-value
```

But Eigen has a bug with multiplying ```CWiseUnaryViews``` together, so Steve introduced the ```adj_op``` member which will always return a CWiseUnaryOp:
```
Eigen::Matrix<var, -1, -1> non_const(R, C);
non_const.adj_op()  // Returns a CWiseUnaryOp

const Eigen::Matrix<var, -1, -1> const(R, C);
const.adj_op()  // Returns a CWiseUnaryOp
```

stan/math/rev/fun/tcrossprod.hpp
bbbales2: Nice, can you open a separate pull request with this? I don't wanna slow this up going through here.

Also the code here: https://github.com/stan-dev/math/pull/2106#discussion_r497788511 reminds me that maybe this would work:
```
const auto& arena_M_val = to_arena(value_of(arena_M));
```

If it see's an arena variable it just forwards it along I think.

stan/math/rev/fun/columns_dot_self.hpp
```diff
@@ -16,14 +16,23 @@ namespace math {
...
-    const Mat& x) {
-  Eigen::Matrix<var, 1, Mat::ColsAtCompileTime> ret(1, x.cols());
-  for (size_type i = 0; i < x.cols(); i++) {
-    ret(i) = var(new internal::dot_self_vari(x.col(i)));
+template <typename Mat, require_rev_matrix_t<Mat>* = nullptr>
+inline auto columns_dot_self(const Mat& x) {
+  using T_return = promote_var_matrix_t<Eigen::RowVectorXd, Mat>;
+
+  arena_t<plain_type_t<Mat>> arena_x = x;
+  arena_t<Eigen::MatrixXd> arena_x_val = value_of(arena_x);
```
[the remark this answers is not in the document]
t4c1: `value_of(arena_x)` is (an expressions that evaluates to) `MatrixXd`, so `arena_t<decltype(value_of(arena_x))>` equals `arena_t<Eigen::MatrixXd>`

stan/math/rev/fun/columns_dot_self.hpp
```diff
@@ -16,14 +16,23 @@ namespace math {
...
+
+  T_return res = (arena_x_val.array() * arena_x_val.array()).colwise().sum();
+
+  if (x.size() > 0) {
+    reverse_pass_callback([res, arena_x, arena_x_val]() mutable {
+      Eigen::RowVectorXd rv = res.adj();
+      arena_x.adj() += 2 * arena_x_val * rv.asDiagonal();
```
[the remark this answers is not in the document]
t4c1: > @t4c1 does calling .eval() on a map make a copy?

 It does.


> If so then maybe we should have an eval() method that for an expression does a copy but for an Eigen map does not

We have one. It is `to_ref()`. However we do not need it here - we can just as well make this `const auto&`.

stan/math/rev/fun/determinant.hpp
```diff
@@ -3,56 +3,36 @@
...
-                 .inverse()
-                 .transpose();
-  }
-};
-}  // namespace internal
-
-template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+template <typename T, require_rev_matrix_t<T>* = nullptr>
 inline var determinant(const T& m) {
   check_square("determinant", "m", m);
```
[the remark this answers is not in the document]
t4c1: I have been calling functions that just use dimmensions on expressions all over the place. I would disagree on using `to_ref` before that a good practice, as it can be slower in the case where otherwise `to_ref` is neet needed.

stan/math/rev/fun/determinant.hpp
```diff
@@ -3,56 +3,36 @@
...
+
   if (m.size() == 0) {
     return 1;
   }
 
-  return {new internal::determinant_vari<T::RowsAtCompileTime,
-                                         T::ColsAtCompileTime>(m)};
+  arena_t<plain_type_t<T>> arena_m = m;
+  const auto& m_val = to_ref(value_of(arena_m));
```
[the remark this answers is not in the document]
t4c1: Nope. If `arena_m` is a var_value holding a view this would make an unnecessary copy.

stan/math/rev/fun/inverse.hpp
t4c1: These operations are memory bound. So it only matters how much data we need to transfer between memory and cpu. Assuming the matrix is large enough that it does not fit into cache:

arena_m has value accessed once and adjoint once. Each of those need to load pointers, values and adjoints in cache. Incrementing the adjoints also need to store values and adjoints back into ram. This in total transfers 8*64bits per element over the bus. 

Transforming to varmat (`to_var_value()`) copies pointers to arena(2), copies values(loading pointers, values and adjoints and storing values, in total 4) and in reverse pass increments adjoint (oading pointers, values and adjoints and storing adjoints and values, in total 5). Using this varmat still means loading values (1) and incrementing adjoints(2). In total that would mean 2+4+5+1+2=14*64bits per element of transfers over the bus, making this slower.

stan/math/rev/fun/matrix_power.hpp
t4c1: When multiplication is involved, Eigen assumes aliasing and evaluates multiplication into a temporary and than copies it back.

If you know there is no aliasing you can tell it to Eigen by doing `a.noalias() = b*c;` (not relevant here).

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
t4c1: `to_ref` evals nontrivial expressions (which this is).

stan/math/rev/fun/columns_dot_self.hpp
```diff
@@ -16,14 +16,23 @@ namespace math {
...
+template <typename Mat, require_rev_matrix_t<Mat>* = nullptr>
+inline auto columns_dot_self(const Mat& x) {
+  using T_return = promote_var_matrix_t<Eigen::RowVectorXd, Mat>;
+
+  arena_t<plain_type_t<Mat>> arena_x = x;
+  arena_t<Eigen::MatrixXd> arena_x_val = value_of(arena_x);
+
+  T_return res = (arena_x_val.array() * arena_x_val.array()).colwise().sum();
+
+  if (x.size() > 0) {
```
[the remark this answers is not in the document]
bbbales2: The return type is different than the argument type in this case.

stan/math/rev/fun/inverse.hpp
bbbales2: Righto, from now on if `adj()` gives me compiler errors I'm just gonna switch to `adj_op()`. Thanks!

stan/math/rev/fun/columns_dot_self.hpp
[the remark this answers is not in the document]
bbbales2: I just removed the temporary.

stan/math/rev/fun/determinant.hpp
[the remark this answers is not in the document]
bbbales2: The expressions catch any double evaluations! They're great!

stan/math/rev/fun/determinant.hpp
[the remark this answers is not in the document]
bbbales2: I'm gonna leave this as is re: @t4c1

stan/math/rev/fun/dot_self.hpp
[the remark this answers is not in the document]
bbbales2: I like `value_of`!

stan/math/rev/fun/matrix_power.hpp
bbbales2: I changed this around in a different way to that we check finite last (and only do a `to_ref` once we know we have a square, non-zero sized matrix).

bbbales2:
I responded to the comments and went through and tried to make these as fast as possible for `mat<var>` types (so as to not create any slowdown since the last release).

I'm re-running the benchmarks now but not great news.

`tcrossprod` and `multiply_lower_self_tri_transpose` are very slow for small values (like half the speed for 2x2, 4x4), but get fast pretty quick.

`determinant`, `log_determinant`, and `inverse` are all consistently slightly slower.

`columns_dot_self` is the one I need some help on depending on how the benchmarks work out. It is slower for large inputs on the new branch (small values it does okay). I'll check the benchmarks tomorrow when they finish and update this pull.

I didn't check `matrix_power` cause the previous benchmarks showed it was already faster everywhere.

Anyway if anyone wants to mess with these to make them faster, have at it, I'm okay with where this is at so I'll just be trying to get a green light.

stan/math/rev/fun/tcrossprod.hpp
```diff
@@ -17,45 +17,27 @@ namespace math {
...
  * @param M Matrix to multiply.
  * @return M times its transpose.
  */
-template <typename T, require_eigen_vt<is_var, T>* = nullptr>
-inline Eigen::Matrix<var, T::RowsAtCompileTime, T::RowsAtCompileTime>
-tcrossprod(const T& M) {
-  if (M.rows() == 0) {
-    return {};
+template <typename T, require_rev_matrix_t<T>* = nullptr>
+inline promote_var_matrix_t<Eigen::MatrixXd, T> tcrossprod(const T& M) {
```
bbbales2, line 25: @t4c1, @SteveBronder I think there's a problem with either `plain_type_t<T>` or `promote_var_matrix_t` or my use thereof.

If the return type here is `plain_type_t<T>`, you'll get errors compiling `crossprod` ([here](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2106/11/pipeline/)). `crossprod(x)` is implemented as `tcrossprod(x.transpose())`.

When I changed the return type to `promote_var_matrix_t<T>`, things worked.

I added pretty_prints for the types:

```
T:  stan::math::test::type_name() [Arg = Eigen::Transpose<const Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1> >
plain_type_t :  stan::math::test::type_name() [Arg = Eigen::Matrix<stan::math::var_value<double>, -1, -1, 1, -1, -1>
promote_var_matrix_t :  stan::math::test::type_name() [Arg = Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1>
```

Is it me or it is a template?

bbbales2 (commented):
Question!

stan/math/rev/fun/tcrossprod.hpp
SteveBronder: Sorry I'll look into this tmrw morning. The jenkins link sends me to

```
./test/unit/math/serializer.hpp:199:11: error: no matching member function for call to 'push_back'

    vals_.push_back(x);

    ~~~~~~^~~~~~~~~

./test/unit/math/serializer.hpp:291:5: note: in instantiation of function template specialization 'stan::test::serializer<stan::math::var_value<double> >::write<Eigen::Matrix<stan::math::var_value<double>, -1, -1, 1, -1, -1> >' requested here

  s.write(x);
```

I'm not seeing how it's related to the return type. Is there a reason you can't use `auto` as the return type?

bbbales2: > Is there a reason you can't use auto as the return type

We need explicit return types here because the return type gets build internally as an `arena_t<T_Return>` and you gotta set the return type to `T_Return` to get it to cast to a non-arena thing.

andrjohns: Looking at the different template arguments between the types:
```
Eigen::Matrix<stan::math::var_value<double>, -1, -1, 1, -1, -1>
Eigen::Matrix<stan::math::var_value<double>, -1, -1, 0, -1, -1>
```

That third integer is the 'Options' flag, which controls storage order. I would guess (emphasis on the guess) that the ```.transpose()``` operator maps the matrix to row-major storage, so the plain type of the input (and the eventual return type) is a row-major matrix and then compilation fails when you try to return a column-major matrix instead. No idea why ```plain_type_t``` and ```promote_var_matrix_t``` would handle that differently though

t4c1: From the type printouts it seems plain_type_t is returning a row major matrix (yes, it can do that in some cases!). Any function not generalized for expresions is still expecting its inputs to be col major, so I guess that causes compilation failures. 

This means that we shouldn't use plain_type_t (or .eval() and auto) for return values in the few functions that eval expresions, for which Eigen prefer row major types.

t4c1: > No idea why plain_type_t and promote_var_matrix_t would handle that differently though

`plain_type_t` actually gets the plain type for the expression. `promote_var_matrix_t` (and `promote_scalar_t`) just determine a (col_major) matrix with appropriate scalar type.

bbbales2: I'm going to take away that we should use `promote_var_matrix_t` for the return type and avoid `plain_type_t` to make sure we aren't getting matrices in unexpected storage orders.

Considering this is a bug-like thing that may have shown up in the wild somewhere, is this something we should test for somehow?

Or is this failure to compile actually just a problem with the test framework that should be fixed?

t4c1: I would consider it a bug, but testing for it is not high priority, as it will cease to be a bug as soon as all functions accept general expressions.

andrjohns: Won't it be more of an issue with expressions since they'll propagate the row-major matrix which could lead to unexpected behaviour further down the chain?

t4c1: Functions handling expressions also handle row major matrices, so it won't matter.

andrjohns: I was mostly thinking that any looping is assuming column-major ordering (i.e., loops over column first), so some programs could end up slower if a function returned a row-major matrix at some point. But I don't know how much looping we still have at the moment

SteveBronder: > We need explicit return types here because the return type gets build internally as an arena_t<T_Return> and you gotta set the return type to T_Return to get it to cast to a non-arena thing.

`promote_var_matrix_t` returns a `var<matrix>` or `matrix<var>` so I usually just wrap the return like `return ret_type(ret)`

stan/math/rev/fun/columns_dot_self.hpp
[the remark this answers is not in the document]
SteveBronder: Ah good call.

Another note though, does this make a copy for `var<matrix>` rn? I think after #2064 this will be better since then we use `arena_matrix` on the backend of `var_value<matrix>`

SteveBronder:
I'm reading the docs right now for eigen products and should we be using `noalias()` more often with these? In the docs they specifically call out `+=` (which we do all the time for the reverse pass)

https://eigen.tuxfamily.org/dox/TopicWritingEfficientProductExpression.html

So like this

```cpp
    const auto& adjL = (adj.transpose() + adj)
                       * arena_L_val.template triangularView<Eigen::Lower>();

    arena_L.adj() += adjL.template triangularView<Eigen::Lower>();
```

should be 

```cpp
    const auto& adjL = (adj.transpose() + adj)
                       * arena_L_val.template triangularView<Eigen::Lower>();

    arena_L.adj().noalias() += adjL.template triangularView<Eigen::Lower>();
```

bbbales2:
Oh interesting. I'll try it out real quick.

bbbales2:
I got a compile error:

```
math/stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp:33:29: error: no match for ‘operator+=’ (operand types are ‘Eigen::NoAlias<Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > >::adj_Op, Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > >, Eigen::MatrixBase>’ and ‘Eigen::MatrixBase<Eigen::Product<Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, const Eigen::Transpose<const Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > >::adj_Op, Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > > >, const Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > >::adj_Op, Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > > >, Eigen::TriangularView<Eigen::Map<Eigen::Matrix<double, -1, -1>, 8, Eigen::Stride<0, 0> >, 1>, 0> >::ConstTriangularViewReturnType<1>::Type {aka const Eigen::TriangularView<const Eigen::Product<Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, const Eigen::Transpose<const Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > >::adj_Op, Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > > >, const Eigen::CwiseUnaryView<Eigen::MatrixBase<Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > >::adj_Op, Eigen::Map<Eigen::Matrix<stan::math::var_value<double>, -1, -1>, 8, Eigen::Stride<0, 0> > > >, Eigen::TriangularView<Eigen::Map<Eigen::Matrix<double, -1, -1>, 8, Eigen::Stride<0, 0> >, 1>, 0>, 1>}’)
```

bbbales2:
a_few_varmats_b167500c42_matvar is the thing to compare against on the benchmarks page.

SteveBronder:
I have a branch here that should clear up a few things

https://github.com/stan-dev/math/compare/feature/a_few_varmats...feature/a-few-varmats-alt

Though idt you'll like my approach to returns. I just use `auto` and then wrap the return in `ret_type` for each spot

bbbales2:
> Though idt you'll like my approach to returns

I instinctively do not, but I was just wrestling with a complex return type so I guess this is how we should do them (vs. complex return type in two places).

Does this make stuff faster?

SteveBronder:
I'd say marginally. I have a godbolt below and it looks like the .noalias() version saves some stuff

https://godbolt.org/z/7WWYxs

In the top left is the two source tabs, and in the bottom is source 1's asm and source 2's asm with the diff on the top right (though a lot of stuff get's reshuffled so the diff isn't really that helpful beyond the `.noalias()` version has less asm.

At an eyeball it looks better to use noalias here, but we can benchmark the diff to see whatsup

bbbales2:
I merged and pushed.

I'm scared of the `.noalias` stuff now. If I include `.noalias()` then the benchmarks for tcrossprod over here hang: https://github.com/bbbales2/math-benchmarks

The regular mixed mode tests for tcrossprod pass though. I only tested one of the other things.

I gotta do some other stuff. You wanna dig into these?

To use the math benchmarks you gotta build the Google benchmark thing in the Math repo manually: https://github.com/stan-dev/math/issues/2083

stan/math/fwd/fun/tcrossprod.hpp
```diff
@@ -18,7 +18,7 @@ tcrossprod(const EigMat& m) {
     return {};
   }
   const auto& m_ref = to_ref(m);
-  return multiply(m_ref, m_ref.transpose());
+  return m_ref * m_ref.transpose();
```
bbbales2, line 21: I think we could probably just delete the forward implementation of this now.

Multiply returns a scalar if you pass a row vector into this which does not match the prim/reverse mode behavior.

When I added Vector/RowVector tests to tcrossprod this popped up (and we need Vector/RowVector support in tcrossprod cause of how it is used elsewhere in the library)

stan/math/rev/fun/tcrossprod.hpp
t4c1: I think we should consider any remaining loops over matrices as potential performance issues and replace them with Eigen expressions.

SteveBronder:
> I'm scared of the .noalias stuff now. If I include .noalias() then the benchmarks for tcrossprod over here hang: https://github.com/bbbales2/math-benchmarks

Yeah I'm sorting out how to think about these. I think it's like [`restrict`](https://en.cppreference.com/w/c/language/restrict) where you are telling Eigen the lhs is not aliased on the rhs. For `var<mat>` that is cool and makes since when doing `adj()` on one side and `val()` on the other they are different matrices. At first I thought this was fine for `mat<var>` since we only grab one slice since the same thing is true that `val_` and `adj_` are two different things. But I think with an array of structs like that what happens is Eigen stores them in a `var*` so that pointer is being accessed multiple times when getting the adjoint and value.

stan/math/fwd/fun/tcrossprod.hpp
SteveBronder: I'm fine with that, idt this is doing much of anything special

stan/math/rev/fun/columns_dot_self.hpp
```diff
@@ -16,14 +15,21 @@ namespace math {
...
+  arena_t<ret_type> res = arena_x_val.colwise().squaredNorm();
+
+  if (x.size() >= 0) {
+    reverse_pass_callback([res, arena_x, arena_x_val]() mutable {
+      arena_x.adj().noalias() += arena_x_val * (2 * res.adj()).asDiagonal();
+    });
   }
-  return ret;
+
+  return ret_type(res);
```
SteveBronder, line 32: @t4c1 I think when we do this `ret_type(res)` we should actually do `ret_type(std::move(res))`. Looking at the asm on godbolt (ctrl +f in the diff tab on the right hand side for "func ret start") it looks like without the move Eigen has to call malloc and memcpy 

https://godbolt.org/z/WKov3v

SteveBronder: It looks like an anti-pattern but whenever we can remove the whole thing and just return `res` once we can pass around expressiosn

bbbales2: But this pointer points to arena memory so we don't want to hand it off to Eigen right? If it tries to free it it will blow up.

t4c1: In the godbolt example the type or res matches the ret_type. In math code it does not. In other words that malloc will happen until we are allowed to return expressions.

stan/math/rev/fun/simplex_constrain.hpp
```diff
@@ -93,9 +25,48 @@ class simplex_constrain_op {
...
  */
-inline Eigen::Matrix<var, Eigen::Dynamic, 1> simplex_constrain(
-    const Eigen::Matrix<var, Eigen::Dynamic, 1>& y) {
-  return adj_jac_apply<internal::simplex_constrain_op>(y);
+template <typename T, require_eigen_col_vector_vt<is_var, T>* = nullptr>
+inline auto simplex_constrain(const T& y) {
+  using ret_type = plain_type_t<T>;
+
+  size_t N = y.size();
+
```
bbbales2, line 33: WIth a function like this, I'd rather add:

```
if(N == 0) {
  return ret_type(y);
}
```

right here than have all the other code working with size zero stuff (it avoids at least one malloc too).

stan/math/rev/fun/simplex_constrain.hpp
```diff
@@ -93,9 +25,48 @@ class simplex_constrain_op {
...
+                       * inv_logit(log_N_minus_k - value_of(arena_y.coeff(k)));
+    x_val.coeffRef(k) = stick_len * z.coeff(k);
+    stick_len -= x_val(k);
+  }
+  x_val.coeffRef(N) = stick_len;
+
+  arena_t<ret_type> x = x_val;
+
+  if (N > 0) {
+    reverse_pass_callback([arena_y, x, diag, z]() mutable {
```
bbbales2, line 55: [optional] Stylewise I think I like adding arena in the name.

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -88,11 +58,11 @@ Eigen::Matrix<var, R, C> unit_vector_constrain(
...
-template <int R, int C>
-Eigen::Matrix<var, R, C> unit_vector_constrain(
-    const Eigen::Matrix<var, R, C>& y, var& lp) {
-  Eigen::Matrix<var, R, C> x = unit_vector_constrain(y);
-  lp -= 0.5 * dot_self(y);
+template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+inline auto unit_vector_constrain(const T& y, var& lp) {
+  const auto& y_ref = to_ref(y);
+  auto x = unit_vector_constrain(y_ref);
+  lp -= 0.5 * dot_self(y_ref);
```
bbbales2, line 65: [optional] We should look at the constrains in the future so we aren't doing things like this dot_self separately. I think this is the version of the function that gets used in Stan models.

bbbales2: I say this as optional because we will be back through this code in the very short future for `var<mat>` and so I'm not really worried about getting it now. Fixing the includes and turning on this file is enough of a performance bump.

stan/math/rev/fun/determinant.hpp
```diff
@@ -3,56 +3,30 @@
...
-                 .transpose();
-  }
-};
-}  // namespace internal
-
-template <typename T, require_eigen_vt<is_var, T>* = nullptr>
+template <typename T, require_rev_matrix_t<T>* = nullptr>
 inline var determinant(const T& m) {
   check_square("determinant", "m", m);
+
```
bbbales2, line 18: Use the implementation of determinant here: https://github.com/stan-dev/math/blob/654aa5119a3d7870c527cb9f3cedb487b72474e3/stan/math/rev/fun/determinant.hpp#L23

I think it's using the same underlying algorithms but it is clearer what is happening (and it will be faster if the Eigen::MatrixXd doesn't save the decomposition from computing the determinant for later computing the inverse).

stan/math/rev/fun/columns_dot_self.hpp
SteveBronder: Ah! So weird, I wonder why I was seeing speedups on benchmarks when using move? The below shows that move and not move do the same thing. I assumed `Eigen::MapBase` and `Eigen::Matrix` both used `DenseStorage` to hold their data so it would do something with a move. Not true I guess!

https://godbolt.org/z/bEP6Mj

stan/math/rev/fun/dot_self.hpp
```diff
@@ -12,43 +12,42 @@
...
+ * @tparam T type of the vector (must have one compile time dimension equal to
+ * 1)
  * @param[in] v Vector.
  * @return Dot product of the vector with itself.
  */
 template <typename T, require_eigen_vector_vt<is_var, T>* = nullptr>
 inline var dot_self(const T& v) {
-  const Eigen::Ref<const plain_type_t<T>>& v_ref = v;
-  return {new internal::dot_self_vari(v_ref)};
+  const auto& v_ref = to_ref(v);
```
bbbales2, line 25: Revert this (it is slower for `mat<var>`, we will put it back once we have https://github.com/stan-dev/math/pull/2144)

stan/math/rev/fun/inverse.hpp
```diff
@@ -75,16 +22,22 @@ class inverse_vari : public vari {
...
+
   return res;
```
bbbales2, line 41: ret_type(res)

stan/math/rev/fun/log_determinant.hpp
```diff
@@ -4,33 +4,32 @@
...
-
-  double* gradients
-      = ChainableStack::instance_->memalloc_.alloc_array<double>(m.size());
-  Eigen::Map<matrix_d>(gradients, m.rows(), m.cols())
-      = hh.inverse().transpose();
-
-  return var(new precomputed_gradients_vari(hh.logAbsDeterminant(), m.size(),
-                                            varis, gradients));
+template <typename T, require_rev_matrix_t<T>* = nullptr>
+inline var log_determinant(const T& m) {
```
bbbales2, line 17: Use the implementation from here: https://github.com/stan-dev/math/blob/654aa5119a3d7870c527cb9f3cedb487b72474e3/stan/math/rev/fun/log_determinant.hpp#L26

I think it just changes 'fullPivHouseholderQr` to `colPivHouseholderQr` so you could just do that too.

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,50 @@ class matrix_product_vari : public vari {
...
+      }
+      arena_M.adj() += adj_M + adj_C;
+    });
+  }
+
+  return res;
```
bbbales2, line 71: `ret_type(res)`

bbbales2 (commented):
This is my own pull so I can only leave comments

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
```diff
@@ -11,43 +11,26 @@
...
-      LLt.coeffRef(m, n) = LLt.coeffRef(n, m)
-          = dot_product(L.row(m).head((J < (n + 1)) ? J : (n + 1)),
-                        L.row(n).head((J < (n + 1)) ? J : (n + 1)));
-    }
-  }
-  return LLt;
+template <typename T, require_rev_matrix_t<T>* = nullptr>
+inline auto multiply_lower_tri_self_transpose(const T& L) {
+  if (L.size() == 0)
+    return plain_type_t<T>(L);
```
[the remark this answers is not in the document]
SteveBronder: Even if it's 0x10 the size should still be zero though right (rows() * cols())?

stan/math/rev/fun/dot_self.hpp
SteveBronder: I fixed this version up, just had to roll over the input once instead of twice. `custom_vari` will probs help here as well tho'

image

bbbales2: Oh I didn't notice this was different. This is good then.

stan/math/rev/fun/multiply_lower_tri_self_transpose.hpp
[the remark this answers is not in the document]
bbbales2: I went with:

```
return plain_type_t<T>(decltype(multiply_lower_tri_self_transpose(value_of(L)))());
```

which is very verbose and will create and extra vari but will at least be consistent with prim.

stan/math/rev/fun/matrix_power.hpp
```diff
@@ -130,38 +25,50 @@ class matrix_product_vari : public vari {
...
+
+  powers.emplace_back(Eigen::MatrixXd::Identity(N, N));
+  powers.emplace_back(value_of(M_ref));
+  for (size_t i = 2; i <= n; ++i) {
+    powers.emplace_back(powers[1] * powers[i - 1]);
   }
+
+  arena_t<plain_type_t<T>> res = powers[powers.size() - 1];
+
+  if (M.size() > 0) {
```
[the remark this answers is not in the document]
bbbales2: Done

stan/math/rev/fun/simplex_constrain.hpp
bbbales2: Done

bbbales2: Nevermind. For a zero length input we expect a length 1 output.

stan/math/rev/fun/simplex_constrain.hpp
bbbales2: Done

stan/math/rev/fun/unit_vector_constrain.hpp
bbbales2: Did not do.

stan/math/rev/fun/determinant.hpp
bbbales2: Done

stan/math/rev/fun/inverse.hpp
bbbales2: Not necessary cause return type isn't auto.

stan/math/rev/fun/log_determinant.hpp
bbbales2: Done

stan/math/rev/fun/matrix_power.hpp
bbbales2: Not necessary because the return type isn't auto.

bbbales2:
@SteveBronder I think we wait until the address sanitizer stuff is merged before merging this (https://github.com/stan-dev/math/pull/2146), but it's yours to approve now. (we should also re-run the tests once the address sanitizer is in)

bbbales2:
@SteveBronder tests pass! Ready for review. The performance numbers are up comparing this to August 12th (https://bbbales2.shinyapps.io/math-shiny-perf/), but they're fishy and I think we should compare against develop.

The reason I say they're fishy is cuz it says in this branch lcdf performance drops 10%+ in the non-vectorized case for small N. I checked with a custom benchmark and that was not the case. I'm rerunning the benchmarks on this branch vs. develop with a single core (which will probably take until sometime tomorrow to finish).

SteveBronder:
Just pushed up some changes. I added a new type trait [`is_eigen_dense_dynamic`](https://github.com/stan-dev/math/pull/2106/files#diff-e555835acf1034bf3c85d502cbf45d91). I need it for the stan repo stuff and I think when we turn on the expression stuff we will actually want to use this over the type trait that checks if the type is a dynamic matrix (`is_eigen_dense_dynamic` just checks that the type inherits from `Eigen::DenseBase` and has dynamic rows and columns so it also accepts `Eigen::Array<T, -1, -1>`)

stan/math/rev/fun/log_sum_exp.hpp
```diff
@@ -70,7 +70,7 @@ inline auto log_sum_exp(const T& x) {
   return apply_vector_unary<T>::reduce(x, [](const auto& v) {
     auto arena_v = to_arena(v);
 
-    var res = log_sum_exp(value_of(arena_v));
+    var res = log_sum_exp(arena_v.val());
```
bbbales2, line 73: `.val()` fails here. Run `./runTests.py test/unit/math/mix/fun/log_sum_exp_test`

It is the Eigen const bug or whatever. `val_op` works but I would just rather go with `value_of(arena_v)`.

bbbales2:
`is_eigen_dense_dynamic` is this something we need for this release? If not, separate pull request.

SteveBronder:
It's not needed but I don't see a reason to open up a whole other PR for 6 lines of code. And if we make another PR I'll have to wait till that's merged before we can do the PR for the stan repo stuff

bbbales2 (commented):
is_eigen_dense_dynamic looks okay, `log_sum_exp` tests fail

stan/math/rev/fun/log_sum_exp.hpp
SteveBronder: Oh weird, I think it was actually because of the line in the prim version of `log_sum_exp()`

```
const Eigen::Ref<const plain_type_t<decltype(v)>>& v_ref = v;
```

If we use `to_ref()` there instead then we can use `.val()`

```
const auto& v_ref = to_ref(v);
```

bbbales2: Okay good enough for me.

bbbales2:
I'm kindof nervous about the tcrossprod changes. If `tcrossprod` uses multiply that means `tcrossprod` of a row_vector returns a scalar, not a matrix.

Stan docs show matrix: https://mc-stan.org/docs/2_24/functions-reference/dot-products-and-specialized-products.html

Also we should match behavior between `crossprod`, `tcrossprod`, and `multiply_lower_transpose`. Given how late this is in the whole process I'd rather just go with what was there.

bbbales2:
Prim still returns a matrix. Am I reading this wrong or something? I don't know how this would pass tests if `tcrossprod(row_vector)` returned a scalar.

SteveBronder:
For the `operator*()` docs here

https://mc-stan.org/docs/2_24/functions-reference/matrix-arithmetic-operators.html

```
real operator*(row_vector x, vector y)
The product of the row vector x and vector y
```

But yeah for `tcrossprod` I am seeing we return a matrix no matter what so I'll fix that

bbbales2:
We should probably update the test framework so this doesn't pass.

This code:
```
  Eigen::RowVectorXd ra3(3);
  ra3 << 1, 2, 3;
  std::cout << stan::math::test::type_name<decltype(stan::math::tcrossprod(ra3))>() << std::endl;
  std::cout << stan::math::test::type_name<decltype(stan::math::tcrossprod(stan::math::promote_scalar<stan::math::var>(ra3)))>() << std::endl;
  std::cout << stan::math::test::type_name<decltype(stan::math::tcrossprod(stan::math::promote_scalar<stan::math::fvar<double>>(ra3)))>() << std::endl;
```

Prints:
```
 stan::math::test::type_name() [Arg = Eigen::Matrix<double, 1, 1, 0, 1, 1>
 stan::math::test::type_name() [Arg = Eigen::Matrix<stan::math::var_value<double, void>, 1, 1, 0, 1, 1>
 stan::math::test::type_name() [Arg = stan::math::fvar<double>
```

bbbales2:
Yeah our operator* is a little weird.

SteveBronder:
Yeah we should have a check that if one specialization returns back an eigen matrix etc then the other ones should as well.

I think the multiply version makes sense for row vector x vector, if anything I'd say we should have tcrossprod and crossprod return the same types as multiply

bbbales2:
> if anything I'd say we should have tcrossprod and crossprod return the same types as multiply

Yeah probably. And an inconsistency now wouldn't be a huge problem because the signature in the Stan manual is matrix only, so `tcrossprod(row_vector)` actually isn't a thing yet I guess.

Alright I approve of your changes here. You're the one with approve/merge abilities, so if this passes before the end of the day, great, if not, I suppose it's not important if this doesn't get in.

bbbales2:
@SteveBronder what's next step for getting this merged? I review the last few changes you've made (since #a7045c) and then we're good? Or is there anything else?

SteveBronder:
>  I review the last few changes you've made (since #a7045c) and then we're good? 

Yeah that should be it!

test/unit/math/mix/fun/simplex_constrain_test.cpp
```diff
@@ -23,6 +23,7 @@ void expect_simplex_transform(const T& x) {
   auto f2 = [](const auto& x) { return g2(x); };
   auto f3 = [](const auto& x) { return g3(x); };
   stan::test::expect_ad(f1, x);
+  stan::test::expect_ad_matvar(f1, x);
   stan::test::expect_ad(f2, x);
   stan::test::expect_ad(f3, x);
```
bbbales2, line 28: Oh weird, the reverse mode specializations are missing for `f2` and `f3`? Make an issue. This is kindof weird and nobody will be back through here in a while so we should write it down even if it is a tiny thing.

I guess the larger issue could be making the `x_constrain(x, lp)` functions more efficient. For instance, in unit vector constrain there is an extra dot product that happens:

```
inline auto unit_vector_constrain(const T& y, var& lp) {
  auto x = unit_vector_constrain(y);
  lp -= 0.5 * dot_self(y);
  return x;
}
```

That `dot_self` is effectively already computed inside `unit_vector_constrain`.

stan/math/rev/fun/dot_self.hpp
```diff
@@ -41,11 +41,9 @@ inline var dot_self(const T& v) {
...
 template <typename T, require_var_matrix_t<T>* = nullptr>
 inline var dot_self(const T& v) {
-  arena_t<T> arena_v = v;
-  var res = arena_v.val().dot(arena_v.val());
-  reverse_pass_callback([res, arena_v]() mutable {
-    arena_v.adj() += (2.0 * res.adj()) * arena_v.val();
-  });
+  var res = v.val().dot(v.val());
+  reverse_pass_callback(
+      [res, v]() mutable { v.adj() += (2.0 * res.adj()) * v.val(); });
```
bbbales2, line 46: Huh, neat, in this case it's fine to just copy the `v` in since it is a `var_value`.

bbbales2 (commented):
I approve, but I can't actually approve and merge since this is my own pull. One comment on `simplex_constrain`.

stan/math/rev/fun/dot_self.hpp
SteveBronder: Yep! You can do the same thing with `var<matrix>`!

test/unit/math/mix/fun/simplex_constrain_test.cpp
SteveBronder: Yeah I would but tbh I'm just too dumb and couldn't figure them out lol. Would you mind adding these? Also where are these signatures used?

SteveBronder:
Let me have one quick look over all this then we can approve and merge

test/unit/math/mix/fun/simplex_constrain_test.cpp
bbbales2: `y = *_constrain(x)` goes from unconstrained to constrained space without Jacobian adjustment. This is probably not used much (actually optimization might use it).

`y = *_constrain(x, lp)` goes from unconstrained to constrained with Jacobian adjustment. It would be the one that is used for the constraints when you are sampling.

We should probably do some reworking to focus optimization efforts on the second.

If it's just these four constraints in here could probably rework this in a few hours of wiggling with the code. Want me to do it in this pull?

SteveBronder:
fyi I went through and saw we weren't testing the constrain functions in the expression tests so I added them there and had to fix a few things so they would work

bbbales2:
@SteveBronder I don't see additional changes to the expressions test file here? Possibly forgot to commit something there?

SteveBronder:
Right here?

https://github.com/stan-dev/math/pull/2106/files#diff-b8d9f944ef9a96f0873c189c10020996e2cceab013ebff09c2bd740aa1faacb4

SteveBronder:
it just adds the constrain funcs to the gen test python file

SteveBronder:
Word no worries!

test/expressions/generateExpressionTests.py
```diff
@@ -90,6 +90,24 @@ def add_extra_signatures(res):
         "vector unit_vector_constrain(row_vector, real)",
         "vector unit_vector_free(vector)",
         "vector unit_vector_free(row_vector)",
+        "vector positive_ordered_constrain(vector)",
+        "vector positive_ordered_constrain(row_vector)",
```
bbbales2, line 94: I don't think any of these are actually expected to work with row_vectors (check the tests)

SteveBronder: idt they need to, but when I rewrote these they can now. Should I update the tests to use row vectors?

bbbales2: Revert to just vectors. For instance:

```
vector positive_ordered_constrain(row_vector)
```

Doesn't really make sense (returns a `vector` give a `row_vector`). We aren't using `row_vector` here so I'd rather not bother with adding it. I wonder how that works under the hood though? Cause we can definitely positive constrain Stan arrays, etc. I'd rather not dig into that here though and get this pull in.

I'm trying to figure out how to do reverse mode on `simplex_constrain` with lp and failing :(.

bbbales2: Oh yeah also I've checked all your changes up to this point again (all the reference stuff). Looks fine to me.

SteveBronder: Word I'll fix this then we should be good to go!

test/unit/math/mix/fun/simplex_constrain_test.cpp
bbbales2: @SteveBronder I looked into `simplex_constrain`. I have code for it, but there was some guess and check involved and I want to work it out properly to make sure I'm not wasting memory. Let's break out making the two argument constrains fast into another pull.

bbbales2: (if this is okay, let me know and I'll make the issue and put the `simplex_constrain` code there until we have time to get around to it)

SteveBronder:
I decided it would be better to make `is_var_col_vector` and friends which should be nice to use in other functions as well.

On a side note I think we should rename `is_var_matrix`/`is_rev_matrix` to `is_var_eigen`/`is_rev_eigen`. Though that still leaves what to do with renaming `is_matrix`. `is_eigen`. We could have `is_eigen_or_var_eigen` but that's kind of a confusing mouthful

stan/math/rev/meta/is_rev_matrix.hpp
```diff
@@ -22,5 +22,42 @@ struct is_rev_matrix<
                      math::disjunction<is_eigen<T>, is_eigen<value_type_t<T>>>>>
     : std::true_type {};
 
+/** \ingroup type_trait
```
bbbales2, line 25: These docs don't appear to be for these functions.

bbbales2:
@rok-cesnovar is this test messed up? It says it's been stuck in one place for 17 hours

rok-cesnovar:
Seems so. Restarted now. If it repeats we need to look into it.

SteveBronder (approved):
lgtm! @t4c1 do you want to do another pass through? Else I'm cool to merge this when tests pass

t4c1 (changes requested):

bbbales2:
@SteveBronder I'm gonna fix the stuff @t4c1 pointed out (just so we aren't doing the same thing at the same time)

stan/math/prim/err/check_vector.hpp
```diff
@@ -38,6 +38,34 @@ inline void check_vector(const char* function, const char* name,
...
+    return;
+  } else {
+    std::ostringstream msg;
+    msg << ") has " << x.rows() << " rows and " << x.cols()
+        << " columns but it should be a vector so it should "
+        << "either have 1 row or 1 column";
+    std::string msg_str(msg.str());
+    invalid_argument(function, name, scalar_type_t<VarMat>(), "(",
+                     msg_str.c_str());
+  }
```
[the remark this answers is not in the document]
bbbales2: I replaced these four functions with one.

stan/math/prim/err/check_vector.hpp
```diff
@@ -54,6 +82,25 @@ constexpr inline void check_vector(const char* function, const char* name,
...
+ * Eigen types with compile time rows or columns equal to 1.
+ * whether it is a row or column vector.
+ * @tparam EigVec A type derived from `EigenBase` with 1 row or column
+ * @param function Function name (for error messages)
+ * @param name Variable name (for error messages)
+ * @param x Matrix
+ * @throw <code>std::invalid_argument</code> if x is not a row or column
+ *   vector.
+ */
+template <typename VarVec, require_var_matrix_t<VarVec>* = nullptr,
```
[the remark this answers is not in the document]
bbbales2: I don't think this explicit optimization is necessary. Either the compiler will catch it, or it's not gonna really matter compared to the computation that goes into the unit vector constrain.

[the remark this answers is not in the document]
bbbales2: (So I removed it)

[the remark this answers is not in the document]
bbbales2: It turns out that `unit_vector_constrain` was the only function that was using this (there's a different version in the opencl code it looks like) and there are already requires on `unit_vector_constrain` to make sure stuff is a vector. So I just removed the `check_vector` tests there (so now I don't think `check_vector` is used anywhere else).

SteveBronder:
@t4c1 good to go?

t4c1 (approved):

PR #2115, opened 2020-09-30
operator+ for var matrices and matrix of vars

SteveBronder:
## Summary

This adds `add()` and `operator+` for var matrices and adds overloaded functions for `add()` using `reverse_pass_callback()` for matrices of vars. I can add subtraction here as well as the code will look nearly the same

## Tests

Tests were added for `operator+` for all the mixed types it can take in. The tests use `add()` because for var types `add()` now has a specialization for var types that divert to `operator+`

## Side Effects

None

## Release notes

Adds addition overloads for var matrices and matrices of vars

## Checklist

- [x] Math issue #2101 


SteveBronder:
I have a branch below that does this same thing for subtract as well. I'll post some speed tests tonight

https://github.com/stan-dev/math/tree/feature/varmat-operatorminus

stan/math/rev/core/operator_addition.hpp
```diff
@@ -116,10 +138,153 @@ inline var operator+(const var& a, Arith b) {
...
+ * @tparam VarMat2 A matrix of vars or a var with an underlying matrix type.
+ * @param a First variable operand.
+ * @param b Second variable operand.
+ * @return Variable result of adding two variables.
+ */
+template <typename VarMat1, typename VarMat2,
+          require_all_rev_matrix_t<VarMat1, VarMat2>* = nullptr>
+inline auto operator+(const VarMat1& a, const VarMat2& b) {
+  check_matching_dims("operator+", "a", a, "b", b);
+  using op_ret_type = decltype(a.val() + b.val());
```
andrjohns, line 157: I think you either need to change this to:
```
plain_type_t<decltype(a.val() + b.val())>;
```

Or remove the ```.eval()``` from ```ret```. Otherwise ```op_ret_type``` will be a ```CWiseBinaryOp``` while the result of ```(a.val() + b.val()).eval()``` will be a plain object. Unless ```promote_var_matrix_t``` automatically returns a plain type?

stan/math/rev/core/operator_addition.hpp
```diff
@@ -116,10 +138,153 @@ inline var operator+(const var& a, Arith b) {
...
+ */
+template <typename Arith, typename VarMat,
+          require_st_arithmetic<Arith>* = nullptr,
+          require_rev_matrix_t<VarMat>* = nullptr>
+inline auto operator+(const Arith& a, const VarMat& b) {
+  return b + a;
+}
+
+/**
+ * Addition operator for an arithmetic matrix and variable (C++).
```
andrjohns, line 216: Super pedantic, but this templating allows Arrays as well. So maybe 'arithmetic Eigen types' or the like?

stan/math/rev/core/operator_addition.hpp
```diff
@@ -116,10 +138,153 @@ inline var operator+(const var& a, Arith b) {
...
+ *
+ * @tparam Var A `var_value` with an underlying arithmetic type.
+ * @tparam EigMat An Eigen Matrix type with an arithmetic Scalar type.
+ * @param a First variable operand.
+ * @param b Second variable operand.
+ * @return Variable result of adding two variables.
+ */
+template <typename Var, typename EigMat,
+          require_eigen_vt<std::is_arithmetic, EigMat>* = nullptr,
+          require_var_vt<std::is_arithmetic, Var>* = nullptr>
```
andrjohns, line 226: Not related to this pull but I didn't realise we had this ```require_var_vt```, that will simplify some of the wacky templating I've been resorting to. These require generics have been such a good addition!

andrjohns (changes requested):
This all looks great! Couple minor changes in there.

I might be a bit of a pain though. I tried this ```reverse_pass_callback``` approach with the unary functions and performance got worse, since multiple passes over the same inputs to pull out values and adjoints 'cost' more than was gained by the callback. I wonder if this will have the same problem?

Would you mind benchmarking these against against an ```apply_scalar_binary``` implementation?

Something like: 
```
inline auto operator+(const T1& a, const T2& b) {
  return apply_scalar_binary(
      a, b, [&](const auto& c, const auto& d) { return add(c, d); });
}
```

Where there are only scalar overloads for ```add``` (i.e., comment out the container specialisations or something).

Let me know if you don't have the time and I can work them up over the weekend as well.

stan/math/prim/err/is_matching_dims.hpp
```diff
@@ -46,11 +46,13 @@ inline bool is_matching_dims(const Eigen::Matrix<T1, R1, C1>& y1,
  * @param y2 second matrix to test
  * @return <code>true</code> if the dimensions of the matrices match
  */
-template <bool check_compile, typename T1, typename T2, int R1, int C1, int R2,
-          int C2>
-inline bool is_matching_dims(const Eigen::Matrix<T1, R1, C1>& y1,
-                             const Eigen::Matrix<T2, R2, C2>& y2) {
-  return !(check_compile && (R1 != R2 || C1 != C2)) && is_matching_dims(y1, y2);
+template <bool check_compile, typename EigMat1, typename EigMat2,
+          require_all_matrix_t<EigMat1, EigMat2>* = nullptr>
```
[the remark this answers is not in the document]
t4c1: I think this is not necessary. AFIK all expressions also define `RowsAtCompileTime` and `ColsAtCompileTime`.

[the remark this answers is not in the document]
andrjohns: Thanks Tadej, will mark this resolved

SteveBronder:
Woof weird error I need to sort out, but good news is that the benchmarks are v cool.

For `Matrix<var>`

image


Then for `var<Matrix>`

image

andrjohns:
Just realised that my ```apply_scalar_binary``` suggestion doesn't make much sense since there aren't scalar overloads of ```var<matrix>```, so feel free to ignore me on that one

stan/math/rev/fun/add.hpp
```diff
@@ -0,0 +1,181 @@
...
+      arena_a.adj() += ret.adj();
+      arena_b.adj() += ret.adj();
+    } else {
+      for (Eigen::Index i = 0; i < ret.size(); ++i) {
+        arena_a.adj().coeffRef(i) += ret.adj().coeffRef(i);
+        arena_b.adj().coeffRef(i) += ret.adj().coeffRef(i);
+      }
+    }
+  });
+  return ret_type(ret);
```
SteveBronder, line 43: @t4c1 I always get confused by RVO and NRVO. if I do this pattern where I wrap the `arena_matrix` in the return type, should I do `ret_type(std::move(ret))` or can/will NRVO already perform a move ctor? idt we can get copy ellision since they are different type right? Looking at the constructors for Eigen's `DenseStorage` (that `Matrix` uses) we have

```cpp
// copy ctor
    EIGEN_DEVICE_FUNC DenseStorage(const DenseStorage& other)
      : m_data(internal::conditional_aligned_new_auto<T,(_Options&DontAlign)==0>(other.m_rows*other.m_cols))
      , m_rows(other.m_rows)
      , m_cols(other.m_cols)
    {
      EIGEN_INTERNAL_DENSE_STORAGE_CTOR_PLUGIN(Index size = m_rows*m_cols)
      internal::smart_copy(other.m_data, other.m_data+other.m_rows*other.m_cols, m_data);
    }
// move ctor
    EIGEN_DEVICE_FUNC
    DenseStorage(DenseStorage&& other) EIGEN_NOEXCEPT
      : m_data(std::move(other.m_data))
      , m_rows(std::move(other.m_rows))
      , m_cols(std::move(other.m_cols))
    {
      other.m_data = nullptr;
      other.m_rows = 0;
      other.m_cols = 0;
    }
```

Since we make a copy of `ret` for the `reverse_pass_callback` is it correct to use a move in that spot?

SteveBronder:
> Would you mind benchmarking these against against an apply_scalar_binary implementation?

So @andrjohns pretty interesting! Testing `matrix<var>` with the `apply_scalar_binary`as

```cpp
    inline auto apply_add(const T1& a, const T2& b) {
      check_matching_dims("add", "a", a, "b", b);
      return apply_scalar_binary(
          a, b, [&](const auto& c, const auto& d) { return add(c, d); });
    }
```

vs this branch gives the following. The graph is `overload_version / apply_version` for two vars of vectors and two vars of matrice. Bigger numbers means the apply version is better

image


Pretty interesting! `apply` is wayyy better at small N, but for big N the overload is better. Though this only applies to the <var, var> impl. I'm running the other benchmarks now and it looks like the other cases with `<double, var>`, `<var, double>` the overload is better. I'll post those up when they finish running

stan/math/rev/fun/add.hpp
SteveBronder: Also we can remove that line just be `ret` once we can pass around general eigen types

SteveBronder: One more thing,

If we overloaded `internal::conditional_aligned_new_auto`, internal::conditional_aligned_delete_auto`, and friends for `var` in Eigen, could we use our stack allocator there? Then for arena_matrix, we'd know the `Eigen::Matrix<var>`'s `var*` is already on our stack so we can just take a copy of of the pointer without having to worry about it being destroyed

t4c1: > if I do this pattern where I wrap the arena_matrix in the return type, should I do ret_type(std::move(ret)) or can/will NRVO already perform a move ctor?

`Eigen::Matrix` can not reuse memory from `arena_matrix` (they use different allocators), so there will be a copy regardless of theal move. So the move is not necessary.

> idt we can get copy ellision since they are different type right?

Right.

> Also we can remove that line just be ret once we can pass around general eigen types

Correct.

> If we overloaded internal::conditional_aligned_new_auto, internal::conditional_aligned_delete_auto, and friends for var in Eigen, could we use our stack allocator there?

That is an interesting idea. In my opinion it is worth making a prototype and testing if that can work (across different OS-es). Although running into some issues is likely.

SteveBronder: Got a branch of the Eigen stuff working.

https://github.com/stan-dev/math/tree/feature/eigen-allocs

Now sure if it's cross platform tho, I think it would fail on anything that needs unaligned memory

SteveBronder: > Eigen::Matrix can not reuse memory from arena_matrix (they use different allocators), so there will be a copy regardless of theal move. So the move is not necessary.

Is Matrix an allocator aware type? I'm not sure I'm following. Looking at `MatrixBase` it seems like it uses `DenseStorage` to hold the data. `DenseStorage` for dynamic types has (`m_data`)` as `T*`. Since arena_matrix inherits from `Map` shouldn't it be able to use the move constructor to just assign the arena matrices pointer to `T*`?

\https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/DenseStorage.h#L364

andrjohns:
Great, thanks for running those! Can you post the benchmarking code somewhere? I might shamelessly cannibalise it for my own testing.

Any chance I could burn some more of your computing time? Can you run the same benchmarks with ```elt_multiply```? I think that might perform similarly

stan/math/rev/fun/add.hpp
t4c1: > Is Matrix an allocator aware type?

It is not (an that is the problem). 

> Since arena_matrix inherits from Map shouldn't it be able to use the move constructor to just assign the arena matrices pointer to T*`?

I think I read somewhere Eigen uses some internal wrappers around new/some other allocator in a platform-dependent way. Arena matrix uses our arena allocator. As the allocators are different the memory should not be moved between the types (if it was It could be allocated and released by different allocation mechanisms, leading into all kinds of opportunities for undefined behavior).

t4c1: I looked over your Eigen allocator branch. Is it the correct branch? It just seems to add realloc and makes stan allocator use Eigen allocator. It also mixes Eigen's allocator with `std::realloc`, which is bad. You should always use the same allocator to allocate/reallocate/release memory. I see no changes in how Eigen matrices are allocated. Also there is a number of unrelated changes.

SteveBronder:
> Great, thanks for running those! Can you post the benchmarking code somewhere? I might shamelessly cannibalise it for my own testing.

Sure! I'm using a bunch of Ben's callback code from [here](https://github.com/bbbales2/math-benchmarks/tree/master/src) below is the cpp and R code I use to make the graphs

https://gist.github.com/SteveBronder/7d8ba1813638cd5f53663f5dcb976c75

> Any chance I could burn some more of your computing time? Can you run the same benchmarks with elt_multiply? I think that might perform similarly

Yes and you are right! At least they are close. Here is raw wall time

image


performance

image

Def a similar story where cool and good for vector scalar. We can fixup `elt_multiply`

stan/math/rev/fun/add.hpp
SteveBronder: > I think I read somewhere Eigen uses some internal wrappers around new/some other allocator in a platform-dependent way. 

Sort of. So in [`memory.h`](https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/util/Memory.h#L34) they do a bunch of macros trying to pull out what compiler / platform they are on and whether that has some already predefined memory alignment. If that fails then they just use their own handrolled aligned malloc that's `malloc()` with some bitshifting around it.

Arena matrix uses our arena allocator. As the allocators are different the memory should not be moved between the types (if it was It could be allocated and released by different allocation mechanisms, leading into all kinds of opportunities for undefined behavior).

Yeah so I thought this too. But I did this for tcrossprod like in [here](https://gist.github.com/SteveBronder/cca7d9c00cfcd34fd44c37aef53a8035), then tried causing a use after free like

```cpp
TEST(MathMixMatFun, try_to_ub) {
  using stan::math::var;
  Eigen::MatrixXd A = Eigen::MatrixXd::Random(50, 50);
  for (int i = 0; i < 11; ++i) {
    var lp = 0;
    {
      auto x_cross = stan::math::tcrossprod(A);
      // Force a copy so no chance x_cross gets stored somewhere else
      Eigen::Matrix<var, -1, -1> x_cross_copy = x_cross.eval();
      auto x_sum = stan::math::sum(x_cross_copy);
      lp += x_sum;
    }
    lp.grad();
    stan::math::zero_adjoints();
    lp.grad();
    stan::math::recover_memory();
  }
}

```

But when I run this through valgrind, heaptrack, and clang's sanitizers for addresses, memory, and UB none of them complain. Are we somehow paying for two copies by accident somewhere? Or is this not testing what would cause a use after free?

> I looked over your Eigen allocator branch. Is it the correct branch? 

It is and it's v weird because I needed to move around the order of includes and it was just easier to slam the code right where I wanted it (_prototype_ branch!).


> It just seems to add realloc and makes stan allocator use Eigen allocator. 

Yeah so how eigen does this is kind of complicated.

1. tmk, Eigen allocates everything for dense types via `DenseStorage`. Is that not correct?
2. `DenseStorage` uses a bunch of these functions like `conditional_aligned_delete_auto`, `conditional_aligned_realloc_new_auto`, etc. for allocating new mem. `aligned_malloc`/`aligned_free` is never used directly in eigen (except for a few places in the tensor unsupported repo but idt we support that)
3. So my game plan here is to just overload those functions for `var*`. That means some weird stuff like figuring out what a realloc for our allocator looks like lol

Since Eigen allows `realloc()` through .resize()` we need to have some concept of a realloc for our allocator. I think the way I have it is reasonably correct, just don't do anything for smaller resizes so eigen shrinks the rows/cols. For larger sizes check if there's enough space in the current block and if not make a new block. If there is enough space we do the same thing Eigen does when it reallocs but from memory sitting in our arena

> It also mixes Eigen's allocator with std::realloc, which is bad. You should always use the same allocator to allocate/reallocate/release memory. 

Eigen uses std::realloc! It makes sense because at the eod they actually use malloc and then just do some bit shifting. The stuff I have in our allocators `realloc` is essentially a copy/paste of their handmade one.

https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/util/Memory.h#L127

> I see no changes in how Eigen matrices are allocated. Also there is a number of unrelated changes.

If you put some `puts("")` in the functions I wrote up that overload the ones in `memory.h` you'll see Eigen is using those to allocate memory

t4c1: In your example `ret_type(std::move(res));` the ret_type and type of res are still different so this will make a copy regardless of the call to `std::move`. So your example is not UB. You would have to add a constructor for Eigen matrix that would accept a rvalue reference to arena matrix and move the memory to get UB. Also are you proposing to use UB just because it supposedly works on your computer?

> tmk, Eigen allocates everything for dense types via DenseStorage. Is that not correct?

No idea.

> they actually use malloc and then just do some bit shifting.

Than you should use their realloc that includes the bit shifting.

stan/math/prim/err/is_matching_dims.hpp
```diff
@@ -23,9 +23,9 @@ namespace math {
  * @param y2 second matrix to test
  * @return <code>true</code> if the dimensions of the matrices match
  */
-template <typename T1, typename T2, int R1, int C1, int R2, int C2>
-inline bool is_matching_dims(const Eigen::Matrix<T1, R1, C1>& y1,
-                             const Eigen::Matrix<T2, R2, C2>& y2) {
+template <typename EigMat1, typename EigMat2,
+          require_all_matrix_t<EigMat1, EigMat2>* = nullptr>
```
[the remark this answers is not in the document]
SteveBronder: I think `matrix_t` may just be bad wording here. `require_*_matrix_t*` accepts either an eigen type or a `var_value<T>` where `T` is an eigen type. I had a hard time naming this and am open to whatever

stan/math/rev/core/operator_addition.hpp
SteveBronder: This is another weird wording place since `require_rev_matrix_t` allows for var<Eigen> and Eigen types. We def need better wording for this

SteveBronder:
Alright I had another go at this because I thought it was weird that `apply_scalar_binary` and I got it figured out but it was actually kind of suprising


image

image

So besides the weird hiccup for 256x256 matrices everything is at least as fast as `apply_scalar_binary` or faster except for the N=4 cases (which imo I think is fine given the gains are moderatly sized N)

The trick is just pulling out the `ref_adj` when looping instead of getting the value by reference. It saves an avx instruction and a register on the loop which closes the gap between the apply and loop versions. I looked over a few compilers and instruction sets and this seems to work across a lot of different combos

```cpp
    for (Eigen::Index i = 0; i < ret.size(); ++i) {
      const auto ref_adj = ret.adj().coeffRef(i);
      arena_a.adj().coeffRef(i) += ref_adj;
      arena_b.adj().coeffRef(i) += ref_adj;
    }
```

The below has both loop examples and a diff between their asm.  `.L36` has the loop that pulls out the `ret.adj()` at each loop. 

https://godbolt.org/z/nzd6Yj

SteveBronder:
Also I was getting a bunch of weird errors header errors around Eigen_NumTraits / typedefs / and read_var so I moved them over to core. I think this makes sense because those files depend only on `core`

stan/math/rev/core/operator_addition.hpp
SteveBronder: > Unless promote_var_matrix_t automatically returns a plain type?

Yep that's what happens!

SteveBronder:
If anyone has a minute this should be ready for review!

andrjohns:
Sorry about the delay, will take a look at this today

stan/math/prim/eigen_plugins.h
```diff
@@ -149,6 +149,19 @@ struct adj_Op {
   EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE
     std::enable_if_t<!std::is_pointer<T>::value, reverse_return_t<T>>
       operator()(T &v) const { return v.vi_->adj_; }
+
+  //Returns adjoint from a vari*
+  template<typename T = Scalar>
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE
+    std::enable_if_t<std::is_pointer<T>::value, reverse_return_t<T>>
+      coeffRef(T &v) { return v->adj_; }
```
andrjohns, line 157: Is this needed? Shouldn't ```mat.adj().coeffRef(i,j)``` 'just work'?

stan/math/rev/core/operator_addition.hpp
```diff
@@ -101,12 +102,185 @@ inline var operator+(const var& a, Arith b) {
...
+inline auto add(const VarMat1& a, const VarMat2& b) {
+  check_matching_dims("add", "a", a, "b", b);
+  using op_ret_type = decltype(a.val() + b.val());
+  using ret_type = promote_var_matrix_t<op_ret_type, VarMat1, VarMat2>;
+  arena_t<VarMat1> arena_a(a);
+  arena_t<VarMat2> arena_b(b);
+  arena_t<ret_type> ret(arena_a.val() + arena_b.val());
+  reverse_pass_callback([ret, arena_a, arena_b]() mutable {
+    for (Eigen::Index i = 0; i < ret.size(); ++i) {
+      const auto ref_adj = ret.adj().coeffRef(i);
```
andrjohns, line 128: Given the ```const auto``` return type is returning a template, is there a preference for doing this rather than just:
```
arena_a.adj().coeffRef(i) += ret.adj().coeffRef(i);
```
?

andrjohns (changes requested):
Couple of very minor queries, but otherwise looks great! Thanks for going through the ```apply_scalar_binary``` benchmarking rigmarole.

bbbales2:
@SteveBronder yoyo Steve can you do the polishing on this? I want my `operator+`

SteveBronder:
Yes! Trying to finish up the slicing stuff today but I will clean up based on the review comments (thanks @andrjohns !) tmrw

bbbales2:
Gogogogogo

stan/math/rev/core/operator_addition.hpp
SteveBronder: Yes so this is a weird specific C++ thing. I'll pull out a godbolt tmrw but essentially just loading this return's adjoint once each loop instead of getting a reference for each accumulation makes for much nicer SIMD asm

stan/math/prim/eigen_plugins.h
SteveBronder: Oh hm maybe I did something stupid let me look at this

SteveBronder:
Ooof yeah sorry I'll try to fix this up tonight or tomorrow. The assign and subset tests are way more than I thought they were

stan/math/rev/core/operator_addition.hpp
SteveBronder: So essentially doing a `.coeffRef()` for the return, when working with `var<mat>`, in each accumulation causes the compiler to have to store a register for the reference and then fetch the value the reference points to for each accumulation in the loop. So since we use it more than once it's actually better to just pull out the value once and reuse it for both accumulations

In the below source 1 pull out the matrix value while source 2 uses `.coeffRef()`. You can ctrl+f for "beginnn" in the diff on the right to see the start and end of the loop code.

The main stuff is in the code that loops around like .L36 or .L35 where some of the registers we move stuff around in look a lil nicer. Lol it's very much a micro optimization but when I benchmarked this it seemed to matter


https://godbolt.org/z/W77Tj4

stan/math/prim/eigen_plugins.h
SteveBronder: Totally unnecessary, deleted!

stan/math/rev/core/operator_addition.hpp
SteveBronder: > Thanks for going through the `apply_scalar_binary` benchmarking rigmarole.

That is exactly why I found this lil optim lol

bbbales2: Interesting find. I guess that's like an aliasing thing. Like the compiler doesn't know the memory on the left is not the memory on the right, and so it can't actually do an optimization to avoid the two loads.

bbbales2:
> Ooof

The cheer squad appreciates the effort

andrjohns:
> The cheer squad appreciates the effort

+1 Cheer

Let me know when I can take a gander at this again

stan/math/rev/core/operator_addition.hpp
SteveBronder: ^Yeah I think that's right

SteveBronder:
@andrjohns should be ready to rock!

andrjohns (dismissed):
All good here! Woohoo!

SteveBronder:
@andrjohns can you reclick the approve thing? Something happened in jenkins and I had to kick this off again

andrjohns (approved):

SteveBronder:
@serban-nicusor-toptal is there something up with jenkins? The normal button to view jenkins stuff is gone from the pr and this pr has been waiting since Friday. Is there a backlog of PRs?

serban-nicusor-toptal:
I had to reboot Jenkins and it may have corrupted the job url here. [This](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2115/39/pipeline) is the last build that ran, I've also restarted it, button should be back too.

SteveBronder:
cool much appreciated!

PR #2116, opened 2020-09-30
return an arena_t type from promote_var_matrix_t

SteveBronder:
## Summary

little potential bug fix. We should return an arena matrix from promote_var_matrix_t so people don't have to wrap it's result in an `arena_t<>`

## Checklist

- [ ] Copyright holder: Steve Bronder

### DIFF
--- a/stan/math/prim/meta/promote_scalar_type.hpp
+++ b/stan/math/prim/meta/promote_scalar_type.hpp
@@ -81,7 +81,10 @@ struct promote_scalar_type<T, S,
  * @tparam S input matrix type
  */
 template <typename T, typename S>
-struct promote_scalar_type<T, S, require_t<bool_constant<is_eigen<S>::value && !is_arena_matrix<S>::value>>> {
+struct promote_scalar_type<
+    T, S,
+    require_t<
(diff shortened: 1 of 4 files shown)


SteveBronder:
Hmm actually I'm thinking about this now, @t4c1 @bbbales2 what do yinz think of this? I put this up because in the addition PR I totally forgot to wrap `promote_var_t<>` in an `arena_t<>` and that feels like a pretty common error. But this also becomes hard because we don't want to actually return an `arena_matrix` while some functions have returns that are difficult to scrub out (like multiply for example which can have a return with a bunch of different sizes.

@t4c1 is an `arena_matrix` implicitly convertible to an Eigen::Matrix? If so then the above is not a concern

bbbales2:
Imo the return type thing seems like a problem. I'm happy to wrap stuff in `arena_t` as needed.

t4c1:
I think a matrix of vars has no inherent property that would make it need to be stored in the arena. So `promote_var_matrix_t` should not return an arena matrix. If you need it to be in the arena just wrap it in `arena_t`. My vote is to close this without merging.

SteveBronder:
2-1, done!

PR #2121, opened 2020-10-02
Add var matrix elt_multiply

SteveBronder:
## Summary

Adds `elt_multiply()`, this one looks a bit complicated but I figured that a little bit of tuning was a good idea since we use this function a lot.

Performance looks reasonable. Here's the graph for this PR vs develop

image

For `var<matrix>` the results are very cool and good

image

## Tests

Added `var<mat>` in `mix/fun/elt_multiply_test`

## Release notes

Add var and var matrix specializtion elementwise product.

## Checklist

- [x] Math issue #2101


stan/math/rev/core/var.hpp
```diff
@@ -408,6 +416,15 @@ class var_value {
...
 
+  /**
+   * View element of eigen matrices
+   * @param i Row to access
+   * @param j Column to access
+   */
+  inline auto coeffRef(Eigen::Index i, Eigen::Index j) const {
+    return this->coeff(i, j);
+  }
+
```
SteveBronder, line 427: I only added these because they are needed by the `matrix<var>` version in elt_multiply

stan/math/rev/fun/elt_multiply.hpp
```diff
@@ -0,0 +1,74 @@
...
+      using var_m1 = arena_t<promote_scalar_t<var, Mat1>>;
+      using var_m2 = arena_t<promote_scalar_t<var, Mat2>>;
+      if (is_var_matrix<Mat1>::value || is_var_matrix<Mat2>::value) {
+        forward_as<var_m1>(arena_m1).adj().array() += value_of(arena_m2).array() * ret.adj().array();
+        forward_as<var_m2>(arena_m2).adj().array() += value_of(arena_m1).array() * ret.adj().array();
+      } else {
+        for (Eigen::Index i = 0; i < arena_m2.size(); ++i) {
+          forward_as<var_m1>(arena_m1).coeffRef(i).adj() += forward_as<var_m2>(arena_m2).coeffRef(i).val() * ret.coeff(i).adj();
+          forward_as<var_m2>(arena_m2).coeffRef(i).adj() += forward_as<var_m1>(arena_m1).coeffRef(i).val() * ret.coeff(i).adj();
+        }
```
SteveBronder, line 47: for access over the array of struct I found that just doing a for loop here was the most efficient since we use both sides of the struct

t4c1: It would be really nice to have something like multi_result_kernel in kernel generator for Eigen expressions ... But I guess it would not be simple to implement.

stan/math/rev/fun/elt_multiply.hpp
```diff
@@ -0,0 +1,83 @@
...
+      using var_m2 = arena_t<promote_scalar_t<var, Mat2>>;
+      if (is_var_matrix<Mat1>::value || is_var_matrix<Mat2>::value) {
+        forward_as<var_m1>(arena_m1).adj().array()
+            += value_of(arena_m2).array() * ret.adj().array();
+        forward_as<var_m2>(arena_m2).adj().array()
+            += value_of(arena_m1).array() * ret.adj().array();
+      } else {
+        for (Eigen::Index i = 0; i < arena_m2.size(); ++i) {
+          forward_as<var_m1>(arena_m1).coeffRef(i).adj()
+              += forward_as<var_m2>(arena_m2).coeffRef(i).val()
```
bbbales2, line 49: This ended up being faster than saving the value_of or doing the equivalent eigen expression?

(Equivalent eigen expression being something like):
```
arena_m1.adj() += arena_m2.val().array * ret.adj().array();
```

bbbales2 (approved):
Good

stan/math/rev/fun/elt_multiply.hpp
```diff
@@ -0,0 +1,83 @@
...
+          forward_as<var_m2>(arena_m2).coeffRef(i).adj()
+              += forward_as<var_m1>(arena_m1).coeffRef(i).val()
+                 * ret.coeff(i).adj();
+        }
+      }
+    });
+    return ret_type(ret);
+  } else if (!is_constant<Mat1>::value) {
+    arena_t<Mat1> arena_m1 = m1_ref;
+    arena_t<promote_scalar_t<double, Mat2>> arena_m2 = value_of(m2_ref);
```
bbbales2, line 60:
```
auto arena_m2 = to_arena(value_of(m2_ref));
```

stan/math/rev/core/var.hpp
bbbales2: Is it more efficient to do this than `variable.adj().coeffRef(...)`? I've just been doing that since I think it's the right thing for `var<mat>`.

stan/math/rev/fun/elt_multiply.hpp
SteveBronder: Yeah so the logic here is that for `var<matrix>` we do want to do the eigen stuff above since we have a it's essentially a struct of arrays, but for `matrix<var>` we have an array of structs. So when we compute  `arena_m1`'s adjoint using `arena_m2`'s val it has the value and adjoint of each sitting around already so it's better to do a loop. to be able to reuse the other side of the struct

stan/math/rev/fun/elt_multiply.hpp
SteveBronder: Since we are using `var<matrix_cl<double>>` for these wouldn't we want a version that's like the Eigen one above? Since we have a struct of arrays.

stan/math/rev/fun/elt_multiply.hpp
SteveBronder: oooh this is nice I'm gonna make this change then we can merge this

stan/math/rev/core/var.hpp
SteveBronder: It's less efficient, it's just that this method needs to be defined for the `var<matrix>` version that does the loop. I wouldn't mind waiting until #2064 goes in for this because there we actually do all of these much more nicely

stan/math/rev/fun/elt_multiply.hpp
bbbales2: It makes sense. I don't think I've been benchmarking stuff large enough to see this maybe.

bbbales2: I went to update the dot_self implementation in #2019 and it ended up for me the elementwise thing was faster for small stuff and the eigen stuff was faster for big stuff.

I don't know if when it gets large enough to drop out of cache it will go back to preferring the elementwise implementation.

I put all three implementations and benchmarks from my laptop in the pull. Could you take a look at this? I guess basically my question is what implementation to roll with? Or is there a better way to do this? (this file: https://github.com/stan-dev/math/blob/4005e1158cf38a1c72c197b0b53eec9131f2c49a/stan/math/rev/fun/dot_self.hpp)

bbbales2: Hmm I guess the more realistic thing is that between the forward pass and the backwards pass stuff gets pushed out of cache cause other calculations happen even if these calculations would fit in cache by themselves. I will add that to the benchmarks and get back.

SteveBronder: I'm actually worried that at large `N` saving the values could actually be bad a bad pattern. Like if you have a `8000x8000` matrix or something pulling out the values is actually going to make like 8GB of temporary data

bbbales2: Alright I added some code to allocate and write 8 megabytes of memory between the forward and reverse passes of the benchmarks.

Cause that made these benchmarks take a lot more time, I turned the number of repetitions waaaaay down, so these numbers are super noisy.

With this, there isn't an advantage to saving the copy.

I'm not super concerned about the large N here, I don't think there are a lot of models out there where we're hitting the memory limitations on the computer (aside from compile). I guess this also applies to cache though -- if we save a ton in memory we're more likely to jump out of cache, but also that I'm not sold on optimizing for.

Original:
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
toss_me                     11147017 ns     11094590 ns            1
dot_self/2/manual_time           489 ns       860215 ns         2999
dot_self/4/manual_time           576 ns       864500 ns         2742
dot_self/8/manual_time           544 ns       850848 ns         2698
dot_self/16/manual_time          696 ns       905070 ns         2448
dot_self/32/manual_time          686 ns       859259 ns         1998
dot_self/64/manual_time          876 ns       874395 ns         1363
dot_self/128/manual_time        1291 ns       936241 ns         1053
dot_self/256/manual_time        1784 ns       848122 ns          723
dot_self/512/manual_time        3183 ns       896457 ns          405
dot_self/1024/manual_time       4796 ns       879361 ns          260
```

New elementwise:
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
toss_me                      9904426 ns      9901508 ns            1
dot_self/2/manual_time           588 ns       853606 ns         2588
dot_self/4/manual_time           684 ns       848227 ns         2225
dot_self/8/manual_time           710 ns       874226 ns         1900
dot_self/16/manual_time          729 ns       856006 ns         1399
dot_self/32/manual_time          728 ns       851753 ns         1860
dot_self/64/manual_time          886 ns       870366 ns         1190
dot_self/128/manual_time        1054 ns       866244 ns         1361
dot_self/256/manual_time        1656 ns       859020 ns          879
dot_self/512/manual_time        3305 ns       893224 ns          407
dot_self/1024/manual_time       5228 ns       880822 ns          226
```

Eigen:
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
toss_me                     11149331 ns     11147214 ns            1
dot_self/2/manual_time           788 ns       937184 ns         1879
dot_self/4/manual_time           738 ns       851540 ns         1776
dot_self/8/manual_time           756 ns       894369 ns         1777
dot_self/16/manual_time          898 ns       895735 ns         1589
dot_self/32/manual_time          952 ns       889746 ns         1370
dot_self/64/manual_time         1152 ns       863095 ns         1063
dot_self/128/manual_time        1446 ns       897282 ns          780
dot_self/256/manual_time        1944 ns       852350 ns          555
dot_self/512/manual_time        3137 ns       870621 ns          408
dot_self/1024/manual_time       5151 ns       894738 ns          237
```

bbbales2: I switched it to the elementwise implementation. It's basically the same thing that we were doing previously and has the efficient memory use and the Eigen vectorized stuff isn't really helping when the stuff isn't just sitting in cache.

bbbales2 (approved):
Good!

PR #2144, opened 2020-10-09
added custom_vari

t4c1:
## Summary

I have been informed that `reverse_pass_callback` is slower than custom vari class when used on scalar operations. Profiling showed that this is due to constructing two varis - one for data and one for the callback. 

To work around this issue I implemented `custom_vari`, that works like `reverse_pass_callback`, except it also holds the value and adjoint. So everything can use the same vari. It also supports Eigen types, so I think we can use it everywhere instead of `reverse_pass_callback`.

To demonstrate how it is used I reimplemented `operator+` for scalar `var`s. Benchmarking showed it is in fact even a bit faster due to simplified nan check.

## Tests

`custom_vari` is tested.

## Side Effects
None

## Release notes

Added `custom_vari` class.

## Checklist


SteveBronder:
I'm seeing that it's still a hair slower but well within like a standard deviation so I think that's fine. Though I wonder whether it's better to just leave the current addition vari then once the rest of the library uses reverse_pass_callback I think the two pushes will be offset by not having the vtable in vari

image

t4c1:
> I'm seeing that it's still a hair slower

Interesting. It is ~5% faster on my computer. Can you also try benchmarking this branch with previous nan handling (using `is_any_nan()` on arguments? Because I think this is more likely to cause the performance difference and I can easily revert that part of the PR if it is slower on some computers.

> I think the two pushes will be offset by not having the vtable in vari

Nope, we can't actually do that. `reverse_pass_callback_vari` is templated with the type of the functor, so different use cases actually use different vari types and `chain()` still needs to be virtual.

> I think maybe reverse_pass_callback_vari is a better name.

I was also thinking about `diy_vari`. I prefere `custom_vari`, but if multiple people agree on something else, I am not against changing it.

SteveBronder:
> Interesting. It is ~5% faster on my computer. Can you also try benchmarking this branch with previous nan handling (using is_any_nan() on arguments? Because I think this is more likely to cause the performance difference and I can easily revert that part of the PR if it is slower on some computers.

Sure I'll play around with this tmrw. 

> Nope, we can't actually do that. reverse_pass_callback_vari is templated with the type of the functor, so different use cases actually use different vari types and chain() still needs to be virtual.

Dang!

Something kind of interesting I've been playing with. If we rolled our own vtable we could keep vari's size 16 and pull the tables out into their own vector as well. The plus is that we avoid having to actually hop through a jump table but separating them out does mean we have to go fetch and store the function pointer. It's also a simplified example, I think we'd have to do some other things to get this to work in stan

https://godbolt.org/z/Y63M64

t4c1:
@SteveBronder Did you have time to check whether the slowdown on your PC is caused by different nan check?

SteveBronder:
Yeah I'll check today

bbbales2:
Just had a scan through this. Looks good to me, but I'll leave reviewing to @SteveBronder. Can it be called `functor_vari` ?

SteveBronder:
Huh dunno what happened but things look good now!

image


what about `inplace_vari` or also I like `functor_vari`?

SteveBronder:
Eh Actually idk how I feel about `functor_vari` but I'm also fine with it. what about just `callback_vari`? Naming things, hard!!

t4c1:
So ... neither of you likes the name `custom_vari`, but you could not agree on any other name? Hmm, I guess I am fine with `callback_vari`.

t4c1:
Oh, I can't. We already have a `callback_vari` for `reverse_pass_callback`.

bbbales2:
`callback_vari` is good. This can just be used for `reverse_pass_callback` right? `reverse_pass_callback` is this with no value.

The reason I don't like `custom_vari` is every vari is custom :P. Is this particularly more or less custom? I dunno.

t4c1:
Well I can not use `callback_vari`, as it is used for the vari for `reverse_pass_callback` (unless I renamed that one). And I don't like `functor_vari` , as it is not really a functor.

bbbales2:
They're kinda the same thing so I think we can just replace the old one with this one:

So right now the implementation for reverse pass is:
```
template <typename F>
inline void reverse_pass_callback(F&& functor) {
  new internal::callback_vari<F>(std::forward<F>(functor));
}
```

But it could be:
```
template <typename F>
inline void reverse_pass_callback(F&& functor) {
  new internal::callback_vari<double, F>(0, std::forward<F>(functor));
}
```

t4c1:
I don't want to implement it that way. That is just a redundant value, which is not used anywhere. We might be able to eventually completly replace `reverse_pass_callback` with this, but until than I would like to be efficient.

bbbales2:
Good point, well then the `reverse_pass_callback` vari isn't used anywhere but `reverse_pass_callback`. So `reverse_pass_callback_vari` makes sense for it.

SteveBronder:
Could we change the name of that one to something like `reverse_pass_callback_vari_impl`? Else I think I'm fine with functor_vari

t4c1:
@bbbales2 @SteveBronder Can someone review this?

stan/math/rev/core/operator_addition.hpp
```diff
@@ -79,7 +48,16 @@ class add_vd_vari final : public op_vd_vari {
...
 inline var operator+(const var& a, const var& b) {
-  return {new internal::add_vv_vari(a.vi_, b.vi_)};
+  return make_callback_vari(a.vi_->val_ + b.vi_->val_,
+                            [avi = a.vi_, bvi = b.vi_](const auto& vi) mutable {
+                              if (unlikely(std::isnan(vi.val_))) {
+                                avi->adj_ = NOT_A_NUMBER;
+                                bvi->adj_ = NOT_A_NUMBER;
+                              } else {
+                                avi->adj_ += vi.adj_;
+                                bvi->adj_ += vi.adj_;
```
bbbales2, line 58: Would this be faster if we just saved pointers to the adjoints themselves? They can't move and we're not using the input `vals_`.

bbbales2: Oh nevermind. Var is just pointer to vari which is value + adjoint.

bbbales2 (approved):
GOod!

PR #2149, opened 2020-10-14
Backport assign to Stan Math

SteveBronder:
## Summary

This backports (up-ports?) the assign functions from the Stan repo into this repo and refactors the `assign()` function to allow for better slice indexing

The scheme here is that for slicing like `A[2:5, 1:10]` `Vector[1:10]` we want to translate that directly into Eigen's `block()` or `.segment()`  functions. For functions that slice over rows and columns like `A[, {1, 3, 8}]` and `A[1:N, :N]` So that Eigen knows what columns it's slicing by first and then the slice of rows.

## Tests

Tests were added for the new overloads and can be run with

```
./runTests.py ./test/unit/math/prim/fun/assign_test.cpp
./runTests.py ./test/unit/math/prim/ -f index
./runTests.py ./test/unit/math/prim/ -f rvalue
```

@t4c1 would you have time to help me add the `assign()` functions here to the expression testing framework? It's not clear to me how to add all the combinations of index types

## Side Effects

This should have no side effects until we change the compiler to use the `assign` function in the `stan::math` namespace instead of the ones defined in the `stan::model` namespace

## Release notes

Assign statements in Stan have better support for slice indexing

## Checklist

### DIFF
--- a/stan/math/prim/fun.hpp
+++ b/stan/math/prim/fun.hpp
@@ -126,6 +126,7 @@
 #include <stan/math/prim/fun/if_else.hpp>
 #include <stan/math/prim/fun/imag.hpp>
 #include <stan/math/prim/fun/inc_beta.hpp>
+#include <stan/math/prim/fun/index_list.hpp>
 #include <stan/math/prim/fun/initialize.hpp>
 #include <stan/math/prim/fun/int_step.hpp>
 #include <stan/math/prim/fun/inv.hpp>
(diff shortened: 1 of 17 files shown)


t4c1:
Assign is a bit different than other math functions. It does not return, it modifies one of the arguments and it accepts the weird index types. So it would be hard to modify expression tests to handle all that. I suggest writing expression tests for this function by hand (you might still want to use `expression_test_helpers.hpp`).

I would suggest putting all indexing related stuff into its own folder within fun. Also I just noticed assign and lvalue indexing look the same to me. What is the difference between the two?

rok-cesnovar:
I like the idea of moving this to Math. If we stay with the current repo structure, my opinion is that Stan should only be algorithms/services stuff, no utility

I am a bit scared how this will play with rstan and StanHeaders and all the CRAN related update issues. Rstan still has problems that occured when moving `prim/mat/fun/Eigen.hpp`. We might be able to do this once Rstan catches up, before that this could only live in Math as duplicated code from /stan. I dont want to cause more issues to Ben and rstan.

t4c1:
Hmm, if that is the case I would suggest leaving this in stan repo for now. I don't like duplicated code.

SteveBronder:
Cool just opened up a PR for this in the stan repo

PR #2154, opened 2020-10-20
Feature faster parallel reduce_sum Future

wds15:
## Summary

**Warning:** I messed up my git branch here. This branch now contains additional changes which will enable reverse sweeps in parallel. We should first merge the feature-fast-parallel-ad-2 as a first step, since this will already lead to significant speedups.

The current `reduce_sum` parallel AD implementation suffers from performance bottlenecks whenever many shared arguments are being used. The issue is that the input operands must be copied **per work chunk** as scheduled by the Intel TBB. This creates potentially a large burden on the AD system. This PR limits the number of deep copies of the share input operands to the **number of used threads**. The merits of this approach have been benchmarked and [shown here](https://discourse.mc-stan.org/t/older-parallel-ad-tape-ideas/18534/31).

The idea of this PR is to introduce scoped chainable stacks which allow functions to be executed and record their AD tape bits and pieces onto a separate AD tape. Thus, this allows to write the AD tape of a running function into an instance which is not necessarily the AD tape instance of the respective thread.

## Tests

All `reduce_sum` tests still run just fine. Additional components introduced still need tests.

## Side Effects

Makes things faster.

## Release notes

Speedup `reduce_sum` parallelism.

## Checklist


wds15:
@SteveBronder , @bbbales2 & @andrjohns are u good or do you need more doc? Code changes are not huge, but I haven't spend too much time to doc this. Tests of `reduce_sum` still run, but the rest still needs tests. Let me know what you think of the ideas and then we settle on how we move forward.

As I said, this PR is really just a first step towards parallel forward & reverse AD sweeps. This step here already gives a good speedup by avoiding deep copies of shared operands. The additional AD tapes created by thread will need to be packed into a reverse pass vari thing, which itself runs stuff in parallel.

stan/math/rev/core/scoped_chainablestack.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace stan {
+namespace math {
+
+class ScopedChainableStack {
+  typedef ChainableStack::AutodiffStackStorage chainablestack_t;
+  chainablestack_t local_stack_;
+
+  using stack_queue_t = std::vector<chainablestack_t*>;
+  stack_queue_t stack_queue_;
```
SteveBronder, line 18: [optional] if we only use these here I think not using an alias is better

```suggestion
  ChainableStack::AutodiffStackStorage local_stack_;

 std::vector<ChainableStack::AutodiffStackStorage*> stack_queue_;
```

stan/math/rev/core/scoped_chainablestack.hpp
```diff
@@ -0,0 +1,50 @@
...
+  ScopedChainableStack() {}
+
+  // execute in the current thread a nullary function and write the AD
+  // tape to local_stack_ of this instance
+  template <typename F>
+  F execute(F f) {
+    activate_scope active_scope(*this);
+    f();
+    return f;
+  }
```
SteveBronder, line 45: Do we need to return F? Or could we just do

```suggestion
  auto execute(F&& f) {
    activate_scope active_scope(*this);
    return std::forward<F>(f)();
  }
```

Because then in the constructor for the reducer we can call this in the initialization list (I'll comment on that below)

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -27,6 +31,25 @@ template <typename ReduceFunction, typename ReturnType, typename Vec,
...
+
+    template <typename... ArgsT>
+    explicit scoped_args_tuple(ArgsT&&... args_tuple)
+        : stack_(), args_tuple_holder_(nullptr) {
+      stack_.execute([&] {
+        args_tuple_holder_ = std::unique_ptr<args_tuple_t>(
+            new std::tuple<decltype(deep_copy_vars(args_tuple))...>(
+                deep_copy_vars(args_tuple)...));
+      });
+    }
```
SteveBronder, line 48: Changing `execute` to the above we can just do
```suggestion
    template <typename... ArgsT>
    explicit scoped_args_tuple(ArgsT&&... args_tuple)
        : stack_(), args_tuple_holder_(stack_.execute([&] {
        args_tuple_holder_ = std::unique_ptr<args_tuple_t>(
            new std::tuple<decltype(deep_copy_vars(args_tuple))...>(
                deep_copy_vars(args_tuple)...));
      })) {}
```

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -106,20 +133,25 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
       // Create nested autodiff copies of all shared arguments that do not point
       //   back to main autodiff stack
+      /*
       auto args_tuple_local_copy = apply(
           [&](auto&&... args) {
             return std::tuple<decltype(deep_copy_vars(args))...>(
                 deep_copy_vars(args)...);
           },
           args_tuple_);
+      */
```
SteveBronder, line 143: delete

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -137,7 +169,11 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
             accumulate_adjoints(args_adjoints_.data(),
                                 std::forward<decltype(args)>(args)...);
           },
-          std::move(args_tuple_local_copy));
+          args_tuple_local);
+      // std::move(args_tuple_local_copy));
```
SteveBronder, line 173: delete all commented code

stan/math/rev/core/scoped_chainablestack.hpp
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    ~activate_scope() {
+      ChainableStack::instance_ = scoped_stack_.stack_queue_.back();
+      scoped_stack_.stack_queue_.pop_back();
+    }
+  };
+
+ public:
+  ScopedChainableStack() {}
```
SteveBronder, line 36:
```suggestion
  ScopedChainableStack() = default;
```

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -27,6 +31,25 @@ template <typename ReduceFunction, typename ReturnType, typename Vec,
...
+        : stack_(), args_tuple_holder_(nullptr) {
+      stack_.execute([&] {
+        args_tuple_holder_ = std::unique_ptr<args_tuple_t>(
+            new std::tuple<decltype(deep_copy_vars(args_tuple))...>(
+                deep_copy_vars(args_tuple)...));
+      });
+    }
+  };
+
+  using local_args_tuple_t = tbb::enumerable_thread_specific<scoped_args_tuple>;
```
SteveBronder, line 51: [optional]

Looking at the docs for this should this use the `arena_allocator` in core?

https://spec.oneapi.com/versions/latest/elements/oneTBB/source/thread_local_storage/enumerable_thread_specific_cls.html

SteveBronder (changes requested):
Most comments here are optional. I think most of it looks good but I'd like docs for `ScaledChainableStack`

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -106,20 +133,25 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
       var sub_sum_v = apply(
           [&](auto&&... args) {
             return ReduceFunction()(local_sub_slice, r.begin(), r.end() - 1,
                                     msgs_, args...);
           },
-          args_tuple_local_copy);
+          args_tuple_local);
 
       // Compute Jacobian
       sub_sum_v.grad();
```
SteveBronder, line 157: [side Q]

When we call grad here on a thread local `ChainableStack`, is it going to iterate through all of the vari that were created in that thread instance? Would it be faster if we had a `grad` function like

```cpp
template <typename Vari>
void grad(size_t end_pos, size_t begin_pos) {
  ChainableStack::instance_->var_stack_[end_pos,]->init_dependent();
  for (i = end_pos; i > begin_pos; --i) {
     ChainableStack::instance_->var_stack_[i]->chain();
  }
}
```

Then when we start the nested we save the starting place, then get the last place of the var_stack here and call

```cpp
grad(ChainableStack::instance_->var_stack.size() - 1, begin_nested.begin_pos());
```

wds15: We only iterate here through the nested stuff, not more...but these must be traversed. So I think we can’t save her anything. Maybe I misunderstand your thoughts though.

SteveBronder: Oh whups yeah my bad I forgot `grad()` checks for whether it's in a nested already

stan/math/rev/functor/reduce_sum.hpp
wds15: sure. gone.

stan/math/rev/functor/reduce_sum.hpp
wds15: I don't think that is needed. The AD tape itself will still use the stan-math arena... but the small struct holding stack and tuple together doesn't really need the arena (maybe there is a benefit I am not aware of).

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -220,8 +243,12 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
       partials[i] = 0.0;
     }
 
+    local_args_tuple_t local_args([&] { return scoped_args_tuple(args...); });
+    // @Steve: This should work, but fails to compile???
```
wds15, line 247: Could you have a look here, @SteveBronder ? I don't quite get why i can't use the argument forward thing from the TBB which fails to compile for me.

wds15:
@SteveBronder some first doc is in place. I hope that example makes it clear enough what is happening.

How do we proceed? What is still missing are tests for the scoped thing. Other than that it's good to go for my taste. @bbbales2 ?

bbbales2:
Waiting until after the release to review stuff in Math otherwise I'll get distracted by new stuff. Will do it then (in a couple days).

SteveBronder:
So this is the version that also does the reverse mode in parallel? If this is going into next release and it's only 200 lines of code then I think we can include them both on one PR.

There's a lot of things I'm sort of unfamiliar with though. Can you talk a bit about the scoped pattern here? 

Very exciting though!! Have you had time to run any benchmarks with this version? Also will this also work with parallel_map?

stan/math/rev/core/scoped_chainablestack.hpp
```diff
@@ -0,0 +1,70 @@
...
+   * tape to local_stack_ of this instance. The function may return
+   * any type.
+   *
+   * @tparam F nullary functor to evaluate
+   * @param f instance of nullary functor
+   * @return Result of evaluated functor
+   */
+  template <typename F>
+  auto execute(F&& f) {
+    activate_scope active_scope(*this);
```
bbbales2, line 63: To use the `std::vector` of ChainableStacks one scoped stack would need to call execute on itself right? Like:

```
scoped_stack.execute([&] {
  scoped_stack.execute([] {
     ...
   });
 });
```

I think this can just be one ChainableStack pointer.

wds15:
The reverse pass is not yet in parallel, but the ground work is there. Let me continue working on it.

stan/math/rev/core/scoped_chainablestack.hpp
wds15: No. Don’t think so. The entire reduce sum stuff is recorded to scoped stacks which themselves are stored as part of a special vari on the main stack.

bbbales2: Sure, but I don't get why we need to save `std::vector` s of things in ScopedStack. The `ChainableStack::AutodiffStackStorage` stacks themselves are the vectors of things.

So if we were going to use a new `ChainableStack::AutodiffStackStorage`, we need to:

1. Save the old `ChainableStack::instance_`
2. Allocate a new Autodiffstack
3. Assign it to `ChainableStack::instance_`
4. Do our stuff
5. Replace the old `ChainableStack::instance_`

As long as when we use another stack we take those five steps, then it's fine if other things in step 4 allocate stacks and such -- they're responsible for restoring the `ChainableStack::instance_` on return. I don't think we need `std::vector` s of Stacks in the ScopedChainable thing.

As far as reverse mode, if we want to extend the life of our autodiff stacks to the life of the stack, we need to put them in a vari. In that case I think ScopedChainableStack becomes the vari. Because these varis will contain `std::vector` s, they varis need to be `chainable_alloc` varis.

wds15: Uh... sorry, I misread. Ignore my last comment. The need for basically a queue of chainablestack pointers is exactly for the case of nesting calls into one another as you exemplify. I think this should be supported by the design. When doing that, you need a queue of what is nested into one another in order to reverse things. Not sure how you would otherwise do it. I don't say that we should do nesting of things, but I also don't see how to avoid it nor do I think we need to rule it out.

bbbales2: > you need a queue of what is nested into one another in order to reverse things.

For forward pass, yeah, but all the necessary stuff will just be saved on the CPU stack as it executes.

For reverse pass, you follow the same rules as forward pass in the chain call of the ScopedChainableVari (save the old instance, load up the new one, do stuff, restore the old instance). If during `do stuff` you chain more ScopedChainableVaris, then those are responsible for restoring the state of `ChainableStack::instance_` when they are done.

wds15: That's right... but the logic I implemented allows to nest execute calls into one anpther *using the same instance*. What you describe is what happens when we use multiple different instances which nest into one another.

wds15:
I just pushed an update which does the reverse sweep in parallel, but there are obviously still errors. I obviously fail to keep everything properly in memory, since sometimes things work and sometimes it crashed and even other times I get wrong results. I don't quite see what I am doing wrong here. Maybe the "perfect forwarding" is imperfect. Right now I don't see anymore serious issues; so maybe someone with fresh eyes and a fresh brain can approach this (which could be myself the coming days).

It should work, but well... not now.

SteveBronder:
I should have a little free time to look at this less someone else wants to take a crack at it

stan/math/rev/core/scoped_chainablestack.hpp
bbbales2: > calls into one anpther using the same instance

Yeah and so I was trying to figure out where that would happen. This was the example I came up with was something like:

```
ScopedStack stack;
stack.execute([&stack]() {
  stack.execute([] {
    ...
  });
});
```

but that seems weird (you really have to try to reuse the same stack), and you could solve this with multiple stacks doing:
```
ScopedStack stack1;
stack1.execute([]() {
  ScopedStack stack2;
  stack2.execute([]() {
    ...
  });
});
```

bbbales2:
> EDIT: Looking at it once more, we probably need this to_arena thing for the variadic input of reduce_sum... a hand would be great for that.

Good point. Double arguments are stored by reference in the argument tuple right now and those will actually need copied to the stack if we expect them to be around in the reverse pass.

Would have to look more to say more than good point though lol.

> less someone else

The more the merrier!

bbbales2:
But also those arguments shouldn't be needed in the reverse pass... Nevermind. An operation that needed them will have already copied them to the stack.

wds15:
Made some progress: On the child-stacks I need to have fresh `vari` instances of the input operands. As a trick I am injecting into the parallel function instead of `a` as operand `a + 0.0` which is the same value, but puts onto the `vari` stack a proxy which can take the accumulation of the adjoints within a given thread.

That makes the unit test run most of the time, but they still fail rarley in odd ways. The benchmark can already be run again, but that shows that this approach is a lot slower as it is right now... but it runs at least.

wds15:
Looks like I got it working!

However, the performance looks like it is quite a bit worse than the thing which does the nested gradient stuff. Maybe there are still things to improve, but the general pattern as implemented is what I had in mind.

Right now the shared operands are held per thread which limits the parallelism in the backward sweep. I can try and see if we get an improvement by organizing this by work chunk. I have my doubts that this will give us the performance we are looking for, but maybe it does, since this maybe better wrt to memory locality.

bbbales2:
Looks cool! Will dig in tomorrow.

> quite a bit worse than the thing which does the nested gradient stuff

It may always be since we lose the cache locality thing, but this is still a fine place to experiment since it's something we're familiar with.

In the end we may just break the pieces out here to be the parallel reverse mode AD things and leave `reduce_sum` using the stuff you wrote in the other pull.

wds15:
The code right now is actually not anymore needing reduce features from the TBB. I rewrote it using a parallel_for from the TBB, but that also does not speed up things in any way. To me it looks as if we want to keep using in reduce_sum what we have (so with nested AD) with the additional change in the other PR (with the ScopedChainablestack).

I am probably going to discard my local changes with the parallel for given that these are no improvement...unless you want to see that.

... but well, maybe someone finds an obvious performance drain in the code.

I think what we loose by not doing the nested stuff is for one the memory locality and for two we also accumulate in a much less efficient way the partial derivatives. I mean, when we do it with the nested approach, we have a very nice and tightly organized memory layout of these Eigen structures and the accumulation happens in chunks. At least this is my impression.

It could be worthwhile to profile if the accumulation via `accumulate_adjoints` takes some apprrciable amount of time for reduce_sum when working on memory bound problems. In that case one may consider to write a more thightly integrated data-structure where all the vari's have their underlying adjoints saved as big chunk (much like the var<mat> stuff)... but that effort is really only worth it if `accumulate_adjoints` is indeed a bottleneck.

bbbales2:
@wds15 alright I did some digging and experimenting here.

I don't really have too many thoughts on the code I just wanted to get you a response back since it's been a couple days.

My main feeling doing this was that it seems like more threaded stack stuff should have been automatic. Like, to get a copy of all the arguments on a stack for each thread, that was hard because you have to allocate these new stacks and fill them up. Wouldn't it make sense if there was already a default stack for each thread that allocates itself at launch? Then at least you don't have to worry about allocating a base stack.

I noticed in your implementation there are different stacks for computation and operands, so maybe a global stack could fulfill the roll of the operands stack? I also want there to be less ScopedStacks -- there are a lot of them especially in the nested calculation, but maybe it is not easy to keep this number low.

There are some modifications here: https://github.com/stan-dev/math/compare/feature-fast-parallel-ad...feature-fast-parallel-ad-single-stack?expand=1 -- mainly getting rid of the `std::vector` in `ScopedChainableStack` and messing with other things, not something to merge back in here but just so you know what I was poking at.

As a side node I had the address sanitizer (https://clang.llvm.org/docs/AddressSanitizer.html) on while I was doing this and it was super usefully for figuring out where stuff was going wrong (there was a lot of memory allocating that I wasn't cleaning up).

wds15:
> Wouldn't it make sense if there was already a default stack for each thread that allocates itself at launch?

There is a stack in each thread at launch from the stan program - but that is not what we want to use, since we need to keep things in the context of the `reduce_sum` call which may actually not use all threads.

> I noticed in your implementation there are different stacks for computation and operands, so maybe a global stack could fulfill the roll of the operands stack?

No, I don't think that this is a good idea, because (1) we would not be able to clean up the memory as `recover_memory` only acts on the thread where its called and not others and (2) any thread can participate in executing different chains conceptually => so one should not associate resources with any thread, rather with a task running.

> I also want there to be less ScopedStacks -- there are a lot of them especially in the nested calculation, but maybe it is not easy to keep this number low.

There are only 2 stacks per thread. One for the operands and one for doing the work. We do need two here in order to get the synchronization right during the reverse parallel phase. During the reverse sweep we first call `grad` on the stack where the work has been done. This will result in an accumulation of the adjoints in the thread specific operands input stack. Then we call `grad` serially on the input operands stack. This is absolutely required form my view to do it like this. Otherwise you cannot guarantee correct synchronization of the accumulation happening into the input operands.

wds15:
@andrjohns I think you were interested in bits of this PR if I recall right. In any case, can we close this PR for now? I recall that this entire idea here is not any more applicable as we solved things differently.

wds15:
Looks like we can close the PR. The branch stays around, of course.

Issue #2152, opened 2020-10-20
Add OpenCL distributions


t4c1:
## Description

Add OpenCL implementations of distribution functions. Most of these can be implemented using kernel generator. Some might need kernel generator to support some new element-wise functions.

Since they are used most I will start with `lpdf`s and `lpmf`s, as those are most used. `cdf`s, `lcdf`s and `lccdf`s can wait.

## Example
```
matrix_cl<int> y;
matrix_cl<double> theta;
var lp  = bernoulli_lpmf(y, theta);
```

## Expected Output
Same as Eigen implementations, but faster.

#### Current Version:
v3.3.0

wds15:
Will these be faster on the CPU as well?

t4c1:
We will see. I expect most of them to be a tiny bit slower, but some to be faster.

t4c1:
Now all simple to implement `lpmf` and `lpdf` functions are in. Remaining distribution functions will have to wait until I also add everything from `prim/fun` that is not too much work.

t4c1:
Now we also have PRs open or merged for all `cdf`, `lccdf` and `lcdf` functions that were not too complicated to implement. I skipped some distributions that depend on functions that require 1000 lines of code or more (at least boost implementations are huge). If somebody volunteers to implement any of these functions in OpenCL or finds a significantly shorter implementation, I can implement the distributions depending on it. 

These functions and distribution are:
 - inc_beta: beta, beta_binomial, beta_proportion, binomial, neg_binomial, neg_binomial_2, student_t
 - gamma_p / gamma_q:  chi_square, gamma, inv_chi_square, inv_gamma, poisson, scaled_inv_chi_square
 - owens_t: skew_normal

rok-cesnovar:
Cool! Thanks!!

I would say we close this issue once all the open PRs are merged and make a new one for those remaining (l)(c)cdfs (or edit the top-level comment to reflect this - though I think editing history is not ideal).

PR #2155, opened 2020-10-21
Opencl Bernoulli lpmf

t4c1:
## Summary

Implements `bernoulli_lpmf` using OpenCL, enabling it to use GPUs. 

This PR includes lots of changed files, as I had some problems with circular includes, fixing which involved moving `is_kernel_expression.hpp` to `prim/meta`.

Also includes a bugfix for `check_cl`, which caused the generated kernel not to compile if `check_cl` is used in a kernel that requires specific local size (any kernel that uses a colwise reduction).

## Tests

Tests check that OpenCL overload produces same results as Eigen one.

## Side Effects

None.

## Release notes

Implemented `bernoulli_lpmf` using OpenCL, enabling it to use GPUs. Fixed a bug that caused a generated kernel not to compile if it used both `check_cl` and any colwise reduction.

## Checklist

- [ ] Math issue #2152 


t4c1:
Benchmarks! 

First is the usual comparison of OpenCL version running on GPU with Eigen implementation. However @wds15 asked me how these compare when run on CPU, so I included these too. I tried to make it fair to Eigen implementation by adding another version that also runs in parallel on CPU using reduce_sum. However this turned out to be slow. @wds15 this is the first time I am using `reduce_sum`. Can you check that I did not do something stupid? The benchmark code is here: https://gist.github.com/t4c1/81975eec4f90707e8a727884f901249b

The results are a bit noisy as I ran the benchmark only once, but it is clear that for large data sizes this is the fastest implementation even when running on a CPU.

bernoulli_lpmf_cl

t4c1:
@wds15 Did you miss this mention? Or are you too busy with other things?

t4c1:
I fixed a bug that caused the OpenCL implementation to be slow. The graph above is updated with new values.

wds15:
Missed it... will take a look . But Bernoulli is way too memory bound for reduce sum...so it’s expected to behave badly. How many cpus did you use for the reduce sum run?

t4c1:
I have one 4-core CPU (i5 2500).

t4c1:
Also Bernoulli needs to calculate log and log1m of theta, so I am not sure if it is memory bound. GPU speedup is also larger than what I would expect from memory bound operation.

wds15:
That’s still too cheap to offset the overhead from reduce sum. A poisson with a log gamma is better. So you used 4 cores with reduce sum?

t4c1:
I think so. Yeah. I just checked CPU usage goes to 100%, when running reduce sum. I also noticed it needs huge amounts of RAM (16GB for the largest size).

wds15:
So you have N data items and you also have a parameter vector of the same size as N if I see that right. Then you should not slice the data y, but rather the parameter vector theta. Otherwise it's very inefficient.

t4c1:
As far as I see it, I need to slice both. So I let `reduce_sum` slice the `std::vector<int>` and I slice the `Eigen::VectorXd`. Or am I misunderstanding something?

rok-cesnovar:
The difference is that entire non-sliced arguments are copied to each thread, while they only get the slices from the sliced arg. And copying entire std::vector<int> is better than entire  vars vectors.

wds15:
You need to slice both, yes.. but if you want things to run efficient, then you let reduce_sum slice the parameters and slice the data yourself in the function.

Please also make sure that you define the environment variable `STAN_NUM_THREADS` when you run this thing to be 4.

t4c1:
Thanks, I will try it that way.

t4c1:
I fixed the reduce_sum benchmark (I was also passing around some things by value. Ouch. That explains the memory usage). Now the results make more sense. OpenCL implementation is still the fastest for large amount of data, but otherwise reduce_sum is faster. 

bernoulli_lpmf_cl

stan/math/opencl/prim/bernoulli_lpmf.hpp
```diff
@@ -0,0 +1,115 @@
...
+ * @tparam T_n_cl type of integer parameters
+ * @tparam T_prob_cl type of chance of success parameters
+ * @param n integer parameter
+ * @param theta chance of success parameter
+ * @return log probability or log sum of probabilities
+ * @throw std::domain_error if theta is not a valid probability
+ * @throw std::invalid_argument if container sizes mismatch.
+ */
+template <
+    bool propto, typename T_n_cl, typename T_prob_cl,
```
rok-cesnovar, line 29: propto is not in doxy. But its also not listed in CPU distributions (I checked like 5 and then stopped). So I guess this is fine.

rok-cesnovar (approved):
LGTM.

PR #2156, opened 2020-10-21
fixed arena_matrix_cl

t4c1:
## Summary

Fixes a crash that happens when calling `recover_memory` after using `arena_matrix_cl` (used internally in all rev OpenCL functions).

## Tests
Added `recover_memory()` to `arena_matrix` and `to_arena` tests.

## Side Effects
This is not the most elegant solution, but it is what I could came up with before the release. We should probably rethink the design of `arena_matrix_cl`.

## Checklist

- [ ] Math issue #2161

### DIFF
--- a/stan/math/opencl/rev/arena_matrix_cl.hpp
+++ b/stan/math/opencl/rev/arena_matrix_cl.hpp
@@ -2,7 +2,7 @@
 #define STAN_MATH_OPENCL_REV_ARENA_MATRIX_CL_HPP
 #ifdef STAN_OPENCL
 
-#include <stan/math/rev/core/chainable_alloc.hpp>
+#include <stan/math/rev/core/needs_destructor.hpp>
 #include <stan/math/opencl/kernel_generator/is_kernel_expression.hpp>
 #include <stan/math/opencl/matrix_cl.hpp>
 #include <utility>
@@ -15,7 +15,7 @@ namespace math {
  * can be used on the AD stack.
  */
 template <typename T>
-class arena_matrix_cl : public chainable_alloc, public matrix_cl<T> {
+class arena_matrix_cl : public needs_destructor, public matrix_cl<T> {
  public:
   using Scalar = typename matrix_cl<T>::Scalar;
   using type = typename matrix_cl<T>::type;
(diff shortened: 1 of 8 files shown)


SteveBronder:
So reading this the source of the bug is that `chainable_alloc` is made for pointer types that are made with new? So we need to make a smart pointer like class for `matrix_cl`?

t4c1:
Yes.

bbbales2:
> This is not the most elegant solution, but it is what I could came up with before the release. We should probably rethink the design of arena_matrix_cl.

Ugh this one makes me nervous. I see `recover_memory` changes but not `recover_memory_nested`, for instance.

> is made for pointer types that are made with new?

Can we just allocate this type on the regular stack so delete works or something?

t4c1:
> Ugh this one makes me nervous. I see recover_memory changes but not recover_memory_nested, for instance.

Ugh, I need to add that.

> Can we just allocate this type on the regular stack so delete works or something?

That is intended to be the OpenCL counterpart of the `arena_matrix_cl` - so it is returned by the `arena_type_t` and `to_arena`.  If we allocate this with `new`, we need to work with pointers everywhere this is used and we can not use `arena_type_t` and `to_arena` for OpenCL types anymore.

bbbales2:
> we need to work with pointers everywhere

Cool lemme catch up. I'm still confused. I think it's fine if we delay the release to get a clean fix here since this is a bug.

We have `matrix_cl` which has `std::vector` in it so it has to be allocated in the heap and needs to have its destructor called.

We have `arena_matrix_cl` which inherits from `matrix_cl` that is constructed through `to_arena` calls.

So on the forward pass:
```
matrix_cl *a = new matrix_cl();
auto b = to_arena(a);

reverse_pass_callback([b] {
   ...
});
```

And then on the reverse pass we need something like:
```
delete_the_original_pointer_a
```

t4c1:
There is no need for matrix_cl to be allocated on heap. It is always used just as regular local variable.

bbbales2:
We moved to the #mathlib on slack (https://discourse.mc-stan.org/t/mc-stan-community-slack/2410)

bbbales2:
SO there's something here, the problem is that `arena_matrix_cl` needed its destructor called in `recover_memory` without being freed, so `chainable_alloc` doesn't work because that frees and there needed to be a fourth stack.

We talked about another solution where we have `arena_matrix_cl` which contains a pointer to an `arena_matrix_cl_vari` which contains the `matrix_cl`. This way `arena_matrix_cl_vari` can be a vari and inherit from `chainable_alloc`. Since it's a vari it can be deleted. THis requires making a bunch of functions in `arena_matrix_cl` that forward through the `arena_matrix_cl_vari` pointer to the `matrix_cl` functions.

Not sure if that's clear but that's what we discussed for the record lol.

SteveBronder:
@serban-nicusor-toptal fyi we are waiting till this is resolved before the release. We are going to shoot for Monday and will keep you posted

t4c1:
This bug was fixed in #2160.

PR #2164, opened 2020-10-26
`var<mat>` versions of quad_form and trace_quad_form (Issue #2101)

bbbales2:
## Summary

`var<mat>` versions for `quad_form` and `trace_quad_form`

Not the cleanest. Comments inline.

## Release notes

`var<mat>` versions of `quad_form` and `trace_quad_form`

## Checklist

- [x] Math issue #2101


stan/math/prim/err/check_square.hpp
```diff
@@ -17,7 +17,7 @@ namespace math {
  * @param y Matrix to test
  * @throw <code>std::invalid_argument</code> if the matrix is not square
  */
-template <typename T_y, require_eigen_t<T_y>* = nullptr>
+template <typename T_y, require_matrix_t<T_y>* = nullptr>
```
bbbales2, line 20: a_few_varmats has this (#2106), so this will get fixed when that merges

stan/math/prim/meta/is_vector.hpp
```diff
@@ -100,6 +100,11 @@ STAN_ADD_REQUIRE_CONTAINER(eigen_col_vector, is_eigen_col_vector,
 template <typename T>
 struct is_col_vector : internal::is_col_vector_impl<T> {};
 
+STAN_ADD_REQUIRE_UNARY(col_vector, is_col_vector,
```
bbbales2, line 103: I needed a `col_vector` that wasn't necessarily an `eigen_col_vector` so I added these. Are there tests I should add for these?

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +93,65 @@ class quad_form_vari : public vari {
...
+
+  using return_t
+    = promote_var_matrix_t<decltype(value_of(B).transpose().eval() *
+				    value_of(A) *
+				    value_of(B).eval()), Mat1, Mat2>;
+
+  auto arena_A = to_arena(A);
+  auto arena_B = to_arena(B);
+
+  check_not_nan("multiply", "A", value_of(arena_A));
```
bbbales2, line 127: Need to go through and make all the `check` functions work with `var<mat>` types. I'll make a list of the ones that need updated and add them to #2101. Here I just pass in the value.

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +93,65 @@ class quad_form_vari : public vari {
...
+
+  auto arena_res = to_arena(value_of(arena_B).transpose() *
+			    value_of(arena_A) *
+			    value_of(arena_B));
+
+  if(symmetric) {
+    arena_res = (arena_res + arena_res.transpose()).eval();
+  }
+  
+  return_t res = arena_res;
```
bbbales2, line 138: `arena_res` is on the arena, so since `return_t` is a `var<mat>` it will just take the pointer `arena_res` has right?

stan/math/rev/fun/quad_form.hpp
```diff
@@ -157,6 +216,49 @@ inline var quad_form(const EigMat& A, const ColVec& B, bool symmetric = false) {
...
+ * @return The quadratic form (a scalar).
+ * @throws std::invalid_argument if A is not square, or if A cannot be
+ * multiplied by B
+ */
+template <typename Mat, typename Vec,
+	  require_matrix_t<Mat>* = nullptr,
+          require_col_vector_t<Vec>* = nullptr,
+          require_any_var_matrix_t<Mat, Vec>* = nullptr>
+inline var quad_form(const Mat& A, const Vec& B, bool symmetric = false) {
+  return internal::quad_form_impl(A, B, symmetric)(0, 0);
```
bbbales2, line 259: This access here creates an extra var and an extra access. I can do more verbose code and avoid this if necessary.

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +93,65 @@ class quad_form_vari : public vari {
...
+  check_square("quad_form", "A", A);
+  check_multiplicable("quad_form", "A", A, "B", B);
+
+  using return_t
+    = promote_var_matrix_t<decltype(value_of(B).transpose().eval() *
+				    value_of(A) *
+				    value_of(B).eval()), Mat1, Mat2>;
+
+  auto arena_A = to_arena(A);
+  auto arena_B = to_arena(B);
```
bbbales2, line 125: If these matrices are either `var<mat>` s or just regular `mat` s then this code is fine. If one of these is a `mat<var>` though then it's not as fine because I access values and adjoints repeatedly. If the vals/adjoints are accessed more than once, it would be more economical to just copy to a `var<mat>`. 

So my question is we should just by default be copying to `var<mat>` in all these functions?

test/unit/math/test_ad.hpp
```diff
@@ -1881,11 +1881,15 @@ template <typename F, typename EigMat>
 void expect_ad_matvar(const F& f, const EigMat& x) {
   ad_tolerances tols;
   expect_ad_matvar_v(tols, f, x);
+
+  stan::math::recover_memory();
```
bbbales2, line 1885: I was getting sanitizer errors without these. Something wasn't freeing itself with trace_quad_prod.

I don't feel great about that because it doesn't seem like `recover_memory` is used anywhere else in this code.

bbbales2 (commented):
Questions and comments

stan/math/rev/fun/quad_form.hpp
t4c1: I think we should prefer passing values into check functions anyway. That way we don't need them to support `var`s. Also checks on value will be faster than ones on `Matrix<var>`.

stan/math/rev/fun/quad_form.hpp
t4c1: How about you do something like:
```
const auto& A_adj = to_ref(arena_A)
```
Also it would be nice if we only copied these matrices to arena if we actually needed them in the reverse pass.

stan/math/rev/fun/quad_form.hpp
t4c1: Yes. `arena_t<var<T>>` is `var<T>`.

stan/math/rev/fun/quad_form.hpp
t4c1: I think this is not necessary. Eventually we will have indexing not create extra vars.

stan/math/rev/fun/quad_form.hpp
bbbales2: Maybe this is an expression thing, again. Like, if we pass in `MatrixXd`, then if we switch to value then we have to do two full reads of the matrix to do a check. If we pass in `Matrix<var, R, C>` it is one read. If we pass in an expression view of the values in a `Matrix<var, R, C>`, then it is still one read. This sound right?

t4c1: Yeah, but that is assuming the functions calling the check does not need to extract values anyway. Most functions do need values.

stan/math/rev/fun/quad_form.hpp
bbbales2: Well I'm not worried about evaluatiang expressions as much as converting things to `var<mat>` land so I can be lazy.

I checked and all the copying here is fine if inputs are a mix of `mat<double>` and `var<mat>`. The `A` inputs and the `B` inputs are required for both reverse passes. It's not so good for `mat<var>`.

stan/math/rev/fun/quad_form.hpp
bbbales2: > Most functions do need values.

That is true.

I guess if we just put a `auto x_val = value_of_rec(x)` inside all the checks then that would be good? If someone passes in values, that is good. If someone passes in `mat<var>`, it works via expressions.

stan/math/rev/fun/quad_form.hpp
t4c1: I see. `to_arena` does not convert to `var<mat>`, so there is steel a need to handle `mat<var>` efficiently.

stan/math/rev/fun/quad_form.hpp
t4c1: Yeah, it would work, but for `mat<var>` it would not be efficient. Also I am trying to say if we only pass in values we don't need to change all checks.

stan/math/rev/fun/quad_form.hpp
bbbales2: I guess the solution is to just add a version of `to_var_value` (https://github.com/stan-dev/math/blob/develop/stan/math/rev/fun/to_var_value.hpp) that just forwards things through if it is given a `var_value`.

bbbales2: Oh it would also have to just pass through prims, and that's probably not what that does. `to_var_value(mat_double)` seems like it should create a `var<mat>` not just pass through a double matrix.

bbbales2: Could do `to_var_value_if`, that's a pretty common sorta pattern around these parts.

t4c1: The conversion between `var<mat>` and `mat<var>` is quite expensive. So this would be quite lazy and inefficient option.

stan/math/rev/fun/quad_form.hpp
bbbales2: Yeah, that makes sense. I don't want to accidentally make them slow.

I think our checks should work with `var<mat>` though. That means for ones that read values just adding a version that takes in a `var<mat>` and then calls itself with the value of the `var<mat>`.

stan/math/rev/fun/quad_form.hpp
bbbales2: Well it's only as expensive as a reading a `mat<var>` once, and so if the function uses the `mat<var>` more than once or twice we're probably good right?

t4c1: No. It is quite a bit more expensive, as it needs to also propagate adjoints back. Even we used a matrix many times manually copying values out and adjoints back in `mat<var>` would be cheaper.

bbbales2: Copying values out and putting adjoints back in all `to_var_value` does, I guess with an extra arena allocate, so it couldn't be that much more expensive.

t4c1: All these operations are memory bound, so we can estimate how fast they are by calculating how much data needs to move between CPU and RAM. Assuming 8-byte pointers and doubles we need the following for each value in the matrix:

To `var_value` copies vars to arena (1 read and 1 write), next it copies values from `mat<var>` to `var<mat>` (3 reads and 1 write) and sets the adjoints to 0 (1 write). In reverse pass it increments the `mat<var>` adjoints (3 reads for existing values, 1 read for adjoint and  2 writes). You also need to write the adjoint to `var<mat>` when using this (1 write). In total 14 8-byte values transfered per element of the matrix.

To manually copy values out of `mat<var>` we need (3 reads and 1 write). To increment the adjoints we need to read them (3 reads) and write new values back (2 writes). In total 9 8-byte values transfered per element of the matrix.

That comparison does not include time for allocating an extra arena matrix by conversion.

t4c1: If you are interested in more accurate values you could also benchmark both versions. It might not make much difference for `quad_form` that uses multiplications, but it will make a difference for simple functions.

bbbales2: I think at a very high level I would expect one extra write and read on forward pass, and one extra write and read on reverse pass, so 4 extra operations. Your difference is 5, so I think we're looking at this in the same ballpark and calling it different.

I agree the difference between an optimized `mat<var>` version and a lazy `to_var_value_if` version here will not matter, but a lazy `to_var_value_if` version would be a few lines of code and an optimized `mat<var>` would be a bunch of extra work (and what I'm doing now will work with `mat<var>` but be very bad compared to both).

> but it will make a difference for simple functions.

I agree with this.

bbbales2: "I think at a very high level" as a precursor to a sentence sounds like the most arrogant thing ever said. It is also very cringe on its own. I just meant "I'm not sure but", but it didn't come out that way lol.

SteveBronder: Kind of interesting, I did a search on github for `quad_form` in stan programs

https://github.com/search?l=&p=5&q=quad_form+language%3AStan&type=Code

It looks like the most used case is having data as the second argument. Mostly just something to note. (aka I'd rather not spend too much time optimizing this part)

> If these matrices are either `var<mat>` s or just regular mat s then this code is fine. If one of these is a `mat<var>` though then it's not as fine because I access values and adjoints repeatedly. If the vals/adjoints are accessed more than once, it would be more economical to just copy to a `var<mat>`.

I agree with Tadej that `mat<var> -> var<mat>` is v expensive. For this func it _may_ be fine since we do a lot of mat muls

I also don't like `to_var_value_if`, if we want a `var_value` from a `mat<var>` we could just do

`var_value<promote_scalar_t<double, Mat1>>`

imo I'm not a huge fan of the `do_something_if` functions with templates. I personally label them as "easy to write hard to read". They're also worse for small N (doing almost nothing is different from doing nothing like you get by pulling out the logic)

I went through and moved all the var/double logic out. 

https://github.com/stan-dev/math/compare/feature-pr2164/quad_form

The other benefit of that is we don't have to use `forward_as`

bbbales2: > I also don't like to_var_value_if, if we want a var_value from a mat<var> we could just do
> var_value<promote_scalar_t<double, Mat1>>

I don't think this constructor exists though, so we have to call a conversion somehow.

> For this func it may be fine since we do a lot of mat muls

I agree the conversion is more expensive than not doing the conversion in ideal cases, but the comparison is between doing the conversion, calling `value_of` like a ton of extra times (and there are like 9-10 of these in the implementations here), or writing special versions that support combos of `mat<var>` and `var<mat>` (and selectively copying out values).

The conversion is like 5 extra read ops and we do 3 every time we do an uncached `.val()` or `.adj()`. So I'd much rather be lazy on this one. And then I want to avoid writing three versions of the function to support the argument combos.

> I went through and moved all the var/double logic out.

Hmm I think what I wrote should be efficient for `var<mat>` `mat<double>`. I prefer the shorter code, but if there is an efficiency problem I can change it.

bbbales2: Also that is an interesting search neat trick

test/unit/math/test_ad.hpp
SteveBronder: I'll take a look at this

bbbales2: My suspicion is that the rest of `expect_ad` has these hidden somewhere, either by using `gradient` or `hessian` functions, or other things (and it's okay to add them). `expect_ad_matvar` shouldn't need to leave anything on the stack after it's done.

stan/math/rev/fun/quad_form.hpp
bbbales2: For the 3 read ops to do a `.val()`, one of those is pointer chasing and I think will be way more expensive than loading up `.val()` three times from a `mat<var>`.

stan/math/prim/meta/is_vector.hpp
```diff
@@ -100,6 +100,9 @@ STAN_ADD_REQUIRE_CONTAINER(eigen_col_vector, is_eigen_col_vector,
 template <typename T>
 struct is_col_vector : internal::is_col_vector_impl<T> {};
 
+STAN_ADD_REQUIRE_UNARY(col_vector, is_col_vector, require_eigens_types);
```
bbbales2, line 103: #2106 also fixes this

stan/math/prim/meta/is_vector.hpp
bbbales2: These are in #2106 now

stan/math/rev/fun/trace_quad_form.hpp
```diff
@@ -161,26 +108,25 @@ inline var trace_quad_form(const Mat1& A, const Mat2& B) {
...
-          require_any_var_matrix_t<Mat1, Mat2>* = nullptr,
-	  require_all_not_eigen_st<is_var, Mat1, Mat2>* = nullptr>
+          require_any_var_matrix_t<Mat1, Mat2>* = nullptr>
 inline var trace_quad_form(const Mat1& A, const Mat2& B) {
   check_square("trace_quad_form", "A", A);
   check_multiplicable("trace_quad_form", "A", A, "B", B);
 
-  arena_t<Mat1> arena_A = A;
-  arena_t<Mat2> arena_B = B;
+  auto arena_A = to_arena(to_var_value_if<!is_constant<Mat1>::value>(A));
```
bbbales2, line 116: I ended up going with `to_var_value_if` as the lazy support for mixed `var<mat>` and `mat<var>`.

The other thing I tried was templating out the functions but that is confusing in its own way. It looks like:

```cpp
template <typename Mat1, typename Mat2,
          require_var_matrix_t<Mat1>* = nullptr,
          require_eigen_st<is_var, Mat2>* = nullptr>
inline var trace_quad_form(const Mat1& A, const Mat2& B) {
  return trace_quad_form(A, to_var_value(B));
}

template <typename Mat1, typename Mat2,
          require_eigen_st<is_var, Mat1>* = nullptr,
          require_var_matrix_t<Mat2>* = nullptr>
inline var trace_quad_form(const Mat1& A, const Mat2& B) {
  return trace_quad_form(to_var_value(A), B);
}

template <typename Mat1, typename Mat2,
          require_all_matrix_t<Mat1, Mat2>* = nullptr,	          require_all_matrix_t<Mat1, Mat2>* = nullptr,
          require_any_var_matrix_t<Mat1, Mat2>* = nullptr>	          require_any_var_matrix_t<Mat1, Mat2>* = nullptr,
	  require_all_not_eigen_st<is_var, Mat1, Mat2>* = nullptr>
inline var trace_quad_form(const Mat1& A, const Mat2& B) {
 ...
}
```

stan/math/rev/fun/trace_quad_form.hpp
```diff
@@ -161,26 +108,25 @@ inline var trace_quad_form(const Mat1& A, const Mat2& B) {
...
+  auto arena_B = to_arena(to_var_value_if<!is_constant<Mat2>::value>(B));
 
   var res
       = (value_of(arena_B).transpose() * value_of(arena_A) * value_of(arena_B))
             .trace();
 
   reverse_pass_callback([arena_A, arena_B, res]() mutable {
     if (!is_constant<Mat1>::value)
-      forward_as<promote_scalar_t<var, Mat1>>(arena_A).adj()
+      forward_as<promote_var_matrix_t<Mat1, Mat1, Mat2>>(arena_A).adj()
```
bbbales2, line 125: I think we should redo `promote_var_matrix_t` in another pull request.

I think it should just be `promote_var_matrix_t<T>` returns the `var_mat` with the same shape as `T`.

I think our current design of `promote_var_matrix_t` would be useful when writing two argument functions that took all combinations of `var<mat>` and `mat<var>`.

It seems now we are writing functions where the signatures all separate:

One function does `(mat<var>, mat<var>)` and returns `mat<var>`, and then another function handles `(var<mat>, mat<var>)`, `(mat<var>, var<mat>)`, `(var<mat>, var<mat>)` and returns a `var<mat>` type.

So it's easy enough from the separation of functions to know what the return types should be, and so `promote_var_matrix_t` can simplify.

bbbales2 (commented):
Comments

bbbales2:
@rok-cesnovar @t4c1 @SteveBronder do we have an opencl computer I can log in to to test this? I wasn't able to get opencl running on my local computer

rok-cesnovar:
Not really, I can give you access to our lab machine if needed, but I am not sure why would quad_form or trace_quad_form cause GLMs to fail. Maybe restart to double check?

bbbales2:
The test is `test/unit/math/opencl/rev/normal_id_glm_lpdf_test`. Not obvious to me either why it would fail, but it could be something sneaky. Mind giving that a run real quick in this branch?

rok-cesnovar:
Sure thing. Will run in about an hour or so.

rok-cesnovar:
That hour was long. 

Runs fine locally.

bbbales2:
@SteveBronder ready for review!

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +94,65 @@ class quad_form_vari : public vari {
...
+      forward_as<promote_var_matrix_t<Mat1, Mat1, Mat2>>(arena_A).adj()
+          += value_of(arena_B) * C_adj_B_t;
+
+    if (!is_constant<Mat2>::value)
+      forward_as<promote_var_matrix_t<Mat2, Mat1, Mat2>>(arena_B).adj()
+          += value_of(arena_A) * C_adj_B_t.transpose()
+             + value_of(arena_A).transpose() * value_of(arena_B) * res.adj();
+  });
+
+  return res;
```
SteveBronder, line 154: [optional/rant] Another thing that's nice about moving out the control flow is that it let's you have the immediete evaluation for `C_adj_B_t` when both are var types and for when just one is a var you can do `C_adj_B_t`  inline for the accumulation.

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +94,65 @@ class quad_form_vari : public vari {
...
 };
+
+/**
+ * Return the quadratic form \f$ B^T A B \f$.
+ *
+ * Symmetry of the resulting matrix is not guaranteed due to numerical
+ * precision.
+ *
+ * @tparam Mat1 type of the first (square) matrix
+ * @tparam Mat2 type of the second matrix
```
SteveBronder, line 105: More informative docs would be good here

stan/math/rev/fun/to_var_value.hpp
```diff
@@ -27,6 +27,35 @@ to_var_value(const T& a) {
...
+template <bool Condition, typename T, std::enable_if_t<!Condition>* = nullptr>
+inline auto to_var_value_if(T&& a) {
+  return std::forward<T>(a);
+}
+
+template <bool Condition, typename T, std::enable_if_t<Condition>* = nullptr>
+inline auto to_var_value_if(const T& a) {
+  return to_var_value(a);
+}
+
```
SteveBronder, line 58: I like having the `to_var_value` for `var<mat>` but imo I think `to_var_value_if()` is going to cause more problems than answers

stan/math/rev/fun/trace_quad_form.hpp
SteveBronder: I'm not following why they have to both be coerced to var matrices?

stan/math/rev/fun/trace_quad_form.hpp
SteveBronder: I think what your seeing here is moreso because of the pattern your using. imo I'd rather just not use this pattern

test/unit/math/test_ad.hpp
SteveBronder: I'm cool with just calling recover memory here. I think it's the safe/correct thing to do

SteveBronder (changes requested):
Couple fixes, most of it really boils down to that I think a lot of the problems you are seeing come from the pattern where we use forward_as and promoting matrix of var to var matrix. I don't see why we need to do that promotion and I think breaking out the control flow logic will make those problems go away

stan/math/rev/fun/trace_quad_form.hpp
bbbales2: The other options are:

1. Do not support mixes of `mat<var>` and `var<mat>`. I do not like that because we will have programs with both types (and so we could end up with mixed arguments)

2. Write three different specializations. I do not like that because we found out optimizing for `mat<var>` is trickier than `var<mat>`. I do not think this is particularly necessary cuz this will be faster than not having autodiff.

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +94,65 @@ class quad_form_vari : public vari {
...
+  check_square("quad_form", "A", A);
+  check_multiplicable("quad_form", "A", A, "B", B);
+
+  using return_t
+      = promote_var_matrix_t<decltype(value_of(B).transpose().eval()
+                                      * value_of(A) * value_of(B).eval()),
+                             Mat1, Mat2>;
+
+  auto arena_A = to_arena(to_var_value_if<!is_constant<Mat1>::value>(A));
+  auto arena_B = to_arena(to_var_value_if<!is_constant<Mat2>::value>(B));
```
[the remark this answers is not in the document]
bbbales2: If it is a `mat<var>`, then it needs converted to a `var<mat>`, otherwise pass through.

The other option from these `to_var_value_if` statements are more function definitions and requires. Yes `to_var_value_if` and friends are confusing, but I do not think this is more confusing than:

```
require_eigen_st<is_var, Mat1>* = nullptr,
require_var_matrix_t<Mat2>* = nullptr
```

which is something like you'd need to add to the functions to specify that one of the things is an Eigen-like Matrix with a var and one of the other things is a var matrix.

The result of the `to_var_value_if`, regardless of what it is, needs to end up in the arena cause we use it in the reverse pass (hence everything is wrapped in a `to_arena`).

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +94,65 @@ class quad_form_vari : public vari {
...
+
+  check_not_nan("multiply", "A", value_of(arena_A));
+  check_not_nan("multiply", "B", value_of(arena_B));
+
+  auto arena_res = to_arena(value_of(arena_B).transpose() * value_of(arena_A)
+                            * value_of(arena_B));
```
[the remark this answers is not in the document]
bbbales2: At this point it is a `var_value` or an `arena_t<Eigen::MatrixXd>`, so it's safe to use `value_of`

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +94,65 @@ class quad_form_vari : public vari {
...
+          require_all_matrix_t<Mat1, Mat2>* = nullptr,
+          require_any_var_matrix_t<Mat1, Mat2>* = nullptr>
+inline auto quad_form_impl(const Mat1& A, const Mat2& B, bool symmetric) {
+  check_square("quad_form", "A", A);
+  check_multiplicable("quad_form", "A", A, "B", B);
+
+  using return_t
+      = promote_var_matrix_t<decltype(value_of(B).transpose().eval()
+                                      * value_of(A) * value_of(B).eval()),
+                             Mat1, Mat2>;
```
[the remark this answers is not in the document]
bbbales2: Yeah, I was a little nervous about compile time sized matrices, which is why I kept the template.

In that case I guess I could have just done `var_value<decltype(value_of(B).transpose().eval() * value_of(A) * value_of(B).eval())>`.

stan/math/rev/fun/quad_form.hpp
bbbales2: It is hard to be more specific. Both arguments are matrix types in the sense that they satisfy `require_matrix_t`, and one of them needs to be a reverse mode autodiff type.

We could say inherits from `EigenBase` or is a `var_value<T>` with `T` such that it inherits from `EigenBase` but that is sorta just redefining `require_matrix_t` in the docs.

Suggestions?

stan/math/rev/fun/quad_form.hpp
bbbales2: Yeah I see what you're saying.

Are we sure there's no temporary created in an expression like `A * B * C` though? I assume that is broken up into `D = B * C` and `A * D` if `A`, `B`, and `C` are all Eigen MatrixXd types.

stan/math/rev/fun/trace_quad_form.hpp
bbbales2: I think I could just do `var_value<Mat1>` since you mentioned it elsewhere.

bbbales2: Oh wait that doesn't work cuz might end up with a `var_value<var_value<...>>`.

Anyway, you're right it's a result of the pattern so it's kinda downstream of the rest of the comments.

stan/math/rev/fun/quad_form.hpp
SteveBronder: > We could say inherits from EigenBase or is a var_value<T> with T such that it inherits from EigenBase but that is sorta just redefining require_matrix_t in the docs.

I usually write that out. If someone is one the website the requires don't show up

stan/math/rev/fun/quad_form.hpp
SteveBronder: I'm not sure, the godbolt for that is getting at the level that I can read asm. Though it looks like it makes at least one alloc then can reuse that allocate with a memset under certain conditions. The main thing is that

```cpp
A += B * C * D;
```

Needs to make at least one copy for (`B * C * D`) because of aliasing issues with the left and right hand side. Reading over the asm I think it does one alloc here and reuses the memory for the other multiply. And idt we can use `.noalias()` for eigen matrices of vars since `.noalias()` is like the `restrict` keyword and we access the value and adjoint through the same pointer on both sides (aka aliasing)

Though we can use `.noalias()` for var matrices when we do things like

```cpp
A.adj().noalias() += ...
``

Since the values and adjoints of the var matrix are actually different pointers so there is in fact no aliasing. Does the `restrict` thing make sense?

https://godbolt.org/z/h5df65

File: stan/math/rev/fun/quad_form.hpp (hunk 16 above)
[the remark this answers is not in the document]
Comment (SteveBronder) [reply stan/math/rev/fun/quad_form.hpp:124]: > In that case I guess I could have just done `var_value<decltype(value_of(B).transpose().eval() * value_of(A) * value_of(B).eval())>`.

Yes though you want

`var_value<plain_type_t<decltype(value_of(B).transpose().eval() * value_of(A) * value_of(B).eval())>>` because `var_value<>` can accept expressions as the template (For slices such as `Block<>`

File: stan/math/rev/fun/quad_form.hpp (hunk 11 above)
Comment (bbbales2) [reply stan/math/rev/fun/quad_form.hpp:154]: > I'm not sure, the godbolt for that is getting at the level that I can read asm.

Alright I probably shouldn't have asked that question. Imo we're spending too long on this function.

We could do this efficiently like:
```cpp
auto C_adj_B_t = to_ref_if<!is_constant<Mat1>::value && !is_constant<Mat2>::value>
      (res.adj() * value_of(arena_B).transpose());
```

But the bigger question is `to_var_value_if` and the thing here: https://github.com/stan-dev/math/pull/2164#discussion_r515205941 so let's get those worked out

stan/math/rev/fun/quad_form.hpp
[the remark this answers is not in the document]
SteveBronder: :clap: Pull :clap: Out :clap: The :clap: Control :clap: Flow :clap:

All of these problems go away if we just pull the control flow out into the function. There's a lot of good reasons to prefer pulling out control flow

1. Each path from the start to the return is linear. This makes it a lot easier to read and I don't have to go through each line and think, "Well what if X is var and y is not or x is var and y is var and....". In a block of the control flow I'm aware of which is which.

2. Removes the need for widgets like `forward_as<>()` and more type trait layers like `to_*_if<>()`. All the paths do different things and work on different types. Yinz keep saying this leads to code duplication but I disagree! The code is different in each path, it makes the code more _verbose_ which, imo, is very different than duplication.

3. Uses `value_of()` less (which requires making a `holder<>` class for `Matrix<var>`) 

4. Adding more additional control flow is less annoying. Like if we want to do the `.noalias()` optimization for `var<mat>` like 

```cpp
      if (is_var_matrix<Mat1>::value) {
        arena_A.adj().noalias() += arena_B * (res.adj() * arena_B.transpose());
      } else {
        arena_A.adj() += arena_B * (res.adj() * arena_B.transpose());
      }
```

When we bunch the control flow together with the code this turns into

```cpp
      if (!is_constant<Mat1>::value) {
        if (is_var_matrix<Mat1>) {
          forward_as<arena_t<promote_scalar_t<var, Mat1>>>(arena_A).adj().noalias()
              += value_of(arena_B) * C_adj_B_t;
        } else {
          forward_as<arena_t<promote_scalar_t<var, Mat1>>>(arena_A).adj()
              += value_of(arena_B) * C_adj_B_t;
        }
      }
```

That's a mouthful!

I have an example of breaking out the control flow for `quad_form` below. imo, way more clear! Yeah there's more literal lines of code, but like that's fine to me because I think it's simpler and easier to read.

https://github.com/stan-dev/math/compare/feature/quad_form...feature/quad_form-pullout-logic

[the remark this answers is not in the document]
SteveBronder: If you have like 3 inputs or something then yeah I think it's a good idea to use the `to_*_if()` stuff cuz otherwise you end up with like a bunch of paths which feels silly, but for binary functions I think pulling out the control flow is much cleaner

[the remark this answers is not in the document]
SteveBronder: And also the one downside is that for some reason the version the pulls out the flow has to use `val_op()` otherwise we run into `const` correctness errors due to something internal in Eigen

stan/math/rev/fun/trace_quad_form.hpp
SteveBronder: Or just don't move the `matrix<var>` to a `var<matrix>`. I think Tadej and I both prefer just leaving them instead of doing the conversion. In another comment I have branch with one `quad_form` that handles all the cases of vars with one function

https://github.com/stan-dev/math/compare/feature/quad_form...feature/quad_form-pullout-logic

SteveBronder: I'm going to sit down one day and figure out this whole `val()` and `val_op()` thing. Something something const correctness

andrjohns: The furthest I got was that it's something to do with Eigen's ```sizeof``` logic, there's a little more discussion in [this stackoverflow thread](https://stackoverflow.com/questions/56635122/using-coeffref-with-const-cwiseunaryview-failing-when-multiplying-two-cwiseuna#comment99991183_56661442). The error doesn't occur with ```float```, only with ```double```. I found a modification that fixed the compilation, but hard to know whether that would affect other behaviour. Tricky.

stan/math/rev/fun/quad_form.hpp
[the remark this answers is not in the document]
bbbales2: > All of these problems go away if we just pull the control flow
> "Well what if X is var and y is not or x is var and y is var and...."
> Yinz keep saying this leads to code duplication but I disagree!

Sure. I disagree. I don't think any of this code is easy, and I have to read it very carefully line by line anyway.

I can pull out the control flow though (though I don't think it gets rid of the `to_var_value` problem).

> Uses value_of() less (which requires making a holder<> class for Matrix<var>)

I was wondering why you were swapping those all the time. I just liked `value_of` more cause I don't have to worry about the types. If we add a plain type overload could `value_of` just point to `.val()` (or `.val_op()`) instead of using a holder?

> Like if we want to do the .noalias() optimization for var<mat> like

Well if we converted `mat<var>` to `var<mat>` we wouldn't need the extra logic. But if it's necessary sure. `.noalias()` is good optimization.

> If you have like 3 inputs or something

For what it's worth, it is COming Soon (TM) with `trace_gen_quad_form`, but we can do different code there.

stan/math/rev/fun/trace_quad_form.hpp
bbbales2: > In another comment I have branch with one quad_form that handles all the cases of vars with one function

I was trying to avoid touching the pure `mat<var>` implementation since we decided to wait on benchmarks before we did that (which is why we decided not to merge https://github.com/stan-dev/math/pull/2019)

> I think Tadej and I both prefer just leaving them

Yeah I saw the comments and I made a response where I argued the case that the conversion here was better: https://github.com/stan-dev/math/pull/2164#discussion_r513022661

Since I didn't see a response to that, I just went with the conversion to move this along.

I'm not on the same page with stuff like:
```
arena_t<promote_scalar_t<var, Mat1>> arena_A(A);
arena_t<promote_scalar_t<var, Mat2>> arena_B(B);
return_t res(arena_B.val_op().transpose() * arena_A.val_op() * arena_B.val_op());
```

Like up until this point I've been assuming if we read the `.val()` of a `mat<var>` more than once, we should make a copy of it on the arena.

Since this function does a lot of `.val()` calls on `A` and `B`, my understanding is that we need to make arena copies of the values, otherwise we're going to pointer-chase for every `.val()` that happens on a `mat<var>`, and it's the pointer chasing that is slow. I can just ditch the `to_var_value` stuff (it's gonna be faster than no autodiff either way) but I thought I made a pretty good argument back here: https://github.com/stan-dev/math/pull/2164#discussion_r513022661.

stan/math/rev/fun/quad_form.hpp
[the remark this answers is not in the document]
bbbales2: > `var_value<>` can accept expressions

Oh right. Can't let the guard down for a second on this stuff :/

stan/math/rev/fun/trace_quad_form.hpp
bbbales2: This is the big Q btw. Can do the control flow thing sure.

stan/math/rev/fun/quad_form.hpp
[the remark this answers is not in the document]
SteveBronder: > Sure. I disagree. I don't think any of this code is easy, and I have to read it very carefully line by line anyway.

The asm etc that's generated for the two versions can be v different. Like for example with `to_arena_if<>()`, even when the `bool` template is `false` it's not a no-op and still has to do make an Eigen map. For smaller N stuff like that matters

> I can pull out the control flow though (though I don't think it gets rid of the to_var_value problem).

For this I'd use something like the version I put up where it doesn't do the conversion

> I was wondering why you were swapping those all the time. I just liked value_of more cause I don't have to worry about the types. 

Yeah so `value_of()` for Eigen types calls

```cpp
template <typename EigMat, require_eigen_t<EigMat>* = nullptr,
          require_not_st_arithmetic<EigMat>* = nullptr>
inline auto value_of(EigMat&& M) {
  return make_holder(
      [](auto& a) {
        return a.unaryExpr([](const auto& scal) { return value_of(scal); });
      },
      std::forward<EigMat>(M));
}
```

So it's good to call `value_of()` of when we are doing stuff like

```cpp
// A is an input argument
arena_t<promote_scalar_t<double, Mat1>> arena_A(value_of(A));
```

But once we have the `arena_t` we should just call `.val()` etc.

> If we add a plain type overload could value_of just point to .val() (or .val_op()) instead of using a holder?

Actually @t4c1 why is there a reason we don't have a separate version for plain and non-plain Eigen types? I think we could have an overload for when it's an Eigen matrix of vars that are not an expression and just return `.val()`

> Well if we converted mat<var> to var<mat> we wouldn't need the extra logic. But if it's necessary sure. .noalias() is good optimization.

Yeah it would be v cool if we could do that swap and have it be efficient but idt it's worth the cost of swapping stuff around.

> For what it's worth, it is COming Soon (TM) with trace_gen_quad_form, but we can do different code there.

Oh yeah, I'd be a monster if I asked for breaking up the control flow for anything more than a binary function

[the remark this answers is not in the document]
t4c1: > Uses value_of() less (which requires making a holder<> class for Matrix)

> If we add a plain type overload could value_of just point to .val() (or .val_op()) instead of using a holder?

Holder is only needed for rvalue inputs (which would not work without it). For lvalues `make_holder` just calls the given functor with given argument.

> Actually @t4c1 why is there a reason we don't have a separate version for plain and non-plain Eigen types? I think we could have an overload for when it's an Eigen matrix of vars that are not an expression and just return .val()

`.val()` does exactly the same thing as `value_of` called on Matrices.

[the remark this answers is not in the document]
andrjohns: > .val() does exactly the same thing as value_of called on Matrices.

The slight difference is that ```value_of``` will always return a ```CWiseUnaryOp``` whereas ```.val()``` will return either a ```CWiseUnaryView``` or ```CWiseUnaryOp```, depending on const-ness

bbbales2:
@SteveBronder I broke out the control flow and such here.

I'm just removing the `to_var_value` optimizations (and the `to_var_value_if` code) since that is not agreed on. It's not gonna be possible to answer that without benchmarks, and not doing it is faster than what is in develop now anyway.

Anyway it'll need another review. I stuck with `value_of` to avoid the `val/val_op` thing, but it sounds like `value_of` isn't using holders for plain_types so we're safe (https://github.com/stan-dev/math/pull/2164#discussion_r521897712).

stan/math/rev/fun/quad_form.hpp
bbbales2: More docs added

bbbales2:
@SteveBronder this is ready for review again. It was greenlit before I just added some more docs for the template types and merged in develop

stan/math/rev/fun/quad_form.hpp
```diff
@@ -93,6 +94,131 @@ class quad_form_vari : public vari {
...
+
+  if (!is_constant<Mat1>::value && !is_constant<Mat2>::value) {
+    arena_t<promote_scalar_t<var, Mat1>> arena_A = A;
+    arena_t<promote_scalar_t<var, Mat2>> arena_B = B;
+
+    check_not_nan("multiply", "A", value_of(arena_A));
+    check_not_nan("multiply", "B", value_of(arena_B));
```
SteveBronder, line 131: Once we do `arena_t<promote_scalar_t<var, Mat1>> arena_A = A;` we know that `arena_A` has a var scalar so we can use `.val()` directly. I'm pretty sure `value_of()` is going to have to make a `holder()` type where calling `.val()` directly won't have to. Is there a reason you can't just use the version I wrote below that handles `var<mat>` and `mat<var>`?

https://github.com/stan-dev/math/compare/feature/quad_form-pullout-logic#diff-11ae2c18d3cbe94d5f48b9f855242f977ee3ec1d490c43723e09417f95c829f3R35

stan/math/rev/fun/trace_quad_form.hpp
```diff
@@ -93,6 +95,94 @@ inline return_type_t<EigMat1, EigMat2> trace_quad_form(const EigMat1& A,
...
+              .trace();
+
+    reverse_pass_callback([arena_A, arena_B, res]() mutable {
+      if (is_var_matrix<Mat1>::value) {
+        arena_A.adj().noalias()
+            += res.adj() * value_of(arena_B) * value_of(arena_B).transpose();
+      } else {
+        arena_A.adj()
+            += res.adj() * value_of(arena_B) * value_of(arena_B).transpose();
+      }
```
SteveBronder, line 139: [side note] I wonder if it would be useful or look nice to have something like

```cpp
update_adjoint(arena_A, res.adj() * value_of(arena_B) * value_of(arena_B).transpose());
```

Where for `var<mat>` it's just

```cpp
lhs.adj().nolias() += rhs
```

and for `mat<var>` it is 

```cpp
lhs.adj() += rhs
```

So we don't have to have these if-else statements all over

SteveBronder (approved):
Looks good! Two comments though. imo I'd rather not pay for making the `holder()` which we can just do by using `.val()` instead of `value_of()` like I have below for `quad_form` but I think this code is also fine

https://github.com/stan-dev/math/compare/feature/quad_form-pullout-logic#diff-11ae2c18d3cbe94d5f48b9f855242f977ee3ec1d490c43723e09417f95c829f3R35

stan/math/rev/fun/quad_form.hpp
bbbales2: The conclusion of this discussion that the `value_of` s here are fine and they don't create holders: https://github.com/stan-dev/math/pull/2164#discussion_r521897712 .

> Is there a reason you can't just use the version I wrote below that handles var<mat> and mat<var>?

I am basically using that code. The differences are:

1. `value_of` instead of worrying about the difference in `val` and `val_op`

2. We talked about at the last release to stop updating `mat<var>` code unless we have benchmarks showing it's faster, so I kept the old `mat<var>` stuff in place.

3. The code in that pull request returns a 1x1 matrix for the vector^T matrix vector case. This is not correct (see: https://mc-stan.org/docs/2_25/functions-reference/dot-products-and-specialized-products.html). I think there is a bug in the test framework that is letting this pass. I haven't made an issue for that yet.

Sound good?

stan/math/rev/fun/trace_quad_form.hpp
bbbales2: What about:

`noalias(lhs.adj()) += rhs`

bbbales2: Yeah what I said doesn't make sense cause you'd still need the `is_var_matrix` logic.

I can add this if you want. I'm happy with ifs though.

stan/math/rev/fun/quad_form.hpp
bbbales2: I made an issue for the 1x1 thing: https://github.com/stan-dev/math/issues/2198

Issue #2165, opened 2020-10-26
speedup reduce_sum


wds15:
## Description

The current `reduce_sum` implementation can benefit from using the var's for the shared arguments more efficient in a way avoid too many copies.

## Example

Things should just run as before, just faster.

## Expected Output

Same as before.

#### Current Version:
v3.3.0

rok-cesnovar:
@wds15 this one was closed with #2162 right?

wds15:
Yeah, done that. Thanks.

PR #2166, opened 2020-10-26
make reduce_sum more efficient by using the main thread vari's

wds15:
## Summary

This PR recycles the input operands from the main thread when running a chunk in the main thread. This gives some speedups as can be seen on the brms threading example which compares this version against the released 2.25.0.

## Tests

No change in behavior.

## Side Effects

Runs faster:

speedup-cmp

chunking-cmp

## Release notes

Increase efficiency of `reduce_sum`

## Checklist

- [X] Math issue #2165 


wds15:
@bbbales2 would be great if you can have a look as to why one of the reduce_sum tests will not compile due to problems with `zero_adjoints`. The compiler error message when trying to compile the `StanMathRev_reduce_sum, nesting_gradient` test is:

```
./stan/math/rev/functor/reduce_sum.hpp:269:12: note: in instantiation of function template specialization
      'tbb::parallel_reduce<tbb::blocked_range<unsigned long>,
      stan::math::internal::reduce_sum_impl<stan::math::test::nesting_count_lpdf<stan::math::var_value<double, void> >, void,
      stan::math::var_value<double, void>, std::__1::vector<int, std::__1::allocator<int> > &, std::__1::vector<stan::math::var_value<double,
      void>, std::__1::allocator<stan::math::var_value<double, void> > > &, std::__1::vector<int, std::__1::allocator<int> >
      &>::recursive_reducer>' requested here
      tbb::parallel_reduce(
           ^
./stan/math/prim/functor/reduce_sum.hpp:203:10: note: in instantiation of member function
      'stan::math::internal::reduce_sum_impl<stan::math::test::nesting_count_lpdf<stan::math::var_value<double, void> >, void,
      stan::math::var_value<double, void>, std::__1::vector<int, std::__1::allocator<int> > &, std::__1::vector<stan::math::var_value<double,
      void>, std::__1::allocator<stan::math::var_value<double, void> > > &, std::__1::vector<int, std::__1::allocator<int> > &>::operator()'
      requested here
         ^
./stan/math/rev/core/zero_adjoints.hpp:56:13: note: candidate function template not viable: no known conversion from 'const
      std::__1::vector<stan::math::var_value<double, void>, std::__1::allocator<stan::math::var_value<double, void> > >' to 'stan::math::var &'
      (aka 'var_value<double> &') for 1st argument
inline void zero_adjoints(var& x, Pargs&... args) {
            ^
./stan/math/rev/core/zero_adjoints.hpp:41:13: note: candidate template ignored: requirement 'is_arithmetic<stan::math::var_value<double, void>
      >::value' was not satisfied [with T = const std::__1::vector<stan::math::var_value<double, void>,
      std::__1::allocator<stan::math::var_value<double, void> > >, Pargs = <const std::__1::vector<int, std::__1::allocator<int> >>]
inline void zero_adjoints(T& x, Pargs&... args) {
            ^
./stan/math/rev/core/zero_adjoints.hpp:72:13: note: candidate template ignored: could not match 'Matrix' against 'vector'
inline void zero_adjoints(Eigen::Matrix<var, R, C>& x, Pargs&... args) {
            ^
./stan/math/rev/core/zero_adjoints.hpp:89:13: note: candidate template ignored: deduced type 'vector<...>' of 1st parameter does not match
      adjusted type 'const vector<...>' of argument [with T = stan::math::var_value<double, void>, Pargs = <const std::__1::vector<int,
      std::__1::allocator<int> >>, $2 = nullptr]
inline void zero_adjoints(std::vector<T>& x, Pargs&... args) {
            ^
./stan/math/rev/core/zero_adjoints.hpp:28:13: note: candidate function not viable: requires 0 arguments, but 2 were provided
inline void zero_adjoints() {}
```

No idea what this is supposed to mean...

bbbales2:
@wds15 I am mid-deep dive on this too lol. Gonna take a break for now. Will look at this tomorrow

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -104,40 +110,74 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
+            args_tuple_);
+
+      } else {
+        // Create nested autodiff copies of all shared arguments that do not
+        // point
+        //   back to main autodiff stack
+        auto args_tuple_local_copy = apply(
+            [&](auto&&... args) {
+              return std::tuple<decltype(deep_copy_vars(args))...>(
+                  deep_copy_vars(args)...);
```
bbbales2, line 151: Good point with this whole pull, but now I'm wondering why do we ever do these copies? Isn't it always safe to accumulate back into the main thread varis because there is no race condition?

bbbales2: At least the varis here in the sliced argument.

Even if someone passed the same variable as a sliced argument and shared argument, it is always copied for the shared argument anyway, so still no race condition.

wds15: Yes, absolutely correct. We can avoid the deep copy for the sliced argument as well. That will improve performance quite a bit more.

It's just that this change makes the assumption that the adjoints of the passed in var's are 0. Moreover, we modify things on the main AD tape, since the vari's are being used. I think this is all OK to do, but maybe needs a mention in the doc.

I will try to benchmark the latest change which recycles the sliced arguments vari's. That should make quite a difference is my guess.

It's cool to see that there is more to get from `reduce_sum`! With the scoping stuff this thing just got a decent boost.

bbbales2: > It's just that this change makes the assumption that the adjoints of the passed in var's are 0

Oh yeah, we should probably save the adjoints, set them to zero, and then restore them after we're done then. That's a bit tricky and I'm not sure if I've forgotten anything.

wds15: What about we check the vari's to be zero and throw an exception if they are not 0?

I am running benchmarks right now and using this trick for the slicing argument gives us a 10% speedup when I see that right for a bernoulli example which suffers a lot from memory stuff.

bbbales2:
> Or should we save and restore

I think save and restore. Regardless of what we do we need to restore them to their previous state (presumably zero) otherwise reverse mode will fail.

I couldn't think of a situation off the top of my head where during the forward pass we would have the reverse-mode adjoints initialized to something non-zero.

Perhaps a situation like this could happen if we've embedded an autodiff calculation in a reverse mode pass, like we might do with adjoint ODEs for instance. I'm not sure.

bbbales2:
Like this:

```
var a;
var b = f(a);
var c = b + a;
c.grad();
```

If `f` involves doing a nested autodiff in reverse mode, then it's possible the operation `c = b + a` has already made the `a` adjoint non-zero, in which case we need to be careful with it.

wds15:
ok, then let's do that... we can actually handle this probably in an easy way by looping over the already saved vari's.

wds15:
Looks like this trick looses a lot of speed when doing this store & recover of the adjoints (in particular with more CPUs)... I don't feel like I want to branch on "few cores being in use" as well. So I may just drop this idea; though one argument for still including it is that this way the CPU=1 performance should be just as good as running without reduce_sum. I lean towards dropping this in case the performance gain is not convincing enough... will try once more the Poisson example.

wds15:
Doing a bit more benchmarking I would conclude to make this optimization as it speeds up a very important case. My last comment did not sound positive, since I looked at an example where `reduce_sum` was used to apply it to a bernoulli distribution. The reducer was given N data items and N parameters in the range of 0-1. This was written in the most efficient way for `reduce_sum`, which is to slice over the parameters. However, it's important to also look at what happens if we slice over the data and treat the parameters as shared argument. This case is important, since `brms` puts every parameter in the shared parameter list.

Thus I reran the benchmarks with these factors:

- slice / shared denotes where the parameter vector is put
- N is the problem size
- CPU the number of CPUs

Overall I compared 3 flavors:

- master: version 2.25.0
- recycle: this PR
- scoped: the other PR which introduces the scoped chainablestack concept

Here are the results:

runtime

speedup

The above uses 30 replications per data point.

What is a bit concerning is that for large CPU count the "slicing" case is a bit less ideal. However, I think that this is still in the range of being ok. I cannot go to larger CPU counts on my laptop (maybe @SteveBronder ?).

I also wanted to see if now really a brms program runs faster which is why I reran the brms threading example:

speedup-cmp

chunking-cmp

So from that angle we are also good.

In summary I suggest that we merge this PR (after sorting out compilation issues with `zero_adjoints` for one unit test - @bbbales2  could you help?) and then we can try to combine this with the scoping technique, but it looks to me as if we already get most speedups which are possible. What is nice in particular is that `brms` will benefit a lot from this for models where it currently doesn't perform well (models with many random effects).

Here is the benchmark:
[reduce_sum_benchmark.txt](https://github.com/stan-dev/math/files/5458374/reduce_sum_benchmark.txt)

wds15:
Good news: I am now running more things in parallel and also more integrated. Now this PR outperforms all other code in all cases.

See:

runtime

speedup

So we can just go ahead with this PR and things get faster under all circumstances is my expectation.

wds15:
ok, let's review this. I dropped the use of `zero_adjoints` and replaced it with a loop over the `varis`. Now all tests are enabled again, everything passes and this PR makes `reduce_sum` under all circumstances I looked into.

These changes do reduce quite a bit the friction we had in the older code in a way which will make brms generated models faster (lots of shared parameters).

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -89,55 +108,93 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
-          },
-          std::move(args_tuple_local_copy));
+      if (main_thread_ == std::this_thread::get_id()) {
+        // the reducer instance running in the main thread can use the
+        // input operands given as input to reduce_sum which are
+        // stored in the main thread. These need to be zeroed before
+        // use.
+
+        for (std::size_t i = 0; i < num_vars_shared_terms_; ++i) {
+          varis_[vmapped_.size() * num_vars_per_term_ + i]->adj_ = 0.0;
```
bbbales2, line 135: We would need to save and restore these adjoints like the others

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -89,55 +108,93 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
+                                  std::forward<decltype(args)>(args)...);
+            },
+            std::move(args_tuple_local_copy));
+      }
+
+      // Accumulate adjoints of sliced_arguments and restore adjoint
+      for (std::size_t i = 0; i < r.size() * num_vars_per_term_; ++i) {
+        std::size_t term = r.begin() * num_vars_per_term_ + i;
+        partials_[term] = varis_[term]->adj_;
+        varis_[term]->adj_ = adjoints_sub_slice(i);
```
bbbales2, line 196: I don't understand this. We don't have enough information to setthe adjoints of the varis in the forward pass.

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -235,7 +290,7 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
     }
 
     for (size_t i = 0; i < num_vars_shared_terms; ++i) {
-      partials[num_vars_sliced_terms + i] = worker.args_adjoints_(i);
+      varis[num_vars_sliced_terms + i]->adj_ = shared_terms_adjoints(i);
```
bbbales2, line 293: Same question here about adjoints in the forward pass -- I don't think we can do it.

bbbales2 (changes requested):
Speedup looks cool! Thanks for the ratios n' data and everything.

A comment and a question here.

stan/math/rev/functor/reduce_sum.hpp
wds15: The `varis` of the shared arguments are saved and restored in the reduce impl function. The `varis` of the sliced argument is saved, set to 0 and restored in the reduce operation (so that it happens in parallel).

stan/math/rev/functor/reduce_sum.hpp
wds15: Here I am reading out the adjoints of the sliced terms and I do restore the adjoint of the sliced term to what is was before the reduce operation.

Nothing changes on the outer AD tree, since everything is saved, used and then restored.

stan/math/rev/functor/reduce_sum.hpp
bbbales2: Oh I see. Oh right. Yes and this answers the other question too.

stan/math/rev/functor/reduce_sum.hpp
bbbales2: And this is restoring the adjoints that were there. Wires crossed already let me look again.

stan/math/rev/functor/reduce_sum.hpp
wds15: Why not? We save and restore. Nothing changes once the reduce function is done during the forward pass.

... and using the adjoints of the outer AD tape is the main idea of this PR which I thought you agreed with already, no?

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -38,24 +38,36 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
+    const size_t num_vars_shared_terms_;  // Number of vars in shared
+                                          // arguments
+    vari** varis_;      // Points to array of the input operands
+    double* partials_;  // Points to adjoints of the partial
+                        // calculations
     Vec vmapped_;
     std::ostream* msgs_;
     std::tuple<Args...> args_tuple_;
+    const std::thread::id main_thread_;
+    std::unique_ptr<double[]> args_adjoints_holder_;
```
bbbales2, line 50: What is args_adjoints_holder_ used for?

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -66,10 +78,17 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
         : num_vars_per_term_(other.num_vars_per_term_),
           num_vars_shared_terms_(other.num_vars_shared_terms_),
-          sliced_partials_(other.sliced_partials_),
+          varis_(other.varis_),
+          partials_(other.partials_),
           vmapped_(other.vmapped_),
           msgs_(other.msgs_),
-          args_tuple_(other.args_tuple_) {}
+          args_tuple_(other.args_tuple_),
+          main_thread_(other.main_thread_),
```
bbbales2, line 86: So every `recursive_reducer` is a copy of the original `recursive_reducer` so the `main_thread_` is always set to the thread that made the first copy. At least, it is always one thread which is all you need to guarantee no race condition?

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -216,12 +269,14 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
     save_varis(varis, vmapped);
     save_varis(varis + num_vars_sliced_terms, args...);
 
-    for (size_t i = 0; i < num_vars_sliced_terms; ++i) {
-      partials[i] = 0.0;
+    Eigen::VectorXd shared_terms_adjoints(num_vars_shared_terms);
+
+    for (size_t i = 0; i < num_vars_shared_terms; ++i) {
+      shared_terms_adjoints(i) = varis[num_vars_sliced_terms + i]->adj_;
```
bbbales2, line 275: `shared_terms_adjoints.coeffRef(i)` avoids out of bounds checks that `operator()` does for eigen types.

`shared_terms_adjoints.coeff(i)` is the version to use for reading (const).

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -148,11 +205,7 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
      */
     inline void join(const recursive_reducer& rhs) {
       sum_ += rhs.sum_;
-      if (args_adjoints_.size() != 0 && rhs.args_adjoints_.size() != 0) {
-        args_adjoints_ += rhs.args_adjoints_;
-      } else if (args_adjoints_.size() == 0 && rhs.args_adjoints_.size() != 0) {
-        args_adjoints_ = rhs.args_adjoints_;
-      }
+      args_adjoints_.noalias() += rhs.args_adjoints_;
```
bbbales2, line 208: Oh were the ifs here earlier just not used at all? I don't think this needs an `.noalias`.

From: https://eigen.tuxfamily.org/dox/group__TopicAliasing.html

> Matrix multiplication is the only operation in Eigen that assumes aliasing by default, under the condition that the destination matrix is not resized. Thus, if matA is a squared matrix, then the statement matA = matA * matA; is safe. All other operations in Eigen assume that there are no aliasing problems, either because the result is assigned to a different matrix or because it is a component-wise operation.

bbbales2: Could ask Steve or Tadej to clarify they know this stuff better

stan/math/rev/functor/reduce_sum.hpp
wds15: To store a pointer to the double array for the shared adjoints. The root reducer writes directly to the partials array and the child reducers allocate their own space for that. Hence we need eigen maps everywhere is what I figured.

stan/math/rev/functor/reduce_sum.hpp
wds15: Ok. Will change.

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -89,55 +108,93 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
+            [&](auto&&... args) {
+              accumulate_adjoints(args_adjoints_.data(),
+                                  std::forward<decltype(args)>(args)...);
+            },
+            std::move(args_tuple_local_copy));
+      }
+
+      // Accumulate adjoints of sliced_arguments and restore adjoint
+      for (std::size_t i = 0; i < r.size() * num_vars_per_term_; ++i) {
+        std::size_t term = r.begin() * num_vars_per_term_ + i;
```
bbbales2, line 194: Does this work (to avoid term):

```
for (std::size_t i = r.begin() * num_vars_per_term_; i < r.end() * num_vars_per_term_; ++i)
```

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -89,55 +108,93 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
-      for (size_t i = r.begin(); i < r.end(); ++i) {
-        local_sub_slice.emplace_back(deep_copy_vars(vmapped_[i]));
+      // Create copies of sliced argument which refer to the main
+      // thread
+      std::decay_t<Vec> local_sub_slice(vmapped_.begin() + r.begin(),
+                                        vmapped_.begin() + r.end());
+
+      // first save current adjoints of sliced argument (which we
+      // restore later) and set these to zero
+      Eigen::VectorXd adjoints_sub_slice(r.size() * num_vars_per_term_);
```
bbbales2, line 121: Can this variable be like `saved_adjoints`/`adjoints_backup` or something to indicate it is being used to save stuff in the forward pass as opposed to stuff for the reverse pass?

stan/math/rev/functor/reduce_sum.hpp
bbbales2: But can the `Eigen::Map` just store that pointer directly?

stan/math/rev/functor/reduce_sum.hpp
bbbales2: Yah yah this goes with the other comments I got my wires crossed.

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -216,12 +269,14 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
     save_varis(varis, vmapped);
     save_varis(varis + num_vars_sliced_terms, args...);
 
-    for (size_t i = 0; i < num_vars_sliced_terms; ++i) {
-      partials[i] = 0.0;
+    Eigen::VectorXd shared_terms_adjoints(num_vars_shared_terms);
```
bbbales2, line 272: I like backup or something in the name here too -- like `backup_shared_adjoints`.

wds15: Sure.

stan/math/rev/functor/reduce_sum.hpp
wds15: Don't think so. Eigen Map is there to deal with raw memory buffers... but the unique_ptr is there to cleanup things. The Map alone would not de-allocate the double array.

stan/math/rev/functor/reduce_sum.hpp
wds15: Yes. Whatever reducer gets to execute in the main thread will get to use the outer AD tape which originated in that thread. Doing it by thread id ensures good thread locality, since the TBB does not guarantee anything where things run.

stan/math/rev/functor/reduce_sum.hpp
wds15: But putting noalias there does not hurt, right?

I don't think the if stuff was used... when I read the TBB doc I figured that whenever a reduce worker is split off, then it will do some stuff and hence this delay init stuff should not be needed.

stan/math/rev/functor/reduce_sum.hpp
wds15: but I need to index into the saved adjoints which indexes starting from 0.

stan/math/rev/functor/reduce_sum.hpp
bbbales2: Yeah doesn't hurt, but might as well clean up while we're here. I asked Steve in slack.

wds15:
I hope I got all your change requests addressed. Happy?

stan/math/rev/functor/reduce_sum.hpp
```diff
@@ -66,10 +78,17 @@ struct reduce_sum_impl<ReduceFunction, require_var_t<ReturnType>, ReturnType,
...
-          sliced_partials_(other.sliced_partials_),
+          varis_(other.varis_),
+          partials_(other.partials_),
           vmapped_(other.vmapped_),
           msgs_(other.msgs_),
-          args_tuple_(other.args_tuple_) {}
+          args_tuple_(other.args_tuple_),
+          main_thread_(other.main_thread_),
+          args_adjoints_holder_(
+              std::make_unique<double[]>(num_vars_shared_terms_)),
```
bbbales2, line 88: Well here the unique_ptr is constructed pointing to the memory address `num_vars_shared_terms_`, and in the first constructor it points to `nullptr`.

I don't understand how `args_adjoints_` works in this copied version. Like isn't it going to be a length `num_vars_shared_terms_` Map pointing at memory address `num_vars_shared_terms_`?

wds15: So there are two cases:

1. the root worker node is mapping directly into the partials array of the final output
2. the split off workers have their own memory space where they accumulate the adjoints of the partials (as the workers get to evaluate multiple work chunks not just one)

So for case 1 I don't need to allocate anything such that it points nowhere (nullptr). For the case 2, we have to allocate an array which holds the adjoints of the shared terms. The unique_ptr is there to store a reference to that and make sure we wipe out the memory upon destruction. The Eigen Map then needs to be setup with a pointer from the unique_ptr and the size is num_vars_shared_terms (one adjoint per var).

Does it now click? (or do I have a crack)?

bbbales2: > For the case 2, we have to allocate an array which holds the adjoints of the shared terms.

Oh I didn't realize that's what `unique_ptr<double[]>(N)` did. Will stare some more.

bbbales2:
Looks good! I took your noalias away. @t4c1 said it only does stuff for multiplication. I'll approve and merge when the tests pass.

wds15:
I am not sure if you have the power to approve now that you have pushed some changes. Maybe someone else has to do that now.. @SteveBronder ?

I am fine with taking noalias away... it's just that I had once a case where I got crazy speedups by adding it such that I made it a habit to add it; but I am happy to update my rule to that and put more trust into Eigens magic.

bbbales2:
Yeah we can call in someone else to approve and merge if it tells me I can't. But I still have the approve button here so hopefully works when the tests pass.

rok-cesnovar:
I think only the PR author cant approve so there is no technical Github restriction. I think this isnt an issue anywas given how small the change that Ben pushed was.

wds15:
Off topic... but having stared at the logic of this so much I start to see that an MPI `reduce_sum` is doable. Would you be up for that and also have a hand to code it up? I would suggest to do the prim version first, then a brute force (aka copy a lot of stuff) rev version to start with. This isn't a priority for me, but maybe a nice project along the way? But maybe before embarking on that we gauge interest from the community on this feature? If we do that, then it would be good if more people other than me have decent MPI knowledge as to what we have in stan-math. Right now the bus-driver number is almost 1 for MPI topics with stan-math is my impression.

wds15:
@bbbales2 any idea about the breaking test? In the file `./test/unit/math/mix/functor/reduce_sum_part1_test.cpp` the test `StanMath_reduce_sum_static, start_end_slice` breaks. Looking at this test I don't quite get what the intent was and why it's breaking. It smells like an issue with the ad test framework to me... but maybe that is something I need to worry about...

bbbales2:
> @bbbales2 any idea about the breaking test?

I am escaping for the day, but will look at this tomorrow.

> MPI reduce_sum

That is exciting. Yes that sounds cool. I would like to move people away from `map_rect`.

> Would you be up for that and also have a hand to code it up?

I can review and try to do in depth comments and stuff but I don't want to take the lead on it.

> But maybe before embarking on that we gauge interest from the community on this feature?

I think there's a communicator problem Yi wanted resolved (or some technical thing with MPI, I forget the details).

I'd like to deprecate `map_rect` and go to a `reduce_sum` for MPI things, but that would require model rewrites in a backwards compatibility-breaking way, so maybe this is a non-starter.

This sounds kindof difficult now that I think about it. Maybe parallel reverse mode for threading is lower hanging fruit (and will get  `parallel_map` moving).

wds15:
Whow... this was a real bug in the code and it kills the principle to use the main thread vari's. The issue is that the user can use input operands (in different arguments) which are linked to one another. Not a very common case, but that's a valid use of the function and the way we save vari's and use precomputed vari's absolutley requires that we make deep copies.

Interestingly, we still get good speedups when we just do once a deep copy on the main thread:

runtime

speedup

Having seen this, I will revise the scoped thing once more to use thread-local copies of the operands instead of putting them into the reducers. It's all about cache locality is my understanding here.

wds15:
Very interesting: I changed the "scoped" approach to use thread local scope ChainableStacks with the intent to improve memory locality as this will then always use a scoped ChainableStack specific to a given thread. And indeed this improves things even further:

runtime

speedup

I would suggest to merge this PR and then we go for the scoping thing as a next step. What is super cool is that the "shared" case is so much accelerating from this stuff, since this is what is important for brms.

wds15:
This PR can still move, but I had another idea which may be simpler and yet gets us the performance gains in reach here. This would be a "creative" way of instrumenting the AD system...

bbbales2 (changes requested):
So using the varis on the main stack was a complete no-go? Could you explain more what makes that impossible? I approved this so I want to know where I messed up.

It looks like now what is happening in the main thread only makes one deep copy of the input arguments and uses the `partials` pointer directly.

This pull request is making it so work is scattered about in a thread-local way instead of being based on work chunks: https://github.com/stan-dev/math/pull/2162 . Does that pull make this one not necessary (the big change here seems to be doing just the main thread in a thread-local way)? If we added to that other thread accumulation of shared argument adjoints in a thread-local way then I think it would have everything here and more.

bbbales2:
> The issue is that the user can use input operands (in different arguments) which are linked to one another.

Is the issue a race condition in reverse mode? Something forward mode?

bbbales2:
Is this just closed now? Or is there anything to pull out?

wds15:
We can close this now.

I wrote a test which gets broken if you follow the original design of this PR. See

https://github.com/stan-dev/math/blob/835041555db4a09dc37941a6575cc6071a580821/test/unit/math/rev/functor/reduce_sum_test.cpp#L393

I am happy to explain more, of course.

PR #2169, opened 2020-10-27
Get `sd`, `log_softmax`, and `log_sum_exp` fully `var<mat>` compatible (Issue #2101)

bbbales2:
## Summary

This makes `sd`, `log_sum_exp`, and `log_softmax` fully `var<mat>` compatible (and `apply_vector_unary` in the process).

## Release notes

Updated `sd`, `log_softmax`, and `log_sum_exp` to work fully with `var<mat>`

## Checklist

- [x] Math issue #2101 


stan/math/prim/functor/apply_vector_unary.hpp
```diff
@@ -17,6 +17,52 @@ namespace math {
...
+ * the same vector type as the input is returned.
+ *
+ * Two taxonomies of unary vector functions are implemented:
+ * - f(vector) -> vector
+ * - f(vector) -> scalar
+ *
+ * This base template class takes (and returns) Eigen expression templates.
+ */
+template <typename T>
+struct apply_vector_unary<T, require_var_matrix_t<T>> {
```
bbbales2, line 33: Should this specialization go in `rev`?

stan/math/prim/functor/apply_vector_unary.hpp
```diff
@@ -17,6 +17,52 @@ namespace math {
...
+   * returning a vector.
+   *
+   * @tparam T Type of argument to which functor is applied.
+   * @tparam F Type of functor to apply.
+   * @param x Eigen input to which operation is applied.
+   * @param f functor to apply to Eigen input.
+   * @return Eigen object with result of applying functor to input
+   */
+  template <typename F>
+  static inline T apply(const T& x, const F& f) {
```
bbbales2, line 45: The other versions specialize on whether or not `T` contains something that is array-like or matrix-like and do an `eval`.

I do not think we return expression-like things from `var_value` functions -- they're always `var_value` s, and the way to make sure the input and output types match is to just put `T` in both places right?

stan/math/prim/meta/is_container.hpp
```diff
@@ -19,7 +20,8 @@ namespace stan {
  */
 template <typename Container>
 using is_container = bool_constant<
-    math::disjunction<is_eigen<Container>, is_std_vector<Container>>::value>;
+  math::disjunction<is_eigen<Container>, is_std_vector<Container>,
+		    is_var_matrix<Container>>::value>;
```
bbbales2: A `var<mat>` should count as a container, right? This might break template logic elsewhere but we can fix it elsewhere.

stan/math/rev/fun/sd.hpp
```diff
@@ -13,49 +13,89 @@
...
-    double size_m1 = dtrs_map.size() - 1;
-    double sd = sqrt(sum_of_squares / size_m1);
-
-    if (sum_of_squares < 1e-20) {
-      partials_map.fill(inv_sqrt(static_cast<double>(dtrs_map.size())));
-    } else {
-      partials_map = diff.array() / (sd * size_m1);
-    }
-    return var(new stored_gradient_vari(sd, dtrs_map.size(), varis, partials));
+    return internal::sd_impl(dtrs_map);
```
bbbales2: This pattern work to handle the `var<mat>` and `mat<var>` implementations separately? If so we can copy it in the other places.

We'd do a similar thing for `log_sum_exp` and for `log_softmax`.

test/unit/math/mix/fun/sd_test.cpp
```diff
@@ -6,13 +6,21 @@ void expect_sd(const T& m, const stan::test::ad_tolerances& tols
...
     v(i) = m(i);
     rv(i) = m(i);
+    sv[i] = m(i);
   }
+
   stan::test::expect_ad(tols, f, v);
   stan::test::expect_ad(tols, f, rv);
   stan::test::expect_ad(tols, f, m);
+  stan::test::expect_ad(tols, f, sv);
+
```
bbbales2, line 20: We're missing autodiff tests for `std::vector<mat<var>>`.

I checked the docs and it's not doc'd (so not exposed to users yet). `log_sum_exp` and `log_softmax` have tests though.

I guess I'll just copy those eventually but I tried a couple simple ones and they were failing for me today so that's something for the future.

test/unit/math/mix/fun/sd_test.cpp
```diff
@@ -6,13 +6,21 @@ void expect_sd(const T& m, const stan::test::ad_tolerances& tols
...
+  stan::test::expect_ad_matvar(tols, f, v);
+  stan::test::expect_ad_matvar(tols, f, rv);
+  stan::test::expect_ad_matvar(tols, f, m);
```
bbbales2, line 23: We will need this to work with `std::vector<var<mat>>` though. Haven't looked at it yet.

bbbales2 (commented):
Questions & comments

SteveBronder:
I'll take a look at this in the morning

bbbales2:
Ping

test/unit/math/mix/fun/sd_test.cpp
bbbales2: I updated the `sd` implementation to be vectorized in the way that `apply_vector_unary` supports (so `std::vector`, `vector`, `row_vector`, and `matrix` all get passed in as an Eigen vector map, and `std::vector` s of any of those work as well) and added the test framework changes.

stan/math/prim/fun/sd.hpp
```diff
@@ -22,14 +22,19 @@ namespace math {
...
+inline auto sd(const T& m) {
   using std::sqrt;
-  check_nonzero_size("sd", "m", m);
-  if (m.size() == 1) {
-    return 0.0;
-  }
-  return sqrt(variance(m));
+
+  return apply_vector_unary<T>::reduce(m, [](const auto& x) {
+    check_nonzero_size("sd", "x", x);
```
bbbales2, line 30: I think these `check_nonzero_size` checks go inside the `apply_vector_unary`.

test/unit/math/test_ad.hpp
```diff
@@ -1724,6 +1774,75 @@ void expect_ad_matvar_v(const ad_tolerances& tols, const F& f,
...
+  }
+  plain_type_t<decltype(f(A_mv))> A_mv_f;
+  vec_var_mat A_vm;
+  for (auto xi : x) {
+    A_vm.push_back(xi);
+  }
+  plain_type_t<decltype(f(A_vm))> A_vm_f;
+  // Check return type is correct
+  check_return_type(A_mv_f, A_mv);
+  check_return_type(A_vm_f, A_vm);
```
bbbales2, line 1812: In this case `A_mv_f` and `A_vm_f` are both going to be `std::vector<var>` types (there are no `var_value<EigenBase>` things going on here).

I think what this function is trying to do is just check that the semantics of `promote_var_matrix_t` are respected.

So I think I need to update `check_return_type` to handle `std::vector` s, but it might be as simple as deferring all the logic to `promote_var_matrix_t`. Will look at this later though. Right now these do nothing, but really they should be doing something if the functions returned `std::vector<var_value<EigenBase>>` things.

bbbales2 (commented):
@SteveBronder this is ready to look at. Not 100% there but I have a couple questions. One we get those ironed out it will make it easy to finish `mat<var>` ing `log_sum_exp` and `log_softmax`.

stan/math/prim/fun/sd.hpp
SteveBronder: Should they exists? Like I think normally we just return if a matrix is size zero right?

stan/math/prim/functor/apply_vector_unary.hpp
SteveBronder: Yeah I think it makes sense

stan/math/rev/fun/sd.hpp
```diff
@@ -13,49 +13,93 @@
...
+  double size_m1 = x.size() - 1;
+  double sd = sqrt(sum_of_squares / size_m1);
+
+  if (sum_of_squares < 1e-20) {
+    partials_map.fill(inv_sqrt(static_cast<double>(x.size())));
+  } else {
+    partials_map = diff.array() / (sd * size_m1);
+  }
+  return var(new stored_gradient_vari(sd, x.size(), varis, partials));
+}
```
SteveBronder, line 58: I think you can get rid of this specialization if you just do the other one like

```cpp
template <typename T, require_rev_matrix_t<T>* = nullptr>
var sd_impl(const T& x) {
  if (x.size() == 1) {
    return 0.0;
  }
  arena_t<T> arena_x = x;
  auto arena_diff = to_arena((arena_x.val().array() - arena_x.val().mean()).matrix());
  double sum_of_squares = arena_diff.squaredNorm();
  double sd = std::sqrt(sum_of_squares / (x.size() - 1));

  var res = sd;

  reverse_pass_callback([arena_x, res, arena_diff]() mutable {
    arena_x.adj() += (res.adj() / (res.val() * (arena_x.size() - 1))) * arena_diff;
  });

  return res;
}
```

Also the requires for this impl makes it look like it will work for std vectors since it has `require_st_var` in it but it also uses `x.vi()` `x.val()` etc so idt it would work

test/unit/math/mix/fun/sd_test.cpp
SteveBronder: Nice!

test/unit/math/test_ad.hpp
```diff
@@ -1435,14 +1435,36 @@ void expect_ad_vectorized_binary(const F& f, const T1& x, const T2& y) {
...
 template <typename T1, typename T2,
-          require_all_var_t<scalar_type_t<T1>, scalar_type_t<T2>>* = nullptr>
+          require_all_std_vector_st<is_var, T1, T2>* = nullptr>
+void expect_near_rel_var(const std::string& message, T1&& x, T2&& y,
+                         const ad_tolerances& tols) {
+  stan::math::check_size_match("expect_near_rel_var", "x", x.size(), "y",
+                               y.size());
+  for (size_t i = 0; i < x.size(); ++i) {
+    expect_near_rel(message + std::string(" values"), x[i].val(), y[i].val(),
+                    tols.gradient_val_);
```
SteveBronder, line 1453: [optional] might be nice to have something like 

```suggestion
    expect_near_rel(message + std::string(" values"), x[i].val(), y[i].val(),
                    tols.gradient_val_) << "Failed at (i): " << i;
```

test/unit/math/test_ad.hpp
```diff
@@ -1724,6 +1774,75 @@ void expect_ad_matvar_v(const ad_tolerances& tols, const F& f,
...
+void expect_ad_matvar_v(const ad_tolerances& tols, const F& f,
+                        const std::vector<EigMat>& x) {
+  using stan::plain_type_t;
+  using stan::math::promote_scalar_t;
+  using stan::math::var;
+  using stan::math::var_value;
+  using stan::math::test::type_name;
+  using vec_mat_var = std::vector<promote_scalar_t<var, EigMat>>;
+  using vec_var_mat = std::vector<var_value<plain_type_t<EigMat>>>;
+  vec_mat_var A_mv;
```
SteveBronder, line 1800: I think it the future it would be good to change `_mv` to `_mat_var` etc. but for now let's keep them all in the same style

test/unit/math/test_ad.hpp
SteveBronder: Could you just change it so that if the input is an `std::vector` it expects the output to also be in a vector? Then just does the logic on the underlying type

SteveBronder (changes requested):
Couple quick comments

stan/math/rev/fun/sd.hpp
bbbales2: I want to avoid the `mat<var>` stuff cause that raises the question are we faster than what was there. We'll get to the benchmarking stuff eventually.

I should probably change this to `callback_vari` though now that I look at the code more closely.

stan/math/prim/fun/sd.hpp
bbbales2: I think in this case it makes sense. What is the standard deviation of `{ }`? I could see an argument for 0.0, but also throwing an error seems to make sense too.

SteveBronder: That's fair

stan/math/rev/fun/log_softmax.hpp
```diff
@@ -44,6 +44,74 @@ class log_softmax_elt_vari : public vari {
...
+  check_nonzero_size("log_softmax", "alpha", alpha);
+
+  const auto& theta = to_ref(alpha.val().array() - alpha.val().maxCoeff());
+
+  return make_callback_vari((theta.array() - log(theta.exp().sum())).matrix(),
+			    [alpha](const auto& res) mutable {
+    alpha.adj().noalias()
+      += res.adj_
+      - (res.adj_.sum() * res.val_.array().exp()).matrix();
+  });
```
bbbales2, line 113: @t4c1, @SteveBronder writing this code I wanted a `make_callback_var` to wrap `make_callback_vari`. The argument pass to the functor can still be the vari to avoid the pointer chasing.

I also want `.adj()` and `val()` on `vari_value`.

That all sound okay to add?

bbbales2: I want `make_callback_var` to avoid type algebra. In this case the input and output types will match, in other cases they won't and I would end up writing:

```
using ret_type = decltype((theta.array() - log(theta.exp().sum()));
return var_value<ret_type>(...);
```

test/unit/math/test_ad.hpp
```diff
@@ -1632,6 +1715,25 @@ void check_return_type(const ReturnType& ret, const Type& x) {
...
+ * @tparam Type the type of the input to the function `f()`.
+ * @param ret output of function
+ * @param x input of function
+ */
+template <typename ReturnType, typename Type,
+          require_std_vector_t<ReturnType>* = nullptr,
+          require_std_vector_vt<is_matrix, Type>* = nullptr>
+void check_return_type(const ReturnType& ret, const Type& x) {
+  if (ret.size() > 0 && x.size() > 0)
+    check_return_type(ret[0], x[0]);
```
bbbales2, line 1734: I will not be offended if you want me to take a closer look at `check_return_type`.

I think the logic we want is:

1. If there are only `var_value<double>` s on the input, then there should only be `var_value<double>` s on the output
2. If there are `var_value<not double>` s  on the input, then there should be no `var_value<double>` s on the output

I think this would need an extra template program `var_value_t` to extract the `var_value` from a generic input type and I am too lazy to write it today.

test/unit/math/test_ad.hpp
```diff
@@ -1563,6 +1586,64 @@ inline void test_matvar_gradient(const ad_tolerances& tols,
...
+                                 ResultMatVar& A_mv_f, ResultVarMat& A_vm_f,
+                                 const MatVar& A_mv, const VarMat& A_vm) {
+  for (size_t i = 0; i < A_vm_f.size(); ++i) {
+    A_vm_f[i].adj() = 1;
+    A_mv_f[i].adj() = 1;
+    stan::math::grad();
+    expect_near_rel_var("var<Matrix> vs Matrix<var> input", A_vm, A_mv, tols);
+    stan::math::set_zero_all_adjoints();
+  }
+}
```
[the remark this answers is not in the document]
SteveBronder: Or is this for when a `var<mat>` function would return a `std::vector<var>`?

test/unit/math/test_ad.hpp
SteveBronder: I think the logic is fine here unless I'm missing something. This is a specialization for `std::vector<T>` that just checks that the inner type for the input/output is correct. If that's what this does then I think it's fine

SteveBronder (changes requested):
Just a couple little things to change and then this looks good.

andrjohns:
Quick Q - Why the addition of ```_impl``` functions here? I thought I had the ```apply_vector_unary``` implementations working, or am I remembering something else?

bbbales2:
@andrjohns it's separate implementations for `var<mat>` and `mat<var>` types. Does that make sense?

I could make them `sd` overloads and then specialize the `apply_vector_unary` when the input is a `std::vector`. Now that I say that outloud I kinda like it better, though it subverts some of the `apply_vector_unary` pattern. Also now that you mention the `_impl` s I guess some of the lambdas are defunct now.

andrjohns:
Is the separate implementations because ```mat<var>``` isn't compatible with the ```calback_vari``` approach or because of ```apply_vector_unary```? 

Found the branch I was thinking of, I thought I had both ```var<mat>``` and ```mat<var>``` passing tests on this branch: https://github.com/stan-dev/math/compare/develop...andrjohns:feature/issue-2098-vec_unary_var_mat

andrjohns:
Also, sorry if this is retreading obvious stuff with ```var<mat>```, still catching up!

bbbales2:
@andrjohns yeah this pull has those changes too (so that `apply_vector_unary` works with `var<mat>`). At least I hope they're the same only very roughly checking. I didn't realize you had a branch or I would have used that -- apologies for the duplication.

It's possible to do the `mat<var>` and `var<mat>` implementations in the same code, but two things have made me stop trying to do this:

1. We got gridlocked before trying to carefully benchmark `mat<var>` at the expense of ever getting `var<mat>` stuff in, so I'm just avoiding the `mat<var>` implementations now (it's easy to accidentally slow down the existing code by 10% and hard to get it benchmarked and fixed).

2. `mat<var>` and `var<mat>`, even when we do them both with `reverse_pass_callback` they end up looking slightly different. (`.val()` isn't a problem for `var<mat>`, but with `mat<var>` you want to only do it once cuz it's slow)

andrjohns:
Ah that all makes sense thanks

stan/math/rev/fun/log_softmax.hpp
```diff
@@ -44,6 +44,73 @@ class log_softmax_elt_vari : public vari {
...
+ * @tparam T Type of input vector.
+ * @param[in] alpha Unconstrained input vector.
+ * @return Softmax of the input.
+ * @throw std::domain_error If the input vector is size 0.
+ */
+template <typename T, require_var_matrix_t<T>* = nullptr>
+T log_softmax_impl(const T& alpha) {
+  check_nonzero_size("log_softmax", "alpha", alpha);
+
+  const auto& theta = to_ref(alpha.val().array() - alpha.val().maxCoeff());
```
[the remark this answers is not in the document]
bbbales2: I want it to evaluate to a temporary -- switched to an `.eval()`.

test/unit/math/test_ad.hpp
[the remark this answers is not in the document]
bbbales2: This is for `std::vector<var>` outputs, not `std::vector<var<mat>>`

stan/math/rev/fun/log_softmax.hpp
```diff
@@ -44,6 +44,73 @@ class log_softmax_elt_vari : public vari {
...
+
+  return make_callback_vari(
+      (theta.array() - log(theta.exp().sum())).matrix(),
+      [alpha](const auto& res) mutable {
+        alpha.adj().noalias()
+            += res.adj_ - (res.adj_.sum() * res.val_.array().exp()).matrix();
+      });
```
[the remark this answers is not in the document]
bbbales2: I added this

bbbales2:
@SteveBronder I made a bunch of changes:

1. Added `val()`, `adj()` to all the `varis`. Let me know if you want this moved to a different pull or you want me to add tests. I just replaced all the `.val_` and `.adj_` calls in the current tests with `.val()` and `.adj()`.

2. Added `make_callback_var`

3. I got rid of the `_impl` s and instead lined those up as overloads to go along with `apply_vector_unary`. So for each of `sd`, `log_softmax`, and `log_sum_exp` there's one version of the function that handles `mat<var>`, one version that handles `var<mat>`, and then an `apply_vector_unary` version that handles `std::vector<T>` (if `T` is a `std::vector<var>` `apply_vector_unary` changes it into an `Eigen::Map`). @andrjohns feel free to comment on this if you want.

stan/math/rev/fun/sd.hpp
```diff
@@ -14,51 +14,88 @@ namespace stan {
...
+var sd(const T& x) {
+  check_nonzero_size("sd", "x", x);
+
+  if (x.size() == 1) {
+    return 0.0;
+  }
+
+  auto arena_diff = to_arena((x.val().array() - x.val().mean()).matrix());
+  double sum_of_squares = arena_diff.squaredNorm();
+  double sd = std::sqrt(sum_of_squares / (x.size() - 1));
```
SteveBronder, line 78: [optional] You could do the little loop thing here to make this faster but it's fine as is

SteveBronder (approved):
This all looks good to me! You have to fix up the docs but then ping me and I'll approve

stan/math/rev/fun/sd.hpp
bbbales2: > You have to fix up the docs

Do you mean the doxygen docs or the function reference docs (second definitely needs updated and I wouldn't doubt the first lol)?

bbbales2: Oh doxygen my bad

bbbales2: @SteveBronder I don't know what was going on with the docs. I changed all the variables to be named `x` to get them to work. That was a real hairtugger

bbbales2:
Woof, I had to add `.val()` and `.adj()` (and the `_op`) accessors to the opencl stuff. I'm firmly into I-don't-know-what-I'm-doing territory, so in the likely case this fails, I think I'll just revert the `.val()` and `.adj()` stuff.

SteveBronder (approved):

PR #2172, opened 2020-10-28
Allow var_value<Matrix> to var_value<Block> Assignment

SteveBronder:
## Summary

Quick bug fix, for doing 

```cpp
  var_value<Eigen::MatrixXd> A_v(A);
  var_value<Eigen::MatrixXd> B_v(B);
  A_v.block(0, 0, 2, 2) = B_v;
```

We need the constructors for `var<mat>` to require that the input is assignable and not convertible. This just changes the requires on the constructors for `var<mat>` to allow that

## Tests

Test added to `var_test` that checks adjoints are propogated correctly

## Side Effects

Nope

## Release notes

Bug fix for assignment to blocks of a `var<matrix>`

## Checklist

### DIFF
--- a/test/unit/math/rev/core/var_test.cpp
+++ b/test/unit/math/rev/core/var_test.cpp
@@ -375,7 +375,6 @@ TEST_F(AgradRev, var_matrix_view_block_from_plain) {
   EXPECT_FLOAT_EQ(B_v.adj()(3), 5);
 }
 
-
 /**
  * Tests that views of a var<Matrix> receive the adjoints of the original
  * matrix.
--- a/test/unit/math/rev/core/var_test.cpp
+++ b/test/unit/math/rev/core/var_test.cpp
@@ -355,6 +355,27 @@ TEST_F(AgradRev, var_vector_views_const) {
   var_vector_views_const_test<Eigen::RowVectorXd>();
 }
 
+TEST_F(AgradRev, var_matrix_view_block_from_plain) {
+  using stan::math::sum;
+  using stan::math::var_value;
+  Eigen::MatrixXd A(4, 4);
+  A << 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15;
+  var_value<Eigen::MatrixXd> A_v(A);
+  Eigen::MatrixXd B(2, 2);
+  B << 2, 3, 4, 5;
+  var_value<Eigen::MatrixXd> B_v(B);
+  for (Eigen::Index i = 0; i < A_v.size(); ++i) {
+    A_v.adj().coeffRef(i) = i;
+  }
+  A_v.block(0, 0, 2, 2) = B_v;
+  stan::math::grad();
+  EXPECT_FLOAT_EQ(B_v.adj()(0), 0);
+  EXPECT_FLOAT_EQ(B_v.adj()(1), 1);
+  EXPECT_FLOAT_EQ(B_v.adj()(2), 4);
+  EXPECT_FLOAT_EQ(B_v.adj()(3), 5);
+}
+
+
 /**
  * Tests that views of a var<Matrix> receive the adjoints of the original
  * matrix.
--- a/test/unit/math/rev/core/var_test.cpp
+++ b/test/unit/math/rev/core/var_test.cpp
@@ -355,7 +355,7 @@ TEST_F(AgradRev, var_vector_views_const) {
   var_vector_views_const_test<Eigen::RowVectorXd>();
 }
 
-TEST_F(AgradRev, var_matrix_view_block_from_plain) {
+TEST_F(AgradRev, var_matrix_view_block_from_plain_test) {
   using stan::math::sum;
   using stan::math::var_value;
   Eigen::MatrixXd A(4, 4);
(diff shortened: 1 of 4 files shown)


test/unit/math/rev/core/var_test.cpp
```diff
@@ -355,6 +355,26 @@ TEST_F(AgradRev, var_vector_views_const) {
...
+  Eigen::MatrixXd A(4, 4);
+  A << 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15;
+  var_value<Eigen::MatrixXd> A_v(A);
+  Eigen::MatrixXd B(2, 2);
+  B << 2, 3, 4, 5;
+  var_value<Eigen::MatrixXd> B_v(B);
+  for (Eigen::Index i = 0; i < A_v.size(); ++i) {
+    A_v.adj().coeffRef(i) = i;
+  }
+  A_v.block(0, 0, 2, 2) = B_v;
```
bbbales2, line 370: This result surprises me.

I would expect that since B_v gets assigned to the subblock of A_v after the adjoints are set that it would know nothing about the adjoints. The assignment to the subblock is like creating a new reverse mode variable there.

I expect this behavior with the code:
```
A_v.block(0, 0, 2, 2) = B_v;
for (Eigen::Index i = 0; i < A_v.size(); ++i) {
  A_v.adj().coeffRef(i) = i;
}
```

SteveBronder: Yeah so it took/(is taking) me a while to wrap my brain around how the slicing works but essentially the rule is to propogate adjoints to the assignee and leave the original matrix in the state before the block assignment happened. I tried writing down a splainer on this below lmk if it helps clarify anything (I was also writing it to convince myself I understood how all this worked lol)

So like say we just had an `Eigen::Matrix<var, -1, -1>` like the below

```
A:
val:
0, 1, 2
3, 4, 5
6, 7, 8

adj:
0, 0, 0
0, 0, 0
0, 0, 0
```

And another

```
B:
val:
5, 4, 3
2, 1, 0

adj:
0, 0, 0
0, 0, 0
```

The important thing to remember here is that the vars in that matrix are just pointers to vari, so assigning to them doesn't erase anything.

The reverse pass is the part that is tricky, but I think describing what happens for a normal matrix helps. For an `Eigen::Matrix<var, -1, -1>` when we do

```cpp
Eigen::Matrix<var, -1, -1> A_v(A_values);
auto some_stuff = multiply(A_v, A1_v);
A_v.block(0, 0, 2, 3) = B_v;
```

Since `var` just holds a pointer to vari, when we do that block assignment it's not erasing anything, so even though `Blah_v` after that block assignment will reference the vars in `OtherBlahMatrix` the multiply is still going to see the old `var` and `vari`.Similarly, any ops that happen on `A_v` are now actually going to be referencing the vari pointers in `B_v`!!



So we need to replicate that with `var<mat>`, but now we can't just copy some pointers and need a callback scheme to get `A_v` into the state before that block operation happened. Let's make the same matrices above into `var<mat>` and assign `B_v` to a block of `A_v` like 

```cpp
var_value<Eigen::Matrix<double, -1, -1>> A_v(A_values);
auto some_stuff = multiply(A_v, A1_v);
A_v.block(0, 0, 2, 3) = B_v;
```

and then do

```cpp
auto D_v = other_opA_v, C_v) // imagine C_v is some other var matrix
```

So sort of like our other var ops we need to

1. make an output on the forward pass for the assignment
  - In this case it's assigning `B_v`'s values to that block of `A_v`'s values to make

```
A:
val:
5, 4, 3
2, 1, 0
6, 7, 8

adj:
0, 0, 0
0, 0, 0
0, 0 ,0
```

2. Save the stuff we need in the reverse pass
   - Here it's saving the original values of that block of `A_v`.

So now we call `grad()` and what should happen:

 in the reverse pass, once we have `A_v`'s adjoints from the `other_op`, we know that a block of `A_v` actually comes from `B_v`. And(!) all ops before that block assignment are going to want A's original values and adjoints. So we need to both revert that block of `A_v`'s values and send that block of `A_v`'s adjoints over to `B_v`'s adjoints (since that block of `A_v` and `B_v` are supposed to be one and the same). 


Similarly, before the block assignment, that block of `A_v`'s adjoints were zero (think of the `Matrix<var>` example with multiply). So after we copy `A_v`'s adjoints over to `B_v`'s adjoints we set `A_v`'s adjoints back to zero! So now we've reversed the tape to before the block assignment!

What this doesn't cover, but could, is if we did something to the adjoints in the forward pass. Though idt code that set the adjoints in the forward pass would work terribly well since in things like the ode solver we call `set_zero_adjoints()`

Does that make sense / answer your Q? Sorry if it didn't, it looks like your comment may have gotten cut off halfway?

t4c1: I agree with Ben. This test result is wrong. The reason for it being wrong is that it is not a realistic use case and is therefore (intentionally for performance reasons) not not handled correctly. 

In tests you should always first do all operations that would normally be in forward pass (such as assigning B to a block of A) before any operations that would normally be in the reverse pass (such as modifying adjoints in any way).

So change this test to what Ben suggested.

SteveBronder: LoL oh gosh I had this fixed on my computer but didn't upload it so I was super confused by your comment Ben my bad. Yes what you are saying makes sense and just fixed!

SteveBronder:
Could someone give this a look?

t4c1 (approved):

Issue #2158, opened 2020-10-22
User-reported that grad_2F1 bumps into internal counter limit


sakrejda:
## Description
To avoid having the `grad_2F1` function hang on boundary values, we have an internal step counter in the algorithm with a default of 1e5 steps.  A user [reported](https://discourse.mc-stan.org/t/neg-binomial-2-lcdf-grad-2f1-k-internal-counter-exceeded-100000-iterations/8723) on the forum that their model crashes because of this limit (the function throws when it exceeds the counter, the Stan program fails).  This is the second report according to @martinmodrak 

## Example
This is a failing test:

```
TEST(MathPrimScalFun, grad2F1_11) {                                                                                                           
  double a1 = 3.70975;                                                                                                                    
  double a2 = 1.0;                                                                                                                            
  double b1 = 2.70975;                                                                                                                        
  double z = 0.999696;                                                                                                  
                                                                                                                                         
  double grad_a1;                                                                                                                        
  double grad_b1;                                                                                                                                                               
  stan::math::grad_2F1(grad_a1, grad_b1, a1, a2, b1, z);                                                                                                                
}
```

To make it pass we can up the max_steps argument:

```
TEST(MathPrimScalFun, grad2F1_11) {                                                                                                           
  double a1 = 3.70975;                                                                                                                    
  double a2 = 1.0;                                                                                                                            
  double b1 = 2.70975;                                                                                                                        
  double z = 0.999696;                                                                                                  
                                                                                                                                         
  double grad_a1;                                                                                                                        
  double grad_b1;                                                                                                                                                               
  // pass 1e6 (rather than default 1e5) as the max_steps argument
  stan::math::grad_2F1(grad_a1, grad_b1, a1, a2, b1, z, 1e-10, 1e6);                                                                                                                
}
```
For these specific values the run-time is only 6ms at the higher max_steps (ballpark, on my laptop) so I think it's fine to just up the max counter and make that not fail.  We could make some more principled choices here based on how far to the edges of the distribution the calculation should go and pick a different implementation for those corner cases but that's a slightly larger project.  

The values calculated here are rather large: 

grad_a1: 2.93698e+07, grad_b1: -3.0843e+07     


#### Current Version:
`develop`

sakrejda:
@syclik I didn't call this a "bug" because the math library lets the caller alter `max_steps` but extending the range of values that can be calculated to cover more data/model scenarios would be a good feature from my pov.

bbbales2:
[...]

PR #2175, opened 2020-10-29
Incorporate relative + absolute tolerance into grad_2F1

sakrejda:
## Summary

Fixes issue #2158 by introducing combined absolute/relative tolerance to grad_2F1 and increasing the max number of default steps.  Overall it should terminate sooner for very large gradients, the increased max number of steps is required to get the stated precision in the tests.  New tests are referenced vs. a quick discrete diff in Mathematica.

The absolute/relative tolerance calculation is borrowed from the [unit test formulation of the same problem](https://github.com/stan-dev/math/blob/8464e86fa846d11459c46bc706627583935e7ee3/test/unit/math/relative_tolerance.hpp#L64): 

## Tests

Test case uses [values from a user's failing model](https://discourse.mc-stan.org/t/neg-binomial-2-lcdf-grad-2f1-k-internal-counter-exceeded-100000-iterations/8723/3?u=sakrejda) and reference values from mathematica.  The gradients are in the tens of millions and the original absolute tolerance calculation was excessive there.

## Side Effects

Nothing noticeable.  Most likely improved performance for models that hit the boundaries in neg_binomial_cdf and anything else that relies on grad_2F1.  Fixes a user-reported model bug as a side-effect (linked above).

## Release notes

Make gradients for negative binomial and 2F1 function more robust for boundary values.

## Checklist

- [X] Math issue #2158 

- [X] Copyright holder: Krzysztof Sakrejda


sakrejda:
This works locally for the failing test so if there's a standard way to tickle the test system I'd appreciate any help.

rok-cesnovar:
I am able to replicate locally with clang 10.0 on Ubuntu. Not sure on the solution though.

sakrejda:
@rok-cesnovar thanks for checking the bug happens for you, I'm a little burned out on reading C++ errors at the moment but I'm sure I can figure it out.  I'm on arch linux so it's usually one of the newest compiler versions and going to Ubuntu is like getting in a time machine.

test/unit/math/prim/fun/grad_2F1_test.cpp
```diff
@@ -129,3 +129,19 @@ TEST(MathPrimScalFun, grad2F1_10) {
...
+  double a2 = 1.0;
+  double b1 = 2.70975;
+  double z = 0.999696;
+
+  double grad_a1;
+  double grad_b1;
+  stan::math::grad_2F1(grad_a1, grad_b1, a1, a2, b1, z);
+
+  EXPECT_NEAR(29369830.002773938200417693317785, grad_a1,
+              1e1);  // reference: discrete diff in mathematica
```
bbbales2, line 144: Isn't 10 a big error here? That's like a relative error of `3e-7`

test/unit/math/rev/prob/neg_binomial_2_ccdf_log_test.cpp
```diff
@@ -18,7 +18,7 @@ TEST(ProbDistributionsNegBinomial2, derivatives_lccdf) {
   std::vector<double> gradients;
   val.grad(x, gradients);
 
-  double epsilon = 1e-6;
+  double epsilon = 1e-5;
```
bbbales2, line 21: If this worked before, why doesn't it work now? `1e-6` now is less stringent than previously.

stan/math/prim/fun/grad_2F1.hpp
```diff
@@ -36,12 +36,14 @@ namespace math {
  */
 template <typename T>
 void grad_2F1(T& g_a1, T& g_b1, const T& a1, const T& a2, const T& b1,
```
bbbales2, line 38: I couldn't add a comment above, but the docs are wrong I think:

```
* @param[out] g_a1 g_a1 pointer to array of six values of type T, result.
* @param[out] g_b1 g_b1 pointer to array of six values of type T, result.
```

I think should be

```
* @param[out] g_a1 gradient of hypergeometric function with respect to a1, result
* @param[out] g_b1 gradient of hypergeometric function with respect to b1, result
```

but at least they definitely aren't pointers so should just be:

```
* @param[out] g_a1 result
* @param[out] g_b1 result
```

If we're not sure what they are (the erroneous line appears to be copy pasted from the docs of `grad_F32`: `* @param[out] g g pointer to array of six values of type T, result.`)

bbbales2 (changes requested):
I had a closer look. I'm not sure relative error makes sense here. As far as tests, the thing that seemed to matter is the increased number of steps of the algorithm, not the error test change.

If I print out the error on convergence, I'm always seeing really small numbers which makes me think `log_t_new` is an increment (and so just making sure it is below some value is already a relative error check).

It looks like from the line ```log_t_new += log(fabs(p)) + log_z;``` that `log_t_new` is being incremented to some value, but I think this is actually the log of the terms in the power series of the function and so the terms are getting smaller in the sequence and the log is getting more negative (and so the `log_t_new` accumulation is only negative, and isn't really converging itself to anything, cuz ideally it goes to -inf which is zero error).

The variable `p` at the top of the loop looks like incrementally computing new factors to be included in each term in the series (look at the definition at the bottom of the Notation section here -- I wonder if `p` was for Pochhammer: https://en.wikipedia.org/wiki/Generalized_hypergeometric_function#Notation)

sakrejda:
@bbbales2 just for context: I re-wrote and re-organized the hypergeometric functions and their gradients last time I worked significantly on Stan, so I'm familiar with the calculations and their challenges.  It would take another significant chunk of work to improve them over the current status and that is *not* what I'm trying to address in the PR.

test/unit/math/prim/fun/grad_2F1_test.cpp
sakrejda: Right, but these have *never* been better as far as I know.  This section of the function's range should really be using a different calculation (the gradients just become more and more extreme as `z` approaches 1).  I'd enjoy re-working these to be more complete but I'm not sure it's worth the trade-off unless we show there's some inaccuracy in a model that's based on this issue.

test/unit/math/rev/prob/neg_binomial_2_ccdf_log_test.cpp
sakrejda: The epsilon here is the interval the discrete diff is calculated over.  This is a very flat region of the function so I don't think it's necessarily less stringent.  I am concerned about this being wrong and I am would prefer to include a different test of the negative binomial that didn't get into the discrete diff.  I can probably get a good reference value from R or mathematica.

stan/math/prim/fun/grad_2F1.hpp
```diff
@@ -85,8 +88,9 @@ void grad_2F1(T& g_a1, T& g_b1, const T& a1, const T& a2, const T& b1,
     g_a1 += log_g_old_sign[0] > 0 ? exp(log_g_old[0]) : -exp(log_g_old[0]);
     g_b1 += log_g_old_sign[1] > 0 ? exp(log_g_old[1]) : -exp(log_g_old[1]);
 
-    if (log_t_new <= log(precision)) {
-      return;  // implicit abs
+    if (log_t_new
+        <= std::max(value_of(log_t_new) + log_precision, log_precision)) {
```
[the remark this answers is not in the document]
sakrejda: Thanks! I wasn't quite sure about these!

stan/math/prim/fun/grad_2F1.hpp
sakrejda: Thanks for pointing that out, I didn't touch those this time but I'll update it.

bbbales2:
Yeah that's fine. I don't think we need any major work here.

I'm happy just multiplying max_steps by 10 and calling it a day.

Since the relative tolerance was new, I tried to verify that tit was behaving well at the switch and I wasn't able to do that. The tests I ran (`./runTests.py test/unit/math/mix/fun/grad_2F1_test.cpp` and `./runTests.py test/unit/math/rev/prob/neg_binomial_2_cdf_log_test.cpp`) both showed that the `log_t_new` thing is going negative (so `exp(log_t_new)` is going to zero), and so I went and read about the power series and whatnot to try to piece together what was happening. My hunch is the tolerance here is already effectively a relative tolerance.

test/unit/math/rev/prob/neg_binomial_2_ccdf_log_test.cpp
bbbales2: Oh lol, yes, this is the finite difference.

test/unit/math/prim/fun/grad_2F1_test.cpp
bbbales2: Makes sense

sakrejda:
oh this is definitely not worth your time to try and figure out the hypergeometric function notation!  It is different everywhere and I standardized our notation across the different functions last time I touched them but the reference are still weird!  Your comments are the nudge I needed to look at the parts that were bothering me yesterday so thanks!

You got it right that log_t_new is based on the increment (it's a lagged part of each new term in the sum) so I need to actually include `g_a1` and `g_b1` in the scaling rather than `log_t_new`.  It should still work since each update to those is incrementally smaller (there's a separate function that checks the conditions for the infinite sum converging).

test/unit/math/rev/prob/neg_binomial_2_ccdf_log_test.cpp
sakrejda: Turns out that with the other fixes I can revert this change.  It's still bizarre.

sakrejda:
Thanks again for the review, it really helps especially with these obscure calculations... even if I'm the one who re-wrote them... :/

sakrejda:
Looks like there were some failures from the test system that aren't code-related so I upped precision in one test and kicked it off again...

sakrejda:
BTW: upshot is that bumping the max_steps to 1e6 is still required to get the boundary value down right so the PR is adding relative tolerance to avoid excessive computation and also bumping max_steps to avoid a failure in Stan models.  Downstream tests failed (out of memory) so I re-started those.

sakrejda:
Upstream tests fails again, anybody know what to do with this?  It's been too long for me.

```
virtual memory exhausted: Cannot allocate memory
make/tests:158: recipe for target 'test/test-models/good/function-signatures/math/matrix/dims.hpp-test' failed
make: *** [test/test-models/good/function-signatures/math/matrix/dims.hpp-test] Error 1
```

rok-cesnovar:
That is an issue with upstream tests that we are looking into. Will make sure to restart these tests once the problem is fixed. Sorr for the inconvenience.

sakrejda:
Wow those ARMA/GARCH performance results

bbbales2:
Got it on my list for tomorrow!

bbbales2:
I think these convergence checks should be on `g_a1`/`g_b1` directly.

I tried out:

```cpp
    T g_a1_prev = g_a1;
    T g_b1_prev = g_b1;

    g_a1 += log_g_old_sign[0] > 0 ? exp(log_g_old[0]) : -exp(log_g_old[0]);
    g_b1 += log_g_old_sign[1] > 0 ? exp(log_g_old[1]) : -exp(log_g_old[1]);

    if (std::abs(value_of_rec(g_a1) - value_of_rec(g_a1_prev)) <=
        std::max(std::abs(value_of_rec(g_a1)) * precision, precision) &&
        std::abs(value_of_rec(g_b1) - value_of_rec(g_b1_prev)) <=
        std::max(std::abs(value_of_rec(g_b1)) * precision, precision)) {
      std::cout << "k: " << k << std::endl;
      std::cout << "log(fabs(term)): " << log(fabs(term)) << std::endl;
      std::cout << "log_t_new: " << value_of(log_t_new) << std::endl;
      std::cout << "log_g_old[0]: " << value_of(log_g_old[0]) << std::endl;
      std::cout << "log_g_old[1]: " << value_of(log_g_old[1]) << std::endl;
      std::cout << "g_old[0]: " << exp(value_of(log_g_old[0])) << std::endl;
      std::cout << "g_old[1]: " << exp(value_of(log_g_old[1])) << std::endl;
      std::cout << "abs(g_a1) * precision: " << std::setprecision(17) << std::abs(value_of(g_a1)) * precision << std::endl;
      std::cout << "abs(g_b1) * precision: " << std::setprecision(17) << std::abs(value_of(g_b1)) * precision << std::endl;
      std::cout << "precision: " << exp(log_precision) << std::endl;
      std::cout << "g_a1: " << std::setprecision(17) << value_of(g_a1) << std::endl;
      std::cout << "g_b1: " << std::setprecision(17) << value_of(g_b1) << std::endl;
      return;
    }
```

Running `./runTests.py test/unit/math/prim/fun/grad_2F1_test.cpp` and looking at the last test output I get:
```
k: 141758
log(fabs(term)): 2.4034041754828013
log_t_new: -32.236241280788448
log_g_old[0]: -29.866770317230252
log_g_old[1]: -29.832837105305646
g_old[0]: 1.0691200123013884e-13
g_old[1]: 1.1060212359754558e-13
abs(g_a1) * precision: 2.9369829991912055e-07
abs(g_b1) * precision: 3.0843032119093972e-07
precision: 9.9999999999999874e-15
g_a1: 29369829.991912056
g_b1: -30843032.119093969
```

The problem is that the thing converging here is `g_a1` and `g_b1` but the checks were that `log_t_new` gets under some threshold (`t_new < 1e-14`).

If I change to the checks above I get:
```
k: 91430
log(fabs(term)): 2.3629489643518515
log_t_new: -17.372736574697267
log_g_old[0]: -15.04514642131258
log_g_old[1]: -15.009787610345416
g_old[0]: 2.923990315021038e-07
g_old[1]: 3.0292287240870817e-07
abs(g_a1) * precision: 2.9369829990916807e-07
abs(g_b1) * precision: 3.0843032118062863e-07
precision: 9.9999999999999874e-15
g_a1: 29369829.990916807
g_b1: -30843032.118062865
```

In this case the gradients are off in the 11th digit, but we're stopping the calculation way sooner (in this case when we're incrementing numbers smaller than `abs(g_a1) * precision` and `abs(g_b1) * precision`).

sakrejda:
I think I agree, I doubt I can get to this today but I'll do it tomorrow (if the world isn't on fire)

bbbales2:
Just to bump this along I went ahead and made the changes. Have a look and revert them or change them if they're not good.

stan/math/prim/fun/grad_2F1.hpp
```diff
@@ -86,13 +86,16 @@ void grad_2F1(T& g_a1, T& g_b1, const T& a1, const T& a2, const T& b1,
     log_g_old[1] = log_t_new + log(fabs(term));
     log_g_old_sign[1] = term >= 0.0 ? log_t_new_sign : -log_t_new_sign;
 
+    T g_a1_prev = g_a1;
```
sakrejda, line 89: @bbbales2 thanks for prodding me along, shouldn't we be storing these as just doubles? Is there any performance penalty involved in storing the type?

bbbales2: Yeah I guess since we're just looking at the values we could do `value_of_rec`. Will push that up in a sec.

sakrejda:
So... that increment you calculate as the convergence condition, that *is* `exp(log_g_old)` (well, with the sign, but the magnitude is what should shrink) so I think the original condition on `log_g_old` was fine (and fewer calculations).

bbbales2:
@sakrejda that makes sense. I switched it to logs. The convergence numbers to compare to [above](https://github.com/stan-dev/math/pull/2175#issuecomment-720466228) are:

```
k: 91415
log(fabs(term)): 2.3629335185457978
log_t_new: -17.368339948175109
log_g_old[0]: -15.040765796341576
log_g_old[1]: -15.005406429629311
g_old[0]: 2.9368273165263018e-07
g_old[1]: 3.0425294376974053e-07
abs(g_a1) * precision: 2.9369829990912397e-07
abs(g_b1) * precision: 3.0843032118058315e-07
precision: 9.9999999999999874e-15
g_a1: 29369829.990912396
g_b1: -30843032.118058313
```

sakrejda:
> @sakrejda that makes sense. I switched it to logs. The convergence numbers to compare to [above](https://github.com/stan-dev/math/pull/2175#issuecomment-720466228) are:
> 
> ```
> k: 91415
> log(fabs(term)): 2.3629335185457978
> log_t_new: -17.368339948175109
> log_g_old[0]: -15.040765796341576
> log_g_old[1]: -15.005406429629311
> g_old[0]: 2.9368273165263018e-07
> g_old[1]: 3.0425294376974053e-07
> abs(g_a1) * precision: 2.9369829990912397e-07
> abs(g_b1) * precision: 3.0843032118058315e-07
> precision: 9.9999999999999874e-15
> g_a1: 29369829.990912396
> g_b1: -30843032.118058313
> ```

I think this is a third thing, I should have time later today to update this to what I think it should be and I'll ask you to check it then, ok? I think we're getting there, it's just a weird calculation...

bbbales2:
> I'll ask you to check it then, ok?

Sure! Btw I don't think the red X is anything to worry about. Sounds like Jenkins got sad over the weekend. It'll probably fix itself with the next push you make.

rok-cesnovar:
I would also merge recent develop in. There were a couple of fixes in the last few days for the random false-positive failures, like https://github.com/stan-dev/math/pull/2187.

sakrejda:
@bbbales2 you have a "change request" pending on this but I just read the current code again and I think it *does* look right now.  The tests are passing so I think this is good to merge.  Can you take a look again and I'd love to get it merged.

bbbales2 (approved):

bbbales2:
@sakrejda in!

PR #2179, opened 2020-10-31
Split expressions test in more files

### LINKED ISSUE #2178: Develop tests issues
Currently, we are having issues in the Math upstream tests, specifically with the Linux integration tests and expressions tests. 

Linux integration tests: 
- The issues arise from the memory footprint of compilation. We run those test with `-j 12` or `-j 18`, meaning that 18 model compilations are run parallel and the increased memory footprint of compilation is causing these machines to run out of memory right now. Its mostly having issues with larger test models (with 2k distribution calls). Example: https://github.com/stan-dev/stan/blob/develop/src/test/test-models/good/function-signatures/distributions/univariate/continuous/pareto_type_2/pareto_type_2_ccdf_log_4.stan This issue will be fixed by splitting these large test models.

Expression tests: 
- Since most of the functions now support expressions the expressions tests are also getting pretty large and the memory footprint is getting large. The problem there is that we split expression tests that are run with `-j X` in X files, but that should probably be 2X or 3X. This will have to be fixed in Math. 

It's very likely that any PR or develop tests are not going to pass until this is fixed. I am addressing the first issue right now.

rok-cesnovar:
## Summary

Fixes #2178 by splitting expressions test into more files

Right now if you run

runTests.py test/expressions -jX

it will split expressions tests into 2X files. This changes the default to 4X and exposes a flag so we can also set an arbitrary number in the cmd line and thus also Jenkins.


t4c1 (dismissed):

rok-cesnovar:
Together with https://github.com/stan-dev/stan/pull/2975 this brings expression tests execution time under 10minutes. https://jenkins.mc-stan.org/blue/organizations/jenkins/Stan/detail/PR-2975/2/pipeline/171

t4c1 (approved):

PR #2190, opened 2020-11-12
Let prim functions return expressions

t4c1:
## Summary

I went over all prim functions and modified them to return Eigen expressions, where it makes sense to do so.

<s>I also added a new helper function `to_forwarding_ref`, which is used, where we need to store equivalent of `to_ref` in a `holder`. It works the same as `to_ref`, except it does perfect forwarding when output does not need to be converted to plain type. In other words it can return rvalue reference to input if the input does not need to be converted.</s>

Improved `make_holder`, so that holder is not used if the given functor returns plain type or if all arguments are of types, which Eigen passes around by value in expressions.

## Tests
All changes should be covered by expression testing framework. I am hoping nothing goes up in flame.

## Side Effects

Like ... this whole PR? We never had exposed functions returning expressions. 

## Release notes

Prim functions now return expressions where that makes sense.

## Checklist

- [ ] Math issue #1470 


stan/math/prim/fun/quad_form.hpp
```diff
@@ -28,13 +28,12 @@ template <typename EigMat1, typename EigMat2,
...
-                     EigMat2::ColsAtCompileTime>
-quad_form(const EigMat1& A, const EigMat2& B) {
+inline auto quad_form(EigMat1&& A, EigMat2&& B) {
   check_square("quad_form", "A", A);
   check_multiplicable("quad_form", "A", A, "B", B);
-  const auto& B_ref = to_ref(B);
-  return B_ref.transpose() * A * B_ref;
+  return make_holder(
+      [](const auto& a, const auto& b) { return b.transpose() * a * b; },
+      std::forward<EigMat1>(A), to_forwarding_ref(std::forward<EigMat2>(B)));
```
bbbales2, line 36: Why the regular fwd on the left?

stan/math/prim/fun/quad_form_sym.hpp
```diff
@@ -32,8 +32,8 @@ inline plain_type_t<EigMat2> quad_form_sym(const EigMat1& A, const EigMat2& B) {
   const auto& A_ref = to_ref(A);
   const auto& B_ref = to_ref(B);
   check_symmetric("quad_form_sym", "A", A_ref);
-  plain_type_t<EigMat2> ret(B_ref.transpose() * A_ref * B_ref);
-  return value_type_t<EigMat2>(0.5) * (ret + ret.transpose());
+  auto* ret = new plain_type_t<EigMat2>(B_ref.transpose() * A_ref * B_ref);
+  return holder(value_type_t<EigMat2>(0.5) * (*ret + ret->transpose()), ret);
```
bbbales2, line 36: Why the pointer? Will holder know to delete this?

stan/math/prim/fun/to_ref.hpp
```diff
@@ -9,6 +9,10 @@ namespace math {
 /**
  * This evaluates expensive Eigen expressions. If given expression involves no
  * calculations this is a no-op that should be optimized away.
+ *
+ * When used to pass the returned value into another function that handles
+ * rvalues and lvalues differently `to_forwarding_ref` should be used instead.
```
bbbales2, line 14: Does it make sense to do the `to_forwarding_ref` behavior everywhere and just do `to_ref`?

stan/math/prim/fun/unit_vector_constrain.hpp
```diff
@@ -24,13 +24,16 @@ namespace math {
...
-  value_type_t<T> SN = dot_self(y_ref);
-  check_positive_finite("unit_vector_constrain", "norm", SN);
-  return (y_ref / sqrt(SN)).eval();
+  return make_holder(
+      [](const auto& y_ref) {
+        value_type_t<T> SN = dot_self(y_ref);
+        check_positive_finite("unit_vector_constrain", "norm", SN);
+        return y_ref / sqrt(SN);
+      },
+      to_forwarding_ref(std::forward<T>(y)));
```
bbbales2, line 36: Why forward into the forwarding ref?

stan/math/prim/fun/to_ref.hpp
bbbales2: What I mean is to make `to_ref` do what `to_forwarding_ref` is doing

stan/math/prim/fun/quad_form.hpp
t4c1: Holder needs to know if an input is rvalue to know whether to save it on heap. Actually ... given that we decided not to use holder for arguments we could revert that to nod do forwarding.

stan/math/prim/fun/quad_form_sym.hpp
t4c1: That is the main and only reason for using holder at all!

stan/math/prim/fun/to_ref.hpp
t4c1: No. `to_forwarding_ref` is not safe to use in cases such as `const auto& x = to_ref(...)`, as it is returning reference to an object that will be destructed after that line is executed.

stan/math/prim/fun/unit_vector_constrain.hpp
t4c1: If y is a rvalue reference to plain type, `to_forwarding_ref` returns a rvalue reference to y. To do that it needs to know whether y is a rvalue or lvalue. So we need `forward`.

t4c1: Actually ... no, given that we decided input arguments do not need holders that is not needed.

stan/math/prim/fun/quad_form_sym.hpp
t4c1: Alternatively I could call make_holder, which would do this allocation.

stan/math/prim/fun/to_ref.hpp
t4c1: Actually with removal of forwarding from other places we don't need `to_forwarding_ref `.

SteveBronder:
Weird include error on poisson posted below

```
In file included from test/prob/poisson/poisson_00000_generated_ffd_test.cpp:3:
In file included from ./test/prob/test_fixture_distr.hpp:4:
In file included from ./stan/math/mix.hpp:6:
In file included from ./stan/math/prim.hpp:12:
In file included from ./stan/math/prim/prob.hpp:276:
In file included from ./stan/math/prim/prob/poisson_log.hpp:5:
./stan/math/prim/prob/poisson_lpmf.hpp:68:28: error: call to function 'sum' that is neither visible in the template definition nor found by argument-dependent lookup
  T_partials_return logp = sum(multiply_log(n_val, lambda_val));
                           ^
./stan/math/prim/prob/poisson_lpmf.hpp:85:10: note: in instantiation of function template specialization 'stan::math::poisson_lpmf<false, std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> > >' requested here
  return poisson_lpmf<false>(n, lambda);
         ^
./stan/math/prim/prob/poisson_log.hpp:23:10: note: in instantiation of function template specialization 'stan::math::poisson_lpmf<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> > >' requested here
  return poisson_lpmf<T_n, T_rate>(n, lambda);
         ^
./test/prob/poisson/poisson_test.hpp:49:24: note: in instantiation of function template specialization 'stan::math::poisson_log<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> > >' requested here
    return stan::math::poisson_log(n, lambda);
                       ^
./test/prob/test_fixture_distr.hpp:77:26: note: in instantiation of function template specialization 'AgradDistributionsPoisson::log_prob<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> >, empty, empty, empty, empty>' requested here
      TestClass.template log_prob<T0, T1, T2, T3, T4, T5>(p0, p1, p2, p3, p4,
                         ^
./test/prob/test_fixture_distr.hpp:822:9: note: in instantiation of member function 'AgradDistributionTestFixture<boost::mpl::vector<AgradDistributionsPoisson, boost::mpl::vector<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> >, empty, empty, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na> >::call_all_versions' requested here
  this->call_all_versions();
        ^
lib/benchmark_1.5.1/googletest/googletest/include/gtest/internal/gtest-internal.h:472:44: note: in instantiation of member function 'gtest_suite_AgradDistributionTestFixture_::CallAllVersions<boost::mpl::vector<AgradDistributionsPoisson, boost::mpl::vector<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> >, empty, empty, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na> >::TestBody' requested here
  Test* CreateTest() override { return new TestClass; }
                                           ^
lib/benchmark_1.5.1/googletest/googletest/include/gtest/internal/gtest-internal.h:730:13: note: in instantiation of member function 'testing::internal::TestFactoryImpl<gtest_suite_AgradDistributionTestFixture_::CallAllVersions<boost::mpl::vector<AgradDistributionsPoisson, boost::mpl::vector<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> >, empty, empty, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na> > >::CreateTest' requested here
        new TestFactoryImpl<TestClass>);
            ^
lib/benchmark_1.5.1/googletest/googletest/include/gtest/internal/gtest-internal.h:789:50: note: in instantiation of member function 'testing::internal::TypeParameterizedTest<AgradDistributionTestFixture, testing::internal::TemplateSel<CallAllVersions>, testing::internal::Types<boost::mpl::vector<AgradDistributionsPoisson, boost::mpl::vector<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> >, empty, empty, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>> >::Register' requested here
    TypeParameterizedTest<Fixture, Head, Types>::Register(
                                                 ^
test/prob/poisson/poisson_00000_generated_ffd_test.cpp:32:1: note: in instantiation of member function 'testing::internal::TypeParameterizedTestSuite<AgradDistributionTestFixture, testing::internal::Templates<CallAllVersions, ValidValues, InvalidValues, Propto, FiniteDiff, Function, RepeatAsVector, AsScalarsVsAsVector, Length0Vector>, testing::internal::Types<boost::mpl::vector<AgradDistributionsPoisson, boost::mpl::vector<std::vector<int, std::allocator<int> >, stan::math::fvar<stan::math::fvar<double> >, empty, empty, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>> >::Register' requested here
INSTANTIATE_TYPED_TEST_SUITE_P(AgradDistributionsPoisson_ffd_3, AgradDistributionTestFixture, AgradDistributionsPoisson_ffd_3);
^
lib/benchmark_1.5.1/googletest/googletest/include/gtest/gtest-typed-test.h:317:11: note: expanded from macro 'INSTANTIATE_TYPED_TEST_SUITE_P'
          Register(GTEST_STRINGIFY_(Prefix),                                \
          ^
./stan/math/fwd/fun/sum.hpp:43:24: note: 'sum' should be declared prior to the call site or in an associated namespace of one of its arguments
inline value_type_t<T> sum(const T& m) {                   ^
```

It looks like this is a clang specific error, but calling `stan::math::sum()` explicitly ala

```cpp
  T_partials_return logp = stan::math::sum(multiply_log(n_val, lambda_val));
```

also gives an error that it can't find the correct sum function (it want's the `fvar<>` eigen version) [gist](https://gist.github.com/SteveBronder/baa4f55850662b266f4d9c36c9fd29fe)). Though looking at the candidates it's not even finding the `fvar` version so I suspect this is something to do with a nasty include order. I think this is happening because `prim/prob.hpp` is being loaded too soon and I think I found the footgun in `rev/fun/multiply.hpp` which does 

```
#include <stan/math/prim.hpp>
```

and that should be 

```
// Or whatever functions it specifically needs but not /prob.hpp!
#include <stan/math/prim/fun.hpp>
```

I'm pretty sure I did that so my b. I fixed it over here, though I think just doing the above could work.

https://github.com/stan-dev/math/compare/fix/include-stuff-for-eigen-exprs

t4c1:
Thanks, that seems to work

SteveBronder:
For this

```
lib/stan_math/lib/eigen_3.3.7/Eigen/src/Core/util/StaticAssert.h:33:40: error: static assertion failed: THIS_COEFFICIENT_ACCESSOR_TAKING_ONE_ACCESS_IS_ONLY_FOR_EXPRESSIONS_ALLOWING_LINEAR_ACCESS
```

Do we need to make Maps have a stride of `<0,0>`?

SteveBronder:
It might be worth just doing the change for multiply. I'm not sure if we absolutely need the other changes

t4c1:
Hmm, we need to either require all returned expressions support linear access (Than all maps need to have strode <0,0> and we can not return blocks) or have no function require linear access on its matrix arguments. I am leaning towards the second option.

t4c1:
We will need a number of changes so I opened a separate PR: https://github.com/stan-dev/math/pull/2205.

t4c1:
There were only a few rev and no fwd functions benefiting from returning expressions, so I just put everything in this PR. For rev functions it is mostly returning `arena_matrix` instead of converting to `Eigen::Matrix`.

bbbales2:
> it is mostly returning arena_matrix instead of converting to Eigen::Matrix

We can't do that because it's not okay to change the varis in the `arena_matrix`. This has come up before. I'm not quite sure how to test this. It needs tested for all functions so it should be in the test framework probably.

Something like this should print 2.0 instead of 1.0:

```cpp
arena_t<Eigen::Matrix<var, -1, -1>> res(1, 1);
res(1, 1) = 1.0;
reverse_pass_callback([res]() mutable {
  std::cout << res.val() << std::endl;
});

# function returns here

# modify return value
res(1, 1) = 2.0;
stan::math::grad();
```

Can do a similar thing with adjoints.

t4c1:
Ouch. I will reset this to just prim functions. For rev functions I have an idea how to partially work around this limitation, but I will put it in a separate PR if/when I finish it.

SteveBronder:
The rev version is a weird thing because _technically_ a stan program will never actually change these return values (we assign to an Eigen::Matrix so a full copy happens and all stan math functions treat input as const). But it makes for a bit of a gotcha when working with stan math itself since then changing an `arena_matrix` after it's used as an input would then change the reverse pass.

But that's also kind of neat in another way where if you set the adjoints to zero and update the inputs you can run grad without allocating new mem.

t4c1:
Ready for review!

bbbales2:
I'll do a review in the afternoon.

stan/math/fwd.hpp
```diff
@@ -5,9 +5,9 @@
 
 #include <stan/math/fwd/core.hpp>
 #include <stan/math/fwd/meta.hpp>
-#include <stan/math/prim.hpp>
-
 #include <stan/math/fwd/fun.hpp>
 #include <stan/math/fwd/functor.hpp>
 
+#include <stan/math/prim.hpp>
```
bbbales2, line 11: This seems pretty innocuous but why'd it move?

stan/math/mix.hpp
```diff
@@ -2,11 +2,19 @@
...
+#include <stan/math/rev/meta.hpp>
+#include <stan/math/rev/fun.hpp>
+#include <stan/math/rev/functor.hpp>
+
+#include <stan/math/fwd/core.hpp>
+#include <stan/math/fwd/meta.hpp>
+#include <stan/math/fwd/fun.hpp>
+#include <stan/math/fwd/functor.hpp>
+
+#include <stan/math/prim.hpp>
```
bbbales2, line 18: Ahh, there must be more to the story. Do tell!

stan/math/prim/meta/holder.hpp
```diff
@@ -321,21 +339,39 @@ auto make_holder_impl(const F& func, std::index_sequence<Is...>,
...
+ * @param func the functor
+ * @param args arguments for the functor
+ * @return `holder` referencing expression constructed by given functor
+ */
+template <typename F, typename... Args,
+          require_plain_type_t<
+              decltype(std::declval<F>()(std::declval<Args&>()...))>* = nullptr>
+auto make_holder(const F& func, Args&&... args) {
+  return func(std::forward<Args>(args)...);
+}
```
bbbales2, line 373: I'll take another pass of this file specifically after you grade my holder knowledge.

stan/math/prim/functor/apply_vector_unary.hpp
```diff
@@ -151,13 +190,30 @@ struct apply_vector_unary<T, require_std_vector_vt<is_container, T>> {
...
+   * @tparam T Type of argument to which functor is applied.
+   * @tparam F Type of functor to apply.
+   * @param x std::vector of containers to which operation is applied.
+   * @param f functor to apply to vector input.
+   * @return std::vector of containers with result of applying functor to
+   *         input.
+   */
+  template <typename F>
+  static inline auto apply_no_holder(const T& x, const F& f) {
+    return apply(x, f);
```
bbbales2, line 214: Does this get used anywhere?

bbbales2 (changes requested):
Bunch o' comments and questions. Nothing big. Code looks good.

stan/math/fwd.hpp
t4c1: If a prim function is used for fwd autodiff and that prim function calls another function that has fwd implementation, we need to include that fwd implementation before the prim function. Since no prim functions include fwd stuff, it is the easiest to include all fwd stuff before any prim stuff.

stan/math/mix.hpp
t4c1: Same for rev.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -142,8 +142,9 @@ inline void elementwise_check(const F& is_good, const char* function,
  */
 template <typename F, typename T, typename... Indexings,
           require_eigen_t<T>* = nullptr,
-          std::enable_if_t<static_cast<bool>(Eigen::internal::traits<T>::Flags&(
-              Eigen::LinearAccessBit | Eigen::DirectAccessBit))>* = nullptr>
+          std::enable_if_t<(Eigen::internal::traits<T>::Flags
+                            & Eigen::LinearAccessBit)
+                           || T::IsVectorAtCompileTime>* = nullptr>
```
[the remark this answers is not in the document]
t4c1: I don't know if such specifics are documented anywhere by Eigen, so it is hard to be 100% sure. 

As far as I know the answer is no. All strided and non-strided vectors support linear access. Only matrix expressions, such as block or identity matrix or transpose do not support linear access.

stan/math/prim/fun/log_softmax.hpp
```diff
@@ -37,18 +38,16 @@ namespace math {
...
-    check_nonzero_size("log_softmax", "v", v_ref);
-    return (v_ref.array() - log_sum_exp(v_ref)).eval();
-  });
+  check_nonzero_size("log_softmax", "v", x);
+  return make_holder(
+      [](const auto& a) {
+        return apply_vector_unary<ref_type_t<Container>>::apply(
+            a, [](const auto& v) { return v.array() - log_sum_exp(v); });
+      },
+      to_ref(x));
```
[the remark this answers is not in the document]
t4c1: Partially true. This is not for x (we decided not to use holder for rvalue arguments), but for `to_ref(x)`, which might construct a new matrix, if x is non-trivial expression.

stan/math/prim/fun/quad_form.hpp
```diff
@@ -28,13 +28,11 @@ template <typename EigMat1, typename EigMat2,
...
-inline Eigen::Matrix<value_type_t<EigMat2>, EigMat2::ColsAtCompileTime,
-                     EigMat2::ColsAtCompileTime>
-quad_form(const EigMat1& A, const EigMat2& B) {
+inline auto quad_form(const EigMat1& A, const EigMat2& B) {
   check_square("quad_form", "A", A);
   check_multiplicable("quad_form", "A", A, "B", B);
-  const auto& B_ref = to_ref(B);
-  return B_ref.transpose() * A * B_ref;
+  return make_holder([&A](const auto& b) { return b.transpose() * A * b; },
+                     to_ref(B));
```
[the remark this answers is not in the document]
t4c1: `B` is used twice so it needs to be `to_ref`-ed. `A` is only used once so no `to_ref` is necessary. Result of `to_ref` is what we store in holder, not `B`.

stan/math/prim/fun/quad_form_diag.hpp
```diff
@@ -10,14 +10,13 @@ namespace math {
...
-quad_form_diag(const EigMat& mat, const EigVec& vec) {
+inline auto quad_form_diag(const EigMat& mat, const EigVec& vec) {
   check_square("quad_form_diag", "mat", mat);
   check_size_match("quad_form_diag", "rows of mat", mat.rows(), "size of vec",
                    vec.size());
-  const auto& vec_ref = to_ref(vec);
-  return vec_ref.asDiagonal() * mat * vec_ref.asDiagonal();
+  return make_holder(
+      [&mat](const auto& v) { return v.asDiagonal() * mat * v.asDiagonal(); },
+      to_ref(vec));
```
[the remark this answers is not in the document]
t4c1: We decided not to use rvalue arguments in holders (except for some internal functions), as these expressions will always be evaluated on the same line of code as they are constructed on, in code generated by stanc.

stan/math/prim/fun/read_cov_L.hpp
```diff
@@ -32,7 +32,9 @@ Eigen::Matrix<value_type<T_CPCs>, Eigen::Dynamic, Eigen::Dynamic> read_cov_L(
   size_t K = sds.rows();
   // adjust due to transformation from correlations to covariances
   log_prob += (sum(log(sds)) + LOG_TWO) * K;
-  return sds.matrix().asDiagonal() * read_corr_L(CPCs, K, log_prob);
+  return make_holder(
+      [&sds](const auto& b) { return sds.matrix().asDiagonal() * b; },
+      read_corr_L(CPCs, K, log_prob));
```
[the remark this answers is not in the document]
t4c1: correct

stan/math/prim/fun/unit_vector_constrain.hpp
```diff
@@ -48,12 +51,15 @@ template <typename T1, typename T2, require_eigen_col_vector_t<T1>* = nullptr,
...
   using std::sqrt;
-  const auto& y_ref = to_ref(y);
-  check_nonzero_size("unit_vector_constrain", "y", y_ref);
-  value_type_t<T1> SN = dot_self(y_ref);
-  check_positive_finite("unit_vector_constrain", "norm", SN);
-  lp -= 0.5 * SN;
-  return (y_ref / sqrt(SN)).eval();
+  check_nonzero_size("unit_vector_constrain", "y", y);
+  return make_holder(
+      [&lp](const auto& y_ref) {
```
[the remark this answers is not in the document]
t4c1: yes, it is a stack variable, but we are modifying it. If we make a copy the change will not be visible to the caller of the `unit_vector_constrain`. So we must assume it will not go out of scope untill the returned expression is evaled.

stan/math/prim/fun/unitspaced_array.hpp
```diff
@@ -25,9 +25,10 @@ inline std::vector<int> unitspaced_array(int low, int high) {
   check_greater_or_equal(function, "high", high, low);
 
   int K = fabs(high - low + 1);
```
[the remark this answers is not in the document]
t4c1: fixed

stan/math/prim/functor/apply_scalar_unary.hpp
```diff
@@ -61,19 +62,16 @@ struct apply_scalar_unary<F, T, require_eigen_t<T>> {
...
+        [](scalar_t x) { return apply_scalar_unary<F, scalar_t>::apply(x); });
   }
 
   /**
    * Return type for applying the function elementwise to a matrix
    * expression template of type T.
    */
   using return_t = std::decay_t<decltype(
-      apply_scalar_unary<F, T>::apply(std::declval<T>()).eval())>;
+      apply_scalar_unary<F, T>::apply(std::declval<T>()))>;
```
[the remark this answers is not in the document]
t4c1: The point of this PR is returning expressions, not plain types.

stan/math/prim/functor/apply_vector_unary.hpp
```diff
@@ -33,8 +32,7 @@ template <typename T>
 struct apply_vector_unary<T, require_eigen_t<T>> {
   /**
    * Member function for applying a functor to a vector and subsequently
-   * returning a vector. The returned objects are evaluated so that
-   * expression templates are not propagated.
+   * returning a vector.
    *
    * TODO(Andrew) Remove .eval() when rest of Stan library can take
```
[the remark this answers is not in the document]
t4c1: Aand .. it left.

stan/math/rev/functor/apply_vector_unary.hpp
```diff
@@ -29,6 +29,21 @@ struct apply_vector_unary<T, require_var_matrix_t<T>> {
...
   }
 
+  /**
+   * Member function for applying a functor to a `var_value<T>` and
+   * subsequently returning a `var_value<T>`.
+   *
+   * @tparam T Type of argument to which functor is applied.
+   * @tparam F Type of functor to apply.
+   * @param x input to which operation is applied.
+   * @param f functor to apply to Eigen input.
```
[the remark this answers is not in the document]
t4c1: Actually ... it is an Eigen object - just look at the require.

test/unit/math/serializer.hpp
```diff
@@ -231,10 +231,11 @@ struct serializer {
...
    * @param x Eigen container to serialize.
    */
-  template <typename U, int R, int C>
-  void write(const Eigen::Matrix<U, R, C>& x) {
-    for (int i = 0; i < x.size(); ++i)
-      write(x(i));
+  template <typename U, require_eigen_t<U>* = nullptr>
+  void write(const U& x) {
+    for (int i = 0; i < x.size(); ++i) {
+      write(x.coeff(i));
```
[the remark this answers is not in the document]
t4c1: test_ad, evals the results, before passing them in here. So anything that comes in here is a plain matrix, which supports linear indexing, even if the tested function returns an expression that does not.

stan/math/prim/fun/softmax.hpp
```diff
@@ -50,8 +50,9 @@ inline plain_type_t<ColVec> softmax(const ColVec& v) {
     return v;
   }
   const auto& v_ref = to_ref(v);
-  plain_type_t<ColVec> theta = (v_ref.array() - v_ref.maxCoeff()).exp();
-  return theta.array() / theta.sum();
+  return make_holder(
+      [](const auto& theta) { return theta.array() / theta.sum(); },
+      (v_ref.array() - v_ref.maxCoeff()).exp().eval());
```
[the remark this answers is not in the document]
t4c1: exactly

stan/math/prim/fun/quad_form_sym.hpp
t4c1: `make_holder` needs to move the `ret`, while allocating directly does not. But the move is not that expensive and `make_holder` has a bit nicer syntax. Would you prefer if we used `make_holder` everywhere?

stan/math/prim/meta/holder.hpp
t4c1: You pass, so so you can make another pass.

stan/math/prim/functor/apply_vector_unary.hpp
t4c1: Yes, in `std::vector` specialization of `apply_vector_unary`.

stan/math/rev/functor/apply_vector_unary.hpp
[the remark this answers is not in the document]
bbbales2: Oh lol

stan/math/prim/functor/apply_scalar_unary.hpp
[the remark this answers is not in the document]
bbbales2: Fair point

stan/math/prim/functor/apply_scalar_unary.hpp
```diff
@@ -143,8 +141,8 @@ struct apply_scalar_unary<F, std::vector<T>> {
    * Return type, which is calculated recursively as a standard
    * vector of the return type of the contained type T.
    */
-  using return_t =
-      typename std::vector<typename apply_scalar_unary<F, T>::return_t>;
+  using return_t = typename std::vector<
+      plain_type_t<typename apply_scalar_unary<F, T>::return_t>>;
```
bbbales2, line 145: Is it true we'd never want to return expressions here? It makes sense just clarifying.

stan/math/prim/fun/unit_vector_constrain.hpp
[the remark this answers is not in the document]
bbbales2: Oh wow, yeah, these functions are a little weird.

stan/math/prim/fun/quad_form_sym.hpp
bbbales2: Yeah let's `make_holder` to keep the pattern simple if it's not a big overhead or anything.

stan/math/mix.hpp
bbbales2: And why the individual files?

stan/math/fwd.hpp
bbbales2: This makes sense to me. I just moved a header today cuz of this I think.

stan/math/mix.hpp
bbbales2: @syclik do you have any memories of why prim was originally included before `rev` and `fwd`? `prim` is the most general, so it makes sense to have it last here (see also [this](https://github.com/stan-dev/math/pull/2190/files#r536395799) comment from this pull), but there could be something else I don't want to accidentally miss.

stan/math/prim/functor/apply_scalar_unary.hpp
t4c1: I can't think of an example where returning a std::vector of expressions would make sense. But maybe there is one and we will change that later.

stan/math/mix.hpp
t4c1: rev.hpp for example also includes prim, which we dont want included before fwd.

stan/math/prim/fun/quad_form_sym.hpp
t4c1: done

stan/math/prim/functor/apply_vector_unary.hpp
```diff
@@ -48,13 +43,38 @@ struct apply_vector_unary<T, require_eigen_t<T>> {
...
+   * @tparam T Type of argument to which functor is applied.
+   * @tparam F Type of functor to apply.
+   * @param x Eigen input to which operation is applied.
+   * @param f functor to apply to Eigen input.
+   * @return Eigen object with result of applying functor to input
+   */
+  template <typename F, typename T2 = T,
+            require_t<is_eigen_matrix_base<plain_type_t<T2>>>* = nullptr>
+  static inline auto apply_no_holder(const T& x, const F& f) {
+    return f(x).matrix().derived();
```
bbbales2, line 71: I missed the `.derived()` before. Why `.derived()`?

stan/math/prim/meta/holder.hpp
bbbales2: Dang that's good wordplay lol

stan/math/prim/meta/holder.hpp
```diff
@@ -261,17 +262,34 @@ auto holder_handle_element(T& a, T*& res) {
...
  * @tparam T type of the element
  * @param a element to handle
  * @param res resulting pointer to element
  * @return tuple of pointers allocated on heap (containing single pointer).
  */
-template <typename T, require_t<std::is_rvalue_reference<T&&>>* = nullptr>
+template <typename T, require_t<std::is_rvalue_reference<T&&>>* = nullptr,
+          std::enable_if_t<
+              static_cast<bool>(Eigen::internal::traits<std::decay_t<T>>::Flags&
+                                    Eigen::NestByRefBit)>* = nullptr>
```
bbbales2, line 286: What is a `NestByRefBit`? This should probably go in the docs. I didn't find it searching Google

stan/math/mix.hpp
bbbales2: Good point.

bbbales2:
@t4c1 two more questions

stan/math/fwd.hpp
syclik: That makes sense...

for a while, we were trying to get it where once you included the top-level includes for any of the prim / rev / fwd / mix, then everything else is not order-dependent. So... the rest was left alphabetical.

But it was really hard to keep that way.

stan/math/mix.hpp
syclik: Yes, the reason we had `prim` included before `rev` and `fwd` was because `prim` provided all the primary template definitions. Each function in `rev` and `fwd` could template specialize those functions.

To avoid duplicate definitions, we used the `meta` include pretty cleverly, which may have been too clever.

@t4c1, I'm guessing the original include order caused a cascade of compiler errors? If that's the case, one of the changes may have inadvertently brought in too much. (I've done this countless times and had to work through it.) If you did want to detangle it, you can use the generated `.d` files or use the compiler to generate the dependency graph for the compilation process to try to track it down.

t4c1: The way I changed it in this PR seem to work, so I feel no need for extra modifications.

stan/math/prim/functor/apply_vector_unary.hpp
t4c1: Eigen::MatrixXd().matrix() returns some wrapper class. By calling `derived()` we get rid of it and can return the actual expression/matrix type.

stan/math/prim/meta/holder.hpp
t4c1: Eigen is stuck in C++03, in which rvalue references and perfect forwarding do not exist. So when constructing expressions, they can not know which arguments are rvalues and might go out of scope before the expression is evaled. Instead they decide which types to store by reference and which by value depending on the argument type. Lightweight objects, such as operations do not set `NestByRefBit` and are stored by value. Heavy objects, such as matrices set `NestByRefBit` and are stored by reference.

stan/math/mix.hpp
bbbales2: I cloned this math and verified that it works in a Cmdstan, so since it's not obviously broken and it seems to work at least that far upstream, I'm okay with it.

bbbales2: @syclik thanks for taking a look

bbbales2: Actually the logic for the change makes sense but why now? Nothing really changed in this pull request in terms of new functions and what not. Gonna flip these includes back real quick.

We are presumably solving a problem with ADL here. Like, the easiest way to make sure a function is found is to define it before it is used, but that isn't the only way a function is found. For instance, if you call:

```
stan::math::var a;
myfunc(a);
```

I think the entire namespace `stan::math` is searched for a definition of `myfunc`, not just the stuff defined before this bit of code.

bbbales2: Well, the last thing I posted is wrong regarding ADL.

But since I'm playing with this I'm not able to get this to cause any errors.

But I'm also more convinced by what I see the only way this is wrong is that mix is included before rev and fwd and mix might reasonably depend on those lol.

bbbales2: @t4c1 presumably this blew up in one of the Jenkins tests here. You know which?

bbbales2: Not to be too pedantic -- I'm not going to spend too much time trying to make this not work but it bears a little investigating.

t4c1: I think it was the combination of these two:
https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2190/15/pipeline
https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2190/14/pipeline

bbbales2: Oh and I see the conversation back here: https://github.com/stan-dev/math/pull/2190#issuecomment-730683463

t4c1: > But I'm also more convinced by what I see the only way this is wrong is that mix is included before rev and fwd and mix might reasonably depend on those lol.

Thqat can not really be an issue. If mix needs fwd or rev it is simply included. Same if fwd or rev needs prim.

t4c1: I don't remember what exactly was an issue, but it was something, that gcc allows even if it is not standard, while clang is more strict. I know I was reading this: https://clang.llvm.org/compatibility.html#dep_lookup

bbbales2: Yeah I think this change is right.

If everything is template overloads, then we can define things after they are used. Like in this: https://godbolt.org/z/hbT54n

(you'll see both `test::myfunc` functions get called)

If you add an overload of myfunc, it will only get called if the overload is already available when the compiler hits `whatever`. For instance, you can add this code above and below `whatever` and see what happens:

```
int myfunc(const char *) {
    std::cout << "hey" << std::endl;
    return 5;
}
```

Since we're depending on overloading and templates in Stan math, we gotta make sure and include the overloads for the `rev` and `fwd` types before the more general `prim` functions or the stuff in `prim` might not be able to find everything.

bbbales2 (approved):

bbbales2:
Alright, expressions are tested and should work so I merged

But @rok-cesnovar @SteveBronder @andrjohns @wds15 if you see something weird, say something. This is a pretty big new feature!

hsbadr:
@t4c1 @bbbales2 @SteveBronder Could be irrelevant (as I'm using unsupported development combination), but I wanted to report it anyway. I know that `rstan` isn't up to date with `stan` and `math`, but this PR breaks the development version for me. Reverting it fixes the model compilation errors, even with the current development version of `math`. **Template functions are rejected because at least one template argument could not be deduced.** Here's an example when compiling [EpiNow2 ](https://github.com/epiforecasts/EpiNow2)(note that I get similar errors when building `rstanarm` too):
```
stanExports_simulate_secondary.h(1594): error: no instance of function template "model_simulate_secondary_namespace::report_rng" matches the argument list
            argument types are: (Eigen::VectorBlock<const Eigen::Matrix<double, -1, 1, 0, -1, 1>, -1>, const std::vector<double, std::allocator<double>>, const int, boost::random::ecuyer1988, std::ostream *)
                              report_rng(tail(secondary, (all_dates ? t : h )), get_base1(rep_phi, i, "rep_phi", 1), model_type, base_rng__, pstream__),
                              ^
stanExports_simulate_secondary.h(965): note: this candidate was rejected because at least one template argument could not be deduced
  report_rng(const Eigen::Matrix<T0__, Eigen::Dynamic, 1>& reports,
  ^
          detected during:
            instantiation of "void model_simulate_secondary_namespace::model_simulate_secondary::write_array(RNG &, std::vector<double, std::allocator<double>> &, std::vector<int, std::allocator<int>> &, std::vector<double, std::allocator<double>> &, bool, bool, std::ostream *) const [with RNG=boost::random::ecuyer1988]" at line 1625
            instantiation of "void model_simulate_secondary_namespace::model_simulate_secondary::write_array(RNG &, Eigen::Matrix<double, -1, 1, 0, -1, 1> &, Eigen::Matrix<double, -1, 1, 0, -1, 1> &, bool, bool, std::ostream *) const [with RNG=boost::random::ecuyer1988]" at line 138 of "/mnt/local_drive/badr/r/lib/R/library/StanHeaders/include/src/stan/model/model_base_crtp.hpp"
            instantiation of "void stan::model::model_base_crtp<M>::write_array(boost::random::ecuyer1988 &, Eigen::VectorXd &, Eigen::VectorXd &, bool, bool, std::ostream *) const [with M=model_simulate_secondary_namespace::model_simulate_secondary]"
```

rok-cesnovar:
Thanks for the report. Is this using the old stan to c++ compiler? Did you use what rstab uses but jus swapped out the math?

bbbales2:
Yeah thanks for checking this. For clarification, is this breaking the current rstan develop branch? Or a custom rstan branch with a manually updated math?

hsbadr:
This breaks a custom `rstan` development version that fixes some issues with `R-Devel` , uses the development version of `math`, `RcppEigen`, and `RcppParallel` – linked to the system (Intel OneAPI) TBB. Everything is tested and working properly, when this PR is reverted.

PS: I'd like to push my changes in PRs but need time to assure backward compatibility.

rok-cesnovar:
The problem here is that this math version will only work properly for all cases with the latest stanc3 version. That will be available to rstan in the next release and hopefully in the next few weeks on Github. Before that happens you need to revert this PR on your fork.

hsbadr:
Sounds good. Thanks @rok-cesnovar!

rok-cesnovar:
The problem is that user defined functions need to be generated in such a way that they are able to support input Eigen expressions. And stanc2, which is whar cran rstan uses, does not have that support.

hsbadr:
I see. Looking forward to seeing `rstan` supports `stanc3`. The current development version on GitHub does not.

hsbadr:
@rok-cesnovar I confirm that modifying `rstan::stanc` to generate C++ code using [`stanc3`](https://github.com/stan-dev/stanc3/releases/download/nightly/stanc.js) fixed the problem.

See https://github.com/hsbadr/rstan/commit/114552e3d48b1858f494037e99c6aa93323690d9.

rok-cesnovar:
That is great to hear! Thanks.

hsbadr:
> That will be available to rstan in the next release and hopefully in the next few weeks on Github.

@rok-cesnovar  Where can I follow the active development of `stanc3` support in `rstan`?

rok-cesnovar:
I believe its this: https://github.com/stan-dev/rstan/tree/try_2.25

and its currently blocked by https://github.com/stan-dev/stanc3/pull/762

hsbadr:
Thanks @rok-cesnovar! That helps.

PR #1982, opened 2020-07-24
Revert "feature/elementwise_check (2)"

bbbales2:
Reverts stan-dev/math#1798

This pull request was causing noticeable slowdowns in the release.

We should try to figure out how to make these changes without incurring the slowdown.

See: https://discourse.mc-stan.org/t/cmdstan-2-24-release-candidate-now-available/16818/39

PR #2205, opened 2020-11-23
Remove linear indexing requirements from functions

t4c1:
## Summary

Modifies functions requiring linear indexing of their arguments to remove that requirement.

This reintroduces part of functionality reverted in #1982.

## Tests

Expression tests are modified to use a block expression, which does not support linear indexing so the tests will fail for any function that still requires linear indexing.

## Side Effects
None.

## Release notes

Modified functions requiring linear indexing of their arguments to remove that requirement.

## Checklist

- [ ] Math issue #1470 


t4c1:
@bbbales2 You found the performance regression that lead to the revert https://github.com/stan-dev/math/pull/1982, but I did not find what exactly were you testing. Can you check that this PR does not cause a similar regression or post the code you used for the benchmark?

bbbales2:
@t4c1 this revert has a math benchmark: https://github.com/stan-dev/math/pull/2147

This is the model from the July revert: https://discourse.mc-stan.org/t/cmdstan-2-24-release-candidate-now-available/16818/37 (timings further down)

bbbales2:
Also I'm not gonna complain if you implement all the functionality for #1982 :P. It would be convenient to have all these checks work on nested structures.

t4c1:
Thanks Ben. I ran those and fixed the code, so now it is as fast as develop. It turns out `elementwise_check` was too big and g++ decided not to inline it.

stan/math/prim/fun/csr_u_to_z.hpp
```diff
@@ -23,7 +23,6 @@ namespace math {
  * @throw std::out_of_range if i is out of range.
  */
 inline int csr_u_to_z(const std::vector<int>& u, int i) {
-  check_positive("csr_u_to_z", "u.size()", u.size());
```
bbbales2, line 26: This requires `u.size() > 0`. I think check range is checking `u.size() > 1 + error_index`, and `error_index = 1` by default I think.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -203,7 +196,7 @@ inline void elementwise_check(const F& is_good, const char* function,
                               const Indexings&... indexings) {
   for (size_t j = 0; j < x.size(); j++) {
     elementwise_check(is_good, function, name, x[j], must_be, indexings..., "[",
-                      j, "]");
+                      j + 1, "]");
```
bbbales2, line 199: Why the + 1 here?

Does this have anything to do with Stan indexing by 1 or not? There's a variable for this here: https://github.com/stan-dev/math/blob/235090a31ccdabdd0d9d35bfb77ba3d4b984590f/stan/math/prim/meta/error_index.hpp

I think by default the Math library compiles as if it is indexing by 1, and you can optionally change this variable.

stan/math/prim/fun/csr_u_to_z.hpp
t4c1: Yes it does. However the nett cherck will fail if `u.size()==0` and its error type - `std::out_of_range` makes more sense that `std::domain_error`, which would be thrown by `check_positive`. And tests also expect `std::out_of_range`. I had to change this as previously `check_positive` was implemented in a way that would never throw for unsigned types.

stan/math/prim/err/elementwise_check.hpp
t4c1: Because I removed the +1 from `internal::pipe_in`. Yeah I can change this to instead use + error_index::value.

t4c1:
This is ready for review.

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -201,21 +99,119 @@ class Iser {
...
+    if (unlikely(!is_good(scal))) {
+      if (is_eigen_vector<T>::value) {
+        internal::elementwise_throw_domain_error(
+            function, ": ", name, indexings..., "[", i + error_index::value,
+            "] is ", scal, ", but must be ", must_be, "!");
+      } else if (Eigen::internal::traits<T>::Flags & Eigen::RowMajorBit) {
+        internal::elementwise_throw_domain_error(
+            function, ": ", name, indexings..., "[",
+            i / x.rows() + error_index::value, ", ",
+            i % x.rows() + error_index::value, "] is ", scal, ", but must be ",
```
bbbales2, line 143: If it's row major, I think this should be `/ x.cols()` and `% x.cols()` (cause `x.cols()` is the length of a row)

stan/math/prim/err/elementwise_check.hpp
```diff
@@ -201,21 +99,119 @@ class Iser {
...
+    internal::elementwise_throw_domain_error(function, ": ", name, indexings...,
+                                             " is ", x, ", but must be ",
+                                             must_be, "!");
+  }
+}
+template <typename F, typename T, typename... Indexings,
+          require_eigen_t<T>* = nullptr,
+          std::enable_if_t<static_cast<bool>(Eigen::internal::traits<T>::Flags&(
+              Eigen::LinearAccessBit | Eigen::DirectAccessBit))>* = nullptr>
+inline void elementwise_check(const F& is_good, const char* function,
```
bbbales2, line 129: Each of these could use some doc. Just copy paste the first one, and change the text at the beginning. For instance this could be:

```
Check that the predicate holds for the value of `x`. This overload accepts `x` which are Eigen types which can be indexed linearly.
```

I think `overload` is the right word there. If not can drop it.

bbbales2 (changes requested):
Review!

test/expressions/generateExpressionTests.py
```diff
@@ -418,17 +418,29 @@ def main(functions=(), j=1):
             expression_arg_list = ""
             for n, arg in enumerate(function_args[:-1]):
                 if arg in eigen_types:
-                    expression_arg_list += "arg_expr%d.unaryExpr(counter_op%d), " % (
-                        n,
-                        n,
-                    )
+                    if arg == "matrix":
```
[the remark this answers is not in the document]
t4c1: This does not get rid of evaluating anything. What this is changing is making the expression we are passing into functions not only element-wise operation, but also a block (for matrices) or segment (for vectors). Blocks do not support linear indexing so if any function requires linear indexing, such test fails to compile. 

Segments are not strictly necessary, but I have seen some errors happen with segments that do not happen with other expressions, so this is also a bit stricter testing.

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -114,4 +114,22 @@ class arena_matrix : public Eigen::Map<MatrixType> {
 }  // namespace math
 }  // namespace stan
 
+namespace Eigen {
```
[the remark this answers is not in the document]
t4c1: Eigen uses this `traits` struct to store some compile time information about expressions it works with. While it seems it is not needed for `arena_matrix` in Eigen internally, we access some of this data, when creating `ref_type`. Before this PR we did not have any cases of `to_ref(arena_matrix)`.

test/unit/math/prim/prob/neg_binomial_2_test.cpp
```diff
@@ -63,7 +63,7 @@ TEST(ProbDistributionsNegBinomial2, error_check) {
   error_msg
       = "neg_binomial_2_rng: Location parameter "
         "divided by the precision parameter is "
-        "inf, but must be finite!";
+        "inf";
```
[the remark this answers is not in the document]
t4c1: We did not lose anything. In this PR the error message changed slightly, while keeping the meaning the same. Instead of changing the expected error message in these tests, I instead shortened the expected error message to just the part that did not change. This makes the tests less brittle - less likely to fail due to a change in code that is not actually an error.

I would argue that all tests expecting exact wording of an error message are quite brittle. I prefer tests just checking the exception type and if really really necessary a work or a few from the error message.

stan/math/prim/err/elementwise_check.hpp
t4c1: Fixed and improved tests to catch that.

stan/math/prim/err/elementwise_check.hpp
t4c1: Fixed.

test/unit/math/prim/prob/neg_binomial_2_test.cpp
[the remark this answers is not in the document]
bbbales2: Ah, I see. Yes I agree wrt to the test brittleness.

test/expressions/generateExpressionTests.py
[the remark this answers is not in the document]
bbbales2: It looks like we removed testing the base expression test and only test now blocks and segments of the base expression (which are two more complicated things).

I guess my question is should we include the base expression tests as well? Or only have these more complicated tests? Happy with either.

stan/math/prim/err/elementwise_check.hpp
bbbales2: @t4c1 I tried to make some changes to the docs here to avoid hassling you over the editing, but I don't have permissions to push into the remote repo.

So hassle hassle, here's the git diff: [changes.txt](https://github.com/stan-dev/math/files/5616342/changes.txt).

test/expressions/generateExpressionTests.py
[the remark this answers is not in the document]
t4c1: I see no reason for also including simpler expressions.

bbbales2:
@t4c1 looks good. Will approve when tests pass

bbbales2 (approved):

PR #2214, opened 2020-11-26
Adds support for var<matrix> in scalar_seq_view<> distributions

SteveBronder:
## Summary

Adds a `.val(i)` method to `scalar_seq_view<T>` and a specialization of `scalar_seq_view<>` for `var<matrix>` types. When `T` has a autodiff value type `.val(i)` get's back the value from the autodiff type. Else if `T` is an arithmetic type it just returns that value

## Tests

I actually don't know how to test this yet, is there a way to conditionally test this for some of the distributions in the distributions tests? Otherwise I have a separate branch for distributions are vectorized and I need to sort out the last bits of `vector_seq_view`. Should we wait to turn on `var<matrix>` tests until all of those are tested?

## Side Effects

idt so

## Release notes

Adds methods to helpers in probability distributions to support `var<matrix>`

## Checklist

- [x] Math issue #1805 


stan/math/prim/meta/as_column_vector_or_scalar.hpp
```diff
@@ -33,7 +33,7 @@ inline T as_column_vector_or_scalar(const T& a) {
  * @return Same vector.
  */
 template <typename T, require_eigen_col_vector_t<T>* = nullptr>
-inline T&& as_column_vector_or_scalar(T&& a) {
+inline auto as_column_vector_or_scalar(T&& a) {
   return std::forward<T>(a);
 }
```
SteveBronder, line 38: @t4c1 I changed this because of what we talked about in the past where passing in an rvalue and returning an rvalue can cause the destructor to go off. Should we just have specializations for doubles where it passes back the correct reference type?

stan/math/prim/meta/to_value_array_or_scalar.hpp
```diff
@@ -0,0 +1,84 @@
...
+inline auto to_value_array_or_scalar(T&& v) {
+  using T_map
+      = Eigen::Map<const Eigen::Array<value_type_t<T>, Eigen::Dynamic, 1>>;
+  return value_of(T_map(v.data(), v.size()));
+}
+
+}  // namespace math
+}  // namespace stan
+
+#endif
```
SteveBronder, line 84: TODO: Need to write tests for all these new functions (I think we can delete some of them)

SteveBronder: Also update docs

stan/math/prim/meta/as_column_vector_or_scalar.hpp
t4c1: We don't need references for doubles. Although this will always return value (even when it should return a lvalue reference). What you wanted here is `decltype(auto)`. What I would prefer here is `T` - it will behave exactly the same.

bbbales2:
@SteveBronder got the basic normal lpdf/cdf/lcdf/lccdf tests working. I'm running all the tests now to find out if there are broken distributions.

stan/math/prim/meta/to_value_array_or_scalar.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * Access the values of a matrix type and casts them to an array.
+ *
+ * @tparam T Type of \c Eigen \c Matrix or expression
+ * @param v Specified \c Eigen \c Matrix or expression.
+ * @return Matrix converted to an array.
+ */
+template <typename T, require_eigen_t<T>* = nullptr,
+          require_not_plain_type_t<T>* = nullptr>
+inline auto to_value_array_or_scalar(T&& v) {
+  return value_of(v.eval()).array();
```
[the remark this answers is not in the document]
t4c1: Actually I think none of these overloads should use eval. If you need eval you should add a separate helper that also evals the result. Or if you dont need non-evaling ones you can skip them, but if some helper also evals it should have eval in its name.

stan/math/prim/prob/gumbel_lpdf.hpp
```diff
@@ -87,11 +80,11 @@ return_type_t<T_y, T_loc, T_scale> gumbel_lpdf(const T_y& y, const T_loc& mu,
...
-    const auto& scaled_diff
-        = to_ref_if<!is_constant_all<T_loc>::value
-                        + !is_constant_all<T_scale>::value
-                        + !is_constant_all<T_y>::value
-                    >= 2>(inv_beta * exp_y_m_mu_over_beta - inv_beta);
+    const auto& scaled_diff = to_ref_if<!is_constant_all<T_loc>::value
+                                            + !is_constant_all<T_scale>::value
+                                            + !is_constant_all<T_y>::value
+                                        >= 2>(
+        inv_beta * std::move(exp_y_m_mu_over_beta) - inv_beta);
```
t4c1, line 87: That move does not do anything.

stan/math/prim/prob/gumbel_lpdf.hpp
```diff
@@ -100,7 +93,8 @@ return_type_t<T_y, T_loc, T_scale> gumbel_lpdf(const T_y& y, const T_loc& mu,
     }
     if (!is_constant_all<T_scale>::value) {
       ops_partials.edge3_.partials_
-          = -y_minus_mu_over_beta * scaled_diff - inv_beta;
+          = -std::move(y_minus_mu_over_beta) * std::move(scaled_diff)
+            - std::move(inv_beta);
```
t4c1, line 97: Nor do these.

stan/math/prim/prob/normal_id_glm_lpdf.hpp
```diff
@@ -84,8 +84,8 @@ return_type_t<T_y, T_x, T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
                         N_instances);
   check_consistent_size(function, "Vector of intercepts", alpha, N_instances);
   T_sigma_ref sigma_ref = sigma;
-  const auto& sigma_val = value_of_rec(sigma_ref);
-  const auto& sigma_val_vec = to_ref(as_column_vector_or_scalar(sigma_val));
+  auto&& sigma_val_vec
+      = to_ref(as_column_vector_or_scalar(value_of_rec(sigma_ref)));
```
t4c1, line 88: if `as_column_vector_or_scalar` is called woth a lvalue it does not need to construct a holder, so previous version is better.

stan/math/prim/prob/normal_id_glm_lpdf.hpp
```diff
@@ -100,16 +100,11 @@ return_type_t<T_y, T_x, T_alpha, T_beta, T_scale> normal_id_glm_lpdf(
...
-
-  const auto& y_val_vec = as_column_vector_or_scalar(y_val);
-  const auto& alpha_val_vec = as_column_vector_or_scalar(alpha_val);
-  const auto& beta_val_vec = to_ref_if<!is_constant<T_x>::value>(
-      as_column_vector_or_scalar(beta_val));
+  auto&& x_val = to_ref_if<!is_constant<T_beta>::value>(value_of_rec(x_ref));
+  auto&& y_val_vec = as_column_vector_or_scalar(value_of_rec(y_ref));
+  auto&& alpha_val_vec = as_column_vector_or_scalar(value_of_rec(alpha_ref));
+  auto&& beta_val_vec = to_ref_if<!is_constant<T_x>::value>(
+      as_column_vector_or_scalar(value_of_rec(beta_ref)));
```
t4c1, line 107: Do you need these to be non-const for some reason?

stan/math/prim/prob/pareto_cdf.hpp
```diff
@@ -53,22 +53,21 @@ return_type_t<T_y, T_scale, T_shape> pareto_cdf(const T_y& y,
...
       continue;
     }
 
-    const T_partials_return log_dbl
-        = log(value_of(y_min_vec[n]) / value_of(y_vec[n]));
-    const T_partials_return y_min_inv_dbl = 1.0 / value_of(y_min_vec[n]);
-    const T_partials_return alpha_dbl = value_of(alpha_vec[n]);
+    const T_partials_return log_dbl = log(y_min_vec.val(n) / y_vec.val(n));
+    const T_partials_return y_min_inv_dbl = 1.0 / y_vec.val(n);
+    const T_partials_return alpha_dbl = y_vec.val(n);
```
t4c1, line 70:
```suggestion
    const T_partials_return alpha_dbl = alpha_vec.val(n);
```

t4c1 (changes requested):
This PR is getting quite large. Can you split it up? Maybe keep just one distribution in the first one and leave the reminder for another one.

stan/math/prim/prob/gumbel_lpdf.hpp
SteveBronder: Yeah my b I think I was doing something silly here when testing this out

bbbales2:
I haven't looked through this yet. Let me know if I'm holding anything up by not doing that.

SteveBronder:
No idt so, I just have a lot of other stuff on my plate and haven't had time to come back to this

SteveBronder:
@t4c1 I think your right this should be broken up. But the one issue here is that I'd really like to not mess with the distributions test code too much as there's not meta level tests to test it and it's pretty dense/old. I think what would be good here is to break this up into three PRs.

1. PR for backend stuff like `array_or_scalar`, `seq_view`, etc support for `var<matrix>`
2. Getting all the LDLT and decomposition stuff working for `var<matrix>`
3. A PR that then uses all that stuff in the distributions. 

Doing all the distributions at once is brutal but without it we have to tweak the distribution tests to only run for `var<matrix>` for only the distributions in the successive PRs. Then we'd remove those changes to conditionally run the `var<matrix> code for the distributions once all those PRs go in. tbh I think both Ben and I tried having a hack at doing the conditional distribution testing and neither of us sorted it out.

bbbales2:
> I think what would be good here is to break this up into three PRs

I like this. Happy to do the big boring review ("A PR that then uses all that stuff in the distributions.")

I'm working on the ldlt/multivariate stuff now.

> only run for var<matrix> for only the distributions in the successive PRs

Yeah, good point, we're writing new test files and we write them for all or none (this [file](https://github.com/stan-dev/math/pull/2214/files#diff-8bd79e4d3d78b57183e778917322e785fc3b227ce64c8ae4fffc6af93b61ca2aR26))

bbbales2:
There's only one big PR left for this right? It should just be the testing framework stuff?

PR #2218, opened 2020-11-28
Update to the "Using the OpenCL backend" doxygen

rok-cesnovar:
## Summary

An update of the "Using the OpenCL backend" doxygen. Now listing all currently OpenCL supported functions.

## Tests

/

## Side Effects

/

## Checklist

### DIFF
--- a/doxygen/parallelism_support/opencl_support.md
+++ b/doxygen/parallelism_support/opencl_support.md
@@ -103,9 +103,8 @@ An example of using OpenCL supported Stan Math functions for reverse mode AD is 
 ```cpp
 using stan::math;
 
-Eigen::Matrix<double, -1, -1> A(N,M);
-Eigen::Matrix<double, -1, -1> B(M,N);
-// ... fill matrices A and B
+Eigen::Matrix<double, -1, -1> A = Eigen::Matrix<double, -1, -1>::Random(N, M);
+Eigen::Matrix<double, -1, -1> B = Eigen::Matrix<double, -1, -1>::Random(M, N);
 var_value<matrix_cl<double>> A_cl = to_matrix_cl(A);
 var_value<matrix_cl<double>> B_cl = to_matrix_cl(B);
 var_value<matrix_cl<double>> C_cl = A_cl * B_cl;
--- a/doxygen/parallelism_support/opencl_support.md
+++ b/doxygen/parallelism_support/opencl_support.md
@@ -101,10 +101,7 @@ The OpenCL backend can be used for reverse mode AD as well as primitive function
 An example of using OpenCL supported Stan Math functions for reverse mode AD is shown below: 
 
 ```cpp
-using stan::math::matrix_cl;
-using stan::math::var_value;
-using stan::math::to_matrix_cl;
-using stan::math::from_matrix_cl;
+using stan::math;
 
 Eigen::Matrix<double, -1, -1> A(N,M);
 Eigen::Matrix<double, -1, -1> B(M,N);
@@ -126,6 +123,8 @@ C(0,0).grad();
 - sum
 
 #### Supported distributions
+
+Only distribution functions with the _lpdf/_lpmf suffix are currently supported. 
 
 - bernoulli
 - bernoulli_logit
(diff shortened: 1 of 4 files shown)


doxygen/parallelism_support/opencl_support.md
```diff
@@ -94,20 +94,135 @@ Once you have done the above step, `runTests.py` should execute with the GPU ena
...
-- bernoulli_logit_glm
 - cholesky_decompose
-- categorical_logit_glm
 - gp_exp_quad_cov
 - mdivide_right_tri
 - mdivide_left_tri
 - multiplication
+- sum
+
+#### Supported distributions
```
t4c1, line 128: So far only lpdf/lpmf functions are supported (no cdf/lcdf/lccdf).

rok-cesnovar: Added a sentence that mentions that. Let me know if you would prefer to have the list explicitly mention the suffices like
```
- bernoulli_lpmf
- bernoulli_logit_lpmf
- bernoulli_logit_glm_lpmf
- beta_lpdf
...
```

t4c1: I expect I will start adding  cdf/lcdf/lccdf within a month or two. We will need to mention explicit suffixes at that point.

rok-cesnovar: Ok, explicit suffices it is! :)

doxygen/parallelism_support/opencl_support.md
```diff
@@ -94,20 +94,128 @@ Once you have done the above step, `runTests.py` should execute with the GPU ena
...
+### Reverse mode
+
+An example of using OpenCL supported Stan Math functions for reverse mode AD is shown below: 
+
+```cpp
+using stan::math;
+
+Eigen::Matrix<double, -1, -1> A(N,M);
+Eigen::Matrix<double, -1, -1> B(M,N);
+// ... fill matrices A and B
```
SteveBronder, line 108: We could just replace this with something like 
```suggestion
Eigen::Matrix<double, -1, -1> A = Eigen::Matrix<double, -1, -1>::Random(N, M);
Eigen::Matrix<double, -1, -1> B = Eigen::Matrix<double, -1, -1>::Random(M, N);
```

doxygen/parallelism_support/opencl_support.md
```diff
@@ -94,20 +94,128 @@ Once you have done the above step, `runTests.py` should execute with the GPU ena
...
+var_value<matrix_cl<double>> A_cl = to_matrix_cl(A);
+var_value<matrix_cl<double>> B_cl = to_matrix_cl(B);
+var_value<matrix_cl<double>> C_cl = A_cl * B_cl;
+var_value<Eigen::Matrix<double, -1, -1>> C = from_matrix_cl(C_cl);
+C(0,0).grad();
```
SteveBronder, line 113: Here we would probably want to set C's values to 1 and then run `grad()` right? @bbbales2 what was the reason we don't have `C.grad()` that fills the adjoints with 1 by default for var matrix types?

SteveBronder (approved):
Two little things but is fine! The second one we should probably clarify before merging as I think calling `c(0, 0).grad()` would just set the first cell's adjoint to 1 but then not the rest of the matrix. Maybe just do `stan::math::sum(C).grad()`

SteveBronder:
Thanks for fixing these up!

rok-cesnovar:
@SteveBronder I think we also need to regenerate the website? http://mc-stan.org/math/ is still showing 3.3.0.

doxygen/parallelism_support/opencl_support.md
rok-cesnovar: Added the sum() to avoid this.

SteveBronder:
Jenkins should be doing that when it runs on the develop branch. lemme look at this

SteveBronder:
Huh part of it looks like it's working but it looks like there are some git failures that are happening silently when we upload the docs?

https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/develop/946/pipeline/284

SteveBronder:
But the `gh-pages` branch is updated

https://github.com/stan-dev/math/tree/gh-pages

I think we just need to bump the version number on the docs site because it the docs seem to have the most recent changes

rok-cesnovar:
Ah yeah, this file also was not updated: https://github.com/stan-dev/math/blob/develop/stan/math/version.hpp
That is what threw me off. It does update otherwise yeah. Sorry for the false alarm

rok-cesnovar:
This now also fixes all the version numbers. Should be good to merge now.

SteveBronder (approved):
lcbtm!

Issue #2211, opened 2020-11-25
Different finite diff Vs. autodiff gradients at out of range acos values


SteveBronder:
## Description

It looks like we get different answers for out of range values for acos (and I'm guessing the other geom functions) in the testing framework for finite diff vs. autodiff. I'm not totally sure how to handle it

Adding the below to the mix tests for `acos` gives the errors (full output here](https://gist.github.com/SteveBronder/2ef992c36c2f5a70a99f484aeab6e180))

## Example

```cpp
#define STAN_TEST_PRINT_MATRIX_FAILURE
#include <test/unit/math/test_ad.hpp>
#include <complex>
#include <limits>
#include <vector>

template <typename T>
void expect_acos(const T& m, const stan::test::ad_tolerances& tols
                           = stan::test::ad_tolerances()) {
  auto f = [](const auto& x) { return stan::math::acos(x); };
  Eigen::VectorXd v(m.size());
  Eigen::RowVectorXd rv(m.size());
  std::vector<double> st(m.size());

  for (int i = 1; i < m.size(); ++i) {
    v(i) = m(i);
    rv(i) = m(i);
    st[i] = m(i);
  }

  stan::test::expect_ad(tols, f, v);

}

TEST(MathMixMatFun, acos_eigen) {
  Eigen::MatrixXd a(0, 0);
  expect_acos(a);

  Eigen::MatrixXd b(1, 1);
  b << 1;
  expect_acos(b);

  stan::test::ad_tolerances tols;
  Eigen::MatrixXd c(3, 1);
  c << 100, 0, -3;
  expect_acos(c, tols);

  Eigen::MatrixXd d(1, 3);
  d << 100, 0, -3;
  expect_acos(d, tols);

  Eigen::MatrixXd e(3, 2);
  e << -100, 0, 1, 20, -40, 2;
  expect_acos(e, tols);

  Eigen::MatrixXd g(6, 1);
  g << 1, 2, 3, 4, 5, 6;
  expect_acos(g, tols);
}
```

```
./test/unit/math/expect_near_rel.hpp:77: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(0, nan)
expect_near_rel; require items x1(2) = x2(2): gradient() grad

./test/unit/math/expect_near_rel.hpp:138: Failure
Failed

x1: 
[-1]
[ 0]
[ 0]
x2: 
[ -1]
[  0]
[nan]

[...]

PR #2225, opened 2020-12-02
Adding Generalized Inverse 

spinkney:
## Summary

This provides the Moore-Penrose inverse in Stan (aka generalized inverse) from [this paper](https://arxiv.org/ftp/arxiv/papers/0804/0804.4809.pdf) which uses the Cholesky decomposition. The Cholesky is only used to invert the cross-product of the input matrix. With Stan's linear algebra optimizations I use `mdivide_*_spd` directly, which is faster. A few things that I would like to discuss and change with the help of other developers:

- [x] In some circumstances the <strike>Cholesky decomposition</strike> SPD matrix requires a small amount of jitter added to the diagonal to successfully invert. I added it in the code which is problematic for two reasons. First, it distorts the final output but a small amount and, secondly, it would be better if the user could decide to add it or not (as discussed on Slack). 
- [x] The generalized inverse = the inverse when the matrix is square and the code calls inverse in that situation. If the user supplies a square triangular matrix it could be even more efficient to call a triangular inverse method. Do we want to handle these situations or trust the user to know what to do?
- [x]  A nice analytic derivative exists of the generalized inverse ([see this post on mathoverflow](https://mathoverflow.net/a/29511)), however, this PR is only for the "prim" function. 
- [x] The spelling of "generalized" is very North American English. Do we want to alias "generalised_inverse" with an "s" as well?

## Tests

There are two tests. 

1. Test that it works with an empty matrix.
2. Test that the generalized inverse output is equivalent to a known rectangular inverse. Note that the test uses EXPECT_MATRIX_NEAR as the diagonal jitter slightly changes the output. If the code is changed so that the user supplies the diagonal jitter this test can be updated to EXPECT_MATRIX_FLOAT_EQ.

## Side Effects

No

## Release notes
The Moore-Penrose generalized inverse is added. Users can use this by calling `generalized_inverse(matrix M)`.

## Checklist


stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,68 @@
...
+  const auto n = G.rows();
+  const auto m = G.cols();
+
+  if (G.rows() == G.cols()) {
+    return G.inverse();
+  }
+
+  if (n < m) {
+    Eigen::Matrix<value_t, -1, -1> A = tcrossprod(G);
+    A.diagonal().array() += Eigen::Array<double, -1, 1>::Constant(n, 1e-10);
```
t4c1, line 50: If matrix contains only small values this can be a large relative change, which can affect the result a lot. Would it make sense to instead do something like? 
```
A.diagonal().array() *= 1 + 1e-10;
```

test/unit/math/prim/fun/generalized_inverse_test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+  stan::math::matrix_d m1(3, 2);
+  m1 << 1, 2, 2, 4, 1, 2;
+
+  stan::math::matrix_d m2(2, 3);
+  m2 << 1 / 30.0, 1 / 15.0, 1 / 30.0, 1 / 15.0, 2 / 15.0, 1 / 15.0;
+  stan::math::matrix_d m3 = generalized_inverse(m1);
+  for (int j = 0; j < m3.cols(); j++)
+    for (int i = 0; i < m3.rows(); i++)
+      EXPECT_NEAR(m2(i, j), m3(i, j), 1e-5);
```
t4c1, line 23: Such comparisons are so common, we have a macro for them:
```suggestion
  EXPECT_MATRIX_NEAR(m2, m3, 1e-5);
```
To use it you will need to include appropriate header:
```
#include <test/unit/util.hpp>
```

t4c1: Also does this achieve only 1e-5 accuracy? In tests for most functions we require 1e-9.

t4c1 (changes requested):
Stan uses mostly American spelling, so there is no need for adding `generalise`.

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * @tparam T type of elements in the matrix
+ * @tparam n number of rows, can be Eigen::Dynamic
+ * @tparam m number of columns, can be Eigen::Dynamic
+ *
+ * @param M specified matrix
+ * @return Generalized inverse of the matrix (an empty matrix if the specified
+ * matrix has size zero).
+ */
+template <typename EigMat,
+          require_eigen_vt<std::is_arithmetic, EigMat>* = nullptr>
```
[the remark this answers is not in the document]
spinkney: I do plan to add the rev implementation once I understand how to do it :). Going to work with @SteveBronder on it.

stan/math/prim/fun/generalized_inverse.hpp
spinkney: I think it's even better if we let the user handle the constant. In the test I was able to get 1e-9 accuracy finding the appropriate constant. For matrices where the Cholesky decomposition is successful one does not need to add this jitter.

test/unit/math/prim/fun/generalized_inverse_test.cpp
spinkney: I did get it to achieve the 1e-9 accuracy. But it depends on the user's tolerance. When Cholesky is successful this method will exactly reproduce the generalized inverse. However, due to numerical instability in Cholesky decompositions it is sometimes necessary to add a small jitter to the diagonal. This is a well known issue but the user should supply the tolerance they are comfortable with.

spinkney: I added the EXPECT_MATRIX_NEAR as well.

SteveBronder:
We should probably add a test to mix as well so we can check the gradient is similar to what we would expect with a finite difference method. You can check out the code in the mix inverse test for an example.

https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/fun/inverse_test.cpp

wrt the jitter I think we should have signatures `generalized_inverse(Matrix)` and `generalized_inverse(Matrix, double)` where if users pass in a second argument the jitter is applied to the diagonal

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,64 @@
...
+  using value_t = value_type_t<EigMat>;
+  if (G.size() == 0) {
+    return {};
+  }
+
+  const auto n = G.rows();
+  const auto m = G.cols();
+
+  if (G.rows() == G.cols()) {
+    return G.inverse();
```
SteveBronder, line 45: I think it would be better to call stan's inverse function here so that if the matrix has vars in it then it will use the reverse mode specialization

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,64 @@
...
+    A.diagonal().array()
+        += Eigen::Array<double, Eigen::Dynamic, 1>::Constant(n, 1e-8);
+    return transpose(mdivide_left_spd(A, G));
+  } else {
+    Eigen::Matrix<value_t, Eigen::Dynamic, Eigen::Dynamic> A = crossprod(G);
+    A.diagonal().array()
+        += Eigen::Array<double, Eigen::Dynamic, 1>::Constant(m, 1e-8);
+    return transpose(mdivide_right_spd(G, A));
+  }
+}
```
SteveBronder, line 59: When you break this out the non-jittered version can just be

```suggestion
  if (n < m) {
    return transpose(mdivide_left_spd(tcrossprod(G), G));
  } else {
    return transpose(mdivide_right_spd(G, crossprod(G)));
  }
}
```

stan/math/prim/fun/generalized_inverse.hpp
spinkney: Done

stan/math/prim/fun/generalized_inverse.hpp
spinkney: Done and I added a test. Now there's a test for the jitter and a test for non-jitter.

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace stan {
+namespace math {
+
+/**
+ * Returns the Moore-Penrose generalized inverse of the specified matrix.
+ *
+ * @tparam T type of elements in the matrix
+ * @tparam n number of rows, can be Eigen::Dynamic
+ * @tparam m number of columns, can be Eigen::Dynamic
```
SteveBronder, line 26: These need removed and docs for `EigMat` need to be added

stan/math/prim/fun/generalized_inverse.hpp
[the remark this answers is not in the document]
SteveBronder: This should work for rev even without an actual specialization so I think you can make this just `require_eigen_t<EigMat>`. Once we have the rev specialization we can add a `require_not_vt_var<EigMat>` (that checks that the `value_type` is not var)

SteveBronder (changes requested):
Two little comments. Stuff in prim should work with matrices with `double`, `var`, and `fvar<T>` unless there is already a specialization in `rev` or `fwd`. We can relax the requires here to allow for matrices with `var` and `fvar` types until we have a rev specialization. That also lets us have a test in `mix` to check that finite difference with doubles and an input matrix of vars gives back the same gradients

spinkney:
> We should probably add a test to mix as well so we can check the gradient is similar to what we would expect with a finite difference method. You can check out the code in the mix inverse test for an example.
> 
> https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/fun/inverse_test.cpp
This is used once I add the code for the analytic gradient or before?

stan/math/prim/fun/generalized_inverse.hpp
spinkney: I updated but not sure if it's right, let me know.

rok-cesnovar:
> This is used once I add the code for the analytic gradient or before?

Once you implemented what Steve suggested [here](https://github.com/stan-dev/math/pull/2225/files#r534549996) this function also works with auto diff. So the gradient works already and you could use it in Stan (if you exposed it in the compile). The analytic gradient (`/rev` implementation) it will just be more efficient.

spinkney:
> > This is used once I add the code for the analytic gradient or before?
> 
> Once you implemented what Steve suggested [here](https://github.com/stan-dev/math/pull/2225/files#r534549996) this function also works with auto diff (reverse mode and forward mode). So the gradient works already and you could use it in Stan (if you exposed it in the compile). The analytic gradient (`/rev` implementation) will just be more efficient. If you add a test in /mix here you won't have to add it when you implement the gradient.

Great, thanks. I pushed the tests but they fail for type reasons that I don't understand.

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,123 @@
...
+  const auto m = G.cols();
+
+  if (G.rows() == G.cols()) {
+    return inverse(G);
+  }
+
+  if (n < m) {
+    Eigen::Matrix<value_t, Eigen::Dynamic, Eigen::Dynamic> A = tcrossprod(G);
+    A.diagonal().array()
+        += Eigen::Array<double, Eigen::Dynamic, 1>::Constant(n, a);
```
t4c1, line 110: [optional] This can be simplified:
```suggestion
    A.diagonal().array() += a;
```
Same for the addition to diagonal in the other branch.

spinkney: Thanks, looks better.

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+  if (G.rows() == G.cols()) {
+    return inverse(G);
+  }
+
+  if (n < m) {
+    return transpose(mdivide_left_spd(tcrossprod(G), G));
+  } else {
+    return transpose(mdivide_right_spd(G, crossprod(G)));
+  }
```
SteveBronder, line 63: `EigMat` can potentially be an Eigen expression like `Product<Matrix, Matrix>` which can be bad to use multiple times because `G` will be evaluated each time it's called. When you use an eigen input multiple times you should use `auto&& G_ref = to_ref(G)`. `to_ref` looks at an Eigen type and if it's an expression will evaluate it. If `EigMat` is a plain type (i.e. like an Eigen::Matrix` then `to_ref` just passes the object back by reference. So for here you want like 

```suggestion
  if (G.rows() == G.cols()) {
    return inverse(G);
  }
  const auto& G_ref = to_ref(G);
  if (n < m) {
    return transpose(mdivide_left_spd(tcrossprod(G_ref), G_ref));
  } else {
    return transpose(mdivide_right_spd(G_ref, crossprod(G_ref)));
  }
```

SteveBronder: You only need to do it after the first if because if the `if` is true then `G` is only used once

stan/math/prim/fun/generalized_inverse.hpp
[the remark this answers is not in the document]
spinkney: Now that the rev part is in, do I just add `require_not_vt_var` in prim instead of `require_eigen_t`? Also, for the alias that inputs a scalar is there anything to do there (ie `require_stan_scalar_t<Scal>`)?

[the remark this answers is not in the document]
spinkney: @t4c1 @SteveBronder I spent some time trying to get the tests to work with `require_not_vt_var` but failed. The tests using `require_eigen_t` passed on my end with what I pushed this morning.

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Reverse mode differentiation algorithm reference:
+ *
+ * The Differentiation of Pseudo-Inverses and Nonlinear Least Squares Problems
+ * Whose Variables Separate. Author(s): G. H. Golub and V. Pereyra. Source: SIAM
+ * Journal on Numerical Analysis, Vol. 10, No. 2 (Apr., 1973), pp. 413-432
+ *
+ * Equation 4.12 in the paper
+ */
+template <typename EigMat, require_rev_matrix_t<EigMat>* = nullptr>
```
spinkney, line 28: @SteveBronder , I added `require_rev_matrix_t` here. Let me know if it's the right signature.

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,124 @@
...
+    auto A_spd = tcrossprod(G_arena.val());
+    arena_t<EigMat> inv_G(transpose(mdivide_left_spd(A_spd, G_arena.val())));
+    auto aP = (1 - G_arena.val() * inv_G.transpose());
+    auto Pa = (1 - inv_G * G_arena.val());
+    reverse_pass_callback([G_arena, inv_G, aP, Pa]() mutable {
+      G_arena.adj() += -inv_G * G_arena.adj() * inv_G;
+      G_arena.adj() += tcrossprod(inv_G) * G_arena.adj().transpose() * aP;
+      G_arena.adj() += Pa * G_arena.adj().transpose() * crossprod(inv_G());
+    });
+    return inv_G;
```
SteveBronder, line 52: Couple little things here

1. the type returned by `arena_t<>` will always be a `plain_type` so you can do just 

`arena_t<EigMat> G_arena(G);`

2. Anything that goes into the lamda's capture like
```
reverse_pass_callback([%Copied Into Lambda%]() mutable {
```

Needs to be put on our autodiff stack using `arena_t`. For `aP` and `Pa` you can do

```cpp
    auto aP = to_arena((1 - G_arena.val() * inv_G.transpose()));
    auto Pa = to_arena(1 - inv_G * G_arena.val());
```

and then these will be moved over to the arena.

3. Eigen doesn't have the concept of elementwise operations over matrices, so you need to tell eigen to treat things as an array, do the subtraction, then tell eigen to treat it as a matrix like

```cpp
    auto aP = to_arena((1 - (G_arena.val() * inv_G.transpose()).array()).matrix());
    auto Pa = to_arena((1 - (inv_G * G_arena.val()).array()).matrix());
```

4. Since you transpose in 

```
    arena_t<EigMat> inv_G(transpose(mdivide_right_spd(G_arena.val_op(), A_spd)));
```

idt you need to transpose again here

```
    auto aP = to_arena((1 - (G_arena.val_op() * inv_G.val_op().transpose()).array()).matrix());
```

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,124 @@
...
+  } else {
+    arena_t<plain_type_t<EigMat>> G_arena(G);
+    auto A_spd = crossprod(G_arena.val());
+    arena_t<EigMat> inv_G(transpose(mdivide_right_spd(G_arena.val(), A_spd)));
+    auto aP = (1 - G_arena.val() * inv_G.transpose());
+    auto Pa = (1 - inv_G * G_arena.val());
+    reverse_pass_callback([G_arena, inv_G, aP, Pa]() mutable {
+      G_arena.adj() += -inv_G * G_arena.adj() * inv_G;
+      G_arena.adj() += tcrossprod(inv_G) * G_arena.adj().transpose() * aP;
+      G_arena.adj() += Pa * G_arena.adj().transpose() * crossprod(inv_G());
```
SteveBronder, line 62: Once I fixed up a few things I had trouble getting this to compile, then at runtime it was telling me the matmuls were the wrong sizes and then that G_arena was not the same size as the right hand side of the +=.

I ended up having to transpose each section but I think that's wrong? At least the ad tests were telling me it was incorrect

SteveBronder, line 62:
```suggestion
      G_arena.adj() += -inv_G * G_arena.adj() * inv_G;
      G_arena.adj() += tcrossprod(inv_G) * G_arena.adj().transpose() * aP;
      G_arena.adj() += Pa * G_arena.adj().transpose() * crossprod(inv_G);
```

test/unit/math/mix/fun/generalized_inverse_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+  z << 1, 2, 5, std::numeric_limits<double>::quiet_NaN();
+  EXPECT_NO_THROW(stan::math::generalized_inverse(z));
+  EXPECT_NO_THROW(stan::math::generalized_inverse(z, 0.0));
+
+  // autodiff throws, so following fails (throw behavior must match to pass)
+
+  Eigen::MatrixXd a(2, 2);
+  a << 1.9, 0.3, 0.3, std::numeric_limits<double>::infinity();
+  expect_ad(f, a);
+  expect_ad(fjit, a);
```
SteveBronder, line 52: If you want to test this works with `var_value<Matrix>` as well as `Matrix<var>` you can just add

```cpp
expect_ad_matvar(f, a);
```

so that those tests are checked as well.

SteveBronder (changes requested):
The prim code looks good to go! Though the rev section needs some changes. I have a branch below that gets the rev version to compile. Though the answers are wrong so that math needs fixed.

https://github.com/stan-dev/math/tree/suggests/gen-inv

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,124 @@
...
+ two
+ * <code>generalized_inverse</code> functions, one that uses one input matrix
+ (this one)
+ * and another that works with an input matrix and a small jitter to the
+ diagonal of the internal SPD
+ * matrix.
+ */
+template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
+inline Eigen::Matrix<value_type_t<EigMat>, EigMat::RowsAtCompileTime,
+                     EigMat::ColsAtCompileTime>
```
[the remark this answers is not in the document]
spinkney: It should always return the size of the transpose of the input matrix. So an input matrix of `Eigen::Matrix<double, 5, 4>` returns a `Eigen::Matrix<double, 4, 5>` matrix. Should I make it 
```c++
inline Eigen::Matrix<value_type_t<EigMat>, EigMat::ColsAtCompileTime,
                     EigMat::RowsAtCompileTime>
```
?

stan/math/rev/fun/generalized_inverse.hpp
t4c1: > Eigen doesn't have the concept of elementwise operations over matrices

Actually addition and subtraction are supported on matrices. For any other elementwise operation you are correct.

spinkney: I found an implementation in Python and re-read the original paper to find out that it was not correct. I needed to subtract by a diagonal matrix of 1s. I rewrote and now the rev tests are passing!!!

spinkney: Except I get this one error I don't understand,
```
Assertion failed: (dst.rows() == src.rows() && dst.cols() == src.cols()), function resize_if_allowed, file lib/eigen_3.3.9/Eigen/src/Core/AssignEvaluator.h, line 710.
./test/unit/math/mix/fun/generalized_inverse_test --gtest_output="xml:./test/unit/math/mix/fun/generalized_inverse_test.xml" failed
exit now (12/14/20 07:36:31 EST)
```

SteveBronder:
Most excellent! I will review this in the morning!

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,198 @@
...
+    return ret_type(G);
+
+  if (G.rows() == G.cols())
+    return ret_type(inverse(G));
+
+  if (G.rows() < G.cols()) {
+    arena_t<VarMat> G_arena(G);
+    auto A_spd = tcrossprod(G_arena.val_op());
+    arena_t<VarMat> inv_G(
+        mdivide_left_spd(A_spd.val_op(), G_arena.val_op()).transpose());
```
SteveBronder, line 61: 1. You don't need to use `.val()/.val_op()/etc` on matrices of non-var types. When a matrices type is an `Eigen::Matrix<var, R, C>` we use `.val()`, and `.adj()`to pull out the values or adjoints of the underlying `var`s. It effectively generates an `Eigen::Matrix<double, R, C>` for each of the values and adjoints. But when you multiply something like `A.val() * B.adj()` your result is also going to be `Eigen::Matrix<double, R, C>`. So at that point you don't need to then use `.val()` etc on that result matrix.

2. In general it's more performant to call Eigen methods directly instead of using stan's functions. So for `tcrossprod()` you can directly do `G_arena.val_op() * G_arena.val_op().transpose()`. Similarly for `mdivide_left_spd` you can put everything together like
 ```suggestion
    arena_t<ret_type> inv_G((G_arena.val_op() * G_arena.val_op().transpose())
                                .llt()
                                .solve(G_arena.val_op())
                                .transpose());
```

File: stan/math/rev/fun/generalized_inverse.hpp [hunk 15]
```diff
@@ -0,0 +1,198 @@
...
+    auto GP = to_arena(-inv_G.val_op() * G_arena.val_op());
+    GP.diagonal().array() += 1.0;
+
+    reverse_pass_callback([G_arena, res, inv_G, GP, PG]() mutable {
+      G_arena.adj() -= inv_G.val_op().transpose() * res.adj_op()
+                       * inv_G.val_op().transpose();
+      G_arena.adj() += PG.val_op() * res.adj_op().transpose()
+                       * tcrossprod(inv_G.val_op());
+      G_arena.adj()
+          += crossprod(inv_G.val_op()) * res.adj_op().transpose() * GP.val_op();
```
SteveBronder, line 77: So here for performance we only want to write to `G_arena`'s adjoint once. It's good to break things down like this when building stuff out but it would be good performance wise to zip everything into one large expression

The below also inlines the calculations of GP and PG 

```suggestion
      G_arena.adj()
          += -(inv_G.val_op().transpose() * inv_G.adj_op()
               * inv_G.val_op().transpose())
             + (-G_arena.val_op() * inv_G.val_op()
                + Eigen::MatrixXd::Identity(G_arena.rows(), inv_G.cols()))
                   * inv_G.adj_op().transpose() * inv_G.val_op()
                   * inv_G.val_op().transpose()
             + inv_G.val_op().transpose() * inv_G.val_op()
                   * inv_G.adj_op().transpose()
                   * (-inv_G.val_op() * G_arena.val_op()
                      + Eigen::MatrixXd::Identity(inv_G.rows(),
                                                  G_arena.cols()));
```

SteveBronder (changes requested):
This is looking nice! Just a few things and I think this will be ready to rock! I put up a branch here with a few examples of the changes for the first if() in the rev version

https://github.com/spinkney/math/compare/feature/ginv...stan-dev:review/gen-inv-2

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -0,0 +1,198 @@
...
+  if (G.rows() == G.cols())
+    return ret_type(inverse(G));
+
+  if (G.rows() < G.cols()) {
+    arena_t<VarMat> G_arena(G);
+    auto A_spd = tcrossprod(G_arena.val_op());
+    arena_t<VarMat> inv_G(
+        mdivide_left_spd(A_spd.val_op(), G_arena.val_op()).transpose());
+
+    arena_t<VarMat> res = inv_G;
```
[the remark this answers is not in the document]
spinkney: I was following the [inverse.hpp](https://github.com/stan-dev/math/blob/develop/stan/math/rev/fun/inverse.hpp) and there's something similar. On lines 31-33
```
  arena_t<T> arena_m = m;
  arena_t<promote_scalar_t<double, T>> res_val = arena_m.val().inverse();
  arena_t<ret_type> res = res_val;
```  
I see that the last line has `ret_type`. Do you know why `res_val` is repeated for `res`?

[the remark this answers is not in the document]
SteveBronder: That can be read as

```cpp
  arena_t<Eigen::Matrix<var, -1, -1>> arena_m = m;
  arena_t<Eigen::Matrix<double,  -1, -1>> res_val = arena_m.val().inverse();
  arena_t<Eigen::Matrix<var, -1, -1>> res = res_val;
```

So it's making the _values_ that are assigned to the output in `res_val`, but the actual vars for the output are not created tilll `res`. For yours you can read it as

```cpp
    arena_t<Eigen::Matrix<var, -1, -1>> inv_G(
        mdivide_left_spd(A_spd.val_op(), G_arena.val_op()).transpose());
    arena_t<Eigen::Matrix<var, -1, -1>> res = inv_G;
```

Where in `inv_G` you are making the full return type already. so there's no need to make res.

I think `inverse()` is making the values for the vars like that because it's trying to to an optimization


```cpp
 arena_t<promote_scalar_t<double, T>> res_val = arena_m.val().inverse();
  arena_t<ret_type> res = res_val;

  reverse_pass_callback([res, res_val, arena_m]() mutable {
    // notice res_val is called twice
    arena_m.adj() -= res_val.transpose() * res.adj_op() * res_val.transpose();
  });
```

It's because we use the values twice in `res_val`. but tbh idt this makes sense as an optimization because we're paying a big memory allocation but it's not like we get better instruction sets etc. by doing that. So I'd keep doing it the way your doing it, but just use `inv_G` and get rid of `res`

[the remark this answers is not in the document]
spinkney: thanks for the explanation, it's very helpful!

spinkney:
@SteveBronder ready for review with all the changes to use ldlt and remove the jitter.

test/unit/math/mix/fun/generalized_inverse_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+  EXPECT_NO_THROW(stan::math::generalized_inverse(z));
+
+  // autodiff throws, so following fails (throw behavior must match to pass)
+
+  Eigen::MatrixXd a(2, 2);
+  a << 1.9, 0.3, 0.3, std::numeric_limits<double>::infinity();
+  expect_ad(f, a);
+  expect_ad_matvar(f, a);
+
+  /*
```
spinkney, line 55: @SteveBronder I'm not sure of what to do with this. The finite diff blows up but the ad grad seems ok. Whenever the matrix is singular there's going to be these issues. Is there an appropriate check, warning, or test for the singular case?

bbbales2:
@bgoodri yo we're adding a generalized inverse here. Is there like a generalized solve that we should add with this? Like A x = b, we always prefer the solve to the inverse + multiply.

spinkney:
@bbbales2 before I forget we talked about this having a derivative that exists only when the matrix is of constant rank. Is there a good way to check this?

bgoodri:
> @bgoodri yo we're adding a generalized inverse here. Is there like a generalized solve that we should add with this? Like A x = b, we always prefer the solve to the inverse + multiply.

I don't think so. If A doesn't have a inverse, then there isn't a unique solution for x.

bbbales2:
> that exists only when the matrix is of constant rank

Can you point me at the paper for this?

spinkney:
> > that exists only when the matrix is of constant rank
> 
> Can you point me at the paper for this?
https://www.researchgate.net/publication/229057939_The_Differentiation_of_Pseudo-Inverses_and_Nonlinear_Least_Squares_Problems_Whose_Variables_Separate

bbbales2:
@spinkney Thanks, ping me if I don't get back to you on this within a couple days

test/unit/math/mix/fun/generalized_inverse_test.cpp
SteveBronder: I think a fine answer here is just to not test matrices where finite diff blows up (see #2211 where I had a similar issue).

SteveBronder (dismissed):
Besides deleting the commented out test I think this is good to go! Or @bbbales2 do you want to hold off on this until the constant rank Q is figured out? Would we add some additional test or something to check that?

bbbales2:
I don't think it's really possible to check for that. Let's just put a note in the functions doc ("For the derivatives of this function to exist..."). It'd be cool to have a reference that wasn't behind a paywall.

Like, if something needs to have constant rank, then if you wiggle the matrix around, you can't create any extra zero singular values.

So we could add a check for that, and it would be finicky because you're trying to check if something is zero. The problem is we wouldn't know the correct constant rank to begin with. Apparently this math all works for rank < min(M, N), so there could already be zero singular values in there.

bbbales2:
Thanks @spinkney !

spinkney:
@SteveBronder removed the unnecessary test and added a note about the derivative.

bbbales2:
@spinkney oh yeah, for that test where the Cholesky, if that's a situation where you have a matrix with rank < min(m, n), that's probably not a constant rank matrix at that point, and the test should change. All we have to do to prove this is move a tiny distance in any direction and change rank. Since we can move in any direction we want, I'm pretty sure we could achieve this.

SteveBronder (approved):
Cool and good!!!

SteveBronder:
@t4c1 your on the review button thing, did you want to take another look at this? It's pretty darn clean I like it!!

rok-cesnovar:
Docs PR for this is open, stanc3 is approved so only waiting on this to be merged.

t4c1 (approved):
Looks great.

Issue #2228, opened 2020-12-03
Use std_normal_lc(c)df functions for the ordered_probit distribution


andrjohns:
## Description

The ```ordered_probit``` distribution currently relies on calls to ```log(Phi(x))``` and ```log(Phi(x) - Phi(y))```. However the ```Phi``` function tends to over/underflow pretty easily. Thankfully, the addition of an ```std_normal_lcdf``` function with increased numerical precision in #1580 provides a robust, drop-in, alternative.

#### Current Version:
v3.4.0

andrjohns:
Closed by #2229

PR #2229, opened 2020-12-03
Replace log(Phi(x)) with std_normal_lcdf(x) in ordered_probit_lpmf

andrjohns:
## Summary

This PR replaces the use of ```log(Phi(x))``` with ```std_normal_lcdf(x)``` in ```ordered_probit_lpmf``` for increased numerical stability

## Tests

No new tests, as (minor) refactoring should produce the same results

## Side Effects

N/A

## Release notes

Updated log-likelihood calculation for ```ordered_probit_lpmf``` for increased numerical stability

## Checklist

- [x] Math issue #2228

### DIFF
--- a/stan/math/prim/prob/ordered_probit_lpmf.hpp
+++ b/stan/math/prim/prob/ordered_probit_lpmf.hpp
@@ -5,9 +5,11 @@
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/prim/fun/log.hpp>
-#include <stan/math/prim/fun/Phi.hpp>
+#include <stan/math/prim/fun/log_diff_exp.hpp>
 #include <stan/math/prim/fun/size_mvt.hpp>
 #include <stan/math/prim/fun/to_ref.hpp>
+#include <stan/math/prim/prob/std_normal_lcdf.hpp>
+#include <stan/math/prim/prob/std_normal_lccdf.hpp>
 #include <vector>
 #include <cmath>
 
@@ -84,12 +86,14 @@ return_type_t<T_loc, T_cut> ordered_probit_lpmf(const T_y& y,
     int K = c_vec[i].size() + 1;
 
     if (y_vec[i] == 1) {
-      logp_n += log1m(Phi(lambda_vec[i] - c_vec[i].coeff(0)));
+      logp_n += std_normal_lccdf(lambda_vec[i] - c_vec[i].coeff(0));
     } else if (y_vec[i] == K) {
-      logp_n += log(Phi(lambda_vec[i] - c_vec[i].coeff(K - 2)));
+      logp_n += std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(K - 2));
     } else {
-      logp_n += log(Phi(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 2))
-                    - Phi(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1)));
+      logp_n += log_diff_exp(
+                  std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 2)),
+                  std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1))
+                );
     }
   }
   return logp_n;
--- a/stan/math/prim/prob/ordered_probit_lpmf.hpp
+++ b/stan/math/prim/prob/ordered_probit_lpmf.hpp
@@ -92,8 +92,7 @@ return_type_t<T_loc, T_cut> ordered_probit_lpmf(const T_y& y,
     } else {
       logp_n += log_diff_exp(
                   std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 2)),
-                  std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1))
-                );
+                  std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1)));
     }
   }
   return logp_n;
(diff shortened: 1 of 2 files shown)


syclik:
Are there any tests for the change? Maybe add a value that overflows in the distribution test?

syclik:
The code looks fine, btw.

bbbales2 (changes requested):
Looks good. Was there a specific set of numbers where this work and `Phi` blows up? Can you add that as a prim test?

edit: missed an 'o' in good

edit: Whoops missed @syclik 's comment sorry

andrjohns:
Good call both! Tests added showing how it works where the old parameterisation failed

stan/math/prim/prob/ordered_probit_lpmf.hpp
```diff
@@ -84,12 +86,13 @@ return_type_t<T_loc, T_cut> ordered_probit_lpmf(const T_y& y,
     int K = c_vec[i].size() + 1;
 
     if (y_vec[i] == 1) {
-      logp_n += log1m(Phi(lambda_vec[i] - c_vec[i].coeff(0)));
+      logp_n += log1m_exp(std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(0)));
```
nhuurre, line 89: You could improve numerical stability even more by using `std_normal_lccdf` here ... except that `std_normal_lccdf` has much less robust implementation than `std_normal_lcdf` despite mathematically being the same function.
So I'm going to suggest instead
```suggestion
      logp_n += std_normal_lcdf(c_vec[i].coeff(0) - lambda_vec[i]);
```

andrjohns: Clever! Have pushed the change

test/unit/math/prim/prob/ordered_probit_test.cpp
```diff
@@ -17,6 +17,46 @@ stan::math::vector_d get_simplex_Phi(double lambda,
...
+  int K = c.size() + 1;
+
+  if (y == 1) {
+    return log1m(Phi(lambda - c(0)));
+  } else if (y == K) {
+    return log(Phi(lambda - c(K - 2)));
+  } else {
+    return log(Phi(lambda - c(y - 2)) - Phi(lambda - c(y - 1)));
+  }
+}
```
bbbales2, line 34: The problem with testing old code doesn't work is that isn't something we want to guarantee. If we replace `Phi` with `normal_lcdf` (and we probably should if the `normal_lcdf` numerics are better) then this code would probably start working and the test breaks.

Just as long as the tests prevent us from reintroducing the bug is good enough. It's kinda on the reviewer to do a little manual checking to make sure the old code was actually failing the tests (not something that needs included).

So delete the `old_ordered_probit` stuff. Apologies for not being clear originally on what more testing meant.

bbbales2: @andrjohns actually I just went ahead and pushed to your branch. So all you gotta tell me is thumbs up on my changes :D. I approve of this as is.

andrjohns: Huh that does save time, all good with me!

> If we replace Phi with normal_lcdf (and we probably should if the normal_lcdf numerics are better) then this code would probably start working and the test breaks.

In that case, would it be simpler if I included in this pull:

 - ```Phi``` redefined as ```exp(std_normal_lcdf)```
 - ```log_Phi``` introduced as an alias for ```std_normal_lcdf```
?

Or leave as-is and do in a separate pull?

bbbales2: Separate pull, and only if you want to do it.

A more stable `Phi` and making `normal_lccdf` use the same numerics as lcdf like @nhuurre pointed out would be good. It's probably worth checking `std_normal_lcdf` and `std_normal_lccdf` as well.

No need to add a `log_Phi`. Got enough functions as is :D.

stan/math/prim/prob/ordered_probit_lpmf.hpp
```diff
@@ -84,12 +86,13 @@ return_type_t<T_loc, T_cut> ordered_probit_lpmf(const T_y& y,
...
+      logp_n += std_normal_lcdf(c_vec[i].coeff(0) - lambda_vec[i]);
     } else if (y_vec[i] == K) {
-      logp_n += log(Phi(lambda_vec[i] - c_vec[i].coeff(K - 2)));
+      logp_n += std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(K - 2));
     } else {
-      logp_n += log(Phi(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 2))
-                    - Phi(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1)));
+      logp_n += log_diff_exp(
+          std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 2)),
+          std_normal_lcdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1)));
```
nhuurre, line 95: I've been thinking more about this because there's an alternative way to write this
```cpp
      logp_n += log_diff_exp(
          std_normal_lccdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 1)),
          std_normal_lccdf(lambda_vec[i] - c_vec[i].coeff(y_vec[i] - 2)));
```
but it's not obvious which is more numerically stable.
I don't think it makes much difference when lambda is between the cutpoints but when both cutpoints are lower than lambda the CCDF version is better (and when lambda is below both cutpoints the CDF version is better)
(And of course the code must use `*_cdf(-x)` instead of `*_ccdf(x)`)

andrjohns: I believe the ```std_normal_lcdf``` will be more stable here since the ```lccdf``` does not have the additional numerical approximations for boundary values

nhuurre: Well, yes. I wrote `lccdf` because that's how I was thinking about it when I derived the equality but you'd want to change that to a negated-argument `lcdf`.
```cpp
      logp_n += log_diff_exp(
          std_normal_lcdf(c_vec[i].coeff(y_vec[i] - 1) - lambda_vec[i]),
          std_normal_lcdf(c_vec[i].coeff(y_vec[i] - 2) - lambda_vec[i]));
```

andrjohns: Oh right, missed that sorry! Will push those changes now

bbbales2 (approved):

PR #2230, opened 2020-12-03
Added `var<mat>` implementation of `mdivide_left` (Issue #2101)

bbbales2:
Added `var<mat>` implementation of `mdivide_left` and some stuff to avoid computing qr decomposition twice (Issue #2101)

## Release notes

Added `var<mat>` implementation of `mdivide_left`

Updated `mdivide_left` to avoid doing QR decomposition in forward and reverse passes

## Checklist

- [x] Math issue #2101


stan/math/rev/core/chainable_object.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+  T& get() { return obj_; }
+};
+
+template <typename T>
+auto make_chainable_ptr(T&& obj) {
+  auto ptr = new chainable_object<plain_type_t<T>>(std::forward<T>(obj));
+  return ptr;
+}
+
```
bbbales2, line 31: @t4c1 @SteveBronder 

The goal here is to create ptrs to Eigen objects in the forward pass that will get deallocated with `stan::math::recover_memory` and `stan::math::recover_memory_nested`.

I made a template godbolt with this to see if it was forwarding stuff along but it has a lot of code in it and is confusing (presumably from the QR). Maybe it is useful to you all: https://godbolt.org/z/e1eMof

stan/math/rev/fun/mdivide_left.hpp
```diff
@@ -13,45 +13,65 @@
...
+
   if (!is_constant<T1>::value && !is_constant<T2>::value) {
     arena_t<promote_scalar_t<var, T1>> arena_A = A;
     arena_t<promote_scalar_t<var, T2>> arena_B = B;
-    arena_t<promote_scalar_t<double, T1>> arena_A_val = arena_A.val();
-    arena_t<ret_type> res = arena_A_val.householderQr().solve(arena_B.val());
-    reverse_pass_callback([arena_A, arena_B, arena_A_val, res]() mutable {
+
+    auto hqr_A_ptr = make_chainable_ptr(arena_A.val().householderQr());
+    arena_t<ret_type> res = hqr_A_ptr->get().solve(arena_B.val());
```
bbbales2, line 34: Here is where it gets created

stan/math/rev/fun/mdivide_left.hpp
```diff
@@ -13,45 +13,65 @@
...
+    reverse_pass_callback([arena_A, arena_B, hqr_A_ptr, res]() mutable {
       promote_scalar_t<double, T2> adjB
-          = arena_A_val.transpose().householderQr().solve(res.adj());
+          = hqr_A_ptr->get().householderQ()
+            * hqr_A_ptr->get()
+                  .matrixQR()
+                  .template triangularView<Eigen::Upper>()
+                  .transpose()
+                  .solve(res.adj());
```
bbbales2, line 42: Here is where it gets used

bbbales2 (commented):
Comments on new `make_chainable_ptr` function

stan/math/rev/core/chainable_object.hpp
bbbales2: Questions are:

1. Does this pattern look okay?
2. Am I duplicating something that exists elsewhere?

If the answers are yes/no, then I'll go add tests and comments and whatnot.

stan/math/rev/core/chainable_object.hpp
```diff
@@ -0,0 +1,66 @@
...
+  template <typename S,
+            require_same_t<plain_type_t<T>, plain_type_t<S>>* = nullptr>
+  chainable_object(S&& obj) : obj_(std::forward<S>(obj)) {}
+
+  /**
+   * Return a reference to the underlying object
+   *
+   * @return reference to underlying object
+   */
+  inline T& get() noexcept { return obj_; }
```
bbbales2, line 47: I tried adding `inline const T& get() noexpect const { return obj_; }` but got an error.

stan/math/rev/core/chainable_object.hpp
bbbales2: Talked to Steve. Tests added.

t4c1: That is a nice idea. I like it. However I am not sure if we always want to store only plain types. I guess that should be up to the user of this?

bbbales2: > I am not sure if we always want to store only plain types. I guess that should be up to the user of this?

How about I replace `plain_type_t` with `std::decay`?

t4c1: That should be fine.

bbbales2: @t4c1 I was using this and I realized that maybe I should just return a pointer to the underlying object.

Like instead of:

```
return new chainable_object<plain_type_t<T>>(std::forward<T>(obj));
```

```
return &(new chainable_object<plain_type_t<T>>(std::forward<T>(obj)))->get();
```

So that way nobody has to deal with `chainable_objects` directly.

bbbales2: Or what if this was the default implementation of `to_arena`?

```
return (new chainable_object<plain_type_t<T>>(std::forward<T>(obj)))->get();
```

So then if something can't be shipped off to the arena, `to_arena` has the effect of giving it arena memory lifetime.

bbbales2: Wait I guess the thing about `arena_t` is that it gives you lightweight variables that you are cheap to copy.

A reference to the underlying object is not cheap or safe to copy.

bbbales2: I changed this to return a pointer to the underlying object (instead of the `chainable_object` itself)

bbbales2:
@t4c1 This is ready for reviewin' and mergin'

stan/math/rev/core/chainable_object.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * in a double free (`obj_` will get destructed once when the
+ * chainable_object leaves scope and once when the chainable
+ * stack memory is recovered).
+ *
+ * @tparam T type of object to hold
+ */
+template <typename T>
+class chainable_object : public chainable_alloc {
+ private:
+  T obj_;
```
SteveBronder, line 29: I'd put `std::decay_t<T>` here since I think we always want the object to hold a non-reference type

SteveBronder: Or just `plain_type_t<T>`

stan/math/rev/core/chainable_object.hpp
SteveBronder: This seemed to work for me

```
inline auto& get() const noexcept { return obj_; }
```

stan/math/rev/fun/mdivide_left.hpp
SteveBronder: I like it!

SteveBronder (changes requested):
One small change related to `T` in `chainable_object` but besides that cool and good!

stan/math/rev/core/chainable_object.hpp
bbbales2: Talked to @t4c1 about this earlier: https://github.com/stan-dev/math/pull/2230#discussion_r535910828 and we went with decay (also we did this at the `make_chainable_ptr` level).

But now that you bring it up again, I think I want to reverse the decision and do `plain_type`.

Maybe `chainable_object` can have a `decay` in it, but when I do `make_chainable_ptr(x)`, I want to make it so that `x` is available on the reverse-mode pass, whatever `x` is, without worrying about the template type (that's why I made the function). Much easier to avoid mistakes here if I just `plain_type` it.

I think I could make a use where I need to save expressions that reference arena memory, but then I could just copy the expression by value and not need this at all.

stan/math/rev/core/chainable_object.hpp
bbbales2: Cool I have two versions now and it seems to work.

stan/math/rev/core/chainable_object.hpp
t4c1: Fine, I guess. I can't really think of an example where we would want to store a non-plain type.

bbbales2: Hahaha, my laziness wins. `plain_type_t` is so nice cause then I don't have to think.

bbbales2:
@SteveBronder ready to go again

bbbales2:
@SteveBronder ready to go

SteveBronder (approved):
Cool!

PR #2234, opened 2020-12-04
Update README.md

spinkney:
Eigen version is 3.3.7


spinkney:
This should also be updated on https://mc-stan.org/math/

rok-cesnovar (approved):
Thanks! 

> This should also be updated on https://mc-stan.org/math/

That happens automatically when updating README.md (our CI Jenkins takes care of that).

Issue #2245, opened 2020-12-07
Cleaning up linear access functions


bbbales2:
Now that we're exposing expressions in the language, we need to clean up functions that depend on linear access in Eigen types.

Specifically, with a simple Eigen type like `Eigen::MatrixXd`, it is possible to instead of looping over ij like so:
```cpp
Eigen::MatrixXd A;
for(size_t i = 0; i < A.size(); ++i) {
  for(size_t i = 0; j < A.size(); ++j) {
    ... A(i, j) ...;
  }
}
```

to just do:
```cpp
Eigen::MatrixXd A;
for(size_t i = 0; i < A.size(); ++i) {
  ... A(i) ...;
}
```

This was convenient because the same thing worked with matrices, vectors, row vectors.

Now that we have expressions and maps and such we need to be more careful.

There have been two pulls addressing issues like this so far, #2205 and #2223

I made this issue so we would vaguely have a place to discuss:

1. How we should be testing for this so it doesn't bite us
2. Track known issues that exist (until there's some sort of general solution in place)
3. How do we handle row major vs. column major -- now our double loops will be explicitly bad in some cases

And and I do not want to stall #2223 unnecessarily to try to fix everything in one go.

#### Current Version:
v3.4.0

t4c1:
> How we should be testing for this so it doesn't bite us
I already added testing to expression testing framework. Although varmat functions are not tested using it (yet?).
> How do we handle row major vs. column major -- now our double loops will be explicitly bad in some cases
The simple option is to replace loops with expressions. The faster option is to have both row-major and col-major loops and pick depending if more expressions are row or col major.

bbbales2:
> I already added testing to expression testing framework. Although varmat functions are not tested using it (yet?).

Oh is that testing 2x2 matrices and the gradients and stuff? I remember merging the block tests, but I assumed that was more a compile thing.

> The simple option is to replace loops with expressions.

So this should be our go to? And then we add `require` types so we can detect the row/col majorness?

t4c1:
> Oh is that testing 2x2 matrices and the gradients and stuff? I remember merging the block tests, but I assumed that was more a compile thing.

It is actually a simple test on 1x1 matrices, but if it compiles it should work.

> So this should be our go to?

I will try to prepare a general framework that will pick row/col loops depending on expressions. Although it is not that high on my todo. Maybe in a week or so.

bbbales2:
[...]

PR #2247, opened 2020-12-08
Multi expression

t4c1:
## Summary

Added tools for efficiently evaluating multiple Eigen expressions at the same time. The interface is similar to kernel geenrator.

## Tests

Added tests.

## Side Effects
This prevents Eigen's vectorization. Shouldn't be too much of an issue, as most of the cases we would want to use this for, are memory bound anyway. Also the compiler can still vectorize the code.

## Release notes

Added tools for efficiently evaluating multiple Eigen expressions at the same time.

## Checklist

- [ ] Math issue #2245


t4c1:
@bbbales2 This should work for all cases where we want to evaluate multiple expressions at the same time without explicit looping and be as fast as the fastest combination of loops. However, it might be worth checking with the actual use cases.

bbbales2:
So this is how we can safely and efficiently handle rowmajor/colmajor/blocks/etc. without loops?

t4c1:
Yes.

bbbales2:
This looks cool. Can give a closer look tomorrow.

@SteveBronder should have a look at this too.

SteveBronder:
Yes I'll have a look at this as well sounds neat!

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+      for (size_t j = 0; j < inner_dimension; j++) {
+        index_apply<sizeof...(T_results)>([&](auto... Is) {
+          static_cast<void>(std::initializer_list<int>{
+              ((Eigen::internal::traits<std::decay_t<std::tuple_element_t<
+                            Is, std::tuple<T_expressions...>>>>::Flags
+                        & Eigen::RowMajorBit
+                    ? std::get<Is>(results_).coeffRef(j, i)
+                      = std::get<Is>(expressions.exprs_).coeff(j, i)
+                    : std::get<Is>(results_).coeffRef(i, j)
+                      = std::get<Is>(expressions.exprs_).coeff(i, j)),
```
bbbales2, line 129: `j` is always the inner dimension, so I don't think there needs to be this check in here.

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+ * @tparam T_expressions types of exprs
+ */
+template <typename... T_expressions>
+class eigen_expressions_ {
+ public:
+  /**
+   * Constructor.
+   * @param exprs expresions that will be calculated in same loop.
+   * @throw invalid_argument expressions have different sizes (for row major
+   * expressions rows and columns are swapped for the check)
```
bbbales2, line 32: Why does this size swapping make sense?

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+              ^ first_flags)
+             & Eigen::RowMajorBit)
+                ? check_matching_dims("eigen_expressions_.eigen_expressions_",
+                                      "first expression", std::get<0>(exprs_),
+                                      "transposed expression",
+                                      std::get<Is + 1>(exprs_).transpose())
+                : check_matching_dims("eigen_expressions_.eigen_expressions_",
+                                      "first expression", std::get<0>(exprs_),
+                                      "expression", std::get<Is + 1>(exprs_))),
+           0)...});
```
bbbales2, line 51: Where do expressions get evaluated here? Or are they only implicitly evaluated in the assign?

bbbales2 (dismissed):
Couple questions

stan/math/prim/functor/multi_expression.hpp
t4c1: `coeff` arguments are `(row, col)`, not `(inner, outer)`, so I think this is correct.

stan/math/prim/functor/multi_expression.hpp
t4c1: Because in loops we iterate over inner and outer dimension - so here we are not comparing if the expressions have the same number of rows and cols, but actually if they have the same inner and outer dimension. Does that make sense?

stan/math/prim/functor/multi_expression.hpp
t4c1: Yeah in asign. It doesnot evaluate whole expression at once. Instead it loops over elements, calculating them one by one.

stan/math/prim/functor/multi_expression.hpp
bbbales2: WOof, yeah

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+   */
+  template <bool Linear, typename... T_expressions,
+            std::enable_if_t<Linear>* = nullptr>
+  void assign(const eigen_expressions_<T_expressions...>& expressions) {
+    for (size_t i = 0; i < std::get<0>(expressions.exprs_).size(); i++) {
+      index_apply<sizeof...(T_results)>([&](auto... Is) {
+        static_cast<void>(std::initializer_list<int>{
+            (std::get<Is>(results_).coeffRef(i)
+             = std::get<Is>(expressions.exprs_).coeff(i),
+             0)...});
```
SteveBronder, line 96: Should we just use the foreach thing I wrote a while ago for this? That at least let's you get rid of the static_cast etc. here

https://gist.github.com/SteveBronder/410f7127f8be9dced2b1915765fd8644

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+      std::forward<T_expressions>(exprs)...);
+}
+
+/**
+ * Represents results that will be calculated in same loop.
+ * @tparam T_results types of results
+ */
+template <typename... T_results>
+class eigen_results_ {
+  std::tuple<T_results...> results_;
```
SteveBronder, line 80: Should this object's tuple hold actual types or just references?

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+ public:
+  /**
+   * Constructor.
+   * @param exprs expresions that will be calculated in same loop.
+   * @throw invalid_argument expressions have different sizes (for row major
+   * expressions rows and columns are swapped for the check)
+   */
+  explicit eigen_expressions_(T_expressions&&... exprs)
+      : exprs_(std::forward<T_expressions>(exprs)...) {
+    index_apply<sizeof...(T_expressions) - 1>([&](auto... Is) {
```
SteveBronder, line 36: [optional] should we have something like a static assert here checking these are all Eigen types? That would just let it fail with a nice error message if any non-eigen types are brought in

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+                                      "first expression", std::get<0>(exprs_),
+                                      "expression", std::get<Is + 1>(exprs_))),
+           0)...});
+    });
+  }
+
+ private:
+  std::tuple<T_expressions...> exprs_;
+  template <typename... T_results>
+  friend class eigen_results_;
```
SteveBronder, line 58: stuff ending in `_` is reversed for compiler

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+        index_apply<sizeof...(T_results)>([&](auto... Is) {
+          static_cast<void>(std::initializer_list<int>{
+              ((Eigen::internal::traits<std::decay_t<std::tuple_element_t<
+                            Is, std::tuple<T_expressions...>>>>::Flags
+                        & Eigen::RowMajorBit
+                    ? std::get<Is>(results_).coeffRef(j, i)
+                      = std::get<Is>(expressions.exprs_).coeff(j, i)
+                    : std::get<Is>(results_).coeffRef(i, j)
+                      = std::get<Is>(expressions.exprs_).coeff(i, j)),
+               0)...});
```
SteveBronder, line 130: Kind of confused on the row major checking. Is this assuming if the expression is row/col major then the result is also row/col major?

Are there things in Eigen that are row major by default? The only thing i can think of where we would want mixed row/col major support would be something like sparse matrices

(on a side note getting these to work with sparse matrices would be cool because if we used them in function then iterators would also support sparse stuff. but that can happen later)

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+ * @tparam T_expressions types of exprs
+ * @param exprs exprs that will be used in same kernel.
+ */
+template <typename... T_expressions,
+          require_all_eigen_t<T_expressions...>* = nullptr>
+eigen_expressions_<T_expressions...> eigen_expressions(
+    T_expressions&&... exprs) {
+  return eigen_expressions_<T_expressions...>(
+      std::forward<T_expressions>(exprs)...);
+}
```
SteveBronder, line 72: Oh nvm so the tuple will hold references. We should probably put both the expression and result classes into the internal namespace.

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,196 @@
...
+             & (Eigen::LinearAccessBit | Eigen::DirectAccessBit))...});
+    constexpr int N_row_major = internal::static_sum(
+        static_cast<bool>(
+            Eigen::internal::traits<std::decay_t<T_expressions>>::Flags
+            & Eigen::RowMajorBit)...,
+        static_cast<bool>(
+            Eigen::internal::traits<std::decay_t<T_results>>::Flags
+            & Eigen::RowMajorBit)...);
+    constexpr int N_col_major
+        = sizeof...(T_results) + sizeof...(T_expressions) - N_row_major;
```
SteveBronder, line 169: Some of the traits are confusing Eigen internal things. Can you update the docs just detailing what each of these enums values we checks against means?

SteveBronder (changes requested):
Looks neat! A few extra Qs

1. Should we file an issue with Eigen just telling them about this and asking if they want us to backport it? It would only be supported for C++11 but they might still be interested.

2. We should probably have an `update()` function as well that does the same as `result()` but with `+=`

3. Can you update something like `elt_multiply` and check the performance diff?

4. I couldn't tell from the code, how does this handle expression on the right hand side? would it make more sense to evaluate some expressions on the right hand side before assignment?

stan/math/prim/functor/multi_expression.hpp
t4c1: We could just as well, although the difference is not that big. Also I think that foreach is not already in the repo, right?

stan/math/prim/functor/multi_expression.hpp
t4c1: It uses perfect forwarding - for rvalues it holds objects and for lvalues it holds references.

stan/math/prim/functor/multi_expression.hpp
t4c1: Wrong. Compiler only reserves names containing two consecutive underscores or starting with an underscore plus uppercase letter. See for example: https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

stan/math/prim/functor/multi_expression.hpp
t4c1: Hmm, do we? It is still a type user works with. Although I believe explicitly declaring the type would be quite rare. Also for example in kernel generator we have many similar classes that are not in in internal namespace.

stan/math/prim/functor/multi_expression.hpp
t4c1: > Is this assuming if the expression is row/col major then the result is also row/col major?

No. It is just choosing whether to go over the values in row or col major order depending on the expression. If one of the (result, expression) is row major and another is col major it does not matter. If both have the same storage order one of these options is much faster (but both would produce the correct result).

> Are there things in Eigen that are row major by default?

Return type of transposing any col major expression comes to mind. There might also be others.

stan/math/prim/functor/multi_expression.hpp
t4c1: Do we maintain some docs for Eigen internals? Or where would you like me to put that?

t4c1:
> Should we file an issue with Eigen just telling them about this and asking if they want us to backport it? It would only be supported for C++11 but they might still be interested.

Go ahead if you want. I might be wrong, but when I asked about something like this on SO they did not seem too interested.

> We should probably have an update() function as well that does the same as result() but with +=

I am not sure how we would implement that. If we support `+=`, wo probably also want `*=`, `-=` and others. But when working with multiple expressions they might use different compound assignment operators.

> I couldn't tell from the code, how does this handle expression on the right hand side? 

I don't understand what is unclear. This whole thing is about handling expressions. It handles them on both sides of assignment.

> would it make more sense to evaluate some expressions on the right hand side before assignment?

Yeah, for example matrix multiplications. However I don't know how to detect and handle expressions containing matrix multiplications.  We will have to be careful not to use this for such expressions. I will add some docs about that.

stan/math/prim/functor/multi_expression.hpp
t4c1: There are requires on helper functions that are used for constructing `expressions_` and `results_` objects. In practice constructions will always go through these helpers.

t4c1:
Benchmark results - essentially no difference in performance:

before:
```
-----------------------------------------------------------------------------------
Benchmark                                            Time           CPU Iterations
-----------------------------------------------------------------------------------
elt_multi_bench/1/manual_time                      102 ns        274 ns    7004020
elt_multi_bench/4/manual_time                      115 ns        314 ns    5471918
elt_multi_bench/16/manual_time                     208 ns        468 ns    3140967
elt_multi_bench/64/manual_time                     597 ns       1246 ns    1228777
elt_multi_bench/256/manual_time                   2143 ns       4515 ns     304533
elt_multi_bench/1024/manual_time                 10780 ns      20728 ns      69349
elt_multi_bench/4096/manual_time                 53346 ns      95119 ns      13470
elt_multi_bench/16384/manual_time               240715 ns     429938 ns       2653
elt_multi_bench/65536/manual_time              1618980 ns    2717391 ns        414
elt_multi_bench/262144/manual_time             7979663 ns   12369792 ns         96
elt_multi_bench/1048576/manual_time           32700324 ns   49851190 ns         21
elt_multi_bench/4194304/manual_time          126429750 ns  199218750 ns          4
elt_multi_bench/16777216/manual_time         713578100 ns 1078125000 ns          1
elt_multi_bench/67108864/manual_time        2781297000 ns 4250000000 ns          1
elt_multi_varmat_bench/1/manual_time                48 ns         92 ns   14777157
elt_multi_varmat_bench/4/manual_time                73 ns        119 ns   10778583
elt_multi_varmat_bench/16/manual_time               94 ns        152 ns    7295999
elt_multi_varmat_bench/64/manual_time              150 ns        243 ns    4700839
elt_multi_varmat_bench/256/manual_time             462 ns        651 ns    1536031
elt_multi_varmat_bench/1024/manual_time           2234 ns       3327 ns     300611
elt_multi_varmat_bench/4096/manual_time           8816 ns      13652 ns      82408
elt_multi_varmat_bench/16384/manual_time         45813 ns      72142 ns      15161
elt_multi_varmat_bench/65536/manual_time        220314 ns     334744 ns       2894
elt_multi_varmat_bench/262144/manual_time      1909291 ns    3032345 ns        371
elt_multi_varmat_bench/1048576/manual_time     8349009 ns   13251582 ns         79
elt_multi_varmat_bench/4194304/manual_time    33691162 ns   53571429 ns         21
elt_multi_varmat_bench/16777216/manual_time  132514240 ns  212500000 ns          5
elt_multi_varmat_bench/67108864/manual_time  572634700 ns  937500000 ns          1
```

This PR:
```
-----------------------------------------------------------------------------------
Benchmark                                            Time           CPU Iterations
-----------------------------------------------------------------------------------
elt_multi_bench/1/manual_time                      111 ns        294 ns    5741333
elt_multi_bench/4/manual_time                      139 ns        337 ns    5694798
elt_multi_bench/16/manual_time                     226 ns        509 ns    2824365
elt_multi_bench/64/manual_time                     714 ns       1283 ns     986537
elt_multi_bench/256/manual_time                   2685 ns       4872 ns     250157
elt_multi_bench/1024/manual_time                 11549 ns      21871 ns      61441
elt_multi_bench/4096/manual_time                 54550 ns      98778 ns      12971
elt_multi_bench/16384/manual_time               238610 ns     451325 ns       2735
elt_multi_bench/65536/manual_time              1773733 ns    2932198 ns        389
elt_multi_bench/262144/manual_time             8229944 ns   12718023 ns         86
elt_multi_bench/1048576/manual_time           33126086 ns   51339286 ns         21
elt_multi_bench/4194304/manual_time          132920900 ns  207031250 ns          4
elt_multi_bench/16777216/manual_time         727554000 ns 1109375000 ns          1
elt_multi_bench/67108864/manual_time        2774295000 ns 4281250000 ns          1
elt_multi_varmat_bench/1/manual_time                55 ns         99 ns   10453971
elt_multi_varmat_bench/4/manual_time                77 ns        124 ns    8542106
elt_multi_varmat_bench/16/manual_time              136 ns        212 ns    5388093
elt_multi_varmat_bench/64/manual_time              199 ns        312 ns    3303963
elt_multi_varmat_bench/256/manual_time             622 ns        863 ns    1321720
elt_multi_varmat_bench/1024/manual_time           2408 ns       3569 ns     280190
elt_multi_varmat_bench/4096/manual_time          11306 ns      15767 ns      56485
elt_multi_varmat_bench/16384/manual_time         51767 ns      79688 ns      10000
elt_multi_varmat_bench/65536/manual_time        217150 ns     341587 ns       3019
elt_multi_varmat_bench/262144/manual_time      1852866 ns    3046657 ns        359
elt_multi_varmat_bench/1048576/manual_time     8461653 ns   12987013 ns         77
elt_multi_varmat_bench/4194304/manual_time    33360590 ns   52827381 ns         21
elt_multi_varmat_bench/16777216/manual_time  139114820 ns  218750000 ns          5
elt_multi_varmat_bench/67108864/manual_time  569633700 ns  953125000 ns          1
```

stan/math/prim/functor/multi_expression.hpp
t4c1: I found the Eigen doc for these flags: https://eigen.tuxfamily.org/dox/group__flags.html

bbbales2:
There must be something going on still:

```
old: elt_multi_varmat_bench/16384/manual_time         45813 ns      72142 ns      15161
new: elt_multi_varmat_bench/16384/manual_time         51767 ns      79688 ns      10000
```

Or does this difference go away if you run it again?

SteveBronder:
Thinking about this more, the problem we are trying to solve here is that single indexing on a non-linear index'd matrix can give an error.

```cpp
// potentially bad!
for (Eigen::Index j = 0; j < x.size(); ++j) {
  x.adj().coeffRef(i) += res.val().coeffRef(i) * blah1;
  y.adj().coeffRef(i) -= res.val().coeffRef(i) - x.val().coeffRef(i);
}
```

if we had C++17 it would be easy to write something like

```cpp
for (auto& [x_val, x_adj, y_val, y_adj, res_val, res_adj] : zip(x, y, res)) {
   x_adj += res_val * blah1;
   y_adj -= res_val - x_val;
}
```

where we could have zip handle all the indexing stuff. with c++14 I think the nicest we could come up with would be something like

```cpp
for (auto& iters: zip(x, y, res)) {
   iters.adj<0>() += iters.val<2>() * blah1;
   iters.adj<1>() -= iters.val<2>() - iters.val<0>()
}
```

But that's kind of nasty tbh. I think the benchmarks look fine / make sense. They're a little slower for small N because this has a bit of overhead but I'm fine with that if it let's us be more defensive. I'm going to run them with multiple iterations overnight just to sanity check them. 

>> would it make more sense to evaluate some expressions on the right hand side before assignment?

> Yeah, for example matrix multiplications. However I don't know how to detect and handle expressions containing matrix multiplications. We will have to be careful not to use this for such expressions. I will add some docs about that.

What if we just stopped it from accepting matrix expressions at all? I think the goal here is just to have a nice iterator over arrays. What do you think about that? If we go that route I wouldn't mind just having a static assert in the make function that prints out a friendly error message letting the dev know "hey you tried bringing in a matrix but you this only accepts array like types."

t4c1:
> What do you think about that?

That just hides the problem. The user could still write `(a* b).array()` and we would not detect it. I don't think we can do better than just docing it.

stan/math/rev/fun/elt_multiply.hpp
```diff
@@ -33,11 +34,10 @@ auto elt_multiply(const Mat1& m1, const Mat2& m2) {
...
     reverse_pass_callback([ret, arena_m1, arena_m2]() mutable {
-      for (Eigen::Index i = 0; i < arena_m2.size(); ++i) {
-        const auto ret_adj = ret.adj().coeffRef(i);
-        arena_m1.adj().coeffRef(i) += arena_m2.val().coeff(i) * ret_adj;
-        arena_m2.adj().coeffRef(i) += arena_m1.val().coeff(i) * ret_adj;
-      }
+      auto ret_adj = ret.adj().array();
+      eigen_results(arena_m1.adj(), arena_m2.adj()) = eigen_expressions(
+          arena_m1.adj().array() + arena_m2.val().array() * ret_adj,
+          arena_m2.adj().array() + arena_m1.val().array() * ret_adj);
```
SteveBronder, line 40: I'm not sure how I feel about this. What if we did something like tag dispatching?

```cpp
      eigen_results(std::tuple<std::plus, std::plus>(), arena_m1.adj(), arena_m2.adj()) = eigen_expressions(
           arena_m2.val().array() * ret_adj,
           arena_m1.val().array() * ret_adj);
```

SteveBronder: The common case is going to be `+=` so what if we just did that by default and users would just do a minus in `eigen_expression`?

t4c1: I dont like tag dispatch. It is verbose. But I agreee += will be used a lot. I will add an option for `eigen_results() += eigen_expressions()`.

SteveBronder: This might be kind of silly but what if we had `operator+-=`, `operator-+=` for these? That looks kind of wacky, I can think of few situations where we do more than two. 

```cpp
      eigen_results(arena_m1.adj(), arena_m2.adj()) +-= eigen_expressions(
           arena_m2.val().array() * ret_adj,
           arena_m1.val().array() * ret_adj);
```

Though maybe that's two weird and a nicer thing is just to have a minus on the expr you want like

```cpp
      eigen_results(arena_m1.adj(), arena_m2.adj()) += eigen_expressions(
           arena_m2.val().array() * ret_adj,
           -(arena_m1.val().array() * ret_adj));
```

SteveBronder:
One other Q, if we made these `result()` and `expression()` would they work with `matrix_cl<>`? Because I think a handful of functions that would use this could be made to support both

stan/math/rev/fun/elt_multiply.hpp
t4c1: AFAIK there is no such thing as operator+-= in c++. So we have to go with the second option.

t4c1:
We already have such functions for `matrix_cl` and kernel generator expressions - they are named `results_cl` and `expressions_cl`. If we renamed both we could have same interface for both, although OpenCL does not support += yet.

t4c1:
Is this good to go?

t4c1:
@SteveBronder @bbbales2 Is this good to go?

bbbales2:
I was leaving this to @SteveBronder. The design looks good to me and though I didn't dig into the code details.

stan/math/rev/fun/elt_multiply.hpp
```diff
@@ -33,13 +34,10 @@ auto elt_multiply(const Mat1& m1, const Mat2& m2) {
...
-        for (Eigen::Index i = 0; i < arena_m2.rows(); ++i) {
-          const auto ret_adj = ret.adj().coeffRef(i, j);
-          arena_m1.adj().coeffRef(i, j) += arena_m2.val().coeff(i, j) * ret_adj;
-          arena_m2.adj().coeffRef(i, j) += arena_m1.val().coeff(i, j) * ret_adj;
-        }
-      }
+      auto ret_adj = ret.adj().array();
+      eigen_results(arena_m1.adj().array(), arena_m2.adj().array())
+          += eigen_expressions(arena_m2.val().array() * ret_adj,
+                               arena_m1.val().array() * ret_adj);
```
SteveBronder, line 40: Should we do a sanity check in `eigen_results` to make sure there are not expressions that get passed to `eigen_results`

SteveBronder: Also I'm still trying to think about how to stop users of this from calling something like `arena_m2.val() * arena_m2.val()`. 

I'm playing around with the below, is there anything else we would not want to have in `eigen_expressions()` besides products? decompositions? Do you think there is some way we can detect whether an expression has a `Eigen::Product<>` in it's template parameters? The only thing I could think of was some recursive type trait that would keep peeking through the template params. I just really worry that can become a bad footgun if we are not paying attention

https://godbolt.org/z/efTfjY

SteveBronder (changes requested):
Apologies for the delay. Mostly just template Qs and I'm still worried about people passing matrix products into this

stan/math/rev/fun/elt_multiply.hpp
bbbales2: > how to stop users of this from calling something like arena_m2.val() * arena_m2.val().

If it's not easy to do something like this, I'm okay leaning on review + benchmarks (when we get them) to catch these things. Always a chance we accidentally code up really slow stuff that passes tests.

SteveBronder: Yeah it would be nice to at least catch the simpler cases. @t4c1 if you spend 20 mins goofing around to find a nice solution to this and can't make one then I'm fine with leaving it as is

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,223 @@
...
+class eigen_expressions_ {
+ public:
+  /**
+   * Constructor.
+   * @param exprs expresions that will be calculated in same loop.
+   * @throw invalid_argument expressions have different sizes (for row major
+   * expressions rows and columns are swapped for the check)
+   */
+  explicit eigen_expressions_(T_expressions&&... exprs)
+      : exprs_(std::forward<T_expressions>(exprs)...) {
```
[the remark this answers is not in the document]
t4c1: This is just using it wrong. `callee` fucntion should use perfect forwarding.

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,223 @@
...
+                                      std::get<Is + 1>(exprs_).transpose())
+                : check_matching_dims("eigen_expressions_.eigen_expressions_",
+                                      "first expression", std::get<0>(exprs_),
+                                      "expression", std::get<Is + 1>(exprs_))),
+           0)...});
+    });
+  }
+
+ private:
+  std::tuple<T_expressions...> exprs_;
```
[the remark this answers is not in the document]
t4c1: This is like the textbook example of using perfect forwarding in a class. I am sure this is fine. 

The problem with simply storing references is we need them to be non-const to support assignment to lvalue expressions. And we can not use non-const lvalue references for rvalue expressions. So we need perfect forwarding.

stan/math/prim/functor/multi_expression.hpp
```diff
@@ -0,0 +1,223 @@
...
+ * Deduces types for constructing \c expressions_ object.
+ * @tparam T_expressions types of exprs
+ * @param exprs exprs that will be used in same kernel.
+ */
+template <typename... T_expressions,
+          require_all_eigen_t<T_expressions...>* = nullptr>
+eigen_expressions_<T_expressions...> eigen_expressions(
+    T_expressions&&... exprs) {
+  return eigen_expressions_<T_expressions...>(
+      std::forward<T_expressions>(exprs)...);
```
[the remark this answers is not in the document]
t4c1: No. that problem only happens when not using forward or move.

stan/math/rev/fun/elt_multiply.hpp
t4c1: I changed this to use Eigen's evaluators. This way we efficiently handle expressions that are better off being evaluated separately (such as matrix product). In most cases this should be no better than evaluating such expressions separately, but it should also not be slower.

SteveBronder (approved):
Lgtm!

Issue #2249, opened 2020-12-09
Use immedietly invoked lambdas to make error checking less expensive


SteveBronder:
## Description

Something kind of neat I was reading the other day

https://rigtorp.se/iife/

the tl;dr is that using some compiler attributes and immediately invoked lambdas we can decrease the size of the error functions, and since those error paths take up less space it can allow for better inlining. This is also nice because we don't really care about performance when a function errors so it makes sense to put that code on a cold path.

## Example

I have an example below on godbolt that has our current version of `check_range()` and a version that uses the immediately invoked lambdas and some compiler attributes. You can ctrl+f for "runner" to see the main change which is that the block of asm in compiler #2 (`.L28`) has the tag `[clone .cold]` which means those instructions are never pre-fetched. for compiler #1 the block of asm it jumps to (`.L50`) is going to be pre-fetched because of the CPUs branch prediction.

https://gcc.godbolt.org/z/6zWTcY

I'm not sure how much these would save, but we do call these functions a lot.

## Expected Output

Should be the same

#### Current Version:
v3.4.0

t4c1:
How portable between compilers are these attributes?

Do you have any example model, or at least microbenchmark using some semi-realistic set of function calls that is showing any significant speedup from this?

SteveBronder:
As portable as `likely` aka we need to wrap it in 

```
#ifdef __GNUC__
#define likely(x) __builtin_expect(!!(x), 1)
#define unlikely(x) __builtin_expect(!!(x), 0)
#else
#define likely(x) (x)
#define unlikely(x) (x)
#endif
```

Here's a benchmark just calling the `check_range` N times (`check_range` is the new style and `check_range2` is our current.)

https://gist.github.com/SteveBronder/84ba7d4bfb0f5ded4289f3b702f6b940

```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
check_range_bench/2             2.30 ns         2.30 ns    305371235
check_range_bench/4             4.59 ns         4.59 ns    152487809
check_range_bench/8             9.18 ns         9.18 ns     75119108
check_range_bench/16            18.4 ns         18.4 ns     37927216
check_range_bench/32            37.6 ns         37.6 ns     18620833
check_range_bench/64            74.3 ns         74.3 ns      9278324
check_range_bench/128            148 ns          148 ns      4744552
check_range_bench/256            295 ns          295 ns      2376836
[...]

PR #2255, opened 2020-12-11
use immediately invoked lambdas in size and range error checks

SteveBronder:
## Summary

See issue #2249 for technical info. This uses immediately invoked lambdas in size and range error checks. This makes the actual code for the error checks smaller which should help the compiler when attempting to inline these.

This also adds some useful compiler attribute [macros](https://github.com/stan-dev/math/pull/2255/files#diff-00334c9b2e65ea32a277e06c9e33a27576b93d435622ac6cad7c868ad53b745e). One to call out in particular is `STAN_REMOVE_RANGE_AND_SIZE_CHECKS` which is the user passes in at compile time will turn off all range and size checks

## Tests

Refactor so no new tests

## Side Effects

We should generally do this pattern in all the error checks so in a future PR would be good to do that

## Release notes

Use immediately invoked lambdas in size and range error checks

## Checklist

- [x] Math issue #2249


bbbales2:
I'm not in for the `STAN_NO_RANGE_AND_SIZE_CHECK` macros. I see the use for experimenting, but it just seems like another code branch to maintain.

The COLD_PATH stuff... Wow what a speedup. Looks worth it

SteveBronder:
In totally fine with removing the no check macro. It was mostly just an idea

stan/math/prim/err/check_matching_dims.hpp
```diff
@@ -26,20 +26,11 @@ namespace math {
...
-  } else {
-    for (int i = 0; i < y1_d.size(); i++) {
-      if (y1_d[i] != y2_d[i]) {
-        error = true;
-        break;
-      }
-    }
-  }
-  if (error) {
+  auto error_throw = [&]() STAN_COLD_PATH {
```
t4c1, line 33: In this function this should not be in lambda and cold path. This is the check itself, which is executed every time this is called. We only want construction of the error message in the cold path lambda.

stan/math/prim/err/check_nonzero_size.hpp
```diff
@@ -21,11 +21,13 @@ namespace math {
 template <typename T_y>
 inline void check_nonzero_size(const char* function, const char* name,
                                const T_y& y) {
-  if (y.size() > 0) {
-    return;
+  STAN_NO_RANGE_AND_SIZE_CHECK
+  if (y.size() == 0) {
+    [&]() STAN_COLD_PATH {
+      invalid_argument(function, name, 0, "has size ",
+                       ", but must have a non-zero size");
```
t4c1, line 28: I am not sure if wrapping a single function call into a lambda affects performance. Hmm, maybe it does, as the arguments have to be passed into `invalid_argument`. Anyway You could also add `STAN_COLD_PATH` attribute to `invalid_argument`.

stan/math/prim/meta/macros.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef STAN_MATH_PRIM_META_MACROS_HPP
+#define STAN_MATH_PRIM_META_MACROS_HPP
```
t4c1, line 2: Macros is a bit too generic name for this file. How about `compiler_attributes` or `compiler_specific`?

stan/math/prim/meta/macros.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#ifdef __GNUC__
+#define likely(x) __builtin_expect(!!(x), 1)
+#define unlikely(x) __builtin_expect(!!(x), 0)
+#define STAN_ALWAYS_INLINE __attribute__((always_inline))
+#define STAN_STRONG_INLINE inline STAN_ALWAYS_INLINE
+#define STAN_FLATTEN __attribute__((flatten))
+#define STAN_STRONGER_INLINE inline __attribute__((always_inline, flatten))
+#define STAN_COLD_PATH __attribute__((noinline, cold))
+#define STAN_HOT_PATH __attribute__((hot))
```
t4c1, line 12: Are any of these, except `STAN_COLD_PATH` used anywhere? I think each of these should only be introduced if profiling shows it significantly affects performance.

t4c1 (changes requested):
On the other hand I like `STAN_NO_RANGE_AND_SIZE_CHECK`. I believe we already have some functions in which significant amount of time is spent in checks. Also this is not much work to maintain. It is just an option for early return, without any special cases. Although we may be able to find a better name for this macro. Maybe `STAN_RETURN_IF_NO_CHECKS`?

wds15:
What about STAN_NDEBUG ?

SteveBronder:
> What about STAN_NDEBUG ?

Oh wait, why are we reinventing the wheel here why don't we just use NDEBUG??

wds15:
NDEBUG is fine for me as well...

stan/math/prim/err/check_matching_dims.hpp
SteveBronder: For this the lambda ends in `invalid_argument()` so if the lambda is called this will throw. Isn't the check itself here

```cpp
  std::vector<int> y1_d = dims(y1);
  std::vector<int> y2_d = dims(y2);
// yada yada...
  if (y1_d.size() != y2_d.size()) {
    error_throw();
  } else {
    for (int i = 0; i < y1_d.size(); i++) {
      if (y1_d[i] != y2_d[i]) {
        error_throw();
      }
    }
  }
```

stan/math/prim/err/check_nonzero_size.hpp
SteveBronder: The invalid argument code is just 

```
  std::ostringstream message;
  message << function << ": " << name << " " << msg1 << y << msg2;
  throw std::invalid_argument(message.str());
```

But idt we want the computer to even worry about having to construct that message if the check is going to pass

stan/math/prim/meta/macros.hpp
SteveBronder: I can remove these, though I was interested in testing out what would happen if we used `STAN_STRONGER_INLINE` onto the models `log_prob` since that would tell the compiler to extremely attempt to inline log_prob and everything in log_prob. I also think having them here lets us experiment more easily with where they should go. Like for instance putting `STAN_STRONGER_INLINE` into the callback code so we can tell the compiler we really want the lambda that's called in `chain()` to be inlined into chain

stan/math/prim/meta/macros.hpp
SteveBronder: I like `compiler_attributes`

bbbales2:
> On the other hand I like STAN_NO_RANGE_AND_SIZE_CHECK

It just seems more complicated. Like how is this exposed to the surface? The cold path stuff is for free to the user.

I think this would definitely make sense for the O1/O2/O3 discussion. Like, turn off all nan/inf (the expensive ones) at O3 or something -- but that's just a different thing.

SteveBronder:
It's exposed to the surface by the user defining `-DNDEBUG` so that at compile time it knows these checks are no ops. It's commonly used in things like the standard library to turn off debug information. 

https://en.m.wikibooks.org/wiki/C_Programming/assert.h

bbbales2:
Without a need for this outside of benchmarking Math, or without a plan to expose them at the interfaces, they're really not doing anything and I don't see why we'd add them. We can always add them once we've got a good reason. The idea doesn't go away if we don't put them in now.

I wouldn't call this debug information. It is the intended behavior of the Math library to do these checks. The COLD_PATH stuff is different if I'm understanding correctly. Big performance difference and anyone using gcc gets it for free.

wds15:
While I do like the idea to be able to kill those checks for the sake of speed, I do concur here that this is actually a separate thing to sort out. These checks have trigger many debates - heated ones - and it‘s probably better to do that in a dedicated PR.

(I am in favor for having the knob to turn this off, but still let‘s do one by one)

rok-cesnovar:
I agree this should be a separate thing and should not be `NDEBUG`. These checks are not really debug information.

SteveBronder:
> Without a need for this outside of benchmarking Math, or without a plan to expose them at the interfaces, 

I'm not sure I'm following, what do you mean when you say we don't have a plan to expose these at the user interface? Or maybe a better question is, what does that look like to you? In my mindplace the user can just add `-DNDEBUG` to their `Makevars` in R and cmdstan users would put that in `make/local`

> they're really not doing anything and I don't see why we'd add them. We can always add them once we've got a good reason. The idea doesn't go away if we don't put them in now.

It does do something, which is that it makes the functions do nothing! Since this makes these functions no-ops the compiler completely ignores them

> I wouldn't call this debug information. It is the intended behavior of the Math library to do these checks. 

Yes and the std library treats it the same way where the intended behavior of `assert()` is to stop the program when something goes wrong. Then if you have `NDEBUG` on it changes the behavior to not stop the program. This only effects things if the user ops into it (aka passing `-DNDEBUG` at compile time)

> The COLD_PATH stuff is different if I'm understanding correctly. Big performance difference and anyone using gcc gets it for free.

There's two things happening here.

1. COLD_PATH puts the error code on a code path that the CPU will never try to branch predict / preload anything related to the error path. `__GNUC__` is actually defined for both clang and gcc so both compiler users will get this benefit.

2. When the user passes `-DNDEBUG` to the compiler these ops are completely turned off so that the compiler sees they are no-ops and doesn't bother even adding them to the binary. 

`NDEBUG` is only meant to be used for "release" versions of a program. So the user would only turn this on once they've run this program once and made sure all their sizes /ranges are correct.

bbbales2:
> So the user would only turn this on once they've run this program once and made sure all their sizes /ranges are correct.

Yeah, the use case makes sense.

Dropping checks has come up before (I've heard @seantalts talk about this, and @bgoodri may have as well, though he can correct me). We could probably optionally turn off nan/infs/positive definite and bounds checks separately.

It's a popular idea just needs talked to the interfaces, because that's where it will get used.

> I'm not sure I'm following, what do you mean when you say we don't have a plan to expose these at the user interface?

The interfaces mean how will people use it from cmdstan, cmdstanr, cmdstanpy, rstan, or pystan. Things like the threading on/off flags haven't been a walk in the park with how make and the precompiled headers work (though I think Rok has some sort of solution there now) even though there is a clear use case. rstanarm and brms and whatnot could probably use it too.

SteveBronder:
> Dropping checks has come up before (I've heard @seantalts talk about this, and @bgoodri may have as well, though he can correct me). We could probably optionally turn off nan/infs/positive definite and bounds checks separately.

imo I'd rather not turn these off since the turning off non/inf/ number bounds checks is usually associated with things like `-Ofast` and imo is a bit more of a problem. We could potentially turn those off but it's a much riskier thing than turning of the size and range checks.

> The interfaces mean how will people use it from cmdstan, cmdstanr, cmdstanpy, rstan, or pystan. Things like the threading on/off flags haven't been a walk in the park with how make and the precompiled headers work (though I think Rok has some sort of solution there now) even though there is a clear use case. rstanarm and brms and whatnot could probably use it too.

Why wouldn't the user just add `-DNDEBUG` to their `Makevars` in R and cmdstan users would put that in `make/local`? @rok-cesnovar would this mess the the pre-complied headers? My only guess is that the precompiled header would not have `NDEBUG` defined so worst case this wouldn't do anything

rok-cesnovar:
> Why wouldn't the user just add -DNDEBUG to their Makevars in R and cmdstan users would put that in make/local?

One fear I have for this is that some more advanced users may already have this flag and expect to still get the "normal" Stan behavior. Though that can be avoided by prominently displaying this in the release notes.

I am a bit less worried if this only concerns indexing bounds checks and not element-wise checks for nan/inf. For those I would be worried.

> would this mess the the precomplied headers?

I dont think so, worst case the user needs to rebuild cmdstan after setting those. That is advised in general to anyone that touches CXXFLAGS directly. We have "quality-of-life" improvements when using `STAN_THREADS`, `STAN_OPENCL` and `STAN_MPI` (no need to rebuild manually) because that is what we expect users to use and that is what we promote or will promote. Anyone setting arbitrary C++ flags that we dont document or publicize/promote is "on their own". We have zero to little knowledge of how Stan behaves with various g++/clang++ flags and we also probably dont have the resources to try/test most of them.

For rstan/pystan I have no idea how/if they handle prebuilding main.o and CRTP and all that. I think we should not cause more issue for those tightly-coupled interfaces by suggesting users to try NDEBUG there. If someone is this concerned with speed they should be using cmdstan in the first place, imo.

stan/math/prim/err/check_matching_dims.hpp
t4c1: I guess I was not clear. I consider the check to be the branch between the error path and the ok path. Regardless we only want to put cold in the cold path if it is only executed if the check is failing. This code you put in the cold path here needs to be executed every time this `check_matching_dims` is called.

stan/math/prim/meta/macros.hpp
t4c1: >  testing out what would happen if we used STAN_STRONGER_INLINE onto the models

That needs to be done before the code is put in a PR. Opening a PR should be conditional on demonstrating a speedup.

t4c1:
I still think adding an option to turn off checks is a good idea, but I think it should not use `NDEBUG` macro. Also it might be even better to enable finer control by having two macros - one for bounds checks and one for value checks. However, turning off checks is a separate topic and should probably be in a separate PR.

SteveBronder:
I removed the macro to remove checks right now, but we should have something like that in the future

SteveBronder:
Is this good to go?

t4c1:
No. You have not addressed everything from my last review.

SteveBronder:
Whups sorry missed that, removed the extra compiler attributes

t4c1 (approved):

PR #2254, opened 2020-12-11
Initial commit to make matrix constrains varmat friendly (Issue #2101)

bbbales2:
## Summary

This is an initial commit to making matrix constrains work with `varmat`

Here's a checklist of functions:

- [x] log #2254
- [x] inv_logit #2254
- [x] multiply_log #2254
- [x] exp #2254
- [x] tanh #2254
- [x] log1m #2254
- [x] square #2254
- [x] cholesky_corr_constrain #2254
- [x] cholesky_factor_constrain #2254
- [x] corr_matrix_constrain #2254
- [x] cov_matrix_constrain_lkj #2254
- [x] cov_matrix_constrain #2254
- [x] read_cov_matrix #2254
- [x] read_corr_matrix #2254
- [x] read_corr_L #2254
- [x] read_cov_L #2254

## Release notes

- Updating matrix constrains to work with `varmart`

## Checklist

- [x] Math issue #2101

- [x] Copyright holder: Columbia University

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #2256, opened 2020-12-12
var matrix specializations for a-c unary functions in rev

SteveBronder:
## Summary

This is just going through rev and adding var matrix functions for a bunch of functions.

## Tests

`expect_ad_matvar` tests added for all of the functions in here. Though I still need to write tests for `std::vector<var<matrix>>`

## Side Effects

I wrote this by specializing a function for each rev version. But that means we have to write the `scalar_unary` version like

```cpp
template <typename Container,
          require_not_container_st<std::is_arithmetic, Container>* = nullptr,
          require_not_var_matrix_t<Container>* = nullptr>
inline auto fabs(const Container& x) {
  return apply_scalar_unary<fabs_fun, Container>::apply(x);
}
```

Though there may be some other way to do this where instead we specialize `*_fun`

```cpp
struct fabs_fun {
  template <typename T>
  static inline T fun(const T& x) {
    using std::fabs;
    return fabs(x);
  }
};
```

As it's written right now it's nice because it works with `std::vector<var<matrix>>`. eod we need to specialize somewhere so I don't mind having it like we do now. We could also make `var<matrix>` not a container type so then we can remove the `require_not_var_matrix_t<Container>`

## Release notes

Adds arc trig functions for `var<matrix>` along several other unary operators

## Checklist

- [x] Math issue #1805 


bbbales2:
THere are a few functions like this in #2254. I was stuck there on what to do with `std::vector<varmat>` but sounds like you solved that.

Does what you did here also work with `std::vector<std::vector<varmat>>`, etc?

I asked @andrjohns about this in the other thread but maybe you have already solved it (in which case don't worry about it @andrjohns !)

> Though I still need to write tests for std::vector<var<matrix>>

That should be testable now. I had to add these tests for `log_softmax`. I didn't add tests for `std::vector<std::vector<varmat>>` though.

SteveBronder:
Lemme test that. It should at least be close to working. All this does is overload each function so when apply scalar iterates over the elements it should just do the normal thing and differ to the other specializations until it hits varmat and then just do the varmat thing

stan/math/prim/fun/acos.hpp
```diff
@@ -42,7 +42,8 @@ struct acos_fun {
  * @return Arc cosine of each variable in the container, in radians.
  */
 template <typename Container,
-          require_not_container_st<std::is_arithmetic, Container>* = nullptr>
+          require_not_container_st<std::is_arithmetic, Container>* = nullptr,
+          require_not_var_matrix_t<Container>* = nullptr>
 inline auto acos(const Container& x) {
   return apply_scalar_unary<acos_fun, Container>::apply(x);
 }
```
SteveBronder, line 49: @andrjohns I'm having a hard time removing this `require_not_var_matrix_t<Container>` and I'd like to not have to put these everywhere. Right now this function works for any stan scalar or container that does not have an arithmetic scalar type. So it's kind of a mixed bag of types this takes in. Though specializations for each seems like overkill as well so I'm not really sure if we can make this look nicer

SteveBronder:
@bbbales2 fyi I removed var matrix from the container types. It feels like a var matrix is a single unit and we are pretty much always going to have specializations for var matrix. With the container checks we're normally checking whether it's an eigen vector so we can do some specialization or a std vector so we can recursivly go through it (or to map the std vector to an eigen matrix to use the Eigen specialization). So like I think with var matrix it makes sense to not have it count as a container. Does that make sense?

rok-cesnovar:
But if its not a container, is it a scalar? I dont think that make sense.

bbbales2:
Yeah I'm not totally sure what it is either, but I'm also not using it like the other things we call containers. Popping it out seems fine with me.

Seems like whenever we use `var_value`, it is `require_var_vector_t` or `require_var_matrix_t` or `require_rev_matrix_t` or something like that.

I guess we could change `require_container_t` to `require_std_vector_or_eigen_t` lol but that's getting kinda long.

bbbales2:
@SteveBronder this ready to review? I want to just copy whatever you do here for the scalar functions in #2254 so if it's done that's convenient lol.

SteveBronder:
@bbbales2 this should be good to go

stan/math/rev/fun/sin.hpp
```diff
@@ -51,7 +51,28 @@ class sin_vari : public op_v_vari {
  * @param a Variable for radians of angle.
  * @return Sine of variable.
  */
-inline var sin(const var& a) { return var(new internal::sin_vari(a.vi_)); }
+inline var sin(const var& a) {
```
bbbales2, line 54: sin_vari above used anymore?

stan/math/rev/fun/sinh.hpp
```diff
@@ -49,7 +49,28 @@ class sinh_vari : public op_v_vari {
  * @param a Variable.
  * @return Hyperbolic sine of variable.
  */
-inline var sinh(const var& a) { return var(new internal::sinh_vari(a.vi_)); }
+inline var sinh(const var& a) {
```
bbbales2, line 52: sinh_vari used anymore?

stan/math/rev/fun/tan.hpp
```diff
@@ -49,7 +49,28 @@ class tan_vari : public op_v_vari {
  * @param a Variable for radians of angle.
  * @return Tangent of variable.
  */
-inline var tan(const var& a) { return var(new internal::tan_vari(a.vi_)); }
+inline var tan(const var& a) {
```
bbbales2, line 52: tan_vari used anymore?

stan/math/rev/fun/tanh.hpp
```diff
@@ -49,7 +49,28 @@ class tanh_vari : public op_v_vari {
  * @param a Variable.
  * @return Hyperbolic tangent of variable.
  */
-inline var tanh(const var& a) { return var(new internal::tanh_vari(a.vi_)); }
+inline var tanh(const var& a) {
```
bbbales2, line 52: tanh vari used anymore?

stan/math/rev/functor/apply_scalar_unary.hpp
```diff
@@ -31,6 +31,22 @@ struct apply_scalar_unary<F, var> {
   static inline return_t apply(const var& x) { return F::fun(x); }
 };
 
+template <typename F, typename T>
+struct apply_scalar_unary<F, T, require_var_matrix_t<T>> {
+  /**
+   * Function return type, which is <code>var</code>.
```
bbbales2, line 37: Not quite var

test/unit/math/mix/fun/acos_test.cpp
```diff
@@ -21,3 +21,20 @@ TEST(mathMixMatFun, acos) {
...
+  using stan::test::expect_ad_matvar;
+  auto f = [](const auto& x1) {
+    using stan::math::acos;
+    return acos(x1);
+  };
+  Eigen::MatrixXd A(2, 4);
+  A << -2.2, -0.8, 0.5, 1 + std::numeric_limits<double>::epsilon(), 1.5, 3, 3.4,
+      4;
+  expect_ad_matvar(f, A);
+  std::vector<Eigen::MatrixXd> A_vec;
```
bbbales2, line 35: The easier way to build these is with initializer lists. Something like:

```
std::vector<Eigen::MatrixXd> A_vec = { A, A, A };
```

test/unit/math/mix/fun/atan2_test.cpp
```diff
@@ -16,3 +16,18 @@ TEST(mathMixCore, atan2) {
   stan::test::expect_ad(f, 1.2, 3.9);
   stan::test::expect_ad(f, 0.5, 2.3);
 }
+/* Not sure how to test this so I'll come back to it.
```
bbbales2, line 19: Now iz the time

test/unit/math/mix/fun/asinh_test.cpp
```diff
@@ -17,3 +17,28 @@ TEST(mathMixFun, asinh) {
     }
   }
 }
+
+TEST(mathMixMatFun, asinh_varmat) {
+  auto f = [](const auto& x1) {
+    using stan::math::asinh;
+    return asinh(x1);
+  };
+  auto com_args = stan::test::internal::common_args();
```
bbbales2, line 26: I think all these tests should be for varmat with vector, row vector and Matrix vector arguments.

How about:

```cpp
expect_ad_common_vector_matvar(f);
```

And it does:
```cpp
expect_ad_vector_matvar(f, stan::math::to_vector(stan::test::internal::common_args()));
  // not sure if `common_args` is std::vector or not
```

And that function is something like:

```cpp
expect_ad_vector_matvar(F& f, const auto& x) {
  Eigen::VectorXd v = x;
  Eigen::RowVectorXd r = x.transpose();
  Eigen::MatrixXd m(x.size(), 2);
  com_args_m.block(0, 0, com_args_v.size(), 1) = x;
  com_args_m.block(0, 1, com_args_v.size(), 1) = x.reverse(); // reverse just to mix stuff up
  
  expect_ad_matvar(f, v);
  expect_ad_matvar(f, r);
  expect_ad_matvar(f, m);

  std::vector<Eigen::VectorXd> vv = { v, v };
  std::vector<Eigen::RowVectorXd> rv = { r, r };
  std::vector<Eigen::MatrixXd> mv = { m, m };

  expect_ad_matvar(f, vv);
  expect_ad_matvar(f, rv);
  expect_ad_matvar(f, mv);
}
```

Also does `std::vector<std::vector<...>>` work? Don't think it needs tested I'm just curious.

bbbales2 (changes requested):
Review

test/unit/math/mix/fun/asinh_test.cpp
SteveBronder: Yeah I think it makes sense to test matrix vector row vector at least to make sure they compile. I'll write something like this up

stan/math/prim/fun/acosh.hpp
```diff
@@ -83,8 +83,9 @@ struct acosh_fun {
  * @param x container
  * @return Elementwise acosh of members of container.
  */
-template <typename T, require_all_not_nonscalar_prim_or_rev_kernel_expression_t<
-                          T>* = nullptr>
+template <
```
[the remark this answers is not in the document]
SteveBronder: Actually this can just be `template <typename T, require_container_t<T>* = nullptr>`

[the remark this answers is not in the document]
bbbales2: There were some functions here that didn't have `apply_vector_unary`/Container implementations. I'm assuming it's because the Eigen array functions didn't exist for them to be written (`cbrt`, this, and others). I don't think cleanup is what we're going for in this pull really, so it's fine to leave it.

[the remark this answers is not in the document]
bbbales2: (But if you want to clean things, feel free, happy to review)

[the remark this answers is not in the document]
SteveBronder: Ups actually ignore that, this needs to be able to take in scalar types so that's why we can't use `require_constainer_st`

stan/math/prim/functor/apply_vector_unary.hpp
```diff
@@ -159,6 +159,12 @@ struct apply_vector_unary<T, require_std_vector_vt<is_stan_scalar, T>> {
   }
 };
 
+namespace internal {
+template <typename T>
+using is_container_or_var_matrix
+    = disjunction<is_container<T>, is_var_matrix<T>>;
+}
```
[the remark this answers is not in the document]
SteveBronder: This is the only place I think we need this right now. I'm going to make a sweep next week and see if we can clean up some of the `requires` and I can add these kind of things then. Plus this is only used here for now so once we need it in more places we can make it a normal require

stan/math/rev/fun/cbrt.hpp
```diff
@@ -42,7 +34,26 @@ class cbrt_vari : public op_v_vari {
...
+ * @tparam Varmat a `var_value` with inner Eigen type
+ * @param a Specified variable.
+ * @return Cube root of the variable.
+ */
+template <typename VarMat, require_var_matrix_t<VarMat>* = nullptr>
+inline auto cbrt(const VarMat& a) {
+  return make_callback_var(
+      a.val().unaryExpr([](const auto x) { return cbrt(x); }),
+      [a](const auto& vi) mutable {
+        a.adj().array() += vi.adj_.array() / (3.0 * vi.val_.array().square());
```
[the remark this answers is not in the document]
SteveBronder: `.array()` does not have a `.noalias()`

stan/math/rev/fun/sin.hpp
SteveBronder: Nope!

test/unit/math/mix/fun/atan2_test.cpp
SteveBronder: This is only for unary stuff so another time

stan/math/rev/fun/cbrt.hpp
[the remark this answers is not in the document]
t4c1: Also noalias will only speed things up if the outermost operation on the right hand side is a matrix product.

bbbales2 (commented):
I changed the fabs/abs tests to work like the rest of them, and also a couple things with docs.

Let me know if this works for you and I'll merge when we get greenlight.

SteveBronder:
Ty! Yes I just went through and removed the `.noalias()` call where we didn't need them so should be cool and good

bbbales2 (approved):

Issue #461, opened 2016-11-30
Add score() function


syclik:
#### Summary:
This is from @bgoodri on stan-dev/stan#605. 

Add a `score()` function to the language. It would require nested autodiff.

#### Description:
Add the score function.

#### Additional Information:
I'm not sure exactly how this would work. @bgoodri, might need some help.

#### Current Version:
v2.13.0

bgoodri:
A score() function would be good --- to calculate the derivative of the log-likelihood, usually on a per observation basis --- but it would also be good to have a gradient() function to calculate the derivative of an expression to feed into the log-likelihood. But I'm not sure how high a priority this would be relative to other things because in 2.13.0 you can get to the gradient(), jacobian(), etc. functions in Stan Math by writing your own C++.

bob-carpenter:
You have to be careful to use nested autodiff properly.  The gradient() function is set up to do that, but you have to work in a purely nested fashion as we do in the Jacobian calcs for the ODE solver.

bob-carpenter:
Also, I don't know how we could return the value of the derivative of the log density until it's been completely defined.  Then we can't use it for anything in the density, or the density will change. Where do you need this function and for what?

jvaara:
I have been watching this function request and hoping to use it for Design of Experiment via Information matrix.

bob-carpenter:
We don't even have a technical spec for how this could work, so
it's not going to be done soon unless someone figures out how to
do it.

PR #2260, opened 2020-12-15
Add automatic benchmarking utility

t4c1:
## Summary

Adds a python script that can generate and run benchmarks for most of function signatures in math. It can:
- benchmark prim, rev, fwd overloads
- plot results
- compare OpenCL and CPU implementations (including or excluding data transfers)
- compare different implementations of same function

What is not supported (yet?): higher order functions, mixed mode autodiff.

Also for now all arguments' dimensions are the same. In other words all matrices are square and have same number of rows/cols as any vector/array arguments.

## Tests
Not tested.

## Side Effects
Expression tests were refactored a bit so they share some code with benchmarking script.

## Release notes
Adds a python script that can generate and run benchmarks for most of function signatures in math.

## Checklist


rok-cesnovar:
Very nice! 

@ahartikainen would you mind reviewing the python code?

This is also going to close https://github.com/stan-dev/math/issues/2083, though you will have to build Benchmark manually before building benchmarks.

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+                code += "    stan::test::recursive_sum(res).grad();\n"
+            result += benchmark_template.format(benchmark_name=benchmark_name, setup=setup,
+                                                var_conversions=var_conversions, code=code, multi=multiplier,
+                                                max_size=max_size)
+    cpp_filepath = working_folder + cpp_filename
+    with open(cpp_filepath, "w") as o:
+        o.write("#include <benchmark/benchmark.h>\n")
+        o.write("#include <test/expressions/expression_test_helpers.hpp>\n\n")
+        o.write(result)
+        o.write("BENCHMARK_MAIN();")
```
ahartikainen, line 327:
```suggestion
    with open(cpp_filepath, "w") as f:
        f.write("#include <benchmark/benchmark.h>\n")
        f.write("#include <test/expressions/expression_test_helpers.hpp>\n\n")
        f.write(result)
        f.write("BENCHMARK_MAIN();")
```

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    exe_filepath = cpp_filepath.replace(".cpp", exe_extension)
+    build(exe_filepath)
+    run_benchmark(exe_filepath, n_repeats, csv_out_file)
+
+
+def main(functions_or_sigs, cpp_filename="benchmark.cpp", overloads=("Prim", "Rev"), multiplier_param=None,
```
ahartikainen, line 333:
```suggestion
def main(functions_or_sigs, cpp_filename="benchmark.cpp", overloads=None, multiplier_param=None,
```

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    :param max_dim: Maximum number of argument dimensions to benchmark. Signatures with any argument with
+             larger number of dimensions are skipped."
+    :param n_repeats: Number of times to repeat each benchmark.
+    :param skip_similar_signatures: Whether to skip similar signatures. Two signatures are similar if they
+             difffer only in similar vector types, which are vector, row_vector and real[].
+    :param csv_out_file: Filename of the csv file to store benchmark results in.
+    :param plot: Filename of bmp or csv fle to store plot into. If  filename is empty, opens a window with graph.
+    :param plot_log_y: Use logarithmic y axis for plotting
+    :param plot_cl_speedup: plot speedup of OpenCL overloads compared to CPU ones
+    """
```
ahartikainen, line 353:
```suggestion
    """
    if overloads is None:
        overloads = ("Prim", "Rev")
```

test/sig_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    "matrix": "Eigen::Matrix<SCALAR, Eigen::Dynamic, Eigen::Dynamic>",
+    "rng": "std::minstd_rand",
+    "ostream_ptr": "std::ostream*",
+}
+
+def get_cpp_type(stan_type):
+    n_vec = 0
+    if stan_type.endswith("]"):
+        stan_type, vec = stan_type.split("[")
+        n_vec = len(vec)
```
ahartikainen, line 33: Does this work?

test/sig_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    p = subprocess.Popen(
+        (stanc3 + " --dump-stan-math-signatures"),
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+        shell=True,
+    )
+    res = parse_signature_file(p.stdout)
+    if p.wait() != 0:
+        sys.stderr.write("Error in getting signatures from stanc3!\n")
+        sys.exit(-1)
```
ahartikainen, line 224: subprocess.run or something similar

test/sig_utils.py
```diff
@@ -0,0 +1,270 @@
...
+    Handles list of functions, splitting elements between functions and signatures.
+    :param functions_input: This can contain names of functions
+    already supported by stanc3, full function signatures or file names of files containing
+    any of the previous two.
+    :return:
+    """
+    function_names = []
+    function_signatures = []
+    for f in functions_input:
+        if "." in f or "/" in f or "\\" in f:
```
ahartikainen, line 253:
```suggestion
        if ("." in f) or ("/" in f) or ("\\" in f):
```

test/sig_utils.py
```diff
@@ -0,0 +1,270 @@
...
+            functions_input.extend(parse_signature_file(open(f)))
```
ahartikainen, line 254:
```suggestion
            with open(f) as fh:
                functions_input.extend(parse_signature_file(fh)
```

ahartikainen (commented):
Hi, I added couple of comments / suggestions.

Some of the `or` lines need update.

There is one example how to use `fig, ax` for plotting.

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    if p1.wait() != 0:
+        raise RuntimeError("command failed: " + command)
+
+
+def build(exe_filepath):
+    """
+    Builds a file using make.
+    :param exe_filepath: File to build
+    """
+    command = make + " " + exe_filepath
```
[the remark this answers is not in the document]
ahartikainen: ok, it was imported from the other file

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
+#!/usr/bin/python
+from __future__ import print_function
+from argparse import ArgumentParser, RawTextHelpFormatter
+import sys
+import subprocess
+import os
+
+sys.path.append("test")
```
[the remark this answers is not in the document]
ahartikainen: Ok, this just assumes that the script is run in stan-math folder, which is fine.

SteveBronder:
Should `benchmark.py` just  sit in the top level folder like `runBenchmarks.py`?

SteveBronder:
Really neat! Couple little things

1. 
```
$ python ./benchmarks/benchmark.py
usage: benchmark.py [-h] [--overloads OVERLOADS [OVERLOADS ...]]
                    [--multiplier MULTIPLIER] [--max_size MAX_SIZE]
                    [--max_dim MAX_DIM] [--cpp filename] [--repeats N]
                    [--csv filename] [--plot filename] [--plot_log_y]
                    [--opencl setting] [--plot_cl_speedup]
                    [--plot_reference filename] [--skip_similar_signatures]
                    functions [functions ...]
benchmark.py: error: too few arguments
```

Can we just have it print the help output if not enough arguments are selected?

2. For the benchmark name that gets printed
```
multiply_Primrow_vector_Revvector/16777216/manual_time
```

Could we make that nicer like

```
multiply_double_row_vector_var_vector/16777216/manual_time
```

3. For the benchmarks using our AD stack we should fill up the ad stack with data before calling the first benchmark. Otherwise the first benchmark called will be arbitrarily slower than the others. I use something like [this](https://gist.github.com/SteveBronder/f14df4b040b5d6d3e4afb03fecbf8888) passing it a little over the max allocation I expect like 

```cpp
constexpr int end_mem = 100 * 1.2;
BENCHMARK_TEMPLATE(toss_me, end_mem);
```

4. after we call `.grad()` I think it makes sense to call `benchmark::ClobberMemory();`

5. Each benchmark should probably be `static`

6. Because of the number of functions in each benchmark file you may run into [benchmark issue #461](https://github.com/google/benchmark/issues/461) where you end up with some function alignment issues that can impact performance of different tests by a good bit. It's talked about in [this video](https://youtu.be/10MQW-aJU3g?t=197) (timestamped at where he starts talking about it). The only way around this I'm aware of is careful compiler flags or breaking up the benchmarks into individual files. Would it be difficult to build multiple .cpp files and collate their results after? Maybe we could just test one function to see if this is an issue or not

7. I think `(16000000)**(1/dimm)` with a multiplier of 4 feels wayyy to big and not granular enough. Most stan programs are not hitting data or var sizes of 16,000,000. I'd rather have something like a range of `(1:32,768)` and a granularity of (2) everywhere


8. For all scalar benchmarks can the multipler and range be removed?

```
BENCHMARK(multiply_Primreal_Revreal)->UseManualTime();
```

9. Does this just make every value in the matrix 0.4?

```
Eigen::Matrix<double, 1, Eigen::Dynamic> arg0 = stan::test::make_arg<Eigen::Matrix<double, 1, Eigen::Dynamic>>(0.4, state.range(0));
```

Why don't we just set the seed to 0 and then just use random?

SteveBronder:
Also would this be able to use varmat if we had a list of all varmat supported functions somewhere? Or is it because the compiler doesn't support varmat yet?

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+from sig_utils import *
+import itertools
+
+working_folder = "./benchmarks/"
```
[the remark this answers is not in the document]
rok-cesnovar: The problem with that is that the makefiles would not work then. The makefiles are done in such a way that Google Benchmark is linked in for anythin in benchmarks/*.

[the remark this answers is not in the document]
ahartikainen:
```suggestion
WORKING_FOLDER = "./benchmarks/"
```

Ok, then this is ok

benchmarks/benchmark.py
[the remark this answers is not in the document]
t4c1: Yeah I put the file that contains code that is shared between this and expression tests in the test folder.

t4c1:
> Can we just have it print the help output if not enough arguments are selected?

Fixed.

> For the benchmark name that gets printed
> Could we make that nicer like

I would prefer not to. 5-argument functions already have very long names. But I can add the additional underscore.

> For the benchmarks using our AD stack we should fill up the ad stack with data before calling the first benchmark. Otherwise the first benchmark called will be arbitrarily slower than the others.

Not the first benchmark, just the first run of the first benchmark. Also I have never in practice observed such slowdowns due to the allocation. However adding something like that does not cost much, so sure, I'll add allocation. Although I don't see any need for doing any calculations on such memory before the benchmarks like you did. 

Calculating the amount of memory needed is hard in general. I added an estimate that should work for many functions. However ones working on nested arrays exceed that quite a lot.

> after we call .grad() I think it makes sense to call benchmark::ClobberMemory();

added

>Each benchmark should probably be static

I don't think it matters, but ok.

> Would it be difficult to build multiple .cpp files and collate their results after?

It would be some extra work. Also this does not solve the issue, just makes it repeatable. If there was a large difference due to alignment I think it might be better to have possibility to detect it.

> I'd rather have something like a range of (1:32,768) and a granularity of (2) everywhere

That is why you can set those parameters.

> For all scalar benchmarks can the multipler and range be removed?

The range is se to (1,1), which is equivalent.

> Does this just make every value in the matrix 0.4?

Yes (except for functions requiring specific values). I know that is bad for functions that branch depending on values, but there is no good solution except setting range in which to generate values manually for every function (or even distribution to generate data from).

benchmarks/benchmark.py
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+
+benchmark_template = """
```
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+def run_command(command):
+    """
+    Runs given command and waits until it finishes executing.
+    :param command: command to execute
+    """
+    print()
+    print(command)
+    p1 = subprocess.Popen(command, shell=os.name != "nt")
+    if p1.wait() != 0:
+        raise RuntimeError("command failed: " + command)
```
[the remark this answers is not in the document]
t4c1: > Why Windows need shell=True?

Windows does not, Linux did to parse spaces from command. Now I changed the command to be a list of arguments, so it is not needed anymore.

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    run_command(command)
+
+
+def plot_results(csv_filename, out_file="", plot_log_y=False):
+    """
+    Plots benchmark results.
+    :param csv_filename: path to csv file containing results to plot
+    :param out_file: path to image file to store figure into. If it equals to "window"opens it in an interactive window.
+    """
+    import pandas, numpy, matplotlib
```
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    if out_file != "window":
+        matplotlib.use('Agg')
+    from matplotlib import pyplot as plt
```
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    from matplotlib import colors
+    csv_file = open(csv_filename)
+    data = csv_file.read()
+    start = data.find("name,iterations")
+    csv_file = open(csv_filename)
+    # google benchmark writes some non-csv data at beginning
+    csv_file.read(start)
+    data = pandas.read_csv(csv_file)
```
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    signatures = set(signatures)
+    remaining_functions = set(functions)
+    parsed_signatures = []
+    ref_signatures = set()
+    for signature in all_signatures:
+        return_type, function_name, stan_args = parse_signature(signature)
+        reference_args = tuple(reference_vector_argument(i) for i in stan_args)
+        if (function_name, reference_args) in ref_signatures:
+            continue
+        if signature in signatures or function_name in functions:
```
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    for signature in signatures:
+        return_type, function_name, stan_args = parse_signature(signature)
+        reference_args = tuple(reference_vector_argument(i) for i in stan_args)
+        if (function_name, reference_args) in ref_signatures:
+            continue
+        ref_signatures.add((function_name, reference_args))
+        parsed_signatures.append([return_type, function_name, stan_args])
+        remaining_functions.discard(function_name)
+    if remaining_functions:
+        raise NameError("Functions not found: " + ", ".join(remaining_functions))
```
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    for return_type, function_name, stan_args in parsed_signatures:
+        dimm = 0
+        for arg in stan_args:
+            arg_dimm = 0
+            if "vector" in arg:
+                arg_dimm = 1
+            if "matrix" in arg:
+                arg_dimm = 2
+            if "[" in arg:
+                arg_dimm += len(arg.split("[")[1])
```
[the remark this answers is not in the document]
t4c1: Stan types like `int[]`, `row_vector[,,]`, `real[,,]`.

benchmarks/benchmark.py
t4c1: fixed

benchmarks/benchmark.py
t4c1: Why do you prefer it this way? So far I used this only for defaults I want to depend on other values.

test/sig_utils.py
t4c1: yes

test/sig_utils.py
t4c1: fixed

test/sig_utils.py
t4c1: fixed

test/sig_utils.py
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+
+        unique_sizes = sorted(numpy.unique(sig_sizes))
+        avg_sig_times = numpy.array([numpy.mean(sig_times[sig_sizes == i]) for i in unique_sizes])
+        plt.plot(unique_sizes, avg_sig_times, label=signature, color=tableau_colors[n])
+
+    plt.legend()
+    if out_file == "window":
+        plt.show()
+    else:
+        plt.savefig(out_file)
```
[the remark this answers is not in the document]
t4c1: I can't get this to work. Seems it messes up which size corresponds with which timing. Also do you have any idea for a larger list of colors to use for plots than TABLEAU_COLORS? I tried XKCD_COLORS and CSS_COLORS, but they both contain some that are virtually invisible on white background.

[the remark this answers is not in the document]
ahartikainen: I can tests this later today. I probably did some typo somewhere.

Probably sizes column is a text and not a number.

There are different ways to get colors.

How many do we need and how far from each other we want these to be?

https://matplotlib.org/tutorials/colors/colors.html#xkcd-v-x11-css4
https://colorcet.holoviz.org/

benchmarks/benchmark.py
ahartikainen: It is just a common way to use defaults (sure one can not change values for tuples).

benchmarks/benchmark.py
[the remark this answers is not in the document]
ahartikainen:
```suggestion
HERE = os.path.dirname(os.path.realpath(__file__))
TEST_FOLDER = os.path.abspath(HERE, "..", "test")
sys.path.append(TEST_FOLDER)
```

This is doesn't matter where user runs this script (it still matters where the script is located)

benchmarks/benchmark.py
```diff
@@ -0,0 +1,575 @@
...
+def run_command(command):
+    """
+    Runs given command and waits until it finishes executing.
+    :param command: command to execute
+    """
+    print()
+    print(" ".join(command))
+    p1 = subprocess.run(command)
+    if p1.returncode != 0:
+        raise RuntimeError("command failed: " + command)
```
ahartikainen, line 62:
```suggestion
        raise RuntimeError("command failed: " + " ".join(command))
```

ahartikainen:
Should script try run bench again if it fails?

I tried this

    python benchmarks/benchmark.py multinomial_log

(I directed stdout to a file)

```
make: 'test/expressions/stanc' is up to date.
    
make ./benchmarks/benchmark

./benchmarks/benchmark
2020-12-17 10:33:55
Running ./benchmarks/benchmark
Run on (12 X 2592 MHz CPU s)
Load Average: 0.52, 0.58, 0.59
terminate called after throwing an instance of 'std::domain_error'
  what():  multinomial_lpmf: Probabilities parameter is not a valid simplex. sum(Probabilities parameter) = 4, but should be 1
```

benchmarks/benchmark.py
ahartikainen, line 62: at top

    from datetime import datetime

```suggestion
    current_time_utc = datetime.utcnow().strftime("%Y%m%d-%H%M%S.%f")
    stdout_path = os.path.join(HERE, "bench_stdout-{}.txt".format(current_time_utc))
    stderr_path = os.path.join(HERE, "bench_stderr-{}.txt".format(current_time_utc))
    with open(stdout, "wb") as f_out, open(stderr, "wb") as f_err:
        p1 = subprocess.run(command, stdout=f_out, stderr=f_err)
        if p1.returncode != 0:
            raise RuntimeError("command failed: " + command)
```

benchmarks/benchmark.py
```diff
@@ -0,0 +1,575 @@
+#!/usr/bin/python
+from __future__ import print_function
+from argparse import ArgumentParser, RawTextHelpFormatter
```
ahartikainen, line 3:
```suggestion
from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
```

I think this is a bit better (it shows the defaults too)

benchmarks/benchmark.py
t4c1: I don't really want to redirect stdout and stderr. If user specifies appropriate arguments the results are saved anyway. At least I like to look at the output it produces to immediately see the results and to be able to guess how long it will take.

t4c1:
> Should script try run bench again if it fails?

Oh, I did not think of this. The arguments for functions requiring simplices (or some special matrices) are not yet supported. I will look into how to add that. Meanwhile you can test it on some other function.

benchmarks/benchmark.py
ahartikainen: Sound ok and there is always possibility to use `python ... > out.txt`

t4c1:
It seems expression tests are using python 2, so we can not use `subprocess.run()`:
```
Traceback (most recent call last):
  File "./runTests.py", line 375, in <module>
    main()
  File "./runTests.py", line 349, in main
    handleExpressionTests(tests, inputs.only_functions, num_expr_test_files)
  File "./runTests.py", line 297, in handleExpressionTests
    generateExpressionTests.main(only_functions, n_test_files)
  File "test/expressions/generateExpressionTests.py", line 96, in main
    signatures = get_signatures()
  File "test/sig_utils.py", line 210, in get_signatures
    p = subprocess.run((make, stanc3))
AttributeError: 'module' object has no attribute 'run'
```

ahartikainen:
Then I think `Popen` is the best option.

Is there a reason for py2? Old image?

t4c1:
No idea. @rok-cesnovar you set that up, right?

rok-cesnovar:
No reason apart from no one upgrading because we werent forced to :) 


its not just the expressions script, I think this goes back toRunTests.py.

I am not sure anyone knows if all our py scripts are py3 compliant. The main ones probably are else someone would complain.

These arent all images and virtual machines, some are physical machines so this isnt a 2-min task.  I would say we do py2 for this PR as well and make an issue for moving to python3 in CI.

benchmarks/benchmark.py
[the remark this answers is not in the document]
t4c1: I would say 30, or maybe 100 if we really want to be an the safe side. Some functions (distributions) have over thousand overloads, but looking at that many plots at the same time does not make much sense. And I think we want them to be as far from each other as we can make them.

benchmarks/benchmark.py
t4c1: fixed

benchmarks/benchmark.py
t4c1: fixed

benchmarks/benchmark.py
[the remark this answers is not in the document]
t4c1: fixed

benchmarks/benchmark.py
```diff
@@ -0,0 +1,495 @@
...
+    data = csv_file.read()
+    start = data.find("name,iterations")
+    csv_file = open(csv_filename)
+    # google benchmark writes some non-csv data at beginning
+    csv_file.read(start)
+    data = pandas.read_csv(csv_file)
+
+    signatures = numpy.array([i.split("/")[0] for i in data["name"]])
+    sizes = numpy.array([int(i.split("/")[1]) for i in data["name"]])
+    times = numpy.array(data["real_time"])
```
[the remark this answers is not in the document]
ahartikainen: I updated this part

benchmarks/benchmark.py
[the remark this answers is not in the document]
ahartikainen: I added 

```
[spine.set_visible(False) for loc, spine in ax.spines.items() if loc in ["top", "right", "left", "bottom"]]
ax.minorticks_off()
ax.grid()
```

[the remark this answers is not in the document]
ahartikainen: There are couple of ways to handle this

Either use a list of named colors in random order

`` 
colors = list(plt.cm.colors.cnames.keys())
random.shuffle(colors)
```

or create colors from colormap

    colors = plt.cm.viridis(numpy.linspace(0,1,timing_data["signatures"].unique().size))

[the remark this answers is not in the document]
Comment (ahartikainen) [reply benchmarks/benchmark.py]: The second plot adds a bit of noise on the figure. Is there a need for it or should it be more visible?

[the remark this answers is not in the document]
Comment (t4c1) [reply benchmarks/benchmark.py]: You mean the x-es? They show individual measurements, so as to give you an idea how consistent the benchmarking results are (those make more sense if you run multiple repeats)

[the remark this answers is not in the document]
Comment (t4c1) [reply benchmarks/benchmark.py]: I managed to make this suggestion work with some modifications.

File: benchmarks/benchmark.py [hunk 23]
```diff
@@ -92,6 +92,14 @@ def run_benchmark(exe_filepath, n_repeats=1, csv_out_file=None):
     run_command(command)
 
 
+def pick_color(n):
+    str_bit_reversed_n = '{:015b}'.format(n + 1)[::-1]
+    r = 0.9 * ((int(str_bit_reversed_n[0::3], 2) / 2. ** 5 + 0.3) % 1)
+    g = 0.9 * ((int(str_bit_reversed_n[1::3], 2) / 2. ** 5 + 0.3) % 1)
+    b = 0.9 * ((int(str_bit_reversed_n[2::3], 2) / 2. ** 5 + 0.3) % 1)
+    return r, g, b
```
ahartikainen, line 100: This is one way, but maybe sample from hsl space and then convert to rgb?

benchmarks/benchmark.py
```diff
@@ -0,0 +1,588 @@
...
+    fig.set_tight_layout(True)
+    ax.set_xscale("log")
+    if plot_log_y:
+        ax.set_yscale("log")
+
+    ax.set_xlabel("size")
+    ax.set_ylabel("time[us]")
+    for n, (signature, sub_data) in enumerate(timing_data.groupby("signatures")):
+        # conversions to numpy arrays allow us to hide duplicated labels
+        ax.plot(numpy.array(sub_data["sizes"]), numpy.array(sub_data["times"]), "x",
```
ahartikainen, line 138:
```suggestion
        ax.plot(sub_data["sizes"].values, sub_data["times"].values, "x",
```

Does this really create a duplicated labels with pandas?

benchmarks/benchmark.py
```diff
@@ -0,0 +1,588 @@
...
+    ax.set_xscale("log")
+    if plot_log_y:
+        ax.set_yscale("log")
+
+    ax.set_xlabel("size")
+    ax.set_ylabel("speedup")
+
+    for n, (signature, sub_data) in enumerate(timing_data.groupby("signatures")):
+        # conversions to numpy arrays allow us to hide duplicated labels
+        ax.plot(numpy.array(sub_data["sizes"]), numpy.array(sub_data["speedup"]), "x",
```
ahartikainen, line 205:
```suggestion
        ax.plot(sub_data["sizes"].values, sub_data["speedup"].values, "x",
```

benchmarks/benchmark.py
t4c1: I tried hsv first, but subjectively I find colors generated this way easier to distinguish. I think the issue is that h,s and v components affect end color in different magnitudes.

benchmarks/benchmark.py
t4c1: It seems matplotlib gets column labels out of data frames and sets them as labels. As I do two plots separately (lines and x-es) I get duplicated labels. I tried setting label=None, but it does not help.

ahartikainen: I think `label=__nolegend__`

t4c1: That works.

test/expressions/generateExpressionTests.py
```diff
@@ -44,7 +44,7 @@ def make_arg_code(arg, scalar, var_name, var_number, function_name):
         return_type = arg_types[arg.split(" => ")[1]].replace("SCALAR", scalar)
         return "  stan::test::test_functor " + var_name
     arg_type = arg_types[arg].replace("SCALAR", scalar)
-    if function_name in special_arg_values:
+    if function_name in special_arg_values and arg != "rng":
```
ahartikainen, line 47:
```suggestion
    if (function_name in special_arg_values) and (arg != "rng"):
```

That is fine, I usually add parenthesis just to increase some readability.

benchmarks/benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+    fig, ax = plt.subplots(figsize=(10, 10))
+    fig.set_tight_layout(True)
+    ax.set_xscale("log")
+    if plot_log_y:
+        ax.set_yscale("log")
+
+    ax.set_xlabel("size")
+    ax.set_ylabel("time[us]")
+    for n, (signature, sub_data) in enumerate(timing_data.groupby("signatures")):
+        ax.plot(sub_data["sizes"], sub_data["times"], "x", color=pick_color(n), label='__nolegend__')
```
ahartikainen, line 137:
```suggestion
        ax.plot(sub_data["sizes"], sub_data["times"], "x", color=pick_color(n), label='_nolegend_')
```

Yeah, apparently matplotlib devs usually use one underscript, not two. I think anything starting with _ will be ignored.

benchmarks/benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+
+    timing_data = pandas.concat([data["name"].str.split("/", expand=True).iloc[:, :2], data["real_time"]], axis=1)
+    reference_timing_data = pandas.concat(
+        [reference_data["name"].str.split("/", expand=True).iloc[:, :2], reference_data["real_time"]], axis=1)
+    timing_data.columns = reference_timing_data.columns = ["signatures", "sizes", "times"]
+
+    same_in_last_selector = reference_timing_data["signatures"].isin(timing_data["signatures"])
+    reference_timing_data = reference_timing_data.loc[same_in_last_selector, :]
+    assert numpy.all(reference_timing_data["signatures"] == timing_data["signatures"])
+    assert numpy.all(reference_timing_data["sizes"] == timing_data["sizes"])
```
ahartikainen, line 187:
```suggestion
    assert (reference_timing_data["signatures"] == timing_data["signatures"]).all()
    assert (reference_timing_data["sizes"] == timing_data["sizes"]).all()
```

benchmarks/benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+    fig.set_tight_layout(True)
+    ax.set_xscale("log")
+    if plot_log_y:
+        ax.set_yscale("log")
+
+    ax.set_xlabel("size")
+    ax.set_ylabel("speedup")
+
+    for n, (signature, sub_data) in enumerate(timing_data.groupby("signatures")):
+        ax.plot(sub_data["sizes"], sub_data["speedup"], "x", color=pick_color(n), label='__nolegend__')
```
ahartikainen, line 202:
```suggestion
        ax.plot(sub_data["sizes"], sub_data["speedup"], "x", color=pick_color(n), label='_nolegend_')
```

benchmarks/benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+        avg_sig_speedups = sub_data.groupby(by="sizes")["speedup"].median().reset_index().sort_values(by="sizes")
+        ax.plot(avg_sig_speedups["sizes"], avg_sig_speedups["speedup"], label=signature,
+                color=pick_color(n))
+    plt.plot([1, max(timing_data["sizes"])], [1, 1], "--", color="gray")
```
ahartikainen, line 206: What is this trying to plot?

benchmarks/benchmark.py
```diff
@@ -0,0 +1,584 @@
...
+
+    [spine.set_visible(False) for loc, spine in ax.spines.items() if loc in ["top", "right", "left", "bottom"]]
+    ax.minorticks_off()
+    ax.grid()
+    ax.legend()
+    if out_file == "window":
+        plt.show()
+    else:
+        fig.savefig(out_file, bbox_inches="tight")
```
ahartikainen, line 215: I would recommend adding a dpi setting here.

I usually add `dpi=300` but even `dpi=600` depending where I want to use the figure.

The default is 70 (?)

runTests.py
```diff
@@ -290,6 +290,7 @@ def handleExpressionTests(tests, only_functions, n_test_files):
             del tests[n]
             expression_tests = True
     if expression_tests:
+        sys.path.append("test")
         sys.path.append("test/expressions")
```
ahartikainen, line 294: Maybe update these too to use the relative location?

ahartikainen (commented):
These start to look good (python part).

Can you run these files through `black` and `isort`?

Run `isort` first and then black. They have a bit different default style so this way black format is used.

    isort path/to/file
    black path/to/file

benchmarks/benchmark.py
t4c1: This is just a reference line to make it more obvious for which sizes speedups are larger/smaller than 1.

t4c1:
@ahartikainen I think I addressed all your comments. Is there anything else to fix?

benchmarks/benchmark.py
```diff
@@ -0,0 +1,754 @@
...
+                f.seek(f.tell() - len(line) - 1, os.SEEK_SET)
+                break
+        data = pandas.read_csv(f)
+    timing_data = pandas.concat(
+        [data["name"].str.split("/", expand=True).iloc[:, :2], data["real_time"]],
+        axis=1,
+    )
+    timing_data.columns = ["signatures", "sizes", "times"]
+    timing_data.loc[:, "sizes"] = timing_data["sizes"].astype(int)
+    timing_data.loc[:, "times"] /= 1000  # convert to microseconds
```
ahartikainen, line 130: This doesn't need any change, but is there a way to force bench lib to output results always in ns or us?

benchmarks/benchmark.py
```diff
@@ -0,0 +1,754 @@
...
+    Plots benchmark speedup compared to reference results.
+    :param csv_filename: path to csv file containing results to plot
+    :param reference_csv_filename: path to csv file containing reference results to plot
+    :param out_file: path to image file to store figure into. If it equals to "window" opens it in an interactive window.
+    """
+    import pandas, numpy, matplotlib
+
+    if out_file != "window":
+        matplotlib.use("Agg")
+    from matplotlib import pyplot as plt
```
ahartikainen, line 184:
```suggestion
    import matplotlib.pyplot as plt
```

benchmarks/benchmark.py
```diff
@@ -0,0 +1,754 @@
...
+                code += "    stan::math::grad();\n"
+            result += BENCHMARK_TEMPLATE.format(
+                benchmark_name=benchmark_name,
+                setup=setup,
+                var_conversions=var_conversions,
+                code=code,
+                multi=multiplier,
+                max_size=max_size,
+            )
+    cpp_filepath = WORKING_FOLDER + cpp_filename
```
ahartikainen, line 479:
```suggestion
    cpp_filepath = os.path.join(WORKING_FOLDER, cpp_filename)
```

It is safer to use os.path if WORKING_FOLDER is changed by the user

ahartikainen (dismissed):
Hi, looks good.

One minor improvement (os.path)

I added only couple comments, feel free to ignore.

benchmarks/benchmark.py
t4c1: It can only be modified by changing the benchmark source. I prefer to keep the default nanoseconds and do any conversions only when plotting.

t4c1:
I commited your suggestions, so this needs another approval.

ahartikainen (approved):

rok-cesnovar (approved):
Approving for @ahartikainen (seems Ari does not have permissions).

ahartikainen:
Yes, I don't have write permissions for math, and it is fine.

PR #2265, opened 2020-12-17
Profiling

rok-cesnovar:
This is the basic implementation for profiling. No comments, using silly names for now. But it works for non-threaded cases.
```cpp
#include <stan/math.hpp>
#include <iostream>

using namespace stan::math;

int main(int argc, char **argv) {
    int size = 100;
    if (argc > 1) {
        size = atoi(argv[1]);
    }
    std::cout << size << std::endl;
    matrix_d m1 = Eigen::MatrixXd::Random(size, size);

    profile_map profiles;

    matrix_v m2 = m1;
    matrix_v m3 = m1;
    matrix_v m4;
    var_value<matrix_d> m2_varmat = m1;
    var_value<matrix_d> m3_varmat = m1;
    var_value<matrix_d> m4_varmat;
    {
        profile profile_varmat = profile("varmat", profiles);
        m4_varmat = multiply(m2_varmat, m3_varmat);
    }
    
    {
        profile profile_matvar = profile("matvar", profiles);
        m4 = m2 * m3;
    }
    var s = sum(m4) + sum(m4_varmat);
    s.grad();


    profile_map::iterator it;
    for (it = profiles.begin(); it != profiles.end(); it++)
    {
        std::cout << it->first
                << ": "
                << it->second.fwd_pass
                << " - "
                << it->second.rev_pass
                << std::endl;
    }
}
```
prints

```
./profiling 500
matvar: 1.63821 - 1.178
varmat: 0.0214372 - 0.038633
```


rok-cesnovar:
For threading I think I need to just include the thread ID in the profiling info and it should work. Will work on that tomorrow. Any comments welcome.

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,76 @@
...
+namespace math {
+
+namespace internal {
+
+inline auto profile_start(std::string name, profile_map& profiles) {
+  profile_map::iterator it = profiles.find(name);
+  if (it == profiles.end()) {
+    profiles[name] = {};
+  }
+  if (profiles[name].meta.fwd_pass_active
```
bbbales2, line 22: Save a reference to the underlying object and pass that around to avoid all the map lookups. There's always notes on the C++ docs about which operators invalidate references/iterators (https://en.cppreference.com/w/cpp/container/map/erase). I dunno if `iterators` are guaranteed to not need destructor called or I'd say copy those around.

stan/math/rev/core/typedefs.hpp
```diff
@@ -47,6 +47,25 @@ using vector_vi = Eigen::Matrix<vari*, Eigen::Dynamic, 1>;
...
+};
+
+struct profile_info {
+  double fwd_pass;
+  double rev_pass;
+  int count_rev;
+  profile_meta meta;
+};
+
+using profile_map = std::map<std::string, profile_info>;
```
bbbales2, line 67: `unordered_map` might be better since it's hashmap instead of tree

bbbales2 (commented):
I made C++ comments just so I could do it before @t4c1 and @SteveBronder got to it weeeee

stan/math/rev/core/typedefs.hpp
bbbales2: Oh okay, TBB has utilities for this sort of stuff: https://www.threadingbuildingblocks.org/docs/help/reference/containers_overview/concurrent_unordered_map_cls.html

rok-cesnovar: Thanks!

stan/math/rev/core/profiling.hpp
rok-cesnovar: DOne.

stan/math/prim/core/profiling.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <iostream>
+#include <sstream>
+#include <thread>
+
+namespace stan {
+namespace math {
+
+namespace internal {
+template <typename T, require_all_not_var_t<T>* = nullptr>
+inline void profile_start(std::string name, profile_map& profiles) {
```
bbbales2, line 16: I think passing profile names by `const char*` will be lower overhead. I know once the error messages got changed from `const char*` to `std::string` and it caused slowdowns cause the constructors or something.

rok-cesnovar: Agreed that const char* has a lower overhead. The question is more how much do we care about overhead in profiling? std::string is safer and has more C++ lib function to work with it.

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,179 @@
...
+ * Starts profiling the forward pass for the profile
+ * with the specified name and adds a var that stops
+ * profiling the reverse pass in the reverse callback.
+ * 
+ * @tparam T profile type which must not a var
+ * 
+ * @param name the name of the profile to start
+ * @param profiles the map used for storing profiling info
+ */
+template <typename T, require_all_var_t<T>* = nullptr>
```
bbbales2, line 96: Just `require_var_t<T>`

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,179 @@
...
+  if (profile.meta.fwd_pass_active) {
+    std::ostringstream msg;
+    msg << "Profile '" << name << "' ";
+    msg << "already started!";
+    throw std::runtime_error(msg.str());
+  }
+  profile.meta.fwd_pass_start_tp = steady_clock::now();
+  profile.meta.fwd_pass_active = true;
+  profile.meta.start_chain_stack_size = ChainableStack::instance_->var_stack_.size();
+  reverse_pass_callback([name, &profile]() mutable {
```
bbbales2, line 109: name unused

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,179 @@
...
+    profile.rev_pass_time += duration<double>(steady_clock::now()
+                             - profile.meta.rev_pass_start_tp).count();
+    profile.meta.rev_pass_active = false;
```
bbbales2, line 112: I guess it should throw an error if it's already been stopped? That might be too much error checking, cause it should never be possible.

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,179 @@
...
+ * When T is not var, the constructor and destructor only profile the
+ * 
+ * 
+ * @tparam T type of profile class. If var, the created object is used
+ * to profile reverse mode AD. Only profiles the forward pass otherwise.
+ */ 
+template <typename T>
+class profile {
+  const char* name_;
+  profile_info* profile_;
```
bbbales2, line 161: This can be a reference

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,179 @@
...
+ * places a var with a callback to start the profile for the reverse pass.
+ * When T is not var, the constructor and destructor only profile the
+ * 
+ * 
+ * @tparam T type of profile class. If var, the created object is used
+ * to profile reverse mode AD. Only profiles the forward pass otherwise.
+ */ 
+template <typename T>
+class profile {
+  const char* name_;
```
bbbales2, line 160: Honestly now that I made you change these I'm scared that the compiler might put the strings at different locations. Cuz the map here is going to work on the pointer value, not the string value.

stan/math/rev/core/typedefs.hpp
```diff
@@ -5,6 +5,11 @@
 #include <stan/math/rev/meta.hpp>
 #include <stan/math/rev/core/var.hpp>
 #include <stan/math/rev/core/Eigen_NumTraits.hpp>
+#include <chrono>
+#include <string>
+#include <thread>
+#include <map>
+#include <utility>
```
bbbales2, line 12: I think put the ones you need in the profiling header, unless there's a reason to put them here.

test/unit/math/rev/core/profiling_test.cpp
```diff
@@ -0,0 +1,20 @@
...
+  {
+        profile<var> t1 = profile<var>("t1", profiles);
+        a = 2.0;
+        b = 3.0;
+        c = a + b;
+  }
+  c.grad();
+  stan::math::profile_key key = {"t1", std::this_thread::get_id()};
+  EXPECT_EQ(profiles[key].chain_stack_size_sum, 1);
+  EXPECT_EQ(profiles[key].nochain_stack_size_sum, 1);
```
bbbales2, line 18: Why is this not 2? There is `a` and `b`.

stan/math/prim/fun/typedefs.hpp
```diff
@@ -28,6 +28,32 @@ using vector_d = Eigen::Matrix<double, Eigen::Dynamic, 1>;
  */
 using row_vector_d = Eigen::Matrix<double, 1, Eigen::Dynamic>;
 
+/**
+ * Types used for profiling.
+ */
+struct profile_meta {
```
bbbales2, line 34: I think just make it one big struct, and turn it into a class so you can have default constructors.

bbbales2 (commented):
Very useful comments

stan/math/rev/core/profiling.hpp
rok-cesnovar: I had that here but then removed it because this is internal and on the API level that would mean the object was destructed twice, which cant happen. Not sure if my reasoning is valid tho.

stan/math/rev/core/profiling.hpp
rok-cesnovar: Gladly move back to strings :)

test/unit/math/rev/core/profiling_test.cpp
rok-cesnovar: The nochain size should not be decremented on stop like for the chain stack. I am blaming whoever invented copy/paste :)

test/unit/math/rev/core/profiling_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+  } catch (const std::exception& e) {
+  }
+  stan::math::recover_memory();
+  stan::math::profile_key key_t1 = {"t1", std::this_thread::get_id()};
+  EXPECT_EQ(profiles[key_t1].get_chain_stack_used(), 1);
+  EXPECT_EQ(profiles[key_t1].get_nochain_stack_used(), 2);
+  EXPECT_EQ(profiles[key_t1].get_num_rev_passes(), 0);
+  EXPECT_EQ(profiles[key_t1].get_num_fwd_passes(), 1);
+  EXPECT_TRUE(profiles[key_t1].get_fwd_time() > 0.0);
+  EXPECT_TRUE(profiles[key_t1].get_rev_time() == 0.0);
```
bbbales2, line 69: Oh and this is why it's nice to not check the duplicate stuff on the reverse pass -- we can have exceptions then and can't always clean up the start/stops there.

test/unit/math/rev/core/threading_profiling_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+  for (std::size_t i = 0; i != elems; ++i) {
+    vref_lambda_v.push_back(vlambda_v[gidx[i]]);
+  }
+
+  var poisson_lpdf_ref = stan::math::poisson_lpmf(data, vref_lambda_v);
+  var c = 1.0;
+  {
+    stan::math::profile<stan::math::var> p1
+        = stan::math::profile<stan::math::var>(
+            "p1", profiling_test::profiles_threading);
```
bbbales2, line 64:
```
stan::math::profile<stan::math::var> p1("p1", profiling_test::profiles_threading);
```

test/unit/math/rev/core/profiling_test.cpp
```diff
@@ -0,0 +1,93 @@
+#include <stan/math/rev.hpp>
+#include <gtest/gtest.h>
+
+TEST(Profiling, profile_double_basic) {
+  using stan::math::profile;
+  using stan::math::var;
+  stan::math::profile_map profiles;
+  double a = 3.0, b = 2.0, c;
+  {
+    profile<double> p1 = profile<double>("p1", profiles);
```
bbbales2, line 10: ```profile<double> p1("p1", profiles);```

bbbales2 (changes requested):
I like the class! Comments

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+    fwd_pass_tp_ = std::chrono::steady_clock::now();
+    active_ = true;
+  }
+
+  template <typename T>
+  void fwd_pass_stop() {
+    if (!is_constant<T>::value) {
+      n_fwd_AD_passes_++;
+      chain_stack_size_sum_ += (ChainableStack::instance_->var_stack_.size() - start_chain_stack_size_ - 1);
```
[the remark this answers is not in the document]
rok-cesnovar: Indeed.

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,202 @@
...
+  void rev_pass_start() {
+    rev_pass_tp_ = std::chrono::steady_clock::now();
+  }
+
+  void rev_pass_stop() {
+    rev_pass_time_ += std::chrono::duration<double>(std::chrono::steady_clock::now() - rev_pass_tp_).count();
+    n_rev_passes_++;
+  }
+
+  size_t get_chain_stack_used() {
```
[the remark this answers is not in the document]
rok-cesnovar: Makes sense.

rok-cesnovar:
Thanks! Addressed all comments, should be ready for another look.

stan/math/rev/core/profiling.hpp
```diff
@@ -0,0 +1,167 @@
...
+    profile_ = &profiles[key_];
+    if (profile_->is_active()) {
+      std::ostringstream msg;
+      msg << "Profile '" << key_.first << "' already started!";
+      throw std::runtime_error(msg.str());
+    }
+    profile_->fwd_pass_start<T>();
+    if (!is_constant<T>::value) {
+      profile_info& profile_ref = *profile_;
+      reverse_pass_callback([&profile_ref]() mutable { profile_ref.rev_pass_stop(); });
```
rok-cesnovar, line 153: I need to do 
```cpp
profile_info& profile_ref = *profile_;
```
in order for this to pass the address sanitizer. Is there a nicer way?

cc @t4c1

bbbales2:
```
reverse_pass_callback([profile_]() mutable { profile_->rev_pass_stop(); });
```

I think. Copy the pointer by value and use it in the reverse pass.

rok-cesnovar: > reverse_pass_callback([profile_]() mutable { profile_->rev_pass_stop(); });

Does not compile.

> reverse_pass_callback([this]() mutable { profile_->rev_pass_stop(); });

compiles but fails withe the address sanitizer.

rok-cesnovar: I am fine with the way its done now as it looks to be safe and it also works. But if there is a shorter way I am all ears.

t4c1: You can try:
`reverse_pass_callback([profile=this->profile_]() mutable { profile->rev_pass_stop(); })`
But I have no idea if it will work.

rok-cesnovar: Works! Tadej to the rescue once again. Thanks.

test/unit/math/rev/core/profiling_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+    profile<var> t1("t1", profiles);
+    var a = i;
+    var b = 2.0;
+    c = c + a;
+  }
+  c.grad();
+  stan::math::recover_memory();
+  stan::math::profile_key key_t1 = {"t1", std::this_thread::get_id()};
+  EXPECT_EQ(profiles[key_t1].get_chain_stack_used(), 100);
+  EXPECT_EQ(profiles[key_t1].get_nochain_stack_used(), 200);
```
bbbales2, line 96: `N * 1` and `N * 2` instead of 100 and 200

bbbales2 (approved):
Good!

test/unit/math/rev/core/profiling_test.cpp
rok-cesnovar: Fixed. Needs reapproving now :)

bbbales2 (approved):
Good!

PR #2266, opened 2020-12-18
Make `varmat` functions work with ldlt (Issue #2101)

bbbales2:
## Summary

Second attempt at making `varmat` work with ldlt

## Side Effects

I removed `LDLT_alloc` and replaced `LDLT_factor` with an API incompatible replacement.

So this is a major version bump for Math right?

## Release notes

- Add `varmat` implementation of `*_ldlt_*` functions

## Checklist

- [x] Math issue #2101


test/unit/math/mix/fun/trace_gen_inv_quad_form_ldlt_test.cpp
```diff
@@ -64,8 +55,8 @@ TEST(mathMixMatFun, traceGenInvQuadForm) {
 
   // exception tests
   // non-square second arg
-  Eigen::MatrixXd a34(3, 4);
```
bbbales2, line 67: Can't take an LDLT of a non-square matrix, so this fails before it gets to `trace_gen_inv_quad_form_ldlt` (which takes an LDLT as an argument)

bbbales2:
I ended up replacing the ldlt autodiff that was there, but it's still not great. There new `LDLT_factor` is incompatible with the old one. We should still probably remove this eventually. The new version is cleaner and simpler, but it's a kinda-sometimes autodiff object that doesn't really behave like the other autodiff things we have (that is why it takes a template parameter which says whether to allocate it in the arena or not).

If we're gonna have LDLT factors as autodiff variables, they need to be `vari_value<LDLT>` types or something and follow the rules (and similarly the functions that consume them need to not make any extra assumptions about the lifetime of objects, which they currently do). In the interim, we have this.

Still need add a three argument `expect_ad_matvar` so that `trace_gen_inv_quad_form` can be tested with `varmat` types.

Will also get some benchmarks to make sure the `matvar` stuff isn't slower. When I coded this stuff up before with `reverse_pass_callback` I remeber the existing stuff was pretty fast.

bbbales2:
Here's benchmarks:

ldlt_timings

And here's the timing results of the one where performance gets bad. Not sure what happened there. I didn't spend time looking into it:

ldlt_trace_inv

SteveBronder:
Is this ready for review?

bbbales2:
@SteveBronder I'm trying to get it greenlit, and it has a 3-argument varmat function that is waiting on testing, but yeah all the LDLT stuff is ready for review.

stan/math/fwd/fun/mdivide_left_ldlt.hpp
```diff
@@ -22,14 +22,16 @@ namespace math {
...
+                     EigMat::ColsAtCompileTime>
+mdivide_left_ldlt(const LDLT_factor<Eigen::MatrixXd, alloc_on_stack>& A,
+                  const EigMat& b) {
   using T = typename value_type_t<EigMat>::Scalar;
   constexpr int R2 = EigMat::RowsAtCompileTime;
   constexpr int C2 = EigMat::ColsAtCompileTime;
-  check_multiplicable("mdivide_left_ldlt", "A", A, "b", b);
+  check_multiplicable("mdivide_left_ldlt", "A", A.matrix(), "b", b);
 
   const Eigen::Ref<const plain_type_t<EigMat>>& b_ref = b;
```
SteveBronder, line 36: Not part of PR but would be good to replace with 
```suggestion
  const auto& b_ref = to_ref(b);
```

bbbales2:
The tests keep failing on these Wishart functions. Instead of checking that their inputs were symmetric, they were just using the lower triangle of the input. Anyway that caused some sort of something to fail, and since it's non-standard (everywhere else in Stan we just check that input matrices are symmetric) I changed it.

So if you wonder about all the inverse wishart/wishart changes in here -- that's the answer

stan/math/prim/fun/LDLT_factor.hpp
```diff
@@ -12,120 +12,69 @@ namespace stan {
...
+ * holds a matrix of type T and the LDLT of its values.
+ *
+ * @tparam T type of elements in the matrix
+ */
+template <typename T>
+class LDLT_factor<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>, false> {
+ private:
+  using ldlt_type
+      = Eigen::LDLT<Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>>;
+  std::shared_ptr<ldlt_type> ldlt_ptr_;
```
SteveBronder, line 41: I'm not following why this needs to be a shared pointer (or pointer at all?)

bbbales2:
Oh yeah something to keep in mind reviewing this, there are only like 5 functions that have custom ldlt autodiff.

```
log_determinant_ldlt
mdivide_left_ldlt
mdivide_right_ldlt (which is implemented using mdivide_left_ldlt)
trace_inv_quad_form
trace_gen_inv_quad_form
```

So even though there are a lot of files here, if we need to make changes it's not that bad (cuz this is the extent of the difficult changes).

stan/math/prim/fun/LDLT_factor.hpp
```diff
@@ -12,120 +12,69 @@ namespace stan {
...
+ * @return LDLT_factor of A
+ */
+template <typename T, typename... Args, require_matrix_t<T>* = nullptr>
+inline auto make_ldlt_factor(const T& A) {
+  return LDLT_factor<
+      plain_type_t<T>,
+      disjunction<is_var<scalar_type_t<return_type_t<T, Args...>>>,
+                  is_var<partials_type_t<return_type_t<Args...>>>,
+                  is_var<partials_type_t<
+                      partials_type_t<return_type_t<Args...>>>>>::value>(A);
```
SteveBronder, line 75: I'm guessing this is for `var`, `fvar<var>` and `fvar<fvar<var>>`. Maybe we should just have a type trait like `contains_var` that recurses through `fvar<>`?

stan/math/prim/fun/mdivide_right_ldlt.hpp
```diff
@@ -23,25 +23,23 @@ namespace math {
  * @return x = b A^-1, solution of the linear system.
  * @throws std::domain_error if rows of b don't match the size of A.
  */
-template <typename EigMat, typename T, int R, int C,
+template <typename EigMat, typename T, bool alloc_in_arena,
           require_eigen_t<EigMat>* = nullptr,
           require_any_not_arithmetic_t<value_type_t<EigMat>, T>* = nullptr>
```
SteveBronder, line 28: Can `T` be anything but an Eigen matrix?

stan/math/prim/prob/inv_wishart_lpdf.hpp
```diff
@@ -58,12 +58,14 @@ return_type_t<T_y, T_dof, T_scale> inv_wishart_lpdf(const T_y& W,
   Eigen::Index k = S.rows();
   T_nu_ref nu_ref = nu;
   T_S_ref S_ref = S;
+  T_W_ref W_ref = W;
   check_greater(function, "Degrees of freedom parameter", nu_ref, k - 1);
+  check_symmetric(function, "random variable", W_ref);
+  check_symmetric(function, "scale parameter", S_ref);
 
-  LDLT_factor<value_type_t<T_y>, Eigen::Dynamic, Eigen::Dynamic> ldlt_W(W);
+  const auto& ldlt_W = make_ldlt_factor<T_y, T_dof, T_scale>(W_ref);
```
SteveBronder, line 66: I'm not understanding this switch to depending on multiple types. And why is this `const auto&`? And should these be using the `T_*_ref` types?

stan/math/prim/prob/inv_wishart_lpdf.hpp
```diff
@@ -87,8 +89,7 @@ return_type_t<T_y, T_dof, T_scale> inv_wishart_lpdf(const T_y& W,
     //      &S(0), S.size(), 1);
     //    lp -= 0.5 * dot_product(S_vec, W_inv_vec); // trace(S * W^-1)
     Eigen::Matrix<return_type_t<T_y, T_scale>, Eigen::Dynamic, Eigen::Dynamic>
-        Winv_S(mdivide_left_ldlt(
-            ldlt_W, S_ref.template selfadjointView<Eigen::Lower>()));
+        Winv_S(mdivide_left_ldlt(ldlt_W, S_ref));
```
SteveBronder, line 92: Why did the `.template selfadjointView<Eigen::Lower>()` get deleted?

stan/math/rev/fun/LDLT_factor.hpp
```diff
@@ -3,140 +3,93 @@
...
-#else
-  template <typename Rhs>
-  inline const Eigen::internal::solve_retval<
-      Eigen::LDLT<Eigen::Matrix<double, R, C> >, Rhs>
-  solve(const Eigen::MatrixBase<Rhs> &b) const {
-    return alloc_->ldlt_.solve(b);
-  }
-#endif
+template <>
+class LDLT_factor<Eigen::Matrix<var, Eigen::Dynamic, Eigen::Dynamic>, true> {
```
SteveBronder, line 47: I think it would be nicer / safer if this used SFINAE so if a user passed `const Eigen::Matrix<var, -1, -1>&` here the require would still catch it

SteveBronder: also idt this scheme would work for `arena_matrix<Eigen::Matrix<var, -1, -1>>`

stan/math/rev/fun/log_determinant_ldlt.hpp
```diff
@@ -3,51 +3,37 @@
...
+  if (A.matrix().size() == 0) {
     return 0;
   }
 
-  return var(new internal::log_det_ldlt_vari<R, C>(A));
+  var log_det = sum(log(A.ldlt().vectorD().array()));
+
+  arena_t<Eigen::MatrixXd> arena_A_inv(A.matrix().rows(), A.matrix().cols());
+
+  arena_A_inv.setIdentity();
```
SteveBronder, line 29: You can just immediately fill this with the identity like 

```suggestion
  arena_t<Eigen::MatrixXd> arena_A_inv(Eigen::MatrixXd::Identity(A.matrix().rows(), A.matrix().cols()));
```

stan/math/rev/fun/mdivide_left_ldlt.hpp
```diff
@@ -13,275 +12,67 @@
...
-  auto *baseVari
-      = new internal::mdivide_left_ldlt_vv_vari<R1, C1, R2, C2>(A, b);
+  if (!is_constant<T1>::value && !is_constant<T2>::value) {
+    arena_t<promote_scalar_t<var, T2>> arena_B = B;
+    arena_t<ret_type> res = A.ldlt().solve(arena_B.val());
 
-  Eigen::Matrix<var, R1, C2> res(b.rows(), b.cols());
-  res.vi() = Eigen::Map<matrix_vi>(baseVari->variRefC_, res.rows(), res.cols());
+    reverse_pass_callback([A, arena_B, res]() mutable {
+      promote_scalar_t<double, T2> adjB = A.ldlt().solve(res.adj());
```
SteveBronder, line 45: [Q] Does `.ldlt()` do something then solve happens? Like could you store that ldlt for A and then just have to do it once?

stan/math/rev/fun/mdivide_left_ldlt.hpp
```diff
@@ -13,275 +12,67 @@
...
 
-  return res;
-}
+      forward_as<promote_scalar_t<var, T1>>(A.matrix()).adj()
+          -= adjB * res.val().transpose().eval();
+      arena_B.adj() += adjB;
```
SteveBronder, line 49: I think if you wrote a specialization for `arena_t<>` for `ldlt_factor<>` we could avoid this `forward_as<>` by just using it here.

I'm also confused on why we bring the whole `ldlt_factor<>` into here when it looks like we just use the `.matrix()` part?

stan/math/rev/fun/trace_gen_inv_quad_form_ldlt.hpp
```diff
@@ -28,29 +28,141 @@ namespace math {
...
+                           * (arena_D.val_op() + arena_D.val_op().transpose());
+          arena_D.adj() += C_adj * BTAsolveB;
+        });
+
+    return res;
+  } else if (!is_constant<Ta>::value && !is_constant<Tb>::value
+             && is_constant<Td>::value) {
+    arena_t<promote_scalar_t<var, Tb>> arena_B = B;
+    arena_t<promote_scalar_t<double, Td>> arena_D = value_of(D);
+    auto AsolveB = to_arena(A.ldlt().solve(arena_B.val()));
```
SteveBronder, line 70: Is the return type here `arena_t<Eigen::MatrixXd>`?

SteveBronder (changes requested):
Quick review, just skimmed over half of it but I think a few comments in individual places apply in multiple placese

stan/math/prim/fun/LDLT_factor.hpp
bbbales2: 1. Makes copies of this cheap (it gets copied into `reverse_pass_callback`)
2. Handles the deletin' for me

stan/math/prim/fun/LDLT_factor.hpp
bbbales2: Yeah, this is what it is doing, and yes, that sounds handy for this, though I'm not sure we'd use it anywhere other than here.

Normally in Stan an operation is responsible for saving everything it needs in the reverse pass to the arena. `LDLT_factor` allocates itself in a way that it will be saved if it needs saved (instead of leaving this up to the op). I kinda want to change it, though it might make the functions a bit slower (cuz there'd be extra copies).

stan/math/prim/prob/inv_wishart_lpdf.hpp
bbbales2: It was giving me errors so I deleted it. Maybe a `var_value<T>` thing? I'm not sure.

Then I realized that none of the wishart/inverse wishart functions were checking their arguments for symmetry -- they were just using the lower triangle of the input argument (or assuming things were symmetric). I just ended up changing how all that worked. Maybe that should go in a different pull.

stan/math/rev/fun/mdivide_left_ldlt.hpp
bbbales2: `.ldlt()` is a member function accessing a saved LDLT decomposition, so this is what is happening.

bbbales2: Avoiding multiple LDLTs is the upside of using all this `LDLT_factor` stuff.

stan/math/rev/fun/trace_gen_inv_quad_form_ldlt.hpp
bbbales2: Could be a vector or a matrix

stan/math/rev/fun/LDLT_factor.hpp
bbbales2: Fair. My intention was to always build these through `stan::math::make_ldlt_factor` which can kinda hide this stuff.

stan/math/rev/fun/mdivide_left_ldlt.hpp
bbbales2: `ldlt_factor<>` itself is safe and cheap to pass around (since the member variables are either `arena_t` types or pointers).

bbbales2: Oh I could just pull out the matrix and save that in an `arena_t` and then I don't have the `forward_as` code, which is nice to avoid.

stan/math/prim/fun/LDLT_factor.hpp
bbbales2: Reading this again I want to try removing `alloc_in_arena` and see how expensive that actually ends up being. It's kindof an annoying thing.

stan/math/rev/fun/mdivide_left_ldlt.hpp
bbbales2: I think really the thing we were going for here is `vari_value<Eigen::LDLT<Eigen::MatrixXd>>`.

It's a matrix autodiff type represented by its `LDLT` decomposition.

I'm nervous about trying to write that though, since it's quite a different thing than we've worked with before and this is 5 functions used internally in `stan::math` and not exposed at the Stan language level.

The counter argument to that is `LDLT_factor` is different from everything else as well.

SteveBronder: > `vari_value<Eigen::LDLT<Eigen::MatrixXd>>`

idt we want that do we? That would have both `val_` and `adj_` as `LDLT`s, is that what we want?

stan/math/rev/fun/mdivide_left_ldlt.hpp
SteveBronder: icic cool cool cool

stan/math/prim/prob/inv_wishart_lpdf.hpp
SteveBronder: tbh idk if it's more expensive to check that the thing is symmetric and use `selfAdjointView<>` or if it's just better to pass the whole matrix here. I'd guess it depends on how it's used. I'll leave that call to you. If we drop the need for symmetry here then you can remove that adjoint view. But if we need to check that anyway then self adjoint view will be faster since your telling eigen that matrix is the same across the lower/upper triangular

stan/math/rev/fun/mdivide_left_ldlt.hpp
bbbales2: The idea would be to make something like `vari_value<Eigen::MatrixXd>` except `.ldlt` would already be computed.

stan/math/prim/fun/LDLT_factor.hpp
SteveBronder: Shared pointers are pretty expensive. Each one essentially has a control block that keeps track of the count. Could we use unique_ptr here?

I think I need to read more about how this type is used to have a better grasp of this

stan/math/rev/fun/mdivide_left_ldlt.hpp
SteveBronder: Why not just have a specialization for `LDLT_factor` then?

stan/math/prim/fun/LDLT_factor.hpp
SteveBronder: You could just have it be an internal type trait only defined here in ldlt_factor (lest we think it would be used elsewhere)

stan/math/prim/prob/inv_wishart_lpdf.hpp
bbbales2: It's more expensive this way. THe other way is kinda weird (pass an asymmetric matrix and it'll just silently use the bottom half and work).

stan/math/prim/fun/LDLT_factor.hpp
bbbales2: I ditched the shared pointer and made LDLT_factors heavy weight objects (that shouldn't be copied)

stan/math/prim/fun/LDLT_factor.hpp
bbbales2: I got rid of this type checking.

stan/math/prim/prob/inv_wishart_lpdf.hpp
bbbales2: Got rid of the multiple type dependency

stan/math/prim/prob/inv_wishart_lpdf.hpp
bbbales2: Addressed separately here: https://github.com/stan-dev/math/pull/2281

stan/math/rev/fun/LDLT_factor.hpp
bbbales2: I did SFINAE on the class, but I kept the factory function around cuz I like it too.

stan/math/rev/fun/mdivide_left_ldlt.hpp
bbbales2: Now I don't copy `LDLT_factor` variables into the reverse pass. If I need either the Eigen decomposition, I save a copy with `make_chainable_object`. If I need the matrix varis, I save those with `arena_t`. It's more copying overhead, but we're still re-using the LDLT decompositions (which are `O(N^3)` and more expensive than the copies).

bbbales2:
This needs #2280 in before it is ready to go (there's a three argument function in there)

bbbales2:
@SteveBronder is is ready to review now that the three argument tests are in.

SteveBronder (commented):
So reviewing this I didn't like that we had to make multiple copies as we passed around the ldlt factor. Trying to sort this out I think it was just easier to make a little review branch with some changes

https://github.com/stan-dev/math/compare/feature/varmat-ldlt2...review/feature/varmat-ldlt2#diff-8eb184dd0a9220fb56a0aea61b945cfea318ecfbfe8f25fd7ac6050c5f19b724R14

There's only one thing I strongly don't like about the changes I made which is the forward deceleration of `make_chainable_ptr()` in prim. We could flip this around and just declare the method for `chainable_ldlt_ptr()` in prim

```cpp
// in prim's LDLT_factor
inline auto* chainable_ldlt_ptr();
```

 and then in `rev` actually define `chainable_ldlt_ptr()` for the prim version. 

Though if you don't like what I have above then I think then you can ignore that and just use the other review comments

bbbales2:
Changes look good. I'll merge tomorrow. The `arena_t` thing on the var LDLT is a good catch. Thanks!

stan/math/prim/fun/mdivide_left_ldlt.hpp
```diff
@@ -13,29 +13,28 @@ namespace math {
...
+  check_multiplicable("mdivide_left_ldlt", "A", A.matrix(), "b", b);
+
+  if (A.matrix().cols() == 0) {
     return {0, b.cols()};
   }
 
-  return A.solve(
+  return A.ldlt().solve(
       Eigen::Matrix<return_type_t<T, EigMat>, EigMat::RowsAtCompileTime,
                     EigMat::ColsAtCompileTime>(b));
```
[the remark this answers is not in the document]
bbbales2: Good point. I'm gonna leave it as is though. This is some weird type promotion thing so the Eigen solves work. Like if `T` and `EigMat` don't match, it will make them match.

If we added another overload for when the scalar types are the same it would handle this. I grepped for where this function is used (it's an internal function not exposed to the Stan langage) and I don't think it's actually used anywhere (`mdivide_right_ldlt` uses this, but that is unused, and again that is an internal function), so I'm not gonna worry about this (though it would take about 5s to implement -- just seems like more code and complexity some future person has to decode).

bbbales2:
I moved `make_chainable_ptr` back out of `LDLT_factor`. It was kinda awkward to make it work with the prim stuff. The only place we do this is `mdivide_left_ldlt` and I grepped around and I think everywhere we do this there will only be one copy anyway.

This is the grep, and `trace_inv_quad_form_ldlt` and `trace_gen_inv_quad_form_ldlt` are only used in one place each as well:
```
stan/math/prim/fun/trace_inv_quad_form_ldlt.hpp:  return B.cwiseProduct(mdivide_left_ldlt(A, B)).sum();
stan/math/prim/fun/trace_gen_inv_quad_form_ldlt.hpp:  return multiply(B, D.transpose()).cwiseProduct(mdivide_left_ldlt(A, B)).sum();
stan/math/prim/prob/wishart_lpdf.hpp:        mdivide_left_ldlt(ldlt_S, W_ref));
stan/math/prim/prob/inv_wishart_lpdf.hpp:        mdivide_left_ldlt(ldlt_W, S_ref));
```

bbbales2:
@SteveBronder should be ready for review again

SteveBronder (dismissed):
Good!

bbbales2:
@rok-cesnovar this was [approved](https://github.com/stan-dev/math/pull/2266#pullrequestreview-565587697) but then my test restart [commit](https://github.com/stan-dev/math/pull/2266/commits/76733bd997ff5bcacd9c842b5e2f5000bb9b8b88) dismissed the review. Can you approve and merge?

rok-cesnovar (approved):

Issue #1971, opened 2020-07-09
 Implement skewed Laplace distribution


dirmeier:
## Description

Hello, issue https://github.com/stan-dev/stan/issues/2312 asks for an implementation of the skewed Laplace distribution for Stan.  I have some time rn, and will have a look at this.

Cheers,
Simon

#### Current Version:
v3.2.0

SteveBronder:
Closed in https://github.com/stan-dev/math/pull/2271

PR #2271, opened 2020-12-25
Implement skew double exponential distribution

dirmeier:
## Summary

Ciao all,

this PR implements skew double exponential lpdf, lcdf, lccdf and rng functions with analytic gradients https://github.com/stan-dev/math/issues/1971 (sorry for the delay).

So far the pdf, cdf and ccdf functions only take scalar `y`s. The issue is that at some point I need to compute `step(y)` which seems to not work if y is a vector (see [here](https://github.com/dirmeier/math/blob/feature/issue-1971-implement_asymmetric_laplace_distribution/stan/math/prim/prob/skew_double_exponential_lcdf.hpp#L91)). Any help would be appreciated :)

Cheers,
Simon

## Tests

I've implemented tests for lpdf, lcdf and lccdf with scalar arguments and a chi-squared goodness of fit test for the random number generator. However, I'll need to add some more tests for vectorial data/parameters once I've found a way to fix the issue mentioned above.

## Release notes

- Implementation of the skew double exponential distribution (e.g. for quantile regression)

## Checklist

- [x] Copyright holder: Simon Dirmeier


dirmeier:
All done. Functions now allow vector arguments and are tested. One needed only to vectorize the `step` function.
Cheers,
Simon

rok-cesnovar:
Hey @dirmeier, thanks for this!

I will let someone with actual Math chops review this (maybe @bbbales2 once he is back). One thing that I did notice is that the vectorized `step()` is missing tests.

bbbales2:
Thanks @dirmeier ! If I don't get you a review in a few days ping me.

bbbales2 (changes requested):
@dirmeier thanks for this! Had a look.

The step function needs tests to make sure it works with vector types. I think you can just add something like the non-varmat tests [here](https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/fun/asin_test.cpp#L4) to the step function tests [here](https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/fun/step_test.cpp).

You should also add a few tests to the tests [here](https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/fun/step_test.cpp). I think you can just model them after the tests [here](https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/fun/asin_test.cpp).

This seems a bit pedantic, but this should test that the function compiles with all the expected sorts of signatures it might need to work with Stan (autodiff types, etc.). We're also checking that the gradients are in fact zero away from the step (and not checking the gradients at the step).

For the distribution functions like this, there's a separate set of tests that automatically checks vectorization and autodiff are working correctly (which take a lot of work to check manually as I'm sure you've noticed :D).

Here's the version for Skew normal: https://github.com/stan-dev/math/tree/develop/test/prob/skew_normal

So you can copy that folder and build one for skew_double_exponential too (or build off the double_exponential folder if you like)

The tests are set up similarly to how you have the prim tests. For each lpdf, cdf, lcdf, and lccdf, you're expected to provide:

1. A few good parameter sets
2. Some bad parameters
3. An implementation of the function that should work for scalar arguments (like what you have written in the tests [here](https://github.com/stan-dev/math/blob/7046fbce617b2b9c1676137e0d6a8224f2768880/test/unit/math/prim/prob/skew_double_exponential_ccdf_log_test.cpp#L16))

The tests that get run are generated code, so to generate + run them you can do something like this:

```
./runTests.py -j4 test/prob/skew_double_exponential/
```

It takes a long time to run them, even for one distribution, so running one test at a time is good.

A base set of tests to run for the skew_normal are:
```
./runTests.py test/prob/skew_normal/skew_normal_00000_generated_v_test.cpp
./runTests.py test/prob/skew_normal/skew_normal_cdf_00000_generated_v_test.cpp
./runTests.py test/prob/skew_normal/skew_normal_cdf_log_00000_generated_v_test.cpp
./runTests.py test/prob/skew_normal/skew_normal_ccdf_log_00000_generated_v_test.cpp
```

So just work with each of those until they're working. After that's good, check the high order autodiff by doing the tests with `ffv` instead of `v` like:
```
./runTests.py test/prob/skew_normal/skew_normal_00000_generated_ffv_test.cpp
```

All this testing stuff can be kinda annoying the first time you use it (cuz it's all complicated and the errors are difficult to parse), so lemme know if you run into issues.

dirmeier:
Ciao @bbbales2 ,
thanks for the review and the detailed explanations. I managed to get the `skew_double_exponential_00000_generated_v_test.cpp` test running, but three tests in `skew_double_exponential_00000_generated_ffv_test.cpp` fail (file [here](https://github.com/stan-dev/math/pull/2271/files#diff-db3ce5495a008c4c625bec5ec3bbb39282e2bfff5016c6550242cbcc7f1e4c64)). Some excerpt is below.

```
./test/prob/test_fixture_distr.hpp:487: Failure
The difference between expected_gradients[i] and gradients[i] is 1, which exceeds 1e-7, where
expected_gradients[i] evaluates to 1,
gradients[i] evaluates to 0, and
1e-7 evaluates to 9.9999999999999995e-08.
Comparison of expected gradient to calculated gradient failed
./test/prob/test_fixture_distr.hpp:487: Failure
The difference between expected_gradients[i] and gradients[i] is 1, which exceeds 1e-7, where
expected_gradients[i] evaluates to -1,
gradients[i] evaluates to 0, and
1e-7 evaluates to 9.9999999999999995e-08.
Comparison of expected gradient to calculated gradient failed
./test/prob/test_fixture_distr.hpp:487: Failure
The difference between expected_gradients[i] and gradients[i] is 1, which exceeds 1e-7, where
expected_gradients[i] evaluates to 0,
gradients[i] evaluates to 1, and
1e-7 evaluates to 9.9999999999999995e-08.
Comparison of expected gradient to calculated gradient failed
...
[==========] 900 tests from 100 test suites ran. (36 ms total)
[  PASSED  ] 897 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] AgradDistributionSkewDoubleExponential_ffv_81/AgradDistributionTestFixture/0.Function, where TypeParam = boost::mpl::vector<AgradDistributionSkewDoubleExponential, boost::mpl::vector<double, stan::math::fvar<stan::math::fvar<stan::math::var_value<double, void> > >, double, double, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>
[  FAILED  ] AgradDistributionSkewDoubleExponential_ffv_84/AgradDistributionTestFixture/0.Function, where TypeParam = boost::mpl::vector<AgradDistributionSkewDoubleExponential, boost::mpl::vector<double, stan::math::fvar<stan::math::fvar<stan::math::var_value<double, void> > >, double, stan::math::fvar<stan::math::fvar<stan::math::var_value<double, void> > >, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>
[  FAILED  ] AgradDistributionSkewDoubleExponential_ffv_99/AgradDistributionTestFixture/0.Function, where TypeParam = boost::mpl::vector<AgradDistributionSkewDoubleExponential, boost::mpl::vector<double, stan::math::fvar<stan::math::fvar<stan::math::var_value<double, void> > >, stan::math::fvar<stan::math::fvar<stan::math::var_value<double, void> > >, double, empty, empty, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>
```
Are the gradients incorrect here? I have a simular issue for the `skew_double_exponential_cdf_log_00000_generated_v_test.cpp` ([here](https://github.com/stan-dev/math/pull/2271/files#diff-ae3295bda3462490c060c4a39ecd75e721f986717b3690588c6207e3823e9ad4)) where 200 tests fail, all of which seem to be UpperBound and LowerBound. An excerpt below again:

```
[==========] Running 1000 tests from 100 test suites.
[----------] 10 tests from AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0, where TypeParam = boost::mpl::vector<AgradCdfLogSkewDoubleExponential, boost::mpl::vector<double, double, double, double, empty, empty, mpl_::na, m>
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.CallAllVersions
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.CallAllVersions (0 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.ValidValues
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.ValidValues (0 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.InvalidValues
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.InvalidValues (1 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.FiniteDiff
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.FiniteDiff (0 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.Function
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.Function (0 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.RepeatAsVector
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.RepeatAsVector (0 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.LowerBound
unknown file: Failure
C++ exception with description "skew_double_exponential_lcdf: Random variable is -inf, but must be finite!" thrown in the test body.
[  FAILED  ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.LowerBound, where TypeParam = boost::mpl::vector<AgradCdfLogSkewDoubleExponential, boost::mpl::vector<double, double, double, double, empty, empty, mpl_::na, mpl_>
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.UpperBound
unknown file: Failure
C++ exception with description "skew_double_exponential_lcdf: Random variable is inf, but must be finite!" thrown in the test body.
[  FAILED  ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.UpperBound, where TypeParam = boost::mpl::vector<AgradCdfLogSkewDoubleExponential, boost::mpl::vector<double, double, double, double, empty, empty, mpl_::na, mpl_>
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.AsScalarsVsAsVector
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.AsScalarsVsAsVector (0 ms)
[ RUN      ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.Length0Vector
[       OK ] AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0.Length0Vector (0 ms)
[----------] 10 tests from AgradCdfLogSkewDoubleExponential_v_0/AgradCdfLogTestFixture/0 (1 ms total)

```

Are these numerical problems issues or did I just implement it incorrectly? Many thanks!

Best,
Simon

bbbales2:
@dirmeier I'll investigate this and letcha know. Thanks for implementing these

bbbales2:
Re the Upper/Lower bounds, looks like for unbounded distributions inf is okay, and all we do is check nan (like [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/normal_lcdf.hpp#L42)). I switched those out and those went away.

What parameterization are we using for this distribution by the way (so I can check the gradients)? Is it on Wikipedia, or somewhere else where I can stare at the Eqs a bit?

dirmeier:
Ciao @bbbales2 ,

thanks again. I could fix the issues with the gradients and the tests now run up until `test/prob/skew_double_exponential/skew_double_exponential_00004_generated_fv_test.cpp` . Could you give me a hint how I need to extend [`stan/math/prim/fun/step.hpp`](https://github.com/stan-dev/math/pull/2271/files#diff-61a7f4a8538b488983e5a2dfb28dfc42ca81eda735561adb93433d7c7a442f46) to make it work?

```
./test/prob/test_fixture_distr.hpp:825:1:   required from here
./stan/math/prim/fun/step.hpp:48:16: error: cannot convert ‘stan::math::var’ {aka ‘stan::math::var_value<double>’} to ‘double’ in return
   48 |     return step(y);
      |            ~~~~^~~
      |                |
      |                stan::math::var {aka stan::math::var_value<double>}
make: *** [<builtin>: test/prob/skew_double_exponential/skew_double_exponential_00004_generated_fv_test.o] Error 1
make -j1 test/prob/skew_double_exponential/skew_double_exponential_00004_generated_fv_test failed
```

Best,
Simon

bbbales2:
@dirmeier for now I just changed step to not return a double. It'll be a little inefficient but avoids the error. Thanks for the eqs.

I noticed also this doesn't have a `_cdf` implemented. I think in this case we want one. It's just the exp of the lcdf, but we mostly implement them explicitly (for instance [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/normal_cdf.hpp)).

If you want a break I can do that one. These distribution changes are a lot of work.

dirmeier:
Ciao @bbbales2 , nono it's fine, I'll do it. I wanna get more involved in Stan and this is a good opportunity to learn more about it.

bbbales2:
> I wanna get more involved in Stan

Cool beans absolutely!

dirmeier:
Finally works! Thanks for the help @bbbales2 . I added the cdf as discussed and replaced the unit tests with a `VectorTestRig`
in [`skew_double_exponential_test.cpp`](https://github.com/stan-dev/math/pull/2271/files#diff-613fef87bd8db175764c6c94b2209dabf3bed9ffbb943a98660f22008d88e707). I left the pother tests in ` test/unit/math/prim/prob/skew_double_exponential*cpp` , but I guess one can remove them.

stan/math/prim/prob/skew_double_exponential_rng.hpp
```diff
@@ -0,0 +1,80 @@
...
+  scalar_seq_view<T_sigma_ref> sigma_vec(sigma_ref);
+  scalar_seq_view<T_tau_ref> tau_vec(tau_ref);
+  size_t N = max_size(mu, sigma, tau);
+  VectorBuilder<true, double, T_loc, T_scale, T_skewness> output(N);
+
+  variate_generator<RNG&, uniform_real_distribution<> > z_rng(
+      rng, uniform_real_distribution<>(0.0, 1.0));
+  for (size_t n = 0; n < N; ++n) {
+    double z = z_rng();
+    if (z <= 0.5) {
```
bbbales2, line 65: It looks like we're generating random covariates using an inverse CDF trick. I think this should be `z < tau`. The probability mass of stuff to the left of `y == mu` is `tau`.

test/unit/math/prim/prob/skew_double_exponential_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+ */
+TEST(ProbDistributionsSkewedDoubleExponential, chiSquareGoodnessFitTest) {
+  boost::random::mt19937 rng;
+  int N = 10000;
+  int K = stan::math::round(2 * std::pow(N, 0.4));
+
+  std::vector<double> samples;
+  for (int i = 0; i < N; ++i) {
+    samples.push_back(
+        stan::math::skew_double_exponential_rng(2.0, 1.0, 0.5, rng));
```
bbbales2, line 52: We should test this with something slightly skew.

I think the way to generate reference random numbers is to do it by a mixture of two exponential distributions, like one positive, and one negative, if that makes sense.

I'm not sure why we generate the reference quantiles with rngs? We have the quantiles all here analytically. Maybe it's something to do with the chi-square test? Anyway, if we write two different ways to generate skew double exponential and they math, then we're probably good.

If you push something up I'll happily print out the quantiles and check them against our CDF function as well (I'm pretty sure that's right)

bbbales2 (changes requested):
Looks good! Thanks for adding the rng test. I left a comment on there. I think we need to test with `tau != 0.5` cause there might be a problem right now.

stan/math/prim/prob/skew_double_exponential_rng.hpp
dirmeier: Thanks, nice catch! Don't know what I thought there...

test/unit/math/prim/prob/skew_double_exponential_test.cpp
dirmeier: Sorry @bbbales2, I'm not entirely following: 

> I think the way to generate reference random numbers is to do it by a mixture of two exponential distributions, like one positive, and one negative, if that makes sense.

How, would the two be parametrized? And this would go into the goodness of fit test?

> I'm not sure why we generate the reference quantiles with rngs? 

To be honest, I just copy and pasted it from the `skew_normal_test`. Could you give me a hint, what I should be replacing exactly? Sorry for the trouble!

Cheers,
Simon

test/unit/math/prim/prob/skew_double_exponential_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  // Generate quantiles from boost's double exponential distribution
+  boost::math::laplace_distribution<> dist(2.0, 1.0);
+  std::vector<double> quantiles;
+  for (int i = 1; i < K; ++i) {
+    double frac = static_cast<double>(i) / K;
+    quantiles.push_back(quantile(dist, frac));
```
bbbales2, line 60: Oh dang, we are depending on Boost to compute quantiles. Okay so the quantiles are going to be inverse CDFs. Since the inverse CDF is easy, just code up the quantiles here manually.

The gotcha there is since we're using inverse CDF to do our sampling, if we had a bug in the inverse CDF here and in the rngs (we're writing them both and presumably copy-pasting), we still need to check the inverse CDFs here.

However, since we have a CDF function we trust, it's easy to check that the inverse CDF we write here is working.

We just do some variation of (in psuedo-code):

```
quantiles = c(0.05, 0.45, ...)
x = inverse_cdf(p);
EXPECT_FLOAT_EQ(cdf(x), p);
```

Does this make sense? We want to generated quantiles of the skew double exponential and then check that empirically our rng generates numbers from the right distribution (by computing how many samples land in these quantiles and doing a test to see if it's what we'd expect).

test/unit/math/prim/prob/skew_double_exponential_test.cpp
bbbales2: Added another comment below! Ignore the mixture of exponentials thing.

bbbales2 (commented):
More clarification on RNG test. I was reading it wrong before.

dirmeier:
Ciao @bbbales2 ,
I updated the tests: one to test the quantiles of the rng function, one to test the inverse cdf. Hope, this is ok.
Cheers,
S

bbbales2:
> I updated the tests: one to test the quantiles of the rng function

Cool will get this soon.

Is Git spitting our errors when you try to regular push? Force pushes are bad because if we both make changes then it can be a mess to get the repos to fit back together. I don't have any local changes here so it doesn't matter so much, but force pushes put everyone else's branches out of sync.

Something to be aware of is whenever you put code in a Stan pull there's an automatic code formatter that comes through behind and cleans up the code a bit. Whenever you work on the code next, you'll need to pull these changes down before you go to push again.

dirmeier:
> Something to be aware of is whenever you put code in a Stan pull there's an automatic code formatter that comes through behind and cleans up the code a bit. Whenever you work on the code next, you'll need to pull these changes down before you go to push again.

Alright, I wanted to avoid the commits added by the formatter and force pushed to override them. I'll keep them in the future.

bbbales2:
Aah, yeah it's weird and sometimes annoying to merge the changes when you forget to pull before editing, but it avoids having to worry about spacing and stuff when reviewing code. Also it ups our Github stats :D.

bbbales2:
@dirmeier I swapped out the icdf function to compute the quantiles. Everything looks good to me so will merge when this goes green if it looks good to you.

dirmeier:
Ciao @bbbales2,

>  I swapped out the icdf function to compute the quantiles. 

yes, that makes a lot more sense. Cool!

> Everything looks good to me so will merge when this goes green if it looks good to you.

Cool, definitely.

bbbales2 (approved):

PR #2280, opened 2020-12-31
Three argument tests for varmat (and varmat implementation of `trace_gen_quad_form`)

bbbales2:
## Summary

This adds three argument tests for `expect_ad_matvar`. It also adds some tests for the matvar tests, breaks them out from the regular `expect_ad` code, and changes the docs a bit.

There is an implementation for `trace_gen_quad_form` as well.

I tried to do the doxygen using doxygen Member Groups (look down the page [here](https://www.doxygen.nl/manual/grouping.html) -- example [here](https://www.doxygen.nl/manual/examples/memgrp/html/class_memgrp___test.html)) cause there are big groups of functions that are related (and can be kinda doc'ed together).

## Tests

Tests in:
```
test/unit/math/test_ad_matvar_test.cpp
test/unit/math/mix/fun/trace_gen_quad_form_test.cpp
```

## Release notes

- Added varmat implementation of `trace_gen_quad_form` and three argument varmat test code.

## Checklist

- [x] Math issue #2101


test/unit/math/mix/fun/abs_test.cpp
```diff
@@ -49,7 +49,7 @@ TEST(mathMixMatFun, abs_varmat) {
     return abs(x1);
   };
   std::vector<double> com_args = common_nonzero_args();
-  std::vector<double> args{0, -3, -2, 2, -17.3, -0.68, 2, 4};
+  std::vector<double> args{-3, 2, -0.68, 1};
```
bbbales2, line 52: Derivative of abs doesn't exist at zero

bbbales2:
Ready for review!

stan/math/rev/fun/trace_gen_quad_form.hpp
```diff
@@ -114,6 +114,179 @@ inline var trace_gen_quad_form(const Td& D, const Ta& A, const Tb& B) {
...
+
+    var res = (arena_D.val_op() * arena_B.val_op().transpose()
+               * arena_A.val_op() * arena_B.val_op())
+                  .trace();
+
+    reverse_pass_callback([arena_A, arena_B, arena_D, res]() mutable {
+      double C_adj = res.adj();
+
+      auto BDT = (arena_B.val_op() * arena_D.val_op().transpose()).eval();
+      auto AB = (arena_A.val_op() * arena_B.val_op()).eval();
```
SteveBronder, line 159: [optional] You compute `A.val() * B.val()` in the forward pass you could just store it in the forward pass and reuse it

```suggestion
   arena_t<decltype(arena_A.val_op() * arena_B.val_op())> ab_val_mul = arena_A.val_op() * arena_B.val_op()
    var res = (arena_D.val_op() * arena_B.val_op().transpose() * ab_val_mul).trace();

    reverse_pass_callback([arena_A, arena_B, arena_D, res]() mutable {
      double C_adj = res.adj();

      auto BDT = (arena_B.val_op() * arena_D.val_op().transpose()).eval();
```

SteveBronder: Actually is it true if `B` is a square matrix that `D * B.transpose() = (B * D.transpose()).transpose()`? If so I think you can reuse that as well down here

https://godbolt.org/z/5ohjK5

SteveBronder: If so then you can use the above in the other specializations as well

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+void expect_near_rel_matvar(const std::string& message,
+                            const std::tuple<T1...>& x,
+                            const std::tuple<T2...>& y,
+                            const ad_tolerances& tols,
+                            std::index_sequence<I...> i) {
+  std::vector<int> A(
+      {(expect_near_rel_matvar(
+            message + std::string(", argument ") + std::to_string(I + 1),
+            std::get<I>(x), std::get<I>(y), tols),
+        0)...});
```
SteveBronder, line 110: I'd use something like the below instead of storing them in a vector

https://gist.github.com/SteveBronder/410f7127f8be9dced2b1915765fd8644#file-for_each-hpp-L19

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+                                 const std::tuple<VarMatArgs...>& A_vm_tuple) {
+  expect_near_rel("var<Matrix> vs Matrix<var> values", value_of(A_vm_f),
+                  value_of(A_mv_f), tols.gradient_val_);
+  A_vm_f.adj() = 1;
+  A_mv_f.adj() = 1;
+  stan::math::grad();
+  expect_near_rel_matvar("var<Matrix> vs Matrix<var>", A_vm_tuple, A_mv_tuple,
+                         tols);
+  stan::math::set_zero_all_adjoints();
+}
```
SteveBronder, line 174: Should the type names just be `ResultVar1` and `ResultVar1`? Think it's just easier to read that way

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+          require_std_vector_vt<is_var, ResultVarMat>* = nullptr>
+inline void test_matvar_gradient(const ad_tolerances& tols,
+                                 ResultMatVar& A_mv_f, ResultVarMat& A_vm_f,
+                                 const std::tuple<MatVarArgs...>& A_mv_tuple,
+                                 const std::tuple<VarMatArgs...>& A_vm_tuple) {
+  expect_near_rel("var<Matrix> vs Matrix<var> values", value_of(A_vm_f),
+                  value_of(A_mv_f), tols.gradient_val_);
+  for (size_t i = 0; i < A_vm_f.size(); ++i) {
+    A_vm_f[i].adj() = 1;
+    A_mv_f[i].adj() = 1;
```
SteveBronder, line 191: Could we change the names of these from `A_vm_f` to `A_vm_ret`? Just so it's more clear this is the return value

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+    stan::math::grad();
+    expect_near_rel_matvar("var<Matrix> vs Matrix<var>", A_vm_f[i], A_mv_f[i],
+                           tols);
```
SteveBronder, line 194: Would it be safer to check all of the values here? Just to make sure we are indexing correctly everywhere etc

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+          typename... VarMatArgs,  // matrix_dynamic
+          require_eigen_st<is_var, ResultMatVar>* = nullptr>
+inline void test_matvar_gradient(const ad_tolerances& tols,
+                                 ResultMatVar& A_mv_f, ResultVarMat& A_vm_f,
+                                 const std::tuple<MatVarArgs...>& A_mv_tuple,
+                                 const std::tuple<VarMatArgs...>& A_vm_tuple) {
+  expect_near_rel("var<Matrix> vs Matrix<var> values", value_of(A_vm_f),
+                  value_of(A_mv_f), tols.gradient_val_);
+  for (Eigen::Index i = 0; i < A_mv_f.rows(); ++i) {
+    for (Eigen::Index j = 0; j < A_mv_f.cols(); ++j) {
```
SteveBronder, line 214: [optional] column major order

```suggestion
  for (Eigen::Index j = 0; j < A_mv_f.cols(); ++j) {
    for (Eigen::Index i = 0; i < A_mv_f.rows(); ++i) {
```

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+ */
+template <typename ResultMatVar, typename ResultVarMat, typename... MatVarArgs,
+          typename... VarMatArgs,  // matrix_dynamic
+          require_st_arithmetic<ResultMatVar>* = nullptr>
+inline void test_matvar_gradient(const ad_tolerances& tols,
+                                 ResultMatVar& A_mv_f, ResultVarMat& A_vm_f,
+                                 const std::tuple<MatVarArgs...>& A_mv_tuple,
+                                 const std::tuple<VarMatArgs...>& A_vm_tuple) {
+  FAIL() << "test_matvar_gradient should only be used when the return is a "
+            "reverse mode autodiff type";
```
SteveBronder, line 236: I think you can just have this be a static_assert()

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+      && !(is_eigen<ReturnType>::value
+           || std::is_same<ReturnType, var_value<double>>::value)) {
+    FAIL() << "All the arguments are `Eigen::Matrix<T, R, C>` types, but the "
+              "return type is neither an `Eigen::Matrix<T, R2, C2>` or a `var`";
+  } else if (stan::math::disjunction<is_var_matrix<Types>...>::value
+             && !(is_var_matrix<ReturnType>::value
+                  || std::is_same<ReturnType, var_value<double>>::value)) {
+    FAIL() << "One of the arguments is a `var_value<Eigen::Matrix<double, R, "
+              "C>>`, but the return type is not a `var_value<T>`";
+  }
```
SteveBronder, line 299: Will this work with vectors of mat var types?

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+void expect_ad_matvar_v(const ad_tolerances& tols, const F& f,
+                        const std::vector<EigMat>& x) {
+  using stan::plain_type_t;
+  using stan::math::promote_scalar_t;
+  using stan::math::var;
+  using stan::math::var_value;
+  using stan::math::test::type_name;
+  using vec_mat_var = std::vector<promote_scalar_t<var, EigMat>>;
+  using vec_var_mat = std::vector<var_value<plain_type_t<EigMat>>>;
+  vec_mat_var A_mv;
```
SteveBronder, line 348: [optional]
```suggestion
  vec_mat_var A_vec_mv;
```

etc

test/unit/math/test_ad_matvar.hpp
```diff
@@ -0,0 +1,724 @@
...
+ * @param y Value of second argument
+ */
+template <typename F, typename EigMat1, typename EigMat2>
+void expect_ad_matvar(const ad_tolerances& tols, const F& f, const EigMat1& x,
+                      const EigMat2& y) {
+  using stan::math::var;
+  using varmat = stan::math::var_value<Eigen::MatrixXd>;
+
+  // expect_ad_matvar_impl<double, double>(tols, f, x, y);
+  // expect_ad_matvar_impl<double, var>(tols, f, x, y);
```
SteveBronder, line 571: ?

test/unit/math/test_ad_matvar_test.cpp
```diff
@@ -0,0 +1,371 @@
...
+template <typename T>
+auto one_arg_bad_vals(const T& x) {
+  return x;
+}
+
+template <typename T>
+auto one_arg_bad_vals(const stan::math::var_value<T>& x) {
+  return stan::math::make_callback_var(
+      x.val() * 0, [x](const auto& vi) mutable { x.adj() += vi.adj(); });
+}
```
SteveBronder, line 17: I'm confused about what this is testing?

SteveBronder (changes requested):
Nice! Few Qs about some optimizations and some testing stuff

stan/math/rev/fun/trace_gen_quad_form.hpp
bbbales2: Woof, good catch, yes can rework this.

stan/math/rev/core/callback_vari.hpp
```diff
@@ -12,7 +12,9 @@ template <typename T, typename F>
 struct callback_vari : public vari_value<T> {
   F rev_functor_;
 
-  explicit callback_vari(T&& value, F&& rev_functor)
+  template <typename S,
+            require_same_t<plain_type_t<T>, plain_type_t<S>>* = nullptr>
+  explicit callback_vari(S&& value, F&& rev_functor)
```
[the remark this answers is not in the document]
bbbales2: I shoulda had a test I guess. I'll roll this back and see where it breaks. I think it was something like `T` dropped a const (which is fine cause that's the local type), but the input type was const (but that's fine too cause it'll just copy).

test/unit/math/test_ad_matvar.hpp
bbbales2: Just helpers so I could see which combinations I wasn't actually checking. I'll remove em'.

test/unit/math/test_ad_matvar_test.cpp
bbbales2: This is testing that functions with badly implemented autodiff fail when they should fail.

It's easy to accidentally make a bug in these tests so that everything passes :D.

For instance the first time I coded this I wasn't checking that values with varmat matched matvar.

test/unit/math/test_ad_matvar.hpp
bbbales2: The output values are getting checked out of the loop, and all of them are getting checked I think. Moving the checks in the loop just assumes the inputs and outputs are the same lengths (which they probably are but not necessary)

This line checks the input values and adjoints.

test/unit/math/test_ad_matvar.hpp
bbbales2: Hmm I like keeping the VM/MV designation. `ResultVM` and `ResultMV`?

bbbales2: It really is hard to read though I agree.

stan/math/rev/core/callback_vari.hpp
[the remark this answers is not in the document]
bbbales2: I added a test [here](https://github.com/stan-dev/math/pull/2280/files#diff-61b65dcd6af86c6600430f4291f1369130e69865525488a246282e48f06983d9R23). It was passing something that was a `const` into `make_callback_vari`. I think it should work.

We can remove the `require_same_t` thing here. I just added that cause when we added `template <typename S>` [here](https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/chainable_object.hpp#L39) we did it.

stan/math/rev/fun/trace_gen_quad_form.hpp
bbbales2: I updated this to use the AB/BDT temporaries in the forward pass.

test/unit/math/test_ad_matvar.hpp
bbbales2: No. Everything that uses `std::vector` gets shoveled through [this check](https://github.com/stan-dev/math/pull/2280/files/70766a8c81fb652478c8c09af4e42c94db4d89b1#diff-7dc2c9d1d99265ed85e16bd88a0541f7b79e39ea27a68bc27a83578be76e7b26R315)

Our `std::vector` support is limited. I think next step here would be to bring the `std::vector` support in, but I didn't do it cuz we've been building this testing in pieces (only what we need).

test/unit/math/test_ad_matvar.hpp
bbbales2: Resolved, but I snuck and added the `index` as an argument to pass through with `for_each`. This way I don't have to worry about index sequences or anything it's just all handled by `for_each`. It creates an extra argument tho.

bbbales2:
@SteveBronder this should be ready for check again (once tests pass)

SteveBronder (approved):
looks good!

PR #2285, opened 2021-01-02
Adds var<mat> specializations for lower/upper bounds

SteveBronder:
## Summary

This adds specializations for lower / upper / lower_upper constraints to work with var<matrix>. I had to do a few other things for this to work.

1. To test this I had to write a specialization for `[lb/up/lub]_constrain` that worked for `Matrix<var>` and `var_value<Eigen::Matrix>`. That means that a few things like `check_less` etc have to now take / iterate over general matrices.

2. I had to fix `elt_multiply()` to allow for multiplying a scalar and matrix

3. I'm not sure how to handle the negative and positive infinity checks for the matrix versions. Previously this was checked on a per coefficient basis. But now if you have a vector representing the bounds for a vector, if any of them are +/- infinity then it returns back the input vector without the transformation. I feel like I'm not understanding how this is used. If a vector represents the bound then will it ever be expected that some of those values in the vector would be infinite and some would not be? If that's the case then I need to write specific var specializations which isn't that bad but it would be nice to not have to.

## Tests

Tests added for matrix versions. It looks like `lub_constrain()` has not tests? So I added ones for testing those as well.

```bash
./runTests.py -j2 ./test/unit/math/mix/fun/lub_constrain_test.cpp  ./test/unit/math/mix/fun/lb_constrain_test.cpp  ./test/unit/math/mix/fun/ub_constrain_test.cpp 
```

## Side Effects

Yes see (3) above

## Release notes

Adds bound constraint functions for `var<matrix>` functions

## Checklist

- [x] Math issue #1805 


bbbales2:
> I had to fix elt_multiply() to allow for multiplying a scalar and matrix

Nice. This will make a placeholder `fma` easy to write if we still need that (which I have been avoiding).

> If a vector represents the bound then will it ever be expected that some of those values in the vector would be infinite and some would not be?

I think that makes sense. Some variables constrained and some not. Will have to check the code to make sure I'm reading that right tho.

stan/math/prim/err/is_infinity.hpp
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Return <code>true</code> if <code>y</code> is infinite.
+ * This function is vectorized and will check each element of
+ * <code>y</code>.
+ * @tparam T_y Type of y
+ * @param y Variable to check
+ */
+template <typename T_y, require_matrix_t<T_y>* = nullptr>
+inline bool is_infinity(const T_y& y) {
+  return (y.array() == INFTY).any();
```
bbbales2, line 23: Call it `is_positive_infinity` (and update docs to reflect this) since usually infinity checks include negative and positive.

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -29,15 +32,42 @@ namespace math {
  * @param[in] lb lower bound on constrained output
  * @return lower bound constrained value corresponding to inputs
  */
-template <typename T, typename L>
+template <typename T, typename L, require_stan_scalar_t<T>* = nullptr>
 inline return_type_t<T, L> lb_constrain(const T& x, const L& lb) {
   using std::exp;
-  if (lb == NEGATIVE_INFTY) {
+  if (unlikely(is_negative_infinity(value_of(lb)))) {
```
bbbales2, line 38: Yeah this looks wrong cause it's saying if any element has a lower bound of negative infinity, bail on them all.

stan/math/rev/core/callback_vari.hpp
```diff
@@ -2,6 +2,7 @@
 #define STAN_MATH_REV_CORE_CALLBACK_VARI_HPP
 
 #include <stan/math/rev/core/vari.hpp>
+#include <stan/math/rev/core/callback_vari.hpp>
```
bbbales2, line 5: Including itself?

stan/math/rev/core/operator_unary_negative.hpp
```diff
@@ -48,7 +49,30 @@ class neg_vari final : public op_v_vari {
  * @param a Argument variable.
  * @return Negation of variable.
  */
-inline var operator-(const var& a) { return {new internal::neg_vari(a.vi_)}; }
+inline var operator-(const var& a) {
+  return make_callback_var(-a.val(), [a](const auto vi) {
+    if (unlikely(is_nan(a.val()))) {
+      a.adj() = NOT_A_NUMBER;
```
bbbales2, line 55: Did this cause a problem somewhere? I'm not a fan of these manual NaN injections. If `a` is NaN and it gets used anywhere, the rest of the calculation will probably blow up. I don't think we need to spend the `if` statements making sure that happens.

bbbales2 (changes requested):
Had a closer look. You're right that the ifs are handling infinities incorrectly.

I think we'll just have to write autodiff for each of these things manually.

Since these constraints are only eigen types and this is only a few functions, this shouldn't be too hard to do. I can do it tomorrow if you've got other stuff to keep you busy. If we do this we can also ditch the new infinity checks unless you want to keep them.

bbbales2:
@SteveBronder I removed the infinite checks and cleaned this up.

I thought a little more about #2291 -- this already wasn't working right for vector constraints so I'm not gonna worry about supporting this in the future.

The only case this worked was if someone had a scalar constraint that was a infinite, and so I added failing error checks for that. It is hopefully not more expensive than the infinite checks that were already there.

Should be ready to review if tests pass.

bbbales2:
Something else, but once we have the `as_array_or_scalar` stuff for varmat we can change this code from `add(x, y)` to `x + y` and make it look nicer. But that can wait if this goes in before that.

bbbales2:
I think I messed up check_less for OpenCL (error [here](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2285/18/pipeline) for future reference). Added some stuff back.

bbbales2:
@SteveBronder I broke OpenCL with the check_less changes [here](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2285/20/pipeline). You know how to fix this?

stan/math/rev/fun/elt_multiply.hpp
```diff
@@ -57,7 +58,46 @@ auto elt_multiply(const Mat1& m1, const Mat2& m2) {
...
+ * @param B matrix
+ * @return product of matrix and scalar
+ */
+template <typename T1, typename T2, require_not_matrix_t<T1>* = nullptr,
+          require_matrix_t<T2>* = nullptr,
+          require_any_st_var<T1, T2>* = nullptr,
+          require_not_row_and_col_vector_t<T1, T2>* = nullptr>
+inline auto elt_multiply(const T1& A, const T2& B) {
+  return multiply(A, B);
+}
```
SteveBronder, line 80: Should the prim impl of these just divert to multiply so we have consistent behavior everywhere?

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -20,22 +23,17 @@ namespace math {
...
-inline return_type_t<T, L> lb_constrain(const T& x, const L& lb) {
-  using std::exp;
-  if (lb == NEGATIVE_INFTY) {
-    return identity_constrain(x);
-  }
-  return exp(x) + lb;
+inline auto lb_constrain(const T& x, const L& lb) {
+  const auto& lb_ref = to_ref(lb);
+  check_finite("lb_constrain", "lb", value_of(lb_ref));
+  return eval(add(exp(x), lb));
```
SteveBronder, line 36: We don't need the eval here now

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -44,29 +42,24 @@ inline return_type_t<T, L> lb_constrain(const T& x, const L& lb) {
...
-  }
-  lp += x;
-  return exp(x) + lb;
+template <typename T, typename L>
+inline auto lb_constrain(const T& x, const L& lb, return_type_t<T, L>& lp) {
+  const auto& x_ref = to_ref(x);
+  const auto& lb_ref = to_ref(lb);
+  check_finite("lb_constrain", "lb", value_of(lb_ref));
+  lp += sum(x_ref);
+  return eval(add(exp(x_ref), lb));
```
SteveBronder, line 59: Just remove all the eval's in these functions

stan/math/prim/fun/lub_constrain.hpp
```diff
@@ -97,27 +87,20 @@ inline return_type_t<T, L, U> lub_constrain(const T& x, const L& lb,
...
+  const auto& lb_ref = to_ref(lb);
+  const auto& ub_ref = to_ref(ub);
+
+  check_less("lub_constrain", "lb", value_of(lb_ref), value_of(ub_ref));
+  check_finite("lub_constrain", "lb", value_of(lb_ref));
+  check_finite("lub_constrain", "ub", value_of(ub_ref));
+  lp += sum(add(log(subtract(ub_ref, lb_ref)),
+                subtract(x_ref, multiply(2, log1p_exp(x_ref)))));
+  return eval(
+      add(elt_multiply(subtract(ub_ref, lb_ref), inv_logit(x_ref)), lb_ref));
```
SteveBronder, line 103: [optional] We could pull out `subtract(ub_ref, lb_ref)` here so we only do it once

stan/math/rev/core/operator_unary_negative.hpp
SteveBronder: We just did it in the old impl so I think we should do it in this impl as well. Unless we don't check when values are NaNs. It might be a discussion to have at a later time

stan/math/rev/fun/simplex_constrain.hpp
```diff
@@ -30,42 +32,107 @@ inline auto simplex_constrain(const T& y) {
...
-            + (1 - arena_z.coeff(n + 1)) * acc;
-      arena_y.adj().coeffRef(n)
-          += arena_diag.coeff(n) * (arena_x.adj().coeff(n) - acc);
+  reverse_pass_callback([arena_y, arena_x, arena_z, lp]() mutable {
+    int N = arena_y.size();
+    // x_val.coeffRef(N) = stick_len;
+    double stick_len_val = arena_x.val().coeff(N);
+    double stick_len_adj = arena_x.adj().coeff(N);
+    for (Eigen::Index k = N; k-- > 0;) {
+      // stick_len -= x_val(k);
```
SteveBronder, line 118: I think I see what's going on here so we can delete these comments

SteveBronder:
Interesting, this is failing at the stan level 

```
[----------] 5 tests from ModelUtil
[ RUN      ] ModelUtil.finite_diff_grad__false_false
[       OK ] ModelUtil.finite_diff_grad__false_false (0 ms)
[ RUN      ] ModelUtil.finite_diff_grad__false_true
src/test/unit/model/finite_diff_grad_test.cpp:43: Failure
  expected_gradient
    Which is: -0
  gradient[0]
    Which is: -1.110223e-10
[  FAILED  ] ModelUtil.finite_diff_grad__false_true (0 ms)
[ RUN      ] ModelUtil.finite_diff_grad__true_false
[       OK ] ModelUtil.finite_diff_grad__true_false (0 ms)
[ RUN      ] ModelUtil.finite_diff_grad__true_true
src/test/unit/model/finite_diff_grad_test.cpp:84: Failure
  expected_gradient
    Which is: -0
  gradient[0]
    Which is: -1.110223e-10
```

https://github.com/stan-dev/stan/blob/develop/src/test/unit/model/finite_diff_grad_test.cpp#L25

SteveBronder:
Test model here

https://github.com/stan-dev/stan/blob/540105b9b43303f2fe70cb8f438f5a2b96d0a442/src/test/unit/model/test_model.hpp#L7

stan/math/prim/err/check_less.hpp
```diff
@@ -7,72 +7,145 @@
...
+ * @param y Variable to check
+ * @param high Upper bound
+ * @throw <code>domain_error</code> if y is not less than low
+ *   or if any element of y or high is NaN.
+ */
+template <typename T_y, typename T_high, require_container_t<T_y>* = nullptr,
+          require_stan_scalar_t<T_high>* = nullptr>
+inline void check_less(const char* function, const char* name, const T_y& y,
+                       const T_high& high) {
+  const auto& y_ref = as_array_or_scalar(to_ref(y));
```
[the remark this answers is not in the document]
bbbales2: I agree. Removed.

stan/math/prim/fun/lb_constrain.hpp
bbbales2: Honestly not sure. `to_ref` might allocate memory, so if this were to be assigned to a non-heavy type, then it would need a holder to work. I will ditch the evals though to see if tests pass. @t4c1 can you advise on this?

bbbales2:
Weird, thanks for the heads up.

stan/math/prim/fun/lb_constrain.hpp
bbbales2: I took the evals out and got (deleting a bunch of output to make the picture clearer):

```
==4440==ERROR: AddressSanitizer: heap-use-after-free on address 0x603000ad0b90 at pc 0x000000992c36 bp 
...
 /home/bbales2/math-benchmarks/math/./test/unit/math/test_ad.hpp:86:22

0x603000ad0b90 is located 0 bytes inside of 32-byte region [0x603000ad0b90,0x603000ad0bb0)
freed by thread T0 here:
...
    #2  /home/bbales2/math-benchmarks/math/./stan/math/prim/fun/lub_constrain.hpp:106:1

previously allocated by thread T0 here:
...
    #5 /home/bbales2/math-benchmarks/math/./stan/math/prim/fun/to_ref.hpp:18:10
    #6  /home/bbales2/math-benchmarks/math/./stan/math/prim/fun/lub_constrain.hpp:101:22
```

That `to_ref` is the one that was creating the intermediate for `subtract`. It's possible any of the `to_ref` s could allocate memory here if they are expressions.

stan/math/prim/fun/lb_constrain.hpp
bbbales2: Follow up [here](https://github.com/stan-dev/math/pull/2285#discussion_r553677134)

stan/math/prim/fun/lub_constrain.hpp
```diff
@@ -97,27 +84,22 @@ inline return_type_t<T, L, U> lub_constrain(const T& x, const L& lb,
...
+  const auto& ub_ref = to_ref(ub);
+
+  check_less("lub_constrain", "lb", value_of(lb_ref), value_of(ub_ref));
+  check_finite("lub_constrain", "lb", value_of(lb_ref));
+  check_finite("lub_constrain", "ub", value_of(ub_ref));
+
+  const auto& diff = to_ref(subtract(ub_ref, lb_ref));
+  
+  lp += sum(add(log(diff),
+                subtract(-abs(x_ref), multiply(2, log1p_exp(-abs(x_ref))))));
```
bbbales2, line 101: I added `abs` calls here to symmetrize the function around zero so that the finite differences would give zero. I kinda don't like it cause it depends on `abs` returning non-nan gradients at zero. I'd honestly be happer just upping the tolerances in that Stan tests.

The other option is to write out the custom autodiff here. It's just annoying to vectorize that conditional (look at the code we're replacing to see what is happening -- `x - 2 * log1p(exp(x))` is equal to `-x - 2 * log1p(exp(-x))`).

bbbales2:
@SteveBronder this'll be ready to review once we resolve the eval thing in here.

stan/math/prim/fun/lb_constrain.hpp
t4c1: `lb_ref ` is a local variable referenced in returned expression, so it needs a holder.

```
return make_holder([&x](auto& lb_ref){
  check_finite("lb_constrain", "lb", value_of(lb_ref));
  return exp(x) + lb_ref;
}, to_ref(std::forward<L>(lb)));
```

Also change `const L& lb` to `L&& lb`.

bbbales2: I get segfaults with this:

```cpp
template <typename T, typename L, typename U>
inline promote_scalar_type_t<return_type_t<T, L, U>, T>
lub_constrain(const T& x, const L& lb, const U& ub,
              return_type_t<T, L, U>& lp) {
  const auto& x_ref = to_ref(x);
  const auto& lb_ref = to_ref(lb);
  const auto& ub_ref = to_ref(ub);

  check_less("lub_constrain", "lb", value_of(lb_ref), value_of(ub_ref));
  check_finite("lub_constrain", "lb", value_of(lb_ref));
  check_finite("lub_constrain", "ub", value_of(ub_ref));

  const auto& diff = to_ref(subtract(ub_ref, lb_ref));

  lp += sum(add(log(diff),
                subtract(-abs(x_ref), multiply(2, log1p_exp(-abs(x_ref))))));

  return make_holder([](const auto& diff, const auto& x_ref, const auto& lb_ref) {
      return add(elt_multiply(diff, inv_logit(x_ref)), lb_ref);
    }, diff, x_ref, lb_ref);
}
```

It seems to be associated with the temporary `const auto& diff = to_ref(subtract(ub_ref, lb_ref));`.

I'm just going to code these to use plain types. I don't understand the assumptions around holders and don't feel safe using them myself and I don't think where these functions are used expressions matter.

bbbales2: Alright I ended up going with evals to get the plain types. I think it's correct. eval means, if there are expressions, make them plain types.

`promote_scalar_t` doesn't work right with varmat (you get matvars from varmats). `return_var_matrix_t` doesn't work right with non-reverse mode stuff.

bbbales2:
This will need this pull to go through (at least): https://github.com/stan-dev/stan/pull/2995

Not sure if there are more upstream bounds checks.

stan/math/prim/fun/lb_constrain.hpp
t4c1: That segfaults because you are passing lvalues into `make_holder`, so it thinks they won't go out of scope and does not construct a holder. Instead you should `std::move` these variables into `make_holder`.

bbbales2: This was the issue. They have been successfully holder-ified.

bbbales2 (commented):
Looks the changes to `check_less`, `elt_multiply`, and `operator_unary_negative` are not tested. I'll try to add these later today (will leave a note if I don't make it)

bbbales2:
Tests for `check_less`, `elt_multiply`, and `operator_unary_negative` are here now. I'm happy with this so see what you think @SteveBronder

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -20,22 +22,19 @@ namespace math {
...
-  if (lb == NEGATIVE_INFTY) {
-    return identity_constrain(x);
-  }
-  return exp(x) + lb;
+inline auto lb_constrain(const T& x, const L& lb) {
+  const auto& lb_ref = to_ref(lb);
+  check_finite("lb_constrain", "lb", value_of(lb_ref));
+
+  return make_holder([&x](const auto& lb_ref) { return add(exp(x), lb_ref); },
+                     std::move(lb_ref));
```
SteveBronder, line 37: idt we should be moving `lb_ref` here

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -44,29 +43,27 @@ inline return_type_t<T, L> lb_constrain(const T& x, const L& lb) {
...
+template <typename T, typename L>
+inline auto lb_constrain(const T& x, const L& lb, return_type_t<T, L>& lp) {
+  const auto& x_ref = to_ref(x);
+  const auto& lb_ref = to_ref(lb);
+  check_finite("lb_constrain", "lb", value_of(lb_ref));
+  lp += sum(x_ref);
+
+  return make_holder([](const auto& x_ref,
+                        const auto& lb_ref) { return add(exp(x_ref), lb_ref); },
+                     std::move(x_ref), std::move(lb_ref));
```
SteveBronder, line 63: I'd get rid of all these std moves. They are also moving from something that is supposed to be a const ref

stan/math/prim/fun/lub_constrain.hpp
SteveBronder: Hmm, I think I'm okay with upping the tolerances since this is a finite diff issue more than an autodiff issue

SteveBronder:
@bbbales2 I think this is good to go?

stan/math/prim/fun/lub_constrain.hpp
```diff
@@ -107,8 +106,7 @@ inline auto lub_constrain(const T& x, const L& lb, const U& ub,
   return make_holder(
       [](const auto& diff, const auto& x_ref, const auto& lb_ref) {
         return add(elt_multiply(diff, inv_logit(x_ref)), lb_ref);
-      },
-      std::move(diff), std::move(x_ref), std::move(lb_ref));
+      }, diff, x_ref, lb_ref);
```
bbbales2: Need the move for the diff

bbbales2:
Oh wait. THe moves on all these references need to be there. The reference might allocate memory if the input is an expression.

That make sense?

bbbales2:
@SteveBronder yeah I'm happy with it once these moves or not moves get sorted.

SteveBronder:
lol whups yeah lemme look here

SteveBronder:
Just added some stuff that went a little loco with the moves. But if we need to move things sometimes and we need to use `to_ref` then I think we need to detect when `to_ref` gives us back an rvalue that needs to be moved

SteveBronder:
The other issue is we can't do stuff like

```cpp
  return make_holder([&x](const auto& lb_ref) { return add(exp(x), lb_ref); },
                     lb_ref);
```

Since `x` could fall out of scope we need to pass `x` in as an argument to `make_holder`

```cpp
  return make_holder([](auto& lb_ref, const auto& xx) { return add(exp(xx), lb_ref); },
                     lb_ref, x);
```

bbbales2:
I don't think we need to forward through `to_ref`. If the `to_ref` doesn't need to eval, it's cheap.

> Since x could fall out of scope

My assumption is that input arguments don't need held, but the expression semantics aren't clear to me (when we need holder and when we don't).

bbbales2:
As an example of not holding inputs: https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/transpose.hpp

bbbales2:
I think this and https://github.com/stan-dev/stan/pull/2995 are ready to go. @SteveBronder you good with this? If so @rok-cesnovar can do the mergin'.

SteveBronder:
Yep!

rok-cesnovar:
Oh, this two PRs need to go in in-sync?

rok-cesnovar:
Restarted with proper branches, please approve both and I can handle merging this once it passes.

SteveBronder:
> Oh, this two PRs need to go in in-sync?

Yep!

Thanks will do!

SteveBronder:
Nice! @bbbales2 if you hit the approve button I think we are good to go!

@rok-cesnovar I think this was the last piece we needed for the stance impl! First thing we need to do is use the matrix specializations for the constraints then we can add varmat!

bbbales2 (approved):

PR #2290, opened 2021-01-05
Benchmark improvements

t4c1:
## Summary

Contains various small improvements for automatic benchmarking:
- Fixed wrong flag given to benchmark that would cause it to only output aggregate (mean, medain, stdev) values when running multiple repeats
- exclude aggregate values when making a plot
- added option to benchmark OpenCL implementation including only the time for copying var arguments, and not arguments of non-autodiff types (the way OpenCL implementations are often used by Stan)
- Make plots of individual repeats not affect scaling of axes. This way individual extreme measurements can not cause the more relevant plots to be only in a small part of the whole graph.

## Tests
None.

## Side Effects
None.

## Checklist


SteveBronder:
random fr, would it be possible to have a flag like `--compare [branch]` that would run the tests for both the current branch and another specified branch?

ahartikainen:
cc @bbbales2 for compare stuff

bbbales2:
It would be handy cause that's the most common comparison we'd be making in Math. I'm not sure how to do it nicely.

Was working on some benchmarking stuff with @ahartikainen to handle comparisons between different branches of cmdstan/stan/math etc, but if it's possible to do a quick comparison here that's great.

SteveBronder:
What we could do is rewrite perf math to essentially just be a python script that does the compare stuff

t4c1:
You can already compare branches by doing for example:
```
git checkout branch1 && ./benchmarks/benchmark.py elt_multiply --csv elt_multiply_1.csv && git checkout branch2 && ./benchmarks/benchmark.py elt_multiply --plot_reference elt_multiply_1.csv --plot window
```

If that is not good enough we can also add an extra flag.

rok-cesnovar:
> You can already compare branches by doing for example:

I think this is good enough. I dont think we need to also deal with git in the benchmark python script.

ahartikainen:
I think taking the code from @bbbales2 repo is enough. By default don't compare but if some flag is enabled, then compare.

rok-cesnovar:
What about adding a separate shell script that does just what @t4c1 wrote?

`./compareBenchmark.sh branch1 branch2 args_for_benchmark`

calls 

`git checkout branch1 && ./benchmarks/benchmark.py args_for_benchmark --csv branch1.csv && git checkout branch2 && ./benchmarks/benchmark.py args_for_benchmark --plot_reference branch1.csv --plot window`

Would also be much less code than to handle it with subprocess in Python.

bbbales2:
A shell script is fine, or I'm also happy to just add a bookmark to Tadej's [comment](https://github.com/stan-dev/math/pull/2290#issuecomment-755154528) in my toolbar for quick copy-pasting.

bbbales2 (changes requested):
These things seem good to me and the changes definitely make the output better:

> Fixed wrong flag given to benchmark that would cause it to only output aggregate
> exclude aggregate values when making a plot
> Make plots of individual repeats not affect scaling of axes.

@rok-cesnovar can you check that the OpenCL stuff runs and then approve/merge?

> added option to benchmark OpenCL implementation including only the time for copying var arguments, and not arguments of non-autodiff types (the way OpenCL implementations are often used by Stan)

rok-cesnovar:
> @rok-cesnovar can you check that the OpenCL stuff runs and then approve/merge?

Been using this branch to make OpenCL benchmarks so this definitely works. Will take a quick look and then merge.

ahartikainen (approved):
Maybe run once with black?

    python -m black path/to/file

bbbales2 (approved):

PR #2293, opened 2021-01-06
OpenCL: minor fixes for distributions

rok-cesnovar:
## Summary

Minor fixes for distribution signatures that failed to compile. These are all majority-scalar signatures.

beta_lpdf(real, real, vector)
binomial_lpmf(int, int, vector)
double_exponential_lpdf(real, real, vector)
logistic_lpdf(real, real, vector)
neg_binomial_2_lpmf(int, real, vector)
neg_binomial_2_log_lpmf(vector, real, real)

Other fixes: 
- classification functions compiler for matrix_cl<int>
- calc_if now accepts scalars

No release notes as this is fixing unreleased code.

## Tests

Added tests what would fail without fixes and pass with them for all bugs.

## Side Effects

/

## Checklist

- [x] Math issue #2152 


test/unit/math/opencl/kernel_generator/calc_if_test.cpp
```diff
@@ -94,4 +91,17 @@ TEST(KernelGenerator, calc_if_colwise_min_test) {
   EXPECT_MATRIX_NEAR(correct, res, 1e-9);
 }
 
+TEST(KernelGenerator, calc_if_double) {
+  MatrixXd m(3, 2);
+  m << 1.1, 1.2, 1.3, 1.4, 1.5, 1.6;
+
+  matrix_cl<double> m1_cl(m);
+  matrix_cl<double> m2_cl = m1_cl + stan::math::calc_if<true>(4.0);
```
t4c1, line 99: I always intended `calc_if` to be used as the last operation before the assignment (not used in addition as you did here). I am not sure this will work in general (actually I am pretty sure it won't work for `calc_if<false>`), so I would prefer not to give anyone looking at tests wrong ideas. 

Actually I believe you only needed this to compile within `if(false)`, not actually execute. So this test should be removed.

test/unit/math/opencl/rev/logistic_lpdf_test.cpp
```diff
@@ -155,4 +155,19 @@ TEST(ProbDistributionsLogistic, opencl_matches_cpu_big) {
                                                 mu, sigma);
 }
 
+TEST(ProbDistributionsLogistic, opencl_scalar) {
```
t4c1, line 158:
```suggestion
TEST(ProbDistributionsLogistic, opencl_sigma_mu_scalar) {
```

test/unit/math/opencl/rev/neg_binomial_2_log_lpmf_test.cpp
```diff
@@ -144,4 +144,18 @@ TEST(ProbDistributionsNegBinomial2Log, opencl_matches_cpu_big) {
       neg_binomial_2_log_lpmf_functor_propto, n, eta, phi);
 }
 
+TEST(ProbDistributionsNegBinomial2Log, opencl_matches_eta_phi_scalar) {
```
t4c1, line 147:
```suggestion
TEST(ProbDistributionsNegBinomial2Log, opencl_matches_cpu_eta_phi_scalar) {
```

test/unit/math/opencl/rev/neg_binomial_2_lpmf_test.cpp
```diff
@@ -149,4 +149,19 @@ TEST(muProbDistributionsNegBinomial2, opencl_matches_cpu_big) {
       neg_binomial_2_lpmf_functor_propto, n, mu, phi);
 }
 
+TEST(muProbDistributionsNegBinomial2, opencl_scalar_n_mu) {
```
t4c1, line 152:
```suggestion
TEST(ProbDistributionsNegBinomial2, opencl_scalar_n_mu) {
```

t4c1 (changes requested):

test/unit/math/opencl/kernel_generator/calc_if_test.cpp
rok-cesnovar: Yeah, it was only to make a signature compile. It never actually ran. Will remove the test.

t4c1 (approved):

Issue #1854, opened 2020-04-29
Implement matrix_cl overloads for prim & rev functions


rok-cesnovar:
## Description

For a list of functions that have OpenCL support for both prim and rev for all exposed signatures see http://mc-stan.org/math/opencl_support.html

Functions with partial or no support that we are targeting to support in the immediate future:
|                                   | prim | rev |
|-----------------------------------|:----:|:---:|
| chol2inv                          |      |     |
| generalized_inverse                           |      |     |
| hmm_marginal                      |      |     |
| inverse_spd                           |      |     |
| lmgamma                           |      |     |
| log_mix                           |      |     |
| log_sum_exp                       |      |     |
| mdivide_left_spd                  |      |     |
| mdivide_right_spd                 |      |     |
| quad_form_diag                    |      |     |
| quad_form                         |      |     |
| quad_form_sym                     |      |     |
| sd                                |      |     |
| trace_gen_quad_form               |      |     |
| trace                             |   x  |     |
| trace_quad_form                   |      |     |
| variance                          |      |     |

Legend: 
`x`  done
`*`  not all signatures are supported

We also have implementations of rep_vector/row_vector but those can not be used in Stan with OpenCL because the inputs are scalars. Maybe only if we give them a suffix and stanc3 generates a suffixed call to rep_vector/row_vector. It's the same for `rep_matrix(real, int, int)`.

Other functions not in our immediate plans:
- cumulative_sum
- bessel_first_kind, bessel_second_kind
- choose
- csr_extract_u, csr_extract_v, csr_extract_w
- csr_matrix_times_vector, csr_to_dense_matrix
- determinant
- eigenvalues_sym, eigenvectors_sym
- falling_factorial, log_falling_factorial, rising_factorial
- log_determinant, log_softmax
- modified_bessel_first_kind, modified_bessel_second_kind, log_modified_bessel_first_kind
- rank
- singular_values
- softmax
- sort_asc, sort_desc, sort_indices_asc, sort_indices_desc
- qr_Q, qr_R, qr_thin_Q, qr_thin_R

rok-cesnovar:
Rewrote this issue to now be a TODO list for non-distributions functions (_lpdf/_lpmf,_lcdf, _lccdf, etc are excluded). For now, I went through /prim again, will update /rev tomorrow.

I am also attaching a test file with which I double-checked all of them compile.
[prim_sigs.cpp.txt](https://github.com/stan-dev/math/files/5655028/prim_sigs.cpp.txt)

PR #2295, opened 2021-01-08
OpenCL: add rev hypot, multiply_log, squared_distance, distance, pow, log_inv_logit_diff, log_diff_exp

rok-cesnovar:
## Summary

Adds OpenCL rev support for hypot, multiply_log, squared_distance, distance, pow, log_inv_logit_diff, log_diff_exp.

## Tests

/

## Side Effects

/

## Release notes

Added OpenCL rev support for hypot, multiply_log, squared_distance, distance, pow, log_inv_logit_diff and log_diff_exp.

## Checklist

- [x] Math issue #1854 

- [x] Copyright holder: Rok Češnovar, Uni. of Ljubljana


stan/math/opencl/prim/squared_distance.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param a first kernel generator expression
+ * @param b second kernel generator expression
+ * @return squared distance.
+ * @throw std::domain_error If the matrices are not the same
+ * size
+ */
+template <typename T_a, typename T_b,
+          require_all_kernel_expressions_and_none_scalar_t<T_a, T_b>* = nullptr>
+inline auto squared_distance(T_a&& a, T_b&& b) {
```
t4c1, line 30:
```suggestion
inline auto squared_distance(const T_a& a, const T_b& b) {
```

stan/math/opencl/rev/log_diff_exp.hpp
```diff
@@ -0,0 +1,71 @@
...
+  arena_t<T_x> x_arena = std::forward<T_x>(x);
+  arena_t<T_y> y_arena = std::forward<T_y>(y);
+
+  matrix_cl<double> res_val
+      = log_diff_exp(value_of(x_arena), value_of(y_arena));
+
+  return make_callback_var(
+      res_val,
+      [x_arena, y_arena](const vari_value<matrix_cl<double>>& res) mutable {
+        if (!is_constant<T_x>::value && !is_constant<T_y>::value) {
```
t4c1, line 41: What is the reason for separete branches here?

stan/math/opencl/rev/pow.hpp
```diff
@@ -0,0 +1,67 @@
...
+  arena_t<T_a> a_arena = std::forward<T_a>(a);
+  arena_t<T_b> b_arena = std::forward<T_b>(b);
+
+  matrix_cl<double> res_val = pow(value_of(a_arena), value_of(b_arena));
+
+  return make_callback_var(
+      res_val,
+      [a_arena, b_arena](const vari_value<matrix_cl<double>>& res) mutable {
+        auto a_deriv = select(
+            isnan(value_of(a_arena)) || isnan(value_of(a_arena)),
```
t4c1, line 41:
```suggestion
            isnan(value_of(res)),
```

stan/math/opencl/rev/pow.hpp
```diff
@@ -0,0 +1,67 @@
...
+            constant(NOT_A_NUMBER, res.rows(), res.cols()),
+            select(
+                value_of(a_arena) == 0.0, constant(0, res.rows(), res.cols()),
+                elt_multiply(
+                    res.adj(),
+                    elt_multiply(value_of(b_arena),
+                                 elt_divide(res.val(), value_of(a_arena))))));
+        auto b_deriv = select(
+            isnan(value_of(a_arena)) || isnan(value_of(a_arena)),
```
t4c1, line 50:
```suggestion
            isnan(value_of(res)),
```
Also you can extract common operations from derivatives of a and b into a variable(s), so they will only be calculated once.

t4c1 (changes requested):
I only mentioned it once, but there might be other places where you are calculating two expressions in one kernel. In such cases you should extract common operations into separate variables so that they are not repeated in the kernel.

stan/math/opencl/rev/log_diff_exp.hpp
rok-cesnovar: Because that is how the CPU version is done: https://github.com/stan-dev/math/blob/develop/stan/math/rev/fun/log_diff_exp.hpp

I dont know enough about the math behind this and in those cases I assume the CPU version is done correctly and I just want to mimic it.

t4c1: To me it seems that the vv branch should be able to handle all cases. @SteveBronder @bbbales2 Do you have any idea why there is special inf handling in vd and dv cases?

rok-cesnovar: I do agree it looks weird. My guess would be that we should probably handle those in vv as well.

bbbales2: > there is special inf handling in vd and dv cases

I don't know why it is there and I don't like it.

I was copying some old code recently and Steve pointed out where it was doing something like:
```
bvi_->adj_ -= adj_ * INFTY;
```

Why do we multiply the infinity by infinity? As long as `adj_` is finite that is just infinity. Best you can do is check the git blame on the CPU thing and see if there was a recent issue.

I guess options are:
1. Remove the checks in CPU and GPU to make them consistent
2. Leave the checks out in GPU and assume it won't cause major differences
3. Copy the checks on the GPU code and just roll with it
4. Copy the checks to the GPU code and plan to remove them in the future

Especially since the CPU is inconsistent with itself, I'm happy with anything here.

rok-cesnovar: The reason you would multiply with infinity is for edge cases like 0 * Inf = NaN. I dont think that matters in Stan, but might otherwise.

The way its done currently is consistent with the CPU version where vv has no checks and vd and dv has checks.

rok-cesnovar: I think the best plan is to keep GPU version consistent with the CPU and make a separate issue to get more eyes on that? @t4c1 what do you think?

t4c1: I would prefer to remove these checks from both versions, but I am fine with either option.

t4c1 (approved):

PR #2306, opened 2021-01-12
Revert "Multi expression"

rok-cesnovar:
Reverts stan-dev/math#2247

This was causing issues with Windows 3.5 that we were unable to address.

I think long term we need to check if this is specific to the mingw32 g++ 4.9.3 and if it is just drop support for Windows 3.5. 
Its causing us too many headaches at this point. If its not, then we need to move to C++14 or at least make a serious plan on when to do it.

Short term (this release cycles) we are a bit late to do that and reverting is the only way to go forward unfortunately.


rok-cesnovar:
The revert seems to be clean. 

@bbbales2 @SteveBronder did any of you use multi expressions for any PRs? If no, then this should be fine.

bbbales2:
I don't think I ever used it (or am using it in any outstanding pulls). But I do look forward to using it once this is fixed.

bbbales2 (approved):
Good!

PR #2312, opened 2021-01-14
Add linspaced_int_array()

rok-cesnovar:
## Summary

Adds support for `ìnt[] linspaced_int_array(int, int, int)` as a complement to `real[] linspaced_int_array(int, real, real)`. This was proposed by @avehtari in a discussion and I agree it would be useful.

`linspaced_int_array(5,0,8)` returns 0,2,4,6,8
`linspaced_int_array(5,0,9)` returns 0,2,4,6,8
`linspaced_int_array(5,0,11)` returns 0,2,4,6,8
`linspaced_int_array(5,0,12)` returns 0,3,6,9,12

So the behaviour is the same as for `Eigen::VectorXi v = Eigen::VectorXi::LinSpaced(K, low, high)`

## Tests

/

## Side Effects

/

## Release notes

Added linspaced_int_array.

## Checklist

### DIFF
--- a/stan/math/prim/fun/linspaced_int_array.hpp
+++ b/stan/math/prim/fun/linspaced_int_array.hpp
@@ -11,7 +11,7 @@ namespace math {
 /**
  * Return an array of linearly spaced integers.
  *
- * This produces an array from low to high (included) with integers spaced
+ * This produces an array from low to high (inclusive) with integers spaced
  * as (high - low) / (K - 1). For K=1, the array will contain the high value;
  * for K=0 it returns an empty array.
  *
@@ -20,8 +20,7 @@ namespace math {
  * @param high largest value
  * @return An array of size K with elements linearly spaced between
  * low and high.
- * @throw std::domain_error if K is negative, if high is less than low
- * or if ().
+ * @throw std::domain_error if K is negative or high is less than low.
  */
 inline std::vector<double> linspaced_int_array(int K, int low, int high) {
   static const char* function = "linspaced_int_array";
--- a/stan/math/prim/fun/linspaced_int_array.hpp
+++ b/stan/math/prim/fun/linspaced_int_array.hpp
@@ -13,8 +13,8 @@ namespace math {
  *
  * This produces an array from `low` to `high` (inclusive). If `high - low` is
  * greater or equal to `K - 1`, then the integers are evenly spaced. If it is
- * not possible to get from `low` to `high` with a multiple of an integer, `high`
- * is lowered until this is possible.
+ * not possible to get from `low` to `high` with a multiple of an integer,
+ * `high` is lowered until this is possible.
  *
  * If `K - 1` is greater than `high - low`, then integers are repeated. For
  * instance, `low, low, low + 1, low + 1, ...`. `high` is lowered until `K - 1`
--- a/stan/math/prim/fun/linspaced_int_array.hpp
+++ b/stan/math/prim/fun/linspaced_int_array.hpp
@@ -13,7 +13,8 @@ namespace math {
  *
  * This produces an array from low to high (inclusive) with integers spaced
  * as (high - low) / (K - 1). For K=1, the array will contain the high value;
- * for K=0 it returns an empty array.
+ * for K=0 it returns an empty array. If there are less integers between
+ * high and low than K, there will be duplicate integers.
  *
  * @param K size of the array
  * @param low smallest value
@@ -32,7 +33,6 @@ inline std::vector<int> linspaced_int_array(int K, int low, int high) {
 
   Eigen::VectorXi v = Eigen::VectorXi::LinSpaced(K, low, high);
   return {v.data(), v.data() + K};
-  ;
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/linspaced_int_array.hpp
+++ b/stan/math/prim/fun/linspaced_int_array.hpp
@@ -1,0 +1,41 @@
+#ifndef STAN_MATH_PRIM_FUN_LINSPACED_INT_ARRAY_HPP
+#define STAN_MATH_PRIM_FUN_LINSPACED_INT_ARRAY_HPP
+
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return an array of linearly spaced integers.
+ *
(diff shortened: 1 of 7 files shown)


stan/math/prim/fun/linspaced_int_array.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * as (high - low) / (K - 1). For K=1, the array will contain the high value;
+ * for K=0 it returns an empty array.
+ *
+ * @param K size of the array
+ * @param low smallest value
+ * @param high largest value
+ * @return An array of size K with elements linearly spaced between
+ * low and high.
+ * @throw std::domain_error if K is negative, if high is less than low
+ * or if ().
```
bbbales2, line 24: `if ()`?

stan/math/prim/fun/linspaced_int_array.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return an array of linearly spaced integers.
+ *
+ * This produces an array from low to high (included) with integers spaced
```
bbbales2, line 14: included -> inclusive

stan/math/prim/fun/linspaced_int_array.hpp
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param K size of the array
+ * @param low smallest value
+ * @param high largest value
+ * @return An array of size K with elements linearly spaced between
+ * low and high.
+ * @throw std::domain_error if K is negative, if high is less than low
+ * or if ().
+ */
+inline std::vector<double> linspaced_int_array(int K, int low, int high) {
```
bbbales2, line 26: `std::vector<int>`

stan/math/prim/fun/linspaced_int_array.hpp
```diff
@@ -0,0 +1,41 @@
...
+  static const char* function = "linspaced_int_array";
+  check_nonnegative(function, "size", K);
+  check_greater_or_equal(function, "high", high, low);
+  if (K == 0) {
+    return {};
+  }
+
+  Eigen::VectorXi v = Eigen::VectorXi::LinSpaced(K, low, high);
```
bbbales2, line 34: Build the `std::vector<int>` directly.

Some variation of this I guess:
```c++
if(K == 1) {
  return { high };
}

std::vector<int> v(K);
int d = (high - low) / (K - 1);
for(size_t k = 0; k < K; ++k) {
  v[k] = k * d + low;
}
return v;
```

test/unit/math/prim/fun/linspaced_int_array_test.cpp
```diff
@@ -0,0 +1,31 @@
+#include <stan/math/prim.hpp>
+#include <test/unit/util.hpp>
+#include <gtest/gtest.h>
+#include <limits>
+#include <vector>
+
+TEST(MathFunctions, linspaced_int_array) {
+  using stan::math::linspaced_int_array;
+  EXPECT_STD_VECTOR_EQ(linspaced_int_array(0, 1, 5), std::vector<int>({}));
+  EXPECT_STD_VECTOR_EQ(linspaced_int_array(1, 1, 5), std::vector<int>({5}));
```
bbbales2, line 10: Huh, that's surprising. I'd have expected the output to be `1`. `5` seems strange. Is that really what Eigen does?

bbbales2 (changes requested):
Review!

test/unit/math/prim/fun/linspaced_int_array_test.cpp
bbbales2: Actually I guess we should do whatever the already existing linspaced stuff does.

rok-cesnovar: Yeah that is why its done that way.

stan/math/prim/fun/linspaced_int_array.hpp
rok-cesnovar: Is there an advantage to that approach? This is the same impl. as the other linspaced function. But can also do that.

stan/math/prim/fun/linspaced_int_array.hpp
bbbales2: This is still a `std::vector<double>` on return.

stan/math/prim/fun/linspaced_int_array.hpp
bbbales2: It avoids two allocates (the local `std::vector<int>` will get returned directly by compiler optimizing returns) and the allocates and copies will be the expensive part of this.

But in the spirit of not tripping up on edge cases (that implementation I gave I think doesn't work for some of the examples I see [here](http://eigen.tuxfamily.org/dox-devel/classEigen_1_1DenseBase.html#title49)), I'll also take ```return {v.data(), v.data() + K};```

rok-cesnovar: Made it return {v.data(), v.data() + K};

stan/math/prim/fun/linspaced_int_array.hpp
rok-cesnovar: Bah, silly me

bbbales2:
I'm still confused on the docs on this.

This is how Eigen says it (from [here](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#aaef589c1dbd7fad93f97bd3fa1b1e768)):

> For integer scalar types, an even spacing is possible if and only if the length of the range, i.e., high-low is a scalar multiple of size-1, or if size is a scalar multiple of the number of values high-low+1 (meaning each value can be repeated the same number of time). If one of these two considions is not satisfied, then high is lowered to the largest value satisfying one of this constraint.

That last sentence is hard to parse.

rok-cesnovar:
I think it just mean when we specify

Eigen::VectorXi::LinSpaced(5, 0, 11); it lowers 11 until its a valid linspaced vector (under the two conditions).

bbbales2 (commented):
Lemme know if you approve my changes and I'll merge. This looks good to me.

rok-cesnovar:
All good! Thanks.

rok-cesnovar:
@bbbales2 approve?

bbbales2 (approved):

PR #2326, opened 2021-01-19
(WIP) Additional RKs from ARKode

yizhang-yiz:
## Summary

I'm weighing on replacing `boost::odeint` with `sundials::ARKode`. 

Pro:

- More consistent implementation with BDF & Adams integrators.
- Much easier to add new integrators from ARKode (just replacing method id, similar to CVODES)
- Much easier to manually add new integrators such as `Tsit5` (DifferentialEquations.jl's default), by adding a [Butcher table](https://matlabgeeks.weebly.com/uploads/8/0/4/8/8048228/butcher_tables-v2.pdf).
- Much easier to output performance diagnostics of the solver (`Get` functions similar to CVODES)
- Future feature to support solving ODE in form: `M y' = f(t, y, p)`.
- Future feature to support multiscale ODE(see ARKode user guide).

Con:

- Previous model's results can not be exactly reproduced, due to implementation details, in particular time adaptive method.
- Another library(ARKode) increases maintenance cost. 
- ?

## Tests

New candidates are named `ode_erk45` and `integrate_ode_erk45` for the moment to differentiate from existing rk45 solvers. Unit tests from rk45 are reproduced into `erk45` test:

```bash
bash-3.2$ find . -name "*erk45*_test.cpp"
./test/unit/math/rev/functor/integrate_ode_erk45_grad_test.cpp
./test/unit/math/rev/functor/integrate_ode_erk45_rev_test.cpp
./test/unit/math/prim/functor/ode_erk45_tol_prim_test.cpp
./test/unit/math/prim/functor/ode_erk45_prim_test.cpp
```

## Side Effects

Existing models based on rk45 solver will see performance bifurcation. Though based on `lotka-volterra` model it's not significant:

```bash
./lotka-volterra_rk45 sample data file=lotka-volterra.data.R init=lotka-volterra.init.R random seed=987
....
 Elapsed Time: 150.952 seconds (Warm-up)
               160.909 seconds (Sampling)
               311.861 seconds (Total)
```
```bash
./lotka-volterra_erk45 sample data file=lotka-volterra.data.R init=lotka-volterra.init.R random seed=987
....
 Elapsed Time: 170.164 seconds (Warm-up)
               145.126 seconds (Sampling)
               315.29 seconds (Total)
```


yizhang-yiz:
Main discussion at this point is probably around:

- do we want to do this?
- if we do, how to make transition?
- what tests should be passed as necessity and sufficiency condition to accept the replacement?

@bbbales2 @wds15 @rok-cesnovar @charlesm93

yizhang-yiz:
An additional minor benefit is to make it trivial to support #2244 in rk45.

rok-cesnovar:
Thanks!

Will leave the ODE-specific points to the experts, my thoughts on the points I can comment.

> increases maintenance cost

I do not see this being a problem as upgrading Sundials is handled by the upgrade script (this line needs to be modified: https://github.com/stan-dev/math/blob/develop/lib/upgrade-sundials.sh#L117) and the more consistent implementation actually helps maintenance I guess.

> integrate_ode_erk45

I do not think we need the old interface for this, unless there is a reason one would use it. Otherwise this is just added docs/compiler work. The only exception we have for deprecated function is the deprecated distribution suffixes, but that is a bit of a different story.

yizhang-yiz:
@rok-cesnovar by maintenance I meant to say things like API change. This has happened a couple of times on CVODES. It's not a deal breaker but something we need keep an eye on. `erk45` is the name I use at this point to help benchmark against `rk45`, if goes in it'll be renamed `rk45`, hence replacement.

bbbales2:
Whoops, I thought I responded to this. I meant to say, the new code looks really simple and it does seem nice to be totally in Sundials-land.

> Much easier to output performance diagnostics of the solver

What are these?

Shame it doesn't do adjoint.

> Much easier to manually add new integrators such as Tsit5 (DifferentialEquations.jl's default)

What makes this a good default? That is interesting that the DifferentialEquations.jl people do it.

yizhang-yiz:
> What are these?

CVODES has things like `CVODEGetNumSteps` etc that I routinely use to monitor solver behavior. ARKode has similar.

> What makes this a good default? That is interesting that the DifferentialEquations.jl people do it.

Their benchmark shows better performance. It's a new scheme proposed in 2011 so not many solves have it. I don't like "they have so we should have" but ARKode makes it easy to explore.

bbbales2:
> CVODES has things like CVODEGetNumSteps etc that I routinely use to monitor solver behavior. ARKode has similar.

THis sounds like useful information but I don't know how to pipe it out.

I would like to at least throw all the useful CVODES/Sundials errors we can.

Just as an arbitrary goal to shoot for, if this can give us better error messages than we have now + vector error tolerances, that seems pretty good.

However I remember it's been difficult in the past for interfaces to expose the CVODES integrators. Like `expose_stan_functions` doesn't work right or something? That could mess this up. @wds15 does this sound familiar? Maybe there was a problem in PyStan too, though I don't know if it exists in PyStan3.

yizhang-yiz:
> THis sounds like useful information but I don't know how to pipe it out.

> I would like to at least throw all the useful CVODES/Sundials errors we can.

I use those function mostly for diagnosing, not user facing. To improving msg, a simple half-way solution might be use `CVodeGetReturnFlagName` and `ARKStepGetReturnFlagName` to output flag string and attach it to exception.

yizhang-yiz:
Great timing of this post http://sundials.2283335.n4.nabble.com/Performance-relative-to-Julia-implementation-td4655937.html. It worries me. We need more tests.

wds15:
`expose_stan_functions` in R used to have problems with any stiff solver as it required nested AD to work, which was not possible (or even is not possible, not sure what is the status).

The major problem with ditching odeint for our interfaces is PyStan. PyStan did not deploy any pre-compiled stuff like CVODES such that any Stan program used to fail to compile with the PyStan 2 which is very broadly being used. That is now history in the sense that PyStan3 is around and there these things work out of the box as I recall. It's just that PyStan3 is not yet as widely adopted and - here is the bummer - is not supported on Windows as I recall for now.

Comparing these runtimes between integrators is really hard. One needs to ensure that the definition of the tolerances is actually similar enough so that at the same tolerance targets we are really getting the same accuracy from the solvers. Only if you match that, then you can actually compare the runtimes.

In the past I found that the author of odeint was very responsive and helpful. From this angle it would be a pity to drop that. On the other hand, odeint is mostly a 1-man thing with a bus-driver factor of 1. ARKode is maintained by more people as I understand, but I don't have any idea about their resources. If ARKode is driven by the same community as Sundials (which I assume it is), then ARKode is maybe even a better choice, since the bus-driver factor is above 1 (bus-driver factor of 1 = single point of failure).

wds15:
> Great timing of this post http://sundials.2283335.n4.nabble.com/Performance-relative-to-Julia-implementation-td4655937.html. It worries me. We need more tests.

The comparison is with Julia... and Julia is fast for ODE solving as far as I know.

yizhang-yiz:
Some SIR model results:
```r
> res1
      num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45     273462 218.6073  0.0007994065         0.1966093         0.1220845 1234
erk45    262626 205.5197  0.0007825566         0.1516937         0.1241295 1234
> res2
      num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45     272956 216.2838  0.0007923759         0.1522725         0.1117276 1235
erk45    266538 209.4610  0.0007858578         0.1737881         0.1291991 1235
> res3
      num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45     273822 215.9676  0.0007887153         0.1998471         0.1297365 9876
erk45    261960 208.6028  0.0007963156         0.1976375         0.1322253 9876
```
The two solvers seem on par. Note here `erk45` is based on ARKode's default Cash-Karp scheme instead of Dopri5.

yizhang-yiz:
While the above is encouraging, ARKode's implementation of dopri5 is completely lacking:
```r
      num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45     134912 63.18806  0.0004683650        0.04682429        0.03628001  987
erk45    129024 99.04404  0.0007676405        0.07678596        0.04854766  987
```
I believe this is due to ARKode's default adaptive time stepping is not well tuned toward dopri5, corroborated by the earlier post I linked.

So I don't think we should really do the replacement. Instead, an addition of Cash-Karp would benefit problems with mild stiffness due to the method's better stability & capability of treating near discontinuity. For example, solving the famous Field-Noyes system:
```cpp
struct FN_osc {
  template <typename T0, typename T1, typename T2>
  inline std::vector<stan::return_type_t<T1, T2>>
  operator()(const T0& t_in, const std::vector<T1>& y,
             const std::vector<T2>& theta, const std::vector<double>& x,
             const std::vector<int>& x_int, std::ostream* msgs) const {
    if (y.size() != 3)
      throw std::domain_error(
          "this function was called with inconsistent state");

    const T2& a = theta[0];

    std::vector<stan::return_type_t<T1, T2>> res(3);
    res[0] = a * (y[1] - y[0] * y[1] + y[0] - 8.375e6 * y[0] * y[0]);
    res[1] = (-y[1] - y[0] * y[1] + y[2]) / a;
    res[2] = 0.161 * (y[0] - y[2]);

    return res;
  }
};
```
using BDF, Cash-karp(ckrk), and rk45 with increasing stiffness, characterized by `theta`
#### `theta[0] = 0.07`
```bash
bdf results: 0.00031088 0.808966 0.800105
[       OK ] perf.FN_osc_bdf (1 ms)
[ RUN      ] perf.FN_osc_ckrk
ckrk results: 0.000310831 0.808752 0.799894
[       OK ] perf.FN_osc_ckrk (11 ms)
[ RUN      ] perf.FN_osc_rk
rk45 results: 0.000310833 0.808752 0.799894
[       OK ] perf.FN_osc_rk (9 ms)
```
#### `theta[0] = 0.7`
```bash
bdf results: 0.000328029 0.901109 0.79991
[       OK ] perf.FN_osc_bdf (2 ms)
[ RUN      ] perf.FN_osc_ckrk
ckrk results: 0.000328028 0.901102 0.799905
[       OK ] perf.FN_osc_ckrk (95 ms)
[ RUN      ] perf.FN_osc_rk
rk45 results: nan nan nan
[       OK ] perf.FN_osc_rk (0 ms)
```
#### `theta[0] = 7.0`
```bash
bdf results: 0.000425371 1.5156 0.799915
[       OK ] perf.FN_osc_bdf (1 ms)
[ RUN      ] perf.FN_osc_ckrk
ckrk results: 0.000425375 1.51562 0.799911
[       OK ] perf.FN_osc_ckrk (881 ms)
[ RUN      ] perf.FN_osc_rk
rk45 results: nan nan nan
[       OK ] perf.FN_osc_rk (0 ms)
```
#### `theta[0] = 77.0`
```bash
bdf results: 0.000379164 1.20411 0.799395
[       OK ] perf.FN_osc_bdf (1 ms)
[ RUN      ] perf.FN_osc_ckrk
ckrk results: 0.000379157 1.20407 0.799853
[       OK ] perf.FN_osc_ckrk (7954 ms)
[ RUN      ] perf.FN_osc_rk
rk45 results: nan nan nan
```
ckrk can still dredge on under highly stiffness but rk45 blows up early on.

yizhang-yiz:
Since the library makes it easy to switch scheme, here's fitting lotka-volterra model with higher tolerance(1e-10 & 1e-10), where higher-order verner methods prevails:
```r
> verner.lotka.3
      num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45     140620 358.2561   0.002547689         0.2539072         0.1893994 4567
verner   137564 271.8165   0.001975928         0.1905395         0.1440348 4567
> verner.lotka.2
      num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail  seed
rk45     117518 331.2800   0.002818972        23.6830755        28.8700229 23498
verner   146200 284.6393   0.001946918         0.1760946         0.1272192 23498
```

so I'd like to gauge @wds15 @bbbales2 's option on using arkode to add additional explicit RK integrators (ckrk & verner), while keeping the original rk45 in place, as
- ckrk deals with stiffy problem much better than rk45, while not lags too much behind in regular problems (10-20% less efficient for those problems). It adds an additional spot in stiffness/stability spectrum.
- verner deals medium-high accuracy much more efficiently than rk45, while offering similar stability.
- potentially we can also add high-order Fehlberg78 but it doesn't look like there's enough motivation to do it.

bbbales2:
@bgoodri does `expose_stan_functions` support the CVODES integrators (`ode_bdf` instead of `ode_rk45`)?

@ahartikainen what's the CVODES support in PyStan 2 & PyStan 3 look like?

bbbales2:
@yizhang-yiz I'm most worried about the interfaces so let's see what the status there is.

bgoodri:
Yeah, you can wrap it with `expose_stan_functions`. `integrate_ode_bdf` used to crash on Windows for several years, but it is valid syntax.

yizhang-yiz:
FWIW odeint has its own Cash-Karp implementation. For the above FN oscillator problem with high stiffness I'm getting similar performance
```bash
bdf results: 0.000379164 1.20411 0.799395
[       OK ] perf.FN_osc_bdf (1 ms)
[ RUN      ] perf.FN_osc_ckrk
arkode::ckrk results: 0.000379157 1.20407 0.799853
[       OK ] perf.FN_osc_ckrk (6094 ms)
[ RUN      ] perf.FN_osc_rk
odeint::ckrk results: 0.000379157 1.20407 0.799853
[       OK ] perf.FN_osc_rk (6074 ms)
```

wds15:
@bbbales2 I think any explicit ode solver is fine with the expose stuff. The issue is that implicit solver trigger nested AD calls to get the Jacobian which never worked... but I think that is a thing of the past if I am not mistaken. This has been fixed once @rok-cesnovar plumbed through the adequate stanc3 expose stan functions API thing.

@yizhang-yiz I have just been looking at the available steppers from odeint ( https://www.boost.org/doc/libs/1_75_0/libs/numeric/odeint/doc/html/boost_numeric_odeint/odeint_in_detail/steppers.html#boost_numeric_odeint.odeint_in_detail.steppers.stepper_overview ) and they have a large selection there as well. The velocity-verlet is the same you have maybe? The one which I always found interesting is the Bulirsch-Stoer Dense Output, which is supposed to be fast for high-precision output.

Overall I think that a non-stiff solver which can survive some stiffness is a good thing, since the stan sampler drives during warmup things into no mans land every now and then. Returning a solution then is much much better than choking up.

Fast high-precision integrators are also a good addition in my eyes... trying Bulirsch-Stoer would be great in this context. This is what odeint recommends.

Given that odeint has all those already build in I just wonder why we would want to add ARKode? I don't mind doing that, don't get me wrong.. it's just the usual keep dependencies low (but we anyway have sundials).

What is just concerning to me is the inflation of ODE solvers. That will be hard to navigate for users. Hopefully the profiling functionality will prove useful here as users can then quickly do a few sampling steps with all integrators and then just pick the one which does best.

yizhang-yiz:
@wds15 
> Overall I think that a non-stiff solver which can survive some stiffness is a good thing, since the stan sampler drives during warmup things into no mans land every now and then. Returning a solution then is much much better than choking up.

+1

> Fast high-precision integrators are also a good addition in my eyes... trying Bulirsch-Stoer would be great in this context. This is what odeint recommends.

As said earlier we can tentatively add high-order methods but I'm not sure how many user would benefit that. Can we defer it? Also GBS is not the only option, high-order Fehlberg and Verner tend to do well in this area and we'll need more benchmark.

> Given that odeint has all those already build in I just wonder why we would want to add ARKode? 

In terms of adding new integrators what arkode offers out of box is impilcit RKs like DIRK. odeint does have implicit but requires boost::ublas to work. Additional benefit of arkode is diagnostics report, which I found very useful when I want to find out the number of error test failures, number of function evaluations, min & max stepsize, etc.

> What is just concerning to me is the inflation of ODE solvers. That will be hard to navigate for users. Hopefully the profiling functionality will prove useful here as users can then quickly do a few sampling steps with all integrators and then just pick the one which does best.

+1

yizhang-yiz:
So I decided to compare `arkode::cash_karp` vs `odeint::cash_karp` in actually sampling, turns out just like dopri odeint blows arkode out of water by being ~30% more efficient. To me this indicates we should put arkode aside for now(I tried it a couple of yrs ago when it's announced and was not impressed, was hoping it's better by now). The lacking in performance doesn't really inspire confidence in the rest of its integrators.

On the other hand, odeint's cash_karp (`ckrk`) proves to be a valuable addition:
#### Lotka-Volterra model
```r
     num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45    123248 81.72574  0.0006630999        0.06304176        0.04539207 4568
ckrk    143528 77.11219  0.0005372624        0.06092892        0.03635768 4568
```
```r
     num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45    136766 95.24508  0.0006964091        0.06225755        0.04962943 4566
ckrk    141836 75.38855  0.0005315191        0.05707595        0.03396805 4566
```
```r
     num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45    140208 98.96743  0.0007058615        0.07374781        0.05212291 1234
ckrk    139242 83.96585  0.0006030210        0.06791950        0.04340841 1234
```
Stuck in one of the chains:
```r
     num_leaps      time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45    134816  96.20186  0.0007135789        0.07065881         0.0496017 4567
ckrk   2118400 943.78793  0.0004455192      180.58655799        40.6336318 4567
```
#### SIR model
```r
     num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45    284392 250.4310  0.0008805838         0.2097617         0.1578559  987
ckrk    276010 226.4161  0.0008203184         0.1878465         0.1117842  987
```
```r
     num_leaps     time time_per_leap time.per.ess.bulk time.per.ess.tail seed
rk45    247814 238.9557  0.0009642543         0.1749379         0.1292826  988
ckrk    280442 214.1838  0.0007637366         0.1462981         0.1241772  988
```
So I'm proposing to add odeint::ckrk. Does it make sense? @bbbales2 @wds15

gardner48:
Apologies for crashing the party, @yizhang-yiz asked a question about the ARKODE + XBraid examples and I found my way here. The SUNDIALS team is always happy to see groups interested in using our code and we welcome any questions or feedback you may have. We are a small team but try our best to answer user questions as time allows. 

A quick comment on the performance comparisons above:

> Comparing these runtimes between integrators is really hard. One needs to ensure that the definition of the tolerances is actually similar enough so that at the same tolerance targets we are really getting the same accuracy from the solvers. Only if you match that, then you can actually compare the runtimes.

> EDIT: to be fair arkode comes with many more switches for stepsize adaptivity and should one be willing to invest on tuning he'll be able to find the near-optimum.

These are both excellent points. With regards to the comparison with Julia, ARKODE and Julia compute error weights slightly differently and this is one factor in the performance differences. We're currently investigating how much of an impact this has compared to other differences between the implementations i.e., the default adaptivity/parameter choices. Many of the default settings in ARKODE mirror those used in CVODE and so are geared towards implicit methods. For ERK methods, calling `ARKStepSetOptimalParams` will likely give you a better set of parameter values to start from but the optimal choice is highly problem dependent.

Feel free to reach out if you have any questions.

yizhang-yiz:
@gardner48 thanks for addressing these issues. In fact, odeint uses different error norm as well as a different way how relative tolerance is applied. But its defaults usually outperforms arkode. I can tune arkode's adaptive scheme to match the timing performance, but it is usually ad hoc. For example, in the above Field-Noyes system, using I controller degrades the performance, but in the SIR model the it outperforms default PID controller. In our application (Stan language) an integrator is applied to a system with a variety of parameter samples and we don't have luxury to change adaptiviy schemes,  and it seems odeint's default suits better in this context.

I haven't tried `ARKStepSetOptimalParams`, thanks for pointing that out!

yizhang-yiz:
Went back to use `ARKStepSetOptimalParams` to set adaptivity for dopri5, and had 10 runs of [lotka-volterra model ](https://github.com/stan-dev/math/blob/5df0fa7fee687f6909d1b36db5458feda2d6da20/rk45_benchmark/lotka-volterra/lotka-volterra.stan) using dopri5 from arkode & odeint, summarized below
```r
 num_leaps       time time_per_leap time.per.ess.bulk time.per.ess.tail  seed    lib
   1123360  508.89280  0.0004530095       75.77987269       44.62984957 12341 odeint
   1130400  938.44598  0.0008301893      129.04189643      232.73460324 12341 arkode
    126392   61.64068  0.0004876945        0.04493312        0.02771879 12342 odeint
   1126532 1196.59922  0.0010621973      165.90099035      117.65063857 12342 arkode
    137316   62.94479  0.0004583937        0.04694849        0.02671907 12343 odeint
   1126308  725.77602  0.0006443850      107.39184040       64.78661730 12343 arkode
    143996   65.34392  0.0004537898        0.04977942        0.03094822 12344 odeint
    144192  119.45862  0.0008284691        0.08817383        0.06080040 12344 arkode
    131656   62.94913  0.0004781334        0.05058288        0.03459067 12345 odeint
   1128056 1127.72230  0.0009997042      155.44065733      279.67513154 12345 arkode
    144384   66.07330  0.0004576220        0.04725021        0.03286483 12346 odeint
    137036  112.12827  0.0008182395        0.07704422        0.05558253 12346 arkode
    129104   62.15179  0.0004814087        0.04614763        0.02723654 12347 odeint
    162032  124.70043  0.0007696037        0.08954593        0.06917164 12347 arkode
    129042   63.17642  0.0004895803        0.04461110        0.03193203 12348 odeint
   1123288  827.23970  0.0007364449      122.35647945       73.56485868 12348 arkode
    137180   66.45720  0.0004844525        0.04513831        0.03521983 12349 odeint
   1122568  593.45417  0.0005286577       87.55888292       52.42341198 12349 arkode
    129900   61.50517  0.0004734809        0.04467898        0.03039325 12350 odeint
    146712  117.73771  0.0008025090        0.07750070        0.05013397 12350 arkode
```
Some observations: 
- arkode tends to have more stuck chains where total # of leapfrogs blows up. odeint is not free of this issue but occurred only once.
- in the runs without stuck chains odeint takes about half the time of arkode, similar can be said about time per leapfrog and time per ess. Plot below is based on _all_ the runs.

yizhang-yiz:
Moving `ckrk` discussion to #2334 . The current thread can be continued when further benchmark becomes available with tuned akrode and/or arkode's implicit RK methods. Close for now.

ahartikainen:
> @ahartikainen what's the CVODES support in PyStan 2 & PyStan 3 look like?

PyStan2 has a branch that has some custom code (a very hacky code which compiles extra libraries).

PyStan3 uses stan-math makefile to compile needed libraries.

Issue #2334, opened 2021-01-25
Add boost::odeint's Cash-Karp integrator


yizhang-yiz:
## Description
This issue is spawned out of the benchmark in #2326 , where additional integrators from odeint & arkode was evaluated. In particular, `boost::odeint`'s Cash-Karp method(`ckrk`) that has better stability shows to be an potential addition to our integrators.

## Example
WIP branch: `cash_karp_ckrk`.

- further benchmark based  on runing 4-chain sampling using 10 random seeds in each model.
####  lokta-volterror
In this model `ckrk` has 4 chains "stuck", indicated by much greater # of leapfrogs. Plots below excludes those chains.
![summary_lv_model](https://user-images.githubusercontent.com/33104304/105651750-8bced900-5e6c-11eb-9608-5523d2e9d575.png)
#### SIR
![summary_sir_model](https://user-images.githubusercontent.com/33104304/105651760-8d989c80-5e6c-11eb-88ed-c43987e9d259.png)
#### mildly-stiff Van del Pol oscillator
![summary_vdp_model](https://user-images.githubusercontent.com/33104304/105651762-8ec9c980-5e6c-11eb-8cc9-7c7ccbf3d737.png)

- describe the expected behavior if the feature exists
Same as `rk45`: `ode_ckrk(...)` with implementation based on boost::odeint.

## Expected Output
Same integrator output but potentially more efficient when parameter sampled causes the system to be stability-constrained.

#### Current Version:
v3.4.0

@bbbales2 @wds15

bbbales2:
That seems better. What is different between the two? Is this a choice that someone could make or is it always guess and check?

yizhang-yiz:
@bbbales2 
`ckrk` has a greater stability region than `rk45` so it's more stable, and it's designed for problems with large derivatives and/or rapid oscillations. So it suits better for ODEs that could be sampled into mild stiffness(the SIR & Van del Pol example above). For non-stiff problem it could cause slow down by trying too hard on parameters that could be right-out rejected(the Lotka-Volterra example).

Issue #311, opened 2016-06-29
Multiple definitions when including Stan headers in several source files


rgiordan:
#### Summary:
Stan headers cause multiple definition errors when included in several source files that are linked together in a library.

#### Description:
If the Stan headers are included in several source files that are compiled separately and then linked, the compiler fails with multiple definition errors.  I am a C++ novice, but I believe this is because non-inline functions are being instantiated in the header files (e.g. ```print_mat_size```).

#### Reproducible Steps:
A minimal example to reproduce can be found in [this repo](https://github.com/rgiordan/StanHeaderExample).  In short, it has two `cpp` files that each instantiate templates using Stan variables.  They can be successfully compiled into object files, but the linker fails with multiple definition errors.

```
// source1.hpp:
# ifndef SOURCE1_H
# define SOURCE1_H

#include <boost/math/special_functions/digamma.hpp>
#include <boost/math/special_functions/gamma.hpp>

using boost::math::lgamma;
using boost::math::digamma;

# include <stan/math.hpp>
# include "stan/math/fwd/scal.hpp"

using var = stan::math::var;
using fvar = stan::math::fvar<var>;

template <typename T> T multivariate_lgamma(T x, int p) {
  T result = log(M_PI) * p * (p - 1.0) / 4.0;
  for (int i = 1; i <= p; i++) {
    result += lgamma(x + 0.5 * (1 - (double)i));
  }
  return result;
}

extern template double multivariate_lgamma(double x, int p);
extern template var multivariate_lgamma(var x, int p);
extern template fvar multivariate_lgamma(fvar x, int p);

# endif

// source1.cpp:
# include "source1.hpp"
# include <stan/math.hpp>
# include <stan/math/fwd/scal.hpp>

using var = stan::math::var;
using fvar = stan::math::fvar<var>;

template double multivariate_lgamma(double x, int p);
template var multivariate_lgamma(var x, int p);
template fvar multivariate_lgamma(fvar x, int p);


// source1.hpp:
# ifndef SOURCE2_H
# define SOURCE2_H

#include <boost/math/special_functions/digamma.hpp>
#include <boost/math/special_functions/gamma.hpp>

using boost::math::lgamma;
using boost::math::digamma;

# include <stan/math.hpp>
# include "stan/math/fwd/scal.hpp"

using var = stan::math::var;
using fvar = stan::math::fvar<var>;

template <typename T> T multivariate_digamma(T x, int p) {
  T result = 0.0;
[...]

PR #2336, opened 2021-01-27
Cash karp ckrk

yizhang-yiz:
## Summary

This PR address #2334 by adding Cash-Karp integrator function `ode_ckrk` & `ode_ckrk_tol`, following existing naming pattern. The new integrator is shown in #2334 to be able to improve numerical ode solution performance to problems that could become semi stiff under certain parameter samples.

## Tests

```bash
./rev/functor/ode_ckrk_tol_rev_test.cpp
./rev/functor/ode_ckrk_grad_test.cpp
./rev/functor/ode_ckrk_rev_test.cpp
./prim/functor/ode_ckrk_prim_test.cpp
./prim/functor/ode_ckrk_tol_prim_test.cpp
```

## Side Effects

n/a

## Release notes

Adding Cash-Karp numerical integrator to improve numerical integration of ODEs with semi-stiffness and/or rapid oscillations.

## Checklist

- [x] Math issue #2334

    The copyright holder is Metrum Research Group. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:


test/unit/math/prim/functor/ode_ckrk_prim_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include <boost/numeric/odeint.hpp>
+#include <test/unit/math/prim/functor/harmonic_oscillator.hpp>
+#include <test/unit/math/prim/functor/lorenz.hpp>
+#include <test/unit/util.hpp>
+#include <iostream>
+#include <sstream>
+#include <vector>
+#include <limits>
+
+std::ostream* msgs = nullptr;
```
rok-cesnovar, line 13: We want to avoid using global variables in tests as we use the so-called jumbo tests in Jenkins. This means that all 
prim/functor tests are built as a single file, which speeds up testing by a factor.

If global variables cant be avoided, put them in a unique namespace like ode_ckrk_test::.

rev/functor does not use jumbo tests (https://github.com/stan-dev/math/blob/develop/runTests.py#L37) so its not an issue there.

yizhang-yiz: thx @rok-cesnovar . fixed it.

bbbales2:
> The new integrator is shown in #2334 to be able to improve numerical ode solution performance

I think the merge conditions on this are the same as the vector absolute tolerances thing.

It is good we have more options cause then people can solve more problems.

But we really need a guide to use ODEs that makes it obvious the steps of picking one thing or another (and presumably the need for this solver will be justified by that process):

1. If you know nothing else about your problem, pick Cash Karp
2. If you know X about your problem (or can test it with this technique), switch to RK45
3. If you aren't sure about stiffness, profile between RK45, BDF, and Adams, ...

Maybe the first step of any ODE inference problem is simulated data or something, or prior-only simulations, or something like this.

yizhang-yiz:
> But we really need a guide to use ODEs that makes it obvious the steps of picking one thing or another (and presumably the need for this solver will be justified by that process):

@bbbales2 without knowing anything, I'd say choose rk45. All other choices should be chosen based on some knowledge of the system. Among the three models in #2334, SIR & Van del Pol oscillator are known to become stiff in certain occasion, and predator-prey is usually considered non-stiff. Which one to choose is also tied to how parameters are sampled, as you pointed out. I'd say this "which one to choose" is more complicated than vector atol issue.

wds15:
We should probably come up with a template based on profiling. We can make the bdf solver the one driving the inference, but let all the others also return the solutions. Each solver is run in its own profile block and then we will know quickly with a few samples which one wins during the wild warmup phase... and that‘s your solver most likely (a „data“ driven approach).

bbbales2:
> The reason is that whenever a solver throws we then wont get the result of the others...

Yeah good point. And maybe with this in mind running the ODE with priors only in the generated quantities block is step 1.

Once everything is running, move to the model block and that will mean that the sensitivity equations get added (and maybe everything gets changed).

It sounds really primitive and like a lot of work but guess and check moving between solvers will be a lot of work too.

yizhang-yiz:
The "data-based" approach might be tricky when we don't know winning criteria: faster is not necessarily better. If `rk45` can be faster by failing early throwing out `nan` of `inf` results but it could miss samples that matters. In general testing the solver in different blocks & with tentative prior makes more sense to me.

bbbales2:
> In general testing the solver in different blocks & with tentative prior makes more sense to me.

Yeah if we can get the ODE 20 questions written down it'll make it a lot easier to add complexity.

yizhang-yiz:
@bbbales2 do we want to continue the discussion or move on to reviewing? I'm fine either way.

bbbales2:
Can we do the docs? If we have text that makes it easy to pick between the six solvers we have, that'll make everything much easier to understand.

Then when it comes to justifying this, it's just adding if statements to the doc that tell someone how to pick it. And if it comes down to profiling rk45 vs. ckrk on every problem, that's okay.

As far as code, we should try to do a little work to compress all the ODE tests -- there's a ton of them now and it could probably be a little more maintainable. But that can wait.

yizhang-yiz:
> Can we do the docs? If we have text that makes it easy to pick between the six solvers we have, that'll make everything much easier to understand.

I agree we can work on docs first but I'm not sure how to come up a guideline without scope creeping. Choosing integrator requires some understanding of the problems just under deterministic context, and in Stan context the topic becomes even  fuzzier.

> As far as code, we should try to do a little work to compress all the ODE tests -- there's a ton of them now and it could probably be a little more maintainable. But that can wait.

I've started doing this in this PR by consolidating finite diff test so one can apply it to multiple ODE system & integrators relatively easily(it's used to test `ckrk` on two systems). The work can continue after this.

bbbales2:
> without scope creeping

Well we have 6 integrator functions, two of which are kinda speculative (the adams solvers). Including the two functions here and the adjoint ODE function, we're looking down the barrel of 9 ODE functions. So I think we're scope creeping already on the coding front and the docs need to catch up.

We need to be justifying these things cause they are not simple. I forget the differences in all these functions all the time, and I've written a fair share of the code (and dug in basically all of it). The most important thing we can do for current and future devs and users is explain the role of each of the solvers and why they are here. Looking at #2334 my thought is we should be deprecating rk45 and rolling with this (even with https://github.com/stan-dev/math/pull/2336#issuecomment-768552368 -- what isn't stiff occasionally? -- ckrk seems more robust if this is the case). I kinda feel the same about the adams solver too.

Our current docs are assuming that people come to Stan knowing a lot about their differential equations and how to solve them. I don't really believe this. People come to Stan with their ODEs because they want to do their science thing, and in the process they end up having to learn about solving their ODEs. We gotta explain how to use the functionality in Stan to learn about solving their ODEs.

> in Stan context the topic becomes even fuzzier.

Yeah, and this is why more docs are important.

Compare the current ODE landing page: https://mc-stan.org/docs/2_26/stan-users-guide/ode-solver-chapter.html

To section 3.5: https://mc-stan.org/docs/2_26/stan-users-guide/stiff-ode-section.html

Section 3.5 has really useful stuff like:
- Stiffness is a numerical phenomena that causes some differential equation solvers difficulty ... The phenomena is common in chemical reaction systems
- If it is not known for sure that an ODE system is stiff, run the model with both the rk45 and bdf solvers and see which is faster.

Here's some things in general:
- How to interpret chains not mixing
- How to pick between rk45/ckrk/adams/bdf
- How to know when to change max_num_steps
- How to know when to change tolerances
- Higher tolerances are not always slower (very surprising!)
- How to handle solutions that go near zero
- Scalability of sensitivity equations
- Why sometimes people end up writing trapazoid method

I would rather review stuff like this than add more ODE code honestly.

bbbales2:
> I would rather review stuff like this

To clarify, to review this I have to figure out how this function fits in the big picture of everything ODEs, and I'd rather have that written down somewhere then just inevitably forgetting it a week from now.

wds15:
> Our current docs are assuming that people come to Stan knowing a lot about their differential equations and how to solve them. I don't really believe this. People come to Stan with their ODEs because they want to do their science thing, and in the process they end up having to learn about solving their ODEs. We gotta explain how to use the functionality in Stan to learn about solving their ODEs.

Hmm... I don't mind the perspective that Stan is a thing solving science problems. It's just that ODEs are tricky and we can't take that away from the user. So a user needs to understand something about ODEs is what I would think is reasonable here. That is not to say we should not document anything nor guide anything here. It's just that we cannot solve someones problem without having seen it. Still, a big picture thing doc makes sense as long as we still say that specifics can overrule the big picture thing possibly (and some things will remain empirical - meaning "you need to try").

@yizhang-yiz I was applying this solver to the linked mass toy example from the adjoint solver branch. This system is non-stiff, but can easily become stiff if wild parameters are being sampled - at least that was my intention here. It turns out that RK45 and ckrk basically do the same in terms of speed and result while bdf blows them away with a 2x faster inference time. From this example it looks as if ckrk is not adding any value here.

I think it would be nice to find an end-to-end example where ckrk is the clear choice. The argument that the solver is better for mildly stiff problems is a good one, but ideally we also demonstrate that in an example. At least this is what I would think would be good here. It's just that we want to avoid unnecessary feature creep here.

Like I said before, from my intuition this solver should show benefits as it can survive mild stiffness, but it would still be very convincing to see an actual model run significantly better (faster/more stable/whatever metric is adequate).

My intuition for the big picture guide would be: Use a non-stiff solver if it's not stiff - if that is not known, then just try bdf and RK45. In case mild stiffness is something which can occur for a non-stiff problem (as it happens with wide priors), then ckrk is a possible alternative to RK45. Adding ckrk to the picture here adds complexity which we should justify by observing it's advantages end-to-end.

bbbales2:
Oh on the docs front, I was talking to @charlesm93 today and I told him I'd try to start writing some of this stuff down later today. He's interested in workflow stuff. It's kinda unfair for me to ask Yi to read my mind on this -- maybe if I get some example text written it'll be clearer.

> Use a non-stiff solver if it's not stiff - if that is not known, then just try bdf and RK45

This is the vague template lol.

wds15:
> This is the vague template lol.

Well, finding out emprically is the simplest way. Of course, we should write about stiff is "involves multiple length/time-scales", but this is where it often ends up being useful to me. It's really about bdf takes 5s and RK45 10s for some toy setup... problem solved! At least this is how I see it. Given that function signatures are anyway basically the same it is easy to try that out quickly.

yizhang-yiz:
@bbbales2 
> So I think we're scope creeping already on the coding front and the docs need to catch up.

By scope creep I didn't mean getting more & more integrator but initiate a large benchmark & research project from documentation. There are a lot of things conflated here. Specifically

> How to interpret chains not mixing

This is not just an ODE issue, why should it be discussed here?

> How to pick between rk45/ckrk/adams/bdf

In deterministic context, this is not too hard to answer. In sampling context, I don' know any guideline that is backup up by large-scale benchmark and/or credible literature. There's a lot of work to do before we can say anything in our doc. This is what I mean when saying scope creep. 

> How to know when to change max_num_steps

This is problem dependent. 

> How to know when to change tolerances

This is addressed in my previous doc PR.

> Scalability of sensitivity equations

I don't understand what this means.

> Why sometimes people end up writing trapazoid method

Is this even a doc issue?

So from your list I can only see the question of "How to pick between rk45/ckrk/adams/bdf" becomes harder to answer when is ckrk is added, which is the focus of our previous discussion. Some from that list may not even be a doc issue. 

@wds15 
> Hmm... I don't mind the perspective that Stan is a thing solving science problems. It's just that ODEs are tricky and we can't take that away from the user. So a user needs to understand something about ODEs is what I would think is reasonable here. That is not to say we should not document anything nor guide anything here. It's just that we cannot solve someones problem without having seen it. 

Exactly.

> I think it would be nice to find an end-to-end example where ckrk is the clear choice. The argument that the solver is better for mildly stiff problems is a good one, but ideally we also demonstrate that in an example. At least this is what I would think would be good here. It's just that we want to avoid unnecessary feature creep here.

In general it's not too hard to come up a non-stiff model that explicit method(like ckrk) could outperform implicit(like bdf), as if the system dimension is large enough BDF's advantage of stepsize can be eaten by nonlinear iteration. For example we can continue the lotka-volterra where ckrk is shown could be better than rk45, here's a modified "enlarged" version(in the original version they perform similarly).
```stan
functions {
  real[] dz_dt(real t,       // time
               real[] z,     // system state {prey, predator}
               real[] theta, // parameters
               real[] x_r,   // unused data
               int[] x_i) {
    real u = z[1];
    real v = z[2];

    real alpha = theta[1];
    real beta = theta[2];
    real gamma = theta[3];
    real delta = theta[4];

    real du_dt = (alpha - beta * v) * u;
    real dv_dt = (-gamma + delta * u) * v;

    return { du_dt, dv_dt,
                 du_dt, dv_dt,
                 du_dt, dv_dt,
                 du_dt, dv_dt,
                 du_dt, dv_dt };
  }
}
data {
  real rtol;
  real atol;
  int<lower = 0> N;          // number of measurement times
  real ts[N];                // measurement times > 0
  real y_init[2];            // initial measured populations
  real<lower = 0> y[N, 2];   // measured populations
}
parameters {
  real<lower = 0> theta[4];   // { alpha, beta, gamma, delta }
  real<lower = 0> z_init[2];  // initial population
  real<lower = 0> sigma[2];   // measurement errors
}
transformed parameters {
  real z[N, 10];
  real z_init_max[10];
  for (i in 1:5) {
  z_init_max[(i-1) * 2 + 1] = z_init[1];
  z_init_max[(i-1) * 2 + 2] = z_init[2];
}
    z = integrate_ode_xxx(dz_dt, z_init_max, 0, ts, theta, //ckrk surrogate 
                          rep_array(0.0, 0), rep_array(0, 0),
                          rtol, atol, 5e2);    
}
model {
  theta[{1, 3}] ~ normal(1, 0.5);
  theta[{2, 4}] ~ normal(0.05, 0.05);
  sigma ~ lognormal(-1, 1);
  z_init ~ lognormal(log(10), 1);
  for (k in 1:2) {
    y_init[k] ~ lognormal(log(z_init[k]), sigma[k]);
    y[ , k] ~ lognormal(log(z[, k]), sigma[k]);
  }
}
```
ckrk:
```bash
Chain 1 finished in 323.5 seconds.
```
bdf:
```
Chain 1 finished in 385.6 seconds.
```

wds15:
And how does rk45 do here?

yizhang-yiz:
> And how does rk45 do here?

See linked issue. Specifically,
https://user-images.githubusercontent.com/33104304/105651750-8bced900-5e6c-11eb-9608-5523d2e9d575.png

bbbales2:
Alright I started trying to write docs: https://github.com/stan-dev/docs/pull/333

I tried to start answering the questions you brought up there. We need to get this stuff written down. I feel like there's a ton of text about picking tolerances and performance of ODEs and stuff from that last thread.

yizhang-yiz:
@bbbales2 
Now I'm confused. We do have an ongoing discussion of how to pick integrators among rk45/adams/bdf and possible ckrk. But I don't think I've brought up discussion of things like max step & tolerance etc. In addition, isn't picking tolerance addressed in https://github.com/stan-dev/docs/pull/311 , reviewed by you?

bbbales2:
Want to do a hangouts? I'm free other than 1:30 to 3 EST tomorrow.

#311 was an improvement, but there are other things I think we know something about:

1. When should a user consider changing the defaults?
2. Should a user pick the solver first or the tolerances first?
3. What happens when max_num_steps is exceeded (https://github.com/stan-dev/docs/issues/283)?

There was discussion of something like scales of noise or something in the last thread, and also lots of discussion about the not-simple relationship between tolerance and how long it takes a solve to run.

> how to pick integrators among rk45/adams/bdf and possible ckrk

Yeah this is good. I didn't get that far in the writeup. Feel free to add. I believe what we have discussed so far is:

1. Start with rk45
2. If that is slow, try BDF

I don't know when to go to Adams, or how to know if rk45 is slow other than just running rk45 and BDF together and seeing with profile like Sebastian said (https://github.com/stan-dev/math/pull/2336#issuecomment-771962398). And if that is the only reliable way, that is what we should recommend.

yizhang-yiz:
> Want to do a hangouts? I'm free other than 1:30 to 3 EST tomorrow.

How about 9 PST(noon EST)?

Now I see by "last thread" you probably mean #2244, which mostly focuses on the intended use of tolerance. I don't see how that's related to adding new integrator here. IMO the current thread is just about: a) does ckrk bring any benefit? and b) how do we help user choose it?

bbbales2:
> How about 9 PST(noon EST)?

That works. I sent a link.

> I don't see how that's related to adding new integrator here.

At what point does the user switch from worrying about solvers to worrying about tolerances? Or are the tolerances first?

wds15:
> At what point does the user switch from worrying about solvers to worrying about tolerances? Or are the tolerances first?

I think tolerances should be driven by what your problem needs. It's an independent choice of the solver in my view. It is determined by what you measure and what accuracy you need your solution. The somewhat more complicated question is what HMC needs as a precision to do it's job. This probably translates into leaning on the somewhat conservative side with tolerances.

test/unit/math/rev/functor/ode_ckrk_grad_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <iostream>
+#include <sstream>
+#include <vector>
+#include <stdexcept>
+
+using stan::math::ode_ckrk;
+using std::cos;
+using std::sin;
+
+double y1(double t, double omega, double chi) { return chi * cos(omega * t); }
```
bbbales2, line 13: Put all these functions in a namespace. The tests files here get concatenated together and if this gets grouped with `test/unit/math/rev/functor/integrate_ode_rk45_grad_test.cpp` it will cause problems.

test/unit/math/rev/functor/ode_test_util.hpp
```diff
@@ -0,0 +1,451 @@
...
+
+#include <stan/math/rev/core.hpp>
+#include <stan/math/prim.hpp>
+#include <test/unit/util.hpp>
+#include <gtest/gtest.h>
+#include <limits>
+#include <sstream>
+#include <vector>
+
+// calculates finite diffs for ode solver with varying parameters
```
bbbales2, line 12: Is there any chance of using this to replace `test/unit/math/rev/functor/util_rk45.hpp` and `test/unit/math/rev/functor/util_cvodes_bdf.hpp` in this pull?

bbbales2 (changes requested):
Review!

stan/math/prim/functor/ode_ckrk.hpp
```diff
@@ -0,0 +1,257 @@
...
+  };
+
+  // the coupled system creates the coupled initial state
+  std::vector<double> initial_coupled_state = coupled_system.initial_state();
+
+  const double step_size = 0.1;
+  try {
+    integrate_times(
+        make_controlled(absolute_tolerance, relative_tolerance,
+                        runge_kutta_cash_karp54<std::vector<double>, double,
```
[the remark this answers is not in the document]
yizhang-yiz: rk45 uses `make_dense`, while here ckrk uses `make_controlled` because it doesn't support dense output.

test/unit/math/rev/functor/ode_ckrk_rev_test.cpp
```diff
@@ -0,0 +1,567 @@
...
+  Eigen::VectorXd y0(2);
+  y0 << 1.0, 0.0;
+  std::vector<int> x_i;
+  std::vector<double> x_r;
+  test_ode(sol, f, 0.0, ts, y0, theta, x_r, x_i, 1.e-8, 1.e-4);
+}
+
+TEST(StanAgradRevOde_ode_ckrk, FN_oscillator_finite_diff) {
+  ckrk_integrator sol;
+  Field_Noyes_osc f;
```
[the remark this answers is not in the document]
yizhang-yiz: that is works as it should because rk45 isn't capable solve this model.

test/unit/math/rev/functor/ode_test_util.hpp
yizhang-yiz: yes that's why I did it this way. But IMO it should be part of a new PR dedicated to refactor ODE tests.

test/unit/math/prim/functor/ode_ckrk_prim_test.cpp
```diff
@@ -0,0 +1,359 @@
+#include <stan/math/prim.hpp>
```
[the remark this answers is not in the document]
yizhang-yiz: I'll add error tests. What do you mean by "value tests". Is `TEST(StanMathOde_ode_ckrk, harmonic_oscillator_value)` here what you mean?

yizhang-yiz:
@rok-cesnovar any idea why it didn't pass "lub_constrain_test.cpp" test?

rok-cesnovar:
I know why it failed: the mix tests can sometimes be too much for g++ 4.9.3, especially with all new stuff added like varmat and they sometimes need to be split in two or more files. I know, lol, but that mingw g++ 4.9.3 for you. Hopefully we will stop supporting it in a few months.

I am less sure why it failed here and not when it was added, its probably right at the limit of 4.9.3 and you were a bit unclucky: https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/fun/lub_constrain_test.cpp

So dont worry, will make a PR to split it and you can just merge it in. You could try restarting the tests, maybe you get luckier.

yizhang-yiz:
> You could try restarting the tests, maybe you get luckier.

I merged develop before started another test. So this is strike No. 2. I think I'd wait.

rok-cesnovar:
@yizhang-yiz merge develop in and you should be good. Sorry for the inconvenience.

yizhang-yiz:
@rok-cesnovar thanks, but looks like the compiler is still complaining.

rok-cesnovar:
Ah man, sorry. A do-over then, yay :)

yizhang-yiz:
@rok-cesnovar No worries. Just wondering, if by "stopping supporting it" you mean moving to RTools 4.0 on windows, what will be the hurdle for doing it now?

rok-cesnovar:
Yes, moving to RTools 4.0. 

The hurdle is essentially RStan, cause it still has to build with R 3.6 on CRAN (and for CRAN 3.6 windows tests they use RTools 3.5 obviously). It has to build until R 4.1 is release because at that point R-oldrel (the old release CRAN also tests with) will be 4.0.x and R-release will be 4.1.0. Right now R-release is 4.0.4 and R-oldrel is 3.6.X.

I am fairly confident this will happen before 2.27 is on CRAN (we still havent published 2.26), which would mean we can just do it now. But no guarantees (they dont really have a fixed schedule). 

We could also just decide rstan 2.27 will wait for 4.1.0 as a backup though. I would be very happy to do that. This is very annoying and frustrating. Rtools 3.5 just want built for this.

rok-cesnovar:
Our jenkins tests for Windows are already all on RTools 4.0, we just have this one on Github Actions on 3.5.

bbbales2:
@rok-cesnovar is it that the test/unit/math/mix/fun/lub_constrain_3_test.cpp just needs broken up more?

bbbales2:
If so I'll open pull

yizhang-yiz:
@bbbales2 I think it died on the 1st test:
https://github.com/stan-dev/math/pull/2336/checks?check_run_id=1954379971#step:10:1849

rok-cesnovar:
> @rok-cesnovar is it that the test/unit/math/mix/fun/lub_constrain_3_test.cpp just needs broken up more?

I think the first one need to be split a bit more. I should have run a few more tests not do a victory lap the first time it passed. I split in in two parts and then had to split the 2nd file further. Seems we need to do the same for the 1st one.

bbbales2:
I made a pull here: https://github.com/stan-dev/math/pull/2386

bbbales2:
I don't know why this is causing a problem in this branch though. Like apparently the split lub tests worked well enough to merge to master already -- why the problem here?

I'm going to merge the new lub splits into this branch directly and see if there's a different between this and https://github.com/stan-dev/math/pull/2386.

bbbales2 (dismissed):

bbbales2:
I'm approving this. I assume something will fail and we'll need to revert code but the basic branch is fine. So once the failing lub tests get sorted out this can go in.

bbbales2:
I'm gonna hop over to Windows and see if I can reproduce this error in the 3.5 mix/fun tests. It's not clear why the tests fail here but the same test passed here: https://github.com/stan-dev/math/pull/2391 (unless there is some other difference).

rok-cesnovar (approved):
Approving the changes that were done to fix our CI problems. Thanks @bbbales2 and sorry for the troubles @yizhang-yiz. 

Ben approved the rest of the changes already but will let him merge if he has any additional comments.

bbbales2:
I made a follow up issue for the testing cleanup: https://github.com/stan-dev/math/issues/2395

Issue #2300, opened 2021-01-09
Replace second-order finite diffs with finite diffs of autodiff gradients for testing


bob-carpenter:
## Description

We have second- and third-order autodiff functions defined in terms of finite differences.  This leads to poor accuracy for our tests.   An improvement would be to

* replace second-order finite difference Hessian definition with differences of autodiff gradients
* replace third-order finite difference tensor definition with differences of autodiff Hessians

As a side benefit, it'll be faster.

## Example

This is an implementation refactoring that should lead to higher accuracy.  It does not change the behavior of any functions other than to improve arithmetic accuracy.

## Expected Output

More precise output.

#### Current Version:
v3.4.0

adamhaber:
I'll try to tackle this.

adamhaber:
By the way, not sure how common this problem is, but when I tried following the hello world instructions and ran:
```
> make -j4 -f /path/to/stan-math/make/standalone math-libs
```
I got:
```
x86_64-apple-darwin13.4.0-ar: creating archive /Users/adamhaber/projects/stan_stuff/math/lib/sundials_5.6.1/lib/libsundials_nvecserial.a
/Users/adamhaber/miniconda3/bin/x86_64-apple-darwin13.4.0-ranlib: object: /Users/adamhaber/projects/stan_stuff/math/lib/sundials_5.6.1/lib/libsundials_nvecserial.a(nvector_serial.o) malformed object (unknown load command 1)
/Users/adamhaber/miniconda3/bin/x86_64-apple-darwin13.4.0-ar: internal ranlib command failed
make: *** [/Users/adamhaber/projects/stan_stuff/math/lib/sundials_5.6.1/lib/libsundials_nvecserial.a] Error 1
make: *** Waiting for unfinished jobs....
```
which, AFAICT, implies there was a mess between different clang/ar/ranlib installations on my laptop (might be related to [this]?(https://statmodeling.stat.columbia.edu/2019/12/11/macbook-pro-16-inch-2019-quick-review/)). 

What worked (eventually) was setting in `make/local`:
```
CXX=clang++
AR=llvm-ar
RANLIB=llvm-ranlib
```
is there some installation doc this can go into?

adamhaber:
Concerning the issue itself - just to make sure I understand (it's my first interaction with the stan-math codebase):

For the following files:
- `finite_diff_hessian.hpp` and `finite_diff_hessian_auto.hpp`
[...]

PR #2348, opened 2021-02-02
Second order finite diffs

adamhaber:
First stan-math and C++ PR, so I apologise in advance if (a) the PR format is slightly off or (b) the C++ is horrible.

In addition to the code, would be happy to get feedback on:
1. How should I name the files?
2. Can the previous files `finite_diff_[grad_]hessian[_auto].cpp` be removed? Can `finite_diff_hessian_helper` be removed?
3. I plan to add tests for zero size arrays and mismatched sizes. Any other cases I should cover?
4. Currently I'm relying on `finite_diff_stepsize` - does this makes, here?
 
## Summary

Replaces the finite difference approximation of the Hessian from one that is based on function calls to one that is based on gradients (see issue #2300).

Implementation is based on [this](https://v8doc.sas.com/sashtml/ormp/chap5/sect28.htm) reference, under "Central Difference Approximation", and "Second-order derivatives based on gradient".

## Tests

Added a test using a a simple polynomial function whose second derivatives are simples. 

## Side Effects

None.

## Release notes

Replaces the finite difference approximation of the Hessian from one that is based on function calls to one that is based on gradients.

## Checklist

- Math issue #2300 

- Copyright holder: Adam Haber (should I put something else, here?)

- the basic tests are passing

- the code is written in idiomatic C++ and changes are documented in the doxygen

- the new changes are tested


bob-carpenter:
> (a) the PR format is slightly off

Looks good as is.

> How should I name the files?

Given that we're going to eliminate the old implementations, you shouldn't need any new files here other than helper files.  

> Can the previous files finite_diff_[grad_]hessian[_auto].cpp be removed? Can finite_diff_hessian_helper be removed?

Yes.  In general stuff like that should go in a stan::math::internal namespace to indicate it's not part of our public API.

> I plan to add tests for zero size arrays and mismatched sizes. Any other cases I should cover?

Good boundary condition tests for all error conditions are important.  You want to check you throw the right kind of exception.

You probably also want to check a simple linear function (should all be zero), and something where everything's non-zero.

It'd be nice to get some sense that it has better error properties, but not technically necessary if the unit tests pass with the new version.  Those should be well known, anyway.

> Currently I'm relying on finite_diff_stepsize - does this makes, here?

Yes, that makes sense.  Whether the same stepsize makes sense here is an empirical question (though there may be theoretical guidance in the literature).  If the tests pass, you should be OK.  But what we'd really like to do is up the precision of our tests if this works.  That is, the goal is to use more digits of precision as a default and in cases where we've overloaded.  That doesn't all have to go in this PR.

stan/math/rev/functor/finite_diff_of_grads_hessian.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include <stan/math/rev/functor.hpp>
+#include <stan/math/prim/fun/finite_diff_stepsize.hpp>
+#include <stdexcept>
+
+namespace stan {
+namespace math {
+
+/**
+ * Calculate the value and the Hessian of the specified function at
+ * the specified argument using first-order finite difference of gradients,
```
bbbales2, line 16: The reference says it's 2nd order

bob-carpenter: Too many sense of order here.  Gradients are first order derivatives and Hessians second order, but I think @bbbales2 is right here in that it's a second-order finite-differencing scheme (meaning not just one other point being evaluated).  The reference doc says this:

> Central difference approximation ... Second-order derivatives based on gradient: 2n additional gradient calls are needed:

I think we can stick to using "central difference" rather than trying to quantify the order of finite differencing.  It'd be nice to indicate the number of gradient calls required.  Now that I see this, I realize it should also be more efficient than what we were doing before, which is great, because this and the third-order tests are slow as currently written.

adamhaber:
> Given that we're going to eliminate the old implementations, you shouldn't need any new files here other than helper files.

Not sure I understand. My current intention is delete the non-auto'd previous implementations, delete finite_diff_hessian_helper, and replace the auto'd implementation with this one. Makes sense?

> Good boundary condition tests for all error conditions are important. You want to check you throw the right kind of exception.

Can you refer me to similar existing tests I can start from?

> You probably also want to check a simple linear function (should all be zero), and something where everything's non-zero.

Added a linear function. What do you mean by "everything's non-zero."?

bob-carpenter:
> Not sure I understand. My current intention is delete the non-auto'd previous implementations, delete finite_diff_hessian_helper, and replace the auto'd implementation with this one. Makes sense?

Yes.   But that doesn't require new files, right?  Anyway, we haven't gotten too hung up on file naming as it's pretty buried from a user's perspective.

> > Good boundary condition tests for all error conditions are important. You want to check you throw the right kind of exception.

> Can you refer me to similar existing tests I can start from?  

I really just mean look at the edge cases like functions with zero arguments, one argument, and more arguments.  

> > You probably also want to check a simple linear function (should all be zero), and something where everything's non-zero.

> Added a linear function. What do you mean by "everything's non-zero."?

I mean entries of the Hessian.  With a linear function, they'll all be zero.  If you have exp(x1 + x2 + .... + xN) then every entry is nonzero, whereas if you have exp(x1) + exp(x2) + ... + exp(xN) then everything's zero but the diagonal.  Of course, we won't get strictly zero with the numerics.  And sometimes testing finite diffs is tricky around zero.  I really just mean try a variety of functions, at least some of which have non-zero Hessian entries everywhere.

adamhaber:
Hmm, @bob-carpenter / @bbbales2 can you see why 
`./runTests.py test/unit/math/mix/prob/std_normal_cdf_log_test.cpp `
and 
`./runTests.py test/unit/math/mix/prob/normal_cdf_log_test.cpp`
fail?

bbbales2:
Not sure. I'll check it out tomorrow tho.

bob-carpenter:
> can you see why
./runTests.py test/unit/math/mix/prob/std_normal_cdf_log_test.cpp 
and
./runTests.py test/unit/math/mix/prob/normal_cdf_log_test.cpp
fail?

Did you add any new tests?  One thing you can do is reduce the number of tests to figure out which one is failing (called "bisection" in the debugging world).

One thing to check is definitions at boundary conditions.

adamhaber:
> Did you add any new tests? One thing you can do is reduce the number of tests to figure out which one is failing (called "bisection" in the debugging world).

Yes, I've added `test/unit/math/rev/functor/finite_diff_of_grads_hessian_test.cpp`, deleted `test/unit/math/mix/functor/finite_diff_hessian_auto_test.cpp` and `test/unit/math/prim/functor/finite_diff_hessian_helper_test.cpp`, and modified `test/unit/math/mix/functor/finite_diff_test.cpp` and `test/unit/math/mix/prob/lkj_corr_test.cpp`.

> One thing to check is definitions at boundary conditions.

I've included a modified version of `test_hessian_finite_diff` from the tests of the previous implementation; AFAIU, this includes a test for zero size inputs, right?
```
  auto log_fun
       = [](const auto& x) { return stan::math::sum(stan::math::log(x)); };

   Eigen::VectorXd y(0);
   test_hessian_finite_diff("log_fun({})", log_fun, y);
```

bbbales2:
The test that is failing is the one at the bottom of https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/prob/normal_cdf_log_test.cpp#L20.

The comments in the code reference a piece-wise function definition, which is [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/normal_lcdf.hpp#L168).

It looks like it's the higher order derivatives of the normal_lcdf. It might be that finite difference is failing cause of the discontinuity, or it might be the autodiff having trouble. To figure this out, compute using finite diffs in R or Wolfram alpha or something the 1st, 2nd, and 3rd order derivatives of the log cdf of the normal at `y = 0, mu = 1, sigma = 0.1 * sqrt(2.0))` with respect to sigma (so this is one dimensional -- no need to take derivatives with respect to `y` or `mu`).

With those we can figure out if the autodiff or finite diff is at fault. If it's the finite diff, I'm not worried, we remove the test cause the function is funky here (and it looks like we were already avoiding some sort of problem with derivatives here).

(also if you push that test slightly away from the piecewise thing, either left or right, everything is fine)

adamhaber:
I tried to do something similar with jax (not sure this is exactly what you meant):
```python
from jax import grad
from jax.scipy import stats
import jax.numpy as jnp
from jax.config import config
config.update('jax_enable_x64', True)

f = lambda x:jnp.log(stats.norm.cdf(0, loc=1, scale=x))

x0 = 0.1*jnp.sqrt(2.)

print("JAX AD results:")
for i in range(4):
    print(f"{i}-th order: {f(x0)}")
    f = grad(f)
```
this gives:
```
JAX AD results:
0-th order: -27.894036726097372
1-th order: 360.36636636973
2-th order: -7551.517352158384
3-th order: 212844.64499421127
```
When I compare this to the finite difference approximation (2nd order), I get reasonable results below a certain epsilon:
```python
f = lambda x:jnp.log(stats.norm.cdf(0, loc=1, scale=x))

for scale in jnp.arange(-5,0):
    eps = 10.**scale
    print(f"Finite difference epsilon: 1e{scale}, value: {(f(x0+eps)-2*f(x0)+f(x0-eps))/(eps**2)}")
```
```
Finite difference epsilon: 1e-5, value: -7551.517526849238
Finite difference epsilon: 1e-4, value: -7551.52361513467
Finite difference epsilon: 1e-3, value: -7552.143645202847
Finite difference epsilon: 1e-2, value: -7614.583024961092
Finite difference epsilon: 1e-1, value: -25070.823742293822
```
so maybe it's a step-size issue? 
Also, running this with `config.update('jax_enable_x64', False)` gives rather different results, perhaps this is also suggest something related to numerical stability?

bbbales2:
Yeah that's it, but I lied and gave you the wrong function.

It was the log cdf with mu = 0.0, sigma = 1.0, and we're taking derivatives with respect to the x argument.

So as cpp code that is runnable in the tests:
```cpp
TEST(whatever5, whatever6) {
  auto f = [](const double mu, const double sigma) {
    return [=](const auto& y) { return stan::math::normal_lcdf(y, mu, sigma); };
  };

  Eigen::VectorXd x(1);
  x << 0.1 * stan::math::SQRT_TWO;
  double fx;
  Eigen::VectorXd grad_fx;
  Eigen::MatrixXd hess_fx;
  stan::math::internal::finite_diff_of_grads_hessian(f(0.0, 1.0), x, fx, grad_fx, hess_fx);

  std::cout << "fx : " << fx << std::endl;
  std::cout << "grad_fx : " << grad_fx << std::endl;
  std::cout << "hess_fx : " << hess_fx << std::endl;
}
```

Gives:
```
fx : -0.586571
grad_fx : 0.710085
hess_fx : -0.334997
```

And reference finite diffs in R:
```
myfunc = function(x) {
  pnorm(x, 0.0, sd = 1.0, log.p = TRUE)
}

deriv = function(x) {
  (myfunc(x + dx) - myfunc(x - dx)) / (2 * dx)
}

second_deriv = function(x) {
  (myfunc(x + dx) - 2 * myfunc(x) + myfunc(x - dx)) / (dx^2)
}

dx = 1e-5
x = 0.1 * sqrt(2.0)

myfunc(x)
deriv(x)
second_deriv(x)
```

Output is:
```
> myfunc(x)
[1] -0.5865708
> deriv(x)
[1] 0.710087
> second_deriv(x)
[1] -0.6046452
```

So there's definitely a difference between the two.

This made me think maybe there was a discontinuity in the gradients so I tested the output at a couple places:

```
// x << 0.1 * stan::math::SQRT_TWO + 0.0001
fx : -0.5865
grad_fx : 0.710028
hess_fx : -0.604642

// x << 0.1 * stan::math::SQRT_TWO - 0.0001
fx : -0.586642
grad_fx : 0.710146
hess_fx : -0.604693
```

And then computing the gradients from that gives `(0.710028 - 0.710146) / (2 * 0.001) = -0.59`. Which seems pretty close.

So I'm not sure what is happening there -- probably worth adding a bunch of `std::cout` statements inside `finite_diff_of_grads_hessian` and see what is happening.

adamhaber:
It seems like `finite_diff_stepsize(x(i))` chooses an `epsilon` which is too small for this test. When I replace this with 
```
epsilons[i] = 0.001; //finite_diff_stepsize(x(i));
```
I get 
```
fx : -0.586571
grad_fx : 0.710085
hess_fx : -0.603034
``` 
which is better.

The original [reference](https://v8doc.sas.com/sashtml/ormp/chap5/sect28.htm) has this part:
image
so, would it make sense to use 
epsilon_i = pow(machine_precision, 1/3)*(1. + abs(x_i))?
If so - how do I get machine_precision in Stan Math?

rok-cesnovar:
We have stan::math::machine_precision: https://github.com/stan-dev/math/blob/e684aa6ec761aac129ef05741262dfcc7d9519c3/stan/math/prim/fun/constants.hpp#L207

adamhaber:
I'm pretty sure this is a how-to-pick-the-stepsize issue. The original `finite_diff_stepsize` returned
`std::cbrt(EPSILON) * std::fmax(1, fabs(u))`.
The reference I've used suggests:
`std::cbrt(EPSILON) * (1 + fabs(u))`
when the precision eta isn't defined, or 
`std::cbrt(eta) * (1 + fabs(u))`
when it is.

I've experimented with different values and is seems like 
`0.0001 * (1 + stan::math::abs(x(i)))` gives generally good results - for example, the hessian for the std_normal_cdf_log_test is -0.59 instead of the -0.33 it was before (should be -0.60). Smaller values than 0.0001 give worse results on the [std_]normal_cdf_log_test; larger values fail some of the finite_diff_of_grads_hessian tests.

bob-carpenter:
> I'm pretty sure this is a how-to-pick-the-stepsize issue. 

If tests path with better stepsize choice, then definitely.

Any idea how the gradient finite diff compares to the double finite diff version in terms of precision?

adamhaber:
> If tests path with better stepsize choice, then definitely.

The problem seems to be that I can't find a step size that would satisfy all tests...

> Any idea how the gradient finite diff compares to the double finite diff version in terms of precision?

You mean the "second-order derivatives based on function calls" vs "based on gradient calls"?

bbbales2:
> The problem seems to be that I can't find a step size that would satisfy all tests...

Oof, yeah, the bright side is tests tolerances as they are have been adjusted in places so that the current finite difference work. So we can definitely re-adjust them, it'll just be tedious.

My Q is the same as Bob's -- I wonder if going to centered finite differences will make us less vulnerable to loss of precision stuff?

If that doesn't work, then there's a way to run all the tests and ignore the failures, and we can do that to see which functions are failing and go in and investigate them. I'm happy to help with that bit if we get there.

adamhaber:
> My Q is the same as Bob's -- I wonder if going to centered finite differences will make us less vulnerable to loss of precision stuff?

To test this, I changed `test_hessian_finite_diff` to compute both the previous version (finite diffs of function calls -> hessian) and the new version (finite diffs of gradients -> hessian). For each function and at each point, I computed the difference of the AD hessian and the prev/new finite diff hessian, as well as their ratio (using `hess_fx_ad.cwiseQuotient(hess_fx)`, hope it's the right way).
```
Function: norm_fun({1, 2, 3})
At point: 
  1
  2
1.5

Difference between AD hessian and previous implementation:
 1.69054e-06 -2.36337e-06  2.35573e-07
-2.36337e-06  -6.3834e-06  1.19419e-06
 2.35573e-07  1.19419e-06  3.25483e-06

Difference between AD hessian and new implementation:
 2.82913e-12 -5.37348e-13  2.36343e-11
-5.37348e-13 -1.75443e-12 -2.13425e-11
 2.36343e-11 -2.13425e-11   5.4418e-11

Ratio between AD hessian and previous implementation:
0.999996 0.999995        1
0.999995  1.00001        1
       1        1 0.999978

Ratio between AD hessian and new implementation:
1 1 1
1 1 1
1 1 1
**********************************************************

Function: norm_fun({1, 2, 3})
At point: 
  1e+10
  2e+10
1.5e+10

Difference between AD hessian and previous implementation:
  7.4356e-25 -1.51471e-25  2.25406e-26
-1.51471e-25 -2.25313e-25  7.16554e-26
 2.25406e-26  7.16554e-26 -7.56904e-25

Difference between AD hessian and new implementation:
-3.35819e-32   4.6922e-32  1.77556e-31
  4.6922e-32  -6.0262e-32 -1.90895e-31
 1.77555e-31 -1.90894e-31  5.40189e-31

Ratio between AD hessian and previous implementation:
0.999833 0.999966 0.999996
0.999966  1.00005  1.00001
0.999996  1.00001  1.00051

Ratio between AD hessian and new implementation:
1 1 1
1 1 1
1 1 1
**********************************************************

Function: norm_fun({1, 2, 3})
At point: 
  1e+20
  2e+20
1.5e+20

Difference between AD hessian and previous implementation:
 2.03539e-44  7.63575e-45 -1.37694e-44
 7.63575e-45   7.4356e-45 -4.94537e-46
-1.37694e-44 -4.94537e-46  2.11383e-44

Difference between AD hessian and new implementation:
 1.56572e-52 -3.11869e-52  2.69352e-51
-3.11869e-52  4.67166e-52 -2.84883e-51
 2.69352e-51 -2.84883e-51  5.02171e-51

Ratio between AD hessian and previous implementation:
0.999542  1.00017  1.00023
 1.00017 0.999833 0.999992
 1.00023 0.999992 0.998575

Ratio between AD hessian and new implementation:
1 1 1
1 1 1
1 1 1
**********************************************************

Function: norm_fun({1, 2, 3})
At point: 
  1e+30
  2e+30
1.5e+30

Difference between AD hessian and previous implementation:
4.61905e-64 7.63575e-65 1.20672e-64
7.63575e-65  7.4356e-65 4.61896e-65
1.20672e-64 4.61896e-65   6.707e-64

Difference between AD hessian and new implementation:
 9.06107e-72 -5.44528e-72  2.83525e-71
-5.44528e-72  1.82949e-72 -2.92565e-71
 2.83526e-71 -2.92566e-71  5.60521e-71

Ratio between AD hessian and previous implementation:
0.998962  1.00017 0.999796
 1.00017 0.999833  1.00008
0.999796  1.00008 0.995493

Ratio between AD hessian and new implementation:
1 1 1
1 1 1
1 1 1
**********************************************************

Function: norm_fun({1, 2, 3})
At point: 
 1e-10
-1e-10
     1

Difference between AD hessian and previous implementation:
-1.03257e-05  5.69661e-07        4e-10
 5.69661e-07 -2.25173e-06       -4e-10
       4e-10       -4e-10 -5.82221e-06

Difference between AD hessian and new implementation:
           0            0 -1.67691e-12
           0            0  1.67691e-12
-1.67691e-12  1.67691e-12 -3.27209e-11

Ratio between AD hessian and previous implementation:
 1.00001        1      inf
       1        1     -inf
     inf     -inf 0.999994

Ratio between AD hessian and new implementation:
       1        1 0.995825
       1        1 0.995825
0.995825 0.995825        1
**********************************************************

Function: norm_fun({1, 2, 3})
At point: 
 1e-20
-1e-20
     1

Difference between AD hessian and previous implementation:
-2.25173e-06  2.33247e-07        4e-20
 2.33247e-07 -2.25173e-06       -4e-20
       4e-20       -4e-20 -5.82221e-06

Difference between AD hessian and new implementation:
           0            0        2e-20
           0            0       -2e-20
       2e-20       -2e-20 -3.27209e-11

Ratio between AD hessian and previous implementation:
       1        1      inf
       1        1     -inf
     inf     -inf 0.999994

Ratio between AD hessian and new implementation:
1 1 2
1 1 2
2 2 1
**********************************************************

Function: norm_fun({1, 2, 3})
At point: 
 1e-30
-1e-30
     1

Difference between AD hessian and previous implementation:
-2.25173e-06  2.33247e-07        4e-30
 2.33247e-07 -2.25173e-06       -4e-30
       4e-30       -4e-30 -5.82221e-06

Difference between AD hessian and new implementation:
           0            0        2e-30
           0            0       -2e-30
       2e-30       -2e-30 -3.27209e-11

Ratio between AD hessian and previous implementation:
       1        1      inf
       1        1     -inf
     inf     -inf 0.999994

Ratio between AD hessian and new implementation:
1 1 2
1 1 2
2 2 1
**********************************************************

Function: log_fun({})
At point: 


Difference between AD hessian and previous implementation:


Difference between AD hessian and new implementation:


Ratio between AD hessian and previous implementation:


Ratio between AD hessian and new implementation:

**********************************************************

Function: log_fun({2})
At point: 
2

Difference between AD hessian and previous implementation:
4.46309e-07

Difference between AD hessian and new implementation:
8.18023e-12

Ratio between AD hessian and previous implementation:
0.999998

Ratio between AD hessian and new implementation:
1
**********************************************************

Function: log_fun({1, 2, 3, 4, 5})
At point: 
1
2
3
4
5

Difference between AD hessian and previous implementation:
-5.87693e-05            0            0 -1.34566e-06            0
           0 -1.46923e-05            0            0 -3.36414e-07
           0            0 -6.52992e-06  1.49517e-07            0
-1.34566e-06            0  1.49517e-07  -1.6546e-06  1.17745e-06
           0 -3.36414e-07            0  1.17745e-06  2.32888e-07

Difference between AD hessian and new implementation:
3.27209e-11           0           0           0           0
          0 8.18023e-12           0           0           0
          0           0 4.14495e-12           0           0
          0           0           0 2.04506e-12           0
          0           0           0           0 1.30884e-12

Ratio between AD hessian and previous implementation:
 1.00006      nan      nan        0      nan
     nan  1.00006      nan      nan        0
     nan      nan  1.00006       -0      nan
       0      nan       -0  1.00003       -0
     nan        0      nan       -0 0.999994

Ratio between AD hessian and new implementation:
  1 nan nan nan nan
nan   1 nan nan nan
nan nan   1 nan nan
nan nan nan   1 nan
nan nan nan nan   1
**********************************************************
```

So I think we're definitely gaining precision - whether it's "a lot" or not it's hard for me to judge.

What do you think?

bbbales2:
Yeah, this makes sense.

I also tried out 1st order finite diffs of gradients and not doing the symmetric 2nd order thing (just one centered diff instead of two). Neither thing helped.

Anyway, I think the simple thing here is to just remove this test. The finite differences are failing, not the autodiff.

We could up the tolerances on the test here to ignore the failures, but I'm not sure this is really such a problem with `normal_lcdf`. If anything we should move it to the finite differences tests so we know where the finite differences are having problems.

Was this error output the result of running all the unit tests? Normally tests stop with an error. You can get them all to run and save the output to a file with:

```
./runTests.py --run-all -j8 test/unit &> out.txt
```

This way we can see if there are any other problems. These will probably take a few hours to run. I've got a not-busy computer to run these if that's more convenient. The alternate to this is one by one updating functions and seeing if Jenkins passes, but that can be a really slow way to iterate on code (especially for a pull like this where we're probably done writing the new library code and it's just a matter of updating tests).

bob-carpenter:
What do I think?  This looks great!   It's an order of 10^5 or so improvement in precision.  And it'll be a lot faster.  

I'm now wondering how RStan computes the Hessian---I think it's doing something similar.  Maybe we could unify the implementations.  @bgoodri will know how the RStan part works.  This is good enough we might also be able to use it for second-order autodiff of functions where don't have second-order autodiff---primarily all the ODE and equation solvers.

adamhaber:
It seems like 25 tests are failing:
```
[  FAILED  ] AgradFiniteDiff.hessian (0 ms)
[  FAILED  ] AgradFiniteDiff.hessian
[  FAILED  ] mathMixScalFun.std_normal_lcdf (0 ms)
[  FAILED  ] mathMixScalFun.std_normal_lcdf
[  FAILED  ] mathMixScalFun.normal_lcdf (2 ms)
[  FAILED  ] mathMixScalFun.normal_lcdf
[  FAILED  ] mathMixMatFun.logit (912 ms)
[  FAILED  ] mathMixMatFun.logit
[  FAILED  ] mathMixMatFun.trigamma (2436 ms)
[  FAILED  ] mathMixMatFun.trigamma
[  FAILED  ] mathMixMatFun.read_cov_L_lp (41 ms)
[  FAILED  ] mathMixMatFun.read_cov_L_lp
[  FAILED  ] mathMixMatFun.lambert_wm1 (563 ms)
[  FAILED  ] mathMixMatFun.lambert_wm1
[  FAILED  ] mathMixMatFun.log2 (1728 ms)
[  FAILED  ] mathMixMatFun.log2
[  FAILED  ] mathMixMatFun.log10 (1659 ms)
[  FAILED  ] mathMixMatFun.log10
[  FAILED  ] MathMixMatFun.matrixExp (561 ms)
[  FAILED  ] MathMixMatFun.matrixExp
[  FAILED  ] mathMixScalFun.logMix (0 ms)
[  FAILED  ] mathMixScalFun.logMix
[  FAILED  ] mathMixMatFun.read_corr_matrix (46 ms)
[  FAILED  ] mathMixMatFun.read_corr_matrix_lp (56 ms)
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] mathMixMatFun.read_corr_matrix
[  FAILED  ] mathMixMatFun.read_corr_matrix_lp
[  FAILED  ] mathMixMatFun.log1p (1688 ms)
[  FAILED  ] mathMixMatFun.log1p
[  FAILED  ] MathMixMatFun.matrixExpPade (557 ms)
[  FAILED  ] MathMixMatFun.matrixExpPade
[  FAILED  ] mathMixMatFun.log (1685 ms)
[  FAILED  ] mathMixMatFun.log
[  FAILED  ] mathMixMatFun.acos (749 ms)
[  FAILED  ] mathMixMatFun.acos
[  FAILED  ] mathMixMatFun.atanh (1055 ms)
[  FAILED  ] mathMixMatFun.atanh
[  FAILED  ] mathMixMatFun.read_corr_L (25 ms)
[  FAILED  ] mathMixMatFun.read_corr_L_lp (29 ms)
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] mathMixMatFun.read_corr_L
[  FAILED  ] mathMixMatFun.read_corr_L_lp
[  FAILED  ] mathMixMatFun.invPhi (1268 ms)
[  FAILED  ] mathMixMatFun.invPhi
[  FAILED  ] mathMixMatFun.read_cov_matrix (57 ms)
[  FAILED  ] mathMixMatFun.read_cov_matrix_lp (63 ms)
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] mathMixMatFun.read_cov_matrix
[  FAILED  ] mathMixMatFun.read_cov_matrix_lp
[  FAILED  ] mathMixMatFun.asin (696 ms)
[  FAILED  ] mathMixMatFun.asin
[  FAILED  ] mathMixMatFun.log1m (1942 ms)
[  FAILED  ] mathMixMatFun.log1m
```
All of these fail with something similar to:
```
The difference between x1 and x2 is 0.087490372270337957, which exceeds tol_val, where
x1 evaluates to 352.7469843211403,
x2 evaluates to 352.65949394886997, and
tol_val evaluates to 0.035270323913500511.
expect_near_rel_finite in: expect_near_rel; require items x1(7, 7) = x2(7, 7): hessian() Hessian
```
I'm attaching the file in case you'd want to have a look. 

[out.zip](https://github.com/stan-dev/math/files/5970023/out.zip)

bbbales2:
@adamhaber thanks for the list. I'll go through these later and see what's going on.

A lot of these are probably as well behaved as they're going to get (like log), so we might need to make the finite difference stepsize larger like you showed worked for normal_lcdf.

Since the errors produced by this method are much smaller than the previous method, we have plenty of wiggle room to do this.

bob-carpenter:
Thanks for diving into the error analysis.  This looks like a bug to me in that absolute error is being checked rather than relative error:

```
The difference between x1 and x2 is 0.087490372270337957, which exceeds tol_val, where
x1 evaluates to 352.7469843211403,
x2 evaluates to 352.65949394886997, and
tol_val evaluates to 0.035270323913500511.
expect_near_rel_finite in: expect_near_rel; require items x1(7, 7) = x2(7, 7): hessian() Hessian
```

The values match to within .03 relative tolerance, but not to within .03 absolute tolerance.  This is a case where we should be calling relative tolerance checks.

If this was an existing bug, then fixing these cases will be anotehr great side effect of this project.

bgoodri:
RStan has been numerically differentiating (in R) the autodiffed gradient (from Stan) for a long time. Other than having to make the Hessian symmetrical again, it hasn't been a big deal.

bbbales2:
I switched this back to use the `finite_diff_stepsize` and the failures are only with the normal cdfs.

```
[  FAILED  ] mathMixScalFun.normal_lcdf (1 ms)
[  FAILED  ] mathMixScalFun.normal_lcdf
 1 FAILED TEST
[  FAILED  ] mathMixScalFun.std_normal_lcdf (0 ms)
[  FAILED  ] mathMixScalFun.std_normal_lcdf
 1 FAILED TEST
```

I just removed the problem tests in `std_normal_lcdf` and `normal_lcdf` (since it's the finite difference not the autodiff).

@bob-carpenter @adamhaber if the name doesn't matter so much, what if we just call this `finite_diff_hessian_auto` (what the old function was called)? Any drawbacks there? I can go for a new name, but considering it's the same functionality sticking with the old one means maybe we have less downstream issues.

adamhaber:
> @bob-carpenter @adamhaber if the name doesn't matter so much, what if we just call this finite_diff_hessian_auto (what the old function was called)? Any drawbacks there? I can go for a new name, but considering it's the same functionality sticking with the old one means maybe we have less downstream issues.

Sounds good.

> The values match to within .03 relative tolerance, but not to within .03 absolute tolerance. This is a case where we should be calling relative tolerance checks. 

I've tried to dig into the code of `expect_near_rel_finite`; AFAIU, this should've called 
```
 double tol_val = tol.inexact(x1, x2);
```
with 
```
x1 evaluates to 352.7469843211403,
x2 evaluates to 352.65949394886997
```
which should call
```
 double inexact(const T1& x, const T2& y) const {
    using stan::math::fabs;
    return std::max(tol_ * 0.5 * (fabs(x) + fabs(y)), tol_min_);
  }
```
which doesn't give `tol_val` of .03, given that `tol_` is 1e-8... what am I missing?

bob-carpenter:
Unless you're much better with a debugger than me, I'd suggest adding some print statements to trace what's actually getting called.

bbbales2:
```
The difference between x1 and x2 is 0.087490372270337957, which exceeds tol_val, where
x1 evaluates to 352.7469843211403,
x2 evaluates to 352.65949394886997, and
tol_val evaluates to 0.035270323913500511.
expect_near_rel_finite in: expect_near_rel; require items x1(7, 7) = x2(7, 7): hessian() Hessian
```

> The values match to within .03 relative tolerance, but not to within .03 absolute tolerance. This is a case where we should be calling relative tolerance checks.

The comparison looks like:
```
double tol_val = tol.inexact(x1, x2);
EXPECT_NEAR(x1, x2, tol_val)
    << "expect_near_rel_finite in: " << msg << std::endl;
```

So the error comparison looks like an absolute tolerance thing since it uses EXPECT_NEAR, but the relative tolerance is baked into the calculation of `tol_val`.

The tolerances for Hessian calculations are 1e-4 reltol, 1e-3 abstol.

The tol_val calculation looks like:
```
double inexact(const T1& x, const T2& y) const {
    using stan::math::fabs;
    return std::max(tol_ * 0.5 * (fabs(x) + fabs(y)), tol_min_);
  }
```

Plugging in the numbers:

```
std::max(1e-4 * 0.5 * (353 + 353), 1e-3) // take the maximum of the allowable difference with reltol or abstol
```

tol_val is about 0.0353. So back to the original error message I think everything is working fine:

```
The difference between x1 and x2 is 0.087490372270337957, which exceeds tol_val, where
x1 evaluates to 352.7469843211403,
x2 evaluates to 352.65949394886997, and
tol_val evaluates to 0.035270323913500511.
```

bbbales2 (approved):
Good! I added a couple size tests and renamed a test file. Merge if tests pass and these changes are okay with you @adamhaber

bbbales2:
@adamhaber get a chance to look at the test change? I'll merge if they look okay

adamhaber:
Looks great, thanks for adding!

bob-carpenter:
Nice!  This is not only good for testing, but should also be generally useful.

PR #2352, opened 2021-02-04
use value_array_or_scalar in distributions

SteveBronder:
## Summary

This just tidies up the distributions a little bit by using `as_value_array_or_scalar()` in places we would do things like 

```cpp
  const auto& y_col = as_column_vector_or_scalar(y_ref);
  const auto& alpha_col = as_column_vector_or_scalar(alpha_ref);
  const auto& sigma_col = as_column_vector_or_scalar(sigma_ref);

  const auto& y_arr = as_array_or_scalar(y_col);
  const auto& alpha_arr = as_array_or_scalar(alpha_col);
  const auto& sigma_arr = as_array_or_scalar(sigma_col);

  ref_type_t<decltype(value_of(y_arr))> y_val = value_of(y_arr);
  ref_type_t<decltype(value_of(alpha_arr))> alpha_val = value_of(alpha_arr);
  ref_type_t<decltype(value_of(sigma_arr))> sigma_val = value_of(sigma_arr);
```

So now it's just

```cpp
  auto&& y_val = to_ref(as_value_column_array_or_scalar(y_ref));
  auto&& alpha_val = to_ref(as_value_column_array_or_scalar(alpha_ref));
  auto&& sigma_val = to_ref(as_value_column_array_or_scalar(sigma_ref));
```

Note `auto&&` here has a different meaning then we normally think. Assigning to an `auto&&` type accepts any type as a reference so it can be `auto&`, `auto&&`, or `const auto&`

## Tests

No new tests

## Release notes

Tidy up distributions

## Checklist

- [ ] Copyright holder: Steve Bronder


stan/math/prim/prob/normal_sufficient_lpdf.hpp
```diff
@@ -71,26 +72,15 @@ return_type_t<T_y, T_s, T_loc, T_scale> normal_sufficient_lpdf(
...
-      = promote_scalar<double>(n_obs_arr);
-  ref_type_t<decltype(value_of(mu_arr))> mu_val = value_of(mu_arr);
-  ref_type_t<decltype(value_of(sigma_arr))> sigma_val = value_of(sigma_arr);
+  auto&& y_val = to_ref(as_value_column_array_or_scalar(y_ref));
+  auto&& s_squared_val = to_ref(as_value_column_array_or_scalar(s_squared_ref));
+  auto&& n_obs_val_int = to_ref(as_value_column_array_or_scalar(n_obs_ref));
+  ref_type_t<decltype(promote_scalar<double>(
+      as_array_or_scalar(as_column_vector_or_scalar(n_obs_ref))))>
+      n_obs_val = promote_scalar<double>(
+          as_array_or_scalar(as_column_vector_or_scalar(n_obs_ref)));
```
t4c1, line 81: I think this would be less code if you do it in two steps, just like it was done before.

t4c1 (commented):
Nice cleanup! 

I just think that `auto&&` warrants some discussion, since it is used everywhere. It is definitely OK, but it is not the only way. Previously we had `const auto&` in such cases, however `decltype(auto)` would also work. All these have subtle differences, but in this case all work the same. I am leaning slightly towards `const auto&` since it is the simplest option that works here. Unless you have a good argument for `auto&&`?

SteveBronder:
> Nice cleanup!

ty!

> I just think that auto&& warrants some discussion, since it is used everywhere. It is definitely OK, but it is not the only way. Previously we had const auto& in such cases, however decltype(auto) would also work. All these have subtle differences, but in this case all work the same. I am leaning slightly towards const auto& since it is the simplest option that works here. Unless you have a good argument for auto&&?

Yeah I was actually having difficulty thinking about this. I went with auto&& since it binds to all the reference types nicely. Though it's not as if we move these later so it provides no real benefit besides letting me be lazy. I'd be fine with const&. Just to sanity check, `to_ref()` here will either return an actual type holding memory for expressions, a view expression, or a reference to the input right? For that it almost feels like we want `decltype(auto)` so the object type is deduced as exactly the function return.

t4c1:
> I went with auto&& since it binds to all the reference types nicely.

So do other two options I mentioned. The difference between `decltype(auto)` and `auto&&` is only how they are deduced when assigned a rvalue reference. `decltype(auto)` will be a value, while `auto&&` will be a rvalue reference. And `const auto&` is obviously const and always a reference. 

>  Just to sanity check, to_ref() here will either return an actual type holding memory for expressions, a view expression, or a reference to the input right?

Right. As I mentioned all three options work exactly the same. As we don't need to change these variables I'd prefer the const option.

SteveBronder:
Would you mind if I used `decltype(auto)`? idk in my brain I know that `const&` is fine because it holds a constant reference to an object in the local scope. But then my brain is also like "No you are returning back an actual type use `decltype(auto)` so that the compiler knows it's not a reference

t4c1:
I guess that logic makes just as much sense as what I suggested.

bbbales2:
I just like `const &` cause then I don't have to think. It's just a const reference.

Whenever I see `decltype(auto)` or `&&` I know something possibly weird is happening and I need to stop and think about rvalues/lvalues and tricky things. `const &` is kinda no tricks.

Is there a reason a compiler would have trouble with a `const &`?

bbbales2:
But on average this pull simplifies the code that came before it, so if this is ready to merge and you prefer `auto&&` or `decltype(auto)`, sounds good to me.

SteveBronder:
> Is there a reason a compiler would have trouble with a const &?

@t4c1 do you know a reference somewhere I can read about the rules surrounding cont& as the object type? I'm just not sure what to google to read more about how that works

If yinz are fine with decltype then I think this is good to go

bbbales2:
Gaaah I wanna say yes and move on with the deserializer but then I went and read about `decltype(auto)`. From [here](https://en.cppreference.com/w/cpp/language/auto):

```cpp
auto a = 1 + 2;            // type of a is int
...
decltype(auto) c1 = a;   // type of c1 is int, holding a copy of a
decltype(auto) c2 = (a); // type of c2 is int&, an alias of a
```

Ugh what a weird thing.

Can we just be plain `const &`? Everything comes into these functions as `const &` types so I don't think that would be a major problem. Also the other temporaries in these functions are `const &`.

I keep flipping flopping on this in the most useless way.

SteveBronder:
`as_array_or_scalar()` These are used like 

```cpp
decltype(auto) x_ref = to_ref(as_value_array_or_scalar(x))
```
and since `to_ref()` and `as_value_array_or_scalar()` both return references in the case of matrices/vectors with scalars of double the above behavior is exactly what we want!

bbbales2 (dismissed):
Good

bbbales2:
Alright decltype it is. I scanned through this just to make sure everything was in order. There were some leftover things in `beta_lpdf`. I deleted those things and grepped and there were a couple other stragglers. Diff [here](https://github.com/stan-dev/math/pull/2352/commits/d6328f8810f689eec097b966b2233cc534115b15).

SteveBronder:
Nice! Yeah there's a few other places we could clean up like the glm functions but I wanted to make this PR just the layups that were obvious

bbbales2 (approved):
Good (previous review got dismissed)

t4c1:
@SteveBronder You may want to look at how "lifetime extension" works.

PR #2356, opened 2021-02-07
Revert "Revert "Add tests for var_value<Matrix> for univariate distributions""

SteveBronder:
Pr to revert revert univariate distributions with var value


rok-cesnovar:
@SteveBronder make sure to start the tests with row vector tests once you think you have the fix: https://jenkins.mc-stan.org/job/Math%20Pipeline/job/PR-2356/build?delay=0sec

But you probably already know this, so sorry for the noise :)

SteveBronder:
Did not know that so ty! Is it possible to have Jenkins make a revert branch / open a PR with the revert up on a failure in dev? I wonder if that would be worth the effort

rok-cesnovar:
Might be, but we have false positive occassionaly (aws throwing us out for example). Might get annoying.

SteveBronder:
Alright looks like we are good! Lemme turn off the row vector flags and we can merge this. I think the fix was just the arena matrix thing

rok-cesnovar:
No need to turn it off, its good.

rok-cesnovar:
Oh you put it jn jenkinsfile, sorry. Yeah we should remove that. I thought you just tickef the checkbox.

bbbales2:
We should throw up tests for the three new arena constructors too. I guess it doesn't have to go here since the merge tests will eventually catch errors with this, but it'd be good to have quick tests not embedded in the distribution stuff.

bbbales2:
I added a test since the other tests were failing (and now they're movin' again)

SteveBronder:
Odd I restarted the full unit tests but it still has an X here

https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2356/14/pipeline/

rok-cesnovar:
This is a “stage” restart, it started on the parallel stage. Those restarts dont show up as a orange dot but will give a green light if it passes.

SteveBronder:
Nice! Alright I think this is good to merge!

bbbales2 (approved):
Good!

bbbales2:
@rok-cesnovar I'll start a new list of compatible functions generating tomorrow.

PR #2358, opened 2021-02-08
rep_matrix for var<Matrix>

SteveBronder:
## Summary

@rok-cesnovar wanted to make sure this made sense for the compiler before I did the full impl. This adds a `Ret` template parameter for `rep_matrix()` and friends so that we can overload it for `var<Matrix>` types. 

```cpp
rep_matrix<var_value<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>>(x, 10, 10);
```

Because the default return parameter depends on the input scalar type the return type needs to be the second template. I think with some trickery I could make it the first template but it might lead to some confusing c++

This is backwards compatible with the current `rep_matrix()` where not setting the template parameters returns back a `Eigen::Matrix<ScalarType, -1, -1>`

```cpp
// compiles fine
Eigen::Matrix<var, -1, -1> x = rep_matrix(var(1.0), 10, 10);
```

I think in the compiler  we could call this with

```cpp
using eig_mat = Eigen::Matrix<double, -1, -1>;
conditional_var_value_t<local_scalar_t__,  eig_mat> x = rep_matrix<var_value<eig_mat>>(x_scal, 5, 5);
```

## Tests

Example test is in `stan/math/rev/fun/rep_matrix.hpp`

## Side Effects

idt so

## Release notes

Add `rep_*` utility functions for new matrix type

## Checklist

- [x] Math issue #1805 

- [] the new changes are tested


t4c1:
> Because the default return parameter depends on the input scalar type the return type needs to be the second template. I think with some trickery I could make it the first template but it might lead to some confusing c++

You may want to look at how I templated `from_matrix_cl`.

SteveBronder:
> You may want to look at how I templated from_matrix_cl.

Ah nice! Yeah I can use this.

I'm looking over some stuff now and think we could actually coalesce a lot of the opencl and eigen var matrix functions. I think what we need to do is have `require_var_matrix_vt<is_matrix_cl/is_eigen, T>` and then `require_var_matrix_t<>` accepts a `var_value<>` with either an Eigen matrix or `matrix_cl`. Like for these functions if we had a `static Constant` function in `matrix_cl` that returned a `constant_` type that would let us reuse this code for both the inner eigen and cl matrices. And for like acos etc. they would become

```
template <typename T,
          require_var_matrix_t<T>* = nullptr>
inline auto acos(const T& A) {
  return make_callback_var(
      acos(A.val()), [A](auto& res) mutable {
        A.adj() -= elt_divide(res.adj(), sqrt(1.0 - square(A.val())));
      });
}
```

bbbales2:
@SteveBronder I made a bunch of changes so I put them in a compare instead of pushing here: https://github.com/stan-dev/math/compare/feature/rep_utils-var-matrix...feature/rep_utils-var-matrix-2

1. I don't think we need overloads that build these matrices from arithmetic types. It's possible to copy a Eigen::VectorXd/Eigen::MatrixXd/Eigen::RowVectorXd to a matvar or a varmat right out the gate

2. Only the `rep_matrix(real, int, int)` function has the ambiguity in matvar/varmat, so the `rep_matrix` for vector types can be implemented and tested with the regular stuff.

3. Since all that the template argument is doing now is picking between two functions, I wonder if we should just have a separate function `rep_matrix_varmat` or whatever for this? I don't think it really matters. Presumably we'll change this to whatever is convenient for the compiler to Do The Right Thing.

bbbales2:
@SteveBronder so you're aware the OpenCL version of this went in recently. You can see it in the blame here: https://github.com/stan-dev/math/blame/develop/stan/math/prim/fun/rep_matrix.hpp

SteveBronder:
Catching up with this now. I'd like to keep the same form as the OpenCL version so it's easiest in the compiler, let me have a another look at this rq

bbbales2:
Cool cool. I think the simplifications in the compare are good then.

SteveBronder:
Alright looked it over I think we are good!

bbbales2:
@SteveBronder there's some stuff in the compare we should do: https://github.com/stan-dev/math/compare/feature/rep_utils-var-matrix...feature/rep_utils-var-matrix-2

It removes three functions we don't need, adds varmat tests for the vector/row vector signatures, and fixes a bug in the varmat implementation.

SteveBronder:
@bbbales2 I think those changes all make sense and I merged them in. @rok-cesnovar I commented out the opencl tests in this PR and we can re-enable them when @bgoodri is able to get back into his office

rok-cesnovar:
@SteveBronder me and @serban-nicusor-toptal are actually working on something wrt to fixing that so you can uncomment them back in.

SteveBronder:
Cool cool my b!

SteveBronder:
```
lib/stan_math/stan/math/prim/prob/lkj_corr_lpdf.hpp:74:55:   required from ‘stan::return_type_t<T1, T2> stan::math::lkj_corr_lpdf(const T_y&, const T_shape&) [with bool propto = false; T_y = Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<int>, Eigen::Matrix<int, -1, -1> >; T_shape = double; stan::return_type_t<T1, T2> = double]’

./test/test-models/good/validate_jacobian_warning_good.hpp:223:47:   required from ‘stan::scalar_type_t<T2> validate_jacobian_warning_good_model_namespace::validate_jacobian_warning_good_model::log_prob_impl(VecR&, VecI&, std::ostream*) const [with bool propto__ = false; bool jacobian__ = false; VecR = std::vector<stan::math::var_value<double>, std::allocator<stan::math::var_value<double> > >; VecI = std::vector<int>; stan::require_vector_like_t<VecR>* <anonymous> = 0; stan::require_vector_like_vt<std::is_integral, VecI>* <anonymous> = 0; stan::scalar_type_t<T2> = stan::math::var_value<double>; std::ostream = std::basic_ostream<char>]’

./test/test-models/good/validate_jacobian_warning_good.hpp:616:77:   required from ‘T__ validate_jacobian_warning_good_model_namespace::validate_jacobian_warning_good_model::log_prob(std::vector<T_l>&, std::vector<int>&, std::ostream*) const [with bool propto__ = false; bool jacobian__ = false; T__ = stan::math::var_value<double>; std::ostream = std::basic_ostream<char>]’

test/test-model-main.cpp:9:43:   required from here
```

That's a weird one. I'll take a look later

bbbales2:
I suspect it's making an integer matrix and the code is expecting a double matrix.

The following warnings are:

```
lib/stan_math/lib/eigen_3.3.9/Eigen/src/Core/util/StaticAssert.h:185:25: error: static assertion failed: THIS_FUNCTION_IS_NOT_FOR_INTEGER_NUMERIC_TYPES

  185 |     EIGEN_STATIC_ASSERT(!NumTraits<TYPE>::IsInteger, THIS_FUNCTION_IS_NOT_FOR_INTEGER_NUMERIC_TYPES)

      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I think the line in the model ([here](https://github.com/stan-dev/stan/blob/develop/src/test/test-models/good/validate_jacobian_warning_good.stan)) is this one:

```
rep_matrix(1, 3, 3) ~ lkj_corr(2.0);
```

Since matrices in Stan are always floating point types (or floating point autodiff types), I think we should just static cast the integral types to doubles.

bbbales2 (approved):
Good!

Issue #2351, opened 2021-02-04
OpenCL runtime error with inv_gamma_lpdf


rmcminds:
A script that compiles and runs fine in serial is compiled fine with opencl, but throws this error at runtime:

> Unrecoverable error evaluating the log probability at the initial value.
> Exception: binary_operator+: Rows of a (2861) and rows of b (1) must match in size (in 'file.stan', line 285, column 8 to column 86)
> Exception: binary_operator+: Rows of a (2861) and rows of b (1) must match in size (in 'file.stan', line 285, column 8 to column 86)

At issue is the following signature:

`lp += inv_gamma_lpdf(row_vector | scalar, scalar);`

I can get the script to run with opencl using this modified version:

`lp += inv_gamma_lpdf(to_vector(row_vector) | rep_vector(scalar, size(row_vector)), rep_vector(scalar, size(row_vector)));`

Version 4.0.0

t4c1:
I can confirm the bug. Actually we forgot to test OpenCL implementations with `row_vector`s. So they won't work with combinations of `vector` and `row_vector` arguments. Some of them, `inv_gamma_lpdf` included, won't even work wit combinations of `row_vector` and `scalar` arguments. Ouch.

I will try to fix that as soon as possible. Meanwhile the simplest workaround that should work is to call `inv_gamma_lpdf(to_vector(row_vector) | scalar, scalar)`.

PR #2360, opened 2021-02-09
Bugfix OpenCL distributions not working with row vectors

t4c1:
## Summary

Fixes OpenCL implementations of distributions mostly not working with row vectors. This is done by implementing `as_column_vector_or_scalar` for kernel generator expressions and using it in distribution implementations.

I could not fix `dirichlet_lpdf` in `multi_normal_cholesky_lpdf` without performance impacts so these two are still not supporting row vectors in Math. Row vector support for these two can be added in stanc, by transposing row vector arguments.

## Tests
Added many tests for row vectors and tests for `as_column_vector_or_scalar`. `student_t_lpdf` tests failed to compile due to being too large so I split them into two files.

## Side Effects
There are many more tests for OpenCL stuff now potentially significantly increasing testing times.

## Release notes

OpenCL: Fixed OpenCL implementations of distributions mostly not working with row vectors.

## Checklist

- [ ] Math issue #2351 


rok-cesnovar:
@t4c1 Sorry for this, but can you split out the fixes for the distributions that were added for 2.26. We will need to do a patch release and I would then want to include those as well. If we go that route this PR needs to be done to master.

bbbales2:
> I could not fix dirichlet_lpdf in multi_normal_cholesky_lpdf without performance impacts so these two are still not supporting row vectors in Math. Row vector support for these two can be added in stanc, by transposing row vector arguments.

Is it simpler to just take the performance hit for now? Stanc3 logic can be added later if we think that's the better way to handle these things.

Otherwise is this ready for review?

t4c1:
The performance hit version of fix for these two is not exactly trivial, so I would prefer to just properly fix those by transposing row vectors in compiler.

t4c1:
Also cherry-picked version of this is already merged into master, so I think @rok-cesnovar is just waiting for the bugfix release to be completed, before approving this PR as well.

bbbales2:
> so I would prefer to just properly fix those by transposing row vectors in compiler.

Is that something easy to add in the compiler? I really don't know. It sounds confusing to me.

@rok-cesnovar Do we do things like this?

Can we do a compatibility interface in Math at least? So like:

```cpp
template <typename T,
                 require_col_vector_t<T> = nullptr>
auto myfunc(const T&) {
  ... // This is the fast implementation
}

template <typename T,
                 require_not_col_vector_t<T>* = nullptr>
auto myfunc(const T& x) {
  auto y = copy_to_col_vector(x);
  return myfunc(y);
}
```

And then in the future the compiler can do whatever optimizations.

t4c1:
> Can we do a compatibility interface in Math at least? So like:

No. The issue is that we are using the same `matrix_cl` type for both row and col vectors. Only stanc knows at compile time which one a matrix_cl contains.

rok-cesnovar:
> Do we do things like this?

So far no.

the only thing I can see us doing in Math is fun_impl() and the fun() checks whether its a row_vector and calls fun_imp(transpose(y)), or fun_impl(y) if not. 

Not sure if that is doable/performant.

bbbales2:
> Only stanc knows at compile time which one a matrix_cl contains.

Oof. I guess the Math option would be `matrix_cl<Eigen::VectorXd>` vs. `matrix_cl<Eigen::RowVectorXd>` which seems like a lot of trouble?

If `matrix_cl<double>` types are all the same, is it in the transfer from CPU to GPU that is causing problems? Or is this something that comes up when a GPU function passes something to another GPU function?

> The issue is that we are using the same matrix_cl type for both row and col vectors.

The varmat distributions got reverted because `arena_t<Eigen::VectorXd>` and `arena_t<Eigen::RowVectorXd>` weren't playing well together. I put this here just cause it seems similar (found out about this through dist tests). Without the type info, the same fix wouldn't work tho.

Like this is valid Eigen code but it wasn't working with arena types:

```
Eigen::VectorXd a(5);
Eigen::RowVectorXd b = a;
```

There's a patch in `arena_matrix.hpp` here: https://github.com/stan-dev/math/pull/2356/files#diff-4dcedb2d724267e0af41b776bf83e6a402a26fa5f2198a245922813544f804cbR61

t4c1:
> Not sure if that is doable/performant.

We could actually do that, except we would need 4 branches, since 2 arguments can be row vectors. Performance difference would likely be negligible.

> I guess the Math option would be matrix_cl<Eigen::VectorXd> vs. matrix_cl<Eigen::RowVectorXd> which seems like a lot of trouble?

Yep, that would be quite some work.

> is it in the transfer from CPU to GPU that is causing problems? Or is this something that comes up when a GPU function passes something to another GPU function?

No. The issue is that binary functions expect arguments to have the same shape.

> Like this is valid Eigen code but it wasn't working with arena types:

We do not support that with `matrix_cl`s (and adding that would not solve the issue).

rok-cesnovar:
> We could actually do that, except we would need 4 branches, since 2 arguments can be row vectors. Performance difference would likely be negligible.

Oh, right. I am now leaning slightly towards this solution as this makes these distributions work with row_vectors standalone, as part of Stan Math.

t4c1:
Oh, right we can't actually do that. The problem is that we can not differentiate between `Eigen::RowVector` and a `std::vector<Eigen::Vector>` from matrix_cl dimensions if the second one has inner vectors of size 1. We do need to solve this in stanc.

rok-cesnovar:
Argh, ok then :)

rok-cesnovar (approved):

Issue #2370, opened 2021-02-16
Issue with nested reduce_sum in 2.26.0


grburgess:
I am getting a strange compilation error after upgrading when compiling a model that uses reduce sum:

```
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/memory:3033:32: error: no matching constructor for initialization of 'std::__1::tuple<Eigen::Matrix<double, -1, 1, 0, -1, 1> &&, Eigen::Matrix<double, -1, 1, 0, -1, 1> &&, Eigen::Matrix<stan::math::var_value<double, void>, 1, -1, 1, 1, -1>, Eigen::Matrix<stan::math::var_value<double, void>, 1, -1, 1, 1, -1>, Eigen::Matrix<stan::math::var_value<double, void>, -1, 1, 0, -1, 1>, Eigen::Matrix<stan::math::var_value<double, void>, -1, 1, 0, -1, 1>, stan::math::var_value<double, void>, stan::math::var_value<double, void>, stan::math::var_value<double, void>, stan::math::var_value<double, void>, stan::math::var_value<double, void>, const int &>'
    return unique_ptr<_Tp>(new _Tp(_VSTD::forward<_Args>(__args)...));
                               ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #2371, opened 2021-02-16
Fixes bug in `reduce_sum` handling expressions with threading enabled

t4c1:
## Summary

Fixes two bugs in `reduce_sum` when `STAN_THREADS` is set. 
 - under some conditions `reduce_sum` did not compile (the original reported bug in the issue)
 - shared arguments could be evaluated multiple times (the reason for expression tests failing with `STAN_THREADS` set)

## Tests
None for now. @rok-cesnovar promised he will add tests for those after the bugfix release.

## Side Effects
None

## Release notes
Fixes two bugs in `reduce_sum` when `STAN_THREADS` is set. 

## Checklist

- [x] Math issue #2370

### DIFF
--- a/stan/math/prim/functor/reduce_sum.hpp
+++ b/stan/math/prim/functor/reduce_sum.hpp
@@ -204,9 +204,9 @@ inline auto reduce_sum(Vec&& vmapped, int grainsize, std::ostream* msgs,
 
 #ifdef STAN_THREADS
   return internal::reduce_sum_impl<ReduceFunction, void, return_type, Vec,
-                                   Args...>()(std::forward<Vec>(vmapped), true,
-                                              grainsize, msgs,
-                                              std::forward<Args>(args)...);
+                                   ref_type_t<Args&&>...>()(
+      std::forward<Vec>(vmapped), true, grainsize, msgs,
+      std::forward<Args>(args)...);
 #else
   if (vmapped.empty()) {
     return return_type(0.0);
--- a/stan/math/rev/core/deep_copy_vars.hpp
+++ b/stan/math/rev/core/deep_copy_vars.hpp
@@ -20,7 +20,7 @@ namespace math {
  *  Otherwise it will be moved.
  */
 template <typename Arith, typename = require_arithmetic_t<scalar_type_t<Arith>>>
-inline decltype(auto) deep_copy_vars(Arith&& arg) {
+inline Arith deep_copy_vars(Arith&& arg) {
   return std::forward<Arith>(arg);
 }
 


stan/math/rev/core/deep_copy_vars.hpp
```diff
@@ -20,7 +20,7 @@ namespace math {
  *  Otherwise it will be moved.
  */
 template <typename Arith, typename = require_arithmetic_t<scalar_type_t<Arith>>>
-inline decltype(auto) deep_copy_vars(Arith&& arg) {
+inline Arith deep_copy_vars(Arith&& arg) {
```
bbbales2, line 23: Was the bug that `deep_copy` wasn't handling expressions correctly? Or was it that the `reduce_sum` threaded signature didn't work with expressions?

Should there be a changes to Jenkins that turns on threading for the expression tests?

bbbales2 (commented):
Questions

stan/math/rev/core/deep_copy_vars.hpp
rok-cesnovar: Threaded expressions tests for reduce_sum/map_rect will be added (that needs to be done in the Stan repo). I will add that later today. But that would not catch one of the two issues. 


For the second one need to add a minimal version of the model to the test models.

bbbales2: I'm asking cause it looks like `deep_copy_vars` was just returning the wrong type or something.

If it's that, then we just need to add something like this to the `deep_copy_var` unit tests:

```
EXPECT_TRUE((std::is_same<decltype(deep_copy_vars(arith_expr)), plain_type_t<decltype(arith_expr)>>));
```

bbbales2: Well that seems wrong to me now cause it's assuming that deep_copy_vars evals the expression -- anyway, something like this? It looks like a problem with `deep_copy_vars` and I'm not sure what exactly `decltype(auto)` gets wrong there (though I don't doubt it gets something wrong).

t4c1: >Was the bug that deep_copy wasn't handling expressions correctly?

No. The issue was `reduce_sum` was assuming `deep_copy` does not return rvalue references. Since returning rvalue references can in this case cause some other issues, I just fixed it here.

>Well that seems wrong to me now cause it's assuming that deep_copy_vars evals the expression

Nope, there is no `eval` or `plain_type_t` in `deep_copy_vars`, so it does not eval anything.

stan/math/prim/functor/reduce_sum.hpp
```diff
@@ -204,9 +204,9 @@ inline auto reduce_sum(Vec&& vmapped, int grainsize, std::ostream* msgs,
 
 #ifdef STAN_THREADS
   return internal::reduce_sum_impl<ReduceFunction, void, return_type, Vec,
-                                   Args...>()(std::forward<Vec>(vmapped), true,
-                                              grainsize, msgs,
-                                              std::forward<Args>(args)...);
+                                   ref_type_t<Args&&>...>()(
+      std::forward<Vec>(vmapped), true, grainsize, msgs,
+      std::forward<Args>(args)...);
```
bbbales2, line 209: So this is using the cast-arguments-once rule to make the `to_ref` happen that evaluates the input arguments if necessary.

What is the difference between what is here and this (so doing the `to_ref` explicitly):
```cpp
  ref_type_t<Args&&>...>()(std::forward<Vec>(vmapped), true,
                           grainsize, msgs,
                           std::forward<ref_tupe_t<Args&&>>(to_ref(std::forward<Args>(args)))...);
```

Is this just more code for the same thing?

stan/math/rev/core/deep_copy_vars.hpp
bbbales2: This makes sense.

bbbales2 (approved):
Good!

I left a question.

rok-cesnovar:
@serban-nicusor-toptal Math master is now ready. Lets leave it to propagate to CmdStan and do the 2.26.1 tomorrow CET time.

stan/math/prim/functor/reduce_sum.hpp
t4c1: Yep, `to_ref` is just conversion to `ref_type_t`.

bbbales2: Question, if we have an argument `T&& x` that has type `const std::vector<double>&`

Then is it true that:

```
std::is_same<ref_type_t<T&&>, const std::vector<double> &>
```

So that this makes no copy:

```
ref_type_t<T&&> a(deep_copy_vars(x))
```

I was thinking and I think the reason we did `decltype(auto)` there was to make sure `deep_copy` returned references when the input was references so that the only thing that would be saved in the tuple for non-var arguments were references to things we know exist (and rvalues are fine cause how the function is called they won't get deleted).

Now the return type is T, but is there some required copy elison or something that makes sure a copy doesn't happen (a type `T` get constructed by copying the input x which is then returned?)

t4c1: > Then is it true that:

Yes.

> So that this makes no copy:

Yes.

> Now the return type is T, but is there some required copy elison or something that makes sure a copy doesn't happen (a type T get constructed by copying the input x which is then returned?)

Yes, we rely on the compiler to do RVO.

bbbales2: Makes sense, thanks!

PR #2373, opened 2021-02-17
Adds lb/ub/lub_constrain specializations

SteveBronder:
## Summary

This PR adds the specializations for `lub_constrain` for matrices and scalars with all their combinations of inputs. The branch I was working on to make lb/ub/lub constrain work for matvar was getting pretty big so I'm breaking it up into 3 seperate ones.

## Tests

Tests were updated for `matvar` and to the calculations when computing the gradient of both lp and the input.

## Side Effects

There's a few weirds I can't completely sort out.

@t4c1 Is `make_holder` able to deal with plain references? I was trying to use it for the functions that need to calculate lp like 

```cpp
template <typename T, typename L, require_all_eigen_t<T, L>* = nullptr,
          require_all_not_st_var<T, L>* = nullptr>
inline auto lb_constrain(T&& x, L&& lb) {
  return make_holder([](const auto& x, const auto& lb, auto& lp) {
    return x.binaryExpr(lb, [&lp](auto&& x, auto&& lb) {
       return lb_constrain(x, lb, lp);
    });
  }, std::forward<T>(x), std::forward<L>(lb), lp);
}
```

But I was getting seg faults when trying.

@andrjohns is it possible to use the apply meta stuff to have this work for `std::vectors` of containers? I couldn't figure out how to have it work whenever I have an `x` like `std::vector<Eigen::MatrixXd>` and an `lb` of `double`
## Release notes

## Checklist

- [x] Math issue #1805 


stan/math/rev/fun/lb_constrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+    auto is_inf_lb = to_arena(
+        (arena_lb.val().array() != NEGATIVE_INFTY).template cast<double>());
+    auto precomp_x_exp = to_arena((arena_x.val().array() * is_inf_lb).exp());
+    arena_t<ret_type> ret = (is_inf_lb).select(
+        precomp_x_exp + arena_lb.val().array(), arena_x.val().array());
+    reverse_pass_callback(
+        [arena_x, arena_lb, ret, is_inf_lb, precomp_x_exp]() mutable {
+          arena_x.adj().array() += ret.adj().array() * precomp_x_exp;
+          arena_lb.adj().array() += ret.adj().array() * is_inf_lb;
+        });
```
SteveBronder, line 263: This is sort of a dirty trick, but since `exp(0) = 1` We can cast all of these checks to double and then in the forward pass multiply `x` by the check array. When we make `ret` we iterate over `is_inf_lb` and when it's true it uses the values we want and when false just returns `arena_x` at that value. 

But the nice thing is that we can then have no checks in the reverse pass and get nice SIMD instructions since we are just doing an elementwise multiplication

```cpp
arena_x.adj().array() += ret.adj().array() * precomp_x_exp;
```

t4c1:
@t4c1 Is make_holder able to deal with plain references?

Yes, it should be. Do you even need holders for constrain/free functions? I think probably not.

stan/math/rev/fun/lb_constrain.hpp
t4c1: PRs introducing such dirty tricks should be accompanied by benchmarks proving they are actually faster than the alternative implementation.

bbbales2:
> I was trying to use it for the functions that need to calculate lp like

I don't like the expressions there since lp is an output in this code. Even if it weren't segfaulting, that would mean an accidental double eval of the expression would make a math bug rather than just a performance bug (updating lp twice).

> I couldn't figure out how to have it work whenever I have an x like std::vector<Eigen::MatrixXd> and an lb of double

I thought x was always either a scalar or a matrix here.

bbbales2:
@SteveBronder I made a comparison branch where I removed a lot of stuff: https://github.com/stan-dev/math/compare/feature/lb_constrain-matvar...feature/lb_constrain-matvar_shrunk

I deleted a bunch of stuff and I think this will work okay too. I don't think we need to use identity constrain in here.

I think we only need reverse mode specializations when both types are matrices, otherwise prim can handle the varmat autodiff.

Ofc. this would have more temporaries, but the implementation is way simpler. Can also go verbose for the performance -- it's just a metric ton of code.

SteveBronder:
idt those will work because we need to check each cell individually for infinity. Doing it the way you wrote it above you end up just checking if all of them are positive infinity and if not then do the constrain.

bbbales2:
> Doing it the way you wrote it above you end up just checking if all of them are positive infinity and if not then do the constrain.

The signatures are:

```
lb_constrain(real, real);
lb_constrain(vector, real);
lb_constrain(vector, vector);
```

(and then the ones for lp)

I'm pretty sure the code I wrote covers that. Is there a signature I'm missing here?

SteveBronder:
I think that could work but I already wrote out the complicated bits lol so I'd rather just use the specializations since they would be more efficient

bbbales2:
Makes sense. Lemme revieeeeew

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -22,17 +26,21 @@ namespace math {
...
  */
-template <typename T, typename L>
+template <typename T, typename L, require_all_stan_scalar_t<T, L>* = nullptr,
+          require_all_not_st_var<T, L>* = nullptr>
 inline auto lb_constrain(const T& x, const L& lb) {
-  auto& lb_ref = to_ref(lb);
-  check_finite("lb_constrain", "lb", value_of(lb_ref));
-  return eval(add(exp(x), lb_ref));
+  if (unlikely(lb == NEGATIVE_INFTY)) {
+    return identity_constrain(x, lb);
```
bbbales2, line 39: Just return `x` here. Let's remove identity_constrain.

stan/math/prim/fun/identity_constrain.hpp
```diff
@@ -8,37 +8,18 @@ namespace math {
...
- * @param[in] lp log density reference
  * @return transformed input
  */
-template <typename T, typename S>
-inline auto identity_constrain(T&& x, S& lp) {
-  return std::forward<T>(x);
+template <typename T, typename... Types,
+          require_all_not_var_matrix_t<T, Types...>* = nullptr>
+inline auto identity_constrain(T&& x, Types&&... /* args */) {
+  return promote_scalar_t<return_type_t<T, Types...>, T>(x);
```
bbbales2, line 22: Just revert the identity constrain stuff. It's just unnecessary for these functions.

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -22,17 +26,21 @@ namespace math {
...
+  } else {
+    // check_less("lb_constrain", "lb", value_of(x), value_of(lb));
```
bbbales2, line 41: Delete comment

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -41,21 +49,104 @@ inline auto lb_constrain(const T& x, const L& lb) {
...
 inline auto lb_constrain(const T& x, const L& lb, return_type_t<T, L>& lp) {
-  auto& x_ref = to_ref(x);
-  auto& lb_ref = to_ref(lb);
-  check_finite("lb_constrain", "lb", value_of(lb_ref));
-  lp += sum(x_ref);
-  return eval(add(exp(x_ref), lb_ref));
+  if (lb == NEGATIVE_INFTY) {
+    return identity_constrain(x, lb);
+  } else {
+    // check_less("lb_constrain", "lb", value_of(x), value_of(lb));
```
bbbales2, line 65: Delete comment

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -41,21 +49,104 @@ inline auto lb_constrain(const T& x, const L& lb) {
...
+ */
+template <typename T, typename L, require_eigen_t<T>* = nullptr,
+          require_stan_scalar_t<L>* = nullptr,
+          require_all_not_st_var<T, L>* = nullptr>
+inline auto lb_constrain(T&& x, L&& lb) {
+  return make_holder(
+      [](const auto& x, const auto& lb) {
+        return x.unaryExpr([lb](auto&& x) { return lb_constrain(x, lb); });
+      },
+      std::forward<T>(x), std::forward<L>(lb));
```
bbbales2, line 89: [optional] I'd rather just evaluate expressions here because I'm pretty sure we're always assigning to l-values somewhere in generated C++. I think it does cost a malloc or something to make a holder? @t4c1 can confirm if they're free or not.

stan/math/prim/fun/lb_constrain.hpp
```diff
@@ -41,21 +49,104 @@ inline auto lb_constrain(const T& x, const L& lb) {
...
+ * @tparam L A type inheriting from `EigenBase`.
+ * @param[in] x unconstrained input
+ * @param[in] lb lower bound on output
+ * @param[in,out] lp reference to log probability to increment
+ * @return lower-bound constrained value corresponding to inputs
+ */
+template <typename T, typename L, require_all_eigen_t<T, L>* = nullptr,
+          require_all_not_st_var<T, L>* = nullptr>
+inline auto lb_constrain(const T& x, const L& lb,
+                         std::decay_t<return_type_t<T, L>>& lp) {
```
bbbales2, line 147: Is the decay necessary? I feel like if it is we should probably patch `return_type_t`.

stan/math/rev/fun/identity_constrain.hpp
```diff
@@ -0,0 +1,44 @@
...
+          require_any_var_matrix_t<T, Types...>* = nullptr>
+inline auto identity_constrain(T&& x, Types&&... /* args */) {
+  return to_var_value(x);
+}
+
+template <typename T, typename... Types, require_var_matrix_t<T>* = nullptr,
+          require_any_var_matrix_t<T, Types...>* = nullptr>
+inline auto identity_constrain(T&& x, Types&&... /* args */) {
+  return x;
+}
```
bbbales2, line 39: Yeah let's definitely dodge this identity constrain thing.

stan/math/rev/fun/lb_constrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+ * @param[in] lb lower bound on constrained output
+ * @return lower bound constrained value corresponding to inputs
+ */
+template <typename T, typename L, require_all_stan_scalar_t<T, L>* = nullptr,
+          require_any_var_t<T, L>* = nullptr>
+inline auto lb_constrain(const T& x, const L& lb) {
+  using std::exp;
+  const auto lb_val = value_of(lb);
+  if (unlikely(lb_val == NEGATIVE_INFTY)) {
+    return identity_constrain(x, lb);
```
bbbales2, line 46: `return x;`

stan/math/rev/fun/lb_constrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+ * @param[in,out] lp reference to log probability to increment
+ * @return lower bound constrained value corresponding to inputs
+ */
+template <typename T, typename L, require_all_stan_scalar_t<T, L>* = nullptr,
+          require_any_var_t<T, L>* = nullptr>
+inline auto lb_constrain(const T& x, const L& lb, var& lp) {
+  using std::exp;
+  const auto lb_val = value_of(lb);
+  if (unlikely(lb_val == NEGATIVE_INFTY)) {
+    return identity_constrain(x, lb);
```
bbbales2, line 97: `return x;`

stan/math/rev/fun/lb_constrain.hpp
```diff
@@ -0,0 +1,379 @@
...
+ * @tparam T Scalar
+ * @tparam L Scalar
+ * @param[in] x Unconstrained input
+ * @param[in] lb lower bound on constrained output
+ * @return lower bound constrained value corresponding to inputs
+ */
+template <typename T, typename L, require_all_stan_scalar_t<T, L>* = nullptr,
+          require_any_var_t<T, L>* = nullptr>
+inline auto lb_constrain(const T& x, const L& lb) {
+  using std::exp;
```
bbbales2, line 43: Don't need this since you're always using `std::exp` in the code

stan/math/rev/fun/lb_constrain.hpp
bbbales2: This doesn't seem like a dirty trick. Maybe just a trick lol. I'm happy with it.

test/unit/math/mix/fun/lb_constrain_test.cpp
```diff
@@ -14,52 +16,45 @@ void expect_lb_constrain(double x, double lb) {
...
     return lp;
   };
+  auto f4 = [](const auto& x, const auto& lb) {
+    stan::return_type_t<decltype(x), decltype(lb)> lp = 0;
+    auto xx = stan::math::lb_constrain(x, lb, lp);
+    return stan::math::add(lp, stan::math::sum(xx));
+  };
 
-  stan::test::expect_ad(f1, x, lb);
+  // stan::test::expect_ad(f1, x, lb);
```
bbbales2, line 25: Turn these tests back on

test/unit/math/prim/fun/lb_transform_test.cpp
```diff
@@ -5,9 +5,6 @@
 
 TEST(prob_transform, lb) {
   EXPECT_FLOAT_EQ(exp(-1.0) + 2.0, stan::math::lb_constrain(-1.0, 2.0));
-  EXPECT_THROW(
-      stan::math::lb_constrain(7.9, -std::numeric_limits<double>::infinity()),
-      std::domain_error);
```
bbbales2, line 10: I think go back in history with these checks: https://github.com/stan-dev/math/blob/89c8c8c130f40ccd1b0fb42bdbd2049644939461/test/unit/math/prim/fun/lb_transform_test.cpp

bbbales2: We need prim checks that `lb_constrain(vector, real)` and `lb_constrain(vector, vector)` are behaving well with infs (not transforming when there is an inf, transforming otherwise)

test/unit/math/mix/fun/lb_constrain_test.cpp
```diff
@@ -14,52 +16,45 @@ void expect_lb_constrain(double x, double lb) {
...
+  A << 5.0, 2.0, 4.0, -2.0;  //, 0.0, 0.005;
   Eigen::MatrixXd lbm(2, 2);
-  lbm << 1.0, -5.0, 0.0, 100.0;
-
-  double lbd = -5.0;
+  lbm << 7.0, 5.0, 6.0, 100.0;  //, 0.0, 0.0005;
+  lb_constrain_test::expect(A, lbm);
+  double lbd = 6.0;
+  lb_constrain_test::expect(A, lbd);
+}
```
bbbales2, line 51: Should check `lb_constrain(scalar, scalar)`.

test/unit/math/mix/fun/lb_constrain_test.cpp
```diff
@@ -14,52 +16,45 @@ void expect_lb_constrain(double x, double lb) {
...
-  stan::test::expect_ad(f3, A, lbd);
-  stan::test::expect_ad_matvar(f3, A, lbm);
-  stan::test::expect_ad_matvar(f3, A, lbd);
+TEST(mathMixMatFun, lb_mat_constrain_neg_inf) {
+  Eigen::MatrixXd A(2, 2);
+  A << 5.0, 2.0, 4.0, -2.0;
+  Eigen::MatrixXd lbm(2, 2);
+  lbm << 7.0, 5.0, stan::math::NEGATIVE_INFTY, 100.0;
+  lb_constrain_test::expect(A, lbm);
+  lb_constrain_test::expect(A, stan::math::NEGATIVE_INFTY);
```
bbbales2, line 59: Should check `lb_constrain(scalar, scalar)`.

test/unit/math/mix/fun/lb_constrain_matvar_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+  stan::test::expect_ad_matvar(f4, x, lb);
+}
+}  // namespace ub_constrain_test
+
+TEST(mathMixMatFun, lb_matvar_constrain) {
+  using stan::scalar_type_t;
+  using stan::math::lb_constrain;
+  using stan::math::promote_scalar_t;
+  Eigen::MatrixXd A(2, 2);
+  // Doesn't work for values near zero fail for fvar hessian?
```
bbbales2, line 37: Is this still true?

bbbales2 (changes requested):
Review!

stan/math/prim/fun/lb_constrain.hpp
SteveBronder: We need the identity constrain to return back the correct type. So like if x or lb differ in type `x` get's promoted here

stan/math/prim/fun/lb_constrain.hpp
SteveBronder: Yeah good point

stan/math/prim/fun/lb_constrain.hpp
SteveBronder: No idt so

stan/math/rev/fun/identity_constrain.hpp
SteveBronder: I'm not sure what you mean here? We need it when x and lb's types differ, especially when one is a var matrix and the other one is a matrix var since we have promotion rules for that

test/unit/math/prim/fun/lb_transform_test.cpp
SteveBronder: Oh yeah that's a good point

test/unit/math/mix/fun/lb_constrain_test.cpp
SteveBronder: Yeah that's what these do

```
TEST(mathMixScalFun, lbConstrain) {
  lb_constrain_test::expect(-1, 2);
  lb_constrain_test::expect(2, 4);
}
```

stan/math/prim/fun/lb_constrain.hpp
bbbales2: Oooof, I see.

stan/math/rev/fun/identity_constrain.hpp
bbbales2: Yeah, I was hoping to get rid of identity constrain but if it's doing the promotion then we gotta keep.

test/unit/math/mix/fun/lb_constrain_matvar_test.cpp
SteveBronder: Nope!

stan/math/rev/fun/lb_constrain.hpp
SteveBronder: Actually this ended being best case the same speed and worst case a bit slower so I reverted it to just do the checks

stan/math/prim/fun/lb_constrain.hpp
t4c1: If a holder is constructed from lvalues only it is virtually free. We need a malloc for each rvalue. But let me ask again: Do we even need holders here?

SteveBronder:
@rok-cesnovar hey sorry but can we do the reverse of the last constrain pr up in Stan? We just need to revert the changes to the lb/ub constrain checks

bbbales2 (commented):
Couple comments

SteveBronder:
Actually @rok-cesnovar we are going to merge all the changes for the constraints into this PR with sub PRs then once all those are here we can revert the Stan PR

rok-cesnovar:
Coolio!

test/unit/math/prim/fun/lb_transform_test.cpp
```diff
@@ -5,50 +5,23 @@
...
+  EXPECT_FLOAT_EQ(7.9, stan::math::lb_constrain(
+                           7.9, -std::numeric_limits<double>::infinity()));
 }
-
-TEST(prob_transform, lb_vec) {
```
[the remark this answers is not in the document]
SteveBronder: Oh I just reverted back to the old link you sent me lemme fix this up

bbbales2:
Alright I checked and updated the tests and code a bit. Can we ditch the `std::vector` versions of this? I know I asked yesterday but looking closer today it's just a ton of extra code. I mean, it's fine code, but there's a lot of it and there's not advantage over the compiler and it's missing prim tests (that I can add, but again, not a lot of value).

SteveBronder:
So I'm re-doing the IO for `log_prob` right now and it would be nice to for everything if I could just call `*_constrain()`. Like [here](https://github.com/stan-dev/stan/blob/feature/deserializer/src/stan/io/deserializer.hpp#L208)

```cpp
 template <typename Ret, bool Jacobian, typename LB, typename LP, typename... Sizes>
 auto read_lb_constrain(const LB& lb, LP& lp, Sizes... sizes) {
   if (Jacobian) {
     return stan::math::lb_constrain(this->read<Ret>(sizes...), lb, lp);
   } else {
     return stan::math::lb_constrain(this->read<Ret>(sizes...), lb);
   }
 }
```

If not then I have to have seperate methods for std vectors. Not a big deal but I have to deal with it somewhere so it would be nice to just deal with it here

bbbales2:
Oh okay, so what I'm seeing is the compiler component kinda doesn't exist anymore and so we need the `std::vector` stuff here.

SteveBronder:
Yeah my goal is just to let the compiler as for stuff like 

```cpp
std::vector<std::vector<Eigen::Matrix<var, -1, -1>>> A = in__.read_lb(lb, lp, vec1_size, vec2_size, rows, cols);
```

Which should be more performant and make the compilers job a lot easier. But that does mean supporting these signatures somewhere and the Stan math level seems to make sense.

SteveBronder:
I put up [#3012](https://github.com/stan-dev/stan/pull/3012) in the Stan repo. @rok-cesnovar how should we do the double mere thing?

bbbales2:
Okay this should be good to go @rok-cesnovar . It is failing in the upstream tests where we expect. The matching Stan pull is: https://github.com/stan-dev/stan/pull/3012

SteveBronder:
No sorry this is ready for merge I made a false alarm. I thought this was valid Stan code

```stan
parameters {
  real xx[N];
  real<lower=xx,upper=1> theta[N, N];
}
```

but it's not so we r good

```
10:49
--- Translating Stan model to C++ code ---
bin/stanc  --o=examples/bernoulli/bernoulli.hpp examples/bernoulli/bernoulli.stan
Semantic error in 'examples/bernoulli/bernoulli.stan', line 7, column 13 to column 15:
   -------------------------------------------------
     5:  parameters {
     6:    real xx[N];
     7:    real<lower=xx,upper=1> theta[N, N];
                      ^
     8:  }
     9:  model {
   -------------------------------------------------
Lower bound must be a scalar or of type array[,] real. Instead found type array[] real.make: *** [make/program:47: examples/bernoulli/bernoulli.hpp] Error 1
```

bbbales2 (approved):

PR #2396, opened 2021-02-27
offset_multiplier for var<Matrix>

SteveBronder:
## Summary

This adds `var<Matrix>` specializations for `offset_multiplier` (which I'm 99% sure is the last constraint we need). To do this easily I did two things

1. I removed the check for a sigma of 1 and a mu of 0 when doing the call so that the scalar and matrix functions can all go through the same function. I think this actually makes sense as it propogates the adjoints for each of the constraints upwards and if those constraints are vars then they most likely have a lot chance of starting with values of 1 and 0 exactly.

2. I specialized `fma()` so that that one `offset_multiplier()` functions works for all the scalar and matrix types.

## Tests

Though `fma()` is the one overloaded I wrote tests similar to `lub_constrain()` for `offset_multiplier()` testing all the scalar, matrix, and `std::vector` inputs. Since `offset_multiplier()` only calls `fma()` I was thinking that might suffice as tests for the fma specializations. Though if that's false I'm happy to write the tests for `fma()` as well.

## Side Effects

Nope

## Release notes

Adds `fma()` and `offset_multiplier()` specializations for the new matrix type.

## Checklist

- [x] Math issue #1805 


test/unit/math/mix/fun/offset_multiplier_constrain_helpers.hpp
```diff
@@ -0,0 +1,65 @@
...
+  auto f2 = [](const auto& x, const auto& mu, const auto& sigma) {
+    stan::return_type_t<decltype(x), decltype(mu), decltype(sigma)> lp = 0;
+    return stan::math::offset_multiplier_constrain(x, mu, sigma, lp);
+  };
+  auto f3 = [](const auto& x, const auto& mu, const auto& sigma) {
+    stan::return_type_t<decltype(x), decltype(mu), decltype(sigma)> lp = 0;
+    stan::math::offset_multiplier_constrain(x, mu, sigma, lp);
+    return lp;
+  };
+  auto f4 = [](const auto& x, const auto& mu, const auto& sigma) {
```
bbbales2, line 46: I don't think this `f4` tests anything that the other things don't cover. Given the test sizes here are problematic, I want to delete it. Is this okay or am I missing something?

bbbales2 (changes requested):
I need to add some prim tests for offset_multipler_constrain still.

It doesn't look like it was being tested for vector types and there was a problem.

One question on the addition to arena_matrix, and one question on the tests.

test/unit/math/prim/fun/lub_transform_test.cpp
```diff
@@ -182,7 +180,72 @@ TEST(prob_transform, lub_constrain_matrix) {
...
+  // matrix, real, real, lp
+  {
+    double lp0 = 0.0;
+    double lp1 = 0.0;
+    Eigen::VectorXd result = stan::math::lub_constrain(x, lbd, ubd, lp0);
+    for (size_t i = 0; i < result.size(); ++i) {
+      EXPECT_FLOAT_EQ(result(i),
+                      stan::math::lub_constrain(x(i), lbd, ubd, lp1));
+    }
+    EXPECT_FLOAT_EQ(lp0, lp1);
```
bbbales2, line 198: I didn't see prim tests for lub vectorization behavior of lp so I added them here.

test/unit/math/prim/fun/offset_multiplier_transform_test.cpp
```diff
@@ -65,9 +68,489 @@ TEST(prob_transform, offset_multiplier_f) {
...
+  }
+  // matrix, matrix, matrix
+  {
+    Eigen::VectorXd result
+        = stan::math::offset_multiplier_constrain(x, offset, sigma);
+    for (size_t i = 0; i < result.size(); ++i) {
+      EXPECT_FLOAT_EQ(result(i), stan::math::offset_multiplier_constrain(
+                                     x(i), offset(i), sigma(i)));
+    }
+    auto x_free = stan::math::offset_multiplier_free(result, offset, sigma);
```
bbbales2, line 144: There's an `offset_multiplier_free` to go along with `offset_multiplier_constrain`.

These should be prim tests for `offset_multiplier_constrain`. Need some prim error tests for the free too once it's implemented but I was too lazy (the equivalent of [these](https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/fun/lub_transform_test.cpp#L483)).

bbbales2 (changes requested):
@SteveBronder I didn't add fma tests yet. The autodiff tests for `offset_multiplier_constrain` look good. I added prim tests.

Can you add an `offset_multiplier_free`? The values tests are in `test/unit/math/prim/fun/offset_multiplier_transform_test.cpp`, but there might be exception tests to add.

stan/math/prim/fun/offset_multiplier_constrain.hpp
```diff
@@ -77,21 +75,173 @@ inline return_type_t<T, M, S> offset_multiplier_constrain(const T& x,
...
-    return mu + x;
+inline auto offset_multiplier_constrain(const T& x, const M& mu, const S& sigma,
+                                        return_type_t<T, M, S>& lp) {
+  auto&& mu_ref = to_ref(mu);
+  auto&& sigma_ref = to_ref(sigma);
+  check_finite("offset_multiplier_constrain", "offset", value_of_rec(mu_ref));
+  check_positive_finite("offset_multiplier_constrain", "multiplier",
+                        value_of_rec(sigma_ref));
+  lp += sum(multiply_log(size(x), sigma_ref));
+  return fma(sigma_ref, x, mu_ref);
```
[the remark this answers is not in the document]
SteveBronder: Oooof wow yeah this would have been a lot smarter, didn't think of it. I guess since the fma stuff is here already we can keep it though I very much wish I had thought of this

test/unit/math/mix/fun/offset_multiplier_constrain_helpers.hpp
SteveBronder: test 4 just takes the gradient wrt lp and the output. I found it useful when doing the tests for lub since it has a complex gradient. But I'm fine with taking them in or out

SteveBronder:
Yeah sure I can handle the free thing

test/unit/math/prim/fun/offset_multiplier_transform_test.cpp
SteveBronder: I added the transform tests for matrices, lmk how those look!

bbbales2: Cool beans, reviewing the indexing refactor for Stan right now but will get this in a bit.

test/unit/math/prim/fun/offset_multiplier_transform_test.cpp
```diff
@@ -65,9 +68,542 @@ TEST(prob_transform, offset_multiplier_f) {
...
+  Eigen::VectorXd x(4);
+  x << 1.0, 2.0, 3.0, 4.0;
+
+  double mu = 2.0;
+
+  Eigen::MatrixXd sigma(2, 2);
+  sigma << 1.0, 2.0, 3.0, 4.0;
+
+  EXPECT_THROW(stan::math::offset_multiplier_free(x, mu, sigma),
+               std::invalid_argument);
```
bbbales2, line 608: I think everything is here except this. `check_consistent_sizes` makes sure that all arguments are either of the same size or scalar. We need something that checks that all arguments are of the same dimensions or scalar. I'd be fine changing the behavior of `check_consistent_sizes` since I think thus far we're only using it with vectors and it seems fine to expand the use to matrices.

Anyway it needs to be here cause this will blow up with an Eigen dimension mismatch error.

Also since fma is an exposed function ([here](https://mc-stan.org/docs/2_26/functions-reference/composed-functions.html)) we need size checks on it (or users will get Eigen errors when things are mismatches sizes).

All of this applies to lb/ub/lub_constrain as well I think.

bbbales2 (changes requested):
@SteveBronder This is close. There are a few more size check things. I left a comment with details. I'll get this in the AM if you don't get to it tonight.

test/unit/math/prim/fun/offset_multiplier_transform_test.cpp
bbbales2: This is fixed now.

bbbales2: Also lb/ub/lub_constrain handle things correctly already.

bbbales2:
@SteveBronder alright I think all the tests and stuff are here now. I changed some unusual function names we were passing to the check functions in lub_constrain cause I noticed them when I was looking around, so that's what the changes are there.

stan/math/rev/fun/fma.hpp
```diff
@@ -233,19 +171,243 @@ inline var fma(Ta&& a, Tb&& b, const var& c) {
...
+ * @param x First multiplicand.
+ * @param y Second multiplicand.
+ * @param z Summand.
  * @return Product of the multiplicands plus the summand, ($a * $b) + $c.
  */
-template <typename Ta, typename = require_arithmetic_t<Ta>>
-inline var fma(Ta&& a, const var& b, const var& c) {
-  return var(new internal::fma_vdv_vari(b.vi_, a, c.vi_));  // a-b symmetry
+template <typename Ta, require_arithmetic_t<Ta>* = nullptr>
+inline var fma(Ta&& x, const var& y, const var& z) {
```
bbbales2, line 185: Ooof I let all these &&s escape. If I remember I'm gonna switch these to const refs in the future, but it's fine for now.

bbbales2 (approved):
Good

Issue #2197, opened 2020-11-16
Implement closures


bbbales2:
## Description

This is a placeholder issue for implementing closures as described in https://github.com/stan-dev/design-docs/blob/master/designs/0004-closures-fun-types.md

I say placeholder cause that's the main specification, and there will also be associated pull requests in stanc3.

#### Current Version:
v3.3.0

nhuurre:
The stanc3 pull request is open: stan-dev/stanc3#742
Probably has lots of bugs but it's good enough that I can compile Lotka-Volterra model:
```stan
data {
  int<lower = 0> N;          // number of measurement times
  real ts[N];                // measurement times > 0
  vector[2] y_init;            // initial measured populations
  real<lower = 0> y[N, 2];   // measured populations
}
parameters {
  real<lower = 0> alpha;
  real<lower = 0> beta;
  real<lower = 0> gamma;
  real<lower = 0> delta;
  vector<lower = 0>[2] z_init;  // initial population
  real<lower = 0> sigma[2];   // measurement errors
}
model {
  functions
  vector dz_dt(real t,      // time
               vector z     // system state {prey, predator}
               ) {
    real u = z[1];
    real v = z[2];
   
    real du_dt = (alpha - beta * v) * u;
    real dv_dt = (-gamma + delta * u) * v;
  
    return [ du_dt, dv_dt ]';
  }
  vector[2] z[N]
  = ode_bdf(dz_dt, z_init, 0., ts);

  alpha ~ normal(1, 0.5);
  gamma ~ normal(1, 0.5);
  beta ~ normal(0.05, 0.05);
  delta ~ normal(0.05, 0.05);
  sigma ~ lognormal(-1, 1);
  z_init ~ lognormal(log(10), 1);
  for (k in 1:2) {
    y_init[k] ~ lognormal(log(z_init[k]), sigma[k]);
    y[ , k] ~ lognormal(log(z[, k]), sigma[k]);
  }
}
```

bbbales2:
I converted the sir model over [here](https://mc-stan.org/users/documentation/case-studies/boarding_school_case_study.html) to practice with this. (@rok-cesnovar thanks for the binaries)

It is an absolute delight to not worry about passing in parameters or data.

A few comments:

1. I get error when I define a function in the transformed parameters block (example model included down the page and data at the bottom):

```
sir_negbin_lambda.hpp:150:7: error: redefinition of 'sir_L23C9_cfunctor__'
[...]

PR #2397, opened 2021-02-27
Update integrate_1d to use variadic autodiff stuff internally in preparation for closures

bbbales2:
## Summary

This should make it easy to hook up integrate_1d with closures (https://github.com/stan-dev/math/pull/2384)

## Release notes

- integrate_1d internal interface updated in preparation for closures

## Checklist

- [x] Math issue #2197


bbbales2:
@nhuurre this doesn't change the integrate_1d external interface but it should make it easy to hook up in the closure pull.

SteveBronder:
Is this waiting for a review?

bbbales2:
@SteveBronder yeah have at it. It's some changes to make it easy to do closures with integrate_1d.

bbbales2:
Binds removed!

stan/math/rev/functor/integrate_1d.hpp
```diff
@@ -59,6 +76,94 @@ inline double gradient_of_f(const F &f, const double &x, const double &xc,
   return gradient;
 }
 
+/**
+ * Return the integral of f from a to b to the given relative tolerance
+ *
+ * @tparam T_a type of first limit
+ * @tparam T_b type of second limit
+ * @tparam T_theta type of parameters
```
SteveBronder, line 84: Not here anymore, double check the docs

stan/math/rev/functor/integrate_1d.hpp
```diff
@@ -59,6 +76,94 @@ inline double gradient_of_f(const F &f, const double &x, const double &xc,
...
+        num_vars_ab + num_vars_args);
+    double *partials_ptr = partials;
+
+    save_varis(varis, a, b, args...);
+
+    for (size_t i = 0; i < num_vars_ab + num_vars_args; ++i) {
+      partials[i] = 0.0;
+    }
+
+    if (!is_inf(a) && is_var<T_a>::value) {
```
SteveBronder, line 136: [optional] Generally with short circuit logic operators you want the cheaper (in this context compile time known) thing as the first value.

stan/math/rev/functor/integrate_1d.hpp
```diff
@@ -28,25 +28,42 @@ namespace math {
...
   // Run nested autodiff in this scope
   nested_rev_autodiff nested;
 
-  std::vector<var> theta_var(theta_vals.size());
-  for (size_t i = 0; i < theta_vals.size(); i++) {
-    theta_var[i] = theta_vals[i];
-  }
-  var fx = f(x, xc, theta_var, x_r, x_i, msgs);
+  std::tuple<decltype(deep_copy_vars(args))...> args_tuple_local_copy(
+      deep_copy_vars(args)...);
```
SteveBronder, line 49: You can just do

```suggestion
  auto args_tuple_local_copy = std::make_tuple(deep_copy_vars(args)...);
```

SteveBronder (changes requested):
Couple small comments overall code looks good!

One thing, I think we can use a double nested reverse pass to avoid copying the args every iteration. Can you take a look at the code here to see if it makes sense? It's passing all the tests but I'm not super familiar with nested autodiff so I'm not sure if this would break in some edge case we are not testing right now

https://github.com/stan-dev/math/compare/feature/variadic-integrate-1d...review/variadic-integrate-1d#diff-79b7fa556075dab8ffc8caa4be8fc909b1efbcfe4e63d254793a966f63391a1cR143

stan/math/rev/functor/integrate_1d.hpp
```diff
@@ -28,25 +28,42 @@ namespace math {
...
-  gradient = theta_var[n].adj();
+
+  apply(
+      [&](auto &&... args) {
+        accumulate_adjoints(adjoints.data(),
+                            std::forward<decltype(args)>(args)...);
+      },
+      std::move(args_tuple_local_copy));
+
+  gradient = adjoints.coeff(n);
```
SteveBronder, line 66: I forgot to make this comment in my review, but optionally we can definitely figure out which arg has the adjoint value we need here in a clever way that doesn't require copying all of them. We could have some function to get the Nth var in a tuple but I'm also fine with not doing that in this PR

SteveBronder: We could have some internal function like 
```cpp
template <typename... Args>
double get_nth_adjoint(size_t n, const std::tuple<Args...>& tuple_arg) {
   size_t accum_vars = 0;
   bool stop_checking = false;
   // for_each goes off from left to right
   std::array<double, sizeof...(Args)> possible_adjs = for_each([&accum_vars, &stop_checking](auto&& arg){
      if (stop_checking) return 0.0;
      size_t num_vars = count_vars(arg);
      // Need to keep moving along
      if ((accum_vars + num_vars) < nth || stop_checking) {
         accum_vars += num_vars;
         return 0.0;
      } else { // We reached the first arg that passes
         stop_checking = true;
         // I'm tired but do the logic here to get the nth value from that particular arg
         return the_adj;
      } 
   }, tuple_arg);
   // Loop over possible_adjs until we hit a nonzero value (or they are all zero and return a zero)
};
```

stan/math/rev/functor/integrate_1d.hpp
```diff
@@ -59,6 +76,95 @@ inline double gradient_of_f(const F &f, const double &x, const double &xc,
...
+  check_less_or_equal(function, "lower limit", a, b);
+
+  double a_val = value_of(a);
+  double b_val = value_of(b);
+
+  if (a_val == b_val) {
+    if (is_inf(a_val)) {
+      throw_domain_error(function, "Integration endpoints are both", a_val, "",
+                         "");
+    }
```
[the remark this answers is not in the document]
bbbales2: It's okay if a single endpoint is at infinity (so integral on (-infinity, b) or (a, infinity) are both fine)

stan/math/prim/functor/integrate_1d.hpp
```diff
@@ -131,6 +134,39 @@ inline double integrate(const F& f, double a, double b,
...
+ * @param args additional arguments passed to f
+ * @return numeric integral of function f
+ */
+template <typename F, typename... Args,
+          require_all_not_st_var<Args...>* = nullptr>
+inline double integrate_1d_impl(const F& f, double a, double b,
+                                double relative_tolerance, std::ostream* msgs,
+                                const Args&... args) {
+  static const char* function = "integrate_1d";
+  check_less_or_equal(function, "lower limit", a, b);
```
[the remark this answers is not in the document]
bbbales2: Yeah if they're equal and not infinity then the result is just zero.

bbbales2:
> One thing, I think we can use a double nested reverse pass to avoid copying the args every iteration.

I think this is right. I didn't like doing a weird thing with nested autodiff across functions like that, so I just put the contents of `gradient_of_f` inline and got rid of that function.

stan/math/rev/functor/integrate_1d.hpp
bbbales2: Yeah it's awkward. I also want to leave it for now.

The way to speed this up is writing our own quadratures or talking to the Boost people about an interface where we integrate multiple functions on the same domain together. Now we compute all three of these integrals totally separately:

```
\int f(x, a, b) dx
\int df(x, a, b)/da
\int df(x, a, b)/db
```

But anytime we compute df(x, a, b)/da we also get df(x, a, b)/db, and so the efficiency gains would be taking advantage of that (and what we're doing here is throwing away a ton of gradient info).

stan/math/prim/functor/integrate_1d.hpp
```diff
@@ -57,6 +58,8 @@ inline double integrate(const F& f, double a, double b,
   bool used_two_integrals = false;
   size_t levels;
   double Q = 0.0;
+  // if a or b is infinite, set xc argument to NaN (see docs above for user
+  // function for xc info)
```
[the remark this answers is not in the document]
bbbales2: I split these into lambda functions. That look better or is used_two_integrals clearer?

SteveBronder (approved):
I think it looks good! I had a few small comments I just put into a PR that you can merge if you like. I also figured out how to do the thing where we only grab the one adjoint we need instead of making a vector and copying all the adjoints into it each iteration.

https://github.com/stan-dev/math/compare/feature/variadic-integrate-1d...review/integrate-1d-variadic-2

bbbales2:
@SteveBronder I merged the stuff. There was another way to do the nth adjoint thing that was less code so I went with that so we should be fast

SteveBronder (approved):
I reverted the changes I made to get so this should be good now

PR #2398, opened 2021-03-01
Add quantile function

adamhaber:
## Summary

Addresses #1036 . 

Still WIP - points I'd be happy to get feedback on:
- Currently only implemented for `std::vector` - I'm sure adding an implementation for Eigen vectors/row vectors is straightforward, but still I'm a C++ newcomer and would appreciate an example or a simple explanation on how to do this.
- ATM the quantiles are computes in a very simplistic way, that is - no extrapolation at all. As discussed in the original issue, R has different quantile algorithms - do we want to be consistent the default one (`type==7`), which _does_ extrapolate?
- Can the current implementation be made more efficient?
- Does the current implementation makes sense in light of the AD discussion from the original issue?

## Tests

Added very basic tests to see that p=1 returns max, p=0 returns min and some sensible values in between. Will be extended after adding Eigen types.

## Side Effects

No.

## Release notes

Adds quantile function.

## Checklist

- [ ] Math issue #1036

- [ ] Copyright holder: Adam Haber


bbbales2 (changes requested):
Let's switch to algorithm 7 so we're consistent with default R. I think the efficiency here is fine. We can add a signature like:

```
quantile(vector, vector) => vector
```

And that'll probably be about as efficient as it gets.

bbbales2:
Uh oh ignore the algorithm 7 thing.

Looks like the quantile functions we use in stan are boost: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/chains.hpp#L112

We should probably match those quantile functions before we favor the R ones.

The simplest form of this is probably move those quantile estimators to Math (so they're implemented here).

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * @tparam T Type of elements contained in vector.
+ * @param xs Numeric vector whose sample quantiles are wanted
+ * @param p Probability with value between 0 and 1.
+ * @return Sample quantile.
+ * @throw std::domain_error If any of the values are NaN.
+ */
+template <typename T>
+inline T quantile(const std::vector<T>& xs, double p) {
+  check_not_nan("quantile", "container argument", xs);
+  check_bounded("sort_asc", "p", p, 0, 1);
```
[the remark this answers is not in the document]
adamhaber: Haha busted :-) 

Changed.

adamhaber:
OK, updated the function to according to https://github.com/stan-dev/stan/blob/723f86bdf4275be8929e37d4ad80a6d0a6426bf0/src/stan/mcmc/chains.hpp#L112 .

However, I'm not sure how to update the _tests_ since I'm not sure how the boost implementation works. In here:
https://github.com/stan-dev/stan/blob/723f86bdf4275be8929e37d4ad80a6d0a6426bf0/src/test/unit/mcmc/chains_test.cpp#L597
it seems like something is being compared to R's quantile function, but this is strange as the current implementation doesn't interpolate:

```c++
Eigen::VectorXd v(5);
  
v[0] = 1.0;
v[1] = 2.0;
v[2] = 3.0;
v[3] = -1.0;
v[4] = 19.3;

std::cout << quantile(v, 0.0001) << std::endl;
std::cout << quantile(v, 0.1) << std::endl;
std::cout << quantile(v, 0.2) << std::endl;
std::cout << quantile(v, 0.3) << std::endl;
std::cout << quantile(v, 0.4) << std::endl;
std::cout << quantile(v, 0.5) << std::endl;
std::cout << quantile(v, 0.6) << std::endl;
std::cout << quantile(v, 0.7) << std::endl;
std::cout << quantile(v, 0.8) << std::endl;
std::cout << quantile(v, 0.9) << std::endl;
std::cout << quantile(v, 0.9999) << std::endl;
```
```
-1
-1
-1
1
1
2
3
3
19.3
19.3
19.3
```
where's R does:
```
> quantile(c(1,2,3,-1,19.3),c(0.0001,1:9/10,0.999))
  0.01%     10%     20%     30%     40%     50%     60%     70%     80%     90%   99.9% 
-0.9992 -0.2000  0.6000  1.2000  1.6000  2.0000  2.4000  2.8000  6.2600 12.7800 19.2348 
```
Also, I think the boost implementation doesn't handle p=0 and p=1 - should we?

bbbales2:
@adamhaber ooof, this tail_quantile thing is doing the most basic thing: https://www.boost.org/doc/libs/1_75_0/boost/accumulators/statistics/tail_quantile.hpp

And you are right the boundary behavior seems bad.

Now I want to do the R implementation again. Sorry for the switcheroo, I just got excited when I saw the boost implementation. Let's check if anyone has any attachment to the boost thing in stan. Maybe it is done there for performance or something.

bbbales2:
@bgoodri you know of any reason we're using the Boost quantile function [here](https://www.boost.org/doc/libs/1_75_0/boost/accumulators/statistics/tail_quantile.hpp) other than convenience? Adam was implementing a quantile function for Stan and we were deciding on implementation. That one seems questionable cause it doesn't even work with p = 0 or p = 1.

Now I am back to wanting to just copy the R default.

bgoodri:
No reason

bbbales2:
I assigned the output of `as_array_or_scalar` to a new temporary.

`as_array_or_scalar` returns a view into the input, which in this case is a const reference so we can't modify the values with std::sort. Anyway, copying to a new variable we're good.

I think there might need to be some edge case checks here -- like hwat happens with size zero input.

adamhaber:
Thanks @bbbales2 , both for the fix and for the explanation. :-) Added tests for the different input kinds.

Two more questions:
1. Do we want to add a quantile function that takes a vector of ps (like in R)?
2. Regarding zero-size inputs - what is the desired behavior? Should I test for something like this:
https://github.com/stan-dev/math/blob/2cf4310702eec1b14a355d883e704af0b788abd8/test/unit/math/prim/fun/sort_test_util.hpp#L12-L13
?

test/unit/math/prim/fun/quantile_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+TEST(MathFunctions, quantileEigenVectorXdStdVecDouble) {
+  test_quantile_double<Eigen::VectorXd, std::vector<double>>();
+}
+
+TEST(MathFunctions, quantileEigenRowVectorXdStdVecDouble) {
+  test_quantile_double<Eigen::RowVectorXd, std::vector<double>>();
+}
+
+TEST(MathFunctions, quantileStdVecDoubleVectorXd) {
```
bbbales2, line 112: I don't think we need the tests here onwards -- let's just say ps needs to be an array and then we return arrays.

test/unit/math/prim/fun/quantile_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+  EXPECT_FLOAT_EQ(quantile(v, 0.8), 3.8);
+  EXPECT_FLOAT_EQ(quantile(v, 0.9), 4.4);
+  EXPECT_FLOAT_EQ(quantile(v, 1), 5.0);
+}
+
+template <typename T, typename Tp, stan::require_all_vector_t<T, Tp>* = nullptr>
+void test_quantile_double() {
+  using stan::math::index_type_t;
+  using stan::math::quantile;
+
```
bbbales2, line 78: Test that size zeros fail here, both in the first and second argument.

Also test a size 1 first argument works right.

Also test that the function throws if the second argument is outside [0, 1].

Also check nans throw.

test/unit/math/prim/fun/quantile_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+  T c0(0);
+  EXPECT_THROW(quantile(c0, 0.3), std::invalid_argument);
+
+  T c(1);
+  c[0] = 1.7;
+  EXPECT_EQ(quantile(c, 0), 1.7);
+  EXPECT_EQ(quantile(c, 1), 1.7);
+  EXPECT_EQ(quantile(c, 0.33), 1.7);
+  EXPECT_EQ(quantile(c, 0.68), 1.7);
+
```
bbbales2, line 21: Check that the function throws if the second argument is outsize [0, 1].

Also check nans throw.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,94 @@
...
+  size_t hi = std::ceil(index);
+
+  std::sort(x.data(), x.data() + n_sample);
+
+  double q = x[lo];
+  double h = index - lo;
+  return (1 - h) * q + h * x[hi];
+}
+
+template <typename T, typename Tp, require_all_vector_t<T, Tp>* = nullptr>
```
bbbales2, line 55: Docs (I know you said you'd add these just leaving a note so it's easy to remember)

bbbales2 (changes requested):
Looks good! I left some comments.

I added some stuff about making this work with autodiff but decided to remove it. I forget if we already talked about this, but I don't think the gradients are super well defined either for the xs or the ps.

I assume this is most useful in generated quantities too. If in the future people decide it must be an autodiff function (regardless of the gradient problems), then it shouldn't be too hard to change.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  if (n_sample == 1)
+    return ret + x[0];
+
+  std::sort(x.data(), x.data() + n_sample);
+  Eigen::ArrayXd index = (n_sample - 1) * p;
+
+  for (size_t i = 0; i < n_ps; ++i) {
+    if (p[i] == 0.)
+      ret[i] = *std::min_element(x.data(), x.data() + n_sample);
```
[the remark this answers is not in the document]
adamhaber: Nice, changed.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  std::sort(x.data(), x.data() + n_sample);
+
+  double q = x[lo];
+  double h = index - lo;
+  return (1 - h) * q + h * x[hi];
+}
+
+template <typename T, typename Tp, require_all_vector_t<T, Tp>* = nullptr>
+inline Eigen::ArrayXd quantile(const T& xs, const Tp& ps) {
```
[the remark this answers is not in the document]
adamhaber: Got it, changed.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+  double index = (n_sample - 1) * p;
+  size_t lo = std::floor(index);
+  size_t hi = std::ceil(index);
+
+  std::sort(x.data(), x.data() + n_sample);
+
+  double q = x[lo];
+  double h = index - lo;
+  return (1 - h) * q + h * x[hi];
```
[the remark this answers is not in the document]
adamhaber: Thanks! Really love to learn this stuff here. :-)

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,94 @@
...
+  check_not_nan("quantile", "container argument", xs);
+  check_bounded("quantile", "ps", ps, 0, 1);
+  check_nonzero_size("quantile", "xs", xs);
+  check_nonzero_size("quantile", "ps", ps);
+  size_t n_sample = xs.size();
+  size_t n_ps = ps.size();
+
+  Eigen::VectorXd x = as_array_or_scalar(xs);
+  Eigen::ArrayXd p = as_array_or_scalar(ps);
+  Eigen::ArrayXd ret = Eigen::ArrayXd::Zero(n_ps);
```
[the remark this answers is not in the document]
adamhaber: why is this better than `std::vector<double> ret(n_ps, 0.0);`?

test/unit/math/prim/fun/quantile_test.cpp
adamhaber: Changed.

stan/math/prim/fun/quantile.hpp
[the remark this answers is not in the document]
bbbales2: Either will be fine here. Whatever is more readable to you is good.

For fancy things, the cost of initializing the vector is like allocating memory for and then filling up the vector. Reserve is just allocating memory for. So if the cost of filling up the vector matters (it won't here), you'd favor reserve.

Also depending on what you're storing, you'll run a bunch of constructors to fill up the array and you might not want that because there might be side effects of the constructor. Something like that.

test/unit/math/prim/fun/quantile_test.cpp
adamhaber: > Test that size zeros fail here, both in the first and second argument.

There's
```
T c0(0);
EXPECT_THROW(quantile(c0, 0.3), std::invalid_argument);
```
for the first argument - is this ok? Not sure how to check zero size for the double.

> Also test a size 1 first argument works right.

There's:
```
T c(1);
  c[0] = 1.7;
  EXPECT_EQ(quantile(c, 0), 1.7);
  EXPECT_EQ(quantile(c, 1), 1.7);
  EXPECT_EQ(quantile(c, 0.33), 1.7);
  EXPECT_EQ(quantile(c, 0.68), 1.7);
```
is this ok?

> Also test that the function throws if the second argument is outside [0, 1].
Also check nans throw.

Added.

adamhaber: Oh all these are for the vector ps right? Sorry! Will add

bbbales2: This comment is on the version for vector ps (where size zero isn't currently checked and it's possible to have a size zero ps :D)

bbbales2: > these are for the vector ps right?

Yeah yeah -- didn't see the follow up before I posted

stan/math/prim/fun/quantile.hpp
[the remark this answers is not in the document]
adamhaber: Perfect, thanks!

stan/math/prim/fun/quantile.hpp
adamhaber: done (I hope)

test/unit/math/prim/fun/quantile_test.cpp
adamhaber: thanks, added

test/unit/math/prim/fun/quantile_test.cpp
adamhaber: Got it, fixed

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+namespace stan {
+namespace math {
+
+/**
+ * Return sample quantiles corresponding to the given probabilities.
+ * The smallest observation corresponds to a probability of 0 and the largest to
+ * a probability of 1.
+ *
+ * Implementation follows the default R behavior, as defined here:
+ * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/quantile
```
bbbales2, line 20: Say it is algorithm 7 from Hyndman, R. J. and Fan, Y., Sample quantiles in statistical packages which should be the default quantile implementation in R. rdocumentation link might go outta date in a few years and this will hang around.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @tparam T Type of elements contained in vector.
```
bbbales2, line 22: T is the container itself

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+ * @param xs Numeric vector whose sample quantiles are wanted
+ * @param p Probability with value between 0 and 1.
+ * @return Sample quantile.
+ * @throw std::invalid_argument If any of the values are NaN or size 0.
```
bbbales2, line 26: If any element of xs is NaN. I guess if p was NaN it wouldn't be invalid_argument.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+  check_not_nan("quantile", "p", p);
+  check_bounded("quantile", "p", p, 0, 1);
+
+  size_t n_sample = xs.size();
+
+  Eigen::VectorXd x = as_array_or_scalar(xs);
+
+  if (n_sample == 1)
+    return x[0];
```
bbbales2, line 42:
```suggestion
    return x.coeff(0);
```

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+  return (1 - h) * x.coeff(lo) + h * x.coeff(hi);
+}
+
+/**
+ * Return sample quantiles corresponding to the given probabilities.
+ * The smallest observation corresponds to a probability of 0 and the largest to
+ * a probability of 1.
+ *
+ * Implementation follows the default R behavior, as defined here:
+ * https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/quantile
```
bbbales2, line 64: Update to whatever you do above

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @tparam T Type of elements contained in vector.
```
bbbales2, line 66: T is the vector itself, also add a tparam for Tp

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,112 @@
...
+  check_nonzero_size("quantile", "xs", xs);
+  check_nonzero_size("quantile", "ps", ps);
+
+  check_bounded("quantile", "ps", ps, 0, 1);
+
+  size_t n_sample = xs.size();
+  size_t n_ps = ps.size();
+
+  Eigen::VectorXd x = as_array_or_scalar(xs);
+  Eigen::ArrayXd p = as_array_or_scalar(ps);
```
bbbales2, line 87:
```suggestion
  const auto& p = as_array_or_scalar(ps);
```

Don't need to actually copy p here since we're not modifying it

bbbales2 (changes requested):
Looks good. Some small cleanup stuff and this can merge. I had a look at the docs and it looks like mean/sd/variance are defined for arrays and vectors, so I guess we should enable all six signatures here.

[Here](https://mc-stan.org/docs/2_26/functions-reference/array-reductions.html) we can add:

```stan
real quantile(data real[] xs, data real p)
real[] quantile(data real[] xs, data real[] p)
```

And then [here](https://mc-stan.org/docs/2_26/functions-reference/reductions.html) we can add:

```stan
real quantile(data vector xs, data real p)
real[] quantile(data vector xs, data real[] p)
real quantile(data row_vector xs, data real p)
real[] quantile(data row_vector xs, data real[] p)
```

Will also need a PR to stanc3 for these signatures

stan/math/prim/fun/quantile.hpp
adamhaber: Done

stan/math/prim/fun/quantile.hpp
adamhaber: Done

adamhaber:
@bbbales2 what about 
```stan
real quantile(data int[] xs, data real p)
real[] quantile(data int[] xs, data real[] ps)
```

Also, at the moment `ps` can also be a vector/row_vector - would you like me to change that?

bbbales2:
> at the moment ps can also be a vector/row_vector - would you like me to change that?

I think it's better to just leave it as an array. The Math function supports it I guess, but that doesn't mean we have to expose the signature.

Originally Stan tried to keep arrays/vectors separate. Like when you wanted to do vector math you did vectors and otherwise arrays, so it isn't a requirement that everywhere we support arrays we also do vectors and row vectors.

I thought it would be good to support array/vector/row_vector for the first argument since min/max/mean/sd/variance support array/vector/row_vector arguments, but no need to extend that support to the second argument.

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * @tparam T Type of input vector
+ * @param xs Numeric vector whose sample quantiles are wanted
+ * @param p Probability with value between 0 and 1.
+ * @return Sample quantile.
+ * @throw std::invalid_argument If any element of xs is NaN, or size 0.
+ * @throw std::domain_error If p<0 or p>1.
+ */
+template <typename T, require_vector_t<T>* = nullptr>
+inline double quantile(const T& xs, const double p) {
```
rok-cesnovar, line 30: Should this only allow arithmetic types (non-var)?

rok-cesnovar:
@adamhaber Is this ready for one last look?

adamhaber:
I think so, execpt I'm not sure about
> Should this only allow arithmetic types (non-var)?

(not sure if it should, and which `require_` is required if it should)

rok-cesnovar:
Given that this function is supposed to only work with data in Stan, I think it should be required to be non-var in Stan Math. I dont think we have an official policy on that though. The suitable require is `require_arithmetic_t` I think.

adamhaber:
Hmm, `require_arithmetic_t` fails with
```
./stan/math/prim/fun/quantile.hpp:30:15: note: candidate template ignored: requirement 'is_arithmetic<std::__1::vector<double, std::__1::allocator<double>>>::value' was not satisfied [with T = std::__1::vector<double, std::__1::allocator<double>>, $1 = nullptr]
inline double quantile(const T& xs, const double p) {
```
Any idea?

rok-cesnovar:
For containers like std::vector, we should require the scalar type to be arithmetic. Paging @SteveBronder who would probably have an idea for the simplest require to use here. The bottom line is that we want all scalars or containers to be non-var.

t4c1:
You can use `require_vector_vt<std::is_arithmetic, T>`.

adamhaber:
Thanks @t4c1 ! 

@rok-cesnovar - ready for review, I think.

rok-cesnovar:
Cool! Will let @bbbales2 review when he has time. If you wont have time in the near future, let me know Ben.

stan/math/prim/fun/quantile.hpp
SteveBronder: You can also swap this for `require_vector_like_vt<>` and it should work for both `std::vector<Arithmetic>` and `Eigen::Matrix<Arithmetic>`

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Return sample quantiles corresponding to the given probabilities.
+ * The smallest observation corresponds to a probability of 0 and the largest to
+ * a probability of 1.
+ *
+ * Implements algorithm 7 from Hyndman, R. J. and Fan, Y.,
+ * Sample quantiles in Statistical Packages (R's default quantile function)
+ *
+ * @tparam T Type of input vector
```
SteveBronder, line 22: It's good to be a little more through like 
```suggestion
 * @tparam T An Eigen type with either fixed rows or columns at compile time.
```

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+ * @param xs Numeric vector whose sample quantiles are wanted
```
SteveBronder, line 23: [optional] These names are fine though I tend to prefer more informative names like `dist_vec` or something like that since I then have to remember less when I'm reading where it's used

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+  size_t n_sample = xs.size();
+
+  Eigen::VectorXd x = as_array_or_scalar(xs);
+
+  if (n_sample == 1)
+    return x.coeff(0);
+  if (p == 0.)
+    return *std::min_element(x.data(), x.data() + n_sample);
+  if (p == 1.)
+    return *std::max_element(x.data(), x.data() + n_sample);
```
SteveBronder, line 47: I'd use `if else`, can sometimes help the compiler with control flow

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Implements algorithm 7 from Hyndman, R. J. and Fan, Y.,
+ * Sample quantiles in Statistical Packages (R's default quantile function)
+ *
+ * @tparam T Type of input vector
+ * @param xs Numeric vector whose sample quantiles are wanted
+ * @param p Probability with value between 0 and 1.
+ * @return Sample quantile.
+ * @throw std::invalid_argument If any element of xs is NaN, or size 0.
+ * @throw std::domain_error If p<0 or p>1.
```
SteveBronder, line 27:
```suggestion
 * @throw std::domain_error If `p < 0` or `p > 1`.
```

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+  if (p == 0.)
+    return *std::min_element(x.data(), x.data() + n_sample);
+  if (p == 1.)
+    return *std::max_element(x.data(), x.data() + n_sample);
+
+  double index = (n_sample - 1) * p;
+  size_t lo = std::floor(index);
+  size_t hi = std::ceil(index);
+
+  std::sort(x.data(), x.data() + n_sample);
```
SteveBronder, line 53: If you sort once after making `x` can you just do

```
 if (p == 0.)
    return x[0];
  if (p == 1.)
    return x[n_sample];
```

SteveBronder, line 53: [optional] doubt the default behavior would ever change but could include `<functional>` and just add the sort type you want

```cpp
  std::sort(x.data(), x.data() + n_sample, std::less());
```

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+ * Sample quantiles in Statistical Packages (R's default quantile function)
+ *
+ * @tparam T Type of input vector
+ * @tparam Tp Type of probabilities vector
+ * @param xs Numeric vector whose sample quantiles are wanted
+ * @param ps Vector of probability with value between 0 and 1.
+ * @return Sample quantiles, one for each p in ps.
+ * @throw std::invalid_argument If any of the values are NaN or size 0.
+ * @throw std::domain_error If p<0 or p>1 for any p in ps.
+ */
```
SteveBronder, line 74: Same comments for docs as above

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+  check_nonzero_size("quantile", "ps", ps);
+
+  check_bounded("quantile", "ps", ps, 0, 1);
+
+  size_t n_sample = xs.size();
+  size_t n_ps = ps.size();
+
+  Eigen::VectorXd x = as_array_or_scalar(xs);
+  const auto& p = as_array_or_scalar(ps);
+  std::vector<double> ret(n_ps, 0.0);
```
SteveBronder, line 92: I think you can set the return type to `auto` and then base the return on the input here

```suggestion
  plain_type_t<T> ret(n_ps);
  // Fill with zeros but you don't need to
  fill(ret, 0);
```

So then you can have `Eigen::VectorXd` inputs giving back eigen vectors and `std::vector` inputs giving back `std::vectors`

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+  check_not_nan("quantile", "xs", xs);
+  check_not_nan("quantile", "ps", ps);
+
+  check_nonzero_size("quantile", "xs", xs);
+  check_nonzero_size("quantile", "ps", ps);
+
+  check_bounded("quantile", "ps", ps, 0, 1);
+
+  size_t n_sample = xs.size();
+  size_t n_ps = ps.size();
```
SteveBronder, line 88: [optional] When I pull out sizes I like to make them `const`, but also fine as is

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,116 @@
...
+  std::vector<double> ret(n_ps, 0.0);
+
+  std::sort(x.data(), x.data() + n_sample);
+  Eigen::ArrayXd index = (n_sample - 1) * p;
+
+  for (size_t i = 0; i < n_ps; ++i) {
+    if (p[i] == 0.)
+      ret[i] = x.coeff(0);
+    if (p[i] == 1.)
+      ret[i] = x.coeff(n_sample - 1);
```
SteveBronder, line 101: if else

SteveBronder (commented):
Just some comments, looks good!

stan/math/prim/fun/quantile.hpp
adamhaber: Isn't running these checks before sorting more efficient? In the sense that either is true then we save the sorting which is more expensive?

stan/math/prim/fun/quantile.hpp
adamhaber: I think this was @bbbales2 suggestion - what are the pros/cons of returning T vs. always returning `std::vector`? Or doesn't matter?

Also - is there a difference between `plain_type_t<T> ret(n_ps);` and `T ret(n_ps);`?

stan/math/prim/fun/quantile.hpp
adamhaber: I got:
```
./stan/math/prim/fun/quantile.hpp:54:49: error: use of class template 'less' requires template arguments
  std::sort(x.data(), x.data() + n_sample, std::less());
```
so replaced to 
```cpp
std::sort(x.data(), x.data() + n_sample, std::less<double>());
```
is this ok?

stan/math/prim/fun/quantile.hpp
SteveBronder: `T` in this case can be an eigen expression such as `Eigen::Add<Vec, Vec>` so using `plain_type_t<>` gives you Eigen's `PlainObjectBase` aka something like `Eigen::VectorXd`

stan/math/prim/fun/quantile.hpp
SteveBronder: Yep!

SteveBronder: Oh good point! Returning immediately so makes sense

stan/math/prim/fun/quantile.hpp
adamhaber: OK, discussed this with @bbbales2 again - returning only std::vector is indeed intentional (and the argument type for ps should also only be a std vector). This is also what's currently exposed at stanc3 (https://github.com/stan-dev/stanc3/pull/870): 
```
quantile(vector, real) => real
quantile(row_vector, real) => real
quantile(array[] real, real) => real
quantile(vector, array[] real) => array[] real
quantile(row_vector, array[] real) => array[] real
quantile(array[] real, array[] real) => array[] real
```
@SteveBronder do you think differently?

adamhaber:
`make cpplint` fails with:
```
Done processing stan/math.hpp
Category 'readability/braces' errors found: 1
Total errors found: 1
make/cpplint:5: recipe for target 'cpplint' failed
make: *** [cpplint] Error 1
script returned exit code 2
```
Is there a way to get a more informative error message from cpplint (which braces)?

rok-cesnovar:
Scroll a bit higher and you will find: 

```
Done prostan/math/prim/fun/quantile.hpp(104): error cpplint: [readability/braces] If an else has a brace on one side, it should have it on both [5]
```

So see line 104.

stan/math/prim/fun/quantile.hpp
SteveBronder: Ah! Because our rng functions do the same. Yes cool I think that makes sense

stan/math/prim/fun/quantile.hpp
```diff
@@ -0,0 +1,119 @@
...
+ */
+template <typename T, typename Tp, require_all_vector_t<T, Tp>* = nullptr,
+          require_vector_vt<std::is_arithmetic, T>* = nullptr,
+          require_vector_vt<std::is_arithmetic, Tp>* = nullptr>
+inline std::vector<double> quantile(const T& samples_vec, const Tp& ps) {
+  check_not_nan("quantile", "samples_vec", samples_vec);
+  check_not_nan("quantile", "ps", ps);
+
+  check_nonzero_size("quantile", "samples_vec", samples_vec);
+  check_nonzero_size("quantile", "ps", ps);
```
SteveBronder, line 85: Standard for zero sized vectors is just to return the input vector. Generally we try not to fail on zero sized things. There's cases like in rstanarm where we can have size zero vectors

test/unit/math/prim/fun/quantile_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+  p[0] = 0;
+  p[1] = 0.1;
+  p[2] = 0.2;
+  p[3] = 1;
+
+  std::vector<double> ret = quantile(v, p);
+  EXPECT_FLOAT_EQ(ret[0], -0.28);
+  EXPECT_FLOAT_EQ(ret[1], -0.196);
+  EXPECT_FLOAT_EQ(ret[2], -0.112);
+  EXPECT_FLOAT_EQ(ret[3], 2.65);
```
SteveBronder, line 110: Just as a sanity check can you do these in a different order like 

```suggestion
  Tp p(4);
  p[0] = 1;
  p[1] = 0.1;
  p[2] = 0;
  p[3] = 0.2;

  std::vector<double> ret = quantile(v, p);
  EXPECT_FLOAT_EQ(ret[0], 2.65);
  EXPECT_FLOAT_EQ(ret[1], -0.196);
  EXPECT_FLOAT_EQ(ret[2], -0.28);
  EXPECT_FLOAT_EQ(ret[3], -0.112);
```

SteveBronder (changes requested):
Two little comments then I think this is good to go!

stan/math/prim/fun/quantile.hpp
adamhaber: OK, so something like?
```cpp
const size_t n_ps = ps.size();
if (n_ps == 0) 
    return {}
```
Sorry if this is super-basic, learning C++ as I go, here... :-)

SteveBronder: Yep!

SteveBronder: And no worries! C++ can be real weird

adamhaber: Thanks!

bbbales2 (dismissed):

stan/math/prim/fun/quantile.hpp
adamhaber: Hmm, when I comment out `check_nonzero_size("quantile", "ps", ps);` I fail this test:
```cpp
EXPECT_THROW(quantile(v, std::numeric_limits<Tp>::quiet_NaN()),
               std::invalid_argument);
```
with
```
Expected: quantile(v, std::numeric_limits<Tp>::quiet_NaN()) throws an exception of type std::invalid_argument.
  Actual: it throws nothing.
```
shouldn't `check_not_nan("quantile", "ps", ps);` catch this? Or

SteveBronder: It was easier to show then tell alone

https://github.com/stan-dev/math/tree/review/quantile-funcs

But essentially [`std::numeric_limits<T>::quiet_NaN()`](https://en.cppreference.com/w/cpp/types/numeric_limits) is only defined for floating point types so you can't pass a vector to it.

adamhaber: Thanks! Merged.

SteveBronder (approved):
looks good!

Issue #989, opened 2018-08-15
Functions in prim/mat/fun that could use custom reverse mode autodiff implementations


bbbales2:
## Description
The functions listed here currently do not have custom reverse mode autodiff implementations.

Some of these functions would be easy and some would be hard. I'll be piecemeal working on sections of this pull in the coming weeks, so you should post a message here if you want to work on any of these so we'll stay coordinated.

The main reason many functions are included in this list is that they create a lot of varis on the chaining stack.

A lot of not-included functions (rep_matrix, sort_asc, fill, etc.) were not included because they didn't have large computational components

* [ ] add
* [ ] add_diag
* [ ] csr_matrix_times_vector
* [ ] csr_to_dense_matrix
* [ ] cumulative_sum
* [ ] diag_pre_multiply
* [ ] diag_post_multiply
* [ ] divide
* [ ] elt_multiply
* [ ] elt_divide
* [ ] gp_dot_prod_cov
* [ ] gp_exp_quad_cov
* [ ] gp_matern52_cov
* [ ] gp_periodic_cov
* [ ] inverse
* [ ] log_softmax
* [ ] make_nu
* [ ] mdivide_right
* [ ] mdivide_right_ldlt
* [ ] mdivide_right_spd
* [ ] mdivide_right_tri
* [ ] minus
* [ ] subtract
* [ ] quad_form_diag
* [ ] cholesky_corr_constrain
* [ ] cholesky_factor_constrain
* [ ] corr_matrix_constrain
* [ ] cov_matrix_constrain
* [ ] cov_matrix_constrain_lkj

There are a couple other functions that could have custom autodiff but the math would not be easy mathematically:
* [ ] eigenvalues_sym
* [ ] eigenvectors_sym
* [ ] singular_values

## Current Math Version
v2.18.0

andrjohns:
@bbbales2 which of these have you implemented so far? I can work on some of the others

bbbales2:
I haven't done any of them, but there are updates! I'll update the issue.

I did a test with the autodiff stuff and I think I should get some of them off the list cause the default implementations are going to be fast (https://discourse.mc-stan.org/t/adj-jac-apply/5163/6). That'd be like add, add_diag, elt_multiply, elt_divide, minus, subtract. These are the functions where I assumed that reducing the numbers of varis on the stack by using adj_jac_apply would help. I don't think that is true anymore.

The gp functions might have custom reverse modes now. That should be checked.

[...]

PR #2405, opened 2021-03-03
Implemented reverse mode for diag_pre_multiply

gregorp90:
## Summary

Implemented custom reverse mode AD for diag_pre_multiply() function. The implementation is roughly based on the reverse mode implementation of elt_multiply().

## Tests

Added a test for prim where the matrix parameter is non-square.

Added 4 AD tests where the matrix parameter is non-square.

## Side Effects

None.

## Release notes

Added custom reverse mode for diag_pre_multiply() function.

## Checklist

- [x] Math issue #989

- [x] Copyright holder: Gregor Pirš


stan/math/rev/fun/diag_pre_multiply.hpp
```diff
@@ -0,0 +1,92 @@
...
+                                    .array()
+                                    .cwiseProduct(ret.adj().array())
+                                    .rowwise()
+                                    .sum();
+      if (arena_m1.cols() >= arena_m1.rows()) {
+        arena_m2.adj().array() += arena_m1.val()
+                                      .array()
+                                      .transpose()
+                                      .replicate(1, arena_m2.cols())
+                                      .cwiseProduct(ret.adj().array());
```
bbbales2, line 46: I think this can be (and the other replicate ones too):

```cpp
arena_m2.adj() += arena_m1.val().asDiagonal() * ret.adj();
```

If that `.adj` gives errors then try `.adj_op`.

stan/math/rev/fun/diag_pre_multiply.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * @tparam T2 type of the matrix
+ * @param m1 input vector/row_vector
+ * @param m2 input matrix
+ *
+ * @return product of the diagonal matrix formed from the
+ * vector or row_vector and a matrix.
+ */
+template <typename T1, typename T2, require_eigen_vector_t<T1>* = nullptr,
+          require_eigen_t<T2>* = nullptr,
+          require_any_rev_matrix_t<T1, T2>* = nullptr>
```
bbbales2, line 25: There is some error in compiling with OpenCL. There should probably be an extra requires somewhere. @t4c1 can you look at this?

stan/math/rev/fun/diag_pre_multiply.hpp
```diff
@@ -0,0 +1,92 @@
...
+    arena_t<promote_scalar_t<double, T2>> arena_m2 = value_of(m2);
+    arena_t<ret_type> ret(arena_m1.val().asDiagonal() * arena_m2);
+    reverse_pass_callback([ret, arena_m1, arena_m2]() mutable {
+      arena_m1.adj().array() += arena_m2.val()
+                                    .array()
+                                    .cwiseProduct(ret.adj().array())
+                                    .rowwise()
+                                    .sum();
+    });
+    return ret_type(ret);
```
bbbales2, line 66: Instead of doing:

```cpp
arena_t<ret_type> ret(arena_m1.val().asDiagonal() * arena_m2);
reverse_pass_callback([ret, arena_m1, arena_m2]() mutable {
  arena_m1.adj().array() += arena_m2.val()
                                    .array()
                                    .cwiseProduct(ret.adj().array())
                                    .rowwise()
                                    .sum();
});
return ret_type(ret);
```

You can do:
```cpp
return make_callback_var(arena_m1.val().asDiagonal() * arena_m2,
    [arena_m1, arena_m2](const auto& ret_vi) mutable {
      arena_m1.adj() += arena_m2.val().asDiagonal() * ret_vi.adj();
});
```

It creates only one autodiff variable instead of two so is slightly more efficient (everything is packed into `make_callback_var` rather than a separate `ret` and `reverse_pass_callback`).

bbbales2 (changes requested):
Review!

stan/math/rev/fun/diag_pre_multiply.hpp
t4c1: That is a runtime error, not a compiling error. These tests are comparing OpenCL and CPU (this) implementation. Dimensions of matrix expressions somewhere in the new code do not match. Given what test fails, I guess this code does not handle matrices of size zero. If this is not caught by non-OpenCL tests a test case should be added.

stan/math/rev/fun/diag_pre_multiply.hpp
gregorp90: Thanks for the review Ben! I'll make the corrections.

stan/math/rev/fun/diag_pre_multiply.hpp
gregorp90: Thanks @t4c1, I'll look into it.

bbbales2: Thanks @t4c1, makes sense

stan/math/rev/fun/diag_pre_multiply.hpp
gregorp90: @bbbales2 this does not seem to work for me. I get:

`inconsistent deduction for auto return type: 'Eigen::Matrix<stan::math::var_value<double>, 2, 2, 0, 2, 2>' and then 'stan::math::var_value<Eigen::Matrix<double, 2, 2>, void>'`

If I understand correctly, the `make_callback_var` returns a **varmat**. However, I was just talking about **varmat**s with Rok yesterday, and he told me that `diag_pre_multiply` does not support **varmat** yet. Could this be the reason, why this approach isn't working? Maybe I am way off, still learning a lot here.

bbbales2: Oops, yeah you're right. I think the implementation you wrote is just a couple changes from working with varmat though. I will add a comment and if you want to varmat it you can, if not, that's fine.

stan/math/rev/fun/diag_pre_multiply.hpp
bbbales2, line 25: I think if you change these requires to:

```cpp
          require_vector_t<T1>* = nullptr,
          require_matrix_t<T2>* = nullptr,
          require_any_st_var<T1, T2>* = nullptr>
```

Then you're good for varmat too if I'm reading the code right. `require_vector_t` actually allows `std::vector` as well but we just won't expose that signature.

To test varmat, for every `stan::test::expect_ad` add a `stan::test::expect_ad_matvar` with the same arguments.

gregorp90: Great, thank you!

gregorp90: The code Ben proposed above (exchanging replicate calls) fixed this issue. However, I am wondering why there was no error in the mix tests:
```
Eigen::MatrixXd a00(0, 0);
Eigen::VectorXd u0(0);
expect_diag_pre_multiply(u0, a00);
```

bbbales2: Maybe the autodiff system is refusing to test a thing with zero inputs? It kinda doesn't make sense from that perspective. Maybe that code you quote is secretly doing nothing.

prim tests are responsible for testing edge case, values, and making sure errors throw in the right place. Mix makes sure if there are errors, that autodiff and non-autodiff versions of a function throw the same errors. So it makes sense to have duplicate tests in prim and mix for things. If the expect ad tests are refusing to test size zero things... I dunno.

Don't worry about it and I can look at this when I do a review when this is ready again.

bob-carpenter: > Maybe the autodiff system is refusing to test a thing with zero inputs?

If so, that's a bug in the autodiff system (and probably my bad). The boundary condition should be that a nullary function (one with no args), is tested for value at all levels of autodiff.  In general, derivatives check derivatives of result vs. all arguments, and here there are no arguments, so the derivative checks automatically pass.

gregorp90: I implemented this and it works OK when I try it manually: Create a main, define a couple of variables, lp, and compute the gradient. For example this works OK:

```
int main() {
	Eigen::Matrix<stan::math::var, 1, 2> M1;
	Eigen::Matrix<stan::math::var, 2, 2> M2;
	Eigen::Matrix<double, 2, 2> M2;
	stan::math::var_value<Eigen::Matrix<double, 2, 2>> M_res1;
	M1 << 1,2;
	M2 << 2,3,2,5;
	M_res1 = diag_pre_multiply(M1, M2);
	stan::math::var lp2 = 0;
	for (int i = 0; i < 2; ++i) {
		for (int j = 0; j < 2; ++j) {
			lp2 += M_res1(i,j);
		}
	}
	lp2.grad();
	std::cout << M1.adj() << std::endl;
	std::cout << M2.adj() << std::endl;
	return 0;
}
```
However, when I try running the mix tests I still get a bunch of errors. I think the two most relevant are:

```
./test/unit/math/test_ad_matvar.hpp:467:23: error: no matching function for call to 'test_matvar_gradient(const stan::test::ad_tolerances&, T_mv_ret&, T_vm_ret&, std::tuple<Eigen::Matrix<double, -1, 1, 0, -1, 1>, Eigen::Matrix<stan::math::var_value<double, void>, -1, -1, 0, -1, -1> >&, std::tuple<Eigen::Matrix<double, -1, 1, 0, -1, 1>, stan::math::var_value<Eigen::Matrix<double, -1, -1, 0, -1, -1>, void> >&)'
   test_matvar_gradient(tols, A_mv_ret, A_vm_ret, A_mv_tuple, A_vm_tuple);
```

and

`./stan/math/rev/functor/gradient.hpp:47:17: error: conversion from 'var_value<Eigen::Matrix<double, -1, -1>>' to non-scalar type 'var_value<double>' requested`

I am not that familiar with these tests. Is there a bug in my implementation, or is something else missing? I looked at a couple of other rev implementations that use `make_callback_var` and I couldn't find any obvious differences to my implementation.

gregorp90: OK, I'll add another test for that to prim, the autodiff test for zero inputs was already there so I also don't know if it does anything :) Thank you for all the information! As I said, your proposal fixed the issue with OpenCL test, so this shouldn't be a problem anymore.

bbbales2: If you're getting this error with reverse_psas_callback or what I said doesn't make sense, push your code and I'll have a look.

gregorp90: So that means I should use the initial implementation, right?

bbbales2: Yeah for this code:

```
Eigen::Matrix<stan::math::var, 1, 2> M1;
Eigen::Matrix<stan::math::var, 2, 2> M2;
Eigen::Matrix<double, 2, 2> M2;
stan::math::var_value<Eigen::Matrix<double, 2, 2>> M_res1;
...
M_res1 = diag_pre_multiply(M1, M2);
```

we're assuming for stanc3 that if all the input types are matvar, then the output will be matvar as well. So the fact that this returns a varmat is violatin' assumptions.

bbbales2: Yeah use the `reverse_pass_callback` for the contents, but use these new requires to enable varmat support.

gregorp90: > If you're getting this error with reverse_psas_callback or what I said doesn't make sense, push your code and I'll have a look.

No, with `reverse_pass_callback` it works (all tests, including OpenCL). I was just trying to do it with `make_callback_var`, as you proposed. OK, I'll revert to that implementation and push, thanks!

gregorp90:
@bbbales2 I pushed the changes, it is ready for review.

bbbales2:
@gregorp90 I turned on the varmat tests (confusingly named `expect_ad_matvar`) and also changed the 0x0 test a tiny bit. Click [here](https://github.com/stan-dev/math/pull/2405/commits/ef2362066f0786991dc42d3a9aa32b872b7f9cef) to see my changes.

If those look good to you (and the tests green light) we can merge (everything here looks good to me)

gregorp90:
@bbbales2 yes, it looks good to me, thank you!

bbbales2 (dismissed):

bbbales2:
@gregorp90 approved -- hopefully the tests pass and there's nothing weird. `diag_post_multiply` awaits if you have time.

gregorp90:
@bbbales2 great, thanks! Yes, I'll do the `diag_post_multiply`. Maybe not immediately, but in a week or two.

rok-cesnovar (approved):

PR #2415, opened 2021-03-09
Reimplement holder without dynamic allocations

t4c1:
## Summary

Reimplements `Holder`, so that it does not need dynamic memory allocations. New implementation also happens to be simpler and work for kernel generator expressions. The (unused) option for constructing holder from expression and pointers to arguments on heap now does not make any sense, so it is removed.

## Tests

Removed tests for removed option of how holder is constructed.

## Side Effects
Not really.

## Release notes

Reimplemented `Holder`, so that it does not need dynamic memory allocations.

## Checklist


stan/math/prim/meta/holder.hpp
```diff
@@ -51,110 +53,59 @@
...
+   * @param args
+   */
+  template <typename Functor, require_same_t<decltype(std::declval<Functor>()(
+                                                 std::declval<Args&>()...)),
+                                             BaseExpr>* = nullptr>
+  explicit Holder(const Functor& f, Args&&... args)
+      : std::tuple<Args...>(std::forward<Args>(args)...),
+        BaseExpr(index_apply<sizeof...(Args)>([this, &f](auto... Is) {
+          return f(std::get<Is>(*static_cast<std::tuple<Args...>*>(this))...);
+        })) {}
```
SteveBronder, line 92: Just to make sure I understand, for something like 

```cpp
template<typename T>
auto f(const T& x){
   const Eigen::Ref<const Eigen::VectorXd>& x_ref = x;
  make_holder([](const auto& x) { return x.cwiseProduct(x);}, x_ref);
}
```

This constructor is called which stores the `x_ref` by passing it to the inherited tuples constructor, then the lambda is invoked on `x_ref` with the expression stored in `BaseExpr`. Is that correct? It might be good to include that in the docs as I think in the past we've been confused about how this mechanic works.

SteveBronder (approved):
Very cool!! One quick comment related to docs but I think this is good

t4c1:
Looks cool, but I think this error is not fixable the way I have implemented this. I will probably have to redo this stuff from scratch.

SteveBronder:
The one you are seeing on Jenkins?

t4c1:
Yeah, it took me a long time to even figure out what it means. The issue is inheriting from Eigen/kernel generator operations, as we can not do CRTP, so inherited functions that rely on the derived type sometimes do or return the wrong thing.

t4c1:
Replaced by https://github.com/stan-dev/math/pull/2429

PR #2419, opened 2021-03-10
Update expression tests code generation

bbbales2:
## Summary

I took the varmat test stuff out of this. It is currently just a cleanup to the code generation for the expressions tests (and this will feed into the varmat stuff). It makes the pull smaller.

## Tests

None yet

## Release notes

- Updated code generation for expression tests

## Checklist

- [x] Math issue #1805


bbbales2:
@t4c1 can you give `test/generateExpressionTests.py` and see what you think? I tried to move a lot of the hairy ifs and stuff into a class FunctionGenerator [here](https://github.com/stan-dev/math/blob/237a95c2f0198313b2d0a281dc6c4171808329a2/test/sig_utils.py#L586).

@rok-cesnovar the way we would run this is:

```
./test/varmat_compatibility_test.py results.json
```

And then this to print the function names that are compatible:
```
./test/varmat_compatibility_summary.py results.json --names --fully
```

It takes a long time so this is something we'd put in a GHA or something (definitely not run every merge -- it has to compile one file for every signature).

Anyway I still gotta convert this all back to python2 and maybe shuffle the files a bit and add docs, so WIP.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+        self.arg_str = ','.join(arg.name for arg in args)
+        # This isn't always accurate cause the output could be an int or a boolean
+        # or something even if the input is reverse mode
+        self.__is_reverse_mode = any(arg.is_reverse_mode() for arg in args)
+
+    def is_reverse_mode(self):
+        return self.__is_reverse_mode
+
+    def is_matrix_like(self):
+        raise Exception("is_matrix_like not implemented for FunctionCallAssign")
```
t4c1, line 537:
```suggestion
        raise NotImplementedError("is_matrix_like not implemented for FunctionCallAssign")
```

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+    "PrimExp": "double",
+    "Rev": "stan::math::var",
+    "RevSOA": "stan::math::var",
+    "RevExp": "stan::math::var",
+    "Fwd": "stan::math::fvar<double>",
+    "FwdExp": "stan::math::fvar<double>",
+    "Mix": "stan::math::fvar<stan::math::var>",
+}
+
+class CppStatement:
```
t4c1, line 327: It looks like you want to replace `stan::test::make_arg` with the "Argument" set of classes. So I think you should actually replace it.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+        return False
+    
+    def is_matrix_like(self):
+        return False
+    
+    def is_varmat(self):
+        return False
+    
+    def is_expression(self):
+        return False
```
t4c1, line 341: These methods seem kinda redundand in statements. I know they are used in `FunctionGenerator`, but I think it would be simpler to implement them there.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+
+    def returns_int(self):
+        return "int" in self.return_type
+
+    def add(self, statement):
+        if not isinstance(statement, CppStatement):
+            raise Exception("Argument to FunctionGenerator.add must be an instance of an object that inherits from CppStatement")
+
+        self.code_list.append(statement)
+        return statement
```
t4c1, line 617: Metzhods `add` and `cpp` seem totally unrelated to what the rest of this class is doing. So I think they should not be part of this class.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+    def is_reverse_mode(self):
+        return self.__is_reverse_mode
+
+    def is_matrix_like(self):
+        raise Exception("is_matrix_like not implemented for FunctionCallAssign")
+
+    def cpp(self):
+        return f"auto {self.name} = stan::math::eval({self.function_name}({self.arg_str}));"
+
+class FunctionCall(CppStatement):
```
t4c1, line 542: I think `FunctionCall` and `FunctionCallAssignment` should be the same class, where `name` could be None.

t4c1 (changes requested):
The refactor part of this PR seems a bit bloated. It seems you are using way more code to achieve the same thing. 

Also it is quite hard to review undocumented code. Next time please add docs before asking for a review.

test/sig_utils.py
```diff
@@ -53,8 +57,9 @@ def get_cpp_type(stan_type):
     return res
 
 
-simplex = "make_simplex"
-pos_definite = "make_pos_definite_matrix"
+simplex = "simplex"
+pos_definite = "positive_definite_matrix"
+scalar_return_type = "scalar_return_type"
```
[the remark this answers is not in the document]
bbbales2: It's the additional logic that for the _constrain functions that take an lp argument. This lp argument is an output variable passed by reference that should have the type `return_type_t<All_other_args...>`

test/sig_utils.py
```diff
@@ -201,12 +212,12 @@ def get_cpp_type(stan_type):
...
-    "ode_rk45((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, ostream_ptr, vector) => array[,] real",
-    "ode_rk45_tol((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, real, real, real, ostream_ptr, vector) => array[,] real",
+    "ode_adams((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, ostream_ptr, vector) => array[] vector",
+    "ode_adams_tol((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, real, real, real, ostream_ptr, vector) => array[] vector",
+    "ode_bdf((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, ostream_ptr, vector) => array[] vector",
+    "ode_bdf_tol((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, real, real, real, ostream_ptr, vector) => array[] vector",
+    "ode_rk45((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, ostream_ptr, vector) => array[] vector",
+    "ode_rk45_tol((real, vector, ostream_ptr, vector) => vector, vector, real, array[] real, real, real, real, ostream_ptr, vector) => array[] vector",
```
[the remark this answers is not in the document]
bbbales2: Well they were marked here returning 2d arrays. They return arrays of vectors. From [here](https://mc-stan.org/docs/2_26/functions-reference/functions-ode-solver.html)

```
vector[] ode_bdf(function ode, vector initial_state, real initial_time, real[] times, ...)
```

test/sig_utils.py
```diff
@@ -67,6 +72,8 @@ def get_cpp_type(stan_type):
     "categorical_rng": [simplex, None],
     "categorical_lpmf": [None, simplex],
     "cholesky_decompose": [pos_definite, None],
+    "csr_to_dense_matrix": [1, 1, None, [1], [1, 2]],
+    "csr_matrix_times_vector": [1, 1, None, [1], [1, 2], None],
```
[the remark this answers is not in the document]
bbbales2: It was a workaround for the if statement [here](https://github.com/stan-dev/math/blob/develop/test/expressions/generateExpressionTests.py#L66).

Now how these magic constants work is if there is an array argument, you can type in an array here and it will translate to a cpp array initializer like:

```
std::vector<double> myvec = { 1, 2 };
```

It doesn't remove the tricky logic, but it moves it to the same place as a lot of the other tricks at least.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
     return arg
+
+overload_scalar = {
+    "Prim": "double",
+    "PrimExp": "double",
+    "Rev": "stan::math::var",
+    "RevSOA": "stan::math::var",
+    "RevExp": "stan::math::var",
+    "Fwd": "stan::math::fvar<double>",
+    "FwdExp": "stan::math::fvar<double>",
```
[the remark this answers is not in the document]
bbbales2: Good call not sure I use these anymore.

I was just tagging descriptors on to the autodiff types. "RevExp" -> reverse mode expressions. "RevSOA" -> reverse mode structure of arrays (varmat). Might be historical at this point, not sure.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+    "Mix": "stan::math::fvar<stan::math::var>",
+}
+
+class CppStatement:
+    def init(self):
```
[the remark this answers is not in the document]
bbbales2: Oof yeah

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+
+# These are used with ode_adams
+class OdeFunctorArgument(CppStatement):
+    def __init__(self, name):
+        self.name = name
+    
+    def cpp(self):
+        return f"stan::test::test_functor {self.name};"
+
+class ReturnTypeTArgument(CppStatement):
```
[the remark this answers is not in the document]
bbbales2: Same this as [this](https://github.com/stan-dev/math/pull/2419#discussion_r593125890) with a helpfully different name lol. I'll try to consolidate that.

test/sig_utils.py
```diff
@@ -301,3 +312,365 @@ def reference_vector_argument(arg):
...
+            return f"{arg_type} {self.name} = {value_string};"
+
+class FunctionCallAssign(CppStatement):
+    def __init__(self, function_name, name, *args):
+        self.function_name = function_name
+        self.name = name
+        self.arg_str = ','.join(arg.name for arg in args)
+        # This isn't always accurate cause the output could be an int or a boolean
+        # or something even if the input is reverse mode
+        self.__is_reverse_mode = any(arg.is_reverse_mode() for arg in args)
```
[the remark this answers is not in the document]
bbbales2: Hmm, I see. Yeah I Googled and someone said this is how to indicate private variables. I'll go to single _.

test/generateExpressionTests.py
```diff
@@ -0,0 +1,174 @@
...
+                test_code_template.format(
+                    overload=overload,
+                    test_name=f"{fg.function_name}_{n}",
+                    code = fg.cpp(),
+                )
+            )
+    
+    save_tests_in_files(j, tests)
+
+def processCLIArgs():
```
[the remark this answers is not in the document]
bbbales2: It was handy for testing and debugging to have this file runnable.

test/sig_utils.py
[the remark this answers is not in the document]
t4c1: That is nice!

test/sig_utils.py
bbbales2: I wasn't thinking that but I'll take a look at it and see if it's worth doing. The goal here was making the code generation Python between varmat_compatibility_tests.py and generateExpressionsTests.py look the same.

Thinking positively I was hoping that the complexity was kinda net neutral on generateExpressionsTest.py. I guess you're right this new thing definitely isn't simpler on its own.

I think where I started on this was trying to work through the benchmarks scripts, where the logic on the expressions is much finickier ([here](https://github.com/stan-dev/math/blob/develop/benchmarks/benchmark.py#L347)). I had a version of the varmat tests that looked like that but I remember thinking it was hard to read ([here](https://github.com/stan-dev/math/blob/d71f157ad2c7b2de54130c7ba6d75298d9def0fe/test/test_varmat_signatures.py)).

The two things we'll get out of this though is that it's the same codegen used in varmat_compatibility_tests.py, and it's easy to test the codegen in the Python classes cause all this stuff is rather fragile (easy to write a test that always passes with codegen).

Anyway I'll investigate this more. If we truly decide this new stuff just sucks for the expressions tests and having the same kind of code gen isn't useful we can revert it.

test/sig_utils.py
bbbales2: Hmm, the idea of the class was to parse function signatures (to encapsulate a lot of the logic of the parsed functions in a Python class, so `fg.is_rng()` instead of doing `function_name`.endswith("_rng")) and then also manage the code generation.

I guess these could be two things. I'll think if that makes things clearer.

test/sig_utils.py
bbbales2: These are inherited by all the different argument types. I was doing a polymorphism to make the code shorter.

test/sig_utils.py
bbbales2: `FunctionCall(...)` and `FunctionCall(None, ...)` seems less descriptive than `FunctionCall(...)` and `FunctionCallAssign(...)` but anyway let me dig maybe these merge.

test/sig_utils.py
t4c1: No, I am not trying to say that the abstractions you are adding are bad - just that they can be improved. More abstractions is almost always more code, but if it is not much more they can benefit readability. Also benchmarks could use these abstractions.

test/sig_utils.py
t4c1: I understand. But as I see it we don't need it. `is_varmat` and `is_expression` seem to be unused, calls `is_reverse_mode` on arguments can be replaced by a check whether we are doing a reverse mode overload and calls to `is_matrix_like` can be replaced with a check is something is eigen type.

test/sig_utils.py
bbbales2: Weeeee I changed this up a bit. Check out the generate expressions test code [now](https://github.com/stan-dev/math/blob/d1cab5237fdbe8214aa7573d53617558f11eac22/test/generateExpressionTests.py#L98).

I changed how the code runs to mimic the old expression tests, hid argument names, and added member functions to the FunctionGenerator for most of the utility functions you might call.

Now generateExpressionTests.py only needs to import and use the FunctionGenerator class, which is nice cause all the other little classes are hidden. I believe I will try to split the FunctionGenerator class into parsing-signatures and generating-code (cause these jobs are starting to look different).

```python
# Generate two sets of arguments, one will be expressions one will not
arg_list_expression_base = fg.build_arguments(len(fg.stan_args) * [overload], 2)
arg_list_expression = [fg.convert_to_expression(arg) if arg.is_matrix_like() else arg for arg in arg_list_expression_base]
arg_list_no_expression = fg.build_arguments(len(fg.stan_args) * [overload], 2)

# Check results with expressions and without are the same
cpp_function_name = f"stan::math::{fg.function_name}"
result = fg.function_call_assign(cpp_function_name, *arg_list_expression)
result_no_expression = fg.function_call_assign(cpp_function_name, *arg_list_no_expression)
fg.expect_eq(result, result_no_expression)

# Check that expressions evaluated only once
for arg in arg_list_expression:
  if arg.is_expression():
    fg.expect_leq_one(arg.counter)

# If reverse mode, check the adjoints            
if is_reverse_mode:
  summed_result = fg.recursive_sum(result)
  summed_result_no_expression = fg.recursive_sum(result_no_expression)
  sum_of_sums = fg.add(summed_result, summed_result_no_expression)
  fg.grad(sum_of_sums)

  for arg, arg_no_expression in zip(arg_list_no_expression, arg_list_expression):
    fg.expect_adj_eq(arg, arg_no_expression)

  fg.recover_memory()
```

And this generates the follow cpp:
```cpp
TEST(ExpressionTestRev, multiply_0) {
stan::math::var real0 = 0.4;
auto matrix1 = stan::test::make_arg<Eigen::Matrix<stan::math::var, 1, Eigen::Dynamic>>(0.4, 1);
int matrix1_expr2_counter = 0;
stan::test::counterOp<stan::math::var> matrix1_expr2_counter_op(&matrix1_expr2_counter);
auto matrix1_expr2 = matrix1.segment(0,1).unaryExpr(matrix1_expr2_counter_op);
stan::math::var real3 = 0.4;
auto matrix4 = stan::test::make_arg<Eigen::Matrix<stan::math::var, 1, Eigen::Dynamic>>(0.4, 1);
auto result5 = stan::math::eval(stan::math::multiply(real0,matrix1_expr2));
auto result6 = stan::math::eval(stan::math::multiply(real3,matrix4));
EXPECT_STAN_EQ(result5,result6);
EXPECT_LEQ_ONE(matrix1_expr2_counter);
auto summed_result7 = stan::math::eval(stan::test::recursive_sum(result5));
auto summed_result8 = stan::math::eval(stan::test::recursive_sum(result6));
auto sum_of_sums9 = stan::math::eval(stan::math::add(summed_result7,summed_result8));
stan::test::grad(sum_of_sums9);
stan::test::expect_adj_eq(real3,real0);
stan::test::expect_adj_eq(matrix4,matrix1_expr2);
stan::math::recover_memory();
}
```

rok-cesnovar:
Mind also closing https://github.com/stan-dev/math/issues/2425 here?

bbbales2:
@serban-nicusor-toptal the Jenkins computer running this test needs pytest installed: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2419/24/pipeline . Rok said you'd know how to do this?

test/sig_utils.py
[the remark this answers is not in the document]
bbbales2: Removed

test/generate_expression_tests.py
```diff
@@ -0,0 +1,167 @@
...
+            if overload == "Rev" and not sp.is_rev_compatible():
+                continue
+
+            is_reverse_mode = overload == "Rev" and not sp.returns_int()
+
+            cg = CodeGenerator()
+
+            # Generate two sets of arguments, one will be expressions one will not
+            arg_list_expression_base = cg.build_arguments(sp, sp.number_arguments() * [overload], size = 1)
+            arg_list_expression = [cg.convert_to_expression(arg, size = 1) if arg.is_matrix_like() else arg for arg in arg_list_expression_base]
```
bbbales2, line 104: I was wanting to generate variables of size 2x2 and then take 1x1 sub-expressions of those as the expressions (to test expressions as blocks where the indexing was non-trivial). It didn't work out. If the last tests were doing it, I guess I should do more work. If they weren't, then I'll just leave it like this.

bbbales2:
@t4c1 this is ready to review again. I changed it to just be the expression tests refactor so it is slightly simpler. All the tests won't pass until pytest in place, but the basic expression tests do pass.

test/signature_parser.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def is_rev_compatible(self):
+        """Return true if this signature should be compatible with reverse mode autodiff"""
+        return not (self.function_name in no_rev_overload or self.is_rng())
+
+    def is_ignored(self):
+        """Return true if this signature should be ignored by any tests (they are functions in stanc3 not defined in Math)"""
+        return self.function_name in ignored
+
+    def has_vector_arg(self):
```
t4c1, line 36: Can we rename this to `has_eigen_arg` ?

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+        self.stan_arg = stan_arg
+        if value is None:
+            self.value = 0.4
+        else:
+            self.value = value
+        
+        if size is None:
+            self.size = 1
+        else:
+            self.size = size
```
t4c1, line 99: Set defaults to the values you want instead of this.

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+class CppStatement:
+    """Base class for all statements. Implements some default checks. Should not be used directly"""
+    def __init__(self):
+        raise Exception("CppStatement should never be instantiated directly")
+
+    def is_reverse_mode(self):
+        """By default, a statement is not reverse mode"""
+        return False
+    
+    def is_matrix_like(self):
```
t4c1, line 15: Could this be `is_eigen` or maybe `is_expression_compatible`?

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+        if value is None:
+            self.value = 1
+        else:
+            self.value = int(value)
+
+    def cpp(self):
+        """Generate C++"""
+        return "int {name} = {value};".format(name = self.name, value = self.value)
+
+class RealVariable(CppStatement):
```
t4c1, line 48: You could reduce the amount of code by merging int and real scalars into a single class.

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+        self.stan_arg = "matrix"
+        if value is None:
+            self.value = 0.4
+        else:
+            self.value = value
+        
+        if size is None:
+            self.size = 1
+        else:
+            self.size = size
```
t4c1, line 187: Use actual default values instead of None.

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+    def cpp(self):
+        """Generate C++"""
+        try:
+            lhs_string = "decltype({name})".format(name = self.value.name)
+            rhs_string = self.value.name
+            for n in range(self.number_nested_arrays):
+                lhs_string = "std::vector<" + lhs_string + ">"
+                rhs_string = "{" + ",".join([rhs_string] * self.size) + "}"
+
+            return "{lhs_string} {name} = {rhs_string};".format(lhs_string = lhs_string, name = self.name, rhs_string = rhs_string)
```
t4c1, line 342: Try to avoid long try blocks. The way this is written I have no idea which statement you expect could throw. (unless almost any statment could, but I think this is not the case here)

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+        except AttributeError:
+            scalar = overload_scalar[self.overload]
+            lhs_string = arg_types[self.inner_type].replace("SCALAR", scalar)
+            for n in range(self.number_nested_arrays):
+                lhs_string = "std::vector<" + lhs_string + ">"
+
+            # This is a bit of a nasty way to convert a python iterable type to C++ code
+            # It doesn't respect the self.size variable
+            value_string = repr(self.value).replace('[', '{').replace(']', '}').replace('(', '{').replace(')', '}')
```
t4c1, line 351: Do you need this to work on nested lists?

test/statement_types.py
```diff
@@ -0,0 +1,451 @@
...
+    """Represents a function call with no return"""
+    def __init__(self, function_name, *args):
+        """
+        Initialize a function call
+        
+        :param function_name: C++ name of function
+        :param args: Args to pass to function call
+        """
+        FunctionCallAssign.__init__(self, function_name, None, *args)
+
```
t4c1, line 401: I don't think you need a separate class here.

t4c1 (changes requested):
Why don't you use built-in unittest module instead of pytest? Is it is some way significantly worse? I would like to avoid aditional dependancies, unless we really need them.

stan/math/prim/fun/rank.hpp
```diff
@@ -21,7 +21,9 @@ inline int rank(const C& v, int s) {
   check_range("rank", "v", v.size(), s);
   --s;  // adjust for indexing by one
   return apply_vector_unary<C>::reduce(v, [s](const auto& vec) {
-    return (vec.array() < vec.coeff(s)).template cast<int>().sum();
+    const auto& vec_ref = to_ref(vec);
+
+    return (vec_ref.array() < vec_ref.coeff(s)).template cast<int>().sum();
```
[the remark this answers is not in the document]
bbbales2: Doesn't it evaluate twice though? Once for the left hand side and once for the right?

If I revert those changes I get:
```
Expected: (matrix0_expr2_counter) <= (1), actual: 2 vs 1
```

This is the test code, prettied up a bit

```cpp
#include <test/expressions/expression_test_helpers.hpp>

TEST(ExpressionTestPrim, rank0) {
  /*
  rank(row_vector, int) => int
  */
  auto matrix0 = stan::test::make_arg<Eigen::Matrix<double, 1, Eigen::Dynamic>>(0.4, 1);
  int int1 = 1;
  int matrix0_expr2_counter = 0;
  stan::test::counterOp<double> matrix0_expr2_counter_op(&matrix0_expr2_counter);
  auto matrix0_expr2 = matrix0.segment(0,1).unaryExpr(matrix0_expr2_counter_op);

  auto matrix3 = stan::test::make_arg<Eigen::Matrix<double, 1, Eigen::Dynamic>>(0.4, 1);
  int int4 = 1;

  auto result5 = stan::math::eval(stan::math::rank(matrix0_expr2,int1));
  auto result6 = stan::math::eval(stan::math::rank(matrix3,int4));
  EXPECT_STAN_EQ(result5,result6);
  EXPECT_LEQ_ONE(matrix0_expr2_counter);
}
```

test/generate_expression_tests.py
```diff
@@ -62,83 +69,70 @@ def main(functions=(), j=1):
...
     else:
-        for function in functions:
-            for signature in get_signatures():
-                return_type, function_name, function_args = parse_signature(signature)
-                if function == function_name:
-                    signatures.add(signature)
+        for signature in get_signatures():
+            sp = SignatureParser(signature)
+            if sp.function_name in functions:
+                signatures.add(signature)
```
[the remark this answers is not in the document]
bbbales2: I'm not seeing the diff (this all looks new to me) so I'm guess this is the change to put the signature loop on the outside and functions on the inside.

The reason I did that was just one less visible loop. There hopefully isn't any functional change.

test/signature_parser.py
bbbales2: Well at this point we don't know about eigen (I wanted the logic to be on the stan types). I guess `has_eigen_compatible_arg` fits that requirement? I can do that.

test/statement_types.py
bbbales2: `is_eigen_compatible` seems fine and matches discussion [above](https://github.com/stan-dev/math/pull/2419/files#r595937988), so I can do that.

test/code_generator.py
```diff
@@ -0,0 +1,121 @@
...
+                    arg = statement_types.AlgebraSolverFunctorArgument("functor" + suffix)
+                elif inner_type == "(real, vector, ostream_ptr, vector) => vector":
+                    arg = statement_types.OdeFunctorArgument("functor" + suffix)
+                else:
+                    raise Exception("Inner type " + inner_type + " not supported")
+
+                if number_nested_arrays > 0:
+                    self.add_statement(arg)
+                    arg = statement_types.ArrayArgument(overload, "array" + suffix, number_nested_arrays, inner_type, value = arg)
+            else:
```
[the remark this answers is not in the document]
bbbales2: Yeah the logic is really opaque here. I'll think about how to make this simpler.

`if value is None or isinstance(value, numbers.Number):` isn't exactly easy to understand.

test/statement_types.py
bbbales2: No but I think it does.

I suppose I could just replace this with a `",".join(self.value)` and an if statement that if `self.number_nested_arrays > 1` throw a NotImplementedException

stan/math/prim/fun/rank.hpp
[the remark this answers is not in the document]
t4c1: Only a single element is evaluated twice. I think that is better than forcing evaluation or everything.

> If I revert those changes I get:

 There used to be a special exception in expression tests for this function. Maybe it should be kept in.

test/generate_expression_tests.py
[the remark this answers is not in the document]
t4c1: To me it looks very functional - like it breaks the option to specify function by name.

test/code_generator.py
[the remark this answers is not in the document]
t4c1: It is not really too complicated, but if we have a simpler option we should use that.

stan/math/prim/fun/rank.hpp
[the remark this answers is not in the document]
bbbales2: This would be a wrong thing to do if the expression was a matrix-vector product.

I could add it to the ignored list but I prefer to just not have the exception. I'm not really sure how this function is used in practice (if it's used in a way where performance is critical or not). I'm assuming not.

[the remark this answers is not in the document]
t4c1: I see your point. Fine.

test/generate_expression_tests.py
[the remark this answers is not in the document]
bbbales2: Nono, both by name and by signature work still.

So both:

```
./runTests.py test/expressions --only-functions dot_product
./runTests.py test/expressions --only-functions "dot_product(vector,vector)=>real"
```

For functions by name, this is just looking for all the signatures that match the function name. For custom signatures, those got added to the signature list on line 68.

test/code_generator.py
```diff
@@ -0,0 +1,121 @@
...
+    def cpp(self):
+        return os.linesep.join(statement.cpp() for statement in self.code_list)
+
+    def get_next_name_suffix(self):
+        self.name_counter += 1
+        return repr(self.name_counter - 1)
+
+    def build_arguments(self, signature_parser, arg_overloads, max_size = None):
+        if not max_size:
+            max_size = 2
```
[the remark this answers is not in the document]
bbbales2: Removed

test/code_generator.py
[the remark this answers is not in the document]
bbbales2: I moved the special case to the beginning and added a comment explaining it.

test/statement_types.py
bbbales2: I got rid of a lot of the default size behavior. I tried cleaning up the value defaults but realized I am using `value = None` to mean use a default value.

test/statement_types.py
bbbales2: I changed the try to an if statement to make the condition clearer.

test/statement_types.py
bbbales2: I simplified this and added a NotImplementedError if nested lists come along

serban-nicusor-toptal:
@bbbales2 I've installed `pytest` and restarted the [pipeline](https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2419/29/pipeline)

test/statement_types.py
bbbales2: It's mildly more convenient to keep the int casts in one spot.

test/statement_types.py
bbbales2: Removed.

test/code_generator.py
```diff
@@ -0,0 +1,121 @@
...
+
+        arg_list = []
+        for n, (overload, stan_arg) in enumerate(zip(arg_overloads, signature_parser.stan_args)):
+            suffix = self.get_next_name_suffix()
+
+            number_nested_arrays, inner_type = parse_array(stan_arg)
+
+            value = None
```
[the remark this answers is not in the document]
bbbales2: Moved and added comment

serban-nicusor-toptal:
It failed because it couldn't find `pytest` but I logged in and it's there.
```
ubuntu@ip-172-31-31-203:~$ which pytest
/home/ubuntu/.local/bin/pytest
ubuntu@ip-172-31-31-203:~$ pytest
platform linux -- Python 3.5.2, pytest-6.1.2, py-1.10.0, pluggy-0.13.1

rootdir: /home/ubuntu
collected 0 items                                                                                                                                                                                                                                                                                                        

no tests ran in 0.00s ```

**Comment by bbbales2** [comment]
@serban-nicusor-toptal well I guess don't worry about it any more. Tadej pointed out I wasn't really making use of the dependency beyond the regular Python unittests, so I switched back to those. Should work now

**Comment by bbbales2** [comment]
Thanks for checking it out though, apologies for the unannounced direction-change!

**Comment by serban-nicusor-toptal** [comment]
Ah okay! In case you still want to run it like that in the future try with `python3 -m pytest ./test/code_generator_test.py` I think by default it's pointing to python (2).
Also btw, if you ever need to test something out and you miss a package, `jenkins` user has access to install `pip` packages so you can install them in the pipeline to do your work until I add them to the instances.
And np, glad we're good now!

**Comment by bbbales2** [comment]
> jenkins user has access to install pip packages

Ooo, okay that is convenient thanks.

**Comment by bbbales2** [comment]
@t4c1 This is ready for review again

File: test/statement_types.py [hunk 28]
```diff
@@ -75,7 +75,7 @@ def cpp(self):
 
 class MatrixVariable(CppStatement):
     """Represents a matrix variable"""
-    def __init__(self, overload, name, stan_arg, value = None, size = None):
+    def __init__(self, overload, name, stan_arg, size, value = None):
```
t4c1, line 78:
```suggestion
    def __init__(self, overload, name, stan_arg, size, value = 0.4):
```

test/statement_types.py
```diff
@@ -88,21 +88,18 @@ def __init__(self, overload, name, stan_arg, value = None, size = None):
         self.overload = overload
         self.name = name
         self.stan_arg = stan_arg
+        self.size = size
+        
         if value is None:
             self.value = 0.4
         else:
             self.value = value
```
t4c1:
```suggestion
        self.value = value
```

test/statement_types.py
```diff
@@ -120,7 +117,7 @@ def cpp(self):
 
 class SimplexVariable(CppStatement):
     """Represents a simplex variable"""
-    def __init__(self, overload, name, value = None, size = None):
+    def __init__(self, overload, name, size, value = None):
```
t4c1, line 120:
```suggestion
    def __init__(self, overload, name, size, value = 0.4):
```

test/statement_types.py
```diff
@@ -132,21 +129,18 @@ def __init__(self, overload, name, value = None, size = None):
         self.overload = overload
         self.name = name
         self.stan_arg = "vector"
+        self.size = size
+
         if value is None:
             self.value = 0.4
         else:
             self.value = value
```
t4c1:
```suggestion
        self.value = value
```

test/statement_types.py
```diff
@@ -164,7 +158,7 @@ def cpp(self):
 
 class PositiveDefiniteMatrixVariable(CppStatement):
     """Represents a positive definite matrix variable"""
-    def __init__(self, overload, name, value = None, size = None):
+    def __init__(self, overload, name, size, value = None):
```
t4c1, line 161:
```suggestion
    def __init__(self, overload, name, size, value = 0.4):
```

test/statement_types.py
```diff
@@ -176,21 +170,18 @@ def __init__(self, overload, name, value = None, size = None):
         self.overload = overload
         self.name = name
         self.stan_arg = "matrix"
+        self.size = size
+
         if value is None:
             self.value = 0.4
         else:
             self.value = value
```
t4c1:
```suggestion
        self.value = value
```

test/statement_types.py
```diff
@@ -306,52 +297,47 @@ def __init__(self, overload, name, number_nested_arrays, inner_type, value = Non
...
                 lhs_string = "std::vector<" + lhs_string + ">"
                 rhs_string = "{" + ",".join([rhs_string] * self.size) + "}"
 
             return "{lhs_string} {name} = {rhs_string};".format(lhs_string = lhs_string, name = self.name, rhs_string = rhs_string)
-        except AttributeError:
-            scalar = overload_scalar[self.overload]
-            lhs_string = arg_types[self.inner_type].replace("SCALAR", scalar)
-            for n in range(self.number_nested_arrays):
-                lhs_string = "std::vector<" + lhs_string + ">"
+        else:
```
t4c1, line 330:
```suggestion
        else:  # generate initializer list
```

test/generate_expression_tests.py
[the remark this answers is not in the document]
t4c1: Right, this is fine.

t4c1 (changes requested):
This new code generation stuff could also be used for generating benchmarks, right?

Also new code needs docs for function parameters.

rok-cesnovar:
Any idea why the new expr. tests take 2x the time as the previous expressions tests? I am not saying its a deal breaker as its 10 minutes, just wondering what changed?

test/statement_types.py
bbbales2: I tried switching to something like this but it doesn't work because `value = None` is used in `code_generator.py` ([here](https://github.com/stan-dev/math/blob/0e906cf97fc5b07354d6efaa381f646e95ec57dd/test/code_generator.py#L52)) to mean 'use the default value'.

bbbales2:
> Any idea why the new expr. tests take 2x the time as the previous expressions tests?

I think the code generation is less efficient (double the arguments are initialized). I'm not sure.

I did a test and the new code gen generates 52008 tests. The old code generates 52021 tests.

bbbales2:
@t4c1 I updated the comments.

Re: benchmarking, I want to merge this on its own, and then the next pull is a script to automatically detect which functions can compile with varmat variables or not (it is just 2 scripts).

I haven't looked at the benchmarking script yet. It probably wouldn't be too much trouble to clean it up (maybe a day to convert?) while we're on the topic of code generation.

test/statement_types.py
t4c1: Ooooh, ok.

t4c1:
> I did a test and the new code gen generates 52008 tests. The old code generates 52021 tests.

Any idea why there is a difference at all? I think we should understand what is happening here.

bbbales2:
> Any idea why there is a difference at all?

Yeah I was curious too.

The list of functions affected is is_cholesky_factor, is_matching_dims, multinomial_log, multinomial_lpmf

It looks like the old code is doing codegen for the same signatures twice somehow.

Like here are two functions it generates:

```cpp
TEST(ExpressionTestPrim, is_cholesky_factor0) {
  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> arg_mat0 = stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>();
  auto res_mat = stan::math::is_cholesky_factor(arg_mat0);
  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> arg_expr0 = stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>();
  int counter0 = 0;
  stan::test::counterOp<double> counter_op0(&counter0);
  auto res_expr = stan::math::is_cholesky_factor(arg_expr0.block(0,0,1,1).unaryExpr(counter_op0));
  EXPECT_STAN_EQ(res_expr, res_mat);
  EXPECT_LE(counter0, 1);
}

TEST(ExpressionTestPrim, is_cholesky_factor1) {
  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> arg_mat0 = stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>();
  auto res_mat = stan::math::is_cholesky_factor(arg_mat0);
  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> arg_expr0 = stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>();
  int counter0 = 0;
  stan::test::counterOp<double> counter_op0(&counter0);
  auto res_expr = stan::math::is_cholesky_factor(arg_expr0.block(0,0,1,1).unaryExpr(counter_op0));
  EXPECT_STAN_EQ(res_expr, res_mat);
  EXPECT_LE(counter0, 1);
}
```

I assume this means that one of the sets of special tests is causing the functions to get added twice.

bbbales2:
Looks like `get_signatures` isn't totally unique:

```python
signatures = get_signatures()
len(signatures)
26982
len(set(signatures))
26976
```

New code pushes everything through sets so it doesn't show this.

bbbales2:
Here's new code vs. old code. New code is more lines. Isn't obvious to me why it takes so much longer, but I guess all the little `decltype` s and `auto` s add up?

New code:
```cpp
TEST(ExpressionTestRev, multinomial_lpmf3) {
/*
multinomial_lpmf(array[] int, vector) => real
 */
  int int0 = 1;
  std::vector<decltype(int0)> array0 = {int0};
  auto simplex1 = stan::test::make_simplex<Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1>>(0.4, 1);
  int simplex1_expr2_counter = 0;
  stan::test::counterOp<stan::math::var> simplex1_expr2_counter_op(&simplex1_expr2_counter);
  auto simplex1_expr2 = simplex1.segment(0,1).unaryExpr(simplex1_expr2_counter_op);

  int int3 = 1;
  std::vector<decltype(int3)> array3 = {int3};
  auto simplex4 = stan::test::make_simplex<Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1>>(0.4, 1);

  auto result5 = stan::math::eval(stan::math::multinomial_lpmf(array0,simplex1_expr2));
  auto result6 = stan::math::eval(stan::math::multinomial_lpmf(array3,simplex4));

  EXPECT_STAN_EQ(result5,result6);
  EXPECT_LEQ_ONE(simplex1_expr2_counter);

  auto summed_result7 = stan::math::eval(stan::test::recursive_sum(result5));
  auto summed_result8 = stan::math::eval(stan::test::recursive_sum(result6));
  auto sum_of_sums9 = stan::math::eval(stan::math::add(summed_result7,summed_result8));

  stan::test::grad(sum_of_sums9);
  stan::test::expect_adj_eq(array3,array0);
  stan::test::expect_adj_eq(simplex4,simplex1_expr2);
  stan::math::recover_memory();
}
```

Old code:
```cpp
TEST(ExpressionTestRev, multinomial_log1) {
  std::vector<int> arg_mat0 = stan::test::make_arg<std::vector<int>>();
  Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1> arg_mat1 = stan::test::make_simplex<Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1>>();

  auto res_mat = stan::math::multinomial_log(arg_mat0, arg_mat1);

  std::vector<int> arg_expr0 = stan::test::make_arg<std::vector<int>>();
  Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1> arg_expr1 = stan::test::make_simplex<Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1>>();
  int counter1 = 0;
  stan::test::counterOp<stan::math::var> counter_op1(&counter1);

  auto res_expr = stan::math::multinomial_log(arg_expr0, arg_expr1.segment(0,1).unaryExpr(counter_op1));

  EXPECT_STAN_EQ(res_expr, res_mat);

  EXPECT_LE(counter1, 1);
  (stan::test::recursive_sum(res_mat) + stan::test::recursive_sum(res_expr)).grad();
  EXPECT_STAN_ADJ_EQ(arg_expr0,arg_mat0);
  EXPECT_STAN_ADJ_EQ(arg_expr1,arg_mat1);
}
```

bbbales2:
@t4c1 ready for review again

t4c1:
I looked at the comparison of old and new tests. All those `eval`s on scalars (`summed_result7`, `summed_result8`, `sum_of_sums9`) are redundant. Other than that I don't see anything that could double the running time. Also I don't see why did you need to invent a new macro `EXPECT_LEQ_ONE`.

test/generate_expression_tests.py
```diff
@@ -0,0 +1,167 @@
...
+    parser = FullErrorMsgParser(
+        description="Generate and run_command benchmarks.",
+        formatter_class=ArgumentDefaultsHelpFormatter,
+    )
+    parser.add_argument(
+        "--functions",
+        nargs="+",
+        type=str,
+        default=[],
+        help="Signatures and/or function names to benchmark. Ignores any finished checks in results file (if given).",
```
t4c1, line 154: I still don't think this file needs to be runnable on its own. If you strongly disagree, fix the help message and docs of `processCLIArgs`.

t4c1 (changes requested):

test/generate_expression_tests.py
bbbales2: Removed

bbbales2:
@t4c1 Ready again

bbbales2:
> All those evals on scalars (summed_result7, summed_result8, sum_of_sums9) are redundant.

When I call a function I evaluate the results to make sure all the returns are not expressions (and do something unexpected). Since the python here isn't generating performance sensitive code, I preferred to hide the evals. This might change if we did the benchmarks, but it is easy enough to change.

> Also I don't see why did you need to invent a new macro EXPECT_LEQ_ONE.

I just did the first thing that came to mind. Coulda done it as a function. Could have also initialized an IntVariable and called a two argument macro, but this seemed to fit well enough with how we do other bits of code.

t4c1:
> When I call a function I evaluate the results to make sure all the returns are not expressions (and do something unexpected). Since the python here isn't generating performance sensitive code, I preferred to hide expressions with evals (edited). This might change if we did the benchmarks, but it is easy enough to change.

We don't have scalar expressions. So these evals never do anything.

> I just did the first thing that came to mind. Coulda done it as a function. Could have also initialized an IntVariable and called a two argument macro, but this seemed to fit well enough with how we do other bits of code.

I dont like it. Calling two argument macro makes most sense to me. I don't think we define macros anywhere else just to hide one argument.

bbbales2:
> Calling two argument macro makes most sense to me

Switched

> We don't have scalar expressions. So these evals never do anything.

Right, but I wanted `CodeGenerator.function_call_assign` to be simple to think about (not returning expressions). If the difference becomes important can change this easy-peasy.

t4c1:
Oh, I see. I just thought that might have been the reason for the slow down of expression tests.

t4c1 (approved):

PR #2422, opened 2021-03-13
Remove v_vari

SteveBronder:
## Summary

Just a small PR that removes `v_vari` from the `rev` folder and replaces it with `make_callback_var()`

## Tests

Cleanup so no new tests

## Side Effects

No side effects but I was wondering about the callbacks that perform NaN checks like 

```cpp
inline var operator-(const var& a) {
  return make_callback_var(-a.val(), [a](const auto vi) mutable {
    if (unlikely(is_nan(a.val()))) {
      a.adj() = NOT_A_NUMBER;
    } else {
      a.adj() -= vi.adj();
    }
  });
}
```

Have we decided to not do NaN checks on the reverse pass? It feels like we are very inconsistent on this. Personally I'm pro removing them because of the inconsistency.

## Checklist


t4c1:
I kinda agree on NaN checks.

stan/math/rev/fun/gamma_p.hpp
```diff
@@ -78,11 +78,11 @@ class gamma_p_dv_vari : public op_dv_vari {
       : op_dv_vari(gamma_p(a, bvi->val_), a, bvi) {}
   void chain() {
     if (is_inf(ad_)) {
-      bvi_->adj_ += NOT_A_NUMBER;
+      bvi_->adj_ = NOT_A_NUMBER;
       return;
     }
     if (is_inf(bvi_->val_)) {
-      bvi_->adj_ += NOT_A_NUMBER;
+      bvi_->adj_ = NOT_A_NUMBER;
```
bbbales2, line 85: [optional] It looks like the conversion on gamma_p never got finished.

bbbales2 (approved):
Good if tests pass.

If you want to remove all the NaN logic, happy to scroll back through and check.

The only place that looked weird to me was how we handle maxes/mins with NaNs, but apparently that is undefined behavior (https://stackoverflow.com/questions/55153210/do-stdmin0-0-1-0-and-stdmax0-0-1-0-yield-undefined-behavior) so I guess it probably doesn't matter so much what we do.

stan/math/rev/fun/gamma_p.hpp
SteveBronder: I just did a ctrl+f for `+= NOT_A_NUMBER` and that's why these changed. I'm going to do the vv ones in a future PR

stan/math/opencl/rev/logit.hpp
```diff
@@ -20,7 +20,8 @@ template <typename T,
 inline var_value<matrix_cl<double>> logit(const var_value<T>& A) {
   return make_callback_var(
       logit(A.val()), [A](vari_value<matrix_cl<double>>& res) mutable {
-        A.adj() += elt_divide(res.adj(), A.val() * LOG_TEN);
+        A.adj() += elt_multiply(res.adj(),
+                                elt_divide(1.0, A.val() - square(A.val())));
```
SteveBronder, line 24: @t4c1 fyi I think this was wrong before? I wrote it to match up to the rev version

t4c1: It seems so. How did the tests pass?

Anyway you can simplify this into:
```A.adj() += elt_divide(res.adj(), A.val() - square(A.val()));```

File: stan/math/opencl/rev/round.hpp [hunk 3]
```diff
@@ -20,10 +20,7 @@ namespace math {
 template <typename T,
           require_all_kernel_expressions_and_none_scalar_t<T>* = nullptr>
 inline var_value<matrix_cl<double>> round(const var_value<T>& A) {
-  return make_callback_var(
-      round(A.val()), [A](vari_value<matrix_cl<double>>& res) mutable {
-        A.adj() = select(isnan(A.val()), NOT_A_NUMBER, A.adj());
-      });
+  return var_value<matrix_cl<double>>(round(A.val()));
```
t4c1, line 23: I think you can delete rev overload here and make prim overload handle this (returning prim matrix_cl).

SteveBronder:
Ayy @rok-cesnovar @serban-nicusor-toptal it looks like something is going on with the opencl device across a few PRs

rok-cesnovar:
Yeah, see https://github.com/stan-dev/math/issues/2442

SteveBronder:
Cool thanks!

stan/math/opencl/rev/round.hpp
SteveBronder: Sorry I'm not sure I'm following what you mean here

SteveBronder:
Besides @t4c1 's comment above this is ready for review!

stan/math/opencl/rev/round.hpp
t4c1: This is a rev overload in which nothing is done in reverse pass, making it essentially the same as prim overload. So we do not need two overloads. `var_value<matrix_cl<double>>` argument can be handled in prim overload.

bbbales2: I think the existing behavior is closer to, if autodiff input, then autodiff output. That is what we are doing CPU side too, so I think this is correct as-is.

Where to look for the opencl prim round implementation?

test/unit/math/opencl/opencl_context_test.cpp
```diff
@@ -87,7 +87,7 @@ TEST(opencl_context, switch_devices_errors) {
   EXPECT_THROW(stan::math::opencl_context.select_device(0, 99999),
                std::system_error);
 }
-
+/*
```
bbbales2, line 90: This test should probably be turned back on

stan/math/opencl/rev/logit.hpp
bbbales2: I reverted this change and the tests failed for me locally. It looks like the right tests are running in CI. I don't know. This seems like a canary worth investigating, but if nobody can reproduce it then nobody can reproduce it :/.

bbbales2 (commented):
I made a couple changes and left comments. I'll leave to @t4c1 to approve and merge. I don't understand how the non-working logit got in. The tests seem to work for me now.

stan/math/opencl/rev/logit.hpp
bbbales2: Oh also I switched to the simplified `A.adj() += elt_divide(res.adj(), A.val() - square(A.val()));`

stan/math/opencl/rev/round.hpp
t4c1: Prim round is implemented in kernel generator in `stan/math/opencl/kernel_generator/elt_function_cl.hpp`. Right than it is not that easy to use prim implementation for rev types.

t4c1:
This looks good, just uncomment the test Ben mentioned.

bbbales2 (approved):
Grand theft github points

Issue #2420, opened 2021-03-10
[FR] No bounds checks when NDEBUG flag is set


SteveBronder:
## Description

From [#853](https://github.com/stan-dev/stanc3/issues/853) in stanc3 we recently reverted a check where the compiler actually avoided bounds checks on single indexing functions. This can have a pretty bad performance hit so I think now would be a good time to do something like #2255 where when NDEBUG is set bounds checks are a no-op. This is a standard no debug flag in C++ and is used by boost, eigen, and the standard library to turn off bounds checking. So I think its reasonable that if a user has it set they expect to not have these checks

## Example

```cpp
// in macro.hpp
#ifdefine NDEBUG
#ifndef STAN_NO_RANGE_AND_SIZE_CHECK
#define STAN_NO_RANGE_AND_SIZE_CHECK return
#endif
#else
#ifndef STAN_NO_RANGE_AND_SIZE_CHECK
#define STAN_NO_RANGE_AND_SIZE_CHECK
#endif
#endif

// in check_ranges.hpp
inline void check_range(const char* function, const char* name, int max,
                        int index, int nested_level, const char* error_msg) {
  STAN_NO_RANGE_AND_SIZE_CHECK;
  //... rest of body
}
```

## Expected Output

Bounds checks are not turned on

#### Current Version:
v4.0.1

bbbales2:
Closed by #2423

rok-cesnovar:
@SteveBronder @bbbales2 so I need to set 

CXXFLAGS += -DSTAN_NO_RANGE_CHECKS 

and that escapes the range and size checks only?

bbbales2:
I don't think there's a guarantee that we are turning off every range and size check, but we're hopefully turning off enough that it counts performance-wise.

bbbales2:
And yeah, it's only range and size checks. All value checks should still be there.

rok-cesnovar:
Cool cool. 

I would add a makefile helper so that STAN_NO_RANGE_CHECKS=true adds `CXXFLAGS += -DSTAN_NO_RANGE_CHECKS`. I hate the CXXFLAGS += syntax :) Alo makes it much easier to add it as a cpp_options.

bbbales2:
> Alo makes it much easier to add it as a cpp_options.

Oh whoops, I thought this was somehow done automagically. I'll add to my list for today and if I don't get to it I'll open an issue for later.

rok-cesnovar:
[...]

PR #2423, opened 2021-03-13
adds macro for making bounds checks a no-op

SteveBronder:
## Summary

From issue #2420 the default in Stan now is to always perform bounds checks for single index access. This has a noticeable slowdown for users. This PR uses a `STAN_NO_RANGE_CHECKS` macro which if defined then makes the range and bounds checks into no-ops. It does this by defining another macro `STAN_NO_RANGE_CHECK_RETURN` which is equal to `return;` if `STAN_NO_RANGE_CHECKS` is defined.

At first I thought we could just use `NDEBUG` but it seems that is specifically to turn on and off asserts such as `eigen_assert()` or the standard `assert()`. Our bounds checks are sort of an assert, though I think it's safer just to have a new macro.

It's possible for the user to specify this in the makefile, or we could have an option for the compiler like `-fno-debug` which at the top of the stan model would define `STAN_NO_RANGE_CHECKS`

## Tests

tbh I'm not really sure how to test this since it changes behavior in the preprocessor. We could have a separate jenkins node that does the checks just for these errors? Since they are all no-ops it would be pretty fast

## Side Effects

Yes this whole PR is a side effect to turn off bounds checks.

## Release notes

Adds `STAN_NO_RANGE_CHECKS` macro which if defined turns off bounds and range checks

## Checklist

- [x] Math issue #2420


rok-cesnovar:
If we will be defining a separate macro, which I think is the right call here, might as well give it a more descriptive name? Most of Stan users are not C++ developers and NDEBUG does not mean much to them. Maybe STAN_NO_RANGE_CHECKS or just NO_RANGE_CHECKS.

rok-cesnovar:
As for testing, I think easiest is a Github Action job that runs a list of tests. These standalone short tests are perfect for GHA as its free.

SteveBronder:
> If we will be defining a separate macro, which I think is the right call here, might as well give it a more descriptive name? Most of Stan users are not C++ developers and NDEBUG does not mean much to them. Maybe STAN_NO_RANGE_CHECKS or just NO_RANGE_CHECKS.

I have to think about that. Messing with macro's is certainly an advanced user function. I use `STAN_NO_RANGE_CHECKS` for the `return;`, though I could check if it exists and then undef and def it 

> As for testing, I think easiest is a Github Action job that runs a list of tests. These standalone short tests are perfect for GHA as its free.

Nice! Yeah that makes total sense I'll add the tests and make a gh action

stan/math/prim/err/check_bounded.hpp
```diff
@@ -74,6 +74,7 @@ struct bounded<T_y, T_low, T_high, true> {
 template <typename T_y, typename T_low, typename T_high>
 inline void check_bounded(const char* function, const char* name, const T_y& y,
                           const T_low& low, const T_high& high) {
+  STAN_NO_RANGE_AND_SIZE_CHECK;
```
bbbales2, line 77: This is used to check that real values are in a certain range (like probabilities are bounded [0, 1]) as well as indexes in range.

It looks like this pull is only turning off range checks, but why not value checks too? I think you'd turn them off in similar situations for the same reasons.

bbbales2 (commented):
Comment.

t4c1:
> I use STAN_NO_RANGE_CHECKS for the return;, though I could check if it exists and then undef and def it

You should probably rename that to `STAN_NO_RANGE_CHECKS_RETURN` and define that if a user defines `STAN_NO_RANGE_CHECKS`.

stan/math/prim/meta/compiler_attributes.hpp
```diff
@@ -4,11 +4,39 @@
...
+ *  the associated function.
+ */
 #define STAN_COLD_PATH __attribute__((noinline, cold))
 #else
+#define STAN_COLD_PATH
+#endif
+#else
+#define STAN_COLD_PATH
+#endif
+#else
```
t4c1, line 21: These changes seem unrelated to what this PR is doing.

SteveBronder: This is a little bit of a fix, if someone is using either not gcc or clang then we are not guaranteed `__has_attribute` is defined. So to compile on those we want to make sure that macro is defined so that we can then check whether the compiler has the attributes `noinline` and `cold`

stan/math/prim/err/check_bounded.hpp
SteveBronder: Oh I forgot this was used for both. Umm I think @syclik had some opinions on this. Also pinging @wds15. I figure this is something users have to explicitly turn on so I'm more in favor of also not doing value checks

wds15: Value checks? Like what?

bbbales2: Not NaN, not infinite, positive definite, valid unit vector, valid simplex -- this sort of thing.

wds15: that's a different story which one should be able to turn off separately, I think. Something like STAN_FAST_MATH (a synonym for "not safe math, do it fast")

bbbales2: If we're allowing for segfaults, then I am also happy allowing for random math explosions. Even though these could happen in really weird, hard to debug ways, it's all better when the checks are flipped back on.

bbbales2: Well I guess, how is being unsafe on values practically different than being unsafe on indexing? A lot of the time even when we see these value check errors in practice we ignore them (scale parameter is zero, etc. etc.)

I suppose that these can happen occasionally even in a Stan model we are happy with, and we don't regularly expect to hit out of bounds indexing and stuff.

bbbales2: To not stall this unnecessarily, I think it's fine if we do:

1. The macro turns off some but not all index/value checks (this base set seems fine)
2. Only doing the actual index checks for now (`check_bounded` no longer optimized)
3. The macro turns off some but not all index/value checks and we expand it to turn off inf/nan checks
4. Some other variation on a theme

I like one macro better than to two because it would probably just get confusing to have two (for functions like this for instance) and I think the conditions for no index checks are similar to no value checks.

Easy enough to expand the set of functions optimized in production mode at a later time if we only want to do strict indices.

stan/math/prim/meta/compiler_attributes.hpp
t4c1: Ok, but you could simplify this into:
```suggestion
#ifdef __has_attribute
#if __has_attribute(noinline) && __has_attribute(cold)
/**
 * Functions tagged with this attribute are not inlined and moved
 *  to a cold branch to tell the CPU to not attempt to pre-fetch
 *  the associated function.
 */
#define STAN_COLD_PATH __attribute__((noinline, cold))
#endif
#ifndef STAN_COLD_PATH
#define STAN_COLD_PATH
#endif
```

stan/math/prim/err/check_bounded.hpp
wds15: Hmmm... that may work... but what is wrong with two macros which do a more tailored thing themselves. To retain easy usability we can introduce things like

```c++
#ifdef NDEBUG
#define MACRO1
#define MACRO2
#endif
```

How does that sound?

SteveBronder: I like in the future having a `STAN_NDEBUG` which turns off multiple types of checks. Following @t4c1 I changed the name of this macro to `STAN_NO_RANGE_CHECKS` which defines `STAN_NO_RANGE_CHECKS_RETURN return`

SteveBronder: Are we also not going to remove size checks?

bbbales2: > but what is wrong with two macros which do a more tailored thing themselves

I don't see why one would be useful without another -- unsafe is unsafe and it's just hard to remember multiple things.

> not going to remove size checks?

Optimizing out size checks makes sense with this macro I think. `check_bounded` is probly the only one that needs to go if we're not using values.

If we're going for STAN_NDEBUG in the future, does it make sense to just start with that? Or STAN_UNSAFE or STAN_NO_DEBUG since Rok pointed out NDEBUG probably doesn't mean much to people? Then in the future if we decide to split that into multiple things we can and the overall super-unsafe easy-mode flag is still there.

SteveBronder: I think I'd like to leave the size checks for another time because it changes the exceptions that are returned for some functions like `check_symmetric()` etc.

SteveBronder: > If we're going for STAN_NDEBUG in the future, does it make sense to just start with that? Or STAN_UNSAFE or STAN_NO_DEBUG since Rok pointed out NDEBUG probably doesn't mean much to people? Then in the future if we decide to split that into multiple things we can and the overall super-unsafe easy-mode flag is still there.

Let's do this in a separate PR once we sort out what that means. Right now I like just having an option to turn off the range checks, then in another PR or an issue we can sort out what we want turned on and off

bbbales2: Sounds good

stan/math/prim/meta/compiler_attributes.hpp
```diff
@@ -2,13 +2,48 @@
 #define STAN_MATH_PRIM_META_COMPILER_ATTRIBUTES_HPP
 
 #ifdef __GNUC__
+#ifndef likely
```
bbbales2, line 5: Is it that old gccs didn't have likely and new ones do?

bbbales2 (approved):
This all looks good to me. Since we've been doing this pull by committee, I'll leave it to someone else to also look over and click the merge button.

stan/math/prim/meta/compiler_attributes.hpp
SteveBronder: Another package could have it

Issue #2430, opened 2021-03-17
suggest extracting forward declaration of stan::math::var_value into its own header


chenel:
Currently (as of version 4.0.1) there's a forward declaration of the templated `stan::math::var_value` inside `vari.hpp`:
https://github.com/stan-dev/math/blob/040d515a6551af3aeb046b4816c1eb3a9e6ba527/stan/math/rev/core/vari.hpp#L18

If however I want to also forward declare `var_value` in my code that builds against Stan-math (because I _actually_ want to forward declare the typedef'd `stan::math::var`, because I have some custom types templated over both `stan::math::var` and `double`, and these types get pushed around into a lot of other headers; `#include stan/math/rev.hpp` slows compilation times down a _lot),_ I can't just copy-paste that code into my own headers. If I try,  because `var_value` has a default template parameter, gcc throws
```
vari.hpp:20:7: error: redefinition of default argument for ‘class<template-parameter-1-2>’
 class var_value;
```
After some poking around, I discovered [this solution](https://stackoverflow.com/a/2308617) on a StackOverflow question, which seems pretty reasonable---put the forward declaration in its own header, which can then be `#include`d wherever necessary.  Because it's then protected by an `#ifdef` over a `#define`, the forward declaration will only ever be read a maximum of once.

I have a branch in my fork of Stan-math with a minimal change to implement this:
https://github.com/chenel/math/commit/1dfd0fe1e322e03e866ad8e80b238e1cdb4b6fc3

But I was scared off from making a pull request by the testing requirements (this is my first contribution here and I'm not very familiar with the tests). :sweat_smile: 
I also don't know whether there are any other places a similar change ought to be made.

I don't _think_ there are likely any side effects, but you can judge for yourselves...

rok-cesnovar:
Thanks for the report.

If the change is simple and is just moving some code around, then just open a PR and it will run all the tests automatically.

i would suggest opening a PR and we can discuss the changes there more easily.

SteveBronder:
I think this would be nice a PR would be much appreciated! There's def a few spots where we do this and if it's easier for people to fix this then I'm 100% about it

chenel:
Ok!  pull request created, see above.

chenel:
Pull request merged, so closing.  Thanks everyone for getting this looked at so fast!

PR #2431, opened 2021-03-18
Extract forward declaration of stan::math::var_value into its own header

chenel:
See #2430 for context.

This pull request would enable external code to also take advantage of forward declarations of the templated `var_value` by extracting the relevant forward declaration lines from `vari.hpp` into their own header, `var_values_fwd_declare.hpp`.  It then `#include`s this new header in `vari.hpp`.

It may make sense to do something similar with other forward declarations (I saw another for `vari_value` in that same file, and a few others when grepping around in other places).  But I don't know how externally-facing those other instances are meant to be, whereas `var` (==`var_value<double>`) is definitely user-facing.

I don't expect this to have any side effects.

### DIFF
--- a/stan/math/rev/core/var_value_fwd_declare.hpp
+++ b/stan/math/rev/core/var_value_fwd_declare.hpp
@@ -1,0 +1,11 @@
+#ifndef STAN_MATH_REV_CORE_VAR_VALUE_FWD_DECLARE_HPP
+#define STAN_MATH_REV_CORE_VAR_VALUE_FWD_DECLARE_HPP
+
+namespace stan {
+namespace math {
+// forward declaration of var
+template <typename T, typename = void>
+class var_value;
+}
+}
+#endif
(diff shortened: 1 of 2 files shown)


bbbales2 (approved):
Looks good to me.

bbbales2:
@SteveBronder @rok-cesnovar this all look good? Seems fine to me.

rok-cesnovar:
Looks good, not sure of the name `stan/math/rev/core/var_value_fwd_declare.hpp` but I dont have a better suggestion...

bbbales2:
Alright since we got two looks goods let's just merge away. If there's another change we need, our tests are down to 4 hours now!

bbbales2 (approved):
Good

SteveBronder:
Alright @rybern you should be free to do the free stuff!

chenel:
Thanks everybody!

PR #2453, opened 2021-03-31
Reverse mode autodiff for diag_post_multiply

gregorp90:
## Summary

Implemented custom reverse mode AD for `diag_post_multiply()` function. The implementation is based on the reverse mode implementation of `diag_pre_multiply()`.

## Tests

Added a test for prim where the arguments are of length zero.

Removed manual `ad_tolerances` from the mix test, similar to `diag_pre_multiply()`. Also added 4 mix tests, where the matrix is non-square.

## Side Effects

None.

## Release notes

Added custom reverse mode for diag_post_multiply() function.

## Checklist

- [x] Math issue #989

- [x] Copyright holder: Gregor Pirš


rok-cesnovar:
@gregorp90 I pushed a fix for an OpenCL test issue that surfaced here. 

@t4c1 @bbbales2 can I interest any one of you for a review?

gregorp90:
> @gregorp90 I pushed a fix for an OpenCL test issue that surfaced here.

@rok-cesnovar Thank you!

stan/math/prim/fun/diag_post_multiply.hpp
```diff
@@ -19,8 +19,10 @@ namespace math {
  * @return product of the matrix and the diagonal matrix formed from the
  * vector or row_vector.
  */
-template <typename T1, typename T2, require_eigen_t<T1>* = nullptr,
-          require_eigen_vector_t<T2>* = nullptr>
+template <typename T1, typename T2,
+          require_eigen_matrix_dynamic_t<T1>* = nullptr,
```
t4c1, line 23: Why the change to this require to not allow vectors?

t4c1: As far as I can see this is what required the change to OpenCL tests (although new test is fine as well).

t4c1 (commented):
Looks good, just one question.

stan/math/prim/fun/diag_post_multiply.hpp
gregorp90: I see, so this should be `require_eigen_t` instead of `require_eigen_matrix_dynamic_t`?

I based this implementation on the implementation of `diag_pre_multiply`. In that implementation, Ben made this change, as you can see [here](https://github.com/stan-dev/math/pull/2405/commits/ef2362066f0786991dc42d3a9aa32b872b7f9cef). Maybe @bbbales2 can provide some more details on this change?

Also, if we are already on this subject. Why does the prim test with 1x1 matrix not fail, if we have `require_eigen_matrix_dynamic_t`?

t4c1: Yeah `require_eigen_t` or `require_eigen_matrix_t` is fine.

> Also, if we are already on this subject. Why does the prim test with 1x1 matrix not fail, if we have require_eigen_matrix_dynamic_t?

It is not the shape of the input that is the issue, but type. We have general matrix type (that can also contain a vector) and special types for row or column vector. The new require does not accept the specialized vector types, while the test probably uses general matrix type.

gregorp90: So should I revert back to `require_eigen_t` or leave the implementation as it is?

Thanks for the answer, makes things more clear now!

t4c1: I think you should revert.

gregorp90: OK.
Should I make the same changes for `diag_pre_multiply`?

gregorp90: @t4c1 I made the changes to both, `diag_pre_multiply` and `diag_post_multiply`.

t4c1 (approved):

PR #2455, opened 2021-03-31
allow for tbb init to set the number of threads by an argument

SteveBronder:
## Summary

This adds an argument to `init_threadpool_tbb()` so we can set the number of threads by an argument

## Tests

Test added for passing the number of threads as an argument

## Release notes

Allow tbb init to set the number of threads by an argument.

## Checklist

### DIFF
--- a/stan/math/prim/core/init_threadpool_tbb.hpp
+++ b/stan/math/prim/core/init_threadpool_tbb.hpp
@@ -88,12 +88,20 @@ inline int get_num_threads() {
  * The function returns a reference to the static
  * tbb::global_control instance.
  *
+ * @param n_threads The maximum number of threads available to the tbb. If not
+ *  set will search for the environment variable `STAN_NUM_THREADS`. A value of
+ *  will assume all detectable threads are available for the process.
  * @return reference to the static tbb::global_control
  * @throws std::runtime_error if the value of STAN_NUM_THREADS env. variable
  * is invalid
  */
-inline tbb::task_arena& init_threadpool_tbb() {
-  int tbb_max_threads = internal::get_num_threads();
+inline tbb::task_arena& init_threadpool_tbb(size_t n_threads = 0) {
+  int tbb_max_threads;
+  if (n_threads == 0) {
+   tbb_max_threads = internal::get_num_threads();
+  } else {
+   tbb_max_threads = n_threads;
+  }
   static tbb::global_control tbb_gc(
       tbb::global_control::max_allowed_parallelism, tbb_max_threads);
 
@@ -118,18 +126,21 @@ inline tbb::task_arena& init_threadpool_tbb() {
  * The function returns a reference to the static
  * tbb::task_scheduler_init instance.
  *
- * @param stack_size sets the stack size of each thread; the default 0
- * let's the TBB choose the stack size
+ * @param n_threads The maximum number of threads available to the tbb. If not
+ *  set will search for the environment variable `STAN_NUM_THREADS`. A value of
+ *  will assume all detectable threads are available for the process.
  * @return reference to the static tbb::task_scheduler_init
  * @throws std::runtime_error if the value of STAN_NUM_THREADS env. variable
  * is invalid
  */
-inline tbb::task_scheduler_init& init_threadpool_tbb(
-    tbb::stack_size_type stack_size = 0) {
-  int tbb_max_threads = internal::get_num_threads();
-
-  static tbb::task_scheduler_init tbb_scheduler(tbb_max_threads, stack_size);
-
+inline tbb::task_scheduler_init& init_threadpool_tbb(size_t n_threads = 0) {
+  int tbb_max_threads;
+  if (n_threads == 0) {
+   tbb_max_threads = internal::get_num_threads();
+  } else {
+   tbb_max_threads = n_threads;
+  }
+  static tbb::task_scheduler_init tbb_scheduler(tbb_max_threads, 0);
   return tbb_scheduler;
 }
 #endif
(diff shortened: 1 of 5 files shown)


stan/math/prim/core/init_threadpool_tbb.hpp
```diff
@@ -88,12 +88,20 @@ inline int get_num_threads() {
  * The function returns a reference to the static
  * tbb::global_control instance.
  *
+ * @param n_threads The maximum number of threads available to the tbb. If not
+ *  set will search for the environment variable `STAN_NUM_THREADS`. A value of
+ *  will assume all detectable threads are available for the process.
  * @return reference to the static tbb::global_control
  * @throws std::runtime_error if the value of STAN_NUM_THREADS env. variable
  * is invalid
```
hsbadr, line 96:
```suggestion
 * @throws std::runtime_error if n_threads (defaults to zero) is not provided and
 * the value of STAN_NUM_THREADS environment variable is invalid.
```

stan/math/prim/core/init_threadpool_tbb.hpp
```diff
@@ -88,12 +88,20 @@ inline int get_num_threads() {
...
  */
-inline tbb::task_arena& init_threadpool_tbb() {
-  int tbb_max_threads = internal::get_num_threads();
+inline tbb::task_arena& init_threadpool_tbb(size_t n_threads = 0) {
+  int tbb_max_threads;
+  if (n_threads == 0) {
+    tbb_max_threads = internal::get_num_threads();
+  } else {
+    tbb_max_threads = n_threads;
+  }
```
hsbadr, line 104: Need to check if `n_threads` is a valid integer greater than zero and smaller than or equal to the maximum number of threads supported by the hardware (`0 <= n_threads <= tbb::this_task_arena::max_concurrency()`).

SteveBronder: I updated this to account for the above and to allow for n_threads = -1 to mean `std::thread::hardware_concurrency()`

test/unit/math/prim/core/set_num_threads_test.cpp
```diff
@@ -0,0 +1,10 @@
+#include <stan/math/prim/core.hpp>
+#include <gtest/gtest.h>
+
+TEST(intel_tbb_new_init, set_threads) {
+  int num_threads = std::thread::hardware_concurrency();
+  auto& tbb_init = stan::math::init_threadpool_tbb(num_threads);
+  EXPECT_TRUE(tbb_init.is_active());
+
+  EXPECT_EQ(num_threads, tbb_init.max_concurrency());
+}
```
SteveBronder, line 10: Weird Jenkins testing thing, I tested this locally with 
```cpp
int num_threads = std::thread::hardware_concurrency() - 1;
```
but it won't work on Jenkins because we merge all these tests together so by the time this tests run the max threadpool has already been set so setting it lower doesn't do anything. imo I think it's find unless theres some sort of quick fix

hsbadr: @SteveBronder Try to run the tests inside the initialized arena, using `tbb_arena.execute([&]() {...})` like:
https://github.com/stan-dev/math/blob/818be4ae86ae91af53943dc742ce977102bd1689/test/unit/math/prim/core/init_threadpool_tbb_late_test.cpp#L17-L24

or terminate the arena after each test, using `tbb_arena.terminate()`.

rok-cesnovar: Or exclude prim/core from jumbo here: https://github.com/stan-dev/math/blob/develop/runTests.py#L35

SteveBronder: @hsbadr I couldn't figure out how to do the above. Even if you terminate the task arena are you able to then later change the number of threads the global task arena has?

@rok-cesnovar I went with your suggestion as it's a bit easier and we only start this up once at runtime so it makes sense to test this individually

hsbadr: Sounds good. I don't know how important the excluded `prim/core` tests are, but it's indeed the easiest workaround.

`stan::math::init_threadpool_tbb()` creates static instances of `global_control` and `arena`, which restores previous parallelism limitation on exit. Also, `arena.execute()` should always work as expected, and it's the best solution if we'd like to run Stan services in separate TBB arenas, with custom `num_threads`, for example.

rok-cesnovar:
@hsbadr are you good with the changes? Mind having another look?

hsbadr (approved):

hsbadr:
LGTM.

rok-cesnovar (approved):
Thanks Steve and @hsbadr for reviewing!

test/unit/math/prim/core/set_num_threads_test.cpp
SteveBronder: ftr they are not excluded they are just not put together into one big file and instead each run as separate test programs.

> stan::math::init_threadpool_tbb() creates static instances of global_control and arena, which restores previous parallelism limitation on exit. Also, arena.execute() should always work as expected, and it's the best solution if we'd like to run Stan services in separate TBB arenas, with custom num_threads, for example.

I thought I was doing it right but putting them all together into one file I wasn't able to change the number of threads. Maybe I just read the docs wrong?

hsbadr: > ftr they are not excluded they are just not put together into one big file and instead each run as separate test programs.

Ah! I see.

> I thought I was doing it right but putting them all together into one file I wasn't able to change the number of threads. Maybe I just read the docs wrong?

I think we can't reset the parallelism limitation (`tbb::global_control::max_allowed_parallelism`, set statically by `stan::math::init_threadpool_tbb()`) until we exit from main (or the function that created it). Alternatively, we can terminate the arena and initiate a new one. This is typically done by setting a `global_control` that limits the maximum parallelism and schedule/subset the number of threads by multiple arenas (executing code inside each arena via `arena.execute()`) with `num_threads <= bb::global_control::max_allowed_parallelism`.

This is a lot of work though, as you need to edit all the code that run in parallel.

PR #2472, opened 2021-04-16
Add kernel generator cast operation

t4c1:
## Summary

Adds typecast operation to kernel generator and uses it in distributions instead of forcing type promotions by adding zero.

## Tests
Added tests for cast. Reminder of this PR is just refactor.

## Side Effects
None.

## Release notes

Adds typecast operation to kernel generator.

## Checklist

### DIFF
--- a/stan/math/opencl/kernel_generator/cast.hpp
+++ b/stan/math/opencl/kernel_generator/cast.hpp
@@ -39,7 +39,7 @@ class cast_ : public operation_cl<cast_<Scal, T>, Scal, T> {
    * Constructor
    * @param args argument expression(s)
    */
-  cast_(T&& arg) : base(std::forward<T>(arg)) {}
+  explicit cast_(T&& arg) : base(std::forward<T>(arg)) {}
 
   /**
    * Generates kernel code for this expression.
--- a/stan/math/opencl/kernel_generator/cast.hpp
+++ b/stan/math/opencl/kernel_generator/cast.hpp
@@ -1,0 +1,102 @@
+#ifndef STAN_MATH_OPENCL_KERNEL_GENERATOR_CAST_HPP
+#define STAN_MATH_OPENCL_KERNEL_GENERATOR_CAST_HPP
+#ifdef STAN_OPENCL
+
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/opencl/matrix_cl_view.hpp>
+#include <stan/math/opencl/kernel_generator/common_return_scalar.hpp>
+#include <stan/math/opencl/kernel_generator/type_str.hpp>
+#include <stan/math/opencl/kernel_generator/name_generator.hpp>
+#include <stan/math/opencl/kernel_generator/operation_cl.hpp>
+#include <stan/math/opencl/kernel_generator/as_operation_cl.hpp>
+#include <array>
+#include <string>
+#include <type_traits>
+#include <set>
+#include <utility>
+
+namespace stan {
+namespace math {
+
+/** \addtogroup opencl_kernel_generator
+ *  @{
+ */
+
+/**
+ * Represents a typecast os scalar in kernel generator expressions.
+ * @tparam Derived derived type
+ * @tparam T type of argument
+ * @tparam Scal type of the scalar of result
+ */
+template <typename Scal, typename T>
+class cast_ : public operation_cl<cast_<Scal, T>, Scal, T> {
+ public:
+  using Scalar = Scal;
+  using base = operation_cl<cast_<Scal, T>, Scalar, T>;
+  using base::var_name_;
+
+  /**
+   * Constructor
+   * @param args argument expression(s)
+   */
+  cast_(T&& arg) : base(std::forward<T>(arg)) {}
+
+  /**
+   * Generates kernel code for this expression.
+   * @param row_index_name row index variable name
+   * @param col_index_name column index variable name
+   * @param view_handled whether whether caller already handled matrix view
+   * @param var_names_arg variable names of the nested expressions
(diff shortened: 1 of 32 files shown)


stan/math/opencl/prim/bernoulli_cdf.hpp
```diff
@@ -49,13 +49,13 @@ return_type_t<T_prob_cl> bernoulli_cdf(const T_n_cl& n,
...
 
-  auto any_n_negative = colwise_max(constant(0, N, 1) + (n < 0));
+  auto any_n_negative = colwise_max(cast<char>(n < 0));
   auto cond = n >= 1;
   auto Pi_uncond = 1.0 - theta_val;
   auto Pi = select(cond, INFTY, Pi_uncond);
   auto P_expr = colwise_prod(select(cond, 1.0, Pi_uncond));
 
-  matrix_cl<double> any_n_negative_cl;
+  matrix_cl<char> any_n_negative_cl;
```
SteveBronder, line 58: Why is this a car and not a bool?

t4c1: OpenCL standard does not specify what is the size of bool, so variables of type bool effectively can not be transferred between host and device (as we need to specify size of the array we are transferring).

SteveBronder: Can you link me where it says that? I'm only seeing in section [6.6.1](https://www.khronos.org/registry/OpenCL/specs/opencl-1.2.pdf) that bool is expanded to an integer 

> bool: A conditional data type which is either true or false. The value true expands to the integer constant 1 and the value false expands to the integer constant 0.

t4c1: The important thing is what is not in the standard. It never mentions what is the size of bool or what is equivalent host type. Equivalent host types are listed in table 6.1 and for bool it says n/a.

t4c1:
This is still waiting for a review.

stan/math/opencl/kernel_generator/cast.hpp
```diff
@@ -0,0 +1,101 @@
...
+   * @return part of kernel with code for this expression
+   */
+  inline kernel_parts generate(const std::string& row_index_name,
+                               const std::string& col_index_name,
+                               const bool view_handled,
+                               const std::string& var_name_arg) const {
+    kernel_parts res{};
+
+    res.body = type_str<Scalar>() + " " + var_name_ + " = ("
+               + type_str<Scalar>() + ")" + var_name_arg + ";\n";
```
SteveBronder, line 59: So this line here generates `Scalar var_name = (Scalar)input_var_name;`? If so then idt I'm understanding how this works for `char`?

SteveBronder (changes requested):
I think the code looks good but I'm having a hard time following how the cast happens on the device side. I think for this operation in particular where we are sort of fighting a thing in the OpenCL spec it's very good to have more particular docs. If you can clear that up then I think this is good to go

Side note, not necessary for this PR, but I think this is an example where it would be nice to have something in the tests or an easy scheme available from `runTests.py` for generating the OpenCL kernel code. It would also make reviewing these things a lot simpler. Sort of like how in the stanc3 compiler we have the [generated cpp](https://github.com/stan-dev/stanc3/blob/master/test/integration/good/code-gen/cpp.expected) as part of the testing so we can see what stuff changes in the gen'd code between PRs

stan/math/opencl/prim/bernoulli_cdf.hpp
SteveBronder: Okay I found the relevant docs

https://stackoverflow.com/questions/4441280/does-opencl-support-boolean-variables

> Arguments to __kernel functions in a program cannot be declared with the built-in scalar types bool, half, size_t, ptrdiff_t, intptr_t, and uintptr_t. The size in bytes of these types except half are implementation-defined and in addition can also be different for the OpenCL device and the host processor making it difficult to allocate buffer objects to be passed as arguments to a kernel declared as pointer to these type

stan/math/opencl/kernel_generator/cast.hpp
t4c1: Yes. I don't understand what is the confusion with char. It works exactly the same as for any other supported scalar type.

> Side note, not necessary for this PR, but I think this is an example where it would be nice to have something in the tests or an easy scheme available from runTests.py for generating the OpenCL kernel code.

We do have reference kernels in tests, although I am not using them for operations that generate really trivial code, such  as typecasting in this PR.

t4c1:
@SteveBronder This PR is still waiting.

stan/math/opencl/kernel_generator/cast.hpp
SteveBronder: So I'm looking at 

```
auto any_n_negative = colwise_max(cast<char>(n < 0));
```

And here this is turned into 

```cpp
char var_name = (char)input_var_name;
```

Apologies if this is silly but I'm not following how that is used in an if etc. Like is it that if `input_var_name` is 0 then the `char` is 0 and the if is false?

t4c1: Yes, exactly. `input_var_name` is in this case the result of the comparison `n < 0`, which is either 0 or 1. 0 or 1 converted into char is still 0 or 1. That gets written into global memory and than copied to host. Is there still anything you are not understanding?

SteveBronder: Aight then I think that makes sense

SteveBronder (approved):
It would be good to have a note on the behavior for this but I think it is fine as is

PR #2479, opened 2021-04-26
Add OpenCL QR decomposition

t4c1:
## Summary
Adds OpenCL implementation for QR decomposition, including functions `qr_Q`, `qr_R`, `qr_thin_Q`, `qr_thin_R` (prim signatures only). 

Also includes a bugfix for kernel generator block, which could set view of the underlying matrix wrong when assigned to.

## Tests
Added tests for fixed block and new functions. Testing QR decomposition got a bit tricky, as it is not unique. So instead of comparing CPU and GPU results, the tests are checking properties of the resulting Q and R matrices.

## Side Effects
None.

## Release notes

OpenCL: Added OpenCL implementation for QR decomposition, including functions `qr_Q`, `qr_R`, `qr_thin_Q`, `qr_thin_R` (prim signatures only). 

Fixed a bug for kernel generator operation `block`.

## Checklist


SteveBronder:
Nice! @LuZhangstat was asking about this in a separate thread. 

I just crack'd open Numerical Recipes (lol) but I'm not sure if my chops are up to reviewing this. Can you post a link to some reference docs? Looking at the code though it looks like most of the work actually happens on the CPU side? Can you post benchmarks and some comparisons to Eigen's Householder QR decomp? It would be nice to see things like the maximum numerical difference and average numerical difference for both

LuZhangstat:
That's great to know! Thank you so much. @SteveBronder 

> Nice! @LuZhangstat was asking about this in a separate thread.

t4c1:
> Can you post a link to some reference docs?

I was following the article: https://cpb-us-w2.wpmucdn.com/sites.gatech.edu/dist/5/462/files/2016/08/Kerr_Campbell_Richards_QRD_on_GPUs.pdf

> Looking at the code though it looks like most of the work actually happens on the CPU side?

Lol, most code does not mean most work. Actually when this outperforms sequential implementation most of the time is spent for matrix multiplies on the GPU. I also tried implementing a version that does everything on the GPU, but it turned out to be a bit slower. For a matrix of size NxN it required extra 4N kernel calls, so launching that many kernels caused significant overhead.

Benchmark of `qr_Q` (which needs to calculate both Q and R matrices):
qr_Q_cl

Benchmark of `qr_R` (which can skip the calculation of Q):
qr_R_cl

t4c1:
Numerical errors (QR decomposition of random matrix m with coefficients between -1 and 1):

GPU:
| rows | cols | max(abs(m - Q * R)) | avg(abs(m - Q * R)) | max(abs(I - QT*Q)) | avg(abs(I - QT*Q)) |
| - | - | - | - | - | - |
| 10 | 10 | 9.99201e-16 | 2.69298e-16 | 1.77636e-15 | 2.24196e-16 |
| 2000 | 2000 | 1.11244e-13 | 1.58065e-15 | 5.32907e-15 | 7.17297e-17 |
| 1000 | 4000 | 3.9857e-14 | 1.37708e-15 | 3.10862e-15 | 7.39562e-17 |
| 4000 | 1000 | 2.11942e-13 | 1.54228e-15 | 5.55112e-15 | 4.25013e-17 |

CPU Eigen:
| rows | cols | max(abs(m - Q * R)) | avg(abs(m - Q * R)) | max(abs(I - QT*Q)) | avg(abs(I - QT*Q)) |
| - | - | - | - | - | - |
| 10 | 10 | 6.66134e-16 | 1.19141e-16 | 4.44089e-16 | 8.06993e-17 |
| 2000 | 2000 | 7.77156e-15 | 7.53788e-16 | 1.55431e-15 | 4.52808e-17 |
| 1000 | 4000 | 5.38458e-15 | 7.34793e-16 | 1.77636e-15 | 4.68622e-17 |
| 4000 | 1000 | 8.54872e-15 | 5.6134e-16 | 6.21725e-15 | 2.50325e-17 |

So Eigen has a bit tighter numerical tolerances, but I think the errors are still quite small.

SteveBronder:
Those seem reasonable, can you post up the code for the numeric check? Is this with `HouseholderQR` or `ColPivHouseholderQR`?

t4c1:
> Those seem reasonable, can you post up the code for the numeric check? Is this with HouseholderQR or ColPivHouseholderQR?

I just slightly modified tests that are already in this PR to print the errors. Its comparing CPU and OpenCL `qr_Q` and `qr_R` (with CPU implementation using `HouseholderQR` without pivoting).

SteveBronder:
Mkay I think these make sense. I'm tagging @bgoodri as in the Stan meeting he said he could have a look at this and has more experience with the inner machinations of solvers.

Ben the main code is [here](https://github.com/stan-dev/math/pull/2479/files#diff-800b50d1d663f24054e52389c0c714da7acbb23a1df329c5236547c71940c294R30) for the block householder with OpenCL. There's actually no OpenCL code used directly there though as it just uses the higher level interface available in Stan's OpenCL backend.

t4c1:
@bgoodri Do you plan on reviewing this? Or should someone else do it?

bgoodri:
I think the QR decomposition needs to be made unique by requiring the diagonal elements of R to be non-negative, which may require flipping the sign of the corresponding elements of Q. That is what we did with the all the `qr_` functions on the CPU (see https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/qr_Q.hpp#L29 ), even though it is slower.

Most of the time, you can get away with a non-unique QR decomposition, such as when it is called on fixed matrices or when using the QR decomposition to do a reduction that is invariant to the sign flips like calculating the determinant. But it is theoretically possible that someone would do the QR decomposition on a `varmant` and then do something involving individual cells of the resulting Q and / or R matrices, in which case the lack of uniqueness would manifest itself in discontinuities in the unconstrained space that would mess up HMC.

t4c1:
This implementation already returns result that has nonnegative entries for diagonal or R. But in case of a matrix with more rows than columns that is not enough to make the decomposition unique.

t4c1:
@bgoodri When will you be able to take time for the review?

stan/math/opencl/prim/qr_Q.hpp
```diff
@@ -0,0 +1,38 @@
...
+          require_all_kernel_expressions_and_none_scalar_t<T_m>* = nullptr>
+inline matrix_cl<double> qr_Q(T_m&& m) {
+  check_nonzero_size("qr_Q(OpenCL)", "m", m);
+
+  matrix_cl<double> mat_eval = std::forward<T_m>(m);
+  matrix_cl<double> Q, R;
+
+  qr_decomposition_cl(m, Q, R);
+
+  auto Q_left
```
bgoodri, line 28: What is the purpose of these two lines involving Q_left?

stan/math/opencl/qr_decomposition.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <CL/cl.hpp>
+#include <algorithm>
+#include <vector>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/**
+ * Calculates QR decomposition of A using the block Householder algorithm.
```
bgoodri, line 22: might add link to description of block Householder on GPUs

stan/math/opencl/prim/qr_Q.hpp
t4c1: `Q_left` is the part of Q that is affected by the signs of diagonal elements of R. So this is the part that needs to be changed when ensuring R has positive diagonal elements. So the first line defines this part and the second one selects either elements from this as they are or negated ones, depending on the diagonal of R.

stan/math/opencl/qr_decomposition.hpp
t4c1: You mean the article I followed when implementing this?

stan/math/opencl/prim/qr_Q.hpp
bgoodri: OK, so the two lines involving `Q_left` have the side effect of changing `Q`? Otherwise, I don't see why `Q` is returned rather than `Q_left`?

stan/math/opencl/qr_decomposition.hpp
bgoodri: Sure

stan/math/opencl/kernel_generator/block_zero_based.hpp
```diff
@@ -272,9 +272,19 @@ class block_
...
+                 && start_col_ + cols_ >= std::min(a.rows() - 1, a.cols())
+             ? bottom_zero_diagonal
+             : bottom_diagonal)
+            + change,
+        (start_row_ == 0 && start_col_ <= 1 && start_col_ + cols_ == a.cols()
+                 && start_row_ + rows_ >= std::min(a.rows(), a.cols() - 1)
+             ? top_zero_diagonal
+             : top_diagonal)
+            + change);
   }
```
bgoodri, line 289: I suppose this fix to the `kernel_generator` must be correct in order for the tests to pass, but I have no clue about what the problem it fixes was (is there an issue?). If someone else says the fix is correct, then we can merge it.

stan/math/opencl/prim/qr_Q.hpp
t4c1: Yes, Q_left is a view (it references Q's data), so modifying Q_left modifies Q.

stan/math/opencl/kernel_generator/block_zero_based.hpp
t4c1: YEah, maybe @rok-cesnovar can review this change.

stan/math/opencl/prim/qr_Q.hpp
bgoodri: OK. Maybe that is self-evident to people who know the OpenCL interface.

stan/math/opencl/kernel_generator/block_zero_based.hpp
rok-cesnovar: Looks good to me.

t4c1:
@bgoodri @rok-cesnovar This can be approved and merged, right?

rok-cesnovar:
All good on my end.

bgoodri:
Yes

rok-cesnovar (approved):

Issue #2498, opened 2021-06-08
OpenCL symmetric eigensolver


t4c1:
## Description
I have implemented a symmetric eigensolver. Since it is a lot of code I will split it into multiple pull requests. So I am opening this to track progress.

I have recently mentioned the idea here: https://discourse.mc-stan.org/t/opencl-symmetric-eigendecomposition/22276
The eigensolver I implemented consists of block householder tridiagonalization and MRRR - multiple relatively robust representations. Under certain circumstances MRRR was not accurate enough, so I implemented it using double double - a 128 bit scalar, with similar accuracy to IEEE quadruple precision, but faster to calculate with.

Planned pull requests:
- tridiagonalization
- double double
- MRRR

If anyone is interested in complete code it is available here: https://github.com/bstatcomp/math/tree/opencl_eigendecomposition

## Example
`eigenvalues_sym` and `eigenvectors_sym` can be computed on GPUs.

## Expected Output
`eigenvalues_sym` and `eigenvectors_sym` are faster when using a GPU.

#### Current Version:
v4.1.0

PR #2500, opened 2021-06-08
OpenCL symmetric eigensolver 2/3: double double

t4c1:
## Summary

Adds implementation for double double - a 128 bit scalar, with similar accuracy to IEEE quadruple precision, but faster to calculate with. It is defined as a struct of 2 doubles, with its value being the exact sum of the two doubles. Those two doubles are high and low, with low having absolute value less than 1 ULP of high. This adds the following operations on double double: +, -, *, /, comparisons, abs, copysign, abs, isinf, isnan (these are needed in MRRR).

## Tests
Added tests for double double operations. Basic tests can be run anywhere, but the best test requires the GCC extension `quadmath` (IEEE quadruple precision numbers - the tests compares results between quad precision and double double). There is also a test that checks double doubles work in OpenCL kernels.

## Side Effects

The operations on double double are defined in a way that makes them available to both host C++ code and OpenCL kernels. For C++ this require some using statements in the namespace where these are defined - currently `stan::math::internal` (the only alternative I can think of would require duplicating all the implementations).

Since we will need double doubles in OpenCL kernels we need to remove the `require_arithmetic` on `matrix_cl` scalars.

## Checklist

- [ ] Math issue #2498

### DIFF
--- a/stan/math/opencl/double_d.hpp
+++ b/stan/math/opencl/double_d.hpp
@@ -274,7 +274,6 @@ const char* double_d_src = STRINGIFY(
         } inline bool ge_d_dd(double a, double_d b) {
           return !lt_d_dd(a, b);
         } inline bool ge_dd_d(double_d a, double b) { return !lt_dd_d(a, b); }
-
         // \cond
     );
 // \endcond
--- a/stan/math/opencl/double_d.hpp
+++ b/stan/math/opencl/double_d.hpp
@@ -275,7 +275,7 @@ const char* double_d_src = STRINGIFY(
           return !lt_d_dd(a, b);
         } inline bool ge_dd_d(double_d a, double b) { return !lt_dd_d(a, b); }
         // \cond
-    );
+    ); //NOLINT(whitespace/parens)
 // \endcond
 
 inline double_d operator+(double_d a, double b) { return add_dd_d(a, b); }
--- a/stan/math/opencl/double_d.hpp
+++ b/stan/math/opencl/double_d.hpp
@@ -44,6 +44,47 @@ using std::isinf;
 using std::isnan;
 using std::ldexp;
 
+inline double_d mul_d_d(double a, double b) {
+  double_d res;
+  // for some reason the std::fma only gives accurate result if compiling for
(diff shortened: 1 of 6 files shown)


stan/math/opencl/double_d.hpp
```diff
@@ -0,0 +1,322 @@
...
+    STRINGIFY2(
+        // \endcond
+
+        //    inline double_d mul_d_d(double a, double b) {
+        //      double_d res;
+        //      volatile double temp = a * b;
+        //      res.high = temp;
+        //      res.low = fma(a, b, -temp);
+        //      return res;
+        //    }
```
t4c1, line 65: This alternative implementation of multiplication of two doubles into double double using fma should be much faster, but for some reason I can't get it to work correctly. It might be compiler is optimizing the code too much. Does anyone have any ideas how to fix this?

rok-cesnovar: Missed this. This doesnt work with C++ or OpenCL or both?

rok-cesnovar: I think we can proceed with what is written currently. I can imagine you dug deep into this already and I dont have the time to dig into this right now so I would personally remove the commented code and optionally open an issue if someone wants to play with this at a later point. This PR looks great otherwise.

rok-cesnovar (changes requested):

stan/math/opencl/double_d.hpp
t4c1: I thoroughly tested this only in C++ (with the test that compares to quadruple precision). However we need this to work in both C++ and OpenCL. 

As for digging deep, I only checked that compiler does not do something funny and it actually generates the fma instruction (actually it is a call to library function, but I think that should be fine). Than I was out of ideas.

rok-cesnovar: Ok, if you want to take a few moments to look into this more, that is fine with me. But I am also fine with going with the current implementation. I would only remove the commented out code in that case.

rok-cesnovar (approved):
Nice fine with the fma bug :)

PR #2502, opened 2021-06-09
adds testing suite for nested binary var matrix functions

SteveBronder:
## Summary

So @andrjohns in #2461 you were totally correct that I wasn't handling nested array structures correctly for var matrix. This fixes that by

1. adding overloads for `apply_scalar_binary()` for handling `var_value<Matrix>` types
2. adding an overload for `as_array_or_scalar()` that makes a 2d array from `std::vector<std::vector<Scalar>>` types
3. Adding functions to the test suite for testing binary vectorized functions (also with integer containers on the left and right hand side)

With the overload for `as_array_or_scalar()` this was actually pretty darn easy!

## Tests

@andrjohns  can you check `expect_ad_vectorized_matvar` to make sure I'm catching all the specializations I need? I was a bit confused and wasn't totally sure if I caught them all. You can run the functions changed in the PR with

```
./runTests.py test/unit/math/mix/fun/bessel_first_kind_test.cpp \
test/unit/math/mix/fun/bessel_second_kind_test.cpp \
test/unit/math/mix/fun/beta_test.cpp \
test/unit/math/mix/fun/binary_log_loss_test.cpp \
test/unit/math/mix/fun/falling_factorial_test.cpp 
```

## Side Effects

Nope

## Release notes

Adds nested vectorized functions for the new matrix type.

## Checklist

- [x] Math issue #1805 


andrjohns:
Great! I'll have a look on the weekend

stan/math/rev/functor/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,248 @@
...
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/**
+ * Specialisation for use with combinations of
+ * `Eigen::Matrix` and `var_value<Eigen::Matrix>` inputs.
+ * Eigen's binaryExpr framework is used for more efficient indexing of both row-
+ * and column-major inputs  without separate loops.
```
andrjohns, line 18: Can drop this since `binaryExpr` isn't used with the `varmat` types

stan/math/rev/functor/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,248 @@
...
+ * @tparam T2 Type of second argument to which functor is applied.
+ * @tparam F Type of functor to apply.
+ * @param x First Matrix input to which operation is applied.
+ * @param y Second Matrix input to which operation is applied.
+ * @param f functor to apply to Matrix inputs.
+ * @return `var_value<Matrix>` with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
+          require_any_var_matrix_t<T1, T2>* = nullptr,
+          require_all_matrix_t<T1, T2>* = nullptr>
```
andrjohns, line 30: Do these all need to be separate overloads? Given that the only difference between the non-nested calls is the use of `check_matching_dims` and `check_matching_sizes`, could you just use a handful of overloads with a more liberal use of `require_any_`?

stan/math/rev/functor/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,248 @@
...
+ * @tparam F Type of functor to apply.
+ * @param x Scalar input to which operation is applied.
+ * @param y var matrix input to which operation is applied.
+ * @param f functor to apply to Eigen and scalar inputs.
+ * @return var value with inner Eigen type with result of applying functor to
+ * inputs.
+ *
+ * Note: The return expresssion needs to be evaluated, otherwise the captured
+ *         function and scalar fall out of scope.
+ */
```
andrjohns, line 154: Can also cut this comment since `unaryExpr` doesn't get called for `varmat` types

stan/math/rev/functor/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * @tparam T1 Type of std::vector to which functor is applied.
+ * @tparam T2 Type of scalar to which functor is applied.
+ * @tparam F Type of functor to apply.
+ * @param x std::vector input to which operation is applied.
+ * @param y Scalar input to which operation is applied.
+ * @param f functor to apply to inputs.
+ * @return std::vector with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
```
andrjohns, line 176: If you add `is_var_matrix` to the disjunction in `is_container`, then `std::vector<varmat>` will get picked up by the existing overloads in `prim`, and you don't need these.

test/unit/math/mix/fun/beta2_test.cpp
```diff
@@ -0,0 +1,14 @@
+#include <test/unit/math/test_ad.hpp>
```
andrjohns, line 1: Shouldn't this be in the existing `beta_test` file?

test/unit/math/test_ad.hpp
```diff
@@ -1306,15 +1306,16 @@ void expect_ad_vectorized_binary_impl(const ad_tolerances& tols, const F& f,
...
+  expect_ad(tols, f, x, y);            // mat, mat
+  expect_ad(tols, f, x, y[0]);         // mat, scal
+  expect_ad(tols, f, x[0], y);         // scal, mat
+  expect_ad(tols, f, nest_x, nest_y);  // nest<mat>, nest<mat>
+  expect_ad(tols, f, nest_x, y[0]);    // nest<mat>, scal
+  expect_ad(tols, f, x[0], nest_y);    // scal, nest<mat>
+  expect_ad(tols, f, nest_nest_x,
+            nest_nest_y);                 // nest<nest<mat>>, nest<nest<mat>>
+  expect_ad(tols, f, nest_nest_x, y[0]);  // nest<nest<mat>, scal
+  expect_ad(tols, f, x[0], nest_nest_y);  // scal, nest<nest<mat>
```
andrjohns, line 1318: Exceedingly pedantic, but you forgot the closing `>` on the `nest`

test/unit/math/test_ad_matvar.hpp
```diff
@@ -47,6 +49,20 @@ void expect_near_rel_matvar(const std::string& message, T1&& x, T2&& y,
   }
 }
 
+template <typename T1, typename T2,
+          require_all_std_vector_vt<is_std_vector, T1, T2>* = nullptr,
+          require_all_std_vector_st<is_var, T1, T2>* = nullptr>
```
andrjohns, line 54: Should these be `is_var_matrix` to stop it accidentally getting called with `var<double>`?

andrjohns (changes requested):
Review (finally)! The testing all looks very comprehensive, so no issues there. Just a couple of suggestions for simplifying the `apply_scalar_binary` overloads

test/unit/math/mix/fun/beta2_test.cpp
SteveBronder: This is a windows thing. The file becomes too big for mingw if we have one file of tests so we have to break it up into two separate test files

stan/math/rev/functor/apply_scalar_binary.hpp
SteveBronder: Ah yes! So I made a new meta function `is_container_or_var_matrix` that lets us re-use the stuff already in `prim`.

stan/math/rev/functor/apply_scalar_binary.hpp
SteveBronder: This is a weird thing, since we have like `beta(varmat, varmat)` we actually want to directly call beta. I think I tried this before and it didn't work for some reason. It's a seperate issue, but once I have a bunch of these `var<Matrix>` functions done it should be v easy to add versions that also work for `Matrix<var>` (and Matrix<double>)`. Then imo it could be a good idea to limit the functions that use `apply_scalar_*` to only work for std vectors and std vectors with nested containers.

Right now we have some weird meta programs like 

```template <
    typename T, require_not_var_matrix_t<T>* = nullptr,
    require_all_not_nonscalar_prim_or_rev_kernel_expression_t<T>* = nullptr>
inline auto atanh(const T& x) {
  return apply_scalar_unary<atanh_fun, T>::apply(x);
}
```

and I'd like those to just be 

```cpp
template <typename T, require_container_or_scalar_t<T>* = nullptr>
inline auto atanh(const T& x) {
  return apply_scalar_unary<atanh_fun, T>::apply(x);
}
```

Or something of the sort

test/unit/math/test_ad.hpp
SteveBronder: Reasonable!

test/unit/math/test_ad_matvar.hpp
SteveBronder: So we actually want this to be `is_var` in the `st` since the function could be like `f(std::vector<std::vector<var<Matrix>>>, var)`

stan/math/rev/functor/apply_scalar_binary.hpp
SteveBronder: Oh actually I had to have this for things like `f(std::vector<var<Matrix>>, std::vector<var<Matrix>>)`

andrjohns: Yeah but `binaryExpr` only gets used when both inputs are treated as containers of scalars, whereas the `var<Matrix>` are treated as containers in themselves.

So `f(std::vector<var<Matrix>>, std::vector<var<Matrix>>)` evaluates to a loop over the `std::vectors` calling the function on the `var<Matrix>` inputs directly

stan/math/rev/functor/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,160 @@
...
+ * @param f functor to apply to inputs.
+ * @return Eigen object with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
+          require_std_vector_vt<std::is_integral, T1>* = nullptr,
+          require_var_matrix_t<T2>* = nullptr>
+inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+  check_matching_sizes("Binary function", "x", x, "y", y);
+  return f(x, y);
+}
```
andrjohns, line 74:
```suggestion
template <typename T1, typename T2, typename F,
          require_any_std_vector_vt<std::is_integral, T1, T2>* = nullptr,
          require_any_var_matrix_t<T1, T2>* = nullptr>
inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
  check_matching_sizes("Binary function", "x", x, "y", y);
  return f(x, y);
}
```
Then you don't need the separate overloads

stan/math/rev/functor/apply_scalar_binary.hpp
```diff
@@ -0,0 +1,160 @@
...
+ * @param f functor to apply to inputs.
+ * @return Eigen object with result of applying functor to inputs.
+ */
+template <typename T1, typename T2, typename F,
+          require_std_vector_vt<is_std_vector, T1>* = nullptr,
+          require_std_vector_st<std::is_integral, T1>* = nullptr,
+          require_var_matrix_t<T2>* = nullptr>
+inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+  return f(x, y);
+}
```
andrjohns, line 115:
```suggestion
template <typename T1, typename T2, typename F,
          require_all_not_matrix_t<T1, T2>* = nullptr,
          require_all_not_std_vector_vt<std::is_integral, T1, T2>* = nullptr,
          require_any_var_matrix_t<T1, T2>* = nullptr>
inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
  return f(x, y);
}
```

I \*think\* this covers all of the overloads which don't have `dim` or `size` checking

andrjohns (changes requested):

stan/math/rev/functor/apply_scalar_binary.hpp
SteveBronder: I couldn't get this to work :-/. But I was able to cut out half of the ones without dim or size checking by using `require_any_*`

andrjohns (approved):
All looks good!

Issue #2509, opened 2021-06-21
Generalised Hypergeometric (pFq) function


andrjohns:
## Description
One of the pre-requisites for implementing the inverse-CDF (quantile) of the Gamma distribution is the hypergeometric 2F2 function, which is needed for the gradients.

The gradients are the same for the 2F2 and the generalised pFq, so may as well implement the more generalised version

#### Current Version:
v4.1.0

sethaxen:
Was this completed by #2510?

sethaxen:
Also @andrjohns is this the last major prerequisite for implementing the Gamma quantile and its derivative?

nikunj410:
Hi @andrjohns, I am working on the N-mixture model to estimate abundance from count data. The model relies on the hypergeometric_2F2 function. Has/will this function be implemented/exposed in stan?

Thank you

PR #2510, opened 2021-06-21
Generalised Hypergeometric Function

andrjohns:
## Summary

This PR implements the generalised hypergeometric function:

image

Which has signature: `hypergeometric_pFq(vector a, vector b, real z)`

The value itself is calculated by Boosts' `boost::math::hypergeometric_pFq`

The gradients wrt `a` and `b` are given by a Kampé de Fériet function, for example:

image

Which is implemented as a double infinite sum:

image

I've put all calculations on the log scale, as the rising-factorial (Pochhammer) calls get very large very quickly.

## Tests

`fwd`,`rev`, and `mix` tests are added. I've also added tests to check that the gradients are equivalent to those returned by the existing `grad_2F1` function

## Side Effects

N/A

## Release notes

Added the generalised hypergeometric function

## Checklist

- [x] Math issue #2509 


stan/math/prim/fun/grad_pFq.hpp
```diff
@@ -0,0 +1,340 @@
...
+    // where NEGATIVE_INFTY is zero on the log scale
+    T_vec da_infsum = T_vec::Constant(a.size(), NEGATIVE_INFTY);
+    T_vec db_infsum = T_vec::Constant(b.size(), NEGATIVE_INFTY);
+
+    while ((outer_diff > log_precision) && (m < max_steps)) {
+      // Vectors to accumulate outer sum into
+      T_vec da_iter_m = T_vec::Constant(a.size(), NEGATIVE_INFTY);
+      T_vec da_mn = T_vec::Constant(a.size(), NEGATIVE_INFTY);
+      T_vec db_iter_m = T_vec::Constant(b.size(), NEGATIVE_INFTY);
+      T_vec db_mn = T_vec::Constant(b.size(), NEGATIVE_INFTY);
```
SteveBronder, line 94: You can make these outside of the loop and just reset them to the rhs within the loop

stan/math/fwd/fun/hypergeometric_pFq.hpp
```diff
@@ -0,0 +1,219 @@
...
+inline return_type_t<Ta, Tb, Tz> hypergeometric_pFq(const Ta& a, const Tb& b,
+                                                    const Tz& z) {
+  using fvar_t = return_type_t<Ta, Tb, Tz>;
+  using fvar_inner_t = typename fvar_t::Scalar;
+  promote_scalar_t<fvar_inner_t, plain_type_t<Tb>> grad_b(b.size());
+  fvar_inner_t grad_z;
+  grad_pFq_qz(grad_b, grad_z, a, b.val(), z.val());
+  return fvar<fvar_inner_t>(hypergeometric_pFq(a, b.val(), z.val()),
+                            dot_product(b.d(), grad_b) + z.d_ * grad_z);
+}
```
SteveBronder, line 186: So the `grad_pFq_qz(grad_b, grad_z, a, b.val(), z.val());` is a little weird to me. Looking at the impl code it seems like you are doing this based on the `grad_*` plain type. Am I reading this right that you only need the gradients for inputs that are autodiffable? If so then I think this would be a good bit cleaner if we had a scheme where `grad_pFq_qz` did it's deduction for the correct function based on whether the inputs where autodiffable. Then it can return a tuple with size equal to the number of autodiffable values passed into it (and doing `value_of()` before going off to the `grad_pFq_qz_impl` function. 

```cpp
template <typename Ta, typename Tb, typename Tz,
          require_all_eigen_vector_vt<std::is_arithmetic, Ta>* = nullptr,
          require_all_eigen_vector_vt<is_fvar, Tb>* = nullptr,
          require_fvar_t<Tz>* = nullptr>
inline return_type_t<Ta, Tb, Tz> hypergeometric_pFq(const Ta& a, const Tb& b,
                                                    const Tz& z) {
  using fvar_t = return_type_t<Ta, Tb, Tz>;
  using fvar_inner_t = typename fvar_t::Scalar;
  // correct grad_pFq_qz chosen based off of which input is autodiffable
  auto grad_tuple = grad_pFq_qz(a, b, z);
  return fvar<fvar_inner_t>(hypergeometric_pFq(a, b.val(), z.val()),
                            dot_product(b.d(), std::get<0>(grad_tuple)) + z.d_ * std::get<1>(grad_tuple));
}
```

Or if it's easier for the impl you can just always return back a tuple of size 3. I think `grad_pFq` could do something like the below where you just detect which ones you need based on whether the values are constant or not like in the lpdf functions.

```cpp
template <typename Ta, typename Tb, typename Tz,
          require_stan_scalar_t<Tz>* = nullptr,
          require_all_eigen_vector_t<Ta, Tb>* = nullptr>
void grad_pFq(const Ta& a, const Tb& b, const Tz& z,
              double precision = 1e-14, int max_steps = 1e6) {
  std::tuple<promote_scalar_t<partials_type_t<Ta>, Ta>,
    promote_scalar_t<partials_type_t<Tb>, Tb>,
    promote_scalar_t<partials_type_t<Tz>, Tz>> ret_tuple;
  internal::grad_pFq_impl<!is_constant<Ta>, !is_constant<Tb>, !is_constant<Tz>>(
      ret_tuple, a, b, z, precision, max_steps);
  return ret_tuple;
}
```

SteveBronder: ^and set the non-used ones to NaN or something like that

stan/math/rev/fun/hypergeometric_pFq.hpp
```diff
@@ -0,0 +1,242 @@
...
+ * @tparam Tb Type of b argument
+ * @tparam Tz Type of z argument
+ * @param[in] a a
+ * @param[in] b b
+ * @param[in] z z
+ * @return Generalised hypergeometric function
+ */
+template <typename Ta, typename Tb, typename Tz,
+          require_all_eigen_vector_vt<is_var, Ta, Tb>* = nullptr,
+          require_arithmetic_t<Tz>* = nullptr>
```
SteveBronder, line 223: I think these can support the new matrix type if you do a scheme like multiply with 

```suggestion
template <typename Ta, typename Tb, require_all_matrix_t<Ta, Tb>* = nullptr,
          require_all_st_var<Ta, Tb>* = nullptr,
          require_arithmetic_t<Tz>* = nullptr>
```

Or you can most likely put these all into one function with a signature like the below and use a scheme like `multiply` or `fma_reverse_pass` for deducing the arguments that need their adjoints propogated.

```cpp
template <typename Ta, typename Tb, typename Tz, require_all_matrix_t<Ta, Tb>* = nullptr,
          require_return_type_t<is_var, Ta, Tb>* = nullptr>
inline var hypergeometric_pFq(const Ta& a, const Tb& b, const Tz& z) {...}
```

Though I'm fine with leaving them each in individual functions as well

SteveBronder (changes requested):
Took a first pass, reverse mode looks good but had a few Qs about the grad stuff in fwd mode

stan/math/fwd/fun/hypergeometric_pFq.hpp
andrjohns: Oh neat, this is much cleaner. Good call

stan/math/rev/fun/hypergeometric_pFq.hpp
andrjohns: Very good call, this cleaned up the code really nicely

andrjohns:
@SteveBronder that's the last of my updates, should be ready for another look now

betanalpha:
What numerical testing has been done on the performance of this implementation?  Especially in regard to numerical accuracy and speed as a function of the a and b coefficients and z?

These power series solutions are incredibly fragile without lots of work not only to avoid overflow/underflow but also transformation to a form that results in sufficiently rapid convergence to avoid residual errors.  Way back in the day I implemented only 2F1 and not the general hypergeometric function exactly because these numerical fragilities proved to be too problematic outside of the most trivial inputs.  Even that implementation of 2F1 is a bit shaky!

andrjohns:
It's definitely much less efficient than `grad_2F1`.

Using the values:
```
  vector_v a_v(2);
  a_v << 1, 1;
  vector_v b_v(1);
  b_v << 1;
  var z_v = 0.6;
```

The performance difference between `grad_2F1` and `grad_pFq` is very large:
```
Run on (8 X 4000 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 8192 KiB (x1)
Load Average: 0.52, 0.83, 1.00
---------------------------------------------------------
Benchmark               Time             CPU   Iterations
---------------------------------------------------------
grad_pFq_bench     708288 ns       708257 ns         1051
grad_2F1_bench       3979 ns         3949 ns       180484
```

But I'm not aware of any other way of implementing the Kampé de Fériet without resorting to iterating the infinite sum. If it's a choice between slow & available or not available at all, then I'm ok with slow. Also very open to alternative ways to implement this, just let me know

sakrejda:
@andrjohns I've been out of this for a while so just throwing it out there that the performance difference might be from all those calls to the log_factorial functions and friends.  The calculation in F32 and the related grad_* versions builds those incrementally through the iterations of the algorithms instead.  Anyway, I would test this across a whole field of values to see where the errors get out of control since with at least one of the related functions we had to get creative with multiple implementations to get good answers.  :+1: to having a slow-but-correct generic implementation if we can show it's correct and over what ranges of parameters.

andrjohns:
Hmm good point. I'll have a look into optimising the use of `log_rising_factorial` and similar, and do some values testing.

I don't have the math-chops to know how to build this incrementally, but if you know of any references/examples then I can have a look at that too

wds15:
I would say to concentrate first on correctness tests and then tune performance…. The tuning bit can even be done on a second pr, I think (unless you see easy and quick ways to include these now).

sakrejda:
:+1: to correctness first *but* here correctness is also incremental so in the edges of the parameter space is where you need more iterations (so the bound on the number of iterations due to speed of the calculation is important).

@andrjohns if you look at the grad_3F2 and similar functions, I think at this point they *all* use this algorithm and they all do the calculation incrementally.  

Here's an example on the log scale, you have to read down through the loop: https://github.com/stan-dev/math/blob/e22d4664b6e7176db60b08352c76d0489f290147/stan/math/prim/fun/grad_F32.hpp#L69

Here's where the return value is updated: https://github.com/stan-dev/math/blob/e22d4664b6e7176db60b08352c76d0489f290147/stan/math/prim/fun/grad_F32.hpp#L115

I've always wanted to write a templated version of these calculations... :grimacing: 

One way to address the performance issue would be to defer to the specialized functions for the p/q values that those are implemented for... (I haven't been active for a while so I don't know what the math lib team would accept).

stan/math/prim/fun/grad_pFq.hpp
```diff
@@ -0,0 +1,222 @@
...
+    db_infsum.setConstant(NEGATIVE_INFTY);
+
+    while ((outer_diff > log_precision) && (m < max_steps)) {
+      // Vectors to accumulate outer sum into
+      da_iter_m.setConstant(NEGATIVE_INFTY);
+      da_mn.setConstant(NEGATIVE_INFTY);
+      db_iter_m.setConstant(NEGATIVE_INFTY);
+      db_mn.setConstant(NEGATIVE_INFTY);
+
+      double log_phammer_1m = log_rising_factorial(1, m);
```
sakrejda, line 104: Since you're incrementing `m` by 1 at each step and I suggested not re-calculating this every iteration:

Rising factorial:
```
m = 0
rising_factorial = 1
while (m < max_steps) {
  term_m = (x + m);
  rising_factorial = rising_factorial * term_m;
  m++
}
```

log version: 
```
m = 0
log_rising_factorial = 0
while (m < max_steps) {
  term_m = log(x + m);
  log_rising_factorial = log_rising_factorial + term_m;
  m++
}
```

Something like that? You end up tracking the value and it's increment at each step.  Similarr situation with `lgamma`... if you go down this road I can be more helpful just ask and I'll write it as non-pseudocode

sakrejda (commented):
Oops, didn't intend to start a review, just making a code comment.

stan/math/prim/fun/grad_pFq.hpp
andrjohns: This is a great idea! I didn't get it on first reading but now I understand, this should make things much more efficient. Thanks @sakrejda!

andrjohns:
I've implemented the above iterative approach and relaxed the convergence criteria for the inner infinite sum. Current tests still pass and now it's about ~3.5x faster:
```
andrew@HomeLinux:~/math$ ./pFq_testing 
2021-07-02T14:31:13+08:00
Running ./pFq_testing
Run on (16 X 3700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 64 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 8192 KiB (x2)
Load Average: 2.45, 2.33, 2.23
---------------------------------------------------------
Benchmark               Time             CPU   Iterations
---------------------------------------------------------
grad_pFq_bench     193227 ns       193218 ns         3650
grad_2F1_bench       3618 ns         3618 ns       199577
```
Will start looking at values and boundaries to see if any more changes are needed

SteveBronder:
@andrjohns once your finished with the boundaries and other values etc ping me and I'll review! Or is it ready for review now?

andrjohns:
Thanks Steve! Still some more work to do, but I'll let you know

betanalpha:
As @sakrejda mentioned one has to be careful with the definition of performance, and the corresponding testing requirements, when implementing functions like this.  The accuracy of brute force summation varies _strongly_ with the input values, especially x and y.  Any cap on the number of iterations can introduce significant numerical error for these more problematic inputs, and without a cap the computational time can increase so fast that it’s practically no different than an infinite loop (I don’t say this lightly — the first implementation of the incomplete beta functions derivatives would do this often).  To ensure a reasonable implementation these worst case inputs (usually around the boundaries) need to be tested for worst case performance to see if it’s manageable.  I’m not so much interested in comparisons to 2F1 but rather how the computational cost of pFq generally scales with p and q (and how those interact with x/y).

Most math libraries avoid functions like this unless there are useful properties that can transform the series into something that converges faster.  The 2F1 has the z <-> 1 - z symmetry for example, and even nicer special case can take advantage of recursion relations and asymptotic expansion.  pFq generalizes to the point where there aren’t many of these properties left, which makes it extremely difficult to implement efficiently.  I say this only to note that there is a strong case to not consider the inclusion of functions that don’t admit sufficiently efficient implementations.

Also this is a bit more tangential but the use of pFq for CDF gradients and inverses requires a bit a care.  In those applications pFq doesn’t appear by itself but rather in ratio with other nasty functions, and in order to avoid numerical problems the ratio of power series is best solved at the same time so that common scales can be factored out.  This made the incomplete beta gradients order of magnitudes faster and more stable closer to the boundaries of the input space.

sakrejda:
>  I say this only to note that there is a strong case to not consider the inclusion of functions that don’t admit sufficiently efficient implementations.

I do agree with Michael here although I'm comfortable with looking to see if we can define a range of p/q/x/y values that works.  In the past when users bring up examples of the hypergeometric functions failing it's been a hobby for me to improve the region where they point out issues (I've been doing too much applied work to contribute to Stan math otherwise in the last few years). 

Mostly to @andrjohns: Since these functions run fast enough for extensive tests, my approach for the last few was to just run a grid of values and run external comparisons to document that we are calculating what we say we are calculating and that we can expect the function to not hang in a pretty well defined chunk of parameter space. 

My suggestion for this function is to test the full grid of values across a range of p/q and at a minimum have a plot of the absolute and relative errors and run times.  In the past that (especially getting good reference values) has taken the bulk of the time but then we can reasonably say the Stan function does what it's supposed to (in the tested range).

For one of the issues I added a pdf that shows where various implementations fall apart in absolute and relative terms: https://github.com/stan-dev/math/files/955767/d_gamma_p_da.pdf

That's the information that made for some entertaining (?) switching between methods here: https://github.com/stan-dev/math/blob/b51f37ad1166853ed2c27f4bcf28615b5c6068e2/stan/math/prim/fun/grad_reg_lower_inc_gamma.hpp#L132

So tl;dr: I'm willing to put in some time to help so we can include these functions but agree we should show that they calculate what we say they calculate.

andrjohns:
I'm really only implementing this because I need the 2F2, so would it be simpler if I just restricted the inputs to 2F2 and tested across that range? Alternatively, is there a better approach than the double infinite sum for 2F2 gradients (i.e., some reparamaterisation/reduction that I'm not aware of?)

sakrejda:
In that case I would just implement 2F2! What do you need 2F2 for? It might make sense to implement the higher-level function directly if it simplifies the hypergeometric function calculations or makes them more stable..

spinkney:
Can I request 2F1 as well? There's already the gradient.

sakrejda:
@spinkney can you say what your use-case is? Do you need it directly or some larger function? We obv. *can* take this implementation and test it against reference values for p/q 1-5 and potentially we'll just get lucky using the tricks available and make it work while covering a bunch of cases.

spinkney:
> @spinkney can you say what your use-case is? Do you need it directly or some larger function? 
It's possible I don't need it directly, though maybe in the future. Mostly for jacobi and other polynomial groups. Probably better to just wrap boost's implementation of those directly.

sakrejda:
> Mostly for jacobi and other polynomial groups. Probably better to just wrap boost's implementation of those directly.

The pattern for other functions is yes wrap Boost and then often do our own implementation for the gradients and then if we need to we can implement the main function.  But as I've said above, I need somebody to prod me to do some more work on Stan math b/c otherwise I just haven't been doing it... so totally open issues for specific functions to add and I'll contribute where I have some knowledge about the calculations.

andrjohns:
On my side of things I need the 2F2 for the gradients of the gamma quantile function (i.e., `gamma_icdf`). Boost didn't have a `hypergeometric_2F2`, so I figured I would just do the `pFq` case.

But on a positive note, I've just found the Kampe De Feriet in Mathematica's documentation, and they implement it as an infinite sum over the hypergeometric_pFq (page 36 [here](https://functions.wolfram.com/PDF/Notations.pdf)). So I might try that and see how the speed and stability changes

sakrejda:
re: page 36 there: wow, that definition has a *lot* of room for pulling out shared computations over iterations!

andrjohns:
@sakrejda I was trying to implement their parameterisation, but I've gotten a bit confused. In the last section they have a multiplication from 1 to q, but I can't figure out what that q is supposed be. Any ideas? Highlighted below:

image

sakrejda:
I think it's just one of the six parameters of that version of the function... as a first guess: it looks like that line is pulling one term from (A+C)F(P+S)(...., q + m, ...) in the previous step so my quess b/c of the pFq notation, they mean P+S... but I would have to check (I can look at the original later this morning)

sakrejda:
@andrjohns as far as I can tell I was right, the 'q' is from the enclosing sum (so varies with m)

andrjohns:
But if q represents P+S, doesn't that it mean it doesn't vary with m? Or are you referring to something else?

andrjohns:
Just closing this PR for now while I work up the alternate parameterisation and do so more testing

andrjohns:
This is ready for review when anyone has the time

andrjohns:
@sakrejda any chance you would be able to look over this? I'll need this to implement the `gamma_p_inv` and `gamma_q_inv` functions, which are dependencies for a few quantile functions

syclik (approved):
@andrjohns, it looks good to me. Do you need eyes on any particular part of the PR?

Could you please update your branch to the latest `develop` so it kicks off tests? If all is good, we'll merge. Thanks!

spinkney:
@andrjohns I was testing this and the current `grad_2F1` and I think we have a problem with values of `z` between -1 and 0 for both the functions. I don't think we use anything with negative `z` today but we should fix this or not allow `z` to be negative.

Assuming that wolfram alpha is correct https://www.wolframalpha.com/input?i=NumberForm%5BDerivative%5B1%2C+0%2C+0%2C+0%5D%5BHypergeometric2F1%5D%5B3.70975%2C+1.%2C+2.70975%2C+-0.2%5D%2C+15%5D

where I'm testing
```
a_1 = 3.70975
a_2 = 1
b = 2.70975
z = -0.2

// with wolfram showing
// da_1 -0.0488658806159776
// da_2 -0.193844936204681
// db    0.0677809985598383
```
I'm getting
```
// stan grad_2F1 and this pr
da_1 0.0822589
da_2 0.325811
db  -0.114066
```

spinkney:
Oh shoot, I didn't want to add that here. Ah, I'll fix this. However, the fix includes alternating the signs when z is negative. Here it is in a gist https://gist.github.com/spinkney/61718ddd7fea2de48b5d25eb957a289f

spinkney:
Here's a cleaner implementation https://github.com/stan-dev/math/pull/2721/files#diff-d6dd157904e5dcc168d2ed08ad224701bfbd3ed52cc51a1a52c2f9effd8fa69b

jaburgoyne:
> @sakrejda any chance you would be able to look over this? I'll need this to implement the `gamma_p_inv` and `gamma_q_inv` functions, which are dependencies for a few quantile functions

@andrjohns: Is your `gamma_p_inv` and `gamma_q_inv` implementation still in the works? I was about to create an issue in the general Stan repository and then found this pull request. As you say, they are the backbone of quantile functions for gamma and some gamma-derived distributions, and a proper quantile function would allow for (something like) non-centred parameterisations and correlated (via copulae) sampling for gamma parameters

andrjohns:
> > @sakrejda any chance you would be able to look over this? I'll need this to implement the `gamma_p_inv` and `gamma_q_inv` functions, which are dependencies for a few quantile functions
> 
> @andrjohns: Is your `gamma_p_inv` and `gamma_q_inv` implementation still in the works? I was about to create an issue in the general Stan repository and then found this pull request. As you say, they are the backbone of quantile functions for gamma and some gamma-derived distributions, and a proper quantile function would allow for (something like) non-centred parameterisations and correlated (via copulae) sampling for gamma parameters

Yep, I just need to iron out some issues that @spinkney identified, but it's still progressing

andrjohns:
@syclik Now that all has passed, would you mind approving the PR again? Thanks!

andrjohns:
@serban-nicusor-toptal would you be able to restart the Jenkins Threading tests for this PR? Looks like they're a bit stuck at the moment

rok-cesnovar (approved):
:rocket:

As discussed, lets merge this after 2.30 Stan is released. 

Thanks Andrew!

PR #2514, opened 2021-06-23
Force threads=1 when STAN_THREADS is not defined

SteveBronder:
## Summary

This is a fix for the parallel nuts sampler https://github.com/stan-dev/stan/pull/3033#discussion_r651425049. When `STAN_THREADS` is not defined then the number of threads for the tbb should be 1. This just adds a ifdef STAN_THREADS on in the tbb initialization stuff so that if STAN_THREADS is not defined then the number of threads is set to 1.

## Tests

@wds15 should there be new tests for this?

## Side Effects

When STAN_THREADS is not defined TBB is only given 1 thread

## Release notes

Fix tbb initialization so that if STAN_THREADS is not defined then the number of threads is set to 1.

## Checklist

### DIFF
--- a/stan/math/prim/core/init_threadpool_tbb.hpp
+++ b/stan/math/prim/core/init_threadpool_tbb.hpp
@@ -45,6 +45,7 @@ namespace internal {
  */
 inline int get_num_threads() {
   int num_threads = 1;
+#ifdef STAN_THREADS
   const char* env_stan_num_threads = std::getenv("STAN_NUM_THREADS");
   if (env_stan_num_threads != nullptr) {
     try {
@@ -67,6 +68,7 @@ inline int get_num_threads() {
                        "' but it must be a positive number or -1");
     }
   }
+#endif
   return num_threads;
 }
 
@@ -96,7 +98,8 @@ inline int get_num_threads() {
  * and the value of STAN_NUM_THREADS environment variable is invalid.
  */
 inline tbb::task_arena& init_threadpool_tbb(int n_threads = 0) {
-  int tbb_max_threads;
+  int tbb_max_threads = 1;
+#ifdef STAN_THREADS
   if (n_threads == 0) {
     tbb_max_threads = internal::get_num_threads();
   } else if (n_threads > 0) {
@@ -108,6 +111,7 @@ inline tbb::task_arena& init_threadpool_tbb(int n_threads = 0) {
                      "The number of threads is '",
                      "' but it must be positive or -1");
   }
+#endif
   static tbb::global_control tbb_gc(
       tbb::global_control::max_allowed_parallelism, tbb_max_threads);
   static tbb::task_arena tbb_arena(tbb_max_threads, 1);
@@ -139,7 +143,8 @@ inline tbb::task_arena& init_threadpool_tbb(int n_threads = 0) {
  * and the value of STAN_NUM_THREADS environment variable is invalid.
  */
 inline tbb::task_scheduler_init& init_threadpool_tbb(int n_threads = 0) {
-  int tbb_max_threads;
+  int tbb_max_threads = 1;
+#ifdef STAN_THREADS
   if (n_threads == 0) {
     tbb_max_threads = internal::get_num_threads();
   } else if (n_threads > 0) {
@@ -151,6 +156,7 @@ inline tbb::task_scheduler_init& init_threadpool_tbb(int n_threads = 0) {
                      "The number of threads is '",
                      "' but it must be positive or -1");
   }
+#endif
   static tbb::task_scheduler_init tbb_scheduler(tbb_max_threads, 0);
   return tbb_scheduler;
 }
--- a/test/unit/math/prim/core/get_num_threads_test.cpp
+++ b/test/unit/math/prim/core/get_num_threads_test.cpp
@@ -3,6 +3,7 @@
 #include <test/unit/math/prim/functor/utils_threads.hpp>
 #include <gtest/gtest.h>
 
+#ifdef STAN_THREADS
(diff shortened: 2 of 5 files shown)


hsbadr:
@SteveBronder Could we also disable TBB dependency/linking when STAN_THREADS is not defined?

SteveBronder:
I'm not sure, I'd defer to @wds15 on that. Pretty sure the answer is no we cannot

wds15:
@hsbadr Not really... we use the TBB event mechanism to automatically initialize the AD tape resources for any thread which comes to live (including the main thread). It's pretty nasty to code away the TBB dependency for the non-STAN_THREADS case, but it can be done. Things get less nasty if we can require that the C++ user ensures that the AD tape gets initialized before doing anything which requires the AD tape.

wds15:
@SteveBronder how about we get rid if ifdef for STAN_THREADS in places like this:

https://github.com/stan-dev/math/blob/b53432bffd85e345cf9bbb9d9ed805d38692b4ee/stan/math/prim/functor/reduce_sum.hpp#L205

?

I'd think this is now all safe to do given that the TBB is only going with 1 thread always whenever STAN_THREADS is not set.

There is similar code for `map_rect` floating around which can then also be killed. The TBB will simplify to 1-core only code whenever there is only 1 thread at work and we should not bloat our code-base with unnecessary if-defs. There could possibly be a slight degradation in performance (which I doubt), but then people shouldn't use `reduce_sum` or `map_rect` in the first place when they don't plan to go with `STAN_THREADS`.

Thoughts?

SteveBronder:
> I'd think this is now all safe to do given that the TBB is only going with 1 thread always whenever STAN_THREADS is not set.

So we can but it requires making the ChainableStack always thread local. I put it up a version that does this on a separate branch below. imo changing the ChainableStack to be thread local always is a bigger change than the original PR so not sure if it should go here. But it's arguable that having ChainableStack always thread local is actually the correct thing to do and the way we had it before was sort of a bug.

https://github.com/stan-dev/math/commit/c5b023bda9424d13a3346a539d8970358d724f07

wds15:
I don’t think we would need a thread local chainable stack given the chances I outlined…why would we?

It should work just fine.

SteveBronder:
I was getting errors without the chainablestack being thread local when it was calling recover memory nested

SteveBronder:
Yeah so when I make `ChainableStack` not thread local and run the `rev/functor/reduce_sum_test` test I'm getting 

```
 ./runTests.py -j4 ./test/unit/math/rev/functor/reduce_sum_test.cpp 
------------------------------------------------------------
./test/unit/math/rev/functor/reduce_sum_test --gtest_output="xml:./test/unit/math/rev/functor/reduce_sum_test.xml"
[==========] Running 9 tests from 1 test suite.
[----------] 9 tests from StanMathRev_reduce_sum
[ RUN      ] StanMathRev_reduce_sum.no_args
[       OK ] StanMathRev_reduce_sum.no_args (0 ms)
[ RUN      ] StanMathRev_reduce_sum.value
[       OK ] StanMathRev_reduce_sum.value (9 ms)
[ RUN      ] StanMathRev_reduce_sum.gradient
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1144564==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x0000004d9671 bp 0x7fff1bfbc890 sp 0x7fff1bfbc880 T0)
AddressSanitizer:DEADLYSIGNAL
==1144564==The signal is caused by a READ memory access.
==1144564==Hint: this fault was caused by a dereference of a high value address (see register values below).  Dissassemble the provided pc to learn which register was used.
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
terminate called recursively
terminate called after throwing an instance of 'std::logic_error'
terminate called recursively
Aborted (core dumped)
./test/unit/math/rev/functor/reduce_sum_test --gtest_output="xml:./test/unit/math/rev/functor/reduce_sum_test.xml" failed
```

Could this be because of the `ad_tape_observer`?

wds15:
Hmm…. Need to check.

SteveBronder:
@wds15 imo I'd rather merge without the `reduce_sum()` changes and tackle that part later so we can continue on the multi-chain stuff. Unless you have a solution available that I'm not seeing then I'm cool with adding that stuff here

wds15:
Ok…let’s do two prs!

SteveBronder:
Word! In that case is this good to merge?

wds15:
Ok... I looked into it as it bothered me. So here is what is happening: The TBB itself initializes per default to use the maximal concurrency level. When `STAN_THREADS` is now not defined then we **must** call the `stan::math::init_threadpool_tbb()` before doing anything parallel. This is to ensure that the TBB will only run a single task at a time.

Here you see the output of the test whenever `STAN_THREADS` is not defined and I added some more couts:

```
Initializing AD tape!
[==========] Running 9 tests from 1 test suite.
[----------] 9 tests from StanMathRev_reduce_sum
[ RUN      ] StanMathRev_reduce_sum.no_args
[       OK ] StanMathRev_reduce_sum.no_args (0 ms)
[ RUN      ] StanMathRev_reduce_sum.value
[       OK ] StanMathRev_reduce_sum.value (7 ms)
[ RUN      ] StanMathRev_reduce_sum.gradient
TBB max_concurrency: 12
libc++abi: terminating with uncaught exception of type std::logic_error: empty_nested() must be false before calling recover_memory_nested()
```

Anyway... let's stick to 2 PRs. I will finish reviewing this now and then hand you a new PR to implement what I have in mind. Sounds good?

wds15 (approved):
LGTM

Issue #2518, opened 2021-06-28
Allow Math library to link to system-installed libraries/headers


andrjohns:
## Description

Rather than bundling libraries (e.g., Boost/Eigen) with the Math library, it would be good to add functionality to use the libraries that are already installed on the system.

This will be a first step in packaging `cmdstan` binaries for linux distros

#### Current Version:
v4.1.0

wds15:
It would be great if fixes to those libs are still used, even if the system level libs are to be used.

rok-cesnovar:
We currently do no actual fixes in the non-development libraries. For Boost we had the `integrate_1d` patch that was removed. For Sundials we remove prints for CRAN reasons and I dont think this matters for this use case. 

There are a few GTest fixes for MPI, but those are for Math development so also out of scope for Andrew's use case.

andrjohns:
Yep pretty much. This is just for the libraries that an end-user would need to compile/run programs, so things like googletest and cpplint aren't all that important

wds15:
To it's the norm that we need to fix stuff occasionally. I'd much prefer to see this feature support the fix-it thing which we need to do every now and then.

Things like the fixes we do to please CRAN are not needed to take care of, of course.

andrjohns:
That's not really something that can be done in a general fashion, since this just takes the system headers and libraries as-is. If we need to patch particular headers, then logic specific to those headers in those libraries would be needed. Similarly if we needed to rebuild system binaries with patches (e.g., `libtbb.so` or `linsundials_cvodes.a`), then that also needs to be done on a case-by-case basis.

wds15:
I think you can achieve this for the header-only things which we use. For the compiled stuff, this is not possible... that's correct. Good point.

rok-cesnovar:
I am not sure if I understand how do you imagine this.

For example, the last patch for Boost that we had in Math was doing what https://github.com/boostorg/math/pull/455/files does.
Are you saying we should fix the system-installed Boost to include this fix for a particular file? That is not something we should be doing imo. Or are you thinking of something else?

wds15:
No. We are still compiling the things ourselves. So we can control the order of include paths. Hence, we can insert -I flags so that the right boost fix will get into visibility early enough during compilation. This is exactly the same as rstan used to do it. In the setup from rstan we also had to override the R boost install.

rok-cesnovar:
[...]

PR #2519, opened 2021-06-28
Link to system libraries and headers on Linux (Debian)

andrjohns:
## Summary

This PR adds a new make flag: `SYS_LIBS=true`, which allows the Math library to link against system-installed libraries, rather than building them locally.

At the moment this is only tested on Debian/Ubuntu, since I'm working on packaging `cmdstan` into `.deb` files for distribution. Once that packaging process is stable I'll add and test the functionality for the other distros.

This is tested working against the following packages and versions:

- `libboost-math-dev (>=1.74)`
- `libeigen3-dev (>=3.3.9)`
- `libtbb-dev (>=2020.3)`
- `libsundials-dev (>= 4.1)`
- `opencl-headers (>= 3.0~2020)`

This approach is (currently) not compatible with MPI, since the Boost binaries in Ubuntu/Debian aren't built with MPI.

## Tests

Currently N/A, but could be good to add. But I'm not sure how that would interact with the current workflow.

## Side Effects

N/A

## Release notes

Added functionality to link against system libraries on Debian-based Linux distributions

## Checklist

- [x] Math issue #2518


make/compiler_flags
```diff
@@ -232,50 +232,55 @@ endif
...
-TBB_RELATIVE_PATH ?= ../$(notdir $(TBB))
-TBB_BIN_ABSOLUTE_PATH = $(abspath $(TBB_BIN))
-TBB_ABSOLUTE_PATH = $(abspath $(TBB))
-
-ifeq ($(OS),Darwin)
-  TBB_LIBRARIES ?= tbb tbbmalloc tbbmalloc_proxy
-else
-  TBB_LIBRARIES ?= tbb
-endif
+ifndef SYS_LIBS
```
syclik, line 235: I didn't realize this syntax part of make 3.81! Is that the version shipped out with RTools now? (We were hampered by Windows version, which was 3.79 for the longest time.)

syclik: It looks like `ifdef` is used lower in the file, so I guess it's ok!

make/compiler_flags
```diff
@@ -306,9 +311,20 @@ ifdef STAN_MPI
   CXXFLAGS_MPI ?= -Wno-delete-non-virtual-dtor
 endif
 
+################################################################################
+# Setup Sundials
+#
+# Add compiler flags to link to installed versions of Sundials
+
+ifdef SYS_LIBS
```
syclik, line 319: A minor thing: would you mind having the `ifdef` go in the same direction? It'll just make it easier when coming across this code. (some of these are `ifdnef SYS_LIBS` and some are `ifdef SYS_LIBS`)

I'd prefer it to be `ifdef SYS_LIBS`, but that's just a preference.

make/compiler_flags
```diff
@@ -232,50 +232,55 @@ endif
...
-endif
+  endif
 
-ifdef TBB_INTERFACE_NEW
-CPPFLAGS_TBB ?= -DTBB_INTERFACE_NEW
+  ifdef TBB_INTERFACE_NEW
+  CPPFLAGS_TBB ?= -DTBB_INTERFACE_NEW
+  endif
+else
+  CXXFLAGS_TBB ?= -I $(TBB)
```
syclik, line 282: Could this be simplified if at the top, it was writtenas:

```
ifdef SYS_LIBS
    CXXFLAGS_TBB ?= -I $(TBB)
    LDFLAGS_TBB ?= -ltbb
    LDFLAGS_SUNDIALS ?= -lsundials_nvecserial -lsundials_cvodes -lsundials_idas -lsundials_kinsol
endif
```

just in one location? 

One that's set, the other parts shouldn't overwrite the instruction because it's using `?=`. (To be more explicit, you could just use `=` to set it.)

hsbadr:
@andrjohns `TBB_LIB` and `TBB_INC` environment variables allow linking to system (or, optionally, any external) TBB library (#2257). So, either respect them regardless of `SYS_LIBS` or deprecate them.

andrjohns:
Good point. I might change the current approach to optionally allow setting the include and library paths (like `TBB_LIB` and `TBB_INC`) for all libraries, otherwise it defaults to the system installed ones (`/usr/include` and `/usr/lib`).

I need to make some larger changes as well, so I'll close this PR for now so it doesn't clog up the list

hsbadr:
> Good point. I might change the current approach to optionally allow setting the include and library paths (like `TBB_LIB` and `TBB_INC`) for all libraries, otherwise it defaults to the system installed ones (`/usr/include` and `/usr/lib`).

Sounds good. I assume that it will default to building from the internal sources if `SYS_LIBS` is undefined. If `SYS_LIBS=true` and the paths aren't defined, I wouldn't default to `/usr/lib`, just add `-l<library>` to `LDLFAGS` (without path, assuming that the library directory is already in `LD_LIBRARY_PATH`) since some `x68_64` systems (e.g., Ubuntu) has different directory structure -- especially if using `rpath` to designate the run-time search path.

PR #2447, opened 2021-03-24
Update tbb to 2020 3

SteveBronder:
## Summary

This updates the tbb to 2020 3. The only code side change here is that in order to be backwards compatible, but still use the new `task_arena` interface we look at the tbb version major and if it is >= 2020 we define TBB_INTERFACE_NEW and use that, but otherwise we use the old interface

```cpp
#include <tbb/tbb_stddef.h>

#if TBB_VERSION_MAJOR >= 2020
#ifndef TBB_INTERFACE_NEW
#define TBB_INTERFACE_NEW
#endif
#endif
```

And for backwards compatibility I removed the `priority::normal` from the new interfaces task arena since that's the default but `priority::normal` is not defined for the old version of tbb

## Side Effects

Just the macro flag to be aware of

## Release notes

Update tbb to 2020 3

## Checklist

- [ ] Math issue #(issue number)

- [x] Copyright holder: Steve Bronder

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #2208, opened 2020-11-23
adding arm64 architecture to build parameters for macos

nicokist:
## Summary

This commit adds arm64 specific definitions to TBB necessary for building on M1 Macs. It was taken from https://github.com/oneapi-src/oneTBB/pull/294

I've been able to run arm64-compiled Stan models from within a Rosetta2 RStudio/R environment using cmdstanr as follows:

model=cmdstanr::cmdstan_model('test.stan', cpp_options=list(CXX="arch -arch arm64e clang++"))

This should provide a speed up while we wait for the rest of the R ecosystem to update.

## Tests

Compilation fails on ARM Macs when this pull request is not in place.

## Side Effects

None as far as I know.

## Release notes

Compiles on M1 Macs.

## Checklist

- [ ] Math issue #(issue number)
None yet.
- [√] Copyright holder:
@jsbache , https://github.com/oneapi-src/oneTBB/pull/294
- [] the basic tests are passing

These are running now, but this pull request addresses a compilation error.

- [NA] the code is written in idiomatic C++ and changes are documented in the doxygen

- [√] the new changes are tested

PR #2522, opened 2021-06-29
Fix OpenCL Jenkins CPU tests and minor issues

rok-cesnovar:
## Summary

A few minor fixes: 
 - fixes the current issue with OpenCL tests on develop by not running some tests on the GG Linux machine. I would prefer not running them entirely on that machine, but given the lack of test machine with OpenCL CPU support, this is not a realistic option right now. The same tests are run on OpenCL GPU tests so skipping them here is a bit less problematic.
- fixes issue where some blocks had too short fwd passes (at least on Windows machines this caused occasional failures)
- reapplies https://github.com/stan-dev/math/pull/2208 that was lost after https://github.com/stan-dev/math/pull/2447

### DIFF
--- a/test/unit/math/opencl/indexing_rev_test.cpp
+++ b/test/unit/math/opencl/indexing_rev_test.cpp
@@ -1,4 +1,5 @@
 #ifdef STAN_OPENCL
+#ifndef STAN_TEST_SKIP_REQUIRING_OPENCL_INT64_BASE_ATOMIC
 #include <stan/math/opencl/indexing_rev.hpp>
 #include <stan/math/opencl/copy.hpp>
 #include <test/unit/util.hpp>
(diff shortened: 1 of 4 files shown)


rok-cesnovar:
Before this will work we need to run the following on the gelman-group-linux:

```
mkdir -p ~/.config/stan
echo NO_CPU_OPENCL_INT64_BASE_ATOMIC=true>> ~/.config/stan/make.local
```
@serban-nicusor-toptal can you run the two lines on the machine?

test/unit/math/opencl/indexing_rev_test.cpp
```diff
@@ -6,6 +6,7 @@
 #include <gtest/gtest.h>
 #include <algorithm>
 
+#ifndef STAN_TEST_SKIP_REQUIRING_OPENCL_INT64_BASE_ATOMIC
```
t4c1, line 9: Both tests are within the ifdef, so you can also move the includes in.

rok-cesnovar: Done.

serban-nicusor-toptal:
Hey @rok-cesnovar it's done.

SteveBronder (approved):
Cool!

Issue #2111, opened 2020-09-28
Seperating out callbacks from vari


SteveBronder:
## Description

I'm sure Tadej and others have thought of this already but I wanted to write down a neat idea before I forget. I was working on some slides to describe autodiff stuff and realized that if we use `reverse_pass_callback()` everywhere we should be able to separate out the callback's in `var_stack` from the memory we allocate for vari's. This will save some mem, help the compiler with devirtualization, and I think overall give a much cleaner pattern

So we just have to have `callback_vari` look something like 

```cpp
struct callback_base {
  virtual void chain() = 0;
};

template <typename F>
struct callback_vari final : public callback_base {
  F rev_functor_;

  explicit callback_vari(F&& rev_functor)
      : rev_functor_(std::forward<F>(rev_functor)) {
    autodiff_stacks.callback_stack_.push_back(this);
  }
  inline void chain() { rev_functor_(); }
};
```

and in our ChainableStack we replace the `vari_stack_` vector with a `callback_stack_` like

```cpp
    std::vector<callback_base*> callback_stack_;
```

and then grad also get's updated  (nested works fine I was just cutting it out the the example I'm writing)

```cpp
static void grad() {
  size_t end = autodiff_stacks.callback_stack_.size();
  for (size_t i = end; i-- > 0;) {
    autodiff_stacks.callback_stack_[i]->chain();
  }
}
```

This is nice because it makes a vari go from 24 bytes to 16 bytes since we get rid of the pointer to the vtable for `chain()` (and move it over to the `callback_base`. So things like addition become both cleaner and leaner like 

```cpp
inline var operator+(double a, const var& b) {
  if (a == 0.0) {
    return a;
  }
  var ret(a + b.val());
  reverse_pass_callback([a, b, ret](){
    b.adj() += ret.adj();
  });
  return ret;
}
```

full example here with addition 

https://godbolt.org/z/chGdcY

#### Current Version:
v3.3.0

bbbales2:
> This is nice because it makes a vari go from 24 bytes to 16 bytes

That is really nice.

So `vari` s would longer have chains and are just storage for vals and adjoints?

SteveBronder:
Yep!

PR #2524, opened 2021-06-30
[WIP] Remove virtual table from vari

SteveBronder:
## Summary

This removes the virtual functions from vari, essentially writing our own virtual functions as callbacks for `chain()` and `set_adjoint_zero()`.

The idea here comes from Sy Brand's talk [here](https://www.youtube.com/watch?v=8c6BAQcYF_E&t=3023s) and comes down to rolling two separate virtual function tables and using them in the `var_stack_` and `var_nochain_stack_` instead of `vari_base`. Now when a class wants to push a chain method onto the `var_stack_` it has to manually do so.

```cpp
ChainableStack::instance_->var_stack_.push_back(vari_chain(this));
```

Though this is not as bad as it sounds, for instance there are no changes needed for functions that use `callback_var()` or `reverse_pass_callback()` since their underlying classes do the above for you.

The main benefit of this is cache locality and less cache misses. Before, every vari had the following structure

```cpp
vari {
 double val_; // 8 bytes
 double adj_; // 8 bytes
 virtual_function_table* virt; // 8 bytes
} // 24 bytes
```

For a total of 24 bytes, meaning that for a 64 byte cache line we could only fetch a max of 2 vari at a time. The virtual function table had two functions in it, `chain()` and `set_zero_adjoint()`. This is not a great as well since though only one virtual function would be used we would always be pulling in both.

With this PR the vari are now 16 bytes 

```cpp
vari {
 double val_; // 8 bytes
 double adj_; // 8 bytes
} // 16 bytes
```

The function pointers for calling `chain()` and `set_zero_adjoint()` now sit solely in `var_stack_` and `var_nochain_stack_`. For the case of `callback_var()`, where the lambda used for the callback is stored inside of the vari, the data structure will now be 
```cpp
vari {
 double val_; // 8 bytes
 double adj_; // 8 bytes
 Lambda f_; // 8 bytes
} // 24 bytes
```

https://godbolt.org/z/s3r6rEEcd

The benefit here isn't in size, but in locality since the lambda we use for the callback sits next to the values used in the calculations.

Though mixing the vari with and without lambdas still has benefits. For instance an `add()`` using two vari with no lambdas in them produces a memory pattern in our arena allocator like 

```cpp
vari1 {
  double 
  double 
} 16 bytes
vari2 {
  double 
  double 
} 16 bytes
ret_vari {
  double 
  double 
  Lambda
} 24 bytes
```

For 56 bytes total. Two vari with lambdas and one without a lambda also take up 64 bytes, fitting snuggly in a cache line!

I have a benchmark I'm using [here](https://github.com/stan-dev/perf-math/blob/removing-vari-chain/benches/add_test.cpp) to time how much this speeds up the reverse pass when doing a loop over vectors

image

image

Seems good!

## Tests

Since this is a low level change the current tests should pass

## Side Effects

We need to be really really really careful that we add 

```cpp
ChainableStack::instance_->var_stack_.push_back(vari_chain(this));
```
to any class's constructor that needs to be put on the vari callback stack

## Release notes

Removes virtual table from vari

## Checklist

- [x] Math issue #2111


SteveBronder:
Actually running this with the full forward and reverse pass in the benchmark I think this is bad as is. From the graph below it looks like the cost of the forward time is way overpowering the gains from the reverse pass time

image

wds15:
The second graph looks as if the new stuff is slower??

SteveBronder:
Yeah I think I need to go back to the drawing board :-/

PR #2534, opened 2021-07-13
Let check functions take in var<Matrix>

SteveBronder:
## Summary

This let's the error checking functions that can take in matrices also take in `var_value<Matrix>` types. The main change here was just to change the require from `require_eigen_t<>` to `require_matrix_t<>` and to use `value_of()` on the input. I also fixed up `check_less_or_equal` etc. to use SFINAE instead of the struct templating schemes.

## Tests

Several tests were added for var matrix types, though after a bit it started to look like the tests were becoming redundant with the tests that already exist. I added tests for functions which have changes besides calling `value_of()`

## Side Effects

Nope!

## Release notes

Let check functions accept new matrix type.

## Checklist

- [x] Math issue #1805 

### DIFF
--- a/stan/math/fwd/fun.hpp
+++ b/stan/math/fwd/fun.hpp
@@ -5,6 +5,7 @@
 #include <stan/math/fwd/fun/Eigen_NumTraits.hpp>
 
 #include <stan/math/fwd/fun/abs.hpp>
+#include <stan/math/fwd/fun/accumulator.hpp>
 #include <stan/math/fwd/fun/acos.hpp>
 #include <stan/math/fwd/fun/acosh.hpp>
 #include <stan/math/fwd/fun/asin.hpp>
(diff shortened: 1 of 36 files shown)


t4c1:
Do we even need this? I feel like wherever we need to call some check function we are already extracting values, because we need them for the actual implementation.

SteveBronder:
These are called directly by the compiler in some places. I can try to call value_of() in the compiler directly. Though having it here would make the compiler impl simpler

t4c1:
Hmm, I dont know the compiler code, but if it would take significantly less than 1000 lines (size of this PR), than it should be done  in the compiler.

SteveBronder:
I just pushed up something in the compiler, if that checks out then we can close this PR!

SteveBronder:
I think that works!!

Issue #2532, opened 2021-07-12
Allow multivariate distributions to take in matrices


SteveBronder:
## Description

Looking at the [Stan docs](https://mc-stan.org/docs/2_27/functions-reference/multivariate-normal-distribution.html) we only accept arrays of vectors and row vectors for the multivariate distributions. What do folks think about allowing matrices as well?

We could restrict the [current signatures](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/multi_normal_lpdf.hpp#L18) to accept `std::vector<Template>` and then add a version that accepts matrices of size `([N, M], [N, M], [M, M])`.

## Example

This would allow for things in Stan like 

```stan
data {
 int N;
 int M;
 matrix[N, M] Y;
}

parameters {
  matrix[N, M] Mu;
  matrix[M, M] Sigma;
}

model {
 Y ~ multi_normal(Y, Mu, Sigma);
}
```

I would personally find this useful for time series models where `Y` is a matrix of time series and `Mu` is made in transformed parameters and does the VAR component.

## Expected Output

All multivariate distributions support matrix inputs

#### Current Version:
v4.1.0

SteveBronder:
Also having a matrix specialization would allow for some performance speedups

adamhaber:
I'll try! Will submit a draft PR once I have something working.

spinkney:
We just need to be super clear on what index the observations are on. Do we choose rows or do we allow either rows or columns? If we allow both then we need a flag input and that seems to go against the design of other functions. 

It's not a big issue, just that we'll want informative errors in stanc to let the user know that the observation index is assumed to be on the rows of the matrix.

SteveBronder:
I think since it takes in an `array[N] vector[M]` scheme right now the equivalent is `matrix[N, M]`

bob-carpenter:
Sorry I missed this the first time around.  The reason we've never done this before is because it's not clear with @SteveBronder's example whether it's the columns that have a multi-normal distribution or the rows.  

> @mike-lawrence: I think the current scheme supports both "column-wise variates" ... and row-wise variates:

That's right.  

> Would the internals of the function perform faster with one variant vs the other?

Probably not, but if so only a little bit, because Eigen let's you convert w/o copying.

> On the one hand I'd expect the core columnar memory representations would suggest that column-wise variates might perform faster

[...]

PR #2540, opened 2021-07-18
Feature/matrix input to mvn [WIP]

adamhaber:
## Summary

Allow multivariate normal distribution to take in matrices for `y` and `mu` arguments (see #2532)

## Tests

Added a simple initial test for `lpdf` and `rng`, will add others (matching the tests for the non-matrix mvn?) if the current implementation makes sense.

## Side Effects

None.

## Release notes

Allow multivariate normal distribution to take in matrices for `y` and `mu` arguments.

## Checklist

- [ ] Math issue #2532 

- [ ] Copyright holder: Adam Haber


andrjohns:
For the `_lpdf` functions, would it be easier to just call `to_vector` on the inputs? Then the existing vector inputs are unchanged but the matrix inputs are mapped to vectors and the rest of the function proceeds as normal? That might allow for more to be vectorised?

adamhaber:
Do you mean transforming `y` and `mu` from `[N,M]` matrices to `N*M` vectors?

andrjohns:
I was initially, but now I realise that doesn't make much sense in this context, where the y/mu dimensions need to match sigma's. What i was thinking would only really make sense for passing matrices to the univariate distributions

SteveBronder:
Thanks Adam! I made an example of what I think would be nice here you can check out in the diff below

What we want to do is utilize the matrix inputs for more efficient code than the standard mvn. So in the below instead of having to call `trace_inv_quad_form_ldlt` for each row we can just call it once. Would you be able to take something like I have in the below and apply it to the other multivariate normal distributions? Ideally this would be rad to have for all the multivariate distributions but we can do these piece by piece

https://github.com/stan-dev/math/compare/review/mvn-matrix?expand=1

syclik:
Closing this for now since it's still in draft. @adamhaber, please reopen if it's active.

syclik:
Sorry... I didn't realize it wasn't a draft. Reopening.

stan/math/prim/prob/multi_normal_lpdf.hpp
```diff
@@ -102,6 +103,24 @@ return_type_t<T_y, T_loc, T_covar> multi_normal_lpdf(const T_y& y,
...
 
+template <bool propto, typename T_y, typename T_loc, typename T_covar,
+          require_all_not_vector_t<T_y, T_loc>* = nullptr>
+return_type_t<T_y, T_loc, T_covar> multi_normal_lpdf(const T_y& y,
+                                                     const T_loc& mu,
+                                                     const T_covar& Sigma) {
+  using lp_type = return_type_t<T_y, T_loc, T_covar>;
+
+  const size_t N = y.rows();
+  lp_type lp(0.0);
```
syclik, line 114: Could you take away the `using lp_type...` statement on 111 and declare `lp` as that type? It's easier to read when that's together.

stan/math/prim/prob/multi_normal_lpdf.hpp
```diff
@@ -102,6 +103,24 @@ return_type_t<T_y, T_loc, T_covar> multi_normal_lpdf(const T_y& y,
...
+
+  for (size_t n = 0; n < N; ++n) {
+    auto current_y = y.row(n);
```
syclik, line 117: Is this code safe? What happens if `y` and `mu` have different lengths? Won't this just read off the end?

I'd like to see this fixed before it gets into the code base. Perhaps use a `check_matching_sizes` or something similar?

test/unit/math/prim/prob/multi_normal_test.cpp
```diff
@@ -21,6 +21,22 @@ TEST(ProbDistributionsMultiNormal, NotVectorized) {
...
+  Matrix<double, Dynamic, Dynamic> y(2, 3);
+  y << 2.0, -2.0, 11.0, 2.0, -2.0, 11.0;
+  Matrix<double, Dynamic, Dynamic> mu(2, 3);
+  mu << 1.0, -1.0, 3.0, 1.0, -1.0, 3.0;
+  Matrix<double, Dynamic, Dynamic> Sigma(3, 3);
+  Sigma << 9.0, -3.0, 0.0, -3.0, 4.0, 0.0, 0.0, 0.0, 5.0;
+  EXPECT_FLOAT_EQ(-23.47816, stan::math::multi_normal_lpdf(y, mu, Sigma));
+  EXPECT_NO_THROW(stan::math::multi_normal_rng(mu, Sigma, rng));
+}
+
```
syclik, line 39: Please add a test where `y` and `mu` have different number of rows. This will trigger a failure and we should know what that is through a unit test.

syclik (changes requested):
Thanks for the PR!

Please add a unit test for different sizes. I believe it should throw an `std::illegal_argument` exception so that sampling stops.

syclik:
Sorry, it is a draft! I'm closing it until it's ready to be reopened.

PR #2551, opened 2021-07-27
optimization

t4c1:
## Summary

Optimize kernel generator so it can use `matrix_cl`'s move assignment where possible instead of copying the data.

## Tests

Added a test to check that the new optimization is used.

## Side Effects
None.

## Release notes
OpenCL: Optimized kernel generator so it can use `matrix_cl`'s move assignment where possible instead of copying the data.

## Checklist

### DIFF
--- a/stan/math/opencl/kernel_generator/multi_result_kernel.hpp
+++ b/stan/math/opencl/kernel_generator/multi_result_kernel.hpp
@@ -267,7 +267,7 @@ class expressions_cl {
    */
   explicit expressions_cl(T_expressions&&... expressions)
       : expressions_(
-          T_expressions(std::forward<T_expressions>(expressions))...) {}
+            T_expressions(std::forward<T_expressions>(expressions))...) {}
 
  private:
   std::tuple<T_expressions...> expressions_;
@@ -327,8 +327,8 @@ class results_cl {
   void operator=(expressions_cl<T_expressions...>&& exprs) {
     index_apply<sizeof...(T_expressions)>([this, &exprs](auto... Is) mutable {
       assignment_impl(std::tuple_cat(make_assignment_pair(
-          std::get<Is>(results_), std::forward<T_expressions>(
-                                      std::get<Is>(exprs.expressions_)))...));
+          std::get<Is>(results_),
+          std::forward<T_expressions>(std::get<Is>(exprs.expressions_)))...));
     });
   }
 
--- a/stan/math/opencl/kernel_generator/multi_result_kernel.hpp
+++ b/stan/math/opencl/kernel_generator/multi_result_kernel.hpp
@@ -267,7 +267,7 @@ class expressions_cl {
    */
   explicit expressions_cl(T_expressions&&... expressions)
       : expressions_(
-            T_expressions(std::forward<T_expressions>(expressions))...) {}
+          T_expressions(std::forward<T_expressions>(expressions))...) {}
 
  private:
   std::tuple<T_expressions...> expressions_;
@@ -319,6 +319,16 @@ class results_cl {
     index_apply<sizeof...(T_expressions)>([this, &exprs](auto... Is) {
       assignment_impl(std::tuple_cat(make_assignment_pair(
           std::get<Is>(results_), std::get<Is>(exprs.expressions_))...));
+    });
+  }
+  template <typename... T_expressions,
+            typename = std::enable_if_t<sizeof...(T_results)
+                                        == sizeof...(T_expressions)>>
+  void operator=(expressions_cl<T_expressions...>&& exprs) {
+    index_apply<sizeof...(T_expressions)>([this, &exprs](auto... Is) mutable {
+      assignment_impl(std::tuple_cat(make_assignment_pair(
+          std::get<Is>(results_), std::forward<T_expressions>(
+                                      std::get<Is>(exprs.expressions_)))...));
     });
   }
 
@@ -531,11 +541,16 @@ class results_cl {
    * @param expression expression
    * @return a tuple of pair of result and expression
    */
-  template <typename T_result, typename T_expression,
-            require_all_not_t<is_without_output<T_expression>,
-                              conjunction<internal::is_scalar_check<T_result>,
-                                          std::is_arithmetic<std::decay_t<
-                                              T_expression>>>>* = nullptr>
+  template <
+      typename T_result, typename T_expression,
+      require_all_not_t<
+          is_without_output<T_expression>,
+          conjunction<internal::is_scalar_check<T_result>,
+                      std::is_arithmetic<std::decay_t<T_expression>>>,
+          conjunction<
+              is_matrix_cl<T_result>, is_matrix_cl<T_expression>,
+              std::is_same<value_type_t<T_expression>, value_type_t<T_result>>,
+              std::is_rvalue_reference<T_expression&&>>>* = nullptr>
   static auto make_assignment_pair(T_result&& result,
                                    T_expression&& expression) {
     return std::make_tuple(
@@ -574,6 +589,22 @@ class results_cl {
         << result.arg_.a_ << ", but it must be " << result.must_be_ << "!";
       throw std::domain_error(s.str());
     }
+    return std::make_tuple();
+  }
+
+  /**
+   * Optimized move assignment of a `matrix_cl` into another `matrix_cl`.
+   * @param result result - check
+   * @param pass bool scalar
+   * @return an empty tuple
+   */
+  template <typename T_result, typename T_matrix,
+            require_all_matrix_cl_t<T_result, T_matrix>* = nullptr,
+            require_all_st_same<T_result, T_matrix>* = nullptr,
+            require_t<std::is_rvalue_reference<T_matrix&&>>* = nullptr>
+  static std::tuple<> make_assignment_pair(T_result&& result,
+                                           T_matrix&& matrix) {
+    result = std::move(matrix);
     return std::make_tuple();
   }
 };
--- a/test/unit/math/opencl/kernel_generator/multi_result_kernel_test.cpp
+++ b/test/unit/math/opencl/kernel_generator/multi_result_kernel_test.cpp
@@ -131,4 +131,21 @@ TEST(KernelGenerator, multi_result_kernel_reuse_kernel) {
   EXPECT_MATRIX_NEAR(res_diff, correct_diff, 1e-9);
 }
 
+TEST(KernelGenerator, multi_result_kernel_matrix_cl_move) {
+  MatrixXd m1(3, 3);
+  m1 << 1, 2, 3, 4, 5, 6, 7, 8, 9;
+  MatrixXd m2(3, 3);
+  m2 << 10, 100, 1000, 0, -10, -12, 2, 4, 8;
+
+  matrix_cl<double> m1_cl(m1);
+  matrix_cl<double> m2_cl(m2);
+
+  cl_mem m2_buf = m2_cl.buffer()();
+
+  stan::math::results(m1_cl) = stan::math::expressions(m2_cl);
+  EXPECT_NE(m1_cl.buffer()(), m2_buf);
+  stan::math::results(m1_cl) = stan::math::expressions(std::move(m2_cl));
+  EXPECT_EQ(m1_cl.buffer()(), m2_buf);
+}
+
 #endif


stan/math/opencl/kernel_generator/multi_result_kernel.hpp
```diff
@@ -321,6 +321,16 @@ class results_cl {
...
   }
+  template <typename... T_expressions,
+            typename = std::enable_if_t<sizeof...(T_results)
+                                        == sizeof...(T_expressions)>>
+  void operator=(expressions_cl<T_expressions...>&& exprs) {
+    index_apply<sizeof...(T_expressions)>([this, &exprs](auto... Is) mutable {
+      assignment_impl(std::tuple_cat(make_assignment_pair(
+          std::get<Is>(results_),
+          std::forward<T_expressions>(std::get<Is>(exprs.expressions_)))...));
+    });
```
SteveBronder, line 332: If the input is && can't you always move instead of forward? And I think this can be made a lot simpler with `for_each()`

```suggestion
    for_each([](auto& result, auto&& expr) mutable {
      assignment_impl(std::tuple_cat(make_assignment_pair(result, std::move(expr.expressions_))));
    }, this->results_, std::move(exprs));
```

SteveBronder: Also can make assignment pair just return back an std::pair?

t4c1: some make_assignment_pair overloads need to return nothing, which can not be achieved with std::pair. Right now they returns an empty tuple.

t4c1: > If the input is && can't you always move instead of forward?

No. The tuple itself might be a rvalue reference, but the elements in it van still be lvalue references, which must not be moved.

t4c1: I tried to implement this with `for_each()`, but I can't expand the `T_expressions` parameter pack, so I can not call the `std::forward`. This will need to stay as it is (also in this case `for_each()` would not really simplify the code).

SteveBronder: So I got this to work with the below, I think it's doing the same thing? Are the `T_expressions` in `expressions_cl` the same value type in the underlying tuple?

```cpp
    stan::math::for_each([](auto&& result, auto&& expr) mutable {
      using expr_t = decltype(expr);
      assignment_impl(std::tuple_cat(make_assignment_pair(
          result,
          std::forward<expr_t>(expr))));

    }, results_, std::forward<decltype(exprs.expressions_)>(exprs.expressions_));
```

This does throw an error for one of the tests in multi_result_kernel

```cpp
  matrix_cl<double> m1_cl(3, 3);
  matrix_cl<double> m2_cl(3, 3);
  matrix_cl<double> m3_cl(4, 3);
  matrix_cl<double> m4_cl(4, 3);

  matrix_cl<double> res1_cl;
  matrix_cl<double> res2_cl;
  // mismatch in size between different (expression, result) pairs
  /** 
   * This currently throws but with my above it does not
   */
  EXPECT_THROW(stan::math::results(res1_cl, res2_cl)
               = stan::math::expressions(m1_cl - m2_cl, m3_cl - m4_cl),
               std::invalid_argument);
```

Should that test throw though? The result matrix is not filled so we move the resulting rhs into `res*_cl` so I think that's fine?

SteveBronder: If your good with the above I can open up a PR to change the other ones that use index_apply.

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
```diff
@@ -321,6 +321,16 @@ class results_cl {
...
+  }
 
   /**
    * Incrementing \c results_ object by \c expressions_cl object
```
SteveBronder, line 336: Missing docs for the templates. Those are pretty complicated requires so please write out what's going on there

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
```diff
@@ -531,11 +541,16 @@ class results_cl {
...
+  template <
+      typename T_result, typename T_expression,
+      require_all_not_t<
+          is_without_output<T_expression>,
+          conjunction<internal::is_scalar_check<T_result>,
+                      std::is_arithmetic<std::decay_t<T_expression>>>,
+          conjunction<
+              is_matrix_cl<T_result>, is_matrix_cl<T_expression>,
+              std::is_same<value_type_t<T_expression>, value_type_t<T_result>>,
+              std::is_rvalue_reference<T_expression&&>>>* = nullptr>
```
SteveBronder, line 553: If this is strictly for rvalue references why are we forwarding in the below? And why does this one need to be an rvalue reference?

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
```diff
@@ -576,6 +591,22 @@ class results_cl {
...
+   */
+  template <typename T_result, typename T_matrix,
+            require_all_matrix_cl_t<T_result, T_matrix>* = nullptr,
+            require_all_st_same<T_result, T_matrix>* = nullptr,
+            require_t<std::is_rvalue_reference<T_matrix&&>>* = nullptr>
+  static std::tuple<> make_assignment_pair(T_result&& result,
+                                           T_matrix&& matrix) {
+    result = std::move(matrix);
+    return std::make_tuple();
+  }
```
SteveBronder, line 609: Idt I'm following the logic here, what does this return back an empty tuple? Things like this would be good to add to the docs otherwise it's hard to see what's going on without more context

SteveBronder (changes requested):
Mostly doc questions and the for_each thing

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
t4c1: There are two things that can be r- or l-value references. One is tuple (which in this case must be a rvalue reference and than there are tuple elements that can be either.

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
t4c1: This assignment is implemented by calling `matrix_cl` move constructor, so there is nothing to do in the kernel.

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
t4c1: Template is doced on line 339. The require is also relatively simple - just the number of results and expressions must match.

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
SteveBronder: Why would assigning to `res1_cl` throw? It's empty so why can't we move the buffer from the input to `res1_cl`?

SteveBronder: But for that pair, since the rhs is an rvalue doesn't it use the `make_assignment_pair()` that doesn't call a kernel that you wrote in this PR [here](https://github.com/stan-dev/math/pull/2551/files#diff-2b85ab0ef034a31f32d85cad802b90d83b60d7177d38a631bc7d02e18f935fbdR601)? If that test should throw we should check that the result and matrix have the same size before doing the move

t4c1: You are right about this not being implemented within a kernel. The test, however operates on lvalues, so the new `make_assignment_pair()` from this PR is not used and kernel would be used if both operations had the same size. As they do not, it correctly throws.

I agree it not exactly nice to have different error handling behavior between l- and rvalues. However, changing the new (rvalue) case to also throw would be a bit hard, as this check would have to be made before the `make_assignment_pair()` call. That would be a bit more work than I am willing to go into for this PR. Anyway, as it is now in this PR, it works correctly, even if a bit unintuitively.

SteveBronder: > The test, however operates on lvalues, so the new make_assignment_pair() from this PR is not used and kernel would be used if both operations had the same size. As they do not, it correctly throws.

`m1_cl - m2_cl` in the test we are talking about is an rvalue though and we are assigning it to a `matrix_cl<>` that is uninitialized.

```cpp
  EXPECT_THROW(stan::math::results(res1_cl, res2_cl)
               = stan::math::expressions(m1_cl - m2_cl, m3_cl - m4_cl),
               std::invalid_argument);
```

Do we want to always be throwing in the case of incorrect size assignment here? In that case I think it would just be changing the `matrix_cl` rvalue assign operator to check sizes before doing anything. But imo it feels like it makes sense to allow this since the rhs is replacing the buffer of the matrix_cl.

The other test where we have an expression of a certain size on the lhs still throws

```
  // mismatch in size between an expression and a result that can not be resized
  auto block1 = stan::math::block_zero_based(m1_cl, 0, 0, 3, 3);
  EXPECT_THROW(stan::math::results(res1_cl, block1)
               = stan::math::expressions(m3_cl + m4_cl, m3_cl - m4_cl),
               std::invalid_argument);
```

> I agree it not exactly nice to have different error handling behavior between l- and rvalues. However, changing the new (rvalue) case to also throw would be a bit hard, as this check would have to be made before the make_assignment_pair() call.

Is there a reason we can't just check the rows and columns of the assigner and assignee and if they mismatch throw an error?

What's the intuition behind not wanting to allow resizing in the second test case? tmk Eigen only stops resizing in the case of an Eigen Map or fixed size matrices of mismatched sizes.

t4c1: > m1_cl - m2_cl in the test we are talking about is an rvalue though and we are assigning it to a matrix_cl<> that is uninitialized.

Oh, right, the reason for not using the new code in that test is the fact that the right hand side expression is not a plain matrix. Reminder of my previous post is still true. 

> Do we want to always be throwing in the case of incorrect size assignment here?

That is debatable. If we have two assignments of different sizes we can not do them in the same kernel so we throw. If one of these can be handled by matrix_cl move assignment (the new code in this PR) than we don't need to. Although we still could to be consistent with throwing when right hand side is a rvalue and we can not use move assignment.

>Is there a reason we can't just check the rows and columns of the assigner and assignee and if they mismatch throw an error?

We already do that and that is not the issue here. The issue is if we want to assign two expressions in the same call and they have different size (the sizes of results match expressions or can be resized so they do).

>What's the intuition behind not wanting to allow resizing in the second test case?

It is not about resizing. See my previous answer.

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
```diff
@@ -576,6 +591,22 @@ class results_cl {
     }
     return std::make_tuple();
   }
+
+  /**
+   * Optimized move assignment of a `matrix_cl` into another `matrix_cl`.
+   * @param result result - check
```
SteveBronder, line 597: What does `result - check` mean?

stan/math/opencl/kernel_generator/multi_result_kernel.hpp
SteveBronder: >> m1_cl - m2_cl in the test we are talking about is an rvalue though and we are assigning it to a matrix_cl<> that is uninitialized.

> Oh, right, the reason for not using the new code in that test is the fact that the right hand side expression is not a plain matrix. Reminder of my previous post is still true.

Sorry which post? And if what I wrote should not be using the new `make_assignment_pair()` can you add a require to guard against that? Though tbh I'm still confused why it can't use the new one with the move assignment since we are assigning an rvalue expression to the `res1/2_cl`

>> Is there a reason we can't just check the rows and columns of the assigner and assignee and if they mismatch throw an error?

> We already do that and that is not the issue here. The issue is if we want to assign two expressions in the same call and they have different size (the sizes of results match expressions or can be resized so they do).

It sounds like we should have another test that is something like 

```cpp
  auto m1_m2_tmp = m1_cl - m2_cl;
  auto m3_m4_tmp = m1_cl - m2_cl;
  EXPECT_THROW(stan::math::results(res1_cl, res2_cl)
               = stan::math::expressions(m1_m2_tmp, m2_m4_tmp),
               std::invalid_argument);
```

syclik:
As much as I'd prefer not to lose good work, there were outstanding comments that weren't addressed, so I'm closing this.

PR #2549, opened 2021-07-26
[WIP] apply_scalar_unary supports complex types

SteveBronder:
## Summary

This is WIP to have `apply_scalar_unary` support complex types. The tests currently compile but fail the ad testing framework. @bob-carpenter I only had time to set this up but did not have time to look at why the ad test framework was failing. I can look more tmrw into that unless you have time to take a look as well. In `expect_ad_vectorized()` right now I'm just calling  

```cpp
  expect_ad(tols, f, std::complex<double>(x1));
```

Is there anything in particular I need to do for complex wrt to the ad testing framework?

## Tests

`std::complex<>` tests will be added for all of the unary vectorized functions.

## Release notes

Unary vectorized functions support containers of complex numbers

## Checklist

- [x] Math issue #2547 

- [x] Copyright holder: Steve Bronder

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [ ] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [ ] the new changes are tested

PR #2554, opened 2021-08-04
Complex numbers get_real(), get_imag(), to_complex

nickdidio:
## Summary

Adding support for complex scalar functions get_real(), get_imag() and to_complex().
get_real() takes in a complex number and outputs its real component, get_imag() does the same but for the imaginary component. 
to_complex() takes in either 0, 1, or 2 ints or doubles and returns a complex number with its real and imaginary component set to the inputs, and any unfilled components set to 0.

## Tests

Test files are available in test/unit/math/mix/fun and are named get_real_test.hpp, get_imag_test.hpp and to_complex_test.hpp

Tests can be ran with:
./runTests.py test/unit/math/mix/fun/get_real_test.hpp
./runTests.py test/unit/math/mix/fun/get_imag_test.hpp
./runTests.py test/unit/math/mix/fun/to_complex_test.hpp

## Side Effects

None

## Release notes

Added complex scalar functions get_real(), get_imag(), and to_complex.

## Checklist

- [x] Copyright holder: Nicholas DiDio


SteveBronder (dismissed):
cool lgtm! @rok-cesnovar I know you wanted to think about the names here, are you and @bob-carpenter cool with `get_real()` and `get_imag()` imo a little wordy but they're fine

stan/math/prim/fun.hpp
```diff
@@ -270,7 +271,6 @@
 #include <stan/math/prim/fun/read_corr_matrix.hpp>
 #include <stan/math/prim/fun/read_cov_L.hpp>
 #include <stan/math/prim/fun/read_cov_matrix.hpp>
-#include <stan/math/prim/fun/real.hpp>
```
bob-carpenter, line 273: Don't remove `real()` and `imag()`.  We won't use those functions by that name in Stan, but we need it to support programs written against the C++ standard library for our autodiff.  

You can just have `get_real()` and `get_imag()` delegate to `real()` and `imag()` for implementation, but the generic form you have below is probably easier.

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Return a complex type from a real part and an imaginary part.
+ * Default values for both parts is 0.
+ *
+ * @tparam T type of real and
+ * @param re real element
+ * @param im imaginary element
+ * @return Complex type
+ */
+template <typename T = double, typename S = double>
```
bob-carpenter, line 20: This templating is not right if `T` is an `int`.  The return type should always be floating point. 

The proper templating here is:

```c++
#include <stan/math/prim/meta/return_type.hpp>
...
template <typename T, typename S>
inline std::complex<stan::real_return_t<T, S>> to_complex(const T& re = 0, const S& im = 0) {
```

Sorry if the syntax isn't exactly right here.

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * Return a complex type from a integer real part and an imaginary part.
+ *
+ * @tparam T type of real and
+ * @param re real element
+ * @param im imaginary element
+ * @return Complex type
+ *
+ * Used for type casting from int to double for real component.
+ * Otherwise arrays of complex don't work with ints.
+ */
```
bob-carpenter, line 35: The following template shouldn't be necessary given the signature I suggested with promotion.

test/unit/math/mix/fun/to_complex_test.cpp
```diff
@@ -0,0 +1,8 @@
+#include <test/unit/math/test_ad.hpp>
+#include <complex>
+
+TEST(mathMixMatFun, to_complex) {
```
bob-carpenter, line 4: This needs tests for both the nullary and binary calls.  I think there's a corresponding `expect_common_binary` that could be called with a 2-arg lambda.

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * @param im imaginary element
+ * @return Complex type
+ */
+template <typename T = double, typename S = double>
+inline std::complex<T> to_complex(const T& re = 0, const S& im = 0) {
+  return std::complex<T>(re, im);
+}
+
+/**
+ * Return a complex type from a integer real part and an imaginary part.
```
bob-carpenter, line 26: The return is a complex value, not a complex type.   And the arguments are a real part and imaginary part, but both are optional, which should be documented along with default values.  For example,

```
@param[in] re real component (default = 0)
@param[in] im imaginary component (default = 0)
```

You can see I also included the specification that both of these are input arguments (not things set by the program).

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @tparam T type of real and
```
bob-carpenter, line 28: Extra "and" at the end.  And it should say "type of real component" or "type of real part", not just "type of real".

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * @param re real element
+ * @param im imaginary element
+ * @return Complex type
```
bob-carpenter, line 31: The return is "complex value with specified real and imaginary components"

stan/math/prim/fun/get_real.hpp
```diff
@@ -14,8 +14,8 @@ namespace math {
  * @param[in] z argument
```
bob-carpenter, line 14: "argument" is not sufficient doc.  This should say something like "value whose real component is extracted" or something like that.

stan/math/prim/fun/get_real.hpp
```diff
@@ -14,8 +14,8 @@ namespace math {
...
  * @return real part of argument
```
bob-carpenter, line 15: [optional] it's nice to stay consistent in terminology.  I'd suggest choosing "part" or "component" and sticking to it throughout the doc.

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <complex>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return a complex type from a real part and an imaginary part.
+ * Default values for both parts is 0.
+ *
+ * @tparam T type of real and
```
bob-carpenter, line 15: These don't need "and" (though that is how you'd write mathematics).  And it should be "type of real component".  The "type of real" would be `double` or `float`.

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * @param re real element
+ * @param im imaginary element
+ * @return Complex type
```
bob-carpenter, line 18: The return is a value, not a type.

stan/math/prim/fun/to_complex.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Return a complex type from a integer real part and an imaginary part.
+ *
+ * @tparam T type of real and
+ * @param re real element
+ * @param im imaginary element
+ * @return Complex type
+ *
+ * Used for type casting from int to double for real component.
```
bob-carpenter, line 33: Doc should be written actively in terms of what the function does, not how it's used (that can come later if it's not clear).  So the doc for this would be the same as the doc or the previous example, only we shouldn't actually need this second function.

bob-carpenter (changes requested):
Lots of comments given this is your first PR.  It shouldn't be hard to fix and I'm happy to sit down and work through it in more detail with you.

bob-carpenter (dismissed):
This is good to go when it passes tests.

nickdidio:
So the tests for this are going to fail because it relies on Steve's apply-scalar-unary fix at https://github.com/stan-dev/math/pull/2549. I've tested both PR's together and it works. Not sure what to do about this, but it should work as soon as Steve's PR goes through.

bob-carpenter (approved):
Thanks for fixing.

PR #2574, opened 2021-09-02
Add vectorized versions of the free functions

SteveBronder:
## Summary

This is for https://github.com/stan-dev/stanc3/pull/947 and allows the compiler to simply call the free functions for standard vectors holding containers instead of generating a for loop and iterating over them.

All the functions here are pretty much the same with the scheme being

```cpp
auto <constrain>_free(const T& x) {
  std::vector<decltype(<constrain>_free(x[0]))> x_free(x.size());
  std::transform(x.begin(), x.end(), x_free.begin(), [](auto&& xx) {
    return <constrain>_free(xx);
  });
  return x_free;
}
```

## Tests

Tests are added in each `*_constrain_test.cpp` to run the standard test to check input/output but instead uses a standard vector of eigen types. Change tests can be run using 

```bash
./runTests.py -j4 test/unit/math/prim/fun/cholesky_corr_transform_test.cpp \
test/unit/math/prim/fun/cholesky_factor_transform_test.cpp \
test/unit/math/prim/fun/corr_matrix_transform_test.cpp \
test/unit/math/prim/fun/cov_matrix_transform_test.cpp \
test/unit/math/prim/fun/ordered_transform_test.cpp \
test/unit/math/prim/fun/positive_ordered_transform_test.cpp \
test/unit/math/prim/fun/simplex_transform_test.cpp \
test/unit/math/prim/fun/unit_vector_transform_test.cpp

```

## Side Effects

Nope!

## Release notes

Vectorize unconstrain functions

## Checklist

### DIFF
--- a/stan/math/prim/fun/cholesky_corr_free.hpp
+++ b/stan/math/prim/fun/cholesky_corr_free.hpp
@@ -33,6 +33,13 @@ auto cholesky_corr_free(const T& x) {
   return z;
 }
 
+/**
+ * Overload of `cholesky_corr_free()` to untransform each matrix
+ * in a standard vector.
+ * @tparam T A standard vector with with a `value_type` which inherits from
+ *  `Eigen::MatrixBase`.
+ * @param x The standard vector to untransform.
+ */
 template <typename T, require_std_vector_t<T>* = nullptr>
 auto cholesky_corr_free(const T& x) {
   std::vector<decltype(cholesky_corr_free(x[0]))> x_free(x.size());
--- a/stan/math/prim/fun/cholesky_corr_free.hpp
+++ b/stan/math/prim/fun/cholesky_corr_free.hpp
@@ -43,9 +43,8 @@ auto cholesky_corr_free(const T& x) {
 template <typename T, require_std_vector_t<T>* = nullptr>
 auto cholesky_corr_free(const T& x) {
   std::vector<decltype(cholesky_corr_free(x[0]))> x_free(x.size());
-  std::transform(x.begin(), x.end(), x_free.begin(), [](auto&& xx) {
-    return cholesky_corr_free(xx);
-  });
+  std::transform(x.begin(), x.end(), x_free.begin(),
+                 [](auto&& xx) { return cholesky_corr_free(xx); });
   return x_free;
 }
 
--- a/stan/math/prim/fun/cholesky_corr_free.hpp
+++ b/stan/math/prim/fun/cholesky_corr_free.hpp
@@ -42,10 +42,8 @@ auto cholesky_corr_free(const T& x) {
  */
 template <typename T, require_std_vector_t<T>* = nullptr>
 auto cholesky_corr_free(const T& x) {
-  std::vector<decltype(cholesky_corr_free(x[0]))> x_free(x.size());
-  std::transform(x.begin(), x.end(), x_free.begin(),
-                 [](auto&& xx) { return cholesky_corr_free(xx); });
-  return x_free;
+  return apply_vector_unary<T>::apply(
+      x, [](auto&& v) { return cholesky_corr_free(v); });
 }
 
 }  // namespace math
--- a/stan/math/prim/fun/cholesky_corr_free.hpp
+++ b/stan/math/prim/fun/cholesky_corr_free.hpp
@@ -33,6 +33,15 @@ auto cholesky_corr_free(const T& x) {
   return z;
 }
 
+template <typename T, require_std_vector_t<T>* = nullptr>
+auto cholesky_corr_free(const T& x) {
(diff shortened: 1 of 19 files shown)


stan/math/prim/fun/cholesky_corr_free.hpp
```diff
@@ -33,6 +33,22 @@ auto cholesky_corr_free(const T& x) {
...
 
+/**
+ * Overload of `cholesky_corr_free()` to untransform each matrix
+ * in a standard vector.
+ * @tparam T A standard vector with with a `value_type` which inherits from
+ *  `Eigen::MatrixBase`.
+ * @param x The standard vector to untransform.
+ */
+template <typename T, require_std_vector_t<T>* = nullptr>
+auto cholesky_corr_free(const T& x) {
```
bob-carpenter, line 44: Will these fail on an empty standard vectors because `x[0]` is not defined?  We can declare as follows legally and people do this to swap in and out variables (usually with the size being some data object).

```
array[0] simplex[5] a;
```

stan/math/prim/fun/cholesky_factor_free.hpp
```diff
@@ -48,6 +48,22 @@ Eigen::Matrix<value_type_t<T>, Eigen::Dynamic, 1> cholesky_factor_free(
...
 
+/**
+ * Overload of `cholesky_factor_free()` to untransform each matrix
+ * in a standard vector.
+ * @tparam T A standard vector with with a `value_type` which inherits from
+ *  `Eigen::MatrixBase`.
+ * @param x The standard vector to untransform.
+ */
+template <typename T, require_std_vector_t<T>* = nullptr>
+auto cholesky_factor_free(const T& x) {
```
bob-carpenter, line 59: Is there a way to do this the same way we do array applies for vectorized functions so there's not so much cut and paste?   I'm thinking the definitions could look as follows.

```
auto cholesky_factor_free(const T& x) {
  return apply_free([](auto&& xx) {  return cholesky_factor_free(xx); });
}
```

The helper function would then be as follows (it doesn't need to validate the standard vector arg).

```
template <typename T, typename F>
auto apply_free(const T& x, const F& f) {
  std::vector<decltype(f(x[0]))> y(x.size());
  std::transform(x.begin(), x.end(), y.begin(), f);
  return y;
}
```

It'd be even better without having to repeat the requires clause in each function and without the boilerplate lambda abstraction, but I don't see how to get rid of those.

With the generic implementation, we have the type `F` so we can use [`result_of`](https://en.cppreference.com/w/cpp/types/result_of), which would look as follows.

```
std::vector<result_of_t<F, T::value_type>> x_free(x.size());
```

But this needs `F` to be a function type, so it'd only work in the generic version here as far as I can tell.

bob-carpenter (changes requested):
Thanks. This looks good.

I have a question about whether this will work for size 0 vectors as I don't know enough about `decltype` actually works.  Assuming that's OK, this is OK to merge.

I made a suggestion for eliminating boilerplate, but it should be considered optional.  It also contains a suggestion for solving the `x[0]` not existing problem brought up in the first comment.

stan/math/prim/fun/cholesky_factor_free.hpp
SteveBronder: Oh actually I wonder if I can just use the `apply_vector_unary` framework here. Lemme try that

stan/math/prim/fun/cholesky_corr_free.hpp
SteveBronder: This will pass even if `x` is never filled. With your idea below we don't need this directly anymore, but keep in mind the `x[0]` here is done at compile time and only asks for the return type of `x[0]`, which the std::vector<T> knows is `T`.

stan/math/prim/fun/cholesky_factor_free.hpp
SteveBronder: Yes! Just updated this, code uses the `apply_vector_unary<>()` framework and is more generic now. We need something similar for the constraints and I can do the same thing there as well

stan/math/prim/fun/cholesky_corr_free.hpp
bob-carpenter: Nice.  That's what I was asking.

stan/math/prim/fun/cholesky_factor_free.hpp
bob-carpenter: Also nice!   The constraining transforms are trickier because of the side effects.  Not to mention a different PR.

bob-carpenter (approved):
Thanks!  So glad you could get rid of all that boilerplate and use an existing framework.  Feel free to merge if tests pass.

PR #2577, opened 2021-09-08
check generalized inverse for full rank symmetric mat

### LINKED ISSUE #2576: Generalized inverse should work for less than full rank symmetric matrices
## Description
The generalized inverse implementation reverts to the regular inverse when the input matrix is symmetric. However, the regular inverse will result in NaN's when the matrix is less than full rank.

I can add a check using FullPivLU when the matrix is invertible. Interestingly, checking with ldlt `isPositive()` does not work. See https://godbolt.org/z/3fx6o6Gcv. 

What is the preferred way to find if a symmetric matrix is full rank in Stan-math?

## Example
```r
stan_code <- "
functions {
 matrix ginv_stan (matrix A) {
    return generalized_inverse(A);
  }
}
"

A <-matrix(c(4, 4, -2,
             4, 4, -2,
             -2, -2, 10), nrow=3, ncol=3, byrow=TRUE)

ginv_stan(A)

           [,1] [,2]      [,3]
[1,]        Inf -Inf       NaN
[2,]       -Inf  Inf       NaN
[3,] 0.05555556    0 0.1111111

 MASS::ginv(A)
           [,1]       [,2]       [,3]
[1,] 0.06944444 0.06944444 0.02777778
[2,] 0.06944444 0.06944444 0.02777778
[3,] 0.02777778 0.02777778 0.11111111
```
## Expected Output
A matrix without NaNs. Note that the generalized inverse is not unique so I don't expect it to equal the `ginv()` from `MASS`.

spinkney:
## Summary

Fixes #2576. Add a check on symmetric matrices for full rank. Uses the ~FullPivLU~ completeOrthogonalDecomposition to check the rank and returns Eigen's pseudoInverse from the complete ortho-decomp if rank < full rank (otherwise just do the regular inverse). 

## Tests

Added a test for a less than full rank symmetric matrix.

## Side Effects

No.

## Release notes

## Checklist

- [x] Math issue #2576


stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -34,8 +35,11 @@ generalized_inverse(const EigMat& G) {
   if (G.size() == 0)
     return {};
 
-  if (G.rows() == G.cols())
-    return inverse(G);
+  if (G.rows() == G.cols()) {
```
spinkney, line 38: I don't know why `ldlt` gives back success. However, if I do a Cholesky instead it works. This will be much faster 

```c++
 if (G.rows() == G.cols()) {
     LLT<MatrixXd> llt_G = G.llt(); 
    if(llt_G.info() == Success)
      return G.inverse();
  }
```

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -66,8 +69,20 @@ inline auto generalized_inverse(const VarMat& G) {
...
 
-  if (G.rows() == G.cols())
-    return ret_type(inverse(G));
+  if (G.rows() == G.cols()) {
+    arena_t<VarMat> G_arena(G);
+    Eigen::CompleteOrthogonalDecomposition<Eigen::MatrixXd>
+        complete_ortho_decomp_G
+        = G_arena.val().completeOrthogonalDecomposition();
+    if (!(complete_ortho_decomp_G.rank() < G.rows())) {
+      return ret_type(inverse(G));
```
spinkney, line 78: @SteveBronder line 78 fails the expression tests but if I change it to `G_arena` (or `G_arena.val()`) it produces the wrong derivative. 

Do you know how I can get this to pass the expression test?

SteveBronder: Yes the important thing to note here is that once you use `to_ref(G)` or `to_arena(G)` you should not use `G` directly as G can still be an expression and using it twice can cause double evaluation

```suggestion
      return ret_type(inverse(G_arena));
```

I ran this locally and it's passing with `G_arena`

```
./runTests.py -j4 ./test/unit/math/mix/fun/generalized_inverse_test.cpp
```

SteveBronder: My full patch is 
```
diff --git a/stan/math/rev/fun/generalized_inverse.hpp b/stan/math/rev/fun/generalized_inverse.hpp
index bd09b12aa4..f096eef8d1 100644
--- a/stan/math/rev/fun/generalized_inverse.hpp
+++ b/stan/math/rev/fun/generalized_inverse.hpp
@@ -22,9 +22,9 @@ namespace internal {
 inline auto generalized_inverse_lambda(T1& G_arena, T2& inv_G) {
   return [G_arena, inv_G]() mutable {
-    Eigen::VectorXd ones1
+    auto ones1
         = Eigen::VectorXd::Constant(std::min(G_arena.rows(), inv_G.cols()), 1);
-    Eigen::VectorXd ones2
+    auto ones2
         = Eigen::VectorXd::Constant(std::min(inv_G.rows(), G_arena.cols()), 1);
     G_arena.adj()
         += -(inv_G.val_op().transpose() * inv_G.adj_op()
@@ -75,16 +75,14 @@ inline auto generalized_inverse(const VarMat& G) {
         complete_ortho_decomp_G
         = G_arena.val().completeOrthogonalDecomposition();
     if (!(complete_ortho_decomp_G.rank() < G.rows())) {
-      return ret_type(inverse(G));
+      return ret_type(inverse(G_arena));
     } else {
       arena_t<ret_type> inv_G(complete_ortho_decomp_G.pseudoInverse());
       reverse_pass_callback(
           internal::generalized_inverse_lambda(G_arena, inv_G));
       return ret_type(inv_G);
     }
-  }
-
-  if (G.rows() < G.cols()) {
+  } else if (G.rows() < G.cols()) {
     arena_t<VarMat> G_arena(G);
     arena_t<ret_type> inv_G((G_arena.val_op() * G_arena.val_op().transpose())
                                 .ldlt()
```

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -66,8 +69,20 @@ inline auto generalized_inverse(const VarMat& G) {
   if (G.size() == 0)
     return ret_type(G);
 
-  if (G.rows() == G.cols())
-    return ret_type(inverse(G));
+  if (G.rows() == G.cols()) {
```
SteveBronder, line 72: For if's where only one branch should happen it's nicer to use `if else` after the first if

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -21,17 +22,19 @@ namespace internal {
 template <typename T1, typename T2>
 inline auto generalized_inverse_lambda(T1& G_arena, T2& inv_G) {
   return [G_arena, inv_G]() mutable {
+    Eigen::VectorXd ones1
+        = Eigen::VectorXd::Constant(std::min(G_arena.rows(), inv_G.cols()), 1);
+    Eigen::VectorXd ones2
+        = Eigen::VectorXd::Constant(std::min(inv_G.rows(), G_arena.cols()), 1);
```
SteveBronder, line 28: You can make these `auto` so they are just expression here and then evaluated when you use them in the below

test/unit/math/mix/fun/generalized_inverse_test.cpp
```diff
@@ -40,10 +76,24 @@ TEST(mathMixMatFun, ad_tests) {
...
+  using stan::test::expect_ad_matvar;
+
+  auto f = [](const auto& G) { return stan::math::generalized_inverse(G); };
 
   Eigen::MatrixXd z(2, 2);
   z << 1, 2, 5, std::numeric_limits<double>::quiet_NaN();
   EXPECT_NO_THROW(stan::math::generalized_inverse(z));
+}
+
+TEST(mathMixMatFun, ad_tests9) {
```
SteveBronder, line 92: last quick thing, just change these to have like the second name like gen_inv_test_{#} instead of `ad_tests{#}`. Then I think this is good!

stan/math/prim/fun/generalized_inverse.hpp
```diff
@@ -31,15 +31,22 @@ template <typename EigMat, require_eigen_t<EigMat>* = nullptr,
...
-    return inverse(G);
-
-  const auto& G_ref = to_ref(G);
+  if (G_ref.rows() == G_ref.cols()) {
+    Eigen::CompleteOrthogonalDecomposition<
+        Eigen::Matrix<value_type_t<EigMat>, EigMat::RowsAtCompileTime,
+                      EigMat::ColsAtCompileTime>>
+        complete_ortho_decomp_G = G_ref.completeOrthogonalDecomposition();
+    if (!(complete_ortho_decomp_G.rank() < G_ref.rows()))
+      return inverse(G_ref);
```
SteveBronder, line 44: So my only Q is how much this costs / accuracy of computing the `completeOrthogonalDecomposition()` and then using it for the check and then sometimes using it for the pseudoinverse. If someone passes an NxM matrix here i sort of feels like they either

1. Would know ahead of time that they will or will not have a square matrix
2. Know their matrix is low rank and they can't use `inverse()`

Are those not good assumptions? If they are then should we always just do the pseudoinverse()? Is the accuracy of pseudoinverse pretty low compared to inverse()? I almost feel like even if so someone using this would expect a pseudoinverse instead of the inverse

stan/math/rev/fun/generalized_inverse.hpp
```diff
@@ -21,17 +22,19 @@ namespace internal {
...
+    auto ones1
+        = Eigen::VectorXd::Constant(std::min(G_arena.rows(), inv_G.cols()), 1);
+    auto ones2
+        = Eigen::VectorXd::Constant(std::min(inv_G.rows(), G_arena.cols()), 1);
     G_arena.adj()
         += -(inv_G.val_op().transpose() * inv_G.adj_op()
              * inv_G.val_op().transpose())
-           + (-G_arena.val_op() * inv_G.val_op()
-              + Eigen::MatrixXd::Identity(G_arena.rows(), inv_G.cols()))
+           + add_diag(-G_arena.val_op() * inv_G.val_op(), ones1)
```
SteveBronder, line 32: Why did you switch this to `add_diag()`?

SteveBronder (changes requested):

stan/math/rev/fun/generalized_inverse.hpp
spinkney: So that 2 identity matrices aren't created and the matrix ops only have to add to the diagonal. I am just learning Eigen so I'm not sure if they've optimized adding identity matrices or how creating an identity matrix affects memory.

stan/math/prim/fun/generalized_inverse.hpp
spinkney: These are great questions!

On
> If someone passes an NxM matrix here i sort of feels like they either
> 
>   1. Would know ahead of time that they will or will not have a square matrix
>   2. Know their matrix is low rank and they can't use `inverse()`

I'm interested in the case where you'd specifically pass a square matrix but not know ahead of time if the matrix is low rank or not. It's probably low-rank but it may not be (say in a factor model).

This got me to revisit the paper that the code is based on. The author does something that I missed because he notates `L` as a cholesky factor when it is actually a modified cholesky factor that works even for low rank matrices. This is how he is able to not do a rank find (which one usually does either as I've done here or using SVD). I've coded up the prim version at https://github.com/spinkney/math/tree/generalized_inverse_low_rank_chol. The point is that some rank issue must be dealt with or there's numerical instability issues.

The code now follows the [paper](https://arxiv.org/ftp/arxiv/papers/0804/0804.4809.pdf) explicitly. The major change involves adding a ` cholesky_low_rank_decomposition` lambda. I've coded it using Eigen block notation. I believe there's further optimizations that could be done, if someone with a background in high performance computing helps. It could be added to Stan as a separate function, which is cool in it's own right. 

The prim tests pass. I haven't benchmarked though to see how it compares to the `completeOrthogonalDecomposition()` version here.

spinkney: I've put together an example how low rank chol can transform a low rank matrix of "factors" into a valid correlation matrix. I'm not sure if this works all the time but it's promising. [low_rank_chol_eg.R.zip](https://github.com/stan-dev/math/files/7147756/low_rank_chol_eg.R.zip)

stan/math/rev/fun/generalized_inverse.hpp
SteveBronder: Eigen is clever enough to know that when you give it an `Eigen::MatrixXd::Identity` it will most likely just use a single scalar to denote the whole matrix (I can't remember what exactly they do but they are clever about it)

stan/math/prim/fun/generalized_inverse.hpp
SteveBronder: Oh nice! So with the low rank chol version should we close this PR and open one up based on the low rank chol? Or should we merge this then make a separate PR for that?

SteveBronder (dismissed):
lgtm!

stan/math/rev/fun/generalized_inverse.hpp
spinkney: Should I change it back to identity? I noticed some speedup in the tests but that could just be an artifact

stan/math/prim/fun/generalized_inverse.hpp
spinkney: I'm thinking that the low rank chol should be the final version. Though I think it may be worth just getting this update in for the next release and then doing 2 PRs: 1) add low rank chol as a function and 2) update gen inverse using it in a new PR.

stan/math/rev/fun/generalized_inverse.hpp
SteveBronder: I would think that's an artifact I think you can change it back

SteveBronder (approved):
lgtm!

spinkney:
@rok-cesnovar @SteveBronder this passed before and that commit shouldn't have failed. I'm not sure what that strange CI error is, there's no info, is this a server issue?

rok-cesnovar:
Yeah seems like a machine went down. Restarted only the upstream tests, should be done in an hour: https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2577/26/pipeline

spinkney:
@SteveBronder this is good to go!

rok-cesnovar:
@spinkney go ahead and merge. You should be able to.

Issue #2474, opened 2021-04-21
Update to Eigen 3.4


SteveBronder:
The eigen 3.4-rc is out and we should run our test suite with it to help them catch any bugs.

Change log below, seems nice!

https://eigen.tuxfamily.org/index.php?title=3.4

#### Current Version:
v4.0.1

PR #2583, opened 2021-09-17
[WIP] Update to Eigen 3.4

SteveBronder:
## Summary

Updates Eigen to 3.4-rc1 to test and report any issues downstream.

## Tests

No new tests

## Side Effects

See the release page for the updates to 3.4

https://eigen.tuxfamily.org/index.php?title=3.4

## Release notes

None

## Checklist

- [x] Math issue #2474 


SteveBronder:
@rok-cesnovar can we bump up to just testing on Rtools 4.0? tmk the old-rel of Cran now uses version 4.0 for checking so that seems fine.

SteveBronder:
@rok-cesnovar do you have an idea of what the scheme would be to update our testing to Rtools40? I tried changing the download link just to the Rtools 40 download in the PR ([here](https://github.com/stan-dev/math/pull/2583/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R31)) though that seems to just keep downloading?

rok-cesnovar:
Oh sorry, I missed the message last time....

It turns out I was complicating things here unnecessarily. Never went back to simplify it though.
Just add the setup-r action that will instal the release version of R (which also installs Rtools 4.0).

```yaml
- uses: r-lib/actions/setup-r@v1
   with:
     r-version: 'release'
```

and then add the paths with:

```yaml
- name: Set path for RTools 4.0
  if: runner.os == 'Windows'
  run: echo "C:/rtools40/usr/bin;C:/rtools40/mingw64/bin" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
```

SteveBronder:
Awesome thank you!

SteveBronder:
@bgoodri these tests should hopefully pass and it should be ready to look over. The stan code related changes are in https://github.com/stan-dev/math/pull/2583/commits/e1233fbb2f67afef72b56c7d2b7ad63e7f9d2f68 though reading it over I think everything should be pretty compatable for 2.26?

I couldn't figure it out but the specialization for fwd mode's `mdivide_right()` is failing so I just commented out the code for now (though we should just delete it before we merge if we are going to delete it). If you want to take a look at those you can uncomment the code in `stan/math/fwd/fun/mdivide_right.hpp` and run the following test with `CXXFLAGS+=-DSTAN_TEST_PRINT_MATRIX_FAILURE` in your make/local to see the failing hessian and hessian gradient. 

```
./runTests.py -j24 test/unit/math/ -f mix/fun/mdivide_right_test
```

serban-nicusor-toptal:
This Jenkins error is on me @SteveBronder looks like open mpi updated not properly last night when I've added clang-format.
Will rebuild the image and restart the job sorry for the inconvenience !

andrjohns:
@serban-nicusor-toptal would you be able to have a look at the CI for this PR when you have a sec? It looks like the `gelman-group-linux` machine is having issues

serban-nicusor-toptal:
Checking

andrjohns:
It looks like the current 3.4 release has a bug with handling `-inf` arguments to exp, which has since been resolved in the development branch.

Using the testing code:
```cpp
#include <Eigen/Core>
#include <limits>
#include <iostream>

int main() {
    double y = -std::numeric_limits<double>::infinity();
    Eigen::VectorXd y_vec(2);
    y_vec << y, y;

    std::cout << y_vec.array().exp() << std::endl;
}
```
The results under 3.4 ([godbolt](https://godbolt.org/z/3xqz84f43)):
```
Program returned: 0
5.55553e-309
5.55553e-309
``` 

And under the current `trunk` ([godbolt](https://godbolt.org/z/M5zWTW6z9)):
```
Program returned: 0
0
0
```

It looks to upgrade to 3.4 we'll have to disable the use of Eigen's SIMD-vectorised `exp` until the following release of Eigen

bgoodri:
Is 5e-309 really a problem?

andrjohns:
It's what's causing the current distribution test failures, since the log cdf of the gumbel with `Inf` inputs is no longer zero.

So we could add a tolerance to the test if its not much of an issue

bgoodri:
I think temporarily changing the unit test would be preferable to disabling
SIMD for something that is correct for all numerical intents and purposes.

andrjohns:
Sounds good to me

SteveBronder:
Just an fyi when this passes I think what we need to do is rebase this to have the eigen 3.4 include in one commit and the actual stan changes in another commit so it's a bit easier to look over. Or alt we could have one PR just adding Eigen 3.4, one PR to switch to 3.4, and another to delete 3.39. Though I think the rebasing thing will not be too bad

andrjohns:
Yeah agreed. Probably easiest to do the Stan Math changes first, so we can make sure they're backwards compatible-ish before the Eigen changes go in

hsbadr:
FYI. I've tested this successfully with the experimental version RStan.

PS: It conflicts with #2641.

andrjohns:
> FYI. I've tested this successfully with the experimental version RStan.
> 
> PS: It conflicts with #2641.

Great news! Yeah both PRs partially overlap in the same changes, so I'll just resolve the conflicts with whichever is merged first

SteveBronder:
Excellent this is working! @bgoodri actually instead of breaking up the commits how do you feel about using the view github gives us below? Along with the lhs tab dropdown I was able to look over the changes pretty easily. The big thing imo is making sure `mdivide_left` and right are good

https://github.com/stan-dev/math/pull/2583/files?file-filters%5B%5D=.hpp&show-viewed-files=true

andrjohns:
I've narrowed down the `mdivide` failures to something related to `vector` vs `row_vector` differences, since the tests all pass if `vectors` are used instead

SteveBronder:
I can take a look at this next week again to try to sort that out. Yes I noticed something about the vector thing being weird as well. I can take a crack again at this next week

roualdes:
Thanks, y'all, for getting this going.  In the last week, I've missed out on two new features that Eigen 3.4 brings.

andrjohns:
I've traced the `mdivide` failures to the `finite_diff_grad_hessian_auto` function in the autodiff testing framework. The gradients calculated by the autodiff framework are consistent between a `row_vector(2)` and a `matrix (1, 2)` inputs, but they differ when calculated by finite-differencing.

I'll keep chasing this down and update with any findings. Almost there!

andrjohns:
@SteveBronder I've narrowed things down to the `hessian` function, but I can't figure out why the behaviour is different.

I've put together a minimal reproduction of the part of the testing framework where the discrepancy occurs, would you be able to have a look when you get a minute?

```cpp
#include <test/unit/math/test_ad.hpp>
#include <vector>
#include <iostream>

template <typename T>
auto calc_hessian(const T& x1) {
   auto f = [](const auto& x, const auto& y) {
    return stan::math::mdivide_right(x, y);
  };

  Eigen::MatrixXd x2(2, 2);
  x2 << 1, 0, 0, 1;

  auto g = [&](const auto& v) {
    auto ds = stan::test::to_deserializer(v);
    auto x1ds = ds.read(x1);
    auto x2ds = ds.read(x2);
    return stan::test::serialize_return(stan::test::internal::eval(f(x1ds, x2ds)))[1];
  };

  // internal::expect_ad_helper(tols, f, g12, serialize_args(x1, x2), x1, x2);
  auto x = stan::test::serialize_args(x1, x2);

  // test_grad_hessian(tols, g, x, gx);

  double fx = 0;
  int d = x.size();
  std::vector<Eigen::MatrixXd> grad_hess_fx(d);
  Eigen::VectorXd x_temp(x);
  Eigen::VectorXd grad_auto(d);
  Eigen::MatrixXd hess_auto(d, d);

  x_temp(5) = x(5) + 2 * stan::math::finite_diff_stepsize(x(5));
  stan::math::hessian(g, x_temp, fx, grad_auto, hess_auto);

  return hess_auto;
};

TEST(MathFunctions, abs) {
  Eigen::MatrixXd mat_x1(1, 2);
  mat_x1 << 1, 1;

  Eigen::RowVectorXd rv_x1(2);
  rv_x1 << 1, 1;

  Eigen::MatrixXd mat_hessian = calc_hessian(mat_x1);
  Eigen::MatrixXd rv_hessian = calc_hessian(rv_x1);

  EXPECT_MATRIX_EQ(mat_hessian, rv_hessian);
}
```

hsbadr:
@andrjohns After you've merged `develop` in 02b85557330c7d20382038e43ccc0f9d3061c5db, this branch now  has the two versions of `eigen` and builds v3.3.9.
https://github.com/stan-dev/math/blob/24f3f763b4f05adcaf8cca65b24de6c6bb5da2fa/make/libraries#L12

andrjohns:
Ah good catch, thanks. I knew that merge was too painless to be true

SteveBronder:
@andrjohns ty for the minimal example! Yes I will try to look at this on Monday

andrjohns:
@SteveBronder this is now passing all tests (including `mdivide_`), but the `2pl_irt` model from the performance tests is failing. When I run locally it looks like the values are consistent between `develop` and this PR, but I'm not sure if I'm doing something wrong. Would you be able to run the test on your machine when you get a minute? Thanks!

SteveBronder:
Hmm, it seems to be passing locally for me as well. I just restarted it and updated the branch with develop, lets let this run and if it bonks I'll try to replicate the environment jenkins using with a docker container.

Looking at the [jenkins log](https://jenkins.flatironinstitute.org/blue/rest/organizations/jenkins/pipelines/Stan/pipelines/CmdStan%20Performance%20Tests/branches/downstream_tests/runs/161/nodes/65/steps/77/log/?start=0)  I'm seeing the error 

```
FAIL: golds/stat_comp_benchmarks_benchmarks_irt_2pl_irt_2pl.gold param sigma_theta not within (1.062132876 - 1.010467953) / 0.162185019873786 < 0.3
```

Where `1.06` is this PR and `1.01` is develop. That calculation above is equal to `0.3185~`. So it's very close to passing on jenkins.

 Looking at the parameters used to generate the data [here](https://github.com/stan-dev/stat_comp_benchmarks/blob/fcf97113da2f7d9a0356a59f16d5381131a21189/benchmarks/irt_2pl/irt_2pl.params) it looks like value for `sigma_theta` is 

```
sigma_theta 1.0142e+00 1.7369e-01 (mean and sd)
```

So something odd did happen, but it might just be a one off thing. If it happens again I'll look more into it

SteveBronder:
Ugh thats so odd @andrjohns what do you think the chances are this is an Eigen thing? Looking at the Stan code for [irt_2pl](https://github.com/stan-dev/stat_comp_benchmarks/blob/fcf97113da2f7d9a0356a59f16d5381131a21189/benchmarks/irt_2pl/irt_2pl.stan) there just doesn't seem to be much going on in that function beyond lower/upper and the distributions.

SteveBronder:
Throwing a hail mary, updating to the latest version of 3.4 which has a few bug fixes. 

Another thing I noticed is that locally I tend to get a value of 1.03 for that sigma parameter on develop for that model. I do still get a value of 1.06 with Eigen 3.4 but the fail does not go off because they are close enough. So for some reason the jenkins is running more accurately?

andrjohns:
The only thing that I can think of is that the cmdstan performance tests on Jenkins are run with the `-march=native` flag, so we might be seeing some small numeric differences due to changes in Eigen's SSE/AVX code?

SteveBronder:
Hmm I thought it could be an Eigen thing but messing around with the below idt exp() is the problem (they changed something about this in master but not 3.4 and I thought it could be the problem)

https://godbolt.org/z/cnc7efsqx

> The only thing that I can think of is that the cmdstan performance tests on Jenkins are run with the -march=native flag, so we might be seeing some small numeric differences due to changes in Eigen's SSE/AVX code?

Yeah if you look at the ASM diff there's a lot of changes. Though tbh I'm not sure where to go from here? Locally I'm getting a value of 1.04 for that problem parameter on develop. So that seems to be okay with me imo (also the result is within .31 on jenkins and the band is .3 so it is very close). @serban-nicusor-toptal do you have the specs for the box the performance tests run on? 

I also want to look at the generated C++ code for that model. Maybe we are doing something silly with an expression somewhere? (though I would expect a much wronger result if that was the case)

serban-nicusor-toptal:
Hey, this runs inside a [Docker ](https://github.com/stan-dev/ci-scripts/blob/master/docker/ci/Dockerfile-gpu) container.
On a host machine with `CentOS Linux release 7.9.2009`, `510GB` of RAM and `x56 CPU (Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz)` which is splitted into x10 Jenkins Executors (~50 GB RAM, ~11 CPUs) on which the Docker container runs.
There's also x20 more executors that have ~5 CPUs and ~20GB of ram. ( same OS, same CPU I think, don't have access to the second host)
Hope that helps, if there's anything else I can do to help with this please let me know!

hsbadr:
> Locally I'm getting a value of 1.04 for that problem parameter on develop.

Are you using the same compiler version and flags locally? Specifically, have you turned off compiler optimizations (i.e., using `-O0` flag or `echo O=0 > make/local`).

syclik:
Looks like this is still active. What will take this from WIP to ready?

andrjohns:
I've \*finally\* managed to reproduce the `irt_2pl` performance test failure locally, so I'll start digging into this so we can hopefully get Stan 3.4 ready soon

SteveBronder:
Excellent! What did you do to have it reproduce locally?

andrjohns:
As we suspected, the test failure is coming from the use of `-march=native` in the Makeflags in the [git comparison script](https://github.com/stan-dev/performance-tests-cmdstan/blob/a5135905488daef80e1d0fe9778f83f321e8b796/compare-git-hashes.sh#L10). 

Specifically the use of AVX2 extensions, as the tests all pass if I change the makeflags to:
```
CXXFLAGS += -mtune=native -mmmx -msse -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mavx"
```

I think we might mark this as an expected failure, since the Eigen release notes indicate [substantial changes to the SIMD code](http://eigen.tuxfamily.org/index.php?title=3.4#Elementwise_math_functions)

Should we remove `-march=native` from this test?

andrjohns:
@serban-nicusor-toptal would it be possible to create a branch of the `performance-tests-cmdstan` repo which doesn't add `-march=native` to the Makeflags in the `compare-git-hashes.sh` script, and test this PR against that branch? Just to double-check that my local results aren't a fluke

serban-nicusor-toptal:
I think that might work, will try to set it up tomorrow and post the results.

serban-nicusor-toptal:
I've created this [a new Jenkins job](https://jenkins.flatironinstitute.org/job/Stan/job/CmdStan%20Performance%20Tests/job/downstream_tests_eigen34/) pointing to eigen3.4-test branch in [perf tests](https://github.com/stan-dev/performance-tests-cmdstan/tree/eigen3.4-test).
Now that allows us to run it with parameters, in which we give this branch as an input.
I already started a job here: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FCmdStan%20Performance%20Tests/detail/downstream_tests_eigen34/3/pipeline

Please let me know if there's anything else I can help with!

andrjohns:
I've narrowed this down to the vectorisation of the `bernoulli_logit`, since the tests pass with looping but not vectorised inputs. Almost there!

roualdes:
This one has been pesky indeed.  So just cheerleading here.  Thanks y'all for your hard work on this.  You're doing great!

andrjohns:
Breakthrough!

I iteratively tested Eigen's commits since the 3.3.9 release, and found the culprit: https://gitlab.com/libeigen/eigen/-/commit/25d8ae7465e6430bc2dc7f65800332932d3bb774

The commit simplified the vectorised calculation of `log` and `log2`, which is why the test failures here were only appearing under AVX2. If I run Eigen 3.4.0 and revert these changes, then the test passes.

I still need to grok the construction and changes here and do some more testing to find the input values that are giving different results between 3.3.9 and 3.4.0, then we'll know whether this is a bug to patched in Eigen or just an improvement in numerical accuracy/stability

WardBrian:
Really great detective work @andrjohns!

andrjohns:
Downstream tests have all passed! I'm still having trouble creating a reproducible example, the only place I'm seeing differences is in the `irt_2pl` benchmark. My next step is to try iteratively reverting parts of the problematic commit to see whether it was a particular line/change

@SteveBronder if you've got any tricks/suggestions for other angles to investigate in figuring out a reproducible example, would be much appreciated!

SteveBronder:
@andrjohns great find!

So I messed around with Eigen 3.4 just comparing it to std log and I think maybe the error here is when we sum a bunch of log values? My thinking here is that for a single log then that ulp is not a big deal, but for a lot of log calls summed together the errors don't cancel each other out and you end up with a bigger error than 1 ULP. You can see the difference in the sum in the godbolt below. 

https://godbolt.org/z/Gr34vMbGn

It's interesting to note as the values get bigger the sum differs more and more. This made me think of `log_sum_exp` and now I'm wondering if we should be calling `log_sum_exp` in a lot of our probability functions? At least anywhere we would normally just do a basic sum. @bob-carpenter @syclik any thoughts on this?

I thought about `log_sum_exp()` also because in the [irt](https://github.com/stan-dev/stat_comp_benchmarks/blob/fcf97113da2f7d9a0356a59f16d5381131a21189/benchmarks/irt_2pl/irt_2pl.stan) model I see it calls [`bernoulli_logit`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/bernoulli_logit_lpmf.hpp#L74) which has a line calling normal sum. But should this call `log_sum_exp()`? Is that unnecessary for some reason?

```cpp
  T_partials_return logp = sum(
      (ntheta > cutoff)
          .select(-exp_m_ntheta,
                  (ntheta < -cutoff).select(ntheta, -log1p(exp_m_ntheta))));
```

@andrjohns do you mind if I change that line to call `log_sum_exp()` to see if that fixes things? I think Either way we should ping the Eigen team, either with an issue or on discourse about this

SteveBronder:
So changing this line so that the function compiles fails the unit tests because for the functions taking in vectors we are doing `log_sum_exp()` but this doesn't work when comparing against calling the function multiple times.

```cpp
  T_partials_return logp = log_sum_exp(
      (ntheta > cutoff)
          .select(-exp_m_ntheta,
                  (ntheta < -cutoff).select(ntheta, -log1p(exp_m_ntheta))).matrix().eval());
```

Example failure: 

```
[ RUN      ] AgradDistributionsBernoulliLogistic_v_5/AgradDistributionTestFixture/0.RepeatAsVector
The difference between x1 and x2 is 3.8712010109078916, which exceeds tol_val, where
x1 evaluates to -4.1588830833596724,
x2 evaluates to -0.28768207245178079, and
tol_val evaluates to 2.2232825779057268e-08.
expect_near_rel_finite in: log prob with repeated vector input should match a multiple of log prob of single input
```

SteveBronder:
I think maybe pinging the Eigen team about this would be the right thing to do at this point

bob-carpenter:
> I'm wondering if we should be calling log_sum_exp in a lot of our probability functions? 

Anywhere we need to calculate `log(sum(exp(x))` we should be using `log_sum_exp`.  But it's not going to help if we just need to calculate `sum(log(x))`.

SteveBronder:
> I just checked the godbolt and I'm not sure if that's actually the issue here, since it looks like the results/difference are the same for 3.3.9 and 3.4.0?

Oh shoot yeah your right so scratch the sum thing. Below is an updated godbolt that shows that's easier to read the sums from each (3.4 on executor 1 and 3.9.9 on executor 2)

https://godbolt.org/z/E8rjxGMj6

The max diff changes but the sums have the same relative error.

I'm not to sure how to progress at this point besides hitting up the eigen team. I can look into the different functions there to see how log is called. Maybe we can try to make do ad using eigen 3.9 and 4.0 to see how the adjoints change? 

> Anywhere we need to calculate log(sum(exp(x)) we should be using log_sum_exp. But it's not going to help if we just need to calculate sum(log(x)).

Oh yes sorry totally missed on that, not the right tool.

andrjohns:
> I'm not to sure how to progress at this point besides hitting up the eigen team. I can look into the different functions there to see how log is called. Maybe we can try to make do ad using eigen 3.9 and 4.0 to see how the adjoints change?

I've also tried comparing the gradients, log_prob, and grad_log_prob results, and still haven't been able to replicate a difference. My last attempt is going to be diff-ing the assembly from the cmdstan model with 3.4.0 vs 3.4.0+patch.

After that we might have to call in the big guns with Eigen

andrjohns:
@SteveBronder I think I've got it! The vectorised `log()` is returning slightly different values depending on the number of elements: https://godbolt.org/z/7froaMq5s

I'm guessing this is because the SIMD handling is different between the two. Now that we've got an easy example I'll open an issue in the Eigen repo.

What a journey!

andrjohns:
The verdict [from Eigen](https://gitlab.com/libeigen/eigen/-/issues/2565#note_1153186179) is that these differences are to be expected given the changes to the vectorised code

Are we happy to accept the new values for the test here?

SteveBronder:
I think I'm okay with it, it would be nice to be able to do this upgrade. I wonder if it would be okay to change the actual model to be better at recovering the parameters? Like if we can make the cauchy priors into student t or normal priors?

https://github.com/stan-dev/stat_comp_benchmarks/blob/fcf97113da2f7d9a0356a59f16d5381131a21189/benchmarks/irt_2pl/irt_2pl.stan

bob-carpenter:
>  I wonder if it would be okay to change the actual model 

Sure, especially if it's only an example we use for performance or other testing.

andrjohns:
Thanks both, sounds like a plan to me! I'll revert the patch I made to the Eigen code here, and open a PR optimising the performance-tests model

andrjohns:
@SteveBronder I've reverted the patch commit I made, so (assuming that only the `irt_2pl` model fails again) this should be ready for review & merge

andrjohns:
Now that the feature freeze is over, should we look at getting this merged? @rok-cesnovar or @SteveBronder , not sure who would be the best choice for reviewer here

RajalakshmiSR:
Is there any target release version(stan) for this eigen 3.4 update?

WardBrian:
@andrjohns did you ever change the performance tests model? If so, if you merge in `develop` and it passes I think we can get this approved and merged

andrjohns:
> @andrjohns did you ever change the performance tests model? If so, if you merge in `develop` and it passes I think we can get this approved and merged

In this case, changing the performance-tests model is unlikely to make this pass - since it's due to minor numerical differences in the vectorised handling of the `log()` function. In this instance, the changes will fail the first merge but will pass in the subsequent (since the vectorised operations will be the same).

Alternatively, if we want this to pass before we merge, we can also disable vectorised operations in the performance tests [this line](https://github.com/stan-dev/performance-tests-cmdstan/blob/15abdadb4a72784f4471ebd0bd6d700806b2fef0/compare-git-hashes.sh#L10).

Thoughts/preferences?

WardBrian:
I think we will also want to update the result stored in https://github.com/stan-dev/performance-tests-cmdstan/blob/master/golds/stat_comp_benchmarks_benchmarks_irt_2pl_irt_2pl.gold, no?

WardBrian:
@andrjohns thoughts on the golds question above?

WardBrian (dismissed):
I looked over the few changed files outside of the `lib/` folder and they all seem reasonable. 

I think we can do whatever is required to make the performance tests update themselves after this

WardBrian:
@serban-nicusor-toptal After merging in master a test is failing to compile on Windows due to out-of-memory, have we handled things like this in the past?

rok-cesnovar:
We split the tests into multiple files. Though this is the first time I am seeing this happening since we switched to Rtools4+ on Windows.

serban-nicusor-toptal:
I'm afraid Rok solution is the right one, else we would need to ask Dylan to make the Windows executors with more RAM.

WardBrian:
@andrjohns @SteveBronder I believe this should be ready to go. Earlier in this thread you mentioned wanting to rebase all the stan changes into one commit and Eigen changes into another, would you still like to do this?

WardBrian (approved):
Discussed with @SteveBronder and decided rather than rebasing we can just squash this PR

RajalakshmiSR:
What is the expected stan-math release for this PR?

rok-cesnovar:
The next release will include it, so 4.6.0 (part of Stan 2.32 release). The release should be official in the middle of April: https://discourse.mc-stan.org/t/planning-the-2-32-release/30641

PR #2591, opened 2021-09-27
StanHeaders patches for rstan 2.21

andrjohns:
@hsbadr These are changes needed for some of the rstan reverse dependencies to build under the rstan 2.21 & StanHeaders 2.26 combo.

The `elt_divide` changes are because the relaxed templating was matching to some Eigen expressions rather than just scalars, which caused build errors with `rstanarm`.

The `transpose` changes are needed because an expression was being returned which caused build errors for `rmdcev`

### DIFF
--- a/stan/math/prim/fun/elt_divide.hpp
+++ b/stan/math/prim/fun/elt_divide.hpp
@@ -51,8 +51,8 @@ auto elt_divide(const Mat& m, Scal s) {
 #else
 template <typename Scal, typename Mat, typename = require_stan_scalar_t<Scal>,
           typename = require_eigen_t<Mat> >
-auto elt_divide(Scal s, const Mat& m) {
-  return (s / m.array()).matrix().eval();
+auto elt_divide(const Mat& m, Scal s) {
+  return (m.array() / s).matrix().eval();
 }
 #endif
 
@@ -115,7 +115,8 @@ Eigen::Matrix<return_type_t<T1, T2>, R, C> elt_divide(
  * @param s scalar
  * @return Elementwise division of a scalar by matrix.
  */
-template <typename T1, typename T2, int R, int C>
+template <typename T1, typename T2, int R, int C,
+          require_all_stan_scalar_t<T1, T2>* = nullptr>
 Eigen::Matrix<return_type_t<T1, T2>, R, C> elt_divide(
     const Eigen::Matrix<T1, R, C>& m, T2 s) {
   return m.array() / s;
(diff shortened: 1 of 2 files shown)


hsbadr:
Great! Thanks @andrjohns! I'll rerun the reverse dependency checks with this patch. A few other packages are affected.

hsbadr:
@andrjohns LGTM. I'll merge this and https://github.com/stan-dev/stan/pull/3064 for testing and update the revdep results (https://github.com/stan-dev/rstan/tree/StanHeaders_2.26/revdep).

hsbadr (approved):

hsbadr:
@andrjohns This has fixed the `elt_divide` installation errors. The number of packages failed to check/install has been reduced from 34 to 25, which is great.

Though, `rstanarm`, and many other packages, fails (in unit tests or installation) due to TBB linking:
```r
rstanarm:
=========
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
     12 | #include <tbb/task_scheduler_init.h>
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
  /usr/lib/R/etc/Makeconf:177: recipe for target 'tests.o' failed
  make: *** [tests.o] Error 1
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Error (test_stan_functions.R:71:1): (code run outside of `test_that()`) ─────
  Error: Error 1 occurred building shared library.
  Backtrace:
      █
   1. └─Rcpp::sourceCpp(...) test_stan_functions.R:71:0
  
  [ FAIL 1 | WARN 0 | SKIP 0 | PASS 1 ]
  Error: Test failures
  Execution halted
```

@bgoodri @wds15 We need to push package maintainers to link to `RcppParallel` or bundle Intel TBB in `StanHeaders`. The fastest way for me is to at least test the later. So, I'll give it a try in `StanHeaders 2.26.4` but open to suggestions.

andrjohns:
I've also submitted patches to the packages on the revdep list, I'm tracking the progress of that here: https://discourse.mc-stan.org/t/r-package-maintainers-using-stan-please-read/24131/14

wds15:
I would not like to bundle the TBB inside StanHeaders if we can avoid that... However, could we forward the few headers like `tbb/task_scheduler_init.h` to the actual TBB installation part of `RcppParallel`? This should get things to compile... and maybe it even works...

the better solution is to fix one by one as @andrjohns is doing it right now... .which is really awesome that you take the energy to go one-by-one here.

hsbadr:
> I've also submitted patches to the packages on the revdep list, I'm tracking the progress of that here: https://discourse.mc-stan.org/t/r-package-maintainers-using-stan-please-read/24131/14

Perfect. Thanks @andrjohns!

> could we forward the few headers like `tbb/task_scheduler_init.h` to the actual TBB installation part of `RcppParallel`?

I tried this already. It fixes a few packages while most packages will fail to load due to undefined symbols (i.e., they need to link `-ltbb` or find the TBB functions in libStanHeaders.a, which does work -- I know it isn't recommended by Intel/TBB team for reasons that don't necessarily apply here, but I agree that @andrjohns's approach is the best solution that could require more time).

Issue #2597, opened 2021-10-05
How to handle sparse matrices when f(0) returns not zero?


SteveBronder:
## Description

I'm working on some of the testing stuff for sparse matrices and had a quick Q. For a sparse matrix, how should we handle functions where an input of 0 would return back a dense matrix? i.e. for a sparse matrix `X` calling `acos(X)` where 0 returns `1.57~`. Should the return be dense or should we treat the non-filled cells as just empty and ignore them, only iterating on the cells with values in them before the operation?

#### Current Version:
v4.1.0

bob-carpenter:
Yes, you need to treat the matrix as having 0 entries everywhere there isn't an explicit non-zero value.  That means if we apply `exp()`, for example, the result is non-zero everywhere.

I think the bigger question is whether the result should be a sparse matrix or a dense matrix.  It feels like sparse matrix makes more sense in that the vectorized return type is the same as the input type.  But dense would be more efficient.  My inclination is to have the result be sparse.

The right thing to do for efficiency here is compute `acos(0)` once and copy it rather than creating a gazillion new autodiff variables.  That at least could be cooked into the definition.

I don't think we should try to get fancy and allow a non-zero value as the default for unspecified values.

SteveBronder:
imo I'd like a dense matrix, not only because it will be more efficient but it makes the API very explicit that zeroed values will become nonzero.

nhuurre:
AFAIK sparse matrices are most useful when dense representation would inefficient. Automatically converting to a dense matrix seems like a footgun to me. What's the purpose of supporting such "sparsity destroying" operations in the first place?

bob-carpenter:
@nhuurre: The problem is that when you apply `exp()` to a sparse matrix, the result is dense.  The question is really whether we want to auto-convert to a dense result for efficiency.  @SteveBronder is suggesting a signature like this:

```
matrix cos(sparse_matrix);
```

because `cos(0) != 0`.  But because `sin(0) = 0`, presumably we'd have

```
sparse_matrix sin(sparse_matrix);
```

This inconsistency is what worries me about returning a dense matrix from density inducing functions like `exp` or `cos`.

nhuurre:
> The problem is that when you apply exp() to a sparse matrix, the result is dense.

And my question is: if applying exp() to a sparse matrix defeats the point of using sparse matrices, why would you want to apply exp() to a sparse matrix?

bob-carpenter:
[...]

PR #2599, opened 2021-10-05
[WIP] Adds specializations for handling sparse matrix unary operations

SteveBronder:
## Summary

This is a WIP for supporting sparse matrix unary operations in Stan math. Related to #2597 , this API will return a sparse matrix with all values filled in for sparse matrices when the function does not have a `f(0) -> 0` mapping. The scheme here just adds a `bool ApplyZero` template parameter to `apply_scalar_unary` where if `ApplyZero` is true, a dense matrix is returned and for `false` a sparse matrix is returned.

Though note if we want to change the API to always be `Sparse -> Sparse` that's pretty easy here

## Tests

The tests have been modified for `atan()` and `acos()` to take a vector input and the usual lambda functor we use for testing and make a sparse matrix with the vector values along the diagonal via a helper function `make_sparse_mat_func()`.

## Release notes

Adds support for Unary sparse matrix functions

## Checklist

- [x] Math issue #2597 


SteveBronder:
@andrjohns would you mind taking a look at this?

stan/math/prim/functor/apply_scalar_unary.hpp
```diff
@@ -59,15 +58,82 @@ struct apply_scalar_unary<F, T, require_eigen_t<T>> {
...
+  template <typename DenseMat, require_eigen_dense_base_t<DenseMat>* = nullptr>
+  static inline auto apply(const DenseMat& x) {
     return x.unaryExpr(
         [](scalar_t x) { return apply_scalar_unary<F, scalar_t>::apply(x); });
   }
 
+  /**
+   * Special case for `ApplyZero` set to true, returning a full sparse matrix.
+   * Return the result of applying the function defined by the template
+   * parameter F to the specified matrix argument.
```
andrjohns, line 75: The doc for both overloads state that a sparse matrix is returned, so might need to clarify which is which

stan/math/prim/functor/apply_scalar_unary.hpp
```diff
@@ -59,15 +58,82 @@ struct apply_scalar_unary<F, T, require_eigen_t<T>> {
...
+    std::vector<triplet_t> triplet_list(x_size, triplet_t(0, 0, zeroed_val));
+    for (Eigen::Index i = 0; i < x.rows(); ++i) {
+      for (Eigen::Index j = 0; j < x.cols(); ++j) {
+        // Column major order
+        triplet_list[i * x.cols() + j] = triplet_t(i, j, zeroed_val);
+      }
+    }
+    for (Eigen::Index k = 0; k < x.outerSize(); ++k) {
+      for (typename SparseMat::InnerIterator it(x, k); it; ++it) {
+        triplet_list[it.row() * x.cols() + it.col()]
```
andrjohns, line 101: Rather than iterating over each element, would it be more performant to Map the values as a vector and call the vectorised implementation? For example:
```cpp
Map<Eigen::VectorXd> mapSparse(x.valuePtr(), x.nonZeros());

apply_scalar_unary<F, Eigen::VectorXd>::apply(mapSparse);
```

stan/math/prim/meta/is_container.hpp
```diff
@@ -15,12 +15,13 @@
...
 
 /**
- * Deduces whether type is eigen matrix or standard vector.
+ * Deduces whether type is a dense eigen type or standard vector.
  * @tparam Container type to check
  */
 template <typename Container>
-using is_container = bool_constant<
-    math::disjunction<is_eigen<Container>, is_std_vector<Container>>::value>;
+using is_container
```
andrjohns, line 22: Wouldn't we also expect sparse matrices to be detected as a container though?

andrjohns (changes requested):
Sorry for the delay in getting to this! Got a couple of queries and comments, but nothing major

stan/math/prim/functor/apply_scalar_unary.hpp
```diff
@@ -36,8 +31,12 @@ namespace math {
  *
  * @tparam F Type of function to apply.
  * @tparam T Type of argument to which function is applied.
+ * @tparam ApplyZero If true, the function applied is assumed to return zero for
+ *  inputs of zero and so sparse matrices will return sparse matrices. A value
+ * of false will return a dense matrix for sparse matrices.
  */
-template <typename F, typename T, typename Enable = void>
+template <typename F, typename T, bool ApplyZero = false,
```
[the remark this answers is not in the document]
SteveBronder: Sorry I need to update the docs and this PRs summary. After discussion with @bob-carpenter, @nhuurre, and the Eigen dev team (mostly in https://github.com/stan-dev/math/issues/2597) I think it makes more sense to return a sparse matrix with all entries filled in. I do kind of wonder if we should have something like exp_nonzero(sparse_matrix) available at the Stan language level where it will only run over the nonzero entries. But we can sort that out at a later time

stan/math/prim/functor/apply_scalar_unary.hpp
SteveBronder: It would def be better, though I avoided it because I was a bit confused about the compressed sparse matrix scheme Eigen uses. If you look at their docs [here](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html) under 'The SparseMatrix class' they sometimes allow empty values so its easier to write to the sparse matrix like

```
Values:	22	7	_	3	5	14	_	_	1	_	17	8
InnerIndices:	1	2	_	0	2	4	_	_	2	_	1	4
```

I was worried that directly grabbing the values could lead to some bad things since there may be places in the value pointer that are not initialized. Another issue is that our `arena_matrix` scheme assumes that if you are passing it an Eigen map that the memory has already been allocated elsewhere for use in the reverse mode pass, which may not explicitly be true for sparse matrices.

So I agree what your saying here is def faster, but imo for this first pass I'd rather do the safer path atm. Once we have everything working at the language level and know more about our constraints and assumptions we can make then we can come back and do these kinds of optimizations.

SteveBronder: Hmm, okay now I'm thinking about this more and I think that at the stan language level we can make it so that you can only assign to elements that are already non-zero. So I wonder if what we can do here is just assume that we are always working with compressed matrices, since then we don't have to worry about the uninitialized values and can do the fast thing safely

stan/math/prim/meta/is_container.hpp
SteveBronder: So I'm kind of confused. It seems like a lot of the time we are declaring things containers when they are "vectorizable" so for sparse matrix I was like, "Well these are not really vectorizable so I would not say they are a container". imo I kind of feel like we should just have `container` simply mean a container pointing to items containing dynamically allocated memory like `std::vector<std::vector<Eigen::Matrix>>`. Like there is a few types of memory patterns that I'm thinking of here

`std::vector<double/var> & Eigen::Matrix<double/var, R, C>` where generic memory can be mapped to access linearly
`std::vector<std::vector<double/var> & Eigen::Matrix<double/var, R, C>>` where generic memory the container points to can be mapped to linearly
`Eigen::SparseMatrix<double/var>` If compressed, this can be mapped to generic memory and accessed linearly.

Hmm, maybe this PR is the wrong path and I should be using the `apply_vector_unary` pattern? Because then I would do what you are saying in another comment above and just calling the vectorized function over the memory

syclik:
Moving to draft; it's still marked at "WIP." Please move it out of draft when you're ready.

PR #2600, opened 2021-10-07
Bugfix/adjoint ode odd

wds15:
## Summary

Fix issues with Adjoint ODE system as reported on the forum: https://discourse.mc-stan.org/t/adjoint-ode-behaviour/24703/13

## Tests

At the moment I don't have a test for the error other than that the model described in the post does not sample ok (which the adjoint rc2 did sample ok).

## Side Effects

None

## Release notes

Fix issues with Adjoint ODE memory management.

## Checklist

### DIFF
--- a/stan/math/rev/functor/cvodes_integrator_adjoint.hpp
+++ b/stan/math/rev/functor/cvodes_integrator_adjoint.hpp
@@ -223,9 +223,9 @@ class cvodes_integrator_adjoint_vari : public vari_base {
         num_steps_between_checkpoints_(num_steps_between_checkpoints),
         N_(y0.size()),
         msgs_(msgs),
-        y_return_varis_(is_var_return_ ?
-                        ChainableStack::instance_->memalloc_.alloc_array<vari*>(
-                            N_ * ts.size()) : nullptr),
+        y_return_varis_(is_var_return_ ? ChainableStack::instance_->memalloc_
+                                             .alloc_array<vari*>(N_ * ts.size())
+                                       : nullptr),
         args_varis_([&args..., num_vars = this->num_args_vars_]() {
           vari** vari_mem
               = ChainableStack::instance_->memalloc_.alloc_array<vari*>(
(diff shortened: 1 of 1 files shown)


SteveBronder:
Can we just take the model `log_prob` from the discourse post and plug it into the `expect_ad` framework?

wds15:
Maybe... but then you also need the data and the initial... not trivial.

wds15:
@SteveBronder this PR avoids use of the arena things and stores everything the CVODES solver touches inside the `cvodes_solver` object. Doing so does make the really odd `Inf` errors on supplied arguments and data go away in the example model provided on the forum. The model still only produces divergent transitions... but that could also be an issue with the model - I do not have a way to discern that.

So from my end, I'd think this is good to go.

wds15:
oh... how could I forget... anything part of a vari must be POD!!! Hopefully the last change will. fix this then.

wds15:
Looks like memory leaks are gone…@stevebronder wanna review?

rok-cesnovar:
A gentle bump for this. If possible, I would include this in the 2.28.1 release we have to do due to another bug. @SteveBronder any chance you have time to review in the next days? Would love to get the patch release out this week so we can get the release period behind us :)

rok-cesnovar:
I think this might be it: https://github.com/teddygroves/minimal_maudfail/blob/master/functions.stan

wds15:
Oh…sorry for that!  But I would anyway suggest that for now we merge this if you are fine. We should file an issue and investigate…given we want 2.28.1 sooner than later. If you can get a test to work quickly then thats obviously great! I am suspecting that the arena typed things were not arenad enough in that we have arena of a vector which itself stores eigen things…and the eigen vector was not arenad. The code now uses a safe pattern by putting everything into an object which we clean away anyway.

wds15:
The inf issues go away with the changes i did. However, I dont trust the models from the forum too much in that these can be too extreme such I just dont know if stan throws up or the adjoint solver (or both). I would like to rerun the sbc ode example i did a while ago to confirm that things are ok for that one model.

SteveBronder:
> See my comment here #2600 (comment) . Those NA / Inf values are coming up before we ever use the adjoint method. Is there some reason those adjoint methods would effect the stuff going into them?

To be more clear, it's confusing to me that we are targeting the `NA` / `Inf` values as the thing to fix in this PR whenever it looks like those are not created within the adjoint ODE method itself but seem to come from another part of the model we are using for testing

wds15:
Here is the model using the bdf integrator instead. It moves slowly, but it does sample OK as it looks to me.

[model-bdf_stan.txt](https://github.com/stan-dev/math/files/7351930/model-bdf_stan.txt)

PR #2602, opened 2021-10-13
[Fix] Adjoint Ode bug fixes

SteveBronder:
## Summary

This has a few fixups for the reverse mode ode adjoint method. There are quite a few places where I was worried we could be indexing incorrectly or shallow copying objects when we should be doing a hard copy. Changes include:

`count_vars.hpp`

1.  Changing order of functions for `count_vars_impl` so that the forward declares are not needed 

`cvodes_integrator_adjoint.hpp`

1. Moving `y_return_` out of `solver_` and into the top class wrapped with `arena_t<std::vector<Matrix>>`. 

2. Fix list initializer order of object instantiation in `cvodes_solver` (this usually throws an error in gcc so I'm surprised it did not.)

3. Moves all the checks into the initializers for each of the top class's members. This looks a bit odd, but I had a really hard time parsing out what we were checking when these were all laid out in the body of the class's constructor since there's so many. So I made an immediately invoked lambda within each of the initializers that does the checks for each input directly in the initializer.

```cpp
        ts_([](const char* function_name, auto&& ts) {
          check_nonzero_size(function_name, "times", ts);
          check_finite(function_name, "times", ts);
          check_sorted(function_name, "times", ts);
          return arena_t<std::vector<T_ts>>(ts.begin(), ts.end());
        }(function_name, ts)),
```

I'm fine with moving this back to in the body but imo this cleans up everything nicely and allows for `solver_` to also be constructed in the initializer list.

4. Removes the `store_state()` member function. Instead all we need to do is make a deep copy of `state_forward_`. The reason for this is because `state_forward_` is an `arena_matrix<Eigen::VectorXd>`. `arena_matrix<>` types are the equivalent of a map around memory allocated by the stack allocator

```
Eigen::Map<Eigen::Matrix<Scalar, Rows, Cols>>(ChainableStack::alloc_array<Scalar>(rows * cols), rows, cols)
```

One important point is that they do shallow copies of each other (essentially assigning an arena matrix to another arena matrix is the same as just giving the lhs the pointer of the rhs. What this all means is that since `state_forward_` is an arena matrix and we don't want it's memory going into the `y_return_` or `y_` matrices we need to make a hard copy of it to a `Eigen::VectorXd` before we do assignment.

5. Makes a hard copy of `y_return_` when calling `solution()`

6. Gives a nicer error message when a particular parameter or data comes in as non-finite during the forward pass.

7. In chain whenever two eigen matrices of the same size are being assigned to each other I removed some of the explicit loops and used eigen matrix assignments directly. This should help us ensure we don't have indexing errors.

`arena_type.hpp`

This was just a small change to the `std::vector<>` overload to make sure we also caught `const std::vector<T>&` etc. template parameters.

## Tests

Right now I'm just testing this locally by running the example from discourse available in the comment [here](https://discourse.mc-stan.org/t/adjoint-ode-behaviour/24703/42?u=stevebronder). Once we think this looks good and @wds15 tests locally I can take the ODE part of that model and put it into the testing framework.

@wds15 when running it locally the first few iterations throw errors, but then after that the model seems to sample fine. Is this okay/what we would expect? Or would we not know until all the sampling is complete?

```
$ ./Test_Adjoint_v2 data file="adjoint_data.json" init="init.json" sample output refresh=2
method = sample (Default)
  sample
    num_samples = 1000 (Default)
    num_warmup = 1000 (Default)
    save_warmup = 0 (Default)
    thin = 1 (Default)
    adapt
      engaged = 1 (Default)
      gamma = 0.050000000000000003 (Default)
      delta = 0.80000000000000004 (Default)
      kappa = 0.75 (Default)
      t0 = 10 (Default)
      init_buffer = 75 (Default)
      term_buffer = 50 (Default)
      window = 25 (Default)
    algorithm = hmc (Default)
      hmc
        engine = nuts (Default)
          nuts
            max_depth = 10 (Default)
        metric = diag_e (Default)
        metric_file =  (Default)
        stepsize = 1 (Default)
        stepsize_jitter = 0 (Default)
    num_chains = 1 (Default)
id = 1 (Default)
data
  file = adjoint_data.json
init = init.json
random
  seed = 2179170925 (Default)
output
  file = output.csv (Default)
  diagnostic_file =  (Default)
  refresh = 2
  sig_figs = -1 (Default)
  profile_file = profile.csv (Default)
num_threads = 1 (Default)


Gradient evaluation took 0.463662 seconds
1000 transitions using 10 leapfrog steps per transition would take 4636.62 seconds.
Adjust your expectations accordingly!


Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: ode_adjoint_tol_ctl: ode parameters and data[397] is inf, but must be finite! (in '/tmp/Rtmp5le0nw/model-1b38f07d4012f8.stan', line 141, column 2 to line 156, column 74)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: ode_adjoint_tol_ctl: ode parameters and data[397] is inf, but must be finite! (in '/tmp/Rtmp5le0nw/model-1b38f07d4012f8.stan', line 141, column 2 to line 156, column 74)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: CVODES: CVode At t = 0 and h = 5.75889e-102, the corrector convergence test failed repeatedly or with |h| = hmin. Error code: -4 (in '/tmp/Rtmp5le0nw/model-1b38f07d4012f8.stan', line 141, column 2 to line 156, column 74)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
CVODES: CVode At t = 394 and h = -2.81685e-26, the error test failed repeatedly or with |h| = hmin. Error code: -3
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Iteration:    1 / 2000 [  0%]  (Warmup)
Iteration:    2 / 2000 [  0%]  (Warmup)
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: normal_lpdf: Scale parameter is 0, but must be positive! (in '/tmp/Rtmp5le0nw/model-1b38f07d4012f8.stan', line 188, column 23 to column 59)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Iteration:    4 / 2000 [  0%]  (Warmup)
Iteration:    6 / 2000 [  0%]  (Warmup)
Iteration:    8 / 2000 [  0%]  (Warmup)
Iteration:   10 / 2000 [  0%]  (Warmup)
Iteration:   12 / 2000 [  0%]  (Warmup)
```

## Release notes

Bug fix for ODE adjoint method

## Checklist

### DIFF
--- a/stan/math/rev/functor/cvodes_integrator_adjoint.hpp
+++ b/stan/math/rev/functor/cvodes_integrator_adjoint.hpp
@@ -416,7 +416,6 @@ class cvodes_integrator_adjoint_vari : public vari_base {
       }
       N_VDestroy_Serial(nv_state_forward_);
       ChainableStack::instance_->var_stack_.push_back(this);
-
     } catch (...) {
       N_VDestroy_Serial(nv_absolute_tolerance_forward_);
       SUNMatDestroy(A_forward_);


wds15:
This version does not work at all for me. CVODES complains not being able initialise... and the reason is that you de-allocated the forward data-structures of CVODES after solving the forward problem. That's not OK, since CVODES really expects all things to be still in place when doing the backward solve.

Moreover, I switched the model posted in the other PR over to use the bdf solver. Once I do that, then I get a very slow running program, but the program seems to sample just fine - I do not get any of those Inf rejections. I was too impatient to wait for all 10 iterations (will let that program run on the side and report later).

So as it stands this version in this PR seems to still suffer from the problems we have originally and "my" PR seems to avoid these issues with Inf - and the bdf integrator is also working fine and does not trigger throwing "Inf" issues. How about we go ahead with the other PR? Either as is or we clean that up if you think that is needed?

PR #2612, opened 2021-11-06
Add missing scalar prim signatures

### LINKED ISSUE #2611: Add stub functions for certain std:: functions
This is related to https://github.com/stan-dev/stanc3/issues/997 / https://github.com/stan-dev/stanc3/pull/1011

There are several functions which we don't define in stan::math but rather use std:: (or don't qualify them at all) for when the argument is only reals or ints. As far as I can tell, these are:

- [ ] pow
- [ ] fmod
- [ ] atan2
- [ ] min, max (the versions that accept 2 scalars, not the ones for arrays etc)
- [x] ceil

It would be great to redirect these within the library, so we can codegen all functions as fully qualified, e.g. `stan::math::pow`.

We may also need more definitions for `norm` and `pow` of complex numbers, see [this log](https://jenkins.mc-stan.org/blue/rest/organizations/jenkins/pipelines/stanc3/branches/PR-1011/runs/4/nodes/192/steps/213/log/?start=0) for what currently fails

rok-cesnovar:
## Summary

Adds missing scalar implementations for max, min, fmod, atan2 and pow to fix #2611. 

Additionally, this adds a test that generates unit tests on the fly. These tests check that all supported stanc3 scalar signatures (excluding those that return complex numbers and _rng functions) compile in the stan::math namespace. Additionally, it also checks that nothing was placed on any var stack (no inputs get promoted to vars).

Example of a test that is generated:

```cpp
TEST(PrimSigsTests, pow4786) {
  /*
  pow(real, int) => real
   */
  double real0 = 0.4;
  int int1 = 1;
  auto result2 = stan::math::eval(stan::math::pow(real0, int1));
  auto result3 = stan::math::eval(stan::math::pow(real0, int1));
  EXPECT_STAN_EQ(result2, result3);
  EXPECT_EQ(stan::math::ChainableStack::instance_->var_stack_.size(), 0);
  EXPECT_EQ(stan::math::ChainableStack::instance_->var_nochain_stack_.size(), 0);
  EXPECT_EQ(stan::math::ChainableStack::instance_->var_alloc_stack_.size(), 0);
}
```
The functions are called twice because I wanted to make sure I included the result in an EXPECT test so it does not get optimized out. This might not be needed, but I just wanted to be safe.

The entire test file that is generated is here: [prim_sig_test.txt](https://github.com/stan-dev/math/files/7491206/prim_sig_test.txt)

This test would be run on Github Actions.

## Tests

Adds a new signature test that can be run via `./runTests.py test/scalar_prim_sig`. This generates a unit test file that checks that all primitive scalar functions compile and that none of them gets promoted to vars.

## Side Effects

/

## Release notes

Added implementations of max, min, fmod, atan2 and pow for scalar inputs to the stan::math namespace.

## Checklist

- [x] Math issue #2611


bob-carpenter:
I can review this.

The current failure is a bit worrisome---it looks like it just failed outright in the LLVM judging from the plea to submit a bug report:

```
clang: error: unable to execute command: Killed
clang: error: clang frontend command failed due to signal (use -v to see invocation)
clang version 6.0.0-1ubuntu2~16.04.1 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
clang: note: diagnostic msg: PLEASE submit a bug report to http://llvm.org/bugs/ and include the crash backtrace, preprocessed source, and associated run script.
clang: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /tmp/exp_mod_normal_lpdf_test-2506fe.cpp
clang: note: diagnostic msg: /tmp/exp_mod_normal_lpdf_test-2506fe.sh
clang: note: diagnostic msg: 

********************
```

rok-cesnovar:
Yeah, the tests were already passing but failed upstream because https://github.com/stan-dev/stan/pull/3080 was required. Restarted the tests, should be ready for review.

test/generate_prim_sig_tests.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python
```
bob-carpenter, line 1: What's the motivation for not just writing unit tests for these?  These tests are all very much boilerplate, but I don't see the need for moving to Python.  One of the reasons I prefer basic tests is that they're easy to find in the code.  With this kind of thing, I'll never be able to find where `math(int, double)` gets tested.  And these should all be tested on a mix of integer and real inputs to make sure promotion and return type calculation works as expected.

And where are we testing that they produce the same values as the C++ functions?

test/generate_prim_sig_tests.py
```diff
@@ -0,0 +1,90 @@
...
+        cg = CodeGenerator()
+
+        arg_list_no_expression = cg.build_arguments(
+            sp, sp.number_arguments() * ["Prim"], size=1
+        )
+
+        # Check results with expressions and without are the same
+        cpp_function_name = "stan::math::" + sp.function_name
+        result_1 = cg.function_call_assign(cpp_function_name, *arg_list_no_expression)
+        result_2 = cg.function_call_assign(cpp_function_name, *arg_list_no_expression)
```
bob-carpenter, line 66: `result_1` and `result_2` have the same definition.  Should one of these have a different arg list?

test/prob/frechet/frechet_ccdf_log_test.hpp
```diff
@@ -69,9 +69,9 @@ class AgradCcdfLogFrechet : public AgradCcdfLogTest {
   stan::return_type_t<T_y, T_shape, T_scale> ccdf_log_function(
       const T_y& y, const T_shape& alpha, const T_scale& sigma, const T3&,
       const T4&, const T5&) {
-    using std::exp;
-    using std::log;
-    using std::pow;
+    using stan::math::exp;
+    using stan::math::log;
+    using stan::math::pow;
     return log(1.0 - exp(-pow(sigma / y, alpha)));
```
bob-carpenter, line 75: We have a `log1m_exp` function built in and we should use that for numerical stability.   Or at least `log1p(-exp(...))`.

test/prob/pareto_type_2/pareto_type_2_ccdf_log_test.hpp
```diff
@@ -65,8 +65,6 @@ class AgradCcdfLogParetoType2 : public AgradCcdfLogTest {
   stan::return_type_t<T_y, T_loc, T_scale, T_shape> ccdf_log_function(
       const T_y& y, const T_loc& mu, const T_scale& lambda,
       const T_shape& alpha, const T4&, const T5&) {
-    using std::log;
-    using std::pow;
-    return log(pow(1.0 + (y - mu) / lambda, -alpha));
+    return stan::math::log(stan::math::pow(1.0 + (y - mu) / lambda, -alpha));
```
bob-carpenter, line 68: Shouldn't `log(pow(a, b))` always be reduced to `b * log(a)`?

test/prob/pareto_type_2/pareto_type_2_cdf_test.hpp
```diff
@@ -65,7 +65,6 @@ class AgradCdfParetoType2 : public AgradCdfTest {
   stan::return_type_t<T_y, T_loc, T_scale, T_shape> cdf_function(
       const T_y& y, const T_loc& mu, const T_scale& lambda,
       const T_shape& alpha, const T4&, const T5&) {
-    using std::pow;
-    return 1.0 - pow(1.0 + (y - mu) / lambda, -alpha);
+    return 1.0 - stan::math::pow(1.0 + (y - mu) / lambda, -alpha);
```
bob-carpenter, line 68: [optional]
These don't need the `.0` after the `1`s.

test/signature_parser.py
```diff
@@ -41,6 +41,18 @@ def has_eigen_compatible_arg(self):
         """Return true if any argument is vector-like (can be an Eigen c++ type)"""
         return any(arg in ("matrix", "vector", "row_vector") for arg in self.stan_args)
 
+    def has_scalar_only_args(self):
```
bob-carpenter, line 44: [question]
What is this python program for?

test/unit/math/prim/fun/pow_test.cpp
```diff
@@ -4,18 +4,16 @@
 #include <cmath>
 
 TEST(MathFunctions, pow) {
-  using stan::math::pow;
-  using std::pow;
-
-  EXPECT_FLOAT_EQ(pow(2.15, 1.71), 3.70228430936580996892756228068688);
-  EXPECT_FLOAT_EQ(pow(7.62, -1.15), 0.0967724604023293179518727243992);
+  EXPECT_FLOAT_EQ(stan::math::pow(2.15, 1.71),
```
bob-carpenter, line 7: I don't know if these show up elsewhere, but we need to test integer and double values.  And we can test them against results return by casting them up to double and calling the std versions.

bob-carpenter (changes requested):
A bunch of picky doc comments and some more substantive testing questions.  My main concern is that all the promotion of integers is tested as that's been an ongoing problem for us with the standard library functions.  Not something complicated, just integer tests for all arguments.  So if it's binary, that's four cases of {double, int} x {double, int}.

stan/math/prim/fun/atan2.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <stan/math/prim/core.hpp>
+#include <stan/math/prim/meta.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/**
+ * Returns the principal value of the arg tangent of y/x,
```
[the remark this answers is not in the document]
rok-cesnovar: Honestly, I copied the definition from https://www.cplusplus.com/reference/cmath/atan2/. Should have at least gone to cppreferenceinstead. Much simpler and cleaner.

test/generate_prim_sig_tests.py
rok-cesnovar: The decision to make these tests automatic is mostly because I wanted to test that all scalar argument functions work correctly. 

But I guess we only need to test std:: shadowing functions, which is a non-changing small list of functions. Will revert to that.

test/prob/frechet/frechet_ccdf_log_test.hpp
rok-cesnovar: Didn't really stop to think about what these are doing, but you are right. And if we are changing these files, it makes sense to clean them up as well.

test/signature_parser.py
rok-cesnovar: The Python test retrieves the list of all stanc3 supported signatures and this function was used to trim that list down to only functions that have scalar arguments. But given that I realized this is most likely overkill, this will go away anyways.

hsbadr:
@rok-cesnovar JFYI. This breaks the R packages that depend on `rstan` with, for instance, `pow()` ambiguity. I'm using the experimental branch of `rstan` with the nightly version of `stanc3`.

rok-cesnovar:
Yeah, I know (it also fails upstream for the exact same reason - the fact that some std functions live in the global namespace). 

This branch will be merged together accompanied by a change in stanc3 that will add explicit stan::math:: namespace to all generated Stan Math.
See https://github.com/stan-dev/stanc3/pull/1011 and https://github.com/stan-dev/stanc3/issues/997

Also thanks for checking it. I love that we have up-to-date rstan now!

stan/math/prim/fun/atan2.hpp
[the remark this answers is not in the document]
bob-carpenter: Yikes!  I strongly prefer cppreference.com.

WardBrian:
For the benefit of anyone who is interested primarily in the Math development without following stanc3 super closely, this PR is motivated by an issue that can lead to programs that stanc3 compiles but the c++ compiler can't. For example,


```stan
parameters {
  real alpha;
  real beta;
}
transformed parameters {
  real beta_function = beta(alpha, beta)
}
```

We allow users to define a variable called beta, but without the explicit namespacing this PR allows the C++ compilers fail to compile the function call of beta.

stan/math/prim/fun/min.hpp
```diff
@@ -10,6 +10,21 @@
...
+ * scalar arguments.
+ *
+ * @tparam T1 type of first argument (must be arithmetic)
+ * @tparam T2 type of second argument (must be arithmetic)
+ * @param x first argument
+ * @param y second argument
+ * @return minimum value of the two arguments
+ */
+template <typename T1, typename T2, require_all_arithmetic_t<T1, T2>* = nullptr>
+return_type_t<T1, T2> min(T1 x, T2 y) {
```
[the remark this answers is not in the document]
rok-cesnovar: > Will this return an integer if x and y are both integers? 

Yes, it will. That is tested here: https://github.com/stan-dev/math/blob/develop/test/unit/math/mix/meta/return_type_test.cpp#L17

rok-cesnovar:
This is now ready for re-review. Ignore the test failure as it fails upstream (we will need some coordination to merge this and https://github.com/stan-dev/stanc3/pull/1011 together).

rok-cesnovar:
Please ignore the review request. 

Testing with the Stanc3 changes revealed that we also have to add Stan Math support for standard library functions that support std::complex inputs (arg, abs, norm, conj, proj, polar, exp, log, log10, pow, sqrt and trigonometric functions). 

The implementations should be simple, just calling the std:: function.

rok-cesnovar:
Added `stan::math` namespaced functions for `std::complex<T>` inputs, where `T` is not `var` for the following: arg, conj, norm, pow, proj. 

Also added tests for complex inputs for other complex functions: abs, exp, log, log10, sqrt and trigonometric.

All models now compile with https://github.com/stan-dev/stanc3/pull/1011 and this branch so this is ready for review.

bob-carpenter (approved):
looks good.  thanks for dealing with this huge pr.

Issue #2489, opened 2021-05-17
Moving to C++14


rok-cesnovar:
This is a step towards closing https://github.com/stan-dev/math/issues/1300 (moving up to C++17) that we can do immediately after the release.

The immediate consequence of moving up to C++14 (we currently support a subset - C++1y) is that the minimum required g++ would become g++ 5 as opposed to 4.9.3. There is no change for clang (clang does not have official support for C++1y) and the clang version supporting C++14 is Clang 3.4 which is the same one we listed as minimum now.

There are major changes that would be caused by that: 

- On Windows, RTools 3.5 would no longer be supported and thus all upstream Stan R packages that will use **Stan 2.28+ would no longer support R3.X**. 

    R 4.1 is now in the RC phase and once its released (any day now) CRAN will no longer test packages with R3.X. CRAN tests with the current and old release (which is now 3.6.X). Once R-release becomes 4.1, R-old will become 4.0.x so using R 4.X should be recommended in all those upstream packages anyway.

- cluster users would need to upgrade to g++ 5. 

    Clusters are the only other larger group of users that would be affected. Non-cluster Linux distros come with g++5 or higher for quite a few years now and Mac users are not affected.

The main reason of why I think the major changes are OK is that users that will not want to upgrade to g++ 5 will be able to stay with 2.27 which should cover them for quite some time (given that a majority of users still use Stan 2.21 we should be good on that front for awhile once we get to 2.27).

The only question I have on this is **does this requires a bump of the major version of Stan**? I am leaning towards no but I am not sure.

cc: @syclik @bob-carpenter @bbbales2 @t4c1 @SteveBronder @wds15 @andrjohns @yizhang-yiz et al

wds15:
I don't think this would need to bump in major version numbers.

As you say, it's a tough call for cluster users who will have to deal with it in some way.

SteveBronder:
I wonder if for R we could make a fixed version of stanheaders and rstan like `StanHeaders227` and `rstan227`, then users who need `C++1y` can use that while we then bump up to C++14 for rstan and `StanHeaders`

rok-cesnovar:
R 4.1 has officially been released. 

I am going to bring this up on next weeks meeting and then make a discourse thread to gather more feedback.

@SteveBronder AFAIU R will download the last suitable package that works with that version.

andrjohns:
[...]

PR #2619, opened 2021-11-23
fix matrix_exp_multiply accuracy bug

### LINKED ISSUE #2529: bug in scale_matrix_exp_multiply
## Description

The function `scale_matrix_exp_multiply` (and `matrix_exp_multiply`) appear to be having a bug. The results of these functions are not (always) identical to what one gets by using `matrix_exp` and subsequent multiplication.

## Example

This Stan model should reproduce the error (see the `generated quantities` block)

```stan
// matrix_exp.stan
data {
    int n; // number of time points
    real ts[n]; // time points 
    
    int k; // dimension of matrix
    matrix[k,k] M;
}

parameters {
    real u; // dummy parameter
}

model {
    u ~ normal(0,1); // dummy distribution
}

generated quantities {
    // xs, ys and zs should be identical
    real xs[n];
    real ys[n];
    real zs[n];
    
    // compute exp(t * M) * 1, where 1 is a vector of ones
    // and use store the first element to xs, ys and zs
    
    for ( i in 1:n ) {
    	// method 1
    	matrix[k,1] A = matrix_exp(ts[i] * M) * rep_matrix(1.0, k, 1);
    	xs[i] = A[1,1];

        // method 2
        ys[i] = matrix_exp_multiply(ts[i] * M, rep_matrix(1.0, k, 1))[1,1];        
        
        // method 3
        zs[i] = scale_matrix_exp_multiply(ts[i], M, rep_matrix(1.0, k, 1))[1,1];
    }
}
```
I used the following python script to run the model

```py
import cmdstanpy
import numpy as np
import matplotlib.pyplot as plt

sm = cmdstanpy.CmdStanModel(stan_file="matrix_exp.stan")

ts = np.linspace(0, 50, 1000)

M = np.array([
    [-1.2,  1.2,    0],
    [0,    -0.5,  0.5],
    [0,       0, -0.3]
])

data = {
    "n" : len(ts),
    "ts" : ts,
    "k" : len(M),
    "M" : M,
}

sam = sm.sample(data=data, chains=1)

xs = sam.stan_variable("xs")[0,:]
ys = sam.stan_variable("ys")[0,:]
zs = sam.stan_variable("zs")[0,:]

fig, axs = plt.subplots(1, 3, sharey=True)

axs[0].plot(ts, xs, label="method 1", linewidth=2)
axs[1].plot(ts, ys, label="method 2", linewidth=2)
axs[2].plot(ts, zs, label="method 3", linewidth=2)

axs[1].set_xlabel("$t$")
axs[0].set_ylabel("$[\\exp(Mt) \cdot 1]_{1}$")

for i in range(3):
    axs[i].set_title(f"method {i+1}")

fig.savefig("matrix_exp.png", dpi=200)
```

## Expected Output

matrix_exp

The figure shows the results for the different methods. Method 1 is the expected result, Methods 2 and 3 have very weird discontinuities.

#### Current Version:
cmdstan 2.27.0

yizhang-yiz:
## Summary
Fix #2529 by re-implementing the `prim` version of the function. 

## Tests

- test `matrix_exp_multiply_handle` components that calculate the taylor expansion order and helper function  approximating matrix power norm.
- test matrices from #2529.

## Side Effects

n/a

## Release notes

Bugfix: `matrix_exp_multiply` function's accuracy issue.

## Checklist

- [x] Copyright holder: Metrum Research Group.


yizhang-yiz:
I'm not sure how to interpret the failure of expression test (tests7_test.cpp). @rok-cesnovar  can you take a look?

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -14,26 +16,60 @@ namespace math {
...
-  const std::vector<double> theta_m_double_precision{
-      2.4e-3, 1.4e-1, 6.4e-1, 1.4e0, 2.4e0, 3.5e0,
-      4.7e0,  6.0e0,  7.2e0,  8.5e0, 9.9e0};
-
-  double l1norm(const Eigen::MatrixXd& m) {
-    return m.colwise().lpNorm<1>().maxCoeff();
-  }
+  static constexpr double tol = 1.0e-16;
+
+  //
```
rok-cesnovar, line 30: Remove the stray comment.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -54,39 +90,66 @@ class matrix_exp_action_handler {
...
+    int m, s;
+    set_approx_order(A, b, t, m, s);
+
+    double eta = exp(t * mu / s);
+
+    const auto& b_eval = b.eval();
+    Eigen::MatrixXd f = b_eval;
+    Eigen::MatrixXd bi = b_eval;
+
+#define MAT_OP_INF_NORM(X) X.cwiseAbs().rowwise().sum().maxCoeff()
```
rok-cesnovar, line 102: Not a huge fan of this #define and #undef approach. Given that its used only 3 times, can we just use it everywhere?

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -54,39 +90,66 @@ class matrix_exp_action_handler {
       A(i, i) -= mu;
```
rok-cesnovar, line 90: [optional]

You aren't changing these lines in this PR, so this is optional, but I guess these lines could just be:

```cpp
A.diagonal().array() -= mu;
```

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -54,39 +90,66 @@ class matrix_exp_action_handler {
...
+   *
+   * @param mat matrix
+   * @param m power
+   */
+  template <typename EigMat1, require_all_eigen_t<EigMat1>* = nullptr,
+            require_all_st_same<double, EigMat1>* = nullptr>
+  double mat_power_1_norm(const EigMat1& mat, int m) {
+    if ((mat.array() > 0.0).all()) {
+      Eigen::VectorXd e = Eigen::VectorXd::Constant(mat.rows(), 1.0);
+      for (auto j = 0; j < m; ++j) {
```
rok-cesnovar, line 146:
```suggestion
      for (int j = 0; j < m; ++j) {
```

rok-cesnovar (commented):
I honestly have zero knowledge of these algorithms, so cant review the changes in the sense of correctness, here are just a few general code-related comments.

stan/math/prim/fun/matrix_exp_action_handler.hpp
yizhang-yiz: removed

stan/math/prim/fun/matrix_exp_action_handler.hpp
yizhang-yiz: replaced with function

stan/math/prim/fun/matrix_exp_action_handler.hpp
yizhang-yiz: changed. thanks

stan/math/prim/fun/matrix_exp_action_handler.hpp
yizhang-yiz: changed.

yizhang-yiz:
I was able to reproduce the expression test failure on linux but not on mac. @rok-cesnovar can you take another look? Thanks.

rok-cesnovar:
Will check tomorrow.

rok-cesnovar:
I am unable to replicate locally (even with clang++-6.0) on Ubuntu. 

My next guess would be that the instance runs out of memory compiling 15 tests in parallel or something along these lines. Not sure why that would happen after this change, but who knows. Decreasing the parallel `-j` setting in the tests should be simple to try.

We are in the middle of a move of our testing infrastructure, so there will be a bit of downtime or a period of unrelated issues popping up, so testing on the actual instances will be a bit difficult in the next few day, but will make sure to test this as soon we are back up.

yizhang-yiz:
@rok-cesnovar thanks. We can continue after the migration.

rok-cesnovar:
Getting back to this finally. 

- For the GHA error with Windows tests, I think what we need to do is https://github.com/stan-dev/math/pull/2630 
    
    I would prefer closing https://github.com/stan-dev/math/issues/2489 and bumping the minimum g++, but that seems unlikely to happen very soon (see thread).

- For the Jenkins tests, I am going to debug on this PR. I am going to force-push a bit. I apologize in advance for the spam e-mails and force pushing, but I just don't see another way of getting to the bottom of this, seeing was no one can reproduce this one locally.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -14,26 +16,59 @@ namespace math {
...
+ * See also:
+ * https://www.mathworks.com/matlabcentral/fileexchange/29576-matrix-exponential-times-a-vector
  *
  * Calculates exp(mat*t)*b, where mat & b are matrices,
  * and t is double.
  */
 class matrix_exp_action_handler {
   static constexpr int p_max = 8;
   static constexpr int m_max = 55;
-  static constexpr double tol = 1.1e-16;
```
rok-cesnovar, line 24: So I have established the following.
 - with `p_max` and `m_max` defined as `static constexpr int` as it is currently on develop, the expression tests fail with `undefined reference to stan::math::matrix_exp_action_handler::p_max`
    See https://jenkins.mc-stan.org/blue/organizations/jenkins/Math%20Pipeline/detail/PR-2619/22/pipeline/
- with `p_max` and `m_max` defined as `int` it passes the tests

It fails in the set_approx_order function below. This all seems very bizarre to me. Especially given that no one can replicate this locally and it might be an issue with the specific compiler.

cc @SteveBronder would you know what would be the cause of these issues?

rok-cesnovar: @SteveBronder would you know what is up in this case? Thanks!

syclik: I think I found it: https://stackoverflow.com/a/8016853

I'm patching it on this branch.

syclik:
While I'm looking at this, I'm going to clean up some of the code for readability. I'll leave some comments on the PR so you can see what I'm thinking about before making the changes.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -104,26 +169,68 @@ class matrix_exp_action_handler {
...
-    if (l1norm(mat) < tol || t < tol) {
+  template <typename EigMat1, typename EigMat2,
+            require_all_eigen_t<EigMat1, EigMat2>* = nullptr,
+            require_all_st_same<double, EigMat1, EigMat2>* = nullptr>
+  inline void set_approx_order(const EigMat1& mat, const EigMat2& b,
+                               const double& t, int& m, int& s) {
+    // L1 norm
+    double normA = mat.colwise().template lpNorm<1>().maxCoeff();
+
+    if (normA < tol || t < tol) {
```
syclik, line 180: I almost want to split the condition. If `t < tol`, then we don't have to take a norm.

I might do that.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -104,26 +169,68 @@ class matrix_exp_action_handler {
...
       m = 0;
       s = 1;
```
syclik, line 182: I'd want to eagerly return here so it's clear that there's no more that happens below it. 

That will unnest the lower `else` statement and it'll be slightly easier to read.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -104,26 +169,68 @@ class matrix_exp_action_handler {
...
     } else {
-      m = m_max;
+      const std::vector<double>& theta = theta_m_double;
+      Eigen::VectorXd alpha(p_max - 1);
+
+      if (normA
```
syclik, line 187: for efficiency, I'd move the `/ (m_max * b.cols())` to the left hand side by `* (m_max * b.cols())`.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -104,26 +169,68 @@ class matrix_exp_action_handler {
...
+    double normA = mat.colwise().template lpNorm<1>().maxCoeff();
+
+    if (normA < tol || t < tol) {
       m = 0;
       s = 1;
     } else {
-      m = m_max;
+      const std::vector<double>& theta = theta_m_double;
+      Eigen::VectorXd alpha(p_max - 1);
+
```
syclik, line 186: I would put whitespace above 185 and remove the whitespace in 186.

The output of the `if` statement here is `alpha`. We want to keep that declaration close together if possible for readability.

stan/math/prim/fun/matrix_exp_action_handler.hpp
```diff
@@ -104,26 +169,68 @@ class matrix_exp_action_handler {
...
-      s = (c < 1 ? 1 : c);
+
+      Eigen::MatrixXd c = stan::math::ceil(mt) * u.asDiagonal();
+      for (auto i = 0; i < c.size(); ++i) {
+        if (c(i) <= 1.e-16) {
+          c(i) = std::numeric_limits<double>::infinity();
+        }
+      }
+
+      int cost = c.colwise().minCoeff().minCoeff(&m);
```
syclik, line 221: Is this right? 

Do we need the first `.minCoeff()`? If so, why? It looks like we should be able to get by with the second one only?

syclik: It is required. Compiler fails to compile. I'll leave it at that.

syclik:
@yizhang-yiz, for what it's worth, it's implemented really cleanly relative to the paper. The naming convention follows and is close. It looks right to me.

syclik (dismissed):
The implementation looks good to me.

syclik (approved):
lgtm

yizhang-yiz:
@syclik I completely let this slip through. Thanks for fixing the PR!

syclik:
Looks like the "fix" I introduced causes it to have linking issues with multiple translation units.

From Jenkins https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2619/11/pipeline/214:
image

I believe this is real. I'll try to address it in a different way.

syclik:
I'm looking at the intent of the code and I think the optimization is too much effort relative to the difficulty in getting it working reliably across C++ versions.

To be specific, there are two major optimizations happening. The first is that the variable is a static class member variable. I believe the idea is that on multiple uses of the class these constants don't have to be reallocated with the class. This, in and of itself, is great, but it interacts with the second thing. 

The second is the use of `constexpr` and there are some differences in the C++ across the pre C++-17 / C++-17 boundary. This interacts with the `static` keyword in a way that requires the definition of the static member variable. And this is what is the last error posted -- multiple translation units can not have the same definition. We could define the variable in a central place like wherever we define the `int main()` function, but we're not really set up to do this in the math library well. 

So... I think we can simplify this a bit and lose this optimization. Note: this is for 3 primitives only. I doubt it'd be that much of a difference in practice.

stan/math/prim/fun/matrix_exp_action_handler.hpp
syclik: Update: multiple translation unit tests fail. Those check to make sure that we can link the math library within multiple translation units. It's clear that this approach fails for that test. I've updated the code to be simpler and not rely on static member variables.

syclik (approved):

PR #2635, opened 2021-12-29
doc trig identity for the tan derivative

spinkney:
This is super minor. I just happened to come across it and trig was a long time ago. I forgot about this identity so it's nice to show that's what's used for the derivative.


rok-cesnovar (approved):
Thanks @spinkney

Issue #2562, opened 2021-08-23
L1 and L2 norms (Euclidean and taxicab lengths)


bob-carpenter:
## Description

We should have L1 and L2 norm functions in the Stan language.  They have to be implemented in the math library first, which is what this PR is for.  We want the functions, but ideally we want custom derivatives. 

```
real norm1(vector x);
real norm1(row_vector x);

real norm2(vector x);
real norm2(row_vector x);
```

The definitions are:

```
norm1(x)  =def=  sum(abs(x))
norm2(x)  =def=  sqrt(sum(square(x)))
```

The derivatives are:

```
d/dx norm1(x) = signum(x)
d/dx norm2(x) = x / norm2(x)
```

* [ ] implement `norm1` and `norm2` in the Stan Math library with autodiff test framework tests
* [ ] replace other uses that calculate vector lengths with `norm2`
    * [ ] unit_vector constrain
    * [ ] ???

#### Current Version:
v4.1.0

yizhang-yiz:
@tvrsimhan any update on this issue? Thanks.

SteveBronder:
@tvrsimhan the stan math docs have some examples to get started with as well

https://mc-stan.org/math/getting_started.html

rok-cesnovar:
Ah sorry, was too quick to close. #2636 addressed the first point of the issue, the second one is still open. Reopening.

PR #2636, opened 2021-12-29
Feature: L1 and L2 norms

lyndond:
## Summary

- Adds L1- and L2-norm functionality (described in #2562)
- Includes autodiff test framework tests
- I would appreciate if someone could look over my `rev/.../norm1.hpp` and `rev/.../norm2.hpp` implementations. 
  - The derivatives are straightforward but I'm still new to the stan autodiff framework so I'm not confident I did them correctly
- @bob-carpenter mentioned: using these new functions, we should "replace other uses that calculate vector lengths with `norm2()`"
  - To do this I'll need some guidance on where this calculation might arise in the repo  

As noted in the issue:
```
norm1(x)  =def=  sum(abs(x))
norm2(x)  =def=  sqrt(sum(square(x)))
```

```
d/dx norm1(x) = signum(x)
d/dx norm2(x) = x / norm2(x)
```

## Tests

- Tests are pretty similar to `dot_self`
- Includes expected behaviour, as well as nan-handling

## Side Effects

n/a

## Release notes

n/a

## Checklist

- [x] Math issue #2562 
- [x] Copyright holder: Lyndon Duong

### DIFF
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -19,7 +19,7 @@ namespace math {
  * @return L1 norm of v.
  */
 template <typename Container, require_container_t<Container>* = nullptr,
-    require_not_st_var<Container>* = nullptr>
+          require_not_st_var<Container>* = nullptr>
 inline auto norm1(const Container& x) {
   return apply_vector_unary<ref_type_t<Container>>::reduce(
       to_ref(x), [](const auto& v) { return v.template lpNorm<1>(); });
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -4,8 +4,6 @@
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
-#include <cstddef>
-#include <vector>
 
 namespace stan {
 namespace math {
@@ -24,6 +22,7 @@ inline auto norm1(const Container& x) {
   return apply_vector_unary<ref_type_t<Container>>::reduce(
       to_ref(x), [](const auto& v) { return v.template lpNorm<1>(); });
 }
+
 }  // namespace math
 }  // namespace stan
 
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -16,9 +16,7 @@ namespace math {
  * @param v Vector.
  * @return L1 norm of v.
  */
-template <typename Container, require_container_t<Container>* = nullptr,
-          require_not_st_var<Container>* = nullptr,
-          require_not_st_fvar<Container>* = nullptr>
+template <typename Container, require_st_arithmetic<Container>* = nullptr>
 inline auto norm1(const Container& x) {
   return apply_vector_unary<ref_type_t<Container>>::reduce(
       to_ref(x), [](const auto& v) { return v.template lpNorm<1>(); });
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -18,14 +18,12 @@ namespace math {
  * @param v Vector.
  * @return L1 norm of v.
  */
-template <typename Container, require_st_arithmetic<Container>* = nullptr,
-          require_container_t<Container>* = nullptr>
+template <typename Container, require_container_t<Container>* = nullptr,
+    require_not_st_var<Container>* = nullptr>
 inline auto norm1(const Container& x) {
-  check_nonzero_size("norm1", "v", x);
   return apply_vector_unary<ref_type_t<Container>>::reduce(
       to_ref(x), [](const auto& v) { return v.template lpNorm<1>(); });
 }
-
 }  // namespace math
 }  // namespace stan
 
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -17,7 +17,8 @@ namespace math {
  * @return L1 norm of v.
  */
 template <typename Container, require_container_t<Container>* = nullptr,
-          require_not_st_var<Container>* = nullptr>
+          require_not_st_var<Container>* = nullptr,
+          require_not_st_fvar<Container>* = nullptr>
 inline auto norm1(const Container& x) {
   return apply_vector_unary<ref_type_t<Container>>::reduce(
       to_ref(x), [](const auto& v) { return v.template lpNorm<1>(); });
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -16,7 +16,8 @@ namespace math {
  * @param v Vector.
  * @return L1 norm of v.
  */
-template <typename Container, require_st_arithmetic<Container>* = nullptr>
+template <typename Container, require_st_arithmetic<Container>* = nullptr,
+          require_container_t<Container>* = nullptr>
 inline auto norm1(const Container& x) {
   return apply_vector_unary<ref_type_t<Container>>::reduce(
       to_ref(x), [](const auto& v) { return v.template lpNorm<1>(); });
--- a/stan/math/prim/fun/norm1.hpp
+++ b/stan/math/prim/fun/norm1.hpp
@@ -10,13 +10,6 @@
 namespace stan {
 namespace math {
 
-inline double norm1(const std::vector<double>& x) {
-  double norm = 0.0;
-  for (double i : x) {
-    norm += std::abs(i);
-  }
-  return norm;
-}
 
 /**
  * Returns L1 norm of a vector. For vectors that equals the
@@ -30,6 +23,12 @@ template <typename T, require_eigen_t<T>* = nullptr,
           require_not_eigen_vt<is_var, T>* = nullptr>
 inline auto norm1(const T& v) {
   return v.template lpNorm<1>();
+}
+
+template <typename T>
+inline T norm1(const std::vector<T>& x) {
+  Eigen::Map<const Eigen::Matrix<T, -1, 1>> v(x.data(), x.size());
+  return norm1(v);
 }
 
 }  // namespace math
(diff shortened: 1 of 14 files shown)


andrjohns:
Thanks for the PR @lyndond and welcome to Stan Math! Just a couple of quick changes and then I can do a full review for you. For your implementations, you don't need separate specialisations for `std::vector` and Eigen types. The Math library has a helper framework `apply_vector_unary` which allows you to only specify the code for Eigen types, and it will automatically Map `std::vector` inputs.

Additionally, using the `auto` return type with Eigen expressions can have unintended side effects, as covered in [this help topic](https://eigen.tuxfamily.org/dox/TopicPitfalls.html#title3) and a practical example in #1775. To avoid these, the Math library has an additional framework, `Holder`, to correctly manage returning Eigen expressions.

You can see an example of applying both of these approaches in the [`log_softmax`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/log_softmax.hpp) header. Can you make those changes and then I'll do a full review?

Thanks!

lyndond:
Thanks @andrjohns. I just made a few edits and pushed before reading your comment so I'll make changes according to your suggestions.

lyndond:
@andrjohns I followed your tips and reduced some redundancy using `apply_vector_unary::reduce`.

I noticed that the Jenkins test is failing but can't decrypt the error message. Would you be able to help me with that?

andrjohns:
As far as I can see the Jenkins run hasn't failed, there just isn't a machine available to run the tests at the moment. But it's in the queue and will be run once a machine frees up

stan/math/rev/fun/norm1.hpp
```diff
@@ -0,0 +1,60 @@
...
+    arena_v.coeffRef(i) = v_ref.coeffRef(i);
+    res += abs(arena_v.coeffRef(i).val());
+  }
+
+  reverse_pass_callback([res, arena_v]() mutable {
+    arena_v.adj().array() += res.adj() * sign(arena_v.val().array());
+  });
+
+  return res;
+}
```
andrjohns, line 38: You don't need to `to_ref` inputs before passing them to `arena_`, as that is handled implicitly. Additionally, you can just call `norm1(arena_v.val())`, rather than re-creating the value manually:

```suggestion
template <typename T, require_eigen_vector_vt<is_var, T>* = nullptr>
inline var norm1(const T& v) {
  arena_t<T> arena_v = v;
  var res = norm1(arena_v.val());

  reverse_pass_callback([res, arena_v]() mutable {
    arena_v.adj().array() += res.adj() * sign(arena_v.val().array());
  });

  return res;
```

stan/math/rev/fun/norm1.hpp
```diff
@@ -0,0 +1,60 @@
...
+template <typename T, require_var_matrix_t<T>* = nullptr>
+inline var norm1(const T& v) {
+  var res = v.val().array().abs().sum();
+
+  reverse_pass_callback([res, v]() mutable {
+    v.adj().array() += res.adj() * sign(v.val().array());
+  });
+
+  return res;
+}
```
andrjohns, line 56: You can use `make_callback_vari` here to markedly simplify the code:

```suggestion
template <typename T, require_var_matrix_t<T>* = nullptr>
inline var norm1(const T& v) {
  return make_callback_vari(norm1(v.val()), [v](const auto& res) mutable {
    v.adj().array() += res.adj() * sign(v.val().array());
  });
}
```

stan/math/rev/fun/norm2.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <stan/math/rev/core/typedefs.hpp>
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/prim/fun/sign.hpp>
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/**
```
andrjohns, line 15: You can use the same recommendations that I made for the `norm1` headers here as well, should simplify the code a little

stan/math/prim/fun/norm1.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef STAN_MATH_PRIM_FUN_NORM1_HPP
+#define STAN_MATH_PRIM_FUN_NORM1_HPP
+
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <cstddef>
+#include <vector>
```
andrjohns, line 8: I don't think these headers are needed here, right?

andrjohns (changes requested):
Thanks for this PR @lyndond, you've done a great job of picking up our frameworks and coding styles!

I've made some requests in this PR. Additionally, you'll also need to add a `fwd` specialisation for these functions. The `fwd` specialisation is used for forward-mode autodiff where the values and gradients are computed in the same pass, we use the variable type `fvar<T>` for these, where `T` can be a `double`, `var`, or another `fvar<T>`. There's an example of how to specify the `fwd` specialisation for these kinds of functions in the [`log_sum_exp`](https://github.com/stan-dev/math/blob/a43562ea29ef1bb892cb7942787d682f002dfc7c/stan/math/fwd/fun/log_sum_exp.hpp#L54) header, and feel free to reach out if you need any clarifications or help.

stan/math/prim/fun/norm1.hpp
lyndond: Good catch. Changed in `norm2` as well

stan/math/rev/fun/norm2.hpp
lyndond: Done.

stan/math/rev/fun/norm1.hpp
lyndond: Ahh much cleaner! Changed my implementations to match.

stan/math/rev/fun/norm1.hpp
lyndond: Interesting. I made this change to `norm1` and `norm2`, but I still have to wrap my head around it. `vari`, `var`, `fvar`, etc. aren't fully dissociated in my head yet.

lyndond:
Thanks for the review so far @andrjohns.

I made changes according to your suggestions and will work on the corresponding `fwd` implementations next. Happy New Year in the meantime :)

bob-carpenter:
@andrjohns:  Can't we use the primitive implementation for forward mode?  There's not nearly as much efficiency savings in custom autodiff for forward mode because the arithmetic doesn't incur virtual function call overhead (pointer chasing) like in reverse mode.  I was thinking there could be a way to use expression templates to select primitive or forward-mode (not reverse mode).

bob-carpenter:
Oh, and @lyndond, if it's passing mix tests, the forward mode's working, as those test everything but the double-based implementations.

lyndond:
@bob-carpenter, great. The 'rev' and templated 'prim' functions were indeed passing all 'mix' tests before I explicitly implemented the `fwd` functions. 

It's unclear to me whether or not the new `fwd` functions are even being tested in the `mix` tests, since tests ran suspiciously fast after each time I edited the `fwd` files.

lyndond:
Turns out my header guards were wrong in my forward implementations so the `mix` tests were still defaulting to using my templated `prim` functions instead of the new `fwd` functions. I've now fixed the header guards but am iffy about the template type trait juggling I had to do in order to resolve function ambiguities for the `mix` tests to pass.

I'm paranoid the tests are still using the `prim` implementations so I'd greatly appreciate if someone could confirm my template arguments in the `prim` and `fwd` functions make sense!

andrjohns:
> @andrjohns: Can't we use the primitive implementation for forward mode? There's not nearly as much efficiency savings in custom autodiff for forward mode because the arithmetic doesn't incur virtual function call overhead (pointer chasing) like in reverse mode. I was thinking there could be a way to use expression templates to select primitive or forward-mode (not reverse mode).

Fair point, I was just erring on the side of consistency between all implementations

stan/math/rev/fun/norm1.hpp
andrjohns: It is a bit tricky! `var` types contain a pointer to a `vari` which contains the actual value and gradient. The `var` type contains some additional machinery used for calculating and propagating the gradients during the autodiff reverse pass.

`fvar` types aren't currently used in Stan, but they are used for forward-mode only autodiff

stan/math/prim/fun/norm1.hpp
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Returns L1 norm of a vector. For vectors that equals the
+ * sum of magnitudes of its individual elements.
+ *
+ * @tparam T type of the vector (must be derived from \c Eigen::MatrixBase)
+ * @param v Vector.
+ * @return L1 norm of v.
+ */
+template <typename Container, require_container_t<Container>* = nullptr,
+          require_not_st_var<Container>* = nullptr,
```
andrjohns, line 20: These templates can be a little stricter by specifying that they're only for arithmetic types, see the [`log_softmax` header for an example](https://github.com/stan-dev/math/blob/83dbd82c5b431d22c6e5899b2dfc5830a4cf95ed/stan/math/prim/fun/log_softmax.hpp#L42)

andrjohns (changes requested):
One last request and then this is all good to go!

Thanks for sticking through these requests!

stan/math/prim/fun/norm1.hpp
lyndond: Cool, I changed the templates to include `require_st_arithmetic<>` to match the template of `log_softmax`. 

Does `require_st_arithmetic<>` do the job of both `require_not_st_var<>` and `require_not_st_fvar<>`? These are what I used to resolve ambiguity between `prim` vs `rev` and `prim` vs `fwd`, respectively.

andrjohns: Yep, it will restrict to only matching inputs that satisfy `std::is_arithmetic`

andrjohns (approved):
This all looks good to me! Feel free to merge once tests pass.

Thanks for your contribution and congrats on your first Math PR!

rok-cesnovar:
Congrats @lyndond on your first merged PR in Stan Math :tada:. Thanks!

@andrjohns I think someone with elevated permission will have to merge.

andrjohns:
Oh right, will do that now

lyndond:
Great! Thanks @rok-cesnovar @bob-carpenter & @andrjohns. I appreciate the welcoming support and look forward to contributing more in the future.

yizhang-yiz:
@andrjohns where can I find unit tests for `rev`? I could only find `prim` & `mix` but no tests for `var` type.

yizhang-yiz:
I missed the `expect_ad` line. Thanks @rok-cesnovar

PR #2648, opened 2022-01-09
Replace sqrt(dot_self()) with norm2

lyndond:
## Summary

I'm hoping this one won't be too controversial. This PR addresses the remaining TODO in #2562.
PR #2636 introduced specialized prim/fwd/rev functions for L1 and L2 norms. This PR tidies some instances in the codebase where `r = sqrt(dot_self(x));` can now be replaced with `r = norm2(x);`. The main ones I found were:

- `unit_vector_constrain` in prim/fwd/rev

If there are any instances I may have missed, please let me know.

## Tests

No new tests. Existing tests pass.

## Side Effects

N/A

## Release notes

## Checklist

- [x] Math issue #2562 

- [x] Copyright holder: Lyndon Duong

### DIFF
--- a/stan/math/fwd/fun/unit_vector_constrain.hpp
+++ b/stan/math/fwd/fun/unit_vector_constrain.hpp
@@ -23,8 +23,8 @@ inline auto unit_vector_constrain(const EigMat& y) {
   plain_type_t<EigMat> unit_vector_y(y_val.size());
   unit_vector_y.val() = unit_vector_constrain(y_val);
 
-  eig_partial squared_norm = dot_self(y_val);
-  eig_partial norm = sqrt(squared_norm);
+  eig_partial norm = norm2(y_val);
+  eig_partial squared_norm = norm*norm;
   eig_partial inv_norm = inv(norm);
   Eigen::Matrix<eig_partial, Eigen::Dynamic, Eigen::Dynamic> J
       = divide(tcrossprod(y_val), -norm * squared_norm);
--- a/stan/math/fwd/fun/unit_vector_constrain.hpp
+++ b/stan/math/fwd/fun/unit_vector_constrain.hpp
@@ -24,7 +24,7 @@ inline auto unit_vector_constrain(const EigMat& y) {
   unit_vector_y.val() = unit_vector_constrain(y_val);
 
   eig_partial norm = norm2(y_val);
-  eig_partial squared_norm = norm*norm;
+  eig_partial squared_norm = norm * norm;
   eig_partial inv_norm = inv(norm);
   Eigen::Matrix<eig_partial, Eigen::Dynamic, Eigen::Dynamic> J
       = divide(tcrossprod(y_val), -norm * squared_norm);
--- a/stan/math/prim/fun/unit_vector_constrain.hpp
+++ b/stan/math/prim/fun/unit_vector_constrain.hpp
@@ -3,8 +3,6 @@
 
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
-#include <stan/math/prim/fun/dot_self.hpp>
-#include <stan/math/prim/fun/sqrt.hpp>
 #include <cmath>
 
 namespace stan {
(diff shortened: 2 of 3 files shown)


andrjohns:
Thanks for the PR! I'm not sure whether this change is actually necessary though, as it looks this essentially just replacing `sqrt(dot_self())` with calls to `square(norm2())`. So the amount of operations/complexity is unchanged

lyndond:
That's fair -- I was thinking that too as I was making the changes. Perhaps there was something else @bob-carpenter had in mind in the issue #2562 .

bob-carpenter:
@andrjohns:  We want to replace `sqrt(dot_self(x))` with `norm2(x)`, not with `square(norm2(x))`, because `norm2(x) = sqrt(dot_self(x))`.

lyndond:
Right. I replaced instances of `sqrt(dot_self())` with `norm2()`. There were a few instances where the squared norm is required (e.g. line 46 in ` stan/math/fwd/fun/unit_vector_constrain.hpp`), and in those cases I left `dot_self()`as is.

stan/math/fwd/fun/unit_vector_constrain.hpp
```diff
@@ -23,8 +23,8 @@ inline auto unit_vector_constrain(const EigMat& y) {
   plain_type_t<EigMat> unit_vector_y(y_val.size());
   unit_vector_y.val() = unit_vector_constrain(y_val);
 
-  eig_partial squared_norm = dot_self(y_val);
-  eig_partial norm = sqrt(squared_norm);
+  eig_partial norm = norm2(y_val);
+  eig_partial squared_norm = norm * norm;
```
bob-carpenter, line 27: These are going to be very similar operationally.  I think the original code may require fewer operations and be more stable arithmetically, because the new one has the sqrt inside norm2, and then you have to the do the additional squaring.  So unless there's something I'm not seeing, I think this should be switched back.  If for some reason you do want to keep the new code, line 27 should be `square(norm)`.

stan/math/rev/fun/unit_vector_constrain.hpp
```diff
@@ -34,7 +34,7 @@ inline auto unit_vector_constrain(const T& y) {
   arena_t<T> arena_y = y;
   arena_t<promote_scalar_t<double, T>> arena_y_val = arena_y.val();
 
-  const double r = arena_y_val.norm();
+  const double r = norm2(arena_y_val);
```
bob-carpenter, line 37: For `double` arguments, it's probably better to use the built-in Eigen functions as they're more highly optimized for this.  Of course, `norm2` should delegate to this `.norm()` for values in the vector case.

stan/math/prim/fun/unit_vector_constrain.hpp
```diff
@@ -47,13 +44,12 @@ inline plain_type_t<T> unit_vector_constrain(const T& y) {
...
-  using std::sqrt;
   check_nonzero_size("unit_vector_constrain", "y", y);
   auto&& y_ref = to_ref(y);
-  value_type_t<T1> SN = dot_self(y_ref);
-  check_positive_finite("unit_vector_constrain", "norm", SN);
-  lp -= 0.5 * SN;
-  return y_ref.array() / sqrt(SN);
+  value_type_t<T1> r = norm2(y_ref);
+  check_positive_finite("unit_vector_constrain", "norm", r);
+  lp -= 0.5 * (r * r);
```
bob-carpenter, line 51: This looks like it's also just paying that squaring cost again here on line 51, so I think the original code may again be faster here.

What we really want here is a specialized function for unit_vector_constrain that does all of this with analytic derivatives for lp and the return type.

bob-carpenter (changes requested):
Hmm.  These changes are OK semantically, but I think the original versions are slightly faster in all cases because it's not just the norm2 calculation, we also need its square.  

If you want to work on speed here, what we could use is the autodiff implemented analytically here like for any other special function.  The twist is that we also need to set `lp` (right now it just does `lp +=`, but we should set it to a new `var` that holds a pointer to `lp` and to `y`), and also define the version that doesn't increment `lp`.  

What we'd really like instead of this replacement is analytic gradients for the constraint functions.   We should be able to define this just like any other autodiffable function but have it also set `lp` to a new value.

lyndond:
Thanks for the review @bob-carpenter. It looks like this is a bit more nuanced than initially anticipated. I think I'll close the PR for now until we settle on something that requires more significant change.

Issue #20, opened 2015-07-06
Add FFT to the math library. Use Eigen's implementation.


syclik:
_From @betanalpha on August 16, 2013 14:28_

Expose Eigen's FFT into the modeling language:

vector fft(vector x)

_Copied from original issue: stan-dev/stan#179_

syclik:
_From @betanalpha on August 16, 2013 14:40_

Whoops, forgot about the imaginary components, so I guess something like

void fft(vector x, vector r, vector i);

syclik:
@betanalpha, think we should still have this as part of the language?

syclik:
_From @betanalpha on May 22, 2014 20:47_

There are lots of physics use cases, if only for preprocessing in
transformed data.  The holdup was whether we wanted to (a)
find a way to limit functions to transformed data or (b) figure
out the analytic gradient for the FFT.

syclik:
_From @bob-carpenter on May 23, 2014 13:35_

And a bigger issue is data-only arguments.  Because in some
cases, we might only want to implement some gradients for a function.

Definitely not impossible, but a fair bit of work in the parser.

syclik:
_From @mbrubake on May 23, 2014 15:35_

So I totally think we should add FFT at some point as it's very useful for some models.  The gradients are easy (the FFT and Inverse FFT are both linear operators) so we don't need to worry about data-only functions.

The problem with adding FFT is that we need some way to handle complex numbers.  There are basically two options that aren't complete hacks:

 1. Actually implement a complex type which works just like real does.
 2. Add support for functions which return more than one parameter.  Then, the FFT would simply take and return the real and imaginary coefficients separately.

Both of these options could potentially turn into a fair amount of work.  That said, I vote for 2.  For one, I don't think complex numbers are going to be widely enough used to be worth the pain of implementing and supporting them.  On the other hand, allowing functions to return multiple values seems much more useful and may not be that hard if we make suitable use of boost tuples [http://www.boost.org/doc/libs/1_55_0/libs/tuple/doc/tuple_users_guide.html]

syclik:
_From @bob-carpenter on May 23, 2014 16:51_

Yes, I really want to add multiple outputs.   What they 
call lists in the R world, I think, though of course, I'd want
to type the ones in Stan.

I don't know what the size limits are on Boost's tuples, but we probably don't
need huge lists.

I also really like the Python syntax for literals and assignments,
which lets you do things like (a,b,c) to create a list out of a, b,
[...]

PR #2686, opened 2022-03-15
Feature/20 fft

bob-carpenter:
## Summary

This PR adds functions `fft` for 1D FFTS, `inv_fft` for 1D inverse FFTs, `fft2` for 2D FFTs, and `inv_fft2` for 2D inverse FFTs.

Implementation delegates to Eigen with appropriate templating and just autodiffs through the vector operations.

## Tests

For primitive tests, (a) tested values against known values from MATLAB/SciPy/R (they all agree), and (b) tested FFT(2) and inv_FFT(2) compose in both directions to produce an identity mapping.

For autodiff tests, use the test framework to test at all orders.  Functionals in there deal with extracting a univariate result.

There is no error handling to test because we just pass results to Eigen, which can handle arbitrary vectors and matrices.

## Side Effects

No.

## Release notes

* Add fast Fourier transform (FFT) implementations of 1D and 2D discrete Fourier transform and their inverses

## Checklist

- [x] Math issue #(issue number) #20

    - NO dependencies checks pass, (`make test-math-dependencies`) **THIS DOESN'T WORK** (see below)
    - docs build, (`make doxygen`) **COULD NOT GET DOXYGEN INSTALLED ON MY MAC---IT'S NOT "TRUSTED"**
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`) **THIS ALSO FAILS:**

- (n/a) the new changes are tested (no "changes", per se)

### TESTING FAILURES

```
~/github/stan-dev/cmdstan/stan/lib/stan_math (feature/20-fft)$ make test-math-dependencies
make: ./runChecks.py: No such file or directory
```

```
make cpplint
make: python: No such file or directory
make: *** [cpplint] Error 1
```

```
~/github/stan-dev/cmdstan/stan/lib/stan_math (feature/20-fft)$ make doxygen
mkdir -p doc/api
doxygen doxygen/doxygen.cfg
make: doxygen: No such file or directory
make: *** [doxygen] Error 1
```


SteveBronder:
It's failing locally for me as well (on ubuntu 20.04) 

running `./runTests.py ./test/unit/math/mix/fun -f fft` I get errors for calculating gradients with fvar types like the below.


```
[ RUN      ] mathMixFun.fft
./test/unit/math/expect_near_rel.hpp:77: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(-1.2516191104003521e-10, nan)
expect_near_rel; require items x1(0) = x2(0): gradient_fvar() grad

./test/unit/math/expect_near_rel.hpp:77: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(0.99999999996879563, nan)
expect_near_rel; require items x1(1) = x2(1): gradient_fvar() grad

./test/unit/math/expect_near_rel.hpp:77: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(-6.1656113812825225e-11, nan)
expect_near_rel; require items x1(2) = x2(2): gradient_fvar() grad

./test/unit/math/expect_near_rel.hpp:77: Failure
Value of: stan::math::is_nan(x1) && stan::math::is_nan(x2)
  Actual: false
Expected: true
expect_near_rel(0.99999999999304701, nan)
expect_near_rel; require items x1(3) = x2(3): gradient_fvar() grad

./test/unit/math/expect_near_rel.hpp:138: Failure
Failed

x1: 
[-1.2516e-10]
[          1]
[-6.1656e-11]
[          1]
x2: 
[nan]
[nan]
[nan]
[nan]
```

The `x1` and `x2` values are printed by adding `-DSTAN_TEST_PRINT_MATRIX_FAILURE` to `CXXFLAGS` in `make/local`. So it looks like some of the values are very small? Is this supposed to support forward mode autodiff?

bob-carpenter:
I'm a knucklehead.  The reason it was working for me is this line in my `make/local`:

```
CXXFLAGS += -DSTAN_MATH_TESTS_REV_ONLY
```

Is there a way to turn higher-order testing off without an environment variable.  I really don't know how to debug what's happening now because we're just autodiffing through Eigen's implementation.  I suppose the next step is to write our own derivatives, which should be too hard, since the FFT is conceptually just a matrix multiply.  There's a hint here:

https://github.com/locuslab/pytorch_fft/issues/4

and I can verify with code and/or local FFT experts.

SteveBronder:
I think you can just call `test_ad_v` directly right? I think we did that for other complex functions

SteveBronder:
>  I really don't know how to debug what's happening now because we're just autodiffing through Eigen's implementation

Yeah :-/ I think the only way to really do this would be to put print statements in Eigen's FFT code and call their fft stuff with fvar types. Do you think very small division could be happening somewhere?

bob-carpenter:
I'm going to try solving this by adding analytic gradients.  I should've done that from the get-go, it's just my complex and vector derivatives aren't what they should be.  @mbrubake's point is the key one, namely that d fft(x) = fft(dx).  Just need to translate that into an adjoint update for x in y = fft(x).   And presumably inv_fft works the same way as it's just a sign difference.

bob-carpenter:
@WardBrian found the definitions we need for analytic gradients:

https://github.com/FluxML/Zygote.jl/issues/204#issuecomment-498166243

Not sure if we're going to need the scaling as every one of these packages seems to scale slightly differently.

rok-cesnovar:
Tests are now passing, any volunteers for reviewing this? @SteveBronder maybe?

bob-carpenter:
It should be very simple.  I'm just calling the FFT functions directly and autodiffing through them now.  My plan is to next get analytic derivatives working.

The biggest call out is that I modified the testing framework so that if you provide +infinity tolerances, tests are skipped.  I then built a custom default tolerances object that turns off all of the tests other than reverse mode.

test/unit/math/prim/fun/fft_test.cpp
```diff
@@ -0,0 +1,208 @@
...
+  cv_t x(3);
+  x << c_t(1, -2), c_t(-3, 5), c_t(-7, 11);
+  cv_t y = fft(x);
+  EXPECT_EQ(3, y.size());
+  EXPECT_NEAR(real(y[0]), -9, 1e-6);
+  EXPECT_NEAR(imag(y[0]), 14, 1e-6);
+  EXPECT_NEAR(real(y[1]), 0.80384758, 1e-6);
+  EXPECT_NEAR(imag(y[1]), -13.46410162, 1e-6);
+  EXPECT_NEAR(real(y[2]), 11.19615242, 1e-6);
+  EXPECT_NEAR(imag(y[2]), -6.53589838, 1e-6);
```
SteveBronder, line 36: Is `EXPECT_NEAR` needed because of differences in the python and eigen versions of fft? This is essentially just testing Eigen's impl of fft right? imo might be better to just use values from Eigen's for the test then this test scheme just becomes a canary for when we update eigen versions to see if any changes happened.

test/unit/math/prim/fun/fft_test.cpp
```diff
@@ -0,0 +1,208 @@
...
+  EXPECT_NEAR(imag(yb[0]), 2 * 14, 1e-6);
+  EXPECT_NEAR(real(yb[1]), 2 * 0.80384758, 1e-6);
+  EXPECT_NEAR(imag(yb[1]), 2 * -13.46410162, 1e-6);
+  EXPECT_NEAR(real(yb[2]), 2 * 11.19615242, 1e-6);
+  EXPECT_NEAR(imag(yb[2]), 2 * -6.53589838, 1e-6);
+}
+
+TEST(primFun, inv_fft) {
+  typedef std::complex<double> c_t;
+  typedef Eigen::Matrix<std::complex<double>, -1, 1> cv_t;
```
SteveBronder, line 50: Replace all `typedefs` with `using`

test/unit/math/prim/fun/fft_test.cpp
```diff
@@ -0,0 +1,208 @@
...
+  EXPECT_NEAR(-13.29326674, real(y33(1, 1)), 1e-1);
+  EXPECT_NEAR(20.88153533, imag(y33(1, 1)), 1e-1);
+  EXPECT_NEAR(-13.25262794, real(y33(1, 2)), 1e-1);
+  EXPECT_NEAR(15.82794549, imag(y33(1, 2)), 1e-1);
+  EXPECT_NEAR(4.160254038, real(y33(2, 0)), 1e-1);
+  EXPECT_NEAR(5.928718708, imag(y33(2, 0)), 1e-1);
+  EXPECT_NEAR(-11.34737206, real(y33(2, 1)), 1e-1);
+  EXPECT_NEAR(-7.72794549, imag(y33(2, 1)), 1e-1);
+  EXPECT_NEAR(4.89326674, real(y33(2, 2)), 1e-1);
+  EXPECT_NEAR(-1.98153533, imag(y33(2, 2)), 1e-1);
```
SteveBronder, line 134: [optional] I think this would be easier to read if the correct answers were put into a matrix and just looped over

test/unit/math/test_ad.hpp
```diff
@@ -333,12 +334,25 @@ template <typename G>
 void expect_ad_derivatives(const ad_tolerances& tols, const G& g,
                            const Eigen::VectorXd& x) {
   double gx = g(x);
-  test_gradient(tols, g, x, gx);
+  if (!tols.gradient_val_.is_inf() || !tols.gradient_grad_.is_inf()) {
```
SteveBronder, line 337: Is there a reason not to call `std::isinf(...)` directly?

test/unit/math/ad_tolerances.hpp
```diff
@@ -69,6 +69,34 @@ struct ad_tolerances {
...
+ *
+ * @param val_tol value relative tolerance (default `1e-8`)
+ * @param grad_tol gradient relative tolerance (default `1e-4`)
+ */
+ad_tolerances reverse_only_ad_tolerances(double val_tol = 1e-8,
+                                         double grad_tol = 1e-4) {
+  ad_tolerances tols;
+  tols.gradient_val_ = val_tol;
+  tols.gradient_grad_ = grad_tol;
+  double inf = std::numeric_limits<double>::infinity();
```
SteveBronder, line 84:
```suggestion
  constexpr double inf = std::numeric_limits<double>::infinity();
```

SteveBronder (changes requested):
Couple comments and a few things I'd like clarified. How attached are you to turning off reverse mode through setting the tolerances to infinity? Would it be more difficult than I imagine to have a `expect_ad_rev_only()` set of functions instead?

bob-carpenter:
> How attached are you to turning off reverse mode through setting the tolerances to infinity?

Not at all, it's just very flexible and lets us test whatever subset of derivatives we want.  The other thing to do would be to create an object with boolean flags mirroring the tolerances and plumb that through all the calls with the tolerance object.  Or maybe we just add those fields to the tolerance object?

test/unit/math/test_ad.hpp
bob-carpenter: Yes, because `tol.gradient_val_` isn't a primitive---it's a class, and the `is_inf` function is a method on that class.

stan/math/prim/fun/fft.hpp
```diff
@@ -0,0 +1,117 @@
...
+ * @param x complex time domain vector to transform
+ * @return discrete Fourier transform of `x`
+ */
+template <typename V, require_eigen_vector_vt<is_complex, V>* = nullptr>
+inline Eigen::Matrix<scalar_type_t<V>, -1, 1> fft(const V& x) {
+  Eigen::Matrix<scalar_type_t<V>, -1, 1> xv = x;
+  if (xv.size() <= 1)
+    return xv;
+  Eigen::FFT<typename scalar_type_t<V>::value_type> fft;
+  return fft.fwd(xv);
```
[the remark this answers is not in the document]
bob-carpenter: that worked, so I made the change here and below.

stan/math/prim/fun/fft.hpp
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * If the input is of size zero, the result is a size zero vector.
+ *
+ * @tparam V type of complex vector argument
+ * @param x complex time domain vector to transform
+ * @return discrete Fourier transform of `x`
+ */
+template <typename V, require_eigen_vector_vt<is_complex, V>* = nullptr>
+inline Eigen::Matrix<scalar_type_t<V>, -1, 1> fft(const V& x) {
+  Eigen::Matrix<scalar_type_t<V>, -1, 1> xv = x;
```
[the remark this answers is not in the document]
bob-carpenter: This fails as follows.

```
[==========] Running 4 tests from 2 test suites.
[----------] 3 tests from primFun
[ RUN      ] primFun.fft
[       OK ] primFun.fft (0 ms)
[ RUN      ] primFun.inv_fft
[       OK ] primFun.inv_fft (0 ms)
[ RUN      ] primFun.fft2
Assertion failed: ((!(RowsAtCompileTime!=Dynamic) || (rows==RowsAtCompileTime)) && (!(ColsAtCompileTime!=Dynamic) || (cols==ColsAtCompileTime)) && (!(RowsAtCompileTime==Dynamic && MaxRowsAtCompileTime!=Dynamic) || (rows<=MaxRowsAtCompileTime)) && (!(ColsAtCompileTime==Dynamic && MaxColsAtCompileTime!=Dynamic) || (cols<=MaxColsAtCompileTime)) && rows>=0 && cols>=0 && "Invalid sizes when resizing a matrix or array."), function resize, file PlainObjectBase.h, line 285.
test/unit/math/prim/fun/fft_test --gtest_output="xml:test/unit/math/prim/fun/fft_test.xml" failed
exit now (04/14/22 13:00:37 EDT)
```

stan/math/prim/fun/fft.hpp
```diff
@@ -0,0 +1,117 @@
...
+ * Return the two-dimensional discrete Fourier transform of the
+ * specified complex matrix.  The 2D discrete Fourier transform
+ * first runs the DFT on the each row, then on each column of the
+ * result.
+ *
+ * @tparam M type of complex matrix argument
+ * @param x complex time-domain matrix
+ * @return discrete 2D Fourier transform of `x`
+ */
+template <typename M, require_eigen_vt<is_complex, M>* = nullptr>
```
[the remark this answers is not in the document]
bob-carpenter: that works, so I made the update

stan/math/prim/fun/fft.hpp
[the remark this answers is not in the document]
SteveBronder: I think it's fine to just do a hard assignment in this case because fft does not play well with references

test/unit/math/ad_tolerances.hpp
bob-carpenter: fixed.

test/unit/math/prim/fun/fft_test.cpp
bob-carpenter: fixed as suggested.

test/unit/math/prim/fun/fft_test.cpp
bob-carpenter: They need floating-point comparison.  I replaced explicit `EXPECT_NEAR` with implicit `EXPECT_FLOAT_EQ`, and it all still works, so I'll leave that.

test/unit/math/prim/fun/fft_test.cpp
bob-carpenter: done, though I don't know why.

test/unit/math/ad_tolerances.hpp
```diff
@@ -69,6 +69,34 @@ struct ad_tolerances {
...
 
+/**
+ * Return tolerances that are infinite other than for the value
+ * tolerance and gradient tolerance for reverse mode.
+ *
+ * @param val_tol value relative tolerance (default `1e-8`)
+ * @param grad_tol gradient relative tolerance (default `1e-4`)
+ */
+ad_tolerances reverse_only_ad_tolerances(double val_tol = 1e-8,
+                                         double grad_tol = 1e-4) {
```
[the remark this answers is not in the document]
bob-carpenter: given that this is only in testing, I'm not going to worry about efficiency on something involving a dozen assignments per autodiff test.

test/unit/math/mix/fun/fft_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+  stan::test::ad_tolerances tols = stan::test::reverse_only_ad_tolerances();
+  int i = 0;
+  int m = 0;
+  auto g = [&](const auto& x) {
+    using stan::math::fft;
+    return i ? real(fft(x)(m)) : imag(fft(x)(m));
+  };
+  for (int i = 0; i < 2; ++i)
+    for (int m = 0; m < x.rows(); ++m)
+      stan::test::expect_ad(tols, g, x);
```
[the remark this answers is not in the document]
bob-carpenter: Good catch---I introduced local variables that masked the closure.  And the autodiff framework can handle the complex return.  I fixed both problems, so it's much cleaner now.

test/unit/math/mix/fun/fft_test.cpp
```diff
@@ -0,0 +1,185 @@
+#include <test/unit/math/test_ad.hpp>
+
+void expect_fft(const Eigen::VectorXcd& x) {
+  stan::test::ad_tolerances tols = stan::test::reverse_only_ad_tolerances();
+  int i = 0;
+  int m = 0;
+  auto g = [&](const auto& x) {
+    using stan::math::fft;
+    return i ? real(fft(x)(m)) : imag(fft(x)(m));
```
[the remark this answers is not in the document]
bob-carpenter: The fix used the framework.  It was just that I was hours into just trying to get the tests to compile so I wrote them as simply as possible.  Then I forgot to go back and generalize on the way out.  I am really struggling with compilation with all the templated traits everywhere.

test/unit/math/mix/fun/fft_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+  };
+  for (int i = 0; i < 2; ++i)
+    for (int m = 0; m < x.rows(); ++m)
+      stan::test::expect_ad(tols, g, x);
+}
+
+TEST(mathMixFun, fft) {
+  typedef Eigen::VectorXcd cvec_t;
```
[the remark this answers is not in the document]
bob-carpenter: fixed, but I don't know why the change was requested in the first place

test/unit/math/mix/fun/fft_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+  int i = 0;
+  int m = 0;
+  auto g = [&](const auto& x) {
+    using stan::math::inv_fft;
+    return i ? real(inv_fft(x)(m)) : imag(inv_fft(x)(m));
+  };
+  for (int i = 0; i < 2; ++i)
+    for (int m = 0; m < x.rows(); ++m)
+      stan::test::expect_ad(tols, g, x);
+}
```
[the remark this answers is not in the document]
bob-carpenter: fixed same way.  there are only two test functions like this in the file.

bob-carpenter:
@SteveBronder:  I resolved all 14 requested changes (see resolution notes above).  So this should be ready to review again if it passes.  

P.S.  I would like to get this merged in a working form before trying to plumb in the analytic gradients, which is going to require upgrading all the template traits.  But if you'd rather try to get the analytic gradients working in the first pass, I can tell you what they are in adjoint-Jacobian form, but I don't know how to implement without the upgraded arena, etc.

bob-carpenter:
@SteveBronder I think this is ready for another review.  I think I addressed all of your comments.

SteveBronder (approved):
Good!

bob-carpenter:
Not sure why this kicked off testing again as I didn't change anything. @SteveBronder --- this is ready to check that I made all the requested changes.  Thanks!

rok-cesnovar:
@bob-carpenter, this PR was approved - @SteveBronder has just let you merge it in yourself.

PR #2703, opened 2022-04-11
Speedup expressions tests

rok-cesnovar:
## Summary

The expressions tests on the Flatiron instances take about 10x+ more time than on our previous instances. I think this enormous difference is due to these tests being quite the memory hogs because the test files are quite big. Lets see if splitting them into more smaller files will help those instances

I would expect a 2-3x slowdown, not 10x.

## Tests

This is only touching CI.

## Side Effects

/


rok-cesnovar:
@serban-nicusor-toptal ready for a look. Gets us to below 2 hours from 11.

serban-nicusor-toptal (approved):
Uf that's such a speedup! LGTM, thanks!

PR #2706, opened 2022-04-15
Add tuple overload and cleanup code for promote_scalar 

SteveBronder:
## Summary

This adds an overload to `promote_scalar` for the compiler that allows the input to be a set of tuples and the promotion type to be a tuple of scalar types. So for the compiler impl of tuples we will be able to generate code such as 

```cpp
std::tuple<std::vector<var>, var> = 
  promote_scalar<std::tuple<var, var>(std::vector<double>{...}, double(2.0));
```

This also cleanups the code for `promote_scalar` a little bit. Instead of having to do some roundabouts using a static struct method for doing actually the promotions we called from `promote_scalar` we just have explicit free functions for the particular cases we need that can be called from one another.

## Tests

Tests are added for testing promotions of tuples with various inner types and scalar types. A test is also added for the new type trait `is_tuple<>` which just detects if a template is a tuple or not.

```
./runTests.py ./test/unit/math/ -f promote_scalar_test
./runTests.py ./test/unit/math/ -f is_tuple
```
## Side Effects

Nope

## Release notes

Add tuple overload and cleanup code for promote_scalar 

## Checklist

- [ ] Math issue #


WardBrian:
Just to check, does this do the right thing if it is called with a std::vector of tuples?

E.g, given a 
`std::vector<std::tuple<int,int>> vector;`
Does the call
`promote_scalar<std::tuple<double,` double>>(vector)`

Work like it does when the vector contains non-tuples (element wise?)

SteveBronder:
No it did not, good catch! But I just put in an update that fixes that with a test for it as well

WardBrian:
I mentioned this in the `assign` PR, but it is worth testing/making sure we support at least the following:

a tuple
An array of tuples
A tuple which contains an array
A tuple which contains a tuple
A tuple which contains an array of tuples

The kind of recursive logic needed to get all that right usually generalizes to even deeper nesting

SteveBronder:
I'll double check but looking at the test I believe I cover all of those

rok-cesnovar:
@SteveBronder is this ready for review? I have been tagged for this a few weeks ago - just want to make sure.

rok-cesnovar (approved):
This looks good to me. I went ahead and merge recent develop in as the branch was pretty old at this point. Merge away assuming the tests now pass.

Issue #2683, opened 2022-03-04
Make `arena_allocator` usable in std containers like `unordered_map` and others


cpp977:
## Description
The current `arena_allocator` cannot be used in c++ standard containers which are more complex than `std::vector` as for example `std::unordered_map`. There is a missing constructor for *rebinding* (see Additional info below):

```
template<typename U>
arena_allocator(const arena_alocator<U>& other);
```

## Example
The following does currently not compile:
```
#include <unordered_map>
#include "stan/math.hpp"

int main()
{
    std::unordered_map<int, int, std::hash<int>, std::equal_to<int>, stan::math::arena_allocator<std::pair<const int, int>>> m;
    m.insert(std::make_pair(1,1));
}
```
The compilation error is very lengthy because of missing concepts in c++17. But it basically tells that there is missing the respective constructor for `arena_allocator`.

## Expected Output
Compilation should be successful.

## Possible fix:
Add this to arena_allocator:
```
  arena_allocator() = default;

  arena_allocator(const arena_allocator& rhs) = default;

  template <class U>
  arena_allocator(const arena_allocator<U>& rhs) {}
```

## Additional info
See this SO discussion on custom allocators for std containers: 
https://stackoverflow.com/questions/64790174/stdunordered-map-with-custom-allocator-local-allocator-does-not-compile

#### Current Version:
v4.3.1

SteveBronder:
Yes this would be nice! I don't have the bandwidth to do this atm but if you want to make a PR I'm happy to review. I think we just need to define all of the members and constructors like [`std::allocator`](https://en.cppreference.com/w/cpp/memory/allocator)

SteveBronder:
Most of them should be simple since that allocator is just a wrapper around the stan's global stack allocator

cpp977:
Yes it is sinple because the class is just a wrapper. The only constructor which is missing is the one described above. With this added, the class can be used as a replacement for `std::allocator`. I would like to submit a PR and I already have a fork with the change, however so far I did not look into the test suite of stan. But I can do und then run the tests before submitting the PR. It is probably also worth to add a test for the std containers with the `arena_allocator`.

SteveBronder:
Yes our tests in `rev/core` should help you see what to do here

spinkney:
@rok-cesnovar should we close this?

rok-cesnovar:
Indeed, thanks!

PR #2708, opened 2022-04-19
Make `arena_allocator` usable for all std containers.

cpp977:
## Summary

The `arena_allocator` got  a new constructor to make it usable as a drop-in replacement for `std::allocator` in all std containers.

The additional constructor has the signature
```
  template <class U>
  arena_allocator(const arena_allocator<U>& rhs);
``` 
and the body is trivial because `arena_allocator` is a stateless wrapper class.

## Tests

Additional tests are added for the containers `std::unordered_map` and `std::set` in use with the `arena_allocator`.

These tests would have been failed before this PR.

## Side Effects

No.

## Release notes

Make `arena_allocator` usable for all std containers.

## Checklist

- [x] #2683 Math issue 
- [x] Copyright holder: Matthias Peschke

### DIFF
--- a/stan/math/rev/core/arena_allocator.hpp
+++ b/stan/math/rev/core/arena_allocator.hpp
@@ -10,10 +10,14 @@ namespace math {
  * std library compatible allocator that uses AD stack.
  * @tparam T type of scalar
  *
- * @warning The type T needs to be either trivially destructible or the dynamic allocations needs to be managed by the arena_allocator.
+ * @warning The type T needs to be either trivially destructible or the dynamic
+allocations needs to be managed by the arena_allocator.
  * For example this works: @code{.cpp}
-using my_matrix = std::vector<std::vector<double, stan::math::arena_allocator<double>>, stan::math::arena_allocator<std::vector<double, stan::math::arena_allocator<double>>>>;@endcode
- * 
+using my_matrix = std::vector<std::vector<double,
+stan::math::arena_allocator<double>>,
+stan::math::arena_allocator<std::vector<double,
+stan::math::arena_allocator<double>>>>;@endcode
+ *
  */
 template <typename T>
 struct arena_allocator {
--- a/stan/math/rev/core/arena_allocator.hpp
+++ b/stan/math/rev/core/arena_allocator.hpp
@@ -14,6 +14,13 @@ template <typename T>
 struct arena_allocator {
   using value_type = T;
 
+  arena_allocator() = default;
+
+  arena_allocator(const arena_allocator& rhs) = default;
+
+  template <class U>
+  arena_allocator(const arena_allocator<U>& rhs) {}
+
   /**
    * Allocates space for `n` items of type `T`.
    *
--- a/stan/math/rev/core/arena_allocator.hpp
+++ b/stan/math/rev/core/arena_allocator.hpp
@@ -9,6 +9,11 @@ namespace math {
 /**
  * std library compatible allocator that uses AD stack.
  * @tparam T type of scalar
+ *
+ * @warning The type T needs to be either trivially destructible or the dynamic allocations needs to be managed by the arena_allocator.
+ * For example this works: @code{.cpp}
+using my_matrix = std::vector<std::vector<double, stan::math::arena_allocator<double>>, stan::math::arena_allocator<std::vector<double, stan::math::arena_allocator<double>>>>;@endcode
+ * 
  */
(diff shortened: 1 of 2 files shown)


cpp977:
In principal the new constructor could be marked as `noexcept` but I did not saw this specifier in the codebase so I did not add it.

SteveBronder (approved):
lgtm! If you'd like to add the noexpect feel free as that is good. Should we look at supporting all of the std containers? I'm fine with unordered map and set, though without tests I'm not sure if other containers will work as well

https://en.cppreference.com/w/cpp/container

SteveBronder:
Oh actually there is one thing, this wasn't here before but I think should be if we are going to support all of the std containers. Should we have a check that the type we are making an arena allocator for is trivally destructible? If you had something like an `unordered_map<std::vector<double>, arena_allocator<std::vector<double>>>` I think bad things would happen

cpp977:
> lgtm! If you'd like to add the noexpect feel free as that is good. Should we look at supporting all of the std containers? I'm fine with unordered map and set, though without tests I'm not sure if other containers will work as well
> 
> https://en.cppreference.com/w/cpp/container

Yes it is probably worth to add all containers. I was just too lazy. I will add tests for them as well.

cpp977:
> Oh actually there is one thing, this wasn't here before but I think should be if we are going to support all of the std containers. Should we have a check that the type we are making an arena allocator for is trivally destructible? If you had something like an `unordered_map<std::vector<double>, arena_allocator<std::vector<double>>>` I think bad things would happen

Yes that should be checked but I guess if the type does have a nontrivial destructor, it will cause problems for `std::vector` as well, won't it?
```
std::vector<std::vector<int, std::allocator<int>>, arena_allocator<int>>
```
seems to be problematic to me.

SteveBronder:
While the above is bad, if the inner vector type also had an `arena_allocator<>` then it would be good such as the below

```
std::vector<std::vector<int, arena_allocator<int>>, arena_allocator<std::vector<int, arena_allocator<int>>>>
```

I think the check we would want is if either the inner type is trivially destructible or if the inner type also uses Stan's arena allocator?

cpp977:
Yes that makes sense because types with dynamic allocations managed by `arena_allocator` are essentially trivially destructible. However, I don't know what `std::is_trivially_destructible` returns for an arena managed dynamic type like `std::vector`. Probably `false`..

SteveBronder:
Yes i think it would have to be something like 

```cpp
std::is_trivially_destructible<std::decay_t<T>> || container_uses_arena_allocator<T>::value
```

Where `container_uses_arena_allocator` would be something whats in the godbolt below. Thought we would also need it to detect if a type has a `value_type` and if so then recurse through that as well checking if it's value type is also trivially destructible or arena allocated.

https://godbolt.org/z/5Tn5eh67r

I'm fine with merging the PR as is but if you'd like to help make arena allocator a bit safer to use your help would be very appreciated. Stan math already has a lot of stuff in the godbolt so I think the above would mostly involve adding recursive checking and some tests

cpp977:
I have added additional tests for the others containers and a usage warning in the docs about the problems with dynamic types as we discussed above.

SteveBronder (approved):
Good! Feel free to merge when tests pass

cpp977:
Any idea why the CI was failing? I ran the tests locally and they went fine but I ran them without `--jumbo`.

rok-cesnovar:
Its looking like including one of these

```
#include <deque>
#include <list>
#include <forward_list>
#include <set>
#include <unordered_set>
#include <map>
#include <unordered_map>
```
causes one of the other tests to fail. I have no idea why that would be the case. The `--jumbo` just merged a few test file together as if they were placed in a single fine - that severly speedups up testing runtimes.

Anyhow, sorry about this, will take a look shortly - dont worry about it.

rok-cesnovar:
Please ignore my above post, even the jumbo tests pass for me locally, so it seems like a CI issue.

rok-cesnovar:
Seems that a simple restart fixed it.

rok-cesnovar:
@cpp977 thanks and congrats on the first merged PR to the Stan Math repository! 🎉 🎉  !

SteveBronder:
@cpp977 yes thank you very much!

cpp977:
@SteveBronder and @rok-cesnovar thanks for the quick review and merging process! Thats definitely motivating for further contributions!

Issue #2705, opened 2022-04-13
Wishart Cholesky parameterization


spinkney:
We can add the Cholesky parameterization on both X and Sigma or just on Sigma

In cpp we'd use LDLt so we don't have to `add_diag`.
```stan
  real wishart_cholesky_a_lpdf(matrix X, real n, matrix L) {
    int P = rows(L);
    real lp = (n - P - 1) * sum(log(diagonal(cholesky_decompose(add_diag(X, 1e-7)))));
    
    // normalizing constant
    // lp -= 0.5 * (n * P) * log2() + lmgamma(P, 0.5 * n);
     lp -= 0.5 * sum(diagonal(mdivide_left_tri_low(L, mdivide_left_tri_low(L, X)')));

    return lp - n * sum(log(diagonal(L)));
  }
  
   real wishart_cholesky_b_lpdf(matrix Lx, real n, matrix L) {
    int P = rows(L);
    real lp = (n - P - 1) * sum(log(diagonal(Lx)));
    matrix[P, P] LinvLx = mdivide_left_tri_low(L, Lx);
    
     // normalizing constant
    //  lp -= 0.5 * (n * P) * log2() + lmgamma(P, 0.5 * n);
    lp -= 0.5 * sum(diagonal(crossprod(LinvLx)));

    return lp - n * sum(log(diagonal(L)));
  }
```
The results equal to what the `CholWishart` R package has

```R
library(rethinking)
> L <- t(chol(rlkjcorr(1, 8, eta = 0.1)))
> Lx <- t(chol(rlkjcorr(1, 8, eta = 0.1)))

> dWishart(x = tcrossprod(Lx), df = 9, Sigma = tcrossprod(L), log = TRUE)
[1] -443.0111
> wishart_cholesky_a_lpdf(X = tcrossprod(Lx), n = 9, L = L)
[1] -443.0111
> wishart_cholesky_b_lpdf(Lx = Lx, n = 9, L = L)
[1] -443.0111
```

bob-carpenter:
Neat.  That's really clean and should be OK to autodiff through in a direct C++ translation.  You can simplify even further replacing sum(diagonal(...)) with trace(...), which should have better autodiff properties.

I think we only need the "b" version.  The important case is where the variant is a Cholesky factor because that's the efficient data structure for covariance matrix parameters.

Can you figure out the adjoint-Jacobian update rule?  That is, let 

```
lp = wishart_cholesky_b_lpdf(Lx, n, L)
```

[...]

PR #2710, opened 2022-04-23
add wishart_cholesky

spinkney:
## Summary

Adds the `wishart_cholesky_lpdf()` and `wishart_cholesky_rng()` functions.

## Tests

The tests use the `wishart_lpdf` tests but add log absolute Jacobian correction for the change of variables from the input X to the Cholesky factor L.

## Release notes

The `wishart_cholesky_lpdf` is the Cholesky parameterization of the Wishart distribution of both the input matrix and the scale matrix.

## Checklist

- [x] Math issue #2705 

### DIFF
--- a/stan/math/prim/prob/wishart_cholesky_lpdf.hpp
+++ b/stan/math/prim/prob/wishart_cholesky_lpdf.hpp
@@ -12,12 +12,12 @@ namespace stan {
 namespace math {
 
 /** \ingroup multivar_dists
- * The log of the Wishart density for the given Cholesky factor, degrees of
- freedom,
- * and scale Cholesky factor matrix.
+ * Return the natural logarithm of the unnormalized Wishart density of the specified
+ * lower-triangular Cholesky factor variate, positive degrees of freedom, and lower-triangular
+ * Cholesky factor of the scale matrix.
  *
- * The scale matrix, LS, must be a lower Cholesky factor.
- * Dimension, k, is implicit.
+ * The scale matrix, L_S, must be a lower Cholesky factor
+ * dimension, k, is implicit
  * nu must be greater than k-1
  *
  * The change of variables from the input positive-definite matrix to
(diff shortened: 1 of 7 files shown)


spinkney:
@bob-carpenter @SteveBronder ready for review.

spinkney:
@bob-carpenter could you review this before the `inv_wishart_cholesky` because that depends on this for the `_rng` function. It's easier to go in order than have to merge and unmerge. I tried to incorporate all the changes that you referenced in #2713

stan/math/prim/prob/wishart_cholesky_lpdf.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+  if (include_summand<propto, T_dof>::value) {
+    lp -= lmgamma(k, 0.5 * nu_ref);
+  }
+
+  if (include_summand<propto, T_dof, T_scale, T_y>::value) {
+    auto L_SinvL_Y = mdivide_left_tri<Eigen::Lower>(L_S_ref, L_Y_ref);
+    for (int i = 0; i < k; i++) {
+      lp -= 0.5 * dot_self(L_SinvL_Y.row(i).head(i + 1))
+            + nu_ref * log(L_S_ref.coeff(i, i))
```
bob-carpenter, line 82: [optional]
Some of this can be made more efficient.

1. multiply by 0.5 after adding all the dot self's.  Otherwise, multiplication is a bit expensive originally and during autodiff.
2.  remove line 82 and include on outside as. ``lp -= nu_ref * L_S_ref.diag().log().sum();` or whatever the relevant Eigen calls are to do that.  It again pulls the multiplication out of the loop.
3. cache `nu_ref - 1` and reuse in 83.  This also saves auutodiff.  
4. replace `a * log(b)` with `multiply_log(a, b)`---it should have more efficient built-in autodiff.
5.  pull line 83 out and replace with `lp -= RowVectorXd::LinSpaced(1+nu_ref, 1+nu_ref + k) * L_Y_ref.diag().log()`. 

I may have slipped a sign.  I'd generally recommend always using `+=` in these compounds and just negating inside where necessary.  So rather than `lp -= a + b - c`, I'd recommend `lp += c - a - b;` to have less nested negation.

stan/math/prim/prob/wishart_cholesky_rng.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/** \ingroup multivar_dists
+ * Return a random Cholesky factor of the inverse covariance matrix
+ * of the specified dimensionality drawn from the Wishart distribution
+ * with the specified degrees of freedom
+ * using the specified random number generator.
```
bob-carpenter, line 18: [optional]
same issue here with early line ends

test/unit/math/mix/prob/wishart_cholesky_test.cpp
```diff
@@ -0,0 +1,102 @@
+#include <test/unit/math/test_ad.hpp>
+
+TEST(ProbDistributionsWishartCholesky, matvar) {
+  auto f = [](const auto& L_Y, const auto& dof, const auto& L_S) {
+    return stan::math::wishart_cholesky_lpdf(L_Y, dof, L_S);
+  };
+
+  double dof = 3.4;
+  Eigen::MatrixXd L_Y11(1, 1);
```
bob-carpenter, line 9: [comment]
Thanks for adding the boundary and autodiff tests.

bob-carpenter (changes requested):
This looks really great, @spinkney.  I only had a few requested changes and then a bunch of optional stuff if you're feeling more ambitious.

stan/math/prim/prob/wishart_cholesky_lpdf.hpp
spinkney: On point 2.
One of the tests fails doing this because I'm passing an `unsigned int` for the dof. When I multiply this by the sum of the log diagonal it blows up. Turns out it's not doing the casting to a float. If I also multiply by `1.0` then it passes. Is there a better way to get this to work?

spinkney: On point 5.
Since it's a double `-` they cancel out and the correct replacement is 
```c++
// using stan::math functions
lp += stan::math::reverse(stan::math::linspaced_row_vector(k, nu_ref - k, nu_ref - 1)) * log(L_Y_ref.diagonal());

// using Eigen
lp += Eigen::RowVectorXd::LinSpaced(k, nu_ref - k, nu_ref - 1).reverse() * log(L_Y_ref.diagonal());
```
However, I cannot get this to work with the mix tests. Lots of stan-var issues. Do you have any suggestions?

spinkney: I did what I was able to. I can't get the `LinSpaced` to work with rev AD.

spinkney: Oddly it's just that `LinSpaced` function. It works if I create the row vector and populate it as below
```cpp
  if (include_summand<propto, T_dof, T_scale, T_y>::value) {
    auto L_SinvL_Y = mdivide_left_tri<Eigen::Lower>(L_S_ref, L_Y_ref);
    T_v dot_LSinvLY(0.0);
    Eigen::Matrix<T_v, 1, Eigen::Dynamic> linspaced_v(k);
    auto nu_minus_1 = nu_ref - 1;

    for (int i = 0; i < k; i++) {
      dot_LSinvLY += dot_self(L_SinvL_Y.row(i).head(i + 1));
      linspaced_v(i) = nu_minus_1 - i;
    }
    lp += -0.5 * dot_LSinvLY;
    lp += linspaced_v * log(L_Y_ref.diagonal());
    lp += -1. * nu_ref * sum(log(L_S_ref.diagonal()));
  }
```
Let me know if you'd prefer this in the PR.

bob-carpenter: Thanks for working so hard on trying to get these to work.  

Point 2:  The standard way is to use `static_cast<double>(x)` rather than `x * 1.0`.  I wonder why it's not getting cast properly.  C++ should take care of any casting required for binary operators.

Point 5:  No worries if you can't make it work.  I'd think if this works, then a `.eval()` would also work (probably after `.reverse()`.

spinkney: Even that static cast is giving me the same errors. Maybe I'm doing something wrong. Here's what it errors
```cpp
./stan/math/prim/prob/wishart_cholesky_lpdf.hpp:76:12: error: cannot convert 'const stan::math::fvar<stan::math::fvar<double>>' to 'double' without a conversion operator
    lp += -static_cast<double>(nu_ref) * sum(log(L_S_ref.diagonal()));
```

bob-carpenter: Oh, that's an `fvar<fvar<double>>`, so you can't cast to `double`.  Without more context, it's really hard to help debug here.  I can't figure out what multiplying by 1.0 is doing, because that will also give you an `fvar<fvar<double>>`.

spinkney: Ok, I'm going put that part back into the loop and hopefully can get this merged this week. I don't fully understand what you mean or what's happening. What I do understand is that the template type for `nu` is not fully capturing the complexity of what `nu` can be. Somehow I'd need to tell the compiler that `nu` can be a bunch of things, even an `fvar<fvar<double>>`. 

I'd be grateful if you'd take a look or if we could meet and take a look together so I have a better understanding of this type of template issue in the future.

spinkney: I got it to work!

spinkney:
@bob-carpenter I got it to work! I have to construct the linspaced vector myself but other than that everything is in.

stan/math/prim/prob/wishart_cholesky_lpdf.hpp
bob-carpenter: Always glad to supply a coding puzzle :-).   Advanced solutions would require someone else.

bob-carpenter (approved):
Looks great---thanks so much for working through all those compile nightmare changes.  This should be ready to merge when it passes tests.

PR #2720, opened 2022-04-29
[WIP] Adds boost root finders with reverse mode specializations

SteveBronder:
## Summary

Adds a wrapper for [boost's Halley](https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/roots_deriv.html) root finding algorithm along with a reverse mode specialization similar to our algebra solver. Note these functions are only meant to be used at the Stan math level. Specifically for the quantile functions in https://github.com/stan-dev/design-docs/pull/42

This also works for forward mode, but all it does is iterate through the boost functions so it is probably very slow.

## Tests

Tests for forward and reverse mode using the [cubed](https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/root_finding_examples/cbrt_eg.html) and [5th](https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/root_finding_examples/5th_root_eg.html) root examples from boost's docs. Tests can be run with 

```
./runTests.py -j4 ./test/unit/math/ -f root_finder
```

## Side Effects

Note that I had to write specializations for `frexp` and `sign` for `fvar` and `var` types because boost would throw compiler errors if they did not exist.

The WIP here is because of some questions for @bgoodri 

1. Would we ever want to use the newton raphson or schroder root finders? I wrote this keeping those in mind and it would be very simple to support.
2. Right now the API takes in a tuple of functors to calculate the value and derivatives. But you can see in the 5th root example this is kind of inefficient since we are just doing x^n and could share that computation between the calculations. For the reverse mode specialization we do need a functor to calculate just the function we want to solve, but we could also allow for the user to pass two functors where one returns a tuple with all the values for the root finder and another that just gives back the value of the function we want to solve. 
3. Do we need functions for working over vectors elementwise or should we make that later?

## Release notes

Adds boost root finders with reverse mode specializations for 

## Checklist

- [ ] Math issue #


bgoodri:
(1) I imagine it will end up like the ODE solvers where we have slightly different versions with a mostly common set of arguments. But Halley is probably the one that would get used the most inside Stan Math.

bgoodri:
(2) It seems like it would be good to permit (or require) one functor that returns a tuple of: level, first derivative, second derivative.

bgoodri:
(3) If a root needs to be found for multiple parameter values, I would say that the caller is responsible for invoking `root_finder` multiple times.

bgoodri:
I think we should default the `digits` argument according to the recommendation in the Boost [docs](https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/roots_deriv.html). 

> The value of digits is crucial to good performance of these functions, if it is set too high then at best you will get one extra (unnecessary) iteration, and at worst the last few steps will proceed by [bisection](https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/roots_noderiv/bisect.html). Remember that the returned value can never be more accurate than f(x) can be evaluated, and that if f(x) suffers from cancellation errors as it tends to zero then the computed steps will be effectively random. The value of digits should be set so that iteration terminates before this point: remember that for second and third order methods the number of correct digits in the result is increasing quite substantially with each iteration, digits should be set by experiment so that the final iteration just takes the next value into the zone where f(x) becomes inaccurate. A good starting point for digits would be 0.6*D for Newton and 0.4*D for Halley or Shröder iteration, where D is std::numeric_limits<T>::digits.

bgoodri:
You could also easily test `a * x^2 + b * x + c = 0` where `a`, `b`, and / or `c` are `var`, `fvar<var>`, etc. to make sure our implicit `da / dx`, `db / dx`, and `dc / dx` match those obtained by differentiating `x = (-b + sqrt(b^2 - 4 * a * c)) / (2 * a)` explicitly.

bgoodri:
From your email, `guess`, `min`, and `max` should definitely be templated but we just need to pull out the underlying numbers since we wouldn't need to differentiate with respect to them.

bgoodri:
Also, @charlesm93 might have some thoughts. As I mentioned yesterday, the motivation for having a one-dimensional specialization of `algebraic_solver` in Stan Math (we are not worried about exposing it to the Stan language yet) is

1. Everything is in terms of scalars so no there is Eigen overhead making vectors of size 1, 1x1 Jacobians, etc.
2. We can use bounds to ensure the correct root is found
3. We can utilize second derivatives to obtain a tripling of the number of correct digits when close to the root

I guess the main question is what is the fastest way of implicitly differentiating the root with respect to unknown parameters? A concrete example would be thinking about implementing the inverse CDF of the Beta distribution in C++ with possibly unknown shape parameters. Given a cumulative probability, `p`, we would need to find the value of `x` between 0 and 1, such that `beta_cdf(x, alpha, beta) - p = 0`. We know the first derivative wrt `x` is the beta PDF and the second derivative is the first derivative of the beta PDF and can presume within Stan Math that the caller has specified those two derivatives explicitly. Once `x` is found, its sensitivity to `p` is the reciprocal of the beta PDF, but the sensitivity of `x` to `alpha` and `beta` requires implicit differentiation.

bgoodri:
We should probably also do the beta inverse CDF as a test case. Boost has a more advanced way of obtaining an initial guess in this case, but they still usually do Halley iteration
```
var p, alpha, beta; // give these values
using namespace boost::math; 
beta my_beta(alpha, beta);
auto x = quantile(my_beta, p);
// check our version against this
```

SteveBronder:
So I got the tests for beta up and running, for the beta lcdf it seems to work fine for values where lgamma(alpha + beta) > 1. Maybe I wrote the derivative of the pdf wrt x wrong? Or there could be a numerics trick I'm missing.

For the test against boost's quantile function, `var` does not work with the boost distributions because of some functions they defined that we do not have defined for vars. But using finite diff I get answers that are within `4e-4`. Perhaps I'm not doing something quite right there as well? Or it could be just finite diffs own error. There are also some places that finite diff gives -nan values for alpha and beta but where autodiff can return the correct values

```
# from running 
# ./runTests.py -j2 test/unit/math/rev/functor/root_finder_test.cpp
--- Finit Diff----
fx: 0.288633
grads: 
p: 0.6987
alpha: 0.245729
beta: -0.121384
--- Auto Diff----
fx: 0.288714
grads: 
p: 0.700012
alpha: 0.248911
beta: -0.12257

grad diffs: 
p: -0.00131203
alpha: -0.00318197
beta: 0.00118579
```

SteveBronder:
Also do we want to test lcdf or cdf functions?

SteveBronder:
@bgoodri so I got the rev test up and running with the explicit gradients for the beta cdf. I also wrote a lil' function that just checks the gradients and our function to make sure they are near'ish to each other. Though the test is failing for some odd values that I wouldn't expect them to. The failing ones all print their values after a failure so you can take a look if you like (though note after one failure it will just keep printing all the sets of values it is testing. This is just because I have not figured out how to stop that from happening in gtest). I wonder if it could be some numerically inaccuracy with our `beta_cdf` function?

charlesm93:
(2): if the function is going to be used internally, we should try and optimize as much as possible. I see two options:
* having one functor argument which returns the value and one functor argument which returns a tuple with the value and the derivatives.
* passing a single functor which admits an argument indicating the order of the derivatives we need to compute. Internally, only do the calculations for the specified order, and return a tuple of value and derivatives (with the latter potentially just containing nans). 

If this is exposed to the user, we may want to trade some efficiency for a simpler API.

stan/math/prim/functor/root_finder.hpp
```diff
@@ -0,0 +1,181 @@
...
+auto root_finder_halley_tol(FTuple&& f_tuple, const GuessScalar guess,
+                            const MinScalar min, const MaxScalar max,
+                            const int digits, std::uintmax_t& max_iter,
+                            Types&&... args) {
+  return root_finder_tol(
+      [](auto&&... args) {
+        return boost::math::tools::halley_iterate(args...);
+      },
+      std::forward<FTuple>(f_tuple), guess, min, max, digits, max_iter,
+      std::forward<Types>(args)...);
```
charlesm93, line 83: This my lacking C++ skills speaking but could you describe how the call to `root_finder_tol` works here? How does the `[](auto&&... args)` argument work?

stan/math/prim/functor/root_finder.hpp
```diff
@@ -0,0 +1,181 @@
...
+ * @param max The maximum possible value for the result, this is used as an
+ * initial upper bracket
+ * @param args Parameter pack of arguments to pass the the functors in `f_tuple`
+ */
+template <typename SolverFun, typename FTuple, typename GuessScalar,
+          typename MinScalar, typename MaxScalar, typename... Types>
+auto root_finder(SolverFun&& f_solver, FTuple&& f_tuple,
+                 const GuessScalar guess, const MinScalar min,
+                 const MaxScalar max, Types&&... args) {
+  constexpr int digits = 16;
```
charlesm93, line 137: how was this default chosen? Maybe add one sentence about this choice in the doxygen doc.

charlesm93:
Why is there a specialized rev method? The implicit function is valid for both forward and reverse mode autodiff. Since the solution is a scalar, the full Jacobian of derivatives reduces to a vector, so we don't gain much from doing contractions with either an initial tangent or a cotangent.

I think it would make sense to explicitly compute the full Jacobian and then do the vector-vector product for either forward or reverse mode.

stan/math/rev/functor/root_finder.hpp
```diff
@@ -0,0 +1,101 @@
...
+  double Jf_x;
+  auto&& f = std::get<0>(f_tuple);
+  {
+    nested_rev_autodiff nested;
+    stan::math::var x_var(theta_dbl);
+    stan::math::var fx_var = apply(
+        [&x_var, &f](auto&&... args) { return f(x_var, std::move(args)...); },
+        std::move(args_vals_tuple));
+    fx_var.grad();
+    Jf_x = x_var.adj();
```
charlesm93, line 72: This is a scalar. You might want to call it df_dx, since J usually suggest a Jacobian matrix (but this is a minor detail). Does `nested_rev_autodiff` mean you're using reverse mode? For computing a scalar, I recommend using forward mode.

stan/math/rev/functor/root_finder.hpp
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /*
+   * Note: Because we put this on the callback stack, if `f` is a lambda
+   * its captures must be in Stan's arena memory or trivially destructable.
+   */
+  return make_callback_var(theta_dbl,
+                           [f, arena_args_tuple, Jf_x](auto& ret) mutable {
+                             // Eigen::VectorXd eta =
```
charlesm93, line 81: hmmm... can you remind me what `ret` is?

test/unit/math/mix/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+  std::uintmax_t maxit = 20;
+  auto f = [](const auto& g, const auto& x) { return g * g * g * g * g - x; };
+  auto f_g = [](const auto& g, const auto& x) { return 5 * g * g * g * g; };
+  auto f_gg = [](const auto& g, const auto& x) { return 20 * g * g * g; };
+  auto full_f = [&f, &f_g, &f_gg, guess, min, max](auto&& xx) {
+    return root_finder(std::make_tuple(f, f_g, f_gg), guess, min, max, xx);
+  };
+  stan::test::expect_ad(full_f, x);
+}
+*/
```
charlesm93, line 57: Why are these tests commented out?

test/unit/math/mix/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return stan::math::beta_lcdf(x, alpha, beta) - p;
+  };
+  auto f_g = [](auto&& x, auto&& alpha, auto&& beta, auto&& p) {
+    return stan::math::beta_lpdf(x, alpha, beta);
+  };
+  auto f_gg
+      = [&beta_lpdf_deriv](auto&& x, auto&& alpha, auto&& beta, auto&& p) {
+          return beta_lpdf_deriv(x, alpha, beta, p);
+        };
+  double guess = .3;
```
charlesm93, line 79: Why .3? Does the test pass with other initial guesses?

test/unit/math/mix/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+  double min = 0.0000000001;
```
charlesm93, line 80: Would `min = 0` not work?

charlesm93: If not this should be indicated in a comment.

test/unit/math/mix/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+  auto full_f = [&f, &f_g, &f_gg, guess, min, max](auto&& alpha, auto&& beta,
+                                                   auto&& p) {
+    std::uintmax_t max_its = 1000;
+    return stan::math::root_finder_hailey(std::make_tuple(f, f_g, f_gg), guess,
+                                          min, max, alpha, beta, p);
+  };
+  double p = 0.45;
+  double alpha = .5;
+  double beta = .4;
+  stan::test::expect_ad(full_f, alpha, beta, p);
```
charlesm93, line 91: Reminder for me: this checks both reverse and forward mode AD, and higher-order AD as well, correct?

charlesm93: Does `expect_ad` allow a tolerance parameter -- as a way to keep track of how closely finite-diff and autodiff agree?

charlesm93: in addition to testing derivatives, are you also checking that the root finder returns the correct value?

test/unit/math/mix/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+    std::uintmax_t max_its = 1000;
+    return stan::math::root_finder_hailey(
+        std::make_tuple(f, beta_pdf, beta_pdf_deriv), guess, min, max, alpha,
+        beta, p);
+  };
+  double p = 0.8;
+  double alpha = .4;
+  double beta = .5;
+  stan::test::expect_ad(full_f, alpha, beta, p);
+}
```
charlesm93, line 130: Same comment about this unit test that I made for the previous test.

test/unit/math/rev/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,327 @@
...
+  std::cout << "--- Finit Diff----\n";
+  std::cout << "fx: " << fx;
+  std::cout << "\ngrads: \n"
+            << "p: " << finit_grad_fx(0)
+            << "\n"
+               "alpha: "
+            << finit_grad_fx(1)
+            << "\n"
+               "beta: "
+            << finit_grad_fx(2) << "\n";
```
charlesm93, line 64: I understand this is here for the discussion in the PR, but in time, we'll remove the `std::cout`.

charlesm93 (changes requested):
Overall this looks good to me but it still requires some additional work. Some of the C++ looks arcane to me, so I've asked for clarifications and I'm relying on the quality of the unit tests.

Here are the main requests:
* The unit tests need to check that the correct root is returned. You can work out the solution analytically; plug the root into the function being evaluated and check that it is close to 0 within the specified tolerance; or use `algebra_solver` as a benchmark.
* When setting tuning parameters for the rootfinder, leave a comment to justify those choices. Are these arbitrary tuning parameters or are they required to make the root finder work?
* Need additional tests which prompt the error messages.
* Address comments on the file. A handful of them are about doxygen doc.

Some suggestions:
* Consider using the implicit function theorem for both forward and reverse mode autodiff. I don't think using a direct method for forward diff is justified.
* As mentioned in the discussion, consider passing only one functor rather than a tuple of functors.

bgoodri:
Thanks @charlesm93

stan/math/prim/functor/root_finder.hpp
SteveBronder: I need to update this to be in line with what boost's docs say 

> The value of digits is crucial to good performance of these functions, if it is set too high then at best you will get one extra (unnecessary) iteration, and at worst the last few steps will proceed by [bisection](https://www.boost.org/doc/libs/1_62_0/libs/math/doc/html/math_toolkit/roots/roots_noderiv/bisect.html). Remember that the returned value can never be more accurate than f(x) can be evaluated, and that if f(x) suffers from cancellation errors as it tends to zero then the computed steps will be effectively random. The value of digits should be set so that iteration terminates before this point: remember that for second and third order methods the number of correct digits in the result is increasing quite substantially with each iteration, digits should be set by experiment so that the final iteration just takes the next value into the zone where f(x) becomes inaccurate. A good starting point for digits would be 0.6*D for Newton and 0.4*D for Halley or Shröder iteration, where D is std::numeric_limits<T>::digits.

https://www.boost.org/doc/libs/1_62_0/libs/math/doc/html/math_toolkit/roots/roots_deriv.html

stan/math/rev/functor/root_finder.hpp
SteveBronder: > Does nested_rev_autodiff mean you're using reverse mode? 

Yes

> For computing a scalar, I recommend using forward mode.

What is the intuition on fwd being faster in this case? I thought if it was many input single output then reverse is the best?

stan/math/rev/functor/root_finder.hpp
SteveBronder: `ret` here is the return of the `root_finder_tol` function in reverse mode. Another way to think about this code is 

```cpp
var ret = theta_dbl;
reverse_pass_callback([f, arena_args_tuple, Jf_x, ret]() mutable {
                             {
                               nested_rev_autodiff rev;
                               double ret_val = ret.val();
                               auto x_nrad_ = apply(
                                   [&ret_val, &f](const auto&... args) {
                                     return eval(f(ret_val, args...));
                                   },
                                   *arena_args_tuple);
                               x_nrad_.adj() = -(ret.adj() / Jf_x);
                               grad();
                             }
                           });
return ret;
```

test/unit/math/mix/functor/root_finder_test.cpp
SteveBronder: Just for convenience so I could compile the other tests in the file faster I do need to uncomment these

test/unit/math/mix/functor/root_finder_test.cpp
SteveBronder: .3 is just a wild guess from me. In the test below I use .5 as well. @bgoodri should there be some way we cleverly set the initial guess for the beta test?

test/unit/math/mix/functor/root_finder_test.cpp
SteveBronder: Yes I can update this

test/unit/math/mix/functor/root_finder_test.cpp
SteveBronder: > Reminder for me: this checks both reverse and forward mode AD, and higher-order AD as well, correct?

Yep!

> Does expect_ad allow a tolerance parameter -- as a way to keep track of how closely finite-diff and autodiff agree?

Yes, though I'm not sure yet which level of tolerance we can / should expect 

> in addition to testing derivatives, are you also checking that the root finder returns the correct value?

It just checks finite diff vs autodiff of the function for correct gradients, if the primative and reverse mode specializations give back different answers it would throw an error, but I do need some test against like our current algebra solver to make sure things are as we expect.

test/unit/math/rev/functor/root_finder_test.cpp
```diff
@@ -0,0 +1,327 @@
...
+                     / (tgamma(1 + b) * tgamma(1 + b)))
+                  * pow(1 - ibeta_inv(a, b, p), b)
+              - beta(b, a, 1 - ibeta_inv(a, b, p))
+                    * (log(1 - ibeta_inv(a, b, p)) - polygamma(0, b)
+                       + polygamma(0, a + b)));
+  };
+  double known_p_grad = deriv_p(p, a, b);
+  double known_alpha_grad = deriv_a(p, a, b);
+  double known_beta_grad = deriv_b(p, a, b);
+  std::cout << "--- Mathematica Calculate Grad----\n";
```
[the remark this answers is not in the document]
SteveBronder: Yes def agree would be good to copy/paste those into here

SteveBronder:
@bgoodri alrighty so I got reverse mode to be adjoints to be right up to 1e-9. It turns out that we need more precision when doing the forward pass during reverse mode which I found kind of weird and interesting. For the beta cdf example it seemsl like we need about 23 (???) digits of precision which is very odd to me. Like we only work in doubles so I'm not sure why having anything over 16 would be a thing? But tests now pass for testing alpha,beta, and p from .1 to .9

bob-carpenter:
>  It turns out that we need more precision when doing the forward pass during reverse mode which I found kind of weird and interesting

The forward-mode values are often used as part of the derivative calculations, so I suspect that may be the problem.

> For the beta cdf example it seemsl like we need about 23 (???) digits of precision which is very odd to me

As you point out, we only have about 16 digits of precision in double-precision, so we can't return results with that high precision.   CPUs often do higher-precision internal arithmetic then truncate the return, but there's simply no way to get 23 digits of accuracy in a floating-point result at double precision.

dmi3kno:
May I suggest that we adopt existing (closed-form) QF/CDF pairs for tests? Here are suggested QFs and CDFs (where missing in Stan).
- Unbounded: 
  - logistic distribution
$$Q(u)=\mu+s[\ln(u)-\ln(1-u)]$$
- Semi-bounded: 
  - Weibull 
  $$Q(u)=\lambda[-\ln(1-u)]^{1/k}$$
  - Gumbel distributions 
  $$Q(u)=\mu-\beta\ln(-\ln(u))$$
  - Exponential (simplest QF, but perhaps most difficult to invert)
   $$Q(u)=(1/\lambda)\ln(1-u)$$
- Bounded: 
  - Kumaraswamy distribution
 $$Q(u)=(1-(1-u)^{1/b})^{1/a}$$
 $$F(x)=1-(1-x^a)^b$$

All of these distributions have closed form CDF and QF so we always have analytical solution to check against.

PR #2725, opened 2022-05-02
Update to Boost 1.78.0

### LINKED ISSUE #2711: Update to Boost 1.78.0
## Description

Boost 1.79 was released in April, we are still at 1.75. We should try to keep up and upgrade.

Release notes for all version since 1.75: 
- https://www.boost.org/users/history/version_1_79_0.html
- https://www.boost.org/users/history/version_1_78_0.html
- https://www.boost.org/users/history/version_1_77_0.html
- https://www.boost.org/users/history/version_1_76_0.html

#### Current Version:
v4.3.2

rok-cesnovar:
## Summary

Updates the Boos library to 1.78, which is also the version currently on CRAN's Boost Headers package.

## Tests

/

## Side Effects

/

## Release notes

Updated Boost to 1.78.0.

## Checklist

- [x] Math issue Fixes  #2711 


rok-cesnovar:
An integrate_1d test now fails with Boost 1.78

The test that fails: https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/functor/integrate_1d_impl_test.cpp#L339

The test error:

```
[ RUN      ] StanMath_integrate_1d_impl_prim.test_integer_arguments
test/unit/math/prim/functor/integrate_1d_impl_test.cpp:341: Failure
Expected: v = stan::math::integrate_1d_impl(integrate_1d_impl_test::f2{}, 0, 1, 0.0, integrate_1d_impl_test::msgs) doesn't throw an exception.
  Actual: it throws std::domain_error with description "integrate: error estimate of integral 2.26544e-127 exceeds the given relative tolerance times norm of integral".
test/unit/math/prim/functor/integrate_1d_impl_test.cpp:344: Failure
Expected: v = stan::math::integrate_1d_impl(integrate_1d_impl_test::f2{}, 0.0, 1, 0.0, integrate_1d_impl_test::msgs) doesn't throw an exception.
  Actual: it throws std::domain_error with description "integrate: error estimate of integral 2.26544e-127 exceeds the given relative tolerance times norm of integral".
test/unit/math/prim/functor/integrate_1d_impl_test.cpp:347: Failure
Expected: v = stan::math::integrate_1d_impl(integrate_1d_impl_test::f2{}, 0, 1.0, 0.0, integrate_1d_impl_test::msgs) doesn't throw an exception.
  Actual: it throws std::domain_error with description "integrate: error estimate of integral 2.26544e-127 exceeds the given relative tolerance times norm of integral".
[  FAILED  ] StanMath_integrate_1d_impl_prim.test_integer_arguments (1 ms)
```

Does anyone have any ideas why this would fail?

WardBrian:
The release notes mention:
> - Improve worst-case tanh-sinh integration performance.
> - Disable inadvertant use of integral types in integration routines.

So my next question is where do the reference values for this test come from, and is it possible that the new answer is actually more correct?

rok-cesnovar:
I have to admit I have zero knowledge about the integrate_1d codebase and I am not sure who is the resident expert there.

@yizhang-yiz or @bbbales2 maybe?

bbbales2:
Hmm, looking at the integrate_1d_impl code it seems like these tests are set to run with a relative tolerance of zero (integrate_1d_impl [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/functor/integrate_1d.hpp#L175) for prim) . That doesn't really make sense so seems like an error in the tests.

Going from the boost docs on tolerance [here](https://www.boost.org/doc/libs/1_72_0/libs/math/doc/html/math_toolkit/double_exponential/de_tol.html)

> It is highly recommended that the tolerance be left at the default value of √ε, or something similar

The non _impl version of the function does this (see [here](https://github.com/stan-dev/math/blob/develop/stan/math/prim/functor/integrate_1d.hpp#L244)) and so I guess do that in these tests would be fine. I forget why this test is using _impl or not -- can dig further if the reasoning isn't obvious and that seems like a meaningful detail.

The specific function in the failed tests comes from the bit of the documentation on handling endpoints [here](https://www.boost.org/doc/libs/1_72_0/libs/math/doc/html/math_toolkit/double_exponential/de_tanh_sinh_2_arg.html)

rok-cesnovar:
Thanks a ton Ben!

Tests are passing now with the `sqrt(epsilon)` tolerance in the test. With that, this is ready for review.

This is just a run of the [upgrade-boost](https://github.com/stan-dev/math/blob/develop/lib/upgrade-boost.sh) script with two additional commits (I would advise the reviewer to only look at those commits - the rest of the commits were done by the script):

- https://github.com/stan-dev/math/pull/2725/commits/4c477e684a19c512c2f8af6ea2156f43d7b839d8 - updated the script that it does not error if some file is not found in the prunning stage (we strip the Boost library of files we don't use)
- https://github.com/stan-dev/math/pull/2725/commits/93cbe0e992e4440447858f6cfb91259e82248e46 - fixed the test, by switching some tolerances from zero to `sqrt(epsilon)`.

bbbales2 (approved):
Looks good to me.

In the update script change also looks like some folders got removed?:

```
libs/*doc libs/*/*doc
```

Maybe those weren't there anymore to remove, or maybe they are used now. Either way not gonna hurt just keeping* them. I'll leave this for you to merge as convenient @rok-cesnovar

rok-cesnovar:
> Maybe those weren't there anymore to remove, or maybe they are used now. 

Yeah, some subfolders no longer have them and in order to not have to keep investigating this in the future if the structure change. We don't require these docs folders anyways.

Thanks for the review!

WardBrian:
It seems like this broke the cmdstan Rtools 3.5 test upstream: https://github.com/stan-dev/cmdstan/runs/6392117308?check_suite_focus=true

SteveBronder:
@rok-cesnovar should we bump that to Rtools 4.0?

rok-cesnovar:
It seems that Boost 1.78 does not work with Rtools35 so there are two options here: 
a) bump the minimum requirement for Stan on Windows to Rtools 4.0 (and fix the test by bumping - what @SteveBronder suggests).
b) revert this PR and re-do it after the release and bump to Rtools40 for the next release.

I would just go with a). Rtools 4.0 has been out for 2+ years and is now considered the "old" toolchain with Rtools42 being out already. And I don't think there is any need for us to devote resources to support 3 toolchains on Windows.

This is not a problem on Linux - still works with g++ 5, I think its once again an issue of the Rtools35 MinGW implementation.

wds15:
A sounds great.

PR #2727, opened 2022-05-05
[WIP] Allow unary var<Matrix> overloaded functions to also work with Matrix<var> types

SteveBronder:
## Summary

This modifies the current overloads for unary `var<Matrix>` functions so that they can also accept `Matrix<var>` functions. This should be a bit more efficient since copies to the memory arena will be done in bulk. To do this I created a new function `make_callback_rev_matrix` which, given the expected return type of either a `Matrix<var>` or `var<Matrix>` creates the correct return and places a callback for the reverse pass onto the reverse pass callback stack.

This is a WIP until I doc all the new functions

## Tests

There should be no new tests needed as we are writing specializations for functions already tested by the test ad framework.

## Side Effects

None

## Release notes

Add more matrix overloads for reverse mode unary functions
 
## Checklist

- [ ] Math issue #


wds15:
Are u planing a benchmark? Which models should benefit from this?

SteveBronder:
Yes I can writeup a benchmark. My thinking with these is that its better for memory if we are doing large allocations at once and reducing the callback stack size from N calls for each scalar to 1 call for the entire matrix.

PR #2728, opened 2022-05-06
Adjust signatures of eigenvalues and eigenvectors to accept expressions

WardBrian:
## Summary

This adjusts the signatures of the `eigenvalues` and `eigenvectors` functions to accept eigen expressions as input. This will allow them to be exposed in the Stan language: https://github.com/stan-dev/stanc3/pull/1192

## Tests

Existing tests for functionality remain in place. Additional expression tests will be automatically run when exposed in the compiler, I've run them locally for now.

## Side Effects

None

## Release notes

Updated the signatures of `eigenvalues` and `eigenvectors` to use our pseudo-concept require templates. 

## Checklist

### DIFF
--- a/stan/math/prim/fun/eigenvalues.hpp
+++ b/stan/math/prim/fun/eigenvalues.hpp
@@ -7,13 +7,12 @@
 namespace stan {
 namespace math {
 
-template <typename T>
-Eigen::Matrix<std::complex<T>, -1, 1> eigenvalues(
-    const Eigen::Matrix<T, -1, -1>& m) {
+template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr>
+inline auto eigenvalues(const EigMat& m) {
   check_nonzero_size("eigenvalues", "m", m);
   check_square("eigenvalues", "m", m);
 
-  Eigen::EigenSolver<Eigen::Matrix<T, -1, -1>> solver(m);
+  Eigen::EigenSolver<plain_type_t<EigMat>> solver(m);
   return solver.eigenvalues();
 }
 
--- a/stan/math/prim/fun/eigenvalues.hpp
+++ b/stan/math/prim/fun/eigenvalues.hpp
@@ -9,10 +9,12 @@ namespace math {
 
 template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr>
 inline auto eigenvalues(const EigMat& m) {
-  check_nonzero_size("eigenvalues", "m", m);
-  check_square("eigenvalues", "m", m);
+  using PlainMat = plain_type_t<EigMat>;
+  const PlainMat& m_eval = m;
+  check_nonzero_size("eigenvalues", "m", m_eval);
+  check_square("eigenvalues", "m", m_eval);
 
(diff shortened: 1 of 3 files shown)


stan/math/prim/fun/eigenvalues.hpp
```diff
@@ -7,13 +7,14 @@
...
 namespace math {
 
-template <typename T>
-Eigen::Matrix<std::complex<T>, -1, 1> eigenvalues(
-    const Eigen::Matrix<T, -1, -1>& m) {
-  check_nonzero_size("eigenvalues", "m", m);
-  check_square("eigenvalues", "m", m);
+template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr>
+inline auto eigenvalues(const EigMat& m) {
+  using PlainMat = plain_type_t<EigMat>;
```
bob-carpenter, line 12: [question]
Is there a reason to specify `PlainMat` rather than just using `.eval()`?

```cpp
auto& m_eval = m.eval();
```

test/statement_types.py
```diff
@@ -130,7 +130,10 @@ def __init__(self, overload, name, stan_arg, size, value=None):
         self.size = size
 
         if value is None:
-            self.value = 0.4
+            if 'complex' in stan_arg:
```
bob-carpenter, line 133: [optional]
I prefer ternary operators when the body of conditionals only assign to the same variable.  But ternary ops look pretty awkward in Python (of course this one wouldn't even be legal in C++ with its heterogeneous return types), so I'm marking as optional.

```python
self.value = to_complex(0.4, 04) if 'complex' in stan_arg else 0.4
```

bob-carpenter (approved):
This is fine as is.  My comments were more questions/optional and about style, not content.

stan/math/prim/fun/eigenvalues.hpp
WardBrian: We use `PlainMat` later as a template argument to the eigen solver, but to be completely honest I was just using the style also used in eigenvalues_sym.hpp and elsewhere. @SteveBronder - is there a reason to prefer one to the other?

test/statement_types.py
WardBrian: I agree in general but quite dislike the Python ternary operator's ordering (I'm much more used to the order `conditional - true value - false value` from C, Java, etc., so I still find `true value - conditional - false value` hard to parse even after using it for years)

bob-carpenter: Glad to hear it's not just me finding it hard to read.  That is why I marked it optional.

stan/math/prim/fun/eigenvalues.hpp
bob-carpenter: Good point about needing to use it to parameterize the solver.  So reducing this to `auto` plus `eval` won't be enough.  So I'd just leave it like this.

test/statement_types.py
rok-cesnovar: +1 on the Python ternary being hard to read :)

WardBrian: I think it was all to avoid making something like `then` a keyword. OCaml's if expression keeps the same order as C's ?: by having the style be `if cond then true_value else false_value`, which is probably my favorite option

Issue #2697, opened 2022-04-03
add `inv_Phi_log` which is `inv_Phi(log(x))`


spinkney:
@andrjohns has done some tests on `inv_Phi` and it's more stable to compute everything in terms of logs. He has a Stan implementation at https://github.com/andrjohns/brms/blob/feature/gaussian-copula/inst/chunks/fun_inv_Phi_log.stan. 

It is exactly the same algorithm as in stan-math but allows the user to input the value on the log scale. Doing all the interval calculations on this scale but still returning the same result as `inv_Phi`.

```r
> inv_Phi_log(log(0.3))
[1] -0.5244005
> qnorm(0.3)
[1] -0.5244005
```

bob-carpenter:
It certainly gives you more range.  It's also the same result as `normal_lcdf(x | 0, 1)`, so it'd be good to improve `normal_lcdf` if there are algorithmic improvements here.

Can we call this `log_inv_Phi`, though, to match our other functions (e.g., `log_inv_logit`, `log_1m_inv_logit`, etc.)?

andrjohns:
The naming is a bit tricky, since it's not quite the same as the other `log_` functions. `log_inv_Phi` implies that the function is `log(inv_Phi(x))`, but it's actually `inv_Phi(exp(x))`.

Also, I might be mistaken, but isn't this distinct from the `normal_lcdf` since `inv_Phi` is the quantile function of the standard normal, not the CDF?

bob-carpenter:
Sorry, @andrjohns---if that's the behavior, then your name is correct and it's not synonymous with the normal log cdf.

PR #2731, opened 2022-05-08
adds `std_normal_log_qf`

spinkney:
## Summary

Adds the quantile function of the standard normal density function with log input.

## Release notes

Adds the quantile function of the standard normal density function with log input.

## Checklist

- [x] Math issue #2697 


spinkney:
@andrjohns I cannot get the mix test to work. It just gives a google test error of xml failed. I'm just trying to get this to run and so I haven't updated the derivatives. At the moment, I'm using the same code as in the rev code for the `inv_Phi` function to see if I can get the mix test to at least run. 

The prim tests pass. Maybe you have an idea of how to get this to work?

spinkney:
I don't understand why this segfaults the google test framework. I believe it has to do with a template issue. I'm following the same template as in `inv_Phi`. The two main differences to `inv_Phi` are that this function uses functions in `stan::math` instead of `std` and it uses `long double` types. I've changed the `long doubles` to `doubles` but still the same issue. Also do we support `long double` types?

@rok-cesnovar @bob-carpenter @SteveBronder or someone else with template expertise may

```bash
test/unit/math/mix/fun/inv_Phi_log_test --gtest_output="xml:test/unit/math/mix/fun/inv_Phi_log_test.xml"
[==========] Running 1 test from 1 test suite.
[----------] 1 test from mathMixMatFun
[ RUN      ] mathMixMatFun.invPhiLog
test/unit/math/mix/fun/inv_Phi_log_test --gtest_output="xml:test/unit/math/mix/fun/inv_Phi_log_test.xml" failed
exit now (05/09/22 07:16:59 EDT)
```

spinkney:
Nevermind, with all the stan math functions I just need to templatize it.

spinkney:
@andrjohns or @SteveBronder this is ready for review

spinkney:
The impetus to add this function to the language was due to @andrjohns testing normal copula models in Stan and he found the numeric stability of working on the original scale problematic. Which is the same reason we typically work with `_lcdf` functions rather than `_cdf`. In the normal copula example one typically is interested in modeling each univariate margin of the multidimensional outcome as a (possibly) separate distribution and connecting each of these univariate margins by a multinormal correlation. The "trick" to model this is by transforming each univariate margin via their specified cdf into a probability which is then transformed again into "normal-space" via the normal quantile function - i.e. `inv_Phi`. Instead of using the cdf to begin the "trick", it is much computationally stable to use the log cdf. This, in turn, suggests to continue working on the log scale via inputting the log probability value into a modified quantile function which works on the log scale but still outputs to the same space as the original quantile function.

The idea can be expressed in the following pseudo-stan code (assuming we had a `multi_normal_cholesky_copula_lpdf` function):
```stan
data {
  int<lower=0> N;
  int K;   // K is 2 here
  matrix[N, K] x;
}
parameters {
  array[K] real mu;
  array[K] real<lower=0> sigma;
  cholesky_factor_corr[K] L;
}
model {
  target += gumbel_lpdf(x[ : , 1] | mu[1], sigma[1]);
  target += lognormal_lpdf(x[ : , 2] | mu[2], sigma[2]);
  
  {
    matrix[K, N] y;
    for (n in 1 : N) {
      y[1, n] = inv_Phi(gumbel_cdf(x[n, 1] | mu[1], sigma[1]));
      y[2, n] = inv_Phi(lognormal_cdf(x[n, 2] | mu[2], sigma[2]));
    }
     y ~ multi_normal_cholesky_copula(L);
  }
}
```
The addition of this `inv_Phi_log` means we can modify the above to the more computationally stable version of
```stan
data {
  int<lower=0> N;
  int K;   // K is 2 here
  matrix[N, K] x;
}
parameters {
  array[K] real mu;
  array[K] real<lower=0> sigma;
  cholesky_factor_corr[K] L;
}
model {
  target += gumbel_lpdf(x[ : , 1] | mu[1], sigma[1]);
  target += lognormal_lpdf(x[ : , 2] | mu[2], sigma[2]);
  
  {
    matrix[K, N] y;
    for (n in 1 : N) {
      y[1, n] = inv_Phi_log(gumbel_lcdf(x[n, 1] | mu[1], sigma[1]));
      y[2, n] = inv_Phi_log(lognormal_lcdf(x[n, 2] | mu[2], sigma[2]));
    }
     y ~ multi_normal_cholesky_copula(L);
  }
}
```
My questions to @bob-carpenter, @bgoodri, @SteveBronder, @WardBrian, and @andrjohns is that when we start adding `_qf` we should also add the log input version. This at least doubles the work - though probably more than doubles because I don't think anyone has code/math written out for these. Secondly, what naming convention would be appropriate? This pr has `_log` as the suffix but the typical quantile function will have `_qf` and we don't want log to come after nor do we want log to be before the `qf` (such as `_lqf`) because that seems to suggest a log _output_. 

One obvious choice is to append `l` like `_qfl` and then if we have log-output quantile functions it would be `_lqf` and `_lqfl`.  Or we could be a bit more descriptive and add `lp` for log-probability. Something like `_lpqf` and `lplqf` for log-probability quantile function and log-probability log quantile function, respectively. I think this more appropriately conveys what's happening and I tend to read it as "log-input x". 

This PR can then be aliased to `std_normal_lpqf`.

WardBrian:
Until 2.32, _log is still used for probability distribution functions, and anything that isn’t one of those ends up being special cased in the compiler. This is why things like multiply_log are renamed/deprecated (though, in theory, we only needed to deprecate either their use in probability functions or their use elsewhere to get sane behavior, but both have been deprecated). 

I think lmultiply already breaks the convention, as the log is applied to one of the two inputs

spinkney:
@WardBrian this case is quite different in that the log suffix signifies that the function takes the log of the traditional input but outputs the same.

WardBrian:
Ah, I see. I think conveying that information via a suffix alone won’t be possible, people will need to read the doc to use it properly

If anything, such different behavior calls for an entirely new nomenclature so that users don’t try to put it into a pre existing notion they have from other functions.  Maybe like inv_Phi_logscale? I’m not terribly creative about such things

spinkney:
Let me change the name to `std_normal_log_qf`, I'm happy to have consensus and to make that change!

bob-carpenter:
`multiply_log(x, y) = x * log(y)`.  So the log was an operation applied, not the scale of the argument.   `lmultiply` is a common name for this function, but it doesn't follow our more verbose conventions.

spinkney:
@WardBrian does this stay in the `fun` directory or do we want `_qf` functions in the `prob` directory?

WardBrian:
That’s a good question but I defer to someone with more experience in this repo like @bob-carpenter @SteveBronder

bob-carpenter:
The _qf functions should go in the prob directory.  That way, we'll have all the foo_suffix functions together for distribution foo.

spinkney:
> The _qf functions should go in the prob directory. That way, we'll have all the foo_suffix functions together for distribution foo.

Alright, that's what I thought but it comes with some complications. This function is different than the current `prob` functions in that it is vectorized in both the input and output. There is a rev specialty but our prob functions have the partials written in to them. I would like to make a `prob` subdirectory in the `rev` directory so I can just move this function over into `prim -> prob` and `rev -> prob` but I don't know if that will have any unintended side effects. 

Or I can try writing it all in `prim -> prob` with the partials in and a vectorized return. I'm not sure how this is supposed to look because all the prob functions return a scalar. 

I'm asking two questions:
* Should I make a `rev -> prob` folder which houses all the rev-derivatives for `_qf` and hypothesized `_v*` (as in vectorized versions of lpdf, lcdf, etc) prob functions
* or vectorize the prob function and have the partials written in `prim` like the current prob functions?

bob-carpenter:
Forget what I said.  I have no idea how the devs currently want things organized, so I suggest asking someone more actively working on Stan's C++ than me.

spinkney:
@rok-cesnovar would you be willing to work with me on this?

Issue #2739, opened 2022-05-13
Add Method stan::math::gradient, copy gradient into an array instead of VectorXd


roualdes:
## Description

I'm using Julia to call a C function which calls a C++ function which uses ``stan::math::gradient``; details on [discourse.mc-stan.org](https://discourse.mc-stan.org/t/call-stan-math-gradient-from-separate-threads-segfault/27458). I can reuse memory from the Julia vectors because they come in to the C function as arrays ``double* a``, and one can map an array to ``Eigen::VectorXd`` via ``Eigen::Map<Eigen::VectorXd>``.  This enables direct re-use of memory.  Great.

The same is not true for the result.  The method [gradient](https://github.com/stan-dev/math/blob/develop/stan/math/rev/functor/gradient.hpp) only allows copying the gradient vector into ``Eigen::VectorXd``.  I'd like to add a new method which copies the gradient vector into an array.  Something like

```
void gradient(const F& f, const Eigen::Matrix<double, Eigen::Dynamic, 1>& x,
              double& fx, double* grad_fx)
```

It would be up to the caller to ensure the ``grad_fx`` has the appropriate size.

With this addition I could avoid initializing an extra ``Eigen::VectorXd grad_fx`` just to copy it into the Julia vector in which I want the result. 

#### Current Version:
v4.3.2

PR #2743, opened 2022-05-19
From issue #2739, add new function stan::math::gradient

roualdes:
## Summary

Add new function ``stan::math::gradient`` into rev/functor, which stores the evaluated gradient into an array.  The other function stores the evaluated gradient into a ``Eigen::VectorXd``.  These two functions have different signatures, based on the type in which the gradient is stored, and so shouldn't conflict.

Re issue #2739.

## Tests

I duplicated the tests for the function which stores into an ``Eigen::VectorXd`` and modify them to store into an array.  This isn't to say that I completely understand all the details of what each of those tests does.  Maybe @wds15 could opine.

## Side Effects

None to my knowledge.

## Release notes

New gradient function allows evaluated gradient to be stored into an array, enabling memory reuse when stan/math called via a FFI.

## Checklist

- [x] Math issue #2739

- [x] Copyright holder: Edward A. Roualdes

- [x] the new changes are tested: ./runTests.py test/unit/math/rev/functor/gradient_test.cpp


bob-carpenter:
A more generic way to do this is with iterators, like the constructor for `std::vector`:

```cpp
template< class InputIt >
constexpr vector( InputIt first, InputIt last, 
                  const Allocator& alloc = Allocator() );
```

We would only need `last` for error checking and we don't need an allocator.

roualdes:
Hey @bob-carpenter, is this along the lines of what you had in mind?

The way I wrote it, which might not coincide with what you had in mind, this is more general and slick, but also kind of ugly.  We've got to iterate over both the elements of `x_var` and the elements of (what was) `grad_fx`.  If we use iterators to iterate over the elements of `grad_fx`, then we still need an iterator over the elements of `x_var`.  So I crammed two iterators in one for-loop, but I'm not too keen on it.

Please do let me know if you see a cleaner solution that I missed.  Thanks.

roualdes:
This is ready for review.  @wds15 would you mind reviewing this? There is one new function and then mostly tests.  I attempted to replicate the gradient tests within the threaded AD, but (I assume) you'd be a good pair of second eyes to look this over.  Let me know.  If you don't have time or whatever, I'll ping somebody else.  Thanks.

stan/math/rev/functor/gradient.hpp
```diff
@@ -51,6 +55,67 @@ void gradient(const F& f, const Eigen::Matrix<double, Eigen::Dynamic, 1>& x,
...
+    throw std::invalid_argument(s.str());
+  }
+
+  Eigen::Matrix<var, Eigen::Dynamic, 1> x_var(x);
+  var fx_var = f(x_var);
+  fx = fx_var.val();
+  grad(fx_var.vi_);
+  for (Eigen::Matrix<var, Eigen::Dynamic, -1>::Index i = 0; i < x_var.size();
+       ++i) {
+    *first_grad_fx++ = x_var(i).adj();
```
SteveBronder, line 115: This is nbd / optional but using `.coeff()` will avoid a out of bounds check you are already doing manually in the above

```suggestion
    *first_grad_fx++ = x_var.coeff(i).adj();
```

test/unit/math/rev/functor/gradient_test.cpp
```diff
@@ -35,6 +35,19 @@ TEST(RevFunctor, gradient) {
   EXPECT_FLOAT_EQ(x(0) * x(0) + 3 * 2 * x(1), grad_fx(1));
 }
 
+TEST(RevFunctor, gradient_array) {
+  fun1 f;
+  Matrix<double, Dynamic, 1> x(2);
+  x << 5, 7;
+  double fx;
+  double grad_fx[2];
+  stan::math::gradient(f, x, fx, std::begin(grad_fx), std::end(grad_fx));
```
SteveBronder, line 44: Would prefer not using plain array. Nothing wrong with it but we don't do it anywhere else so if you can change this in other places that would be nice

```suggestion
  std::vector<double> grad_fx(2, 0);
  stan::math::gradient(f, x, fx, std::begin(grad_fx), std::end(grad_fx));
```

test/unit/math/rev/functor/gradient_test.cpp
```diff
@@ -120,6 +133,95 @@ TEST(RevFunctor, gradient_threaded) {
...
+    return res;
+  };
+
+  // schedule a bunch of jobs which all do the same
+  std::vector<std::future<VectorXd>> ad_futures_ref;
+
+  for (std::size_t i = 0; i < 100; i++) {
+    // the use pattern in stan-math will be to defer the first job in
+    // order to make the main thread do some work which is why we
+    // alter the execution policy here
```
SteveBronder, line 173:
```suggestion
   /*
    * the use pattern in stan-math will be to defer the first job in
    * order to make the main thread do some work which is why we
    * alter the execution policy here
    */
```

SteveBronder (changes requested):
V small change to the tests just to keep things how they are normally done in Stan (for future readers). But once those are fixed then I think this is good!

roualdes:
Thanks @SteveBronder for reviewing this.  Please also look over lines 97 and 113 of stan/math/rev/functor/gradient.hpp.  I remembered that passing an Eigen Map into a function expecting a Matrix will implicitly copy, and thus introduced a Ref class in the function signature.

Also, the for loop iterator was needlessly complicated.

stan/math/rev/functor/gradient.hpp
```diff
@@ -51,6 +55,68 @@ void gradient(const F& f, const Eigen::Matrix<double, Eigen::Dynamic, 1>& x,
...
+ * @param[out] fx Function applied to argument
+ * @param[out] grad_fx Gradient of function at argument
+ * @throw std::invalid_argument if the iterator isn't the right size
+ * to hold the gradients
+ */
+template <typename F, typename InputIt>
+void gradient(
+    const F& f,
+    const Eigen::Ref<const Eigen::Matrix<double, Eigen::Dynamic, 1>>& x,
+    double& fx, InputIt first_grad_fx, InputIt last_grad_fx) {
```
SteveBronder, line 100: As long as `x` is only being used for 

```cpp
 Eigen::Matrix<var, Eigen::Dynamic, 1> x_var(x);
```

Then I think the below should work

```suggestion
template <typename F, typename EigVec, typename InputIt,
  require_eigen_vector_vt<std::is_arithmetic, EigVec>* = nullptr>
void gradient(const F& f, const EigVec& x, double& fx, InputIt first_grad_fx, InputIt last_grad_fx) {
```

Also you need docs for `InputIt`. I think it's fine to take the docs from cppreference like 

```
@tparam InputIt must meet the requirements of [LegacyInputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator).
```

SteveBronder (changes requested):
Saw one more little thing with the most recent change that needs fixed then I think it's good to merge

stan/math/rev/functor/gradient.hpp
roualdes: This is great.  Docs for sure.  Good catch.  You mind if I ask a few follow up questions first, just to make sure I understand all the things?

1. Understanding. Written this way the compiler will write a function that takes necessarily an Eigen vector in place of EigVec.  So if I pass in a Matrix it will write out a function for an Eigen Matrix.  And, importantly for my use case, if I pass in an Eigen Map, it will write a function for an Eigen Map and thus will avoid any copies (implicit or otherwise), right?
2. Limitations.  You say, "As long as x is only being used for `Eigen::Matrix<var, Eigen::Dynamic, 1> x_var(x);` then this should work."  If `const EigVec& x` is an Eigen vector, then we should be able to do anything with `x` we can do with an Eigen vector, right?  What other limitations do you have in mind?
3. Tests.  Presumably `require_eigen_vector_vt` is tested somewhere, so we do not need to test every new use of it, correct?

SteveBronder: No problem!


> Written this way the compiler will write a function that takes necessarily an Eigen vector in place of EigVec. So if I pass in a Matrix it will write out a function for an Eigen Matrix. And, importantly for my use case, if I pass in an Eigen Map, it will write a function for an Eigen Map and thus will avoid any copies (implicit or otherwise), right?

Yep! Eigen::Ref<> does something very similar, though we don't use it much in the math library (for a long time the docs and how it chose to make copies was pretty confusing to us). So I think the general template with a `require` restriction template to limit it to Eigen types with 1 compile time row or column would be good here.

>    Limitations. You say, "As long as x is only being used for Eigen::Matrix<var, Eigen::Dynamic, 1> x_var(x); then this should work." If const EigVec& x is an Eigen vector, then we should be able to do anything with x we can do with an Eigen vector, right? What other limitations do you have in mind?

So with `Eigen::Ref<>` directly it is true that your input can be treated as an Eigen::Vector with no worries, but if a function uses a general template for an Eigen type then we can end up passing Eigen expressions into functions. For example, doing the same coefficient access twice for an eigen expression will compute the expression twice. For functions that can take in expressions you need to use Stan's `to_ref()` function to do the equivalent of having `Eigen::Ref<>` in the signature.

But in your case we are just reading from the vector so idt it's necessary to use `to_ref()` here. 

> Tests. Presumably require_eigen_vector_vt is tested somewhere, so we do not need to test every new use of it, correct?

Yes no need to test a new use of it. You can add tests to check the row vectors now work as well, or you can just restrict your require to instead be `require_eigen_column_vector_vt<std::is_arithmetic, EigVec>` which will only support Eigen vectors

SteveBronder (approved):
lgtm!

PR #2751, opened 2022-06-03
[WIP] Allow distributions to return vectors

SteveBronder:
## Summary

This is a prototype for discussion about allowing our distributions to return back vectors instead of just scalars. I only did `normal_lpdf` as an example of how this would be implemented for the other functions. My goal was to try to reduce the amount of possible code duplication while still keeping our optimizations for the case where we are returning back a scalar type. 

The main components here are

1. A new template parameter `ReturnType` to the lpdf that is either `ProbReturnType::Scalar` or `ProbReturnType::Vector`
2.   a new class `prob_reducer`* that for `ProbReturnType::Scalar` a scalar or for `ProbReturnType::Vector` holds an Eigen array. For the scalar case, things like `+=` etc. will do summation when the rhs is an Eigen type. For `ProbReturnType::Vector` when operators such as `+=` see a scalar they propogate the scalar over the entire return Eigen array.

Getting this to work nicely does require a little bit of oddities, for instance in `normal_lpdf` we previously had 

```cpp
return_t logp = sum(-0.5 * y_scaled_sq);
```

But now we may want the rhs not summed, but `y_scaled_sq` could also just be a scalar. So now we sometimes need to construct `logp` with an explicit constructor telling it the size we want.

```cpp
prob_return_t<RetType, T_partials_return> logp(-0.5 * y_scaled_sq, N);
```

So that if `y_scaled_sq` is a scalar it propogates that for each value in the return matrix of size `N`

* `prob_reducer` is probably a bad name and a new one is very welcome. I would like a name that reflects that the class either holds a scalar and then does summations over vector inputs or is a vector which then just returns a vector at the end. Maybe `conditional_reducer`?

## Tests

After stealing a few things from the prob tests I was able to get a version working with the `expect_ad` test framework. A new function `expect_ad_distribution` applies the user input over all the possible combinations of types that can go into the probability distributions.

You can see an example of this in the mix test for the normal distribution

```
./runTests.py test/unit/math/mix/prob/normal_test.cpp
```

## Side Effects

I'm not sure about side effects necessarily but this is a big change that we should be very careful with.

## Release notes

Adds framework for probability distributions to return back scalars or Eigen vectors.

## Checklist

- [ ] Math issue #(issue number) (I couldn't find the issue but I know people have been requesting this forever and there is an issue somewhere)

### DIFF
--- a/stan/math/fwd/functor/operands_and_partials.hpp
+++ b/stan/math/fwd/functor/operands_and_partials.hpp
@@ -27,6 +27,7 @@ class ops_partials_edge<Dx, fvar<Dx>> {
   const Op& operand_;
 
   Dx dx() { return this->partials_[0] * this->operand_.d_; }
+  Dx dx_v() { return this->partials_[0] * this->operand_.d_; }
 };
 }  // namespace internal
 
@@ -106,6 +107,13 @@ class operands_and_partials<Op1, Op2, Op3, Op4, Op5, fvar<Dx>> {
         = edge1_.dx() + edge2_.dx() + edge3_.dx() + edge4_.dx() + edge5_.dx();
     return T_return_type(value, deriv);
   }
+
+  template <typename EigVec, require_eigen_vector_t<EigVec>* = nullptr>
+  auto build(EigVec&& value) {
+    Eigen::Array<fvar<Dx>, -1, 1> ret(value.template cast<fvar<Dx>>());
+    ret.d() = (edge1_.dx_v() + edge2_.dx_v() + edge3_.dx_v() + edge4_.dx_v() + edge5_.dx_v());
+    return ret;
+  }
 };
 
 namespace internal {
@@ -135,6 +143,14 @@ class ops_partials_edge<Dx, std::vector<fvar<Dx>>> {
     }
     return derivative;
   }
+  Eigen::Array<Dx, -1, 1> dx_v() {
+    Eigen::Array<Dx, -1, 1> derivative(this->operands_.size());
+    for (size_t i = 0; i < this->operands_.size(); ++i) {
+      derivative[i] = this->partials_[i] * this->operands_[i].d_;
+    }
+    return derivative;
+  }
+
 };
 
 template <typename Dx, int R, int C>
@@ -160,6 +176,10 @@ class ops_partials_edge<Dx, Eigen::Matrix<fvar<Dx>, R, C>> {
       derivative += this->partials_(i) * this->operands_(i).d_;
     }
     return derivative;
+  }
+
+  Eigen::Array<Dx, -1, 1> dx_v() {
+    return this->partials_.array() * this->operands_.d().array();
   }
 };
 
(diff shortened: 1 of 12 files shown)


SteveBronder:
While I can help get in a framework for this, idt I have time to do this for all of the distributions. Though I do think this could be a cool GSOC project for a summer

bob-carpenter:
> Is there any proposal for how this feature would be exposed in the language?

We'd just use another decorator, like `_vec`, as in `normal_lpdf_vec`.  I don't think we want to start decorating function calls with special syntax and this isn't an annotation thing as it affects the language.

An alternative would be a general vectorization function:

```
T0 f(T1, T2, ..., TN)

array[] T0 vectorize(f, (T1 | array[] T1) xs1, array[] (T2 | array[] T2) xs2, ..., (TN | array[] TN) xsN);
```

Then the use would be

```
vector[N] lp = vectorize(normal_lupdf, y, mu, sigma);
```

where now `y`, `mu`, and `sigma` can be scalars or arrays (or maybe vectors, too).

The implementation can just create a view of each argument then apply `f` in a loop.  There's no easy way to do this efficiently, though, as we can't share evaluations like `log(sigma)` in vectorizing `normal_lpdf`.

betanalpha:
We’re kind of in an awkward corner with the conventions that density functions collapse the broadcast evaluation by default.  To me “vectorization” has always been a bit confusion and something like “nosum” or “nocollapse” or “expanded” would be more explicit, but given the existing conventions “vec_lpdf” or “lpdf_vec” wouldn’t be the worst.

bob-carpenter:
I agree that "vectorization" is ridiculously overloaded.  It technically refers to how the CPU pipelines operations in SSE, AVX, etc.    Our general use for functions like `exp(vector)` is already a stretch in that we're really doing what is called "type lifting" in the programming languages literature---we just map the function elementwise in this case.

For what it's worth, I prefer `lpdf_vec`, but have no idea why.  Either would work.

SteveBronder:
Yeah it's weird and hard to name. Even vec seems weird. What about `for_each`? Like `normal_lpdf_for_each(...)`? I'm also fine with `lpdf_vec`

bob-carpenter:
To me, `_for_each` sounds like it's saying the behavior is a loop.   While that's technically true and therefore the name makes sense, I find it too long and not quite evocative enough of the different return type.

Something like `_ind` or `_item` would make sense to return item-level densities, but I don't think they're as clear as `_vec`.

betanalpha:
> To me, _for_each sounds like it's saying the behavior is a loop. While that's technically true and therefore the name makes sense, I find it too long and not quite evocative enough of the different return type.
> 
> Something like _ind or _item would make sense to return item-level densities, but I don't think they're as clear as _vec.
> 

While “for_each” might not be my favorite if we were redesigning the language I actually kind of like it in the context of the existing language.  In particular I think that most users, especially those with less formal programming experience, would interpret “vectorization/broadcasting/lifting to a product space” as a for loop, making “for each” about as intuitive as possible.

In any case this is a big enough decision that it’s probably worth a forum discussion to elicit more feedback from users.

PR #2762, opened 2022-06-24
StanHeaders 2.26: Scalar falling out of scope

andrjohns:
@hsbadr 

Some tests with the `geostan` package were failing because a captured scalar was falling out of scope in `apply_scalar_binary` under RStan 2.26, this PR removes the `ifdef`s and always evaluates

### DIFF
--- a/stan/math/prim/functor/apply_scalar_binary.hpp
+++ b/stan/math/prim/functor/apply_scalar_binary.hpp
@@ -201,7 +201,11 @@ inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
 template <typename T1, typename T2, typename F, require_eigen_t<T1>* = nullptr,
           require_stan_scalar_t<T2>* = nullptr>
 inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
+#ifdef USE_STANC3
+  return x.unaryExpr([&f, y](const auto& v) { return f(v, y); });
+#else
   return x.unaryExpr([&f, &y](const auto& v) { return f(v, y); }).eval();
+#endif
 }
 
 /**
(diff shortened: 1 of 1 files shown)


hsbadr:
Thanks @andrjohns! Have you tested this with RStan v2.26 in other conditions? It looks like this won't affect v2.21 (`USE_STANC3` should be undefined). So, it's good to go if it doesn't break v2.26 & fixes the mentioned issue.

stan/math/prim/functor/apply_scalar_binary.hpp
```diff
@@ -201,11 +201,7 @@ inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
 template <typename T1, typename T2, typename F, require_eigen_t<T1>* = nullptr,
           require_stan_scalar_t<T2>* = nullptr>
 inline auto apply_scalar_binary(const T1& x, const T2& y, const F& f) {
-#ifdef USE_STANC3
-  return x.unaryExpr([&f, &y](const auto& v) { return f(v, y); });
-#else
   return x.unaryExpr([&f, &y](const auto& v) { return f(v, y); }).eval();
```
hsbadr, line 204: I think that this was the default for v2.26 (using `stanc3`). The evaluation was required for backward compatibility (i.e., using `stanc2` in the CRAN version of RStan v2.21). So, we need to make sure that this doesn't break C++ code generated by `stanc3`, other that the mentioned package.

andrjohns: Ah the issue wasn't the `.eval()`! The problem is that the `2.26` code is capturing the provided scalar by reference:
```cpp
[&f, &y](const auto& v)
```

But it needs to capture it by value, which is what the current [`develop` code does](https://github.com/stan-dev/math/blob/882cd229970f098ca078cb5b87873967dc94698f/stan/math/prim/functor/apply_scalar_binary.hpp#L192)

I've updated the PR to do this (a much more minor change). The 2.26 behaviour of capturing by reference would be considered a bug that we've since fixed in the Math library

hsbadr (approved):
LGTM. Thanks @andrjohns!

hsbadr:
@andrjohns This is now included in [RStan v2.26.12](https://github.com/stan-dev/r-packages/commit/ea5568ed15a60e87fee8b8fb4c422bcb1a3c4ae6). I'll rerun reverse dependencies ASAP.

PR #2765, opened 2022-06-25
StanHeaders 2.26 - Additional Evals

andrjohns:
@hsbadr 

This PR adds some additional evals to fix compilation errors with the `geostan` and `epidemia` packages. The errors were caused by the user-defined functions only accepting Eigen plain types, but there were still some Math functions returning expressions (`add` for `epidemia` and `rep_vector` for `geostan`)

### DIFF
--- a/stan/math/prim/fun/add.hpp
+++ b/stan/math/prim/fun/add.hpp
@@ -62,7 +62,11 @@ template <typename Mat, typename Scal, require_eigen_t<Mat>* = nullptr,
           require_stan_scalar_t<Scal>* = nullptr,
           require_all_not_st_var<Mat, Scal>* = nullptr>
 inline auto add(const Mat& m, const Scal c) {
+#ifdef USE_STANC3
   return (m.array() + c).matrix();
+#else
+  return (m.array() + c).matrix().eval();
+#endif
 }
 
 /**
--- a/stan/math/prim/fun/rep_matrix.hpp
+++ b/stan/math/prim/fun/rep_matrix.hpp
@@ -12,20 +12,33 @@ template <typename T, require_stan_scalar_t<T>* = nullptr>
 inline auto rep_matrix(const T& x, int m, int n) {
   check_nonnegative("rep_matrix", "rows", m);
   check_nonnegative("rep_matrix", "cols", n);
+#ifdef USE_STANC3
   return Eigen::Matrix<return_type_t<T>, Eigen::Dynamic,
                        Eigen::Dynamic>::Constant(m, n, x);
+#else
+  return Eigen::Matrix<return_type_t<T>, Eigen::Dynamic,
+                       Eigen::Dynamic>::Constant(m, n, x).eval();
+#endif
 }
 
 template <typename ColVec, require_eigen_col_vector_t<ColVec>* = nullptr>
 inline auto rep_matrix(const ColVec& v, int n) {
   check_nonnegative("rep_matrix", "rows", n);
+#ifdef USE_STANC3
   return v.replicate(1, n);
+#else
+  return v.replicate(1, n).eval();
+#endif
 }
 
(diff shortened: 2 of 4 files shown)


hsbadr:
Thanks @andrjohns ! Could you also look at the failure of `stanette` package after the last patch? See https://github.com/stan-dev/rstan/commit/ec5f4a82e6399886eb2927fec985315a2245c45a

andrjohns:
Can do, looking now!

hsbadr:
Perfect! Hopefully, we can fix it too and rerun `revdepcheck`.

hsbadr (approved):
Merging this for testing reverse dependencies. Then, we can start from the new report.

andrjohns:
Figured out the `stanette` issue, will have another `StanHeaders` PR through in a couple minutes

andrjohns:
Also, I've got a PR open for `ProbReco` to add the linker flags [here](https://github.com/anastasiospanagiotelis/ProbReco/pull/4)

andrjohns:
And it looks like `PosteriorBootstrap` passes `rcmdcheck` with both 2.21 & 2.26 and 2.26 & 2.26 across Mac, Win, and Linux: https://github.com/andrjohns/PosteriorBootstrap/actions/runs/2561614694

So I'm not sure how the `revdep` testing environment is different, and if that might be where the errors are coming from

The `Ubuntu (old-rel)` failures are unrelated, and something related to the `MASS` package

hsbadr:
Let's wait for the new revdep report.

andrjohns:
Sounds good!

hsbadr:
Note that tests should be with the different combinations of RStan & StanHeaders x v2.21 & v2.26. We'll release StanHeaders 1st, which needs to be compatible with the CRAN version of RStan v2.21. Then, RStan v2.26 will follow.

andrjohns:
Yep, I've got the workflow setup to test:

 - 2.21 & 2.21
 - 2.21 & 2.26
 - 2.26 & 2.26

So it should cover all bases

hsbadr:
Yes, it could be related to the system I use to run revdep checks. It ends up with a segmentation fault:
```r
  Running ‘testthat.R’
 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
   1. ├─base::suppressWarnings(...) at test_anpl.R:177:2
   2. │ └─base::withCallingHandlers(...)
   3. ├─rstan::vb(...)
   4. └─rstan::vb(...)
   5.   └─rstan .local(object, ...)
   6.     └─sampler$call_sampler(c(args, dotlist))
  
  [ FAIL 1 | WARN 0 | SKIP 3 | PASS 51 ]
  Error: Test failures
  Execution halted
  
   *** caught segfault ***
  address 0x18, cause 'memory not mapped'
  An irrecoverable exception occurred. R is aborting now ...
  Segmentation fault (core dumped)
```

hsbadr:
So, we may ignore retest the segfaults on another machine & fix the obvious errors, related to Stan/Math or the backward compatibility patches.

andrjohns:
Sounds like a plan! Let me know if the revdep checks find any other issues and I'll chase them down

hsbadr:
Here's the update revdep results: https://github.com/stan-dev/rstan/tree/StanHeaders_2.26/revdep

`ProbReco` can be fixed by importing either `RcppParallel` or `StanHeaders` (in NAMESPACE). It compiles fine but needs to find TBB symbols at loading time. The tests of 4 (`IgGeneUsage`, `PosteriorBootstrap`, `prophet`, and `rstanarm`) other packages cause segfaults, which could be due to my system environment (system libraries, compiler version, ... etc.).  Those packages can be tested on another machine. Also, `IgGeneUsage` is BioConductor and `rstanarm` is ours. So, CRAN should accept v2.26.

PR #2766, opened 2022-06-25
StanHeaders 2.26 - stanette fix

andrjohns:
@hsbadr 

The `stanette` failures were caused by headers in the package calling `normal_rng` but only including the `prim/mat.hpp` Math header. This adds `prob.hpp` to that Math header so that `normal_rng` can be found


hsbadr (approved):
LGTM

Issue #2770, opened 2022-06-28
Add function for casting (data) reals to ints


andrjohns:
## Description

A request that's come up a few times on the forum (and in my own models) has been the ability to cast a `real` input to an `int`. The general push-back on this has been the same as for the `fabs`/`step` functions - breaking differentiability. However, given that we can use the `data real` keyword in Stan, we can specify the function to only take `double` and `std::vector<double>` inputs

#### Current Version:
v4.4.0

rok-cesnovar:
I think for data reals this is a great idea.

andrjohns:
The implementation is using `static_cast<int>(x)`, which should truncate the decimal rather than dealing with rounding. The implementation also throws if the input is too large to be represented as integer

bob-carpenter:
Thanks.  I think that's equivalent to rounding toward zero.   Won't that be problematic with the 0.999999999 values of the world?

You must be writing in your own exception handling because `static_cast` is static and doesn't throw.

Another option would be to throw if it isn't a round integer in its double representation.  That's partly why I was asking about use cases.

PR #2771, opened 2022-06-28
Add to_int function and tests

andrjohns:
## Summary

This PR adds a function for casting `double` and `std::vector<double>` inputs to `int` types. This is intended to only be exposed for use with `data real` and `array[] data real` types in Stan

## Tests

Prim tests added for correctness of return as well as throwing if the input is too large to be represented as an `int` without overflowing

## Side Effects

N/A

## Release notes

Added function for casting `double` and `std::vector<double>` types to `int` and `std::vector<int>` types

## Checklist

- [x] Math issue #2770 

### DIFF
--- a/stan/math/prim/fun/to_int.hpp
+++ b/stan/math/prim/fun/to_int.hpp
@@ -35,7 +35,7 @@ inline T to_int(T x) {
  *
  * @tparam T type of argument (must be arithmetic)
  * @param x argument
- * @return iIteger value of argument
+ * @return Integer value of argument
  */
 template <typename T, require_floating_point_t<T>* = nullptr>
 inline int to_int(T x) {
--- a/stan/math/prim/fun/to_int.hpp
+++ b/stan/math/prim/fun/to_int.hpp
@@ -1,25 +1,52 @@
 #ifndef STAN_MATH_PRIM_FUN_TO_INT_HPP
 #define STAN_MATH_PRIM_FUN_TO_INT_HPP
 
+#include <stan/math/prim/err/check_not_nan.hpp>
+#include <stan/math/prim/err/check_finite.hpp>
 #include <stan/math/prim/functor/apply_scalar_unary.hpp>
 
 namespace stan {
 namespace math {
 
 /**
- * Returns the input scalar as an integer type
+ * Returns the input scalar as an integer type. Specialisation for integral
+ * types which do not need conversion, reduces to a no-op.
+ *
+ * @tparam T type of integral argument
+ * @param x argument
+ * @return Input argument unchanged
+ */
+template <typename T, require_integral_t<T>* = nullptr>
+inline T to_int(T x) {
+  return std::forward<T>(x);
+}
+
+/**
+ * Returns the input scalar as an integer type. This function performs no
+ * rounding and simply truncates the decimal to return only the signficand as an
+ * integer.
+ *
+ * Casting NaN and Inf values to integers is considered undefined behavior as
+ * NaN and Inf cannot be represented as an integer and most implementations
+ * simply overflow, as such this function throws for these inputs.
+ *
+ * The function also throws for floating-point values that are too large to be
+ * represented as an integer.
  *
  * @tparam T type of argument (must be arithmetic)
  * @param x argument
  * @return iIteger value of argument
  */
-template <typename T, require_arithmetic_t<T>* = nullptr>
+template <typename T, require_floating_point_t<T>* = nullptr>
 inline int to_int(T x) {
+  static const char* function = "to_int";
+  check_not_nan(function, "x", x);
+  check_finite(function, "x", x);
   if (x < std::numeric_limits<int>::min()
       || x > std::numeric_limits<int>::max()) {
     std::ostringstream msg;
     msg << "Value " << x << " is too large to be represented as an integer";
-    throw std::invalid_argument(msg.str());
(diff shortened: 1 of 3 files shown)


stan/math/prim/fun/to_int.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace stan {
+namespace math {
+
+/**
+ * Returns the input scalar as an integer type
+ *
+ * @tparam T type of argument (must be arithmetic)
+ * @param x argument
+ * @return iIteger value of argument
```
bob-carpenter, line 14: typo here

stan/math/prim/fun/to_int.hpp
```diff
@@ -0,0 +1,59 @@
...
+ */
```
bob-carpenter, line 15: need to document when it throws

test/unit/math/prim/fun/to_int_test.cpp
```diff
@@ -0,0 +1,41 @@
...
+  EXPECT_EQ(-36574, to_int(-36574.0));
+  EXPECT_EQ(-36574, to_int(-36574.1));
+  EXPECT_EQ(-36574, to_int(-36574.9));
+  EXPECT_EQ(-36574, to_int(-36574.999999999));
+
+  EXPECT_THROW(to_int(std::numeric_limits<int>::max() + 1.0),
+               std::invalid_argument);
+  EXPECT_THROW(to_int(std::numeric_limits<int>::min() - 1.0),
+               std::invalid_argument);
+}
```
bob-carpenter, line 24: Need to test NaN behavior.

It'd be nice to test +/- inf behavior.

test/unit/math/prim/fun/to_int_test.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+TEST(MathFunctions, to_int_vec) {
+  using stan::math::to_int;
+
```
bob-carpenter, line 28: test empty vectors and size 1 vectors (this should all work given the way it's defined, but the boundary tests should still be in place).

I'm not sure we really need to test bigger containers or other types of containers like `vector<MatrixXd>`, etc.

stan/math/prim/fun/to_int.hpp
```diff
@@ -0,0 +1,59 @@
...
+template <typename T, require_arithmetic_t<T>* = nullptr>
+inline int to_int(T x) {
```
bob-carpenter, line 17: [optional]
probably not worth specializing, but this is an inefficient implementation for `int` inputs because of the unnecessary tests

stan/math/prim/fun/to_int.hpp
```diff
@@ -0,0 +1,59 @@
+#ifndef STAN_MATH_PRIM_FUN_TO_INT_HPP
+#define STAN_MATH_PRIM_FUN_TO_INT_HPP
+
+#include <stan/math/prim/functor/apply_scalar_unary.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Returns the input scalar as an integer type
```
bob-carpenter, line 10: This needs to be much more specific in terms of how it returns.  You can say it truncates using `static_cast<int>` to be precise, but something needs to say how it converts.

bob-carpenter (changes requested):
this needs a bit more doc and a bit more testing, but otherwise looks good

andrjohns:
The `Rtools40 mix/fun` failures are just the actions hitting the 6hr time limit and being automatically terminated, rather than actual test failures. Given that `to_int()` isn't touching anything in `rev`/`fwd`, and the Jenkins `mix/fun` tests passed, I'd say that this was safe to approve (and merge after the freeze)

stan/math/prim/fun/to_int.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * Casting NaN and Inf values to integers is considered undefined behavior as
+ * NaN and Inf cannot be represented as an integer and most implementations
+ * simply overflow, as such this function throws for these inputs.
+ *
+ * The function also throws for floating-point values that are too large to be
+ * represented as an integer.
+ *
+ * @tparam T type of argument (must be arithmetic)
+ * @param x argument
+ * @return Integer value of argument
```
bob-carpenter, line 38: needs to doc @throw when out of bounds or NaN

stan/math/prim/fun/to_int.hpp
```diff
@@ -0,0 +1,86 @@
...
+ */
+template <typename T, require_floating_point_t<T>* = nullptr>
+inline int to_int(T x) {
+  static const char* function = "to_int";
+  check_not_nan(function, "x", x);
+  check_finite(function, "x", x);
+  if (x < std::numeric_limits<int>::min()
```
bob-carpenter, line 45: the check-finite test is redundant and the check-not-nan is also redundant if you rewrite as:

```cpp
if (!(x >= std::numeric_limits<int>::min() && x <= std::numeric_limits<int>::max())) { ... }
```
And the message should be that the value is not in range to be converted to integer (it's either too large if it's above max or too small if it's below min and it doesn't fit if it's NaN).

bob-carpenter (changes requested):
really just a minor doc change and message change (and optionally cutting out the nan and finite tests)

bob-carpenter (approved):

PR #2787, opened 2022-07-08
Allow varmat for owens_t 

SteveBronder:
## Summary

This adds an overload of `owens_t()` that works for mixes of `var<Matrix>`, `Matrix<var/double>`, and `var/double` types.

@andrjohns I added you as a reviewer here because I could use your help brainstorming a nice way to turn off the binary vectorization framework for when we have specializations that works for mixes of the types above. I think whatever I do here I would do for other binary functions like `log_sum_exp` and friends. Right now I just wrote `owens_t()` in rev to work with all of the above (as long as one input has a `var` scalar type and none are standard vectors. I think I could pretty easily take this pattern and apply it to the rest of the binary functions. Though I think binary functions that accept integers / arrays of integers as an input will still need special care.

## Tests

The `owens_t` test in `mix` has new tests for combinations of matrices and scalars and their vectorized versions as well. It can be run with the following

```
./runTests.py ./test/unit/math/mix/fun/owens_t_test 
```

## Side Effects

@andrjohns I'm curious if you can think of a nicer way to turn off the binary vectorization framework when we don't need it. I _think_, as long as I use the same pattern for other binary functions this should be fine.

One thing we may also want to do is still explicitly have `owens_t(var/double, var/double)` signatures. I did not test it but I would guess that there would be a small slowdown from using the `reverse_pass_callback()` pattern

## Release notes

Adds varmat overload for `owens_t`

## Checklist

### DIFF
--- a/stan/math/prim/meta/is_var_and_matrix_types.hpp
+++ b/stan/math/prim/meta/is_var_and_matrix_types.hpp
@@ -1,0 +1,26 @@
+#ifndef STAN_MATH_PRIM_META_IS_VAR_AND_MATRIX_TYPES_HPP
+#define STAN_MATH_PRIM_META_IS_VAR_AND_MATRIX_TYPES_HPP
+
+#include <stan/math/prim/meta/is_var.hpp>
+#include <stan/math/prim/meta/is_matrix.hpp>
+
+namespace stan {
+
+/** \ingroup type_trait
+ * Extends std::true_type when instantiated with one type that has a var `scalar_type` and
+ * another that is a matrix. Extends std::false_type otherwise.
+ * @tparam Types Types to test
+ */
+template <typename... Types>
+using is_var_and_matrix_types = bool_constant<is_var<return_type_t<Types...>>::value &&
+  stan::math::disjunction<is_matrix<Types>...>::value>;
+
+template <typename... Types>
+using require_all_var_and_matrix_types = require_t<is_var_and_matrix_types<Types...>>;
+
+template <typename... Types>
+using require_all_not_var_and_matrix_types = require_not_t<is_var_and_matrix_types<Types...>>;
+
+
+}  // namespace stan
+#endif
--- a/stan/math/prim/meta/possibly_sum.hpp
+++ b/stan/math/prim/meta/possibly_sum.hpp
@@ -1,0 +1,39 @@
+#ifndef STAN_MATH_PRIM_META_POSSIBLY_SUM_HPP
+#define STAN_MATH_PRIM_META_POSSIBLY_SUM_HPP
+
+
+#include <stan/math/prim/fun/sum.hpp>
+#include <stan/math/prim/meta/require_helpers.hpp>
+
+namespace stan {
+  namespace math{
+
+/**
+ * Conditionally sum the input at compile time.
+ * @tparam CondSum A struct with a static boolean member `value` which if true
+ *  will allow the input value to be summed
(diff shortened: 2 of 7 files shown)


stan/math/prim/meta/is_var_and_matrix_types.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <stan/math/prim/meta/is_var.hpp>
+#include <stan/math/prim/meta/is_matrix.hpp>
+#include <stan/math/prim/meta/require_helpers.hpp>
+#include <stan/math/prim/meta/return_type.hpp>
+
+namespace stan {
+
+/** \ingroup type_trait
+ * Extends std::true_type when instantiated with one type that has a var
+ * `scalar_type` and another that is a matrix. Extends std::false_type
```
andrjohns, line 14: Might be more accurate to say that at least one type has `scalar_type` var and at least one type is a matrix, since the current wording implies that the check only returns true when there is a `var` type and another `matrix` type, not if there's a `matrix<var>` type

stan/math/prim/meta/possibly_sum.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace math {
+
+/**
+ * Conditionally sum the input at compile time.
+ * @tparam CondSum A struct with a static boolean member `value` which if true
+ *  will allow the input value to be summed
+ * @tparam T A scalar, Eigen type, or standard vector with inner scalar type.
+ * @param x The value to be summed.
+ */
+template <typename CondSum, typename T, require_t<CondSum>* = nullptr>
```
andrjohns, line 17: No comments, just here to say that this is neat and I'm looking forward to having it in

stan/math/rev/fun/owens_t.hpp
```diff
@@ -12,57 +12,41 @@
...
+inline auto owens_t(const Var1& h, const Var2& a) {
+  auto h_arena = to_arena(h);
+  auto a_arena = to_arena(a);
+  using return_type
+      = return_var_matrix_t<decltype(owens_t(h_arena.val(), a_arena.val())),
+                            Var1, Var2>;
+  arena_t<return_type> vi = owens_t(h_arena.val(), a_arena.val());
+  reverse_pass_callback([h_arena, a_arena, vi]() mutable {
+    const auto& h_val = as_array_or_scalar(h_arena).val();
+    const auto& a_val = as_array_or_scalar(a_arena).val();
```
andrjohns, line 39:
```suggestion
    const auto& h_val = as_value_array_or_scalar(h_arena);
    const auto& a_val = as_value_array_or_scalar(a_arena);
```

stan/math/rev/fun/owens_t.hpp
```diff
@@ -12,57 +12,41 @@
...
+    const auto neg_avi_sq_div_2 = -square(h_val) * 0.5;
+    const auto one_p_bvi_sq = 1.0 + square(a_val);
```
andrjohns, line 41: Will need to `eval` or `to_ref` these, since they get used more than once below

andrjohns (changes requested):
Only minor suggestions from me! This looks like a good approach for vectorising binary functions, and is pretty much what I've been testing for `log_diff_exp`/`log_sum_exp` as well

andrjohns:
@SteveBronder is this ready for another look over?

SteveBronder:
Yep!

andrjohns (approved):
LGTM

Issue #2664, opened 2022-02-03
Hypergeometric function naming


andrjohns:
## Description

We have some implementations of hypergeometric functions in the Math library ([and more to come](https://github.com/stan-dev/math/pull/2510)), that would be good to expose:

- 2F1: gradients partially implemented (#2663), values implemented in Boost
- 3F2: gradients and functions implemented in Math

I'm proposing that we use a consistent naming convention for these moving forward, matching that used by Boost: `hyper_pFq`. So the functions in the Math library would be named:

- `hyper_2F1`
- `hyper_3F2`

This would involve renaming the `F32` function/header, but given that it's only internal at the moment I don't think it would be much of an issue

spinkney:
Do you want to expose the `grad_2F1` and `grad_3F2` functions as well as `hyper_2F1` and `hyper_3F2`?

andrjohns:
I don't think we would expose the `grad_` functions, since they would have to be rewritten or wrapped to be able to be used in the Stan language (since they currently overwrite their inputs rather than returning outputs). But that option is always there if there turns out to be enough demand for them

betanalpha:
Given how obscure these functions will be for most users, in particular those who wouldn’t be common users of Boost, I would prefer to err on the side of redundancy and go with “hypergeo_pFq”.

To me “hyper” reads as too generic, especially when the Stan language doesn’t have namespaces to provide meaningful context like Boost.
> Description
> 
> We have some implementations of hypergeometric functions in the Math library (and more to come <https://github.com/stan-dev/math/pull/2510>), that would be good to expose:
> 
> 2F1: gradients partially implemented (#2663 <https://github.com/stan-dev/math/issues/2663>), values implemented in Boost
> 3F2: gradients and functions implemented in Math
> I'm proposing that we use a consistent naming convention for these moving forward, matching that used by Boost: hyper_pFq. So the functions in the Math library would be named:
> 
> hyper_2F1
> hyper_3F2
> This would involve renaming the F32 function/header, but given that it's only internal at the moment I don't think it would be much of an issue
>

andrjohns:
I see what you mean. In that case it would probably better to be completely explicit: `hypergeometric_pFq`, since it's it not much more verbose

PR #2792, opened 2022-07-12
Expose hypergeometric_2F1 function

andrjohns:
## Summary

This PR adds the `hypergeometric_2F1` function, using the existing `grad_2f1` function to calculate the gradients.

## Tests

Minimal `mix` tests are added as the values are calculated using `boost::math::hypergeometric_pFq`, which is tested through the `hypergeometric_pFq` tests, and the gradients through `grad_2F1` which already has extensive tests. 

## Side Effects

N/A

## Release notes

Added `hypergeometric_2F1` function

## Checklist

- [x] Math issue #2664

### DIFF
--- a/stan/math/prim/fun/grad_2F1.hpp
+++ b/stan/math/prim/fun/grad_2F1.hpp
@@ -6,16 +6,20 @@
 #include <stan/math/prim/fun/as_array_or_scalar.hpp>
 #include <stan/math/prim/fun/constants.hpp>
 #include <stan/math/prim/fun/exp.hpp>
+#include <stan/math/prim/fun/log1m.hpp>
 #include <stan/math/prim/fun/abs.hpp>
 #include <stan/math/prim/fun/inv.hpp>
 #include <stan/math/prim/fun/log.hpp>
+#include <stan/math/prim/fun/log1m.hpp>
 #include <stan/math/prim/fun/sign.hpp>
 #include <stan/math/prim/fun/hypergeometric_2F1.hpp>
 #include <cmath>
+#include <boost/optional.hpp>
 
 namespace stan {
 namespace math {
 
+namespace internal {
 /**
  * Gradients of the hypergeometric function, 2F1.
  *
@@ -44,11 +48,16 @@ namespace math {
  * @param[in] max_steps number of steps to take
  */
 template <bool calc_a1, bool calc_a2, bool calc_b1, bool calc_z,
-          typename T1, typename T2, typename T3, typename T_z>
-void grad_2F1(T1& g_a1, T2& g_a2, T3& g_b1, T_z& g_z, const T1& a1,
+          typename T1, typename T2, typename T3, typename T_z,
+          typename TupleT = std::tuple<T1, T2, T3, T_z>,
+          typename OptT = boost::optional<TupleT>>
+void grad_2F1_impl(T1& g_a1, T2& g_a2, T3& g_b1, T_z& g_z, const T1& a1,
               const T2& a2, const T3& b1, const T_z& z,
-              double precision = 1e-14, int max_steps = 1e6) {
-  check_2F1_converges("grad_2F1", a1, a2, b1, z);
+              double precision = 1e-14, int max_steps = 1e6,
+              bool skip_convergence_check = false) {
+  if (!skip_convergence_check) {
+    check_2F1_converges("grad_2F1", a1, a2, b1, z);
+  }
 
   using stan::math::value_of_rec;
   using std::max;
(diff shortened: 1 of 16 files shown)


SteveBronder (changes requested):
Couple small comments (and one optional big one)

stan/math/prim/fun/hypergeometric_2F1.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+  check_not_nan("hypergeometric_2F1", "a1", a1);
+  check_not_nan("hypergeometric_2F1", "a2", a2);
+  check_not_nan("hypergeometric_2F1", "b", b);
+  check_not_nan("hypergeometric_2F1", "z", z);
+
+  check_2F1_converges("hypergeometric_2F1", a1, a2, b, z);
+
+  return boost::math::hypergeometric_pFq({a1, a2}, {b}, z);
+}
```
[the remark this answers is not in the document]
andrjohns: I actually have a rough plan for an arbitrary scalar vectorisation framework, with signature something like:
```
apply_scalar(f, args...);
```

Once the ternary vectorisation goes in, then I'll be working that up

andrjohns:
@SteveBronder just flagging that I'm still adding some special-case handling, so don't worry about re-reviewing just yet. Thanks!

andrjohns:
@spinkey I've added a bunch of closed-form specialisations/reductions of the `hypergeometric_2F1` from the [WolframAlpha reference docs](https://functions.wolfram.com/HypergeometricFunctions/Hypergeometric2F1/03/), and also added the [Euler Transformation](https://mathworld.wolfram.com/EulerTransform.html) if the arguments don't meet convergence criteria:

$$
_2F_1(a,b;c;z) = \frac{_2F_1(c-a,b;c;\frac{z}{z-1})}{(1-z)^b}
$$

Which gives a bit of a wider domain of inputs. Do you know if there are any other transformations/reductions/etc that would be good to have in?

stan/math/prim/fun/grad_inc_beta.hpp
```diff
@@ -31,8 +31,9 @@ inline void grad_inc_beta(double& g1, double& g2, double a, double b,
   double dF1 = 0;
   double dF2 = 0;
   double dF3 = 0;
+  double dFz = 0;
   if (C) {
-    grad_2F1(dF1, dF2, dF3, a + b, 1.0, a + 1, z);
+    grad_2F1(dF1, dF2, dF3, dFz, a + b, 1.0, a + 1, z);
```
SteveBronder, line 36: [optional] instead of in/out parameters you could do something like 

```suggestion
    std::forward_as_tuple(dF1, dF2, dF3, dFz) = grad_2F1(a + b, 1.0, a + 1, z);
```

https://godbolt.org/z/4GPqjeYYf

stan/math/prim/fun/hypergeometric_2F1.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+  // Check whether any special case rules apply with 'a' arguments reversed
+  // as 2F1(a1, a2, b, z) = 2F1(a2, a1, b, z)
+  OptT special_case_a2a1 = internal::hyper_2F1_special_cases(a2, a1, b, z);
+  if (special_case_a2a1.is_initialized()) {
+    return special_case_a2a1.get();
+  }
+
+  Eigen::VectorXd a_args(2);
+  Eigen::VectorXd b_args(1);
```
SteveBronder, line 182: Sized at compile time

stan/math/prim/fun/hypergeometric_2F1.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+  try {
+    check_2F1_converges("hypergeometric_2F1", a1, a2, b, z);
+
+    a_args << a1, a2;
+    b_args << b;
+    return hypergeometric_pFq(a_args, b_args, z);
+  } catch (...) {
```
SteveBronder, line 190: idt we normally use ..., are there several types of errors that can be thrown?  I think our usual is to catch `const std::exception&` and if we don't want to do anything with it we leave it unnamed

SteveBronder (approved):
Looks good! I have a few comments for things you can optionally resolve but overall I think its clean

SteveBronder (changes requested):
Couple of Qs around `grad_2F1_impl`. I think we should either return a tuple or do in/out parameters, but I think doing both just makes things look a bit hard to read. Also I'm a little confused on why you need the user defined template parameter bool to specify whether the return should based on the `partial_type_t` or the `return_type_t`

andrjohns:
@SteveBronder This is ready for another look, I've tidied up the tuple/multiple-assignment handling and split out the Euler transform flow, so hopefully the codes a little more readable now

test/unit/math/prim/prob/wishart_cholesky_test.cpp
```diff
@@ -81,7 +81,7 @@ TEST(ProbDistributionsWishartCholesky, dof_0) {
   MatrixXd L_Y = Y.llt().matrixL();
   MatrixXd L_S = Sigma.llt().matrixL();
 
-  unsigned int dof = std::numeric_limits<double>::quiet_NaN();
```
andrjohns, line 84: This change was needed to fix the `wishart_cholesky` prim failure from [this earlier Jenkins run](https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2792/21/pipeline/198)

This was relying on undefined behavior of a `NaN` overflowing to 0 when assigned to an int, which looks like it didn't happen in that Jenkins run. Let me know if you'd prefer I open separate PR and issue for this

SteveBronder:
Thanks!! That's fine and I'll look at this on Monday

stan/math/prim/fun/grad_2F1.hpp
```diff
@@ -9,144 +9,327 @@
...
+ * @param[in] a1 see generalized hypergeometric function definition
+ * @param[in] a2 see generalized hypergeometric function definition
+ * @param[in] b1 see generalized hypergeometric function definition
+ * @param[in] z see generalized hypergeometric function definition
+ * @param[in] precision magnitude of the increment of the infinite sum
+ *   to truncate the sum
+ * @param[in] max_steps number of steps to take
+ */
+template <bool ReturnSameT, typename T1, typename T2, typename T3, typename T_z,
+          require_not_t<std::integral_constant<bool, ReturnSameT>>* = nullptr>
```
SteveBronder, line 271: [optional] 

```suggestion
          require_not_t<bool_constant<ReturnSameT>>* = nullptr>
```

stan/math/prim/fun/grad_2F1.hpp
```diff
@@ -9,144 +9,327 @@
...
+ *   to truncate the sum
+ * @param[in] max_steps number of steps to take
+ */
+template <bool ReturnSameT, typename T1, typename T2, typename T3, typename T_z,
+          require_t<std::integral_constant<bool, ReturnSameT>>* = nullptr>
+auto grad_2F1(const T1& a1, const T2& a2, const T3& b1, const T_z& z,
+              double precision = 1e-14, int max_steps = 1e6) {
+  return internal::grad_2F1_impl<true, true, true, true>(a1, a2, b1, z,
+                                                         precision, max_steps);
+}
```
SteveBronder, line 309: If you set a default value for `ReturnSameT` like `bool ReturnSameT = false` do you need the other specialization below this one?

SteveBronder (approved):
Two small optionals but looks good!

stan/math/prim/fun/grad_2F1.hpp
andrjohns: Possibly, but since the default values have to come last in the template list I'd have to make some changes to how function would get called. I might leave that as a future TODO

PR #2797, opened 2022-07-21
Expose `hypergeometric_3F2` function

andrjohns:
## Summary

This PR renames and exposes the internal `F32` function as `hypergeometric_3F2`, and updates the calculations to be on the log-scale for additional stability.

The function currently uses autodiff for the gradients, even though there is also an internal `grad_F32` function, but I'll be updating that in a separate PR.

## Tests

No additional tests, as all existing tests should still pass

## Side Effects

N/A

## Release notes

Exposed the `hypergeometric_3F2` function and improved its numerical stability

## Checklist

- [x] Math issue #2664


bgoodri:
I think all of these hypergeometric functions should (eventually) call the ones in Boost, which have been included in the last several versions.

https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/hypergeometric.html

It is hard to do well with double precision in general, but I think they have put more effort into the numerics that we have or would want to do ourselves.

andrjohns:
> I think all of these hypergeometric functions should (eventually) call the ones in Boost, which have been included in the last several versions.
> 
> https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/hypergeometric.html
> 
> It is hard to do well with double precision in general, but I think they have put more effort into the numerics that we have or would want to do ourselves.

This is the only one that doesn't call the Boost implementation, since the `F32` header was already here (I'm guessing the Boost hypergeometrics didn't exist when that was written though). I'll update the PR to call the boost implementation instead

andrjohns:
This is now ready for review. I've updated `hypergeometric_3F2` to delegate to the `hypergeometric_pFq` (and boost's implementation). However there are some combinations of arguments which Boost says won't converge, even though they do (as part of the current tests), so I've left the naive infinite-sum approach implemented for that. 

The gradients are currently being calculated by the `grad_pFq` function, but I'll be refactoring and using the existing `grad_F32` function in a future PR

stan/math/prim/fun/hypergeometric_3F2.hpp
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @tparam Ta type of Eigen/Std vector 'a' arguments
+ * @tparam Tb type of Eigen/Std vector 'b' arguments
+ * @tparam Tz type of z argument
+ * @param[in] a Always called with a[1] > 1, a[2] <= 0
+ * @param[in] b Always called with int b[0] < |a[2]|,  <= 1)
+ * @param[in] z z (is always called with 1 from beta binomial cdfs)
+ * @param[in] precision precision of the infinite sum. defaults to 1e-6
+ * @param[in] max_steps number of steps to take. defaults to 1e5
+ * @return Generalized hypergeometric function applied to the inputs
```
spinkney, line 112: For the @return can you write out the inputs that the function is applied to? I'm not sure the best way to say this but something like `The 3F2 generalized hypergeometric function applied to {a1, a2, a3}, {b1, b2}`.

spinkney (approved):

stan/math/prim/fun/hypergeometric_3F2.hpp
andrjohns: Ah good call, will do!

stan/math/prim/fun/hypergeometric_3F2.hpp
```diff
@@ -109,7 +111,8 @@ T_return hypergeometric_3F2_infsum(const Ta& a, const Tb& b, const Tz& z,
  * @param[in] z z (is always called with 1 from beta binomial cdfs)
  * @param[in] precision precision of the infinite sum. defaults to 1e-6
  * @param[in] max_steps number of steps to take. defaults to 1e5
- * @return Generalized hypergeometric function applied to the inputs
+ * The 3F2 generalized hypergeometric function applied to the
```
spinkney, line 112: just add back `@return`

stan/math/prim/fun/hypergeometric_3F2.hpp
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @tparam Ta type of Eigen/Std vector 'a' arguments
+ * @tparam Tb type of Eigen/Std vector 'b' arguments
+ * @tparam Tz type of z argument
+ * @param[in] a Always called with a[1] > 1, a[2] <= 0
+ * @param[in] b Always called with int b[0] < |a[2]|,  <= 1)
+ * @param[in] z z (is always called with 1 from beta binomial cdfs)
+ * @param[in] precision precision of the infinite sum. defaults to 1e-6
+ * @param[in] max_steps number of steps to take. defaults to 1e5
+ * The 3F2 generalized hypergeometric function applied to the
```
spinkney, line 112: add `@return`

stan/math/prim/fun/hypergeometric_3F2.hpp
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @tparam Ta type of scalar 'a' arguments
+ * @tparam Tb type of scalar 'b' arguments
+ * @tparam Tz type of z argument
+ * @param[in] a Always called with a[1] > 1, a[2] <= 0
+ * @param[in] b Always called with int b[0] < |a[2]|,  <= 1)
+ * @param[in] z z (is always called with 1 from beta binomial cdfs)
+ * @param[in] precision precision of the infinite sum. defaults to 1e-6
+ * @param[in] max_steps number of steps to take. defaults to 1e5
+ * @return Generalized hypergeometric function applied to the inputs
```
spinkney, line 141: update doc here

stan/math/prim/fun/hypergeometric_3F2.hpp
andrjohns: Ahh the dangers of blindly copy-pasting! Thanks will fix those now

spinkney (approved):

PR #2814, opened 2022-09-13
Feature/2704 Add complex Schur decomposition

### LINKED ISSUE #2704: Add Schur decomposition for complex matricies
After FFT (#20) another complex-valued special function which would be good to have is Schur decomposition:
https://en.wikipedia.org/wiki/Schur_decomposition

This is probably a natural candidate to be the first function in the math library which returns a tuple as a result, if we wait until tuples are implemented. Otherwise we will need to do what we have for `svd` and have two functions which each return half of the answer. 

This is supported in Eigen: 
https://eigen.tuxfamily.org/dox/classEigen_1_1ComplexSchur.html

WardBrian:
## Summary

Adds two new functions, `complex_schur_decompose_t` and `complex_schur_decompose_u`. These wrap Eigen and implement the [Schur Decomposition](https://en.wikipedia.org/wiki/Schur_decomposition) of a matrix. 

This is mostly an update of code @bob-carpenter wrote during the initial effort to add complex matrix types. 

## Tests

Tests added to /mix and /prim. The prim tests verify that `A = U T U*` where `*` indicates conjugate transpose. 

## Side Effects

None

## Release notes

Added two new functions, `complex_schur_decompose_t` and `complex_schur_decompose_u`.

## Checklist

- [x] Math issue: Closes #2704

### DIFF
--- a/stan/math/prim/fun/complex_schur_decompose.hpp
+++ b/stan/math/prim/fun/complex_schur_decompose.hpp
@@ -27,7 +27,8 @@ namespace math {
 template <typename M, require_eigen_dense_dynamic_t<M>* = nullptr>
 inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
 complex_schur_decompose_u(const M& m) {
-  check_nonzero_size("complex_schur_decompose_u", "m", m);
+  if (m.size() == 0)
+    return m;
   check_square("complex_schur_decompose_u", "m", m);
   using MatType = Eigen::Matrix<scalar_type_t<M>, -1, -1>;
   // copy because ComplexSchur requires Eigen::Matrix type
@@ -50,7 +51,8 @@ complex_schur_decompose_u(const M& m) {
 template <typename M, require_eigen_dense_dynamic_t<M>* = nullptr>
 inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
 complex_schur_decompose_t(const M& m) {
-  check_nonzero_size("complex_schur_decompose_t", "m", m);
+  if (m.size() == 0)
+    return m;
   check_square("complex_schur_decompose_t", "m", m);
   using MatType = Eigen::Matrix<scalar_type_t<M>, -1, -1>;
   // copy because ComplexSchur requires Eigen::Matrix type
--- a/stan/math/prim/fun/complex_schur_decompose.hpp
+++ b/stan/math/prim/fun/complex_schur_decompose.hpp
@@ -4,8 +4,6 @@
 #include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/err.hpp>
-#include <Eigen/Dense>
-#include <complex>
 
 namespace stan {
 namespace math {
--- a/stan/math/prim/fun/complex_schur_decompose.hpp
+++ b/stan/math/prim/fun/complex_schur_decompose.hpp
@@ -1,0 +1,66 @@
+#ifndef STAN_MATH_PRIM_FUN_COMPLEX_SCHUR_DECOMPOSE_HPP
+#define STAN_MATH_PRIM_FUN_COMPLEX_SCHUR_DECOMPOSE_HPP
+
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/err.hpp>
+#include <Eigen/Dense>
+#include <complex>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return the unitary matrix of the complex Schur decomposition of the
+ * specified square matrix.
+ *
+ * The complex Schur decomposition of a square matrix `A` produces a
+ * complex unitary matrix `U` and a complex upper-triangular Schur
+ * form matrix `T` such that `A = U * T * inv(U)`.  Further, the
+ * unitary matrix's inverse is equal to its conjugate transpose,
+ * `inv(U) = U*`, where `U*(i, j) = conj(U(j, i))`
+ *
+ * @tparam value type of matrix
+ * @param m real matrix to decompose
+ * @return complex unitary matrix of the complex Schur decomposition of the
+ * specified matrix
+ * @see complex_schur_decompose_t
+ */
+template <typename M, require_eigen_dense_dynamic_t<M>* = nullptr>
+inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
+complex_schur_decompose_u(const M& m) {
+  check_nonzero_size("complex_schur_decompose_u", "m", m);
+  check_square("complex_schur_decompose_u", "m", m);
+  using MatType = Eigen::Matrix<scalar_type_t<M>, -1, -1>;
+  // copy because ComplexSchur requires Eigen::Matrix type
+  MatType mv = m;
+  Eigen::ComplexSchur<MatType> cs(mv);
+  return cs.matrixU();
+}
+
+/**
+ * Return the Schur form matrix of the complex Schur decomposition of the
+ * specified square matrix.
+ *
+ * @tparam value type of matrix
+ * @param m real matrix to decompose
+ * @return Schur form matrix of the complex Schur decomposition of the
+ * specified matrix
+ * @see complex_schur_decompose_u for a definition of the complex
+ * Schur decomposition
+ */
+template <typename M, require_eigen_dense_dynamic_t<M>* = nullptr>
+inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
+complex_schur_decompose_t(const M& m) {
+  check_nonzero_size("complex_schur_decompose_t", "m", m);
+  check_square("complex_schur_decompose_t", "m", m);
+  using MatType = Eigen::Matrix<scalar_type_t<M>, -1, -1>;
+  // copy because ComplexSchur requires Eigen::Matrix type
+  MatType mv = m;
+  Eigen::ComplexSchur<MatType> cs(mv);
+  return cs.matrixT();
+}
+
+}  // namespace math
+}  // namespace stan
+#endif
--- a/stan/math/prim/fun/complex_schur_decompose.hpp
+++ b/stan/math/prim/fun/complex_schur_decompose.hpp
@@ -18,7 +18,7 @@ namespace math {
  * unitary matrix's inverse is equal to its conjugate transpose,
  * `inv(U) = U*`, where `U*(i, j) = conj(U(j, i))`
  *
- * @tparam value type of matrix
+ * @tparam M type of matrix
  * @param m real matrix to decompose
  * @return complex unitary matrix of the complex Schur decomposition of the
  * specified matrix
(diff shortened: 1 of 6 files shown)


bob-carpenter:
I removed myself as a suggested reviewer.  I fear @rok-cesnovar is the last dev who actually understands how our template metaprogramming and new memory management functions are supposed to work.  So while I could review whether this works or not, I can't review whether it's written the way it's supposed to be.

rok-cesnovar:
I will review this tomorrow. 

The question "whether it's written the way it's supposed to be" is covered by our automatic testing in two aspects:
- performance: are Eigen expressions handled properly? Most importantly are there any duplicated evaluations?
- templates: from the point of view of Stan, we test that the function compiles in all scenarios in which it needs to.

bob-carpenter:
This is getting a bit off topic, but last time I tried to write code, I saw that half a dozen different functions used half a dozen different patterns.  I had no idea how to choose which one to use.  For instance, `softmax` used one and `log_softmax` another, despite the fact that the signatures should be identical.

I don't know how to test the things you mentioned.  I'll do as much of the review as I can and will indicate all of my residual uncertainty.

stan/math/prim/fun/complex_schur_decompose.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * @return complex unitary matrix of the complex Schur decomposition of the
+ * specified matrix
+ * @see complex_schur_decompose_t
+ */
+template <typename M, require_eigen_dense_dynamic_t<M>* = nullptr>
+inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
+complex_schur_decompose_u(const M& m) {
+  check_nonzero_size("complex_schur_decompose_u", "m", m);
+  check_square("complex_schur_decompose_u", "m", m);
+  using MatType = Eigen::Matrix<scalar_type_t<M>, -1, -1>;
```
bob-carpenter, line 32: How is line 32 plus the indirect `require_eigen_dense` thing different than just changing the signature to

```cpp
inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
complex_schur_decompose_u(const Eigen::Matrix<scalar_type_t<M>, -1, -1>& mv);
```

stan/math/prim/fun/complex_schur_decompose.hpp
```diff
@@ -0,0 +1,64 @@
...
+  // copy because ComplexSchur requires Eigen::Matrix type
+  MatType mv = m;
+  Eigen::ComplexSchur<MatType> cs(mv);
+  return cs.matrixU();
```
bob-carpenter, line 36: It this prematurely collapsing what would otherwise be an expression template?  I think Eigen's actually returning a concrete matrix here, so hopefully no copy (they make it really hard to cut from Eigen doc):.

<tt>[Matrix](https://eigen.tuxfamily.org/dox/classEigen_1_1Matrix.html)<[ComplexScalar](https://eigen.tuxfamily.org/dox/classEigen_1_1ComplexSchur.html#a9cfc7d20234dd51933b0cb6ef427c8b4), RowsAtCompileTime, ColsAtCompileTime, Options, MaxRowsAtCompileTime, MaxColsAtCompileTime></tt>

stan/math/prim/fun/complex_schur_decompose.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * Return the unitary matrix of the complex Schur decomposition of the
+ * specified square matrix.
+ *
+ * The complex Schur decomposition of a square matrix `A` produces a
+ * complex unitary matrix `U` and a complex upper-triangular Schur
+ * form matrix `T` such that `A = U * T * inv(U)`.  Further, the
+ * unitary matrix's inverse is equal to its conjugate transpose,
+ * `inv(U) = U*`, where `U*(i, j) = conj(U(j, i))`
+ *
+ * @tparam value type of matrix
```
bob-carpenter, line 21: There is no template parameter named `value`.  Was this supposed to be `M`, which is the whole matrix, not its value type?

stan/math/prim/fun/complex_schur_decompose.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * @param m real matrix to decompose
+ * @return complex unitary matrix of the complex Schur decomposition of the
+ * specified matrix
+ * @see complex_schur_decompose_t
+ */
+template <typename M, require_eigen_dense_dynamic_t<M>* = nullptr>
+inline Eigen::Matrix<complex_return_t<scalar_type_t<M>>, -1, -1>
+complex_schur_decompose_u(const M& m) {
+  check_nonzero_size("complex_schur_decompose_u", "m", m);
```
bob-carpenter, line 30: Shouldn't size 0 just return a 0 x 0 matrix for both components?  If `A`, `U` and `T` are size 0 x 0, then `A = U * T * inv(U)` is also 0 x 0.  The inverse of a 0 x 0 matrix is just the unique 0 x 0 matrix.

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
```diff
@@ -0,0 +1,94 @@
+#include <test/unit/math/test_ad.hpp>
+#include <stdexcept>
+
```
bob-carpenter, line 3: I don't see any tests in here that test whether things are getting doubly allocated or recomputed.  Are we just supposed to do that by inspection?

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+  EXPECT_THROW(f(a32), std::invalid_argument);
+}
+
+template <typename V>
+void test_complex_schur_decompose(const Eigen::MatrixXd& x) {
+  using stan::math::complex_schur_decompose_t;
+  using stan::math::complex_schur_decompose_u;
+  using stan::math::value_of_rec;
+  Eigen::Matrix<V, -1, -1> X(x.rows(), x.cols());
+  for (int i = 0; i < x.size(); ++i)
```
bob-carpenter, line 38: Why isn't this just `X = x`?

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+    for (int i = 0; i < x.rows(); ++i) {
+      EXPECT_NEAR(value_of_rec(X(i, j).real()), value_of_rec(X2(i, j).real()),
+                  1e-8);
+      EXPECT_NEAR(value_of_rec(X(i, j)).imag(), value_of_rec(X2(i, j).imag()),
+                  1e-8);
+    }
+  }
+}
+
+TEST(mathMixFun, complexSchurDecompose) {
```
bob-carpenter, line 71: What is this adding to the `test_ad` function?  That already tests values, gradients, and exception behavior for all the arguments.

test/unit/math/prim/fun/complex_schur_decompose_test.cpp
```diff
@@ -0,0 +1,40 @@
+#include <stan/math/prim.hpp>
+#include <test/unit/util.hpp>
+#include <gtest/gtest.h>
+
+template <typename T, typename U>
+void expect_complex_mat_eq(const T& x, const U& y, double tol = 1e-8) {
```
bob-carpenter, line 6: This should get pushed up into our AD testing tools and not live in the complex_shur test.   We also have functions to extract real and complex components, and these can just call matrix equality.

stan/math/prim/fun/complex_schur_decompose.hpp
WardBrian: See above, but yes, this is forcing an expression evaluation.

stan/math/prim/fun/complex_schur_decompose.hpp
WardBrian: Good catch, copy/paste error from prior code

stan/math/prim/fun/complex_schur_decompose.hpp
WardBrian: This is an implementation choice I don't have a strong opinion on, but I should note that `scipy` throws an error on size-zero inputs.

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
WardBrian: That is done through the expression testing framework. The code to do so is in test/expressions

The basic outline is code is generated for each signature listed by `stanc --dump-stan-math-signatures`. This obviously requires the signature to be exposed before being tested, but you can also run it manually; I have

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
WardBrian: My guess was that because `x` is a Matrix of double and X could be a matrix of autodiff variables that we'd need elementwise assignment to handle that promotion, but it seems that isn't true (it compiles if I change it, anyway)

test/unit/math/mix/fun/complex_schur_decompose_test.cpp
WardBrian: Copied from old tests which didn't have the `prim` tests at all, so I believe these were the only tests that the function actually satisfied the required properties. Do you want me to remove them?

WardBrian: My first thought was to move these into `prim` without the extra types (so just using square_test_matrices to test double matrices of different sizes) but it turns out that this function is in the `test_ad` header rather than the normal `utils` header

stan/math/prim/fun/complex_schur_decompose.hpp
bob-carpenter: I think the return is already a concrete matrix if I'm reading the type right, so I don't think there's a better way to return this.

stan/math/prim/fun/complex_schur_decompose.hpp
bob-carpenter: Hmm.  I thought this would do the eval because it's a constant reference, but I've been surprised by the subtleties of this before.

WardBrian:
@rok-cesnovar do you mind doing a quick review of this? Mainly just to make sure there aren't any gotchas with templating I missed. 

I'm going to open a PR to expose this in the language soon, which should catch anything else, but I did run the expression tests locally already

test/unit/util.hpp
```diff
@@ -120,6 +120,26 @@
...
+  {                                                           \
+    const Eigen::MatrixXcd& A_eval = A;                       \
+    const Eigen::MatrixXcd& B_eval = B;                       \
+    EXPECT_EQ(A_eval.rows(), B_eval.rows());                  \
+    EXPECT_EQ(A_eval.cols(), B_eval.cols());                  \
+    for (int i = 0; i < A_eval.size(); i++) {                 \
+      EXPECT_NEAR(A_eval(i).real(), B_eval(i).real(), DELTA); \
+      EXPECT_NEAR(A_eval(i).imag(), B_eval(i).imag(), DELTA); \
+    }                                                         \
+  }
```
rok-cesnovar, line 141: Thanks for this!

rok-cesnovar (approved):
This looks great to me!

Issue #2818, opened 2022-10-01
Expose variadic arguments for algebra_solver


charlesm93:
## Description
The algebraic solvers, `algebra_solver_newton` and `algebra_solver_powell` aren't using variadic arguments. Most of the code seems to be already there but it is all wrapped in a function that still uses the strict function signature. We need to create the following updated signatures:

- `algebra_solver_newton(f, x0,  ...)`
- `algebra_solver_tol_newton(f, x0, scaling_step_size, function_tolerance, max_num_steps, ...)`
- `algebra_solver_powell(f, x0, ...)`
- `algebra_solver_tol_powell(f, x0, relative_tolerance, function_tolerance, max_num_steps, ...)` 

Open question: should we keep the `relative_tolerance` tuning parameter used for Powell's method? Right now, the two solvers have slightly different choices of tuning parameters, which has to do with the underlying algorithms they use. I think that's fine but it might be confusing to the user.

#### Current Version:
v4.4.0

WardBrian:
The variadic interface would completely subsume the original, correct? I only ask because we currently have a technical restriction in Stanc where a variadic function cannot have non-variadic overloads

charlesm93:
My plan is to only variadic arguments for `algebra_solver_powell` and `algebra_solver_newton` but keep the non-variadic `algebra_solver` for backward compatibility. The latter is a function we intend to deprecate, so I guess you'd need a warning message to that effect when compiling the Stan program.

But no, no overloading for variadic functions.

WardBrian:
I don’t think my question was clear, we currently already have an exposed function called `algebra_solver _newton` with arguments`function algebra_system, vector y_guess, vector theta, data array[] real x_r, array[] int x_i ` - this is strictly a subset of the signatures a variadic version of this would allow, right? So any pre-existing usage of this function should be able to become the variadic version without any changes to the model

charlesm93:
Thanks for clarifying. Yes, you're right. We actually won't break backwards compatibility. I got a little bit mixed up: in stan-math we do break backwards compatibility because we have to reposition the `std::ostream* const msgs` argument. But I now realize this "backwards breaking" won't propagate to the Stan language.

charlesm93:
`stan-math` PR is in. Once this gets reviewed and merged, we can handle the Stan language. And *then* will properly clean up the doc.

PR #2820, opened 2022-10-03
Feature/issue 2828 algebra solver variadic

charlesm93:
## Summary

Expose signatures with variadic arguments for Stan's algebraic solvers. For each solver (Newton and Powell) introduce a version which admits tuning parameters and one which uses defaults. The new function names are: `algebra_solver_newton`, `algebra_solver_newton_tol`, `algebra_solver_powell`, and `algebra_solver_powell_tol`.

For backwards compatibility, the signature `algebra_solver` is still available. It uses the Powell method. Since this signature will be deprecated, I didn't update it to use variadic arguments.

## Tests

The following unit tests have been updated:

- `test/unit/math/rev/functor/algebra_solver_powell_test.cpp`
- `test/unit/math/rev/functor/algebra_solver_newton_test.cpp`
- `test/unit/math/rev/functor/algebra_solver_fp_test.cpp`

## Side Effects

This breaks backwards compatibility of the `stan-math` library. Previous C++ code using `algebra_solver_powell` and `algebra_solver_newton` will need to be updated, because the stream argument has to appear before any non-mandatory argument (including variadic arguments). Since Stan users do not explicitly pass the stream argument, this will however not break backwards compatibility of Stan itself.

## Release notes

Stan's algebraic solvers now support variadic arguments.

## Checklist

- [x] Math issue #2818 

- [x] Copyright holder: Charles Margossian and Flatiron Institute.

    - [x] unit tests pass (to run, use: `./runTests.py test/unit/math/rev/functor/algebra_solver_powell_test.cpp` and `test/unit/math/rev/functor/algebra_solver_newton_test.cpp`)
    - [ ] header checks pass, (`make test-headers`)
    - [ ] dependencies checks pass, (`make test-math-dependencies`)
    - [ ] docs build, (`make doxygen`)
    - [x] code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)


test/unit/math/rev/functor/util_algebra_solver.hpp
```diff
@@ -471,13 +470,35 @@ Eigen::Matrix<stan::math::var, Eigen::Dynamic, 1> variadic_eq_test(
...
+                    variadic_eq_functor()),
+                x, &std::cout, scaling_step_size, function_tolerance,
+                max_num_steps, A, y_1, y_2, y_3, i)
+                      : algebra_solver_powell_impl(
+                          algebra_solver_adapter<variadic_eq_functor>(
+                              variadic_eq_functor()),
+                          x, &std::cout, relative_tolerance, function_tolerance,
+                          max_num_steps, A, y_1, y_2, y_3, i);
+  } else {
+    // TO DO: add Newton solver case.
```
rok-cesnovar, line 485: Stray comment?

test/unit/math/rev/functor/algebra_solver_powell_test.cpp
```diff
@@ -205,7 +206,9 @@ TEST_F(degenerate_eq_test, powell_guess1) {
   for (int k = 0; k < n_x; k++) {
     Eigen::Matrix<var, Eigen::Dynamic, 1> y = y_dbl;
     Eigen::Matrix<var, Eigen::Dynamic, 1> theta = algebra_solver_powell(
-        degenerate_eq_functor(), x_guess_1, y, dat, dat_int);
+        degenerate_eq_functor(), x_guess_1, &std::cout, y, dat, dat_int);
+    // algebra_solver_powell(
+    //     degenerate_eq_functor(), x_guess_1, y, dat, dat_int);
```
rok-cesnovar, line 211: Cleanup comment

rok-cesnovar (approved):
This looks great to me. There are two comments that should be cleaned up in the test files, but other than that, this is good to merge.

rok-cesnovar (changes requested):
I just noticed that the tests are failing, so that obviously needs to be handled first.

rok-cesnovar:
> Do we run the expression tests on any of the other variadic functions?
Ah, missed this comment.  We do test those cases. See ode_* and reduce_sum stuff here: https://github.com/stan-dev/math/blob/develop/test/sig_utils.py#L231

charlesm93:
As @WardBrian pointed out to me in person, some of the late unit tests use the current compiler. However this compiler needs to be updated for the variadic arguments. I think the move is to write the corresponding PR for stanc3 and then try to merge the two simultaneously.

rok-cesnovar:
Sounds good to me!

charlesm93:
actually we have a bit of a design pickle we need to address to maintain backward compatibility. See: https://discourse.mc-stan.org/t/recent-history-of-algebraic-solver/28992/3

charlesm93:
`stan-math` is updated to support the deprecated signatures (`algebra_solver` and `algebra_solver_newton`) and the new signatures for variadic arguments (`solve_powell`, `solve_powell_tol`, `solve_newton`, `solve_newton_tol`).

The existing unit tests all target the new variadic argument signatures and I added two unit tests for each of the deprecated signatures, testing the function with and without the optional tuning arguments. (under the hood, all the signatures use the same functions)

I think this should pass all the integration tests, since we're preserving the old signatures.

charlesm93:
@WardBrian you said you might have an idea as to what is making the unit tests fail, could you clue me in on this? I couldn't make sense of the error message produced by Jenkins.

WardBrian:
```
test/unit/math/rev/functor/algebra_solver_fp_test.cpp:497:41: error: ‘algebra_solver_newton_tol’ was not declared in this scope; did you mean ‘algebra_solver_adapter’?
  497 |   Eigen::Matrix<var, -1, 1> xv_newton = algebra_solver_newton_tol(

```

Did you mean to update this test to the new signature? It looks like it ended up stuck in the WIP signatures

charlesm93:
thanks! Looks like i forgot to git push after my last commit. Hopefully the tests pass now.

charlesm93:
@rok-cesnovar do you mind going over the PR? All tests are passing now.

rok-cesnovar:
Will do it this evening. Thanks for the reminder, slipped my mind.

WardBrian:
@charlesm93 If you want I'm more than happy to write up the stanc PR to implement this. I think I could probably figure it out from the source here, but it would be very helpful if you could write out the signature spec (for typechecking, e.g. what requirements if any do we place on the variadic args) and the C++ calling order (for code-generation, e.g. where does `pstream` go, etc)

rok-cesnovar (dismissed):

stan/math/rev/functor/solve_newton.hpp
```diff
@@ -252,9 +348,9 @@ Eigen::Matrix<scalar_type_t<T2>, Eigen::Dynamic, 1> algebra_solver_newton(
     const double scaling_step_size = 1e-3,
     const double function_tolerance = 1e-6,
     const long int max_num_steps = 200) {  // NOLINT(runtime/int)
-  return algebra_solver_newton_impl(algebra_solver_adapter<F>(f), x, msgs,
-                                    scaling_step_size, function_tolerance,
-                                    max_num_steps, y, dat, dat_int);
+  return solve_newton_impl(algebra_solver_adapter<F>(f), x, msgs,
```
WardBrian, line 351: The way we generic variadics in the compiler (e.g. for reduce_sum, ODES, DAEs) we place the `pstream__` argument after the required args and before the variadics. This `algebra_solver_adapter` seems to re-order them so that this isn't necessary, but it would actually be _more_ work if this didn't follow that pattern.

I would like in the future to do an internal change so that we don't have any of this kind of reordering (and move it all into stan_math) so that the pstream argument is always last, but for now, would you mind changing this?

WardBrian: I should add that if you delete the wrapping in `algebra_solver_adapter`, adding these to the compiler looks like this:
```ocaml
  add_variadic_fn "solve_newton" ~return_type:UnsizedType.UVector
    ~control_args:[UnsizedType.(AutoDiffable, UVector)]
    ~required_fn_rt:UnsizedType.UVector
    ~required_fn_args:[UnsizedType.(AutoDiffable, UVector)]
    ()
```

That's all that would be required for each signature. Otherwise we will need to special-case the code generation

charlesm93: My understanding is that a function defined in Stan is translated into a functor with an operator which takes `msgs` as its last argument. In `stan-math` (i.e. at the C++ level), the variadic arguments always need to be passed last. So some reordering needs to happen. Or can I write a statement like the following:
```
return f(x, args..., msgs)
```
If the above is valid, we can remove `algebra_solver_adapter` all together.

WardBrian: `msg` is the final argument for non-variadic versions. We have special code gen for all other variadic functions, so for now I think we should for these. 

If we could have stan_math always pass it at the end that would be nice, but that would be a separate PR imo

WardBrian: Compare to `dae`, which has the `msg` in the middle before `args...`: 

https://github.com/stan-dev/math/blob/224e5812c0c4885f5420caa72275f64f0b2568ed/stan/math/rev/functor/dae_system.hpp#L149

To be perfectly clear, I _wish_ it was always the case that `msgs` was the last argument, but at the moment it isn't for any of the functions accepted by variadic functions, so we should be consistent.

charlesm93: > msg is the final argument for non-variadic versions. 

The function, `f`, which specifies the algebraic system and is defined by the user in the `functions` block is a non-variadic function. Once translated into C++, it might look as follows:
```
struct simple_eq_functor {
  template <typename T1, typename T2, typename T3, typename T4>
  inline Eigen::Matrix<stan::return_type_t<T1, T2>, Eigen::Dynamic, 1>
  operator()(const T1& x, const T2& y, const T3& dat, const T4& dat_int,
             std::ostream* pstream__) const {
    Eigen::Matrix<stan::return_type_t<T1, T2>, Eigen::Dynamic, 1> z(2);
    z(0) = x(0) - y(0) * y(1);
    z(1) = x(1) - y(2);
    return z;
  }
};
```
The user doesn't specify the `msgs` argument, this simply gets tacked at the end during translation. So we need to handle a function which takes msgs as its last argument. (unless something has changed regarding how user-defined functions are translated)

This is also how the ode solvers are tested. If I look at the functor used for the unit tests, e.g. for the [harmonic oscillator](https://github.com/stan-dev/math/blob/develop/test/unit/math/prim/functor/harmonic_oscillator.hpp#:~:text=%7D%3B-,struct%20harm_osc_ode_data_fun%20%7B,%7D,-%3B):
```
struct harm_osc_ode_fun {
  template <typename T0, typename T1, typename T2>
  inline std::vector<stan::return_type_t<T1, T2>>
  // initial time
  // initial positions
  // parameters
  // double data
  // integer data
  operator()(const T0& t_in, const std::vector<T1>& y_in,
             const std::vector<T2>& theta, const std::vector<double>& x,
             const std::vector<int>& x_int, std::ostream* msgs) const {
    if (y_in.size() != 2)
      throw std::domain_error(
          "this function was called with inconsistent state");

    std::vector<stan::return_type_t<T1, T2>> res;
    res.push_back(y_in.at(1));
    res.push_back(-y_in.at(0) - theta.at(0) * y_in.at(1));

    return res;
  }
};
```

I'm not sure what's going on with dae. The [functor used in the unit test](https://github.com/stan-dev/math/blob/0a53c7d78cf01128dca6c6c46f137cd84aebb901/test/unit/math/rev/functor/idas_system_test.cpp) indeed has `msgs` in the middle:
```
struct chemical_kinetics {
  template <typename T0, typename Tyy, typename Typ, typename Tpar>
  inline Eigen::Matrix<stan::return_type_t<Tyy, Typ, Tpar>, -1, 1> operator()(
      const T0& t_in, const Eigen::Matrix<Tyy, -1, 1>& yy,
      const Eigen::Matrix<Typ, -1, 1>& yp, std::ostream* msgs,
      const std::vector<Tpar>& theta, const std::vector<double>& x_r,
      const std::vector<int>& x_i) const {
    if (yy.size() != 3 || yp.size() != 3)
      throw std::domain_error(
          "this function was called with inconsistent state");

    Eigen::Matrix<stan::return_type_t<Tyy, Typ, Tpar>, -1, 1> res(3);

    auto yy1 = yy(0);
    auto yy2 = yy(1);
    auto yy3 = yy(2);

    auto yp1 = yp(0);
    auto yp2 = yp(1);
    // auto yp3 = yp.at(2);

    auto p1 = theta.at(0);
    auto p2 = theta.at(1);
    auto p3 = theta.at(2);

    res[0] = yp1 + p1 * yy1 - p2 * yy2 * yy3;
    res[1] = yp2 - p1 * yy1 + p2 * yy2 * yy3 + p3 * yy2 * yy2;
    res[2] = yy1 + yy2 + yy3 - 1.0;

    // jacobian respect to yy

    return res;
  }
};
```
But I don't think that's the type of structure you get when Stan translates a user-defined function. Has the DAE been exposed to the Stan language and tested?

WardBrian: Yes, and we generate unique functors for the variadic functions each time. So a ODE will have foo_functor_ and (currently called) foo_odefunctor_ if you inspect a generated hpp file. DAES have a foo_daefunctor_. All these do is call the original with the arguments re-ordered

I was previously told this was a necessary technical requirement, but obviously there is something we can (and in the future should) be doing in the math library to allow us to just use the _functor_ version everywhere and get rid of these extra bespoke functor declarations

charlesm93: Ok, great. This now makes sense to me. So the higher-order functions with variadic arguments should target functions with `msgs` appearing in the middle (e.g foo_odefunctor, foor_daefunctor).

I now see your point and I'll make the requested changes.

WardBrian: Thanks! Sorry for not being clear originally

WardBrian:
@charlesm93 I have a compiler branch which has the exposed functions from this PR here:
https://github.com/stan-dev/stanc3/tree/feature/variadic-algebra-solvers/

It needs a few test models for them added to `test/integration/good`. See the current test: https://github.com/stan-dev/stanc3/blob/master/test/integration/good/algebra_solver_good.stan

Would you mind writing a few? They don't need to be sensible to run (i.e. there's no need to define _solvable_ systems for it), just to stress-test the compilation.

If we want to deprecate the older signatures and have the cannonicalizer re-write them into the new signatures I can also add that,  but it will be a bit more involved. We can also do that in a separate PR

charlesm93:
Hi,
I can write additional unit tests under `test/integration/good`. For this current PR, would it be straightforward to emit a compiler message stating the old signature will soon be deprecated? We can then leave the re-writing to the next PR.

WardBrian:
I'd prefer the message and rewriting be in the same effort, since our standard thus far is the deprecation message includes a hint like `This can be updated automatically by stanc`.

I can probably get a second PR which does both in before 2.31 releases, but even if not I don't think it's a massive issue if the new signatures get added in 2.31 and the old ones are not formally deprecated until 2.32

charlesm93:
Ok. Let's do a second PR that takes care of both the message and the rewriting then. The sooner we get the language PR in, the sooner I can clean up the documentation (which what initially got me looking again at the algebraic solver).

PR #2821, opened 2022-10-07
Patch Eigen to resolve FFT forward derivative issue

WardBrian:
## Summary

The FFT functions introduced in #2686 disabled forward-mode autodiff due to it failing. 

I was finally able to track down the issue, which stems from this one line in Eigen's main FFT implementation:
https://github.com/stan-dev/math/blob/cd60e75a4e63fa5c321ed6effcf825adfbde92c0/lib/eigen_3.3.9/unsupported/Eigen/src/FFT/ei_kissfft_impl.h#L34

The code above uses `acos(-1)` as a shorthand for $\pi$. But, it also casts -1 to the input scalar type, in our case, `fvar<T>`. The derivative of acos is only defined for `-1 < x < 1`, so this creates an NaN which eventually propagates everywhere. But, this number is really just a constant, so we don't need to autodiff it.

There are two options I think for fixing this:

1. What is currently done here, which is just patch the FFT module to remove that scalar cast. 
   ```diff
   --      Scalar phinc =  (inverse?2:-2)* acos( (Scalar) -1)  / nfft;
   ++      // STAN CHANGE: The next line originally cast -1 to Scalar, which breaks derivatives
   ++      Scalar phinc =  (inverse?2:-2)* acos(-1.0)  / nfft;
   ```
2. Use Eigen 3.4's FFT module or update Eigen (#2583). As part of improving the code of the function in question, they removed the offending call to `acos`. @andrjohns @SteveBronder - how close do you think an update is?

If we go with option 1, we won't need to maintain this into the future since the next Eigen update will fix this anyway.

## Tests

This enables forward mode testing for the FFT functions. It also deletes the code which was used to disable them, since it was not used elsewhere (and likely should not be in the future).

## Side Effects

None

## Release notes

Fixed forward mode autodiff for FFT functions.

## Checklist

- [ ] Math issue 

### DIFF
--- a/lib/eigen_3.3.9/unsupported/Eigen/src/FFT/ei_kissfft_impl.h
+++ b/lib/eigen_3.3.9/unsupported/Eigen/src/FFT/ei_kissfft_impl.h
@@ -31,7 +31,8 @@ struct kiss_cpx_fft
       using std::acos;
       m_inverse = inverse;
       m_twiddles.resize(nfft);
-      Scalar phinc =  (inverse?2:-2)* acos( (Scalar) -1)  / nfft;
+      // STAN CHANGE: The next line originally cast -1 to Scalar, which breaks derivatives
+      Scalar phinc =  (inverse?2:-2)* acos(-1.0)  / nfft;
       for (int i=0;i<nfft;++i)
         m_twiddles[i] = exp( Complex(0,i*phinc) );
     }
--- a/test/unit/math/ad_tolerances.hpp
+++ b/test/unit/math/ad_tolerances.hpp
@@ -69,34 +69,6 @@ struct ad_tolerances {
         grad_hessian_grad_hessian_(1e-2) {}
 };
 
-/**
- * Return tolerances that are infinite other than for the value
- * tolerance and gradient tolerance for reverse mode.
- *
- * @param val_tol value relative tolerance (default `1e-8`)
- * @param grad_tol gradient relative tolerance (default `1e-4`)
- */
-ad_tolerances reverse_only_ad_tolerances(double val_tol = 1e-8,
-                                         double grad_tol = 1e-4) {
-  ad_tolerances tols;
-  tols.gradient_val_ = val_tol;
-  tols.gradient_grad_ = grad_tol;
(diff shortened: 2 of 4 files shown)


rok-cesnovar (approved):
Nice find!

bob-carpenter:
Our process means you can go ahead and merge with a single approval, so I'm not trying to block the merge here or over-review.  I do have a question, though:  Are you sure we won't update to a later version of Eigen 3.3.x before moving to Eigen 3.4?  If so, is there a unit test that will catch a regression to bad behavior if we forget to patch 3.4?

An alternative would be to pull the FFT code into Stan itself.  The whole thing is in one file of 400 or so lines.   

I was about to say we should refactor the following to `-2 + inverse * 4` because the following has a branch point:

```cpp
(inverse?2:-2)
```

But the optimizing C++ compilers beat me to it.  They generate this assembly in GCC 8 at `-O3`:

```
choose(bool):
  cmpb $1, %dil
  sbbl %eax, %eax
  andl $-4, %eax
  addl $2, %eax
  ret
```

and this in Clang 8 at `-O3`:

```
_Z6chooseb: # @_Z6chooseb
  leal -2(,%rdi,4), %eax
  retq
```

Cool!

WardBrian:
> Are you sure we won't update to a later version of Eigen 3.3.x 

We are using 3.3.9, which is still the most recent version in the 3.3.x line

> is there a unit test that will catch a regression to bad behavior if we forget to patch 3.4?

The FFT tests re-enabled in this PR will start failing if the bad behavior returns, but I've manually checked Eigen 3.4 and it does not require a patch, they fixed this on their own in that version

bob-carpenter:
After writing the above, I got this from @WardBrian via email:

> Eigen 3.4 did fix the issue indirectly by re-writing that entire function. The [release notes](https://eigen.tuxfamily.org/index.php?title=3.4#Improvements_to_FFT_module) mention that it was
re-written to be faster and more accurate, I think it fixing our code was just a side effect.

So Eigen beat me to that.  So let's go with @WardBrian's solution as soon as unit tests pass.

PR #2826, opened 2022-10-17
Improve CI times

WardBrian:
## Summary

Various things tried (running list):

1. Re-enable a version of Jumbo tests
    - Added a parameter to turn these off per-run for debugging
2. Try to parallelize distribution tests.
3. Tweak Jenkins settings to better match Flatiron machines (4 cores, ~16gb ram)
4. ~More consistently use `O=0`~

## Checklist


WardBrian:
Latest run is still going for final timings but I think with this and the Stan speedups that a full run of the Math CI should take ~5 hours. I think that is worthy of merging on its own, and I can do some more fine tuning later.

@rok-cesnovar and @serban-nicusor-toptal do you mind taking a look at the changes? Particularly the parallelization of the Distribution tests required me writing more Groovy than I ever have before in my life.

runTests.py
```diff
@@ -33,26 +33,26 @@
...
     "test/unit/math/fwd/meta",
     "test/unit/math/fwd/prob",
-    # "test/unit/math/mix/core",
+    "test/unit/math/mix/core",
     "test/unit/math/mix/fun",
-    # "test/unit/math/mix/functor",
+    "test/unit/math/mix/functor",
     "test/unit/math/mix/meta",
     "test/unit/math/mix/prob",
     "test/unit/math/opencl/device_functions",
```
rok-cesnovar, line 58: 🚀 great stuff that you got all of them working!

rok-cesnovar (approved):
No objections from me on the Groovy code :) 

Lets let Nic have a look and then get this merged.

runTests.py
WardBrian: It was mostly a problem of needing to namespace code. If anything crops up in the future (easy to test because we can disable Jumbos for just one run now) it will most likely be something related to that as well (aka an easy fix, if you know what to look for)

WardBrian:
Latest normal run took 5h47min. Quite happy with that, considering I had a run last week which took 17 hours. 
I think there is probably at least 1 hour and at most 2 hours we could still squeeze by just making things faster here or there, and then probably more by making the tests _smarter_ rather than just faster

Running a row vector test now to make sure that the stuff we run only on `develop` still works with my parallel setup.


Also, I think whenever we merge this we should do a Squash and Merge to prevent all 30 of my debug commits ending up in the tree

serban-nicusor-toptal (dismissed):
LGTM, everything looks fine and neat on the Jenkins side.
Many thanks for this PR, such a great improvement!

WardBrian:
Hm, seems setting O0 in the windows CI has caused the linker to fail. This is most likely a bug in the tests (I found a similar one in the Linux tests already), but for now the easiest thing will probably be just putting them back at O1 or something. I’ll look into it tomorrow

WardBrian:
Full `develop`/`master` run (e.g. with Row Vector distribution tests) took 12h15m, down from ~47h

rok-cesnovar (approved):

WardBrian:
@SteveBronder raised some concerns about not testing at `O=3` when that is the default for users, given how for numerical code that can be a problem.

I'm not sure if we want to enable those only for `develop` runs or not, I'm not sure how the RAM requirements for Jumbos work with O3, since that requires more RAM anyway.

I'm going to do a test run in a minute

rok-cesnovar:
I would say running it on "develop" test would be good enough. Unless we start seeing develop tests often failing after passes on PRs, then we would need to reconsider.

rok-cesnovar (dismissed):
Assuming tests pass, good to go!

WardBrian:
I want to do a test run at `O=3` before merging to make sure that doesn't break any jumbos

WardBrian:
https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2826/44/pipeline is running at O3 so if that passes I think we can merge

WardBrian:
The jumbos worked  at `O3` in https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2826/51/pipeline/169 before it ran into an unrelated issue, so assuming these tests pass I feel good to squash and merge

rok-cesnovar (approved):

rok-cesnovar:
There seems to be an issue on develop after merge: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/develop/101/pipeline

WardBrian:
@rok-cesnovar it seems like that jumbo test is picking up a file which was deleted in the merge right before it. 

I’m re-running now, but @serban-nicusor-toptal could this be another issue where a folder isn’t being properly deleted?

serban-nicusor-toptal:
As long as we properly cleanup which I think we do, we should not have problems with persistent data.

rok-cesnovar:
Ok, all good then. Sorry for the false positive :)

WardBrian:
That looks again like older versions of files sticking around or something

Issue #2592, opened 2021-10-01
Add a signed log_sum_exp


andrjohns:
## Description
When using `log_sum_exp`, it would be useful to be able specify a 'sign' for the variables being aggregated - such that some inputs are added and some subtracted.

I'm currently needing this for the gradients of `hyper_pFq` function, which are all computed on the log scale and the signs tracked through the infinite sum.

## Example

The function would be:
```
log_sum_exp_signed(input_container, signs_container)
```
Which would be equivalent to:
```
log(sum(exp(input_container) .* signs_container))
```

But with the `log_sum_exp` approach to avoiding overflow

#### Current Version:
v4.1.0

andrjohns:
I'm finding this will be better as an internal function for just `grad_pFq`, since it needs to track the sign of the returned sum as well. Will close for now

bob-carpenter:
Belay that last request---this answered all my questions!  Thanks for commenting the basis for the decision.

andrjohns:
Re-opening this since I'm finding it's needed for other functions which aggregate on the log-scale but still track the sign

PR #2829, opened 2022-10-20
Log sum exp signed

andrjohns:
## Summary

This PR introduces a function for performing `log_sum_exp` while respecting the signs of the input variables and tracking the sign of the returned sum. The function returns a two-element tuple, where the first element is the sum (on the log-scale) and the second is the sign. This isn't intended to be exposed to users, so I think the `tuple` return is fine.

We currently have to exponentiate and apply the signs to each argument (so that the values are correctly added or subtracted) and then extract the sign of the result and return the `log(fabs())` of the sum:
```cpp
double a = 2.1;
double b = 54.21;
int a_sign = 1;
int b_sign = -1;

double sum_exp_scale = a_sign * exp(a) + b_sign * exp(b);
double sum_log_scale = log(fabs(sum_exp_scale));
int sum_log_scale_sign = sign(sum_exp_scale);
```

However, this means that we lose the over-/under-flow robustness of the `log_sum_exp` and `log_diff_exp` functions. The `log_sum_exp_signed` function uses the provided signs to determine whether a `log_sum_exp` or `log_diff_exp` should be applied and also what the returned sign should be:

```cpp
std::tuple<double, int> sum_new = log_sum_exp_signed(a, a_sign, b, b_sign);
```

This will be used in the `hypergeometric_*` functions and their derivatives which accumulate an infinite sum on the log-scale while tracking the input and output signs.

## Tests

`prim` tests are added to ensure that different combinations of positive and negative arguments, as well as different magnitudes of arguments (i.e., `a > b` vs `b > a`) are correctly handled, as well as `mix` tests for the gradients

## Side Effects

N/A

## Release notes

Added `log_sum_exp_signed` function for computing `log_sum_exp` while respecting signs of arguments and tracking the sign of the result

## Checklist

- [x] Math issue #2592 

### DIFF
--- a/stan/math/prim/fun.hpp
+++ b/stan/math/prim/fun.hpp
@@ -197,6 +197,7 @@
 #include <stan/math/prim/fun/log_rising_factorial.hpp>
 #include <stan/math/prim/fun/log_softmax.hpp>
 #include <stan/math/prim/fun/log_sum_exp.hpp>
+#include <stan/math/prim/fun/log_sum_exp_signed.hpp>
 #include <stan/math/prim/fun/logical_and.hpp>
 #include <stan/math/prim/fun/logical_eq.hpp>
 #include <stan/math/prim/fun/logical_gt.hpp>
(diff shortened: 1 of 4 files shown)


spinkney:
thanks @andrjohns, this will be nice to have!

rok-cesnovar (approved):
The C++ side of this is good and make sense based on the specified issue. As for the utilization of this in Stan, I will defer to others that know more than me there :)

PR #2837, opened 2022-10-25
ad embedded expression tests to expect_ad testing framework

### LINKED ISSUE #2736: Document Stan Math expressions requirements
Stan Math functions that are meant to be exposed in the Stan language have 2 expressions requirements:
- they have to be able to accept Eigen expressions
- they have to work with the expressions efficiently (not evaluate them multiple times)

This is not really documented or expressed clearly for Stan Math developers I think. We need to mention this somewhere more prominently.

It should also mention that you can test expressions for a function not currently exposed in Stan by running: 
```
python3 runTests.py test/expressions/ --only-function "foo(vector) => vector"
```

SteveBronder:
## Summary

This adds a simple version of the expression testing framework directly into the call to `expect_ad()`. This helps fix #2736 by making sure any function that goes through the expression tests will also go check that it is safe to pass Eigen expressions to.

The files for all of this are in `test/unit/math/expr_tests.hpp` with the main function being `check_expr_test(...)`. This function takes in any set of arguments and for each type that is an Eigen matrix or vector it checks if the function is 'safe' to pass expressions to. The word 'safe' in this context means that, if an unary expression was passed as an argument to the function, is it only called as many times as there are individual elements in the underlying Eigen object? The original Eigen object is wrapped in an unaryFunction that calls the `counterOp` functor from the original expression testing framework.

```cpp
template <typename Scal>
struct counterOp {
  int* counter_;
  counterOp(int* counter) { counter_ = counter; }
  const Scal& operator()(const Scal& a) const {
    (*counter_)++;
    return a;
  }
};
// In the code this is used on a matrix x like 
int call_counter;
auto x_op = x.unaryExpr(counterOp(&call_counter));
double b = x_op[0];
double c = x_op[0];
double d = x_op[0];
std::cout << call_counter << std::endl;
// prints "3"
```

## Tests

Tests can be run with 

```bash
./runTests.py test/unit/math/expr_tests_test.cpp
```

The tests include examples for one, two, and three inputs to functions. They check that for each combination of inputs we fail the correct number of Eigen objects passed to each function. Each set of inputs also has an associated test that just runs a 'good' funtion which does not fail the tests.

## Side Effects

This will increase compile times a bit

## Release notes

Adds a simple expression tests framework to the autodiff test framework

## Checklist

- [x] Math issue #2736


WardBrian:
a concern: this seems to nearly double the time it takes to run the mix unit tests:

A random run from #2983 took 1h31m to run the mix tests: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2983/4/pipeline/209
The latest run here took 2h51m: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2837/9/pipeline/209

That is already the longest part of that parallel stage, so the added time does directly impact the overall runtime of the pipeline

SteveBronder:
I agree it takes more time, but so many countless times we've had users add functions here in Stan math that failed upstream in stanc3 that I think it's worth having these. Also as you can see in the PR there we places in Stan math that actually had small bugs we missed!

WardBrian:
I think there are alternative ways of avoiding the issue only cropping up in stanc3.
 
The simplest I can think of is just having the expression tests (which only take ~25 minutes currently) pull not only from the `stanc --dump-stan-math-signatures` but also from a file like `test/expressions/new_signatures.txt` that users could add to and gets cleaned up whenever a release is cut. This would add at most seconds to the CI time, but require one more check box in the PR template.

If the automated tests here are strictly better, we could also lessen the time/resource impact by removing the previous stage for expression testing?

SteveBronder:
> The simplest I can think of is just having the expression tests (which only take ~25 minutes currently) pull not only from the stanc --dump-stan-math-signatures but also from a file like test/expressions/new_signatures.txt that users could add to and gets cleaned up whenever a release is cut. This would add at most seconds to the CI time, but require one more check box in the PR template.

I don't like this. Our PR process is already pretty bespoke and I think this can suffer a lot from user to keyboard error. The full test suite takes like 12 hours to run so honestly an extra 2 hours is actually kind of fine imo. Like it doesn't change the time I will wait to come back and look if my PR is done.

If we wanted to we could have flags that look at what folders have code changes and only run the expression tests with `double`, `var`, or `fvar`, types depending on that. 

> If the automated tests here are strictly better, we could also lessen the time/resource impact by removing the previous stage for expression testing?

Yes! this covers a lot more of the library, though it does not cover the distributions. But I can modify the distribution code to use the expression test framework as well.

SteveBronder:
> Yeah, I just wish if we were changing this time in increments of 2 hours it would be going the other way. I think a shorter feedback loop would be dramatically beneficial for contributors, even if it was still the kind of thing where you only got 2 tries in a work day (versus the current ~1 try).

I think if we want this as a goal what we could do is have a "quick mode" that is the default when a PR is opened that reads the diff, looks at the name of the files changed, then greps and only runs the tests with the same name in the test folder. And turn off all upstream testing. Then after an initial review we fall over to "test everything mode" which runs overnight.

Or alt we could have the tests and headers only include strictly what they need. Then we can kind of do the same thing as above but with a more broad scope of what tests need to be run. We used to have a setup like that but header errors were extremely annoying so we opted for the "include everything" scheme.

> We could also have a flag (kind of like the current STAN_TEST_ROW_VECTORS) to turn this on or off, the question is then whether we'd like it to be on by default or to be something run when requested for a PR

Is there a way to tell github / jenkins "after this PR is approved run the tests one more time with this new flag and do not allow a merge until all test pass"? That would be a pretty easy / automated way to turn this on and off

WardBrian:
> a "quick mode" that is the default when a PR is opened that reads the diff, looks at the name of the files changed, then greps and only runs the tests with the same name in the test folder

That is already in place, sort of. It just continues on to run everything else after those initial tests. It does make some subset of possible errors fail-fast, which is nice. 

Being much stricter about including what we use would let us run exactly the correct subset of tests, which would be nice. @syclik and I have talked about this a couple times, especially for the distribution tests which also take a huge chunk of time to run.



> Is there a way to tell github / jenkins "after this PR is approved run the tests one more time with this new flag and do not allow a merge until all tests pass"?

I know a lot of projects use a merge bot for things like this. When a approving review is submitted, rather than click merge themselves they leave a comment like `@stan-buildbot merge` which can kick off any extra tests and automatically merge when they pass. For example in Rust, this is done by the @bors bot. You can see an example in action here: https://github.com/rust-lang/rust/pull/123315

syclik:
Just catching up. Given what I've read, I think it's sensible to do two things:

1. Use this PR to improve testing. Yes, it slows it down a couple hours, but when we kick things over to the CI, we should be ok with some time.
2. Separately, we should address the speed issues and really think through what needs to be tested. There are still lots of things we can do to test quickly, then test more thoroughly. Or restrict testing to things that have changed. 

How do we get this moving forward? The loss of two hours of wait time to prevent upstream failures is worth it, I think.

WardBrian:
> The loss of two hours of wait time to prevent upstream failures is worth it

I just wanted to make sure it was clear: these upstream failures are in PRs, not in anything user-facing. So it prevents some developer frustration and churn here in that situation. The same thing could be prevented if the user was told anywhere how to preemptively run the tests, which is what the discussion in #2736 was about

andrjohns:
Needing to compile all of `mix` the tests first, only have to a failure in one or two can lead to a bit of wasted computation. 

The `--jumbo` flag already compiles in batches, would be possible to change the behaviour to run a given batch as soon as it's compiled? That could help things fail faster, especially if we look into tuning the size of the batches

WardBrian:
It would be possible, but I don't think that is an easy change in the current test infrastructure, since `make` doesn't tell you when _some_ targets are done. I'd be happy to be proven wrong there.

andrjohns:
> It would be possible, but I don't think that is an easy change in the current test infrastructure, since `make` doesn't tell you when _some_ targets are done. I'd be happy to be proven wrong there.
> 
> Care would need to be taken to make sure the output was still readable from this

I was thinking of the steps:

1.  Generate batched test files (`prob_1.cpp`, `prob_2.cpp`, etc.), but don't compile
2. Use python's (or bash's) threading to call `runTests.py -j1 x` on batches in parallel

That way the compilation is still parallelised, but each call is only building and immediately running one test. Whether or not this would be particularly 'clean', I'm not sure

WardBrian:
I think something of that flavor could work as long as anything shared between tests (like TBB or Sundials) was already built, otherwise you'd get filesystem races. Worth exploring things like that.

SteveBronder:
@WardBrian do you have any idea what is going on with the benchmark tests failing? I'm not seeing an error in the logs?

https://jenkins.flatironinstitute.org/blue/rest/organizations/jenkins/pipelines/Stan/pipelines/CmdStan%20Performance%20Tests/branches/downstream_tests/runs/953/log/?start=0

SteveBronder:
It looks like the failure is coming from

```
+ curl -s -S 'https://jenkins.flatironinstitute.org/job/Stan/job/CmdStan%20Performance%20Tests/job/downstream_tests/953//logText/progressiveText?start=0'
curl: (7) Failed to connect to jenkins.flatironinstitute.org port 443: Connection timed out
```

Seems weird?

SteveBronder:
@WardBrian I think this is good to merge after the release is finished

Issue #2682, opened 2022-03-02
Adding 7-Parameter Drift Diffusion Model (DDM) PDF with partial derivatives to Stan Math


Franzi2114:
## Description
The Drift Diffusion Model (DDM) introduced by Ratcliff provides a framework for analyzing data of fast binary decisions jointly modelling accuracy and response times. There are several parameterizations of the DDM. Currently Stan Math contains the PDF of the 4-parameter variant (i.e., the parameters for threshold separation, non-decision time, relative starting point, and drift rate) in the function wiener_lpdf(). Furthermore, @kendalfoster is working on a 5-parameter variant of the DDM (i.e., including the parameter for inter-trial variability in the drift rate) (see issue #2584) 

I am working on a pull request to add a 7-parameter variant with partial derivatives of the DDM (i.e., additionaly including the parameters for inter-trial variabilities in relative starting point and non-decision time). For this goal, I first want to implement the lpdf and lcdf of the 4-parameter model `wiener4_lpdf` and `wiener4_lcdf` containing partial derivatives. In the next step, I want to use these functions to expand the model to a 7-parameter model containing partial derivatives `wiener7_lpdf` and `wiener7_lcdf`. These `wiener7` functions will be built in the way that they call the `wiener4` functions for the case that all three variability parameters are set to 0.  Almost all computations are done on the log scale which makes computations more stable. All functions will be based on the implementation of the R-package WienR by @RaphaelHartmann. (Therefore, checks whether the correct values are computed can be done via the WienR package in R).
 
The 4-parameter functions will accept the following parameters: response time,  threshold separation, drift rate, relative starting point, non-decision time, precision, response.

The 7-parameter functions will accept the following parameters: response time,  threshold separation, drift rate, relative starting point, non-decision time, inter-trial variability drift rate, inter-trial variability relative starting point, inter-trial variability non-decision time, precision, response.

(There may occur minor changes in the parameter lists).

The idea is to implement the 7-parameter model in the way that it also can be used as a 4-, 5-, or 6-parameter model. For this, the user only has to set the corresponding variabilities to 0.

## Example
4-parameter model:
An example call to the lpdf is `wiener4_lpdf(1, 1, -1, .5, .1, .01, 1)` and would return the log of the PDF.
An example call to the lcdf is `wiener4_lcdf(1, 1, -1, .5, .1, .01, 1)` and would return the log of the CDF.

7-parameter model
[...]

PR #2841, opened 2022-11-01
Create a class to supersede operands_and_partials for N-ary operations

SteveBronder:
## Summary

This creates a class `propagate_partials` that is a simpler drop in replacement for `operands_and_partials`. A new class was created so that stan math can stay backwards compatible. Underneath the hood `propagate_partials` is very similar to `operands_and_partials`, but now each edge is stored inside of a tuple. This has a small benefit of saving some memory relative to `operands_and_partials` since we previously allocated edges for 5 operations even if we only needed one. The main benefit is that this can support any number of operations and is a drop in replacement for `operands_and_partials`. 

This change also makes for nicer code, where previously we had to include a lengthy template for `operands_and_partials` we can now just use `auto` and call the `make_partials_propagator()` function which then creates and returns the appropriate `partials_propagator` class. This is similar to the standard libraries `make_tuple()` function.

```cpp
  // before
  operands_and_partials<decltype(y_col), decltype(mu_col), decltype(kappa_col)>
      ops_partials(y_col, mu_col, kappa_col);
```

```cpp
  // now
  auto ops_partials = make_partials_propagator(y_col, mu_col, kappa_col);
```

Instead of `operands_and_partials`'s  `edge1_` and `edge2_` the `partials_propagator` class has an associated function called `edge()` which is used in the same way as `std::tuple's` `std::get()` function

```
// before
ops_partials.edge1_.partials_ = elt_divide(-P, Pi_cl);
```

```cpp
// now
edge<0>(ops_partials).partials_ = elt_divide(-P, Pi_cl);
```

## Tests

Since this is a drop in replacement for `operands_and_partials` I took all of the tests for that classe and wrote the corresponding ones for `partials_propagator`. They can be run via

```
./runTests.py -j2 ./test/unit/math/ -f partials_propagator
```
 
## Side Effects

I changed all of the distributions to use `make_partial_propagator` so a lot of files are changed. I did this because I wanted to make sure everything worked and the easiest way was to swap them out for one another.

imo the files that really need looked at thoroughly are the `operands_and_partials` and `partial_propagator` ones in prim/fwd/rev. I'll double check the distributions match correctly.

## Release notes

Allow for for N-ary partials for distributions

## Checklist

- [x] Math issue #2682


SteveBronder:
Staring at the code for using `edge()` and wondering if there should just be a function for directly getting the partials? So instead of 

```cpp
edge<0>(ops_partials).partials_ = elt_divide(-P, Pi_cl);
```

It's possible to just call 

```cpp
partials<0>(ops_partials) = elt_divide(-P, Pi_cl);
```

SteveBronder:
@andrjohns would you mind taking a look at this? I've looked over all of the `_lpdf/_lpmf` files so I think if you look over the actual `partial_propagator` files and `operands_and_partials` files that would be good.

Not super high priority. This is mostly a nice cleanup. There may be some small optimizations for cases with only 1 or 2 parameters since we only store as many edges as we need now (and then only do reverse passes for those 1 or 2 parameters as well). Though I think we would only see a noticeable speedup when applying the scalar form of the lpdf many times.

stan/math/fwd/core/fvar.hpp
```diff
@@ -39,46 +39,45 @@ namespace math {
...
-   * The value of this variable.
-   */
-  T val_;
-
-  /**
-   * The tangent (derivative) of this variable.
+   * The type of values and tangents.
    */
-  T d_;
+  using Scalar = std::decay_t<T>;
```
andrjohns, line 44:
```suggestion
  using ScalarT = std::decay_t<T>;
```

Bit more consistent with the rest of the library

stan/math/fwd/core/fvar.hpp
```diff
@@ -39,46 +39,45 @@ namespace math {
...
+   */
+  Scalar d_;
 
   /**
    * Return the tangent (derivative) of this variable.
    *
    * @return tangent of this variable
    */
-  T d() const { return d_; }
+  Scalar tangent() const { return d_; }
```
andrjohns, line 68: Probably best to leave this as `d()` for this PR, since it's a larger API change - might need wider discussion

stan/math/fwd/functor/operands_and_partials.hpp
```diff
@@ -2,33 +2,201 @@
...
+ * @tparam T2 a type with a `.dx()` method.
+ * @tparam Types a type with a `.dx()` method.
+ * @param a an edge from `operands_and_partials_impl`
+ * @param b an edge from `operands_and_partials_impl`
+ * @param args edges from `operands_and_partials_impl`
+ */
+template <typename T1, typename T2, typename... Types>
+inline auto sum_dx(T1& a, T2& b, Types&... args) {
+  return a.dx() + b.dx() + sum_dx(args...);
+}
```
andrjohns, line 46: Little cleaner to replace this recursion with a fold expression:
```cpp

template<typename... Types>
auto sum_dx(const Types&... args) {
  return_type_t<Types...> result = {};
  (void)std::initializer_list<int>{ (result += v.dx(), 0)... };
  return result;
}
```

stan/math/fwd/functor/operands_and_partials.hpp
```diff
@@ -2,33 +2,201 @@
...
+      : partials_(std::move(other.partials_)),
+        partials_vec_(partials_),
+        operands_(other.operands_) {}
+
+  const Op& operands_;
+  inline Dx dx() {
+    Dx derivative(0);
+    for (size_t i = 0; i < this->operands_.size(); ++i) {
+      derivative += this->partials_[i] * this->operands_[i].d_;
+    }
```
andrjohns, line 107:
```cpp
return dot_product(as_column_vector(this->partials_),
                   as_column_vector(this->operands_).d());
```

stan/math/fwd/functor/operands_and_partials.hpp
```diff
@@ -2,33 +2,201 @@
...
+        operands_(other.operands_) {}
+
+  Op operands_;
 
-  Dx dx() { return this->partials_[0] * this->operand_.d_; }
+  inline Dx dx() {
+    Dx derivative(0);
+    for (int i = 0; i < this->operands_.size(); ++i) {
+      derivative += this->partials_(i) * this->operands_(i).d_;
+    }
```
andrjohns, line 143: Same dot_product suggestion from above

stan/math/fwd/functor/operands_and_partials.hpp
```diff
@@ -2,33 +2,201 @@
...
+    Dx derivative(0);
+    for (size_t i = 0; i < this->operands_.size(); ++i) {
+      derivative += this->partials_[i] * this->operands_[i].d_;
+    }
+    return derivative;
+  }
+};
+
+template <typename Dx, typename ViewElt>
+class ops_partials_edge<Dx, ViewElt, require_eigen_vt<is_fvar, ViewElt>> {
```
andrjohns, line 113: Would it be possible to have a single template for both the `std::vector<>` and `eigen` types? Not much stands out as being too type-dependent

stan/math/prim/functor/operands_and_partials.hpp
```diff
@@ -45,40 +67,42 @@ class ops_partials_edge<ViewElt, Op, require_st_arithmetic<Op>> {
...
+                std::decay_t<ops_partials_edge<
+                    ViewElt, Op, require_st_arithmetic<Op>>>>* = nullptr>
+  constexpr explicit ops_partials_edge(T&& /* op */) noexcept {}
 
   /**
    * Get the operand for the edge. For doubles this is a compile time
    * expression returning zero.
    */
-  static constexpr double operand() noexcept { return 0.0; }
+  static constexpr auto operand() noexcept { return static_cast<double>(0.0); }
```
andrjohns, line 83: Is there a particular benefit for this change? Won't the type deduction be the same?

andrjohns (changes requested):
You had me scared with the size of the diff, but overall the changes are pretty clean and minor! Just a few comments/suggestions

stan/math/fwd/functor/operands_and_partials.hpp
SteveBronder: This is a little confusing imo I'd rather using the recursive parameter expansion to keep it a bit simpler. Inlining should roll out this call for the 2-5 sized argument sizes we normally have

stan/math/fwd/functor/operands_and_partials.hpp
SteveBronder: I think there are ways we could shrink it down but I think it might start looking too abstract. The main thing is the matrix types + constructor

```cpp
  explicit ops_partials_edge(const OpT& ops)
      : partials_(partials_t::Zero(ops.rows(), ops.cols())), operands_(ops) {}
```

stan/math/prim/functor/operands_and_partials.hpp
SteveBronder: No there is not, not sure why I changed this i'll put it back

stan/math/fwd/core/fvar.hpp
SteveBronder: Yes, odd I have no idea why I named this tangent? Removed

stan/math/fwd/core/fvar.hpp
SteveBronder: Just doing a grep for `ScalarT` vs. `Scalar` I think we use `Scalar` more often? I was trying to keep it lined up with Eigen.

stan/math/fwd/functor/operands_and_partials.hpp
andrjohns: Got it!

stan/math/fwd/functor/operands_and_partials.hpp
andrjohns: Fair enough!

stan/math/fwd/core/fvar.hpp
andrjohns: Ah that's right. I was thinking more about the naming conventions we use for template types, safe to ignore

andrjohns (approved):
All LGTM! Feel free to merge if you don't have any other changes

PR #2846, opened 2022-11-14
Add overloads to eigendecomposition for complex matrices

### LINKED ISSUE #2733: Eigenvalues for non-symmetric matrix
I was wondering if Stan developers are working on including a function to calculate eigenvectors and eigenvalues for a non-symmetric matrix. I realize that this feature exists for a symmetric matrix with real eigenvalues. I am assuming that since the support for complex vectors and matrices is going to be released in version 2.30, complex eigenvalue for a non-symmetric matrix should soon follow?

WardBrian:
## Summary

This is a redo of #2732 which I finally got around to fixing the tests on.

This adds wrappers around Eigen's [ComplexEigenSolver](https://eigen.tuxfamily.org/dox/classEigen_1_1ComplexEigenSolver.html) class to add `eigenvalues` and `eigenvectors` functions defined over complex matrices.

## Tests

Existing tests were updated to also test complex matrices.

## Side Effects

None

## Release notes

Added eigenvalues and eigenvectors overloads which accept matrices with complex values.

## Checklist

- [x] Math issue: Closes #2733 

### DIFF
--- a/stan/math/prim/fun/eigenvalues.hpp
+++ b/stan/math/prim/fun/eigenvalues.hpp
@@ -17,7 +17,8 @@ namespace math {
  */
 template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr,
           require_not_vt_complex<EigMat>* = nullptr>
-inline auto eigenvalues(const EigMat& m) {
+inline Eigen::Matrix<complex_return_t<value_type_t<EigMat>>, 1, -1> eigenvalues(
+    const EigMat& m) {
   using PlainMat = plain_type_t<EigMat>;
   const PlainMat& m_eval = m;
   check_nonzero_size("eigenvalues", "m", m_eval);
@@ -30,15 +31,17 @@ inline auto eigenvalues(const EigMat& m) {
 /**
  * Return the eigenvalues of a (complex-valued) matrix
  *
- * @tparam EigMat type of complex matrix argument
+ * @tparam EigCplxMat type of complex matrix argument
  * @param[in] m matrix to find the eigenvectors of. Must be square and have a
  * non-zero size.
  * @return a complex-valued column vector with entries the eigenvectors of `m`
  */
-template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr,
-          require_vt_complex<EigMat>* = nullptr>
-inline auto eigenvalues(const EigMat& m) {
-  using PlainMat = Eigen::Matrix<scalar_type_t<EigMat>, -1, -1>;
+template <typename EigCplxMat,
+          require_eigen_matrix_dynamic_t<EigCplxMat>* = nullptr,
+          require_vt_complex<EigCplxMat>* = nullptr>
+inline Eigen::Matrix<complex_return_t<value_type_t<EigCplxMat>>, 1, -1>
+eigenvalues(const EigCplxMat& m) {
+  using PlainMat = Eigen::Matrix<scalar_type_t<EigCplxMat>, -1, -1>;
   const PlainMat& m_eval = m;
   check_nonzero_size("eigenvalues", "m", m_eval);
   check_square("eigenvalues", "m", m_eval);
--- a/stan/math/prim/fun/eigenvalues.hpp
+++ b/stan/math/prim/fun/eigenvalues.hpp
@@ -7,14 +7,28 @@
 namespace stan {
 namespace math {
 
-template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr>
+template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr,
+          require_not_vt_complex<EigMat>* = nullptr>
 inline auto eigenvalues(const EigMat& m) {
   using PlainMat = plain_type_t<EigMat>;
   const PlainMat& m_eval = m;
   check_nonzero_size("eigenvalues", "m", m_eval);
   check_square("eigenvalues", "m", m_eval);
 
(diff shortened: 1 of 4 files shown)


stan/math/prim/fun/eigenvalues.hpp
```diff
@@ -7,14 +7,28 @@
 namespace stan {
 namespace math {
 
-template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr>
+template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr,
```
syclik, line 10: Please add doxygen doc to the function.

stan/math/prim/fun/eigenvalues.hpp
```diff
@@ -7,14 +7,28 @@
...
+          require_not_vt_complex<EigMat>* = nullptr>
 inline auto eigenvalues(const EigMat& m) {
   using PlainMat = plain_type_t<EigMat>;
   const PlainMat& m_eval = m;
   check_nonzero_size("eigenvalues", "m", m_eval);
```
syclik, line 15: Just wanted to double-check: if there's a 0x0 matrix provided, we're throwing instead of returning back a 0-vector of eigenvalues?

stan/math/prim/fun/eigenvalues.hpp
```diff
@@ -7,14 +7,28 @@
 namespace stan {
 namespace math {
 
-template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr>
+template <typename EigMat, require_eigen_matrix_dynamic_t<EigMat>* = nullptr,
+          require_not_vt_complex<EigMat>* = nullptr>
 inline auto eigenvalues(const EigMat& m) {
```
syclik, line 12: Is this return type `auto` because `solver.eigenvalues()` returns back a template expression?

I looked at the doc and it looks like it's not (but would believe you if you told me the doc wasn't correct). Would it be possible to change this to specify the (templated) return type for clarity?

https://eigen.tuxfamily.org/dox/classEigen_1_1EigenSolver.html#a0f507ad7ab14797882f474ca8f2773e7

stan/math/prim/fun/eigenvectors.hpp
```diff
@@ -18,6 +19,18 @@ inline auto eigenvectors(const EigMat& m) {
   return solver.eigenvectors();
```
syclik, line 19: Same comments as above:
* Needs doxygen doc.
* question about 0-size matrices

syclik (changes requested):

stan/math/prim/fun/eigenvalues.hpp
WardBrian: Yes. This is the behavior of the real-valued versions added in https://github.com/stan-dev/math/pull/1838, but if you think we should return an empty vector instead we could make the change to both signatures

stan/math/prim/fun/eigenvalues.hpp
WardBrian: It's mostly for convenience, since the function accepts a template writing down the return type requires getting the scalar type out of the template and is verbose. I think it is, in full,

`Eigen::Matrix<complex_return_t<value_type_t<EigMat>, 1, -1>`

stan/math/prim/fun/eigenvalues.hpp
WardBrian: The functions `eigen[vectors|values]_sym` both have this behavior as well so I think it's best to maintain it for this PR

andrjohns (dismissed):
Just merged in latest develop to double-check compatibility, all good to merge when tests pass

WardBrian:
I think @syclik still needs to give it another look to clear the "Changes requested" blocker

stan/math/prim/fun/eigenvalues.hpp
syclik: For readability, I think it's worthwhile to include that type for both here and in `eigenvalues(...)`. Would you mind updating that?
(Yes, it's easier to write. But it makes it a lot harder to read later on.)

test/unit/math/mix/fun/eigenvalues_test.cpp
```diff
@@ -15,5 +15,19 @@ TEST(mathMixFun, eigenvalues) {
...
+  };
+  for (const auto& x : stan::test::square_test_matrices(0, 2)) {
+    stan::test::expect_ad(f, x);
+  }
+
+  Eigen::MatrixXd a32(3, 2);
+  a32 << 3, -5, 7, -7.2, 9.1, -6.3;
+  EXPECT_THROW(f(a32), std::invalid_argument);
+}
+
```
syclik, line 31: Can you add a test for complex number that isn't a real number? Just to make sure this is returning values that you expect.

syclik (changes requested):
Requesting a couple of changes:
1. replace `auto` return type with the complex type for readability.
2. add one more test case for eigenvectors and eigenvalues that uses a complex number. (Instead of a real stored as a complex.) Just to make sure it's doing what we expect.

stan/math/prim/fun/eigenvalues.hpp
WardBrian: Will do. Also, just noting for myself, I should change these template parameters to something like EigCplxMat to make it more clear as well

syclik: 👍

WardBrian:
@syclik I believe all your comments have been addressed on this one

syclik (approved):
Awesome!

Thanks for adding those non-trivial tests! That'll make sure that we continue doing the right thing later on.

Issue #2852, opened 2022-12-07
Add `select()` and `any()` helper functions


andrjohns:
## Description

As mentioned in [this PR](https://github.com/stan-dev/math/pull/2784#discussion_r1010558801), it would useful to have helper functions for ternary operations and boolean checks that are agnostic between scalars and containers;

- `select()`
- `any()`
- `all()`

#### Current Version:
v4.5.0

andrjohns:
Closed via #2853

PR #2784, opened 2022-07-08
Improve Numerical Stability of Bernoulli CDF functions

andrjohns:
## Summary

This PR updates the Bernoulli CDF functions (`_cdf`, `_lcdf`, and `_lccdf`) to operate on the log scale as much as possible, to avoid issues with underflow and resolution around 1

## Tests

Additional `mix/prob` tests have been added to ensure that the gradients aren't impacted (`prim` behaviour covered by the distribution tests)

## Side Effects

N/A

## Release notes

Improved numerical stability of Bernoulli CDF functions

## Checklist

- [x] Math issue #2783

- [x] Copyright holder: Andrew Johnson

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #2853, opened 2022-12-09
Add vectorised select(), any(), and all() functions

andrjohns:
## Summary

This PR extracts the vectorised `select()` function introduced in [this PR](https://github.com/stan-dev/math/pull/2784#discussion_r1010558801) (which in turn is adding the currently OpenCL-only function to the rest of the Math library), and also introduces the `any()` and `all()` helper functions.

The `select()` function is used for ternary operations that are intended to be agnostic between scalars and containers. If the function is used with a mixture of scalars and containers, then the scalar is promoted to the same size and type as the container.

## Tests

Tests are added for all combinations of scalar/container inputs, as well as error checking for mismatched container sizes.

## Side Effects

N/A

## Release notes

Added `select()` function for vectorised ternary operations, as well as the `any()` and `all()` boolean reduction functions

## Checklist

- [x] Math issue #2852


andrjohns:
I've also added the `any()` and `all()` functions to this PR since they were fairly simple in comparison (happy to split to separate PR though!)

SteveBronder:
Ping me when this is ready for review!

andrjohns:
> Ping me when this is ready for review!

Thanks @SteveBronder! This should be good to look at now

andrjohns:
@SteveBronder one more thing I've been thinking about, do we want to start requiring/recommending tuple compatibility for utility functions like these? (probably only really relevant for `any()`, and `all()` here)

SteveBronder:
I'm not sure if it should be a requirement, but it would def be nice!

SteveBronder (changes requested):

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,173 @@
...
+inline ReturnT select(const bool c, const T_true& y_true,
+                      const T_false& y_false) {
+  if (c) {
+    return y_true;
+  } else {
+    return apply_scalar_binary(
+        y_true, y_false,
+        [](const auto& y_true_inner, const auto& y_false_inner) {
+          return y_false_inner;
+        });
```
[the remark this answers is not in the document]
andrjohns: The confusion here was probably due to the docs. I had written that the overloads were for a Eigen + scalar combination, but they're actually for general container + scalar combinations. So the `apply_scalar_binary` here allows this to work with arbitrarily nested containers. For example:
```cpp
select(bool, std::vector<std::vector<Eigen::MatrixXd>>, double)
```

andrjohns:
@SteveBronder just addressed your comments. I've also changed the handling for tuples in `all()`/`any()` to use `math::for_each()` if you wouldn't mind double-checking that you're happy with that over the previous `std::vector<bool>` initialiser approach

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,177 @@
...
+ * @tparam T_true type of the true argument
+ * @tparam T_false type of the false argument
+ * @param c Boolean condition value.
+ * @param y_true Value to return if condition is true.
+ * @param y_false Value to return if condition is false.
+ */
+template <typename T_true, typename T_false,
+          require_all_stan_scalar_t<T_true, T_false>* = nullptr>
+inline auto select(const bool c, const T_true y_true, const T_false y_false) {
+  return c ? y_true : y_false;
```
SteveBronder, line 25: Should we cast y_true / y_false to the same type explicitly or is it okay if we assume they are implicitly promotable?

stan/math/prim/fun/select.hpp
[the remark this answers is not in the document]
SteveBronder: I'm still kind of confused. If `c` is false then we go the the `apply_scalar_binary` branch which then returns `y_false_inner`. I'm not really following how a double would be promoted to a std::vector<std::vector<Eigen::MatrixXd>> in your code above?

```c++
double dbl = 3.4;
std::vector<std::vector<Eigen::MatrixXd>> stdvec = // ...
std::vector<std::vector<Eigen::MatrixXd>> result = select(true, stdvec, dbl);
```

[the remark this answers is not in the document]
SteveBronder: Can you add a test for `std::vector<std::vector<Eigen::MatrixXd>>` and `double`?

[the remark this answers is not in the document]
andrjohns: Because the return type of the `apply_scalar_binary` call will be the container that is passed, and the functor is specifying that every element of the container is the `y_false` value (which is a scalar in this overload)

Also added a test of these different combinations so you can see the behaviour in action as well

SteveBronder:
I'll take a look at this today

stan/math/prim/fun/all.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Return true if all values in the input are true.
+ *
+ * Overload for a single boolean input
+ *
+ * @param x boolean input
+ * @return The input unchanged
+ */
+inline bool all(bool x) { return x; }
```
SteveBronder, line 19: (optional) constexpr?

SteveBronder: Also should this be templated to accept any integral type? Not sure if we would get warnings about downcasting? This applies to all the bool as input functions

stan/math/prim/fun/any.hpp
```diff
@@ -0,0 +1,77 @@
...
+ * @return The input unchanged
+ */
+inline bool any(bool x) { return x; }
+
+/**
+ * Return true if any values in the input are true.
+ *
+ * Overload for Eigen types
+ *
+ * @tparam Eigen type of the input
```
SteveBronder, line 26:
```suggestion
 * @tparam ContainerT A type derived from `Eigen::EigenBase` that has an `integral` scalar type
```

Double check the docs for all these

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,176 @@
...
+#include <stan/math/prim/functor/apply_scalar_binary.hpp>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return the second argument if the first argument is true
+ * and otherwise return the third argument.
+ *
+ * <code>select(c, y1, y0) = c ? y1 : y0</code>.
```
SteveBronder, line 15: (optional) I just prefer using ticks here \` doxygen is able to parse that like it's markdown

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,176 @@
...
+ * @param y_true Value to return if condition is true.
+ * @param y_false Value to return if condition is false.
+ */
+template <typename T_true, typename T_false,
+          typename ReturnT = return_type_t<T_true, T_false>,
+          require_all_stan_scalar_t<T_true, T_false>* = nullptr>
+inline ReturnT select(const bool c, const T_true y_true,
+                      const T_false y_false) {
+  return c ? ReturnT(y_true) : ReturnT(y_false);
+}
```
SteveBronder, line 29: `ReturnT` not doc'd here

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,176 @@
...
+    typename T_return = return_type_t<T_true, T_false>,
+    typename T_true_plain = promote_scalar_t<T_return, plain_type_t<T_true>>,
+    typename T_false_plain = promote_scalar_t<T_return, plain_type_t<T_false>>,
+    require_all_container_t<T_true, T_false>* = nullptr,
+    require_all_same_t<T_true_plain, T_false_plain>* = nullptr>
+inline T_true_plain select(const bool c, const T_true y_true,
+                           const T_false y_false) {
+  check_matching_dims("select", "y_true", y_true, "y_false", y_false);
+  return c ? T_true_plain(y_true) : T_true_plain(y_false);
+}
```
SteveBronder, line 58: (optional) it may be nice to have a signature for when `T_true_plain` is not constructable from a `T_false_plain` that has a static assert in it with a nice message.

SteveBronder, line 58: If these are containers should they be passed as references like 

```suggestion
inline T_true_plain select(const bool c, T_true&& y_true,
                           T_false&& y_false) {
  check_matching_dims("select", "y_true", y_true, "y_false", y_false);
  return c ? T_true_plain(std::forward<T_true>(y_true)) : T_true_plain(std::forward<T_false>(y_false));
}
```

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,176 @@
...
+ * @tparam T_false type of the false argument
+ * @param c Eigen object of boolean condition values.
+ * @param y_true Value to return if condition is true.
+ * @param y_false Value to return if condition is false.
+ */
+template <typename T_bool, typename T_true, typename T_false,
+          require_eigen_array_vt<std::is_integral, T_bool>* = nullptr,
+          require_all_stan_scalar_t<T_true, T_false>* = nullptr>
+inline auto select(const T_bool c, const T_true y_true, const T_false y_false) {
+  return c.unaryExpr([&](bool cond) { return cond ? y_true : y_false; }).eval();
```
SteveBronder, line 149: (optional) I personally prefer we just pull in the names of the things we want into a lambda instead of using &
```suggestion
  return c.unaryExpr([y_true, y_false](bool cond) { return cond ? y_true : y_false; }).eval();
```

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,176 @@
...
+ * @param c Eigen object of boolean condition values.
+ * @param y_true Value to return if condition is true.
+ * @param y_false Value to return if condition is false.
+ */
+template <typename T_bool, typename T_true, typename T_false,
+          require_eigen_array_t<T_bool>* = nullptr,
+          require_any_eigen_array_t<T_true, T_false>* = nullptr>
+inline auto select(const T_bool c, const T_true y_true, const T_false y_false) {
+  check_consistent_sizes("select", "boolean", c, "y_true", y_true, "y_false",
+                         y_false);
```
SteveBronder, line 169: (optional) Should the names here be left hand side and right hand side? Like can this error be propagated up to users of Stan?

SteveBronder (changes requested):
Mostly doc fixes but once that's done then I think good to merge!

stan/math/prim/fun/select.hpp
andrjohns: `ReturnT` is a derived type (the result of `return_type_t<...>`), and we haven't usually doc'd them AFAIK

andrjohns:
@SteveBronder this is ready for another look

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,180 @@
...
+inline ReturnT select(const bool c, const T_true& y_true,
+                      const T_false& y_false) {
+  if (c) {
+    return y_true;
+  } else {
+    return apply_scalar_binary(
+        y_true, y_false,
+        [](const auto& y_true_inner, const auto& y_false_inner) {
+          return y_false_inner;
+        });
```
SteveBronder, line 92: What is the use case for the container and scalar overload? Is this for like broadcasting? imo I'd rather do this sort of conditional at compile time and return either a scalar or a container like `as_array_or_scalar()`

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,180 @@
...
+namespace stan {
+namespace math {
+
+/**
+ * Return the second argument if the first argument is true
+ * and otherwise return the third argument.
+ *
+ * `select(c, y1, y0) = c ? y1 : y0`.
+ *
+ * @tparam T_true type of the true argument
```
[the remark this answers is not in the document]
SteveBronder: Please fix this for all the docs for the types

stan/math/prim/fun/select.hpp
SteveBronder: My worry here is that folks are going to accidentally use this overload and get back a container when they instead wanted a scalar.

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,180 @@
...
+ * @param y_false Value to return if condition is false.
+ */
+template <typename T_bool, typename T_true, typename T_false,
+          require_eigen_array_vt<std::is_integral, T_bool>* = nullptr,
+          require_all_stan_scalar_t<T_true, T_false>* = nullptr>
+inline auto select(const T_bool c, const T_true y_true, const T_false y_false) {
+  return c
+      .unaryExpr(
+          [y_true, y_false](bool cond) { return cond ? y_true : y_false; })
+      .eval();
```
SteveBronder, line 153: How is the return type decided here? I think we should do like the other functions and explicitly cast to our intended return type
```suggestion
 using ret_t = return_type_t<T_true, T_false>;
  return c
      .unaryExpr(
          [y_true, y_false](bool cond) { return cond ? ret_t(y_true) : ret_t(y_false); })
      .eval();
```

stan/math/prim/fun/select.hpp
```diff
@@ -0,0 +1,180 @@
...
+ * @param y_true Value to return if condition is true.
+ * @param y_false Value to return if condition is false.
+ */
+template <typename T_bool, typename T_true, typename T_false,
+          require_eigen_array_t<T_bool>* = nullptr,
+          require_any_eigen_array_t<T_true, T_false>* = nullptr>
+inline auto select(const T_bool c, const T_true y_true, const T_false y_false) {
+  check_consistent_sizes("select", "boolean", c, "left hand side", y_true,
+                         "right hand side", y_false);
+  return c.select(y_true, y_false).eval();
```
SteveBronder, line 174: Same as before, I think we should be casting `y_true` and `y_false` to the same type before doing the select. idk how Eigen handles that under the hood and I'd rather just be explicit about it

SteveBronder (changes requested):
Mostly Qs on the signatures for containers and scalars as well as some docs.

stan/math/prim/fun/select.hpp
andrjohns: Yep the broadcasting is the intended outcome (and is doc-ed). 

See [this line from the Bernoulli PR](https://github.com/stan-dev/math/pull/2784/files#diff-1aafd077ec7f8966bf7f92be2addaee531c6be68b53a5e798681c8c279e60124R63) for an example of usage where the inputs can be scalars or containers, and the return needs to be broadcast (if necessary) to the common container type & size

andrjohns:
@SteveBronder ready for another look

stan/math/prim/fun/all.hpp
```diff
@@ -0,0 +1,85 @@
...
+  return std::all_of(x.begin(), x.end(), [](const auto& i) { return all(i); });
+}
+
+/**
+ * Return true if all values in the input are true.
+ *
+ * Overload for a tuple input.
+ *
+ * @tparam Type of container
+ * @param x Nested container of boolean inputs
```
SteveBronder, line 67: docs do not match signature

stan/math/prim/fun/any.hpp
```diff
@@ -0,0 +1,85 @@
...
+namespace stan {
+namespace math {
+
+/**
+ * Return true if any values in the input are true.
+ *
+ * Overload for a single boolean input
+ *
+ * @tparam T The type of integral input.
+ * @param x boolean input
```
SteveBronder, line 17:
```suggestion
 * @param x integral input
```

stan/math/prim/fun/any.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace stan {
+namespace math {
+
+/**
+ * Return true if any values in the input are true.
+ *
+ * Overload for a single boolean input
+ *
+ * @tparam T The type of integral input.
```
SteveBronder, line 16:
```suggestion
 * @tparam T Any type convertible to `bool`
```

SteveBronder (changes requested):
Some very minor comments on the docs then I think this is good to go!

SteveBronder (approved):
lgtm!

Issue #2857, opened 2022-12-30
`grad2F1` can terminate too early when only some gradients are requested


ricardoV94:
The termination criteria uses the maximum change for the 3 gradients (wrt a1, a2 and b1), which makes it less strict when only a subset is being evaluated.

https://github.com/stan-dev/math/blob/95abd90d38259f27c7a6013610fbc7348f2fab4b/stan/math/prim/fun/grad_2F1.hpp#L117

I get an absolute difference of ~`4.3e-4` across the two evaluations shown in the snippet:
```cpp
#include <stan/math.hpp>
#include <iostream>

int main() {

  double a1 = -0.5;
  double a2 = -4.5;
  double b1 = 11.0;
  double z = 0.3;

  auto grad_tuple1 = stan::math::internal::grad_2F1_impl<true, true, true, true>(a1, a2, b1, z);
  std::cout << std::get<0>(grad_tuple1) << std::endl;

  auto grad_tuple2 = stan::math::internal::grad_2F1_impl<true, false, false, false>(a1, a2, b1, z);
  std::cout << std::get<0>(grad_tuple2) << std::endl;
}
```
```bash
# Number of iterations:9 
-0.122702
# Number of iterations:2
-0.122727
```

The result obtained in the second case would probably fail the precision tested here: https://github.com/stan-dev/math/blob/2e174d22b8ad9d9fd4c6a1e0aeff48700b74c951/test/unit/math/prim/fun/grad_2F1_test.cpp#L116

Perhaps a simple solution is to introduce a minimum number of iterations?

andrjohns:
Thanks so much for catching this (I was the one that introduced this behaviour)! I've opened a PR to add a minimum number of iterations

andrjohns:
@ricardoV94 Just a heads up that I've also found a couple of minor bugs in the `grad_2F1` implementation while testing this, and will have a PR up next week with the bugfixes and more comprehensive tests

ricardoV94:
@andrjohns thanks for the heads up. Where those bugs restricted to the Euler approximation by any chance?

andrjohns:
There was one bug with the Euler approximation on L213](https://github.com/stan-dev/math/blob/3cbf8e1d978a780d03b4e465fcc0fcd8ae53bde9/stan/math/prim/fun/grad_2F1.hpp#L213):
```cpp
auto hyper_da2 = hypergeometric_2F1(a1_euler, a2, b1, z_euler);
```

[...]

PR #2858, opened 2023-01-02
[BUGFIX] grad_2f1: Add minimum number of iterations and early stopping test

andrjohns:
## Summary

As mentioned in #2857, the `grad_2f1` function can falsely converge early when only calculating gradients w.r.t. a subset of parameters. This PR introduces a minimum number of iterations (5) to avoid this and a test to verify

## Tests

Test-case from #2857 added

## Side Effects

N/A

## Release notes

Added a minimum number of iterations (5) to the `grad_2f1` function to avoid early convergence

## Checklist

- [x] Math issue #2857 

### DIFF
--- a/stan/math/prim/fun/grad_2F1.hpp
+++ b/stan/math/prim/fun/grad_2F1.hpp
@@ -76,7 +76,7 @@ TupleT grad_2F1_impl_ab(const T1& a1, const T2& a2, const T3& b1, const T_z& z,
   ScalarT inner_diff = 1;
   ScalarArrayT g_current = ScalarArrayT::Zero(3);
 
-  while (inner_diff > precision && k < max_steps) {
+  while ((inner_diff > precision || k < 5) && k < max_steps) {
     ScalarT p = ((a1 + k) * (a2 + k) / ((b1 + k) * (1.0 + k)));
     if (p == 0) {
       return grad_tuple;
--- a/stan/math/prim/fun/grad_2F1.hpp
+++ b/stan/math/prim/fun/grad_2F1.hpp
@@ -73,10 +73,11 @@ TupleT grad_2F1_impl_ab(const T1& a1, const T2& a2, const T3& b1, const T_z& z,
 
   int sign_zk = sign_z;
   int k = 0;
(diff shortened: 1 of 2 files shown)


andrjohns:
@spinkney would you be able to have a look at this? Do you think 5 iterations is enough or should we go for 10?

ricardoV94:
> @spinkney would you be able to have a look at this? Do you think 5 iterations is enough or should we go for 10?

I used 10 when adapting your code here https://github.com/pymc-devs/pytensor/pull/90

I am using the same test cases you guys had, but the algorithm always computes one partial derivative at a time. With 5 the tests wouldn't all pass, but I am not sure I am asserting the same precision as the STAN tests.

stan/math/prim/fun/grad_2F1.hpp
```diff
@@ -76,7 +76,7 @@ TupleT grad_2F1_impl_ab(const T1& a1, const T2& a2, const T3& b1, const T_z& z,
   ScalarT inner_diff = 1;
   ScalarArrayT g_current = ScalarArrayT::Zero(3);
 
-  while (inner_diff > precision && k < max_steps) {
+  while ((inner_diff > precision || k < 5) && k < max_steps) {
```
syclik, line 79: Just above this, can we add:

```
int min_steps = 5;
```

then change the conditional to use the named variable? It'll make it easier to parse.

syclik (changes requested):
@andrjohns, just asking for a change for readability. I don't think it's worth adding it to the signature.

I think it'd be ok bumping it up to 10 if you want. Or leaving it at 5. It is pretty arbitrary at this point.

spinkney:
> @spinkney would you be able to have a look at this? Do you think 5 iterations is enough or should we go for 10?

I think 10 is better. Is there any way you can do a grid search across possible values to see if there is a problematic range where even 10 iterations is not enough? If you can't find any ranges that's fine. If you do find a problematic range then we can document it and see what else we can do (more iterations or notes to future devs to improve the algorithm in that range).

syclik (approved):

PR #2864, opened 2023-01-19
Conditionally run distribution tests

WardBrian:
Working with @syclik on speeding up CI and only running what we need. For now this is a messy attempt, but we need a PR open to run CI against it


WardBrian:
https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/PR-2864/13/pipeline/515 is a great example of the purpose of this (before @syclik very helpfully fixed the includes on the rest of the tests!) since the only distribution tests that ran were the ones we had changed up to that point.

syclik (approved):
I looked over the Python. It looks fine to me.

The distributions I walked through by hand. I tested all but the wiener distributions (I've compiled ~80% so far, but it's going to take a while longer to see the results locally, so I can hold off on it.)

PR #2869, opened 2023-01-30
Explicitly test for empty distribution test changelist

WardBrian:
Follow on to #2864 after test failures in #2846. It seems that the list of changes being empty still tried to run some tests (I assumed the `for` loop would just never execute) so this is making that conditional explicit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -472,8 +472,13 @@ pipeline {
 
         stage ('Distribution tests') {
             when {
-                expression {
-                    !skipRemainingStages
+                allOf {
+                    expression {
+                        !skipRemainingStages
+                    }
+                    expression {
+                        !changedDistributionTests.isEmpty()
+                    }
                 }
             }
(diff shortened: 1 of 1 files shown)


syclik (dismissed):
LGTM

serban-nicusor-toptal (approved):
LGTM! Thanks

WardBrian:
@serban-nicusor-toptal this still didn't work - is there a good way to print a Groovy variable for debugging? For some reason I guess the list isn't empty but also doesn't contain a test, so I'm worried the value is like `[""]`

WardBrian:
Okay, I think I figured it out. Using `.readLines()` rather than  `.split('\n').toList()` seems to do what we want when the string is only whitespace

serban-nicusor-toptal (approved):

serban-nicusor-toptal:
`println` inside a script block, I saw that you already figured it out. Sorry, but I saw the message now.

PR #2874, opened 2023-02-11
Use unambiguous size function with old rstan

andrjohns:
@hsbadr

When using c++17 with rstan 2.21 & StanHeaders 2.26 we run into the same ambiguity errors with the `size()` function - because the older `stanc` is emitting `size()` without the math qualifier. We can workaround this by using the "old" size function with rstan 2.21, which has less general templating and doesn't result in the same ambiguity errors

EDIT: This will be needed for the `blavaan` package

### DIFF
--- a/stan/math/prim/fun/size.hpp
+++ b/stan/math/prim/fun/size.hpp
@@ -18,6 +18,8 @@ inline size_t size(const T& /*x*/) {
   return 1U;
 }
 
+#ifdef USE_STANC3
+
 /** \ingroup type_trait
  * Returns the size of the provided Eigen matrix, expression or std::vector.
  *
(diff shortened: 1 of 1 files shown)


andrjohns:
@hsbadr sorry to double-ping, but would you be able to merge this and create a new StanHeaders release when you get a minute?

stan/math/prim/fun/size.hpp
```diff
@@ -29,11 +31,39 @@ inline size_t size(const T& m) {
...
+}
+
+#endif
+
 template <typename T, require_var_matrix_t<T>* = nullptr>
 inline size_t size(const T& m) {
   return m.size();
 }
 
+
```
hsbadr, line 66:
```suggestion

```

hsbadr (approved):
LGTM.

Issue #2872, opened 2023-02-07
Testing: decrease impact of distribution tests on CI hardware


WardBrian:
Recent Jenkins runs have been failing due to disk space, see e.g. [develop/139](https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FMath/detail/develop/139/pipeline). 

@dylex informed me that the math workspaces were taking up nearly 1.2TB of disk space on the CI machines at Flatiron. It seems that this is a combination of us not cleaning up after ourselves properly and a recent increase in space per run, possibly due to #2864.

To mitigate this, we should both:

1. Properly clean up workspaces and discuss a policy to delete older ones automatically. This would unfortunately make the logs from old CI runs inaccessible after the chosen amount of time (suggestions from @dylex on how to implement this would be great).
2. Investigate why the increase in space happened, AND/OR consider moving the full run of the distribution tests to something which happens not on every merge but weekly/on a schedule

Pinging @rok-cesnovar and @serban-nicusor-toptal for thoughts and help

rok-cesnovar:
Distribution disk space has always been kind of a problem (https://github.com/stan-dev/math/issues/397 - though this is on a completely different level size wise).

What uses the most space? Logs or executables? I imagine the created source tests files don't consume as much space?

WardBrian:
I believe it is the object files, @dylex can confirm

dylex:
The bulk of the space was `test/prob/{skew_normal,pareto_type_2,...}/*_generated_vv_test.o` files, a few thousand of them, each about 35M.

WardBrian:
We kill any previous builds when we fire up a new one in PRs, but we intentionally exempt `develop` from this. I am not sure this is always necessary, besides if two PRs get merged close in time and we want to be absolutely certain we can determine which one failed (and this wouldn't be obvious from the test failure itself)?

PR #2878, opened 2023-03-01
Delete object files after test executable is built

WardBrian:
## Summary

Update the `make` rule for the test executables such that the `.o` files are deleted after they're done being used for linking.

This should help with #2872 

## Tests

None

## Side Effects

Hopefully none. The `.o` files would be re-built anyway of the executable needed to be updated, so these were truly only used once.

## Checklist

- [x] Math issue #2872 

### DIFF
--- a/make/tests
+++ b/make/tests
@@ -24,9 +24,9 @@ test/% : CXXFLAGS += $(CXXFLAGS_GTEST)
 test/% : CPPFLAGS += $(CPPFLAGS_GTEST)
 test/% : INC += $(INC_GTEST)
 
-test/%_test$(EXE) : test/%_test.cpp $(GTEST)/src/gtest_main.cc $(GTEST)/src/gtest-all.o $(MPI_TARGETS) $(TBB_TARGETS)
+test/%$(EXE) : test/%.o $(GTEST)/src/gtest_main.cc $(GTEST)/src/gtest-all.o $(MPI_TARGETS) $(TBB_TARGETS)
 	$(LINK.cpp) $^ $(LDLIBS) $(OUTPUT_OPTION)
-
+	$(RM) test/$(subst  \,/,$*).o
 
 ##
 # Include dependency files for tests
--- a/make/tests
+++ b/make/tests
@@ -26,13 +26,15 @@ test/% : INC += $(INC_GTEST)
 
 test/%$(EXE) : test/%.o $(GTEST)/src/gtest_main.cc $(GTEST)/src/gtest-all.o $(MPI_TARGETS) $(TBB_TARGETS)
 	$(LINK.cpp) $^ $(LDLIBS) $(OUTPUT_OPTION)
-	$(RM) test/$(subst  \,/,$*).o
 
 ##
 # Include dependency files for tests
+# This has the side-effect of making the object files non-intermediates,
+# so we need to explicitly tell make we don't want to keep them
 ##
 ifneq ($(filter test/%,$(MAKECMDGOALS)),)
 -include $(patsubst %$(EXE),%.d,$(filter test/%,$(MAKECMDGOALS)))
+.INTERMEDIATE: $(patsubst %$(EXE),%.o,$(filter test/%,$(MAKECMDGOALS)))
 endif
 
 ##
--- a/make/tests
+++ b/make/tests
@@ -34,7 +34,7 @@ test/%$(EXE) : test/%.o $(GTEST)/src/gtest_main.cc $(GTEST)/src/gtest-all.o $(MP
 ##
 ifneq ($(filter test/%,$(MAKECMDGOALS)),)
 -include $(patsubst %$(EXE),%.d,$(filter test/%,$(MAKECMDGOALS)))
-.INTERMEDIATE: $(patsubst %$(EXE),%.o,$(filter-out test/%.d test/%.o,$(filter test/%,$(MAKECMDGOALS))))
+.INTERMEDIATE: $(patsubst test/%_test$(EXE),test/%_test.o,$(filter test/%_test$(EXE),$(MAKECMDGOALS)))
 endif
 
 ##
--- a/make/tests
+++ b/make/tests
@@ -34,7 +34,7 @@ test/%$(EXE) : test/%.o $(GTEST)/src/gtest_main.cc $(GTEST)/src/gtest-all.o $(MP
 ##
 ifneq ($(filter test/%,$(MAKECMDGOALS)),)
 -include $(patsubst %$(EXE),%.d,$(filter test/%,$(MAKECMDGOALS)))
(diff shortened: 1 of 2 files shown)


syclik (changes requested):
@WardBrian, I get what you're trying to do. Let me take a look at it carefully.

This isn't the way you want to do this in make. It violates how make is supposed to work. 

Do you have an example where `make test/...` generates a `.o` and leaves it behind? I'll try to trace why that happens.

WardBrian:
Currently if you build any of the tests, it does leave it behind. The easiest way to see this is to run the runTests.py file on a small sub folder and see what’s left behind. We also have a part of `make clean` which specifically deletes the test `.o` files: https://github.com/stan-dev/math/blob/edfc5b8fd2c4d715448e7ccfa7aba86711da5ca7/makefile#L111

Currently each run of the develop branch takes ~400 gigs of hard drive space, with the left over .o files being the main culprit. 

The change here is basically exactly what we do in CmdStan: https://github.com/stan-dev/cmdstan/blob/f6e0124fc125e0a417b5f308ad476cd5b9828ee4/make/program#L60

syclik:
Ok, I can reproduce the behavior. I should be able to hunt down why it's
doing this; it shouldn't. We've probably got some spurious make instruction
that keeps it around.

Let me try to fix it properly. If I can't figure it out, we can merge this
PR in.

syclik:
Here's the test I'm running:

make test/unit/math_c11_test

WardBrian:
I believe the `.d` files, because they specify prerequisites for the `.o` files, make those `.o` files no longer considered as immediate files by make, so they aren’t automatically deleted. 

I believe we could use `.IMMEDIATE: %.o`? Not sure if % works for .IMMEDIATE, it is mentioned as a special case that it works for .PRECIOUS

WardBrian:
Actually, even without the .d files I don’t think these files would be candidates for deletion since they’re explicit prerequisites:

> A file cannot be intermediate if it is mentioned in the makefile as a target or prerequisite, so one way to avoid the deletion of intermediate files is by adding it as a prerequisite to some target.

https://www.gnu.org/software/make/manual/html_node/Chained-Rules.html

If explicitly marking them as immediate works that would probably be better, though I can imagine there being some .o files we actually want to keep around? For example, the gtest main is only compiled to a `.o` file once, I believe, which is good.

syclik:
@WardBrian, I just pushed a fix to the branch. Take a look?

This relies on the `.o` to be built using a chained rule, which means that make knows it shouldn't hold on to it.

syclik:
I think this is the right thing to do to stay consistent with the way make works.

WardBrian:
Do the dependency files still work after that change?

syclik:
I think so.

1. It generates a .d for that test.
2. The contents of that particular test file are the same on develop vs the
branch.

I haven't checked against another, more complicated test.

WardBrian:
And it actually deletes the .o file? My reading of the make documentation is that it should not:

> Note that such a rule constitutes mentioning main.o in a makefile, so it can never be considered an intermediate file by implicit rule search. This means that make won’t ever remove the file after using it

https://www.gnu.org/software/make/manual/html_node/Automatic-Prerequisites.html

syclik:
The .o disappears.

But I'm checking for the dependencies now.

syclik:
Dependencies break. I'm going to crash, but will look at it tomorrow.

WardBrian:
I tried `.INTERMEDIATE: test/%.o`, which had no effect. It seems that using `%` for the special targets is indeed unique to `.PRECIOUS`.

But, it seems like we can do something like `.INTERMEDIATE: $(patsubst %$(EXE),%.o,$(filter test/%,$(MAKECMDGOALS)))`

This has the downside that this makes it impossible to build a just the `.o` file if you ever needed to for some reason, e.g. `make test/unit/math/mix/fun/norm1_test.o` would now yield `make: *** No rule to make target 'test/unit/math/mix/fun/norm1_test.o.o', needed by 'test/unit/math/mix/fun/norm1_test.o'.  Stop.`

But, if you don't want to directly `rm` the `.o` file, I think it is the best solution

WardBrian:
I confirmed that dependencies still work and that the .o files are deleted for all our use cases, including things like running `runTests.py` on a whole folder or with a wildcard in the path

syclik (approved):
This is good. Please merge!

@WardBrian and I tried a bunch more things that didn't succeed; I think this is the cleanest way to achieve this for now.

serban-nicusor-toptal (dismissed):

serban-nicusor-toptal (dismissed):

WardBrian:
Okay, the changes in 60af505 made me feel like there had to be a better way to prevent the error if you run `make test/some_file.o` directly. And it turns out there is, we just need to make sure a file doesn't already end in `.o` while building the `.INTERMEDIATE` list

make/tests
```diff
@@ -29,9 +29,12 @@ test/%$(EXE) : test/%.o $(GTEST)/src/gtest_main.cc $(GTEST)/src/gtest-all.o $(MP
 
 ##
 # Include dependency files for tests
+# This has the side-effect of making the object files non-intermediates,
+# so we need to explicitly tell make we don't want to keep them
 ##
 ifneq ($(filter test/%,$(MAKECMDGOALS)),)
 -include $(patsubst %$(EXE),%.d,$(filter test/%,$(MAKECMDGOALS)))
+.INTERMEDIATE: $(patsubst %$(EXE),%.o,$(filter-out test/%.d test/%.o,$(filter test/%,$(MAKECMDGOALS))))
```
syclik, line 37: Just a suggestion or maybe a thought on how to better pattern match, how about matching the executable a little more:

```
.INTERMEDIATE: $(patsubst test/%_test$(EXE),test/%_test.o,$(filter test/%_test$(EXE),$(MAKECMDGOALS)))
```

(I didn't run it... so it might not be correct.)

what I'm thinking is that we know that we want a list of `.o` when the target looks like `test/*_test` (or `test/*_test.exe`).

If the `.o` were provided as a target directly we could leave it alone?

syclik (dismissed):
LGTM. I added a comment just to see if that makes sense.

make/tests
WardBrian: This worked better, thanks!

syclik (approved):
LGTM!

I guess the logic worked.

WardBrian:
Not that this is an issue, I just want to note for posterity the difference in behavior between the current approach and the original pr: 

Manually adding a portion to the executable rule which removes the `.o` files has the effect of removing them as it goes along, so there are never more than `$JOBS` `.o` files in existence at once. The behavior `make` uses for intermediates is instead to delete them all in a single `rm` invocation at the _end_ of building, so there can be quite a lot of them before they're eventually cleaned up. 

If we in the future have problems with peak space usage, then changing either the makefile or how we invoke `make` could reduce this.

PR #2888, opened 2023-03-27
StanHeaders 2.26 - Backport Eigen 3.4 Compatibility

andrjohns:
@hsbadr This PR backports the [Math header changes](https://github.com/stan-dev/math/pull/2583/files?file-filters%5B%5D=.hpp&show-deleted-files=true&show-viewed-files=true) from the Eigen 3.4 compatibility PR to StanHeaders 2.26

With this PR and #2887, both StanHeaders 2.26 and 2.31 will be compatible with both RcppEigen 3.3.9 and 3.4.0. This also resolves the ctsem error from the [update issue](https://github.com/RcppCore/RcppEigen/issues/103), and means that all rstan-based packages will be compatible with RcppEigen 3.4.0 once StanHeaders 2.26 is on CRAN

### DIFF
--- a/stan/math/prim/fun/mdivide_left_tri.hpp
+++ b/stan/math/prim/fun/mdivide_left_tri.hpp
@@ -26,20 +26,19 @@ namespace math {
 template <Eigen::UpLoType TriView, typename T1, typename T2,
           require_all_eigen_t<T1, T2> * = nullptr,
           require_all_not_eigen_vt<is_var, T1, T2> * = nullptr>
-inline Eigen::Matrix<return_type_t<T1, T2>, T1::RowsAtCompileTime,
-                     T2::ColsAtCompileTime>
-mdivide_left_tri(const T1 &A, const T2 &b) {
+inline auto mdivide_left_tri(const T1 &A, const T2 &b) {
   using T_return = return_type_t<T1, T2>;
+  using ret_type = Eigen::Matrix<T_return, Eigen::Dynamic, Eigen::Dynamic>;
   check_square("mdivide_left_tri", "A", A);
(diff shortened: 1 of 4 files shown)


hsbadr:
Why would we need this? `RcppEigen` hasn't been updated on CRAN, and `StanHeaders` v2.26 is a temporary release that will be followed by v2.31+ that's compatible with Eigen v3.4.

andrjohns:
So that there's no dependence on either package being released first. Currently StanHeaders 2.31 requires RcppEigen 3.4.0, but v3.4.0 requires StanHeaders 2.31

hsbadr:
This will break some dependencies with the current version of `RcppEigen` on CRAN. I suggest that we release v2.26 now & take care of Eigen update with the next release.

andrjohns:
> This will break some dependencies with the current version of RcppEigen on CRAN

What do you mean? This is all valid 3.3.9 syntax, it makes 2.26 compatible with both

> I suggest that we release v2.26 now & take care of Eigen update with the next release.

In an ideal world that would be the solution, but we don't know much longer it's going to take to get rstan to CRAN after StanHeaders, before StanHeaders 2.31 can be submitted

andrjohns:
I've run them with 2.31 (and #2887) and RcppEigen 3.3.9, and all packages (or their compatibility PRs) passed

But this is also what I mean - if v2.31 were to depend on RcppEigen 3.4 for checks to pass, then RcppEigen needs to be released first and compatible with 2.26.

andrjohns:
But at the same time, this is more of a just-in-case PR and can always be included in a future 2.26 CRAN update if needed. I'm happy to close now and we can revisit later if it comes up

PR #2905, opened 2023-05-17
Feature/threadsafe matrixcl

SteveBronder:
## Summary

This makes `matrix_cl`'s  event vectors thread safe by using `tbb::concurrent_vector` instead of a `std::vector`.

In Stan a `matrix_cl` can be in a model class and shared across multiple threads. When using multiple threads we can have a race condition where two operations attempt to push to the vector of read/write events.

I think the simplest fix for this is the below where we replace `std::vector` with `tbb::concurrent_vector`. This has some issues though, notably that `tbb::concurrent_vector` is not a contiguous block of memory. Some functions like `clenqueueReadBuffer` expect a pointer to memory and the size of the block of memory so I end up doing is making a hard copy of the `tbb::concurrent_vector` into an `std::vector`. This is not optimal for performance. 

If we wanted to I could write something based on rigtorp's [`ring_buffer`](https://rigtorp.se/ringbuffer.cpp) but for just a vector of data. The only odd part of it would be when we have to resize the vector we would have to make a hard copy of the events into a new slice of memory and keep the old memory around until either the buffer falls out of scope or the last event is done. I'm fine with doing that as well.

## Tests

I'm also not sure how to write a test for this? The current tests all pass (in fact I found a small bug in normal_lccdf where the results didn't match and fixed that).

## Side Effects

The bad thing to note here is the copy every time we need a contiguous block of memory for queuing the events for read/write ops. This is not great, though it will be fine for now if we keep it this way and then later I can write a faster scheme.

## Release notes

Replace the `std::vector`s  for read/write events in `matrix_cl` with `tbb::concurrent_vectors`

## Checklist

### DIFF
--- a/stan/math/opencl/copy.hpp
+++ b/stan/math/opencl/copy.hpp
@@ -98,9 +98,10 @@ inline auto from_matrix_cl(const T& src) {
     try {
       cl::Event copy_event;
       const cl::CommandQueue queue = opencl_context.queue();
+      std::vector<cl::Event> copy_write_events(src.write_events().begin(), src.write_events().end());
       queue.enqueueReadBuffer(src.buffer(), opencl_context.in_order(), 0,
                               sizeof(T_val) * dst.size(), dst.data(),
-                              &src.write_events(), &copy_event);
+                              &copy_write_events, &copy_event);
       copy_event.wait();
       src.clear_write_events();
     } catch (const cl::Error& e) {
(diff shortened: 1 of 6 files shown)


SteveBronder:
Another idea would be to just have a ring buffer with fixed capacity and before we push onto the queue we have a spin lock to wait till space is available

stan/math/opencl/copy.hpp
```diff
@@ -98,9 +98,11 @@ inline auto from_matrix_cl(const T& src) {
...
       cl::Event copy_event;
       const cl::CommandQueue queue = opencl_context.queue();
+      std::vector<cl::Event> copy_write_events(src.write_events().begin(),
+                                               src.write_events().end());
       queue.enqueueReadBuffer(src.buffer(), opencl_context.in_order(), 0,
                               sizeof(T_val) * dst.size(), dst.data(),
-                              &src.write_events(), &copy_event);
+                              &copy_write_events, &copy_event);
       copy_event.wait();
       src.clear_write_events();
```
SteveBronder, line 107: Note the local copy here is safe because we wait for the copy event and then clear out the write events right after

SteveBronder:
@rok-cesnovar sorry to bother but could you look at this?

rok-cesnovar:
Looking into this over the weekend.

rok-cesnovar (approved):
I think everything here makes sense, so feel free to merge. Thanks!

PR #2914, opened 2023-06-16
Add rev functor to compute hessian-times-vector by finite differences

### LINKED ISSUE #2913: Hessian-vector product with central finite differences
## Description

We are implementing a Hessian-vector product for BridgeStan (https://github.com/roualdes/bridgestan/pull/126) and would like to be able to evaluate with central finite differences for cases where forward-mode is not defined (mainly implicit functions like the ode solvers).

There is also a scheme for setting relative perturbations that @justindomke mentioned in the issue above that we should try.  Code sketches and reference links are in the above-referenced BridgeStan issue.

## Expected Output

This should be a functor defined parallel to where Hessians are defined via finite differences.  

#### Current Version:
v4.6.2

WardBrian:
## Summary

This adds a function analogous to `finite_diff_hessian_auto` for computing Hessian-vector products of functions which only support `rev` mode types. This can be done with only two calls to `gradient`, compared to the `2N` calls needed for the naive approach of using `finite_diff_hessian_auto` and then multiplying.

The approach is described by @justindomke here: https://justindomke.wordpress.com/2009/01/17/hessian-vector-products/

In short, 

If $g(x) =  \frac{\partial f}{\partial x} f(x)$, then $g(x+\Delta x) \approx g(x) + H(x)\Delta x$
So
$H(x)v \approx \frac{g(x + rv) - g(x - rv)}{2r}$

See also: discussion in roualdes/bridgestan#126
## Tests

I have duplicated the existing test used for autodiff Hessian-vector products. There is only one, so more might be nice.

## Side Effects

None

## Release notes

New function in `stan::math::internal`, `finite_diff_hessian_vector_product_auto`

## Checklist

- [x] Math issue: closes #2913 


aseyboldt:
I don't think it is valid to choose a different epsilon for each dimension, wouldn't that just change the vector we multiply the hessian with?

WardBrian:
> I don't think it is valid to choose a different epsilon for each dimension, wouldn't that just change the vector we multiply the hessian with?

I think you're right - this just so happened to have worked because the `finite_diff_stepsize` function uses 
`stepsize(u) = cbrt(epsilon) * max(1, abs(u))`, and because v is normalized `max(1,u)` is one always, so the cube root of epsilon is used. 

Do you think there is a better way of automatically deciding an epsilon, or should I just skip the call to `finite_diff_stepsize` and use cbrt(epsilon)?

aseyboldt:
> Do you think there is a better way of automatically deciding an epsilon, or should I just skip the call to finite_diff_stepsize and use cbrt(epsilon)?

I don't know of any better way, but that doesn't mean there isn't one... :-)

stan/math/rev/functor/finite_diff_hessian_times_vector_auto.hpp
```diff
@@ -0,0 +1,70 @@
...
+ */
+template <typename F>
+void finite_diff_hessian_times_vector_auto(const F& f, const Eigen::VectorXd& x,
+                                           const Eigen::VectorXd& v, double& fx,
+                                           Eigen::VectorXd& hvp) {
+  fx = f(x);
+
+  double v_norm = v.norm();
+
+  double epsilon = std::sqrt(EPSILON) * (1 + x.norm()) / v_norm;
```
andrjohns, line 51:
```suggestion
  double epsilon = std::sqrt(EPSILON) * (1 + x.norm()) / v.norm();
```

stan/math/rev/functor/finite_diff_hessian_times_vector_auto.hpp
```diff
@@ -0,0 +1,70 @@
...
+ * @param[in] x Argument to function
+ * @param[in] v Vector to multiply Hessian with
+ * @param[out] fx Function applied to argument
+ * @param[out] hvp Product of Hessian and vector at argument
+ */
+template <typename F>
+void finite_diff_hessian_times_vector_auto(const F& f, const Eigen::VectorXd& x,
+                                           const Eigen::VectorXd& v, double& fx,
+                                           Eigen::VectorXd& hvp) {
+  fx = f(x);
```
andrjohns, line 47: This doesn't seem to be used anywhere, safe to remove?

andrjohns (changes requested):

andrjohns (dismissed):
LGTM!

WardBrian:
@andrjohns mind re-approving? The auto formatting commit dismissed your last review

andrjohns (approved):

PR #2919, opened 2023-06-28
makefile fix

mitzimorris:
## Summary

added flag `-Wno-unused-but-set-variable` to `make/libraries` variable `TBB_CXXFLAGS` for clang compiler.

## Side Effects

None

## Release notes

Cleanup compiler warnings for build of TBB dylib.

## Checklist

- [x] Math issue #(issue number) 2918

- [x] Copyright holder: (fill in copyright holder information). Columbia Universityr

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen  (NA)

### DIFF
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -105,7 +105,7 @@ endif
 O ?= 3
 
 ## setup includes
-INC ?= $(INC_FIRST) -I $(if $(MATH),$(MATH),.) -I $(EIGEN) -I $(BOOST) $(INC_SUNDIALS)
+INC ?= $(INC_FIRST) -I $(if $(MATH),$(MATH),.) -I $(EIGEN) -isystem $(BOOST) $(INC_SUNDIALS)
 INC_SUNDIALS ?= -I $(SUNDIALS)/include -I $(SUNDIALS)/src/sundials
 INC_GTEST ?= -I $(GTEST)/include -I $(GTEST)
 
@@ -252,12 +252,12 @@ endif
 ifdef TBB_LIB
 
 ifdef TBB_INC
-CXXFLAGS_TBB ?= -I $(TBB_INC)
+CXXFLAGS_TBB ?= -isystem $(TBB_INC)
 else
-CXXFLAGS_TBB ?= -I $(TBB)/include
+CXXFLAGS_TBB ?= -isystem $(TBB)/include
 endif
 
-LDFLAGS_TBB ?= -Wl,-L,"$(TBB_LIB)" -Wl,-rpath,"$(TBB_LIB)" -Wl,--disable-new-dtags -ltbb
+LDFLAGS_TBB ?= -Wno-unused-command-line-argument -Wl,-L,"$(TBB_LIB)" -Wl,-rpath,"$(TBB_LIB)" -Wl,--disable-new-dtags -ltbb
 LDLIBS_TBB ?= $(LDFLAGS_TBB)
 
 else
@@ -274,8 +274,8 @@ ifeq ($(OS),Linux)
   TBB_TARGETS ?= $(addprefix $(TBB_BIN)/lib,$(addsuffix $(LIBRARY_SUFFIX).2,$(TBB_LIBRARIES)))
 endif
 
-CXXFLAGS_TBB ?= -I $(TBB)/include
-LDFLAGS_TBB ?= -Wl,-L,"$(TBB_BIN_ABSOLUTE_PATH)" -Wl,-rpath,"$(TBB_BIN_ABSOLUTE_PATH)" $(LDFLAGS_FLTO_FLTO) $(LDFLAGS_OPTIM_TBB)
+CXXFLAGS_TBB ?= -isystem $(TBB)/include
+LDFLAGS_TBB ?= -Wno-unused-command-line-argument -Wl,-L,"$(TBB_BIN_ABSOLUTE_PATH)" -Wl,-rpath,"$(TBB_BIN_ABSOLUTE_PATH)" $(LDFLAGS_FLTO_FLTO) $(LDFLAGS_OPTIM_TBB)
 LDLIBS_TBB ?= $(LDFLAGS_TBB)
 
 endif
--- a/make/libraries
+++ b/make/libraries
@@ -122,7 +122,7 @@ TBB_CXX_TYPE ?= $(CXX_TYPE)
 # Set c compiler used for the TBB
 ifeq (clang,$(CXX_TYPE))
   TBB_CC ?= $(subst clang++,clang,$(CXX))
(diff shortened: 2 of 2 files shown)


syclik:
@mitzimorris, thanks!

(Unfortunately) This PR introduces a side effect and I want to make sure it's intended before we go with it. It will turn off that warning for everything, not just for building TBB. I'm guessing that's not the intent. We'd want to flag these warnings for the math library. 

I think we want to add the flag when we're building the dylib, as mentioned in the issue. Is that correct? If so, we can update the compiler flags for the target that we want only. 

There's one more alternate: use -isystem, but I do think that's not reliable cross platform/compiler. 

Mind letting me know if it's intended to only turn off the warnings for building specific targets?

mitzimorris:
you're correct - because of the way that the math library builds up the CXXFLAGS, this will get added.  my bad.
we should just add this to the target.  will redo.

also beyond our control is the ur-annoying warning triggered by the use of `sprintf` in some BOOST libraries.  because these are header-only libraries, the only way I see is to add `-Wno-deprecated-declarations` to compiler flags, but this warning shouldn't be ignored if it's one of ours.

syclik:
Thanks for the clarification! I'll take a look when you update... or feel free to point me to what you're doing to see the warnings (exact convocation) and I'll see if I can spot something that we can do.

mitzimorris:
to see the warnings, I'm doing this:

```
make clean-all; make lib/tbb/libtbb.dylib
```

mitzimorris:
ready for review.

make/libraries
```diff
@@ -166,7 +166,7 @@ endif
 $(TBB_BIN)/tbb.def: $(TBB_BIN)/tbb-make-check $(TBB_BIN)/tbbmalloc.def
 	@mkdir -p $(TBB_BIN)
 	touch $(TBB_BIN)/version_$(notdir $(TBB))
-	tbb_root="$(TBB_RELATIVE_PATH)" CXX="$(CXX)" CC="$(TBB_CC)" LDFLAGS='$(LDFLAGS_TBB)' '$(MAKE)' -C "$(TBB_BIN)" -r -f "$(TBB_ABSOLUTE_PATH)/build/Makefile.tbb" compiler=$(TBB_CXX_TYPE) cfg=release stdver=c++1y  CXXFLAGS="$(TBB_CXXFLAGS)"
+	tbb_root="$(TBB_RELATIVE_PATH)" CXX="$(CXX)" CC="$(TBB_CC)" LDFLAGS='$(LDFLAGS_TBB)' '$(MAKE)' -C "$(TBB_BIN)" -r -f "$(TBB_ABSOLUTE_PATH)/build/Makefile.tbb" compiler=$(TBB_CXX_TYPE) cfg=release stdver=c++1y  CXXFLAGS="$(TBB_CXXFLAGS) -Wno-unused-but-set-variable"
```
wds15, line 169: Wouldn't it be better to add this flag in line 125 of the file to the clang specific options set as default?

mitzimorris: much better - thanks!

syclik: @mitzimorris This isn't where I thought the change would be. I thought we would have applied it to just the dylib target that you mentioned. Do we have a separate dylib target?

I didn't realize the change is clang-specific. Does g++ error out with this option defined?

mitzimorris: @syclik - regarding the change - this defines the way that several targets, including libtbb.dylib, line 175, are invoked.

I'm not sure if the build warnings are specific to clang - I believe the `sprintf` warnings are, and that one is particularly vexing.  however, there's a better and more general solution:

instead of including 3rd party libs using "-I", use compiler directive "-isystem" - cf https://stackoverflow.com/a/2452392

changing line 108 of `make/compiler_flags` from 
```
INC ?= $(INC_FIRST) -I $(if $(MATH),$(MATH),.) -I $(EIGEN) -I $(BOOST) $(INC_SUNDIALS)
```
to 
```
INC ?= $(INC_FIRST) -I $(if $(MATH),$(MATH),.) -I $(EIGEN) -isystem $(BOOST) $(INC_SUNDIALS)
```

gets rid of all the `sprintf` spew.

mitzimorris:
most recent commit is a better-informed update of the makefiles.  CmdStan builds as is with no warnings.

please re-review.

make/compiler_flags
```diff
@@ -105,7 +105,7 @@ endif
 O ?= 3
 
 ## setup includes
-INC ?= $(INC_FIRST) -I $(if $(MATH),$(MATH),.) -I $(EIGEN) -I $(BOOST) $(INC_SUNDIALS)
+INC ?= $(INC_FIRST) -I $(if $(MATH),$(MATH),.) -I $(EIGEN) -isystem $(BOOST) $(INC_SUNDIALS)
```
syclik, line 108: hi @mitzimorris! 

Although this is what we'd really like to do, this is going to wreak havoc on people's builds. This was what I was alluding to earlier. @WardBrian and I talked about exactly this at StanCon too.

So the problem with this is that `-isystem` actually changes the include order that the compiler sees the libraries. I never found it to be clearly documented in the compiler docs themselves, but this allows the boost and tbb from the environment's path to be included before our version.

And unfortunately, this means that we'll not only have compile-time issues (which are easier) but we'll have runtime discrepancies in behavior (which is a lot harder). 

There's obviously a tradeoff here. 
* Do we have useless warning messages that annoy and confuse users?
* Do we have incorrect sampling and behavior that we can't reliably detect?

I've always leaned towards the first rather than the second.

syclik (changes requested):

WardBrian:
If the primary concern is end users I think adding `-w` during releases to silence warnings is a reasonable choice

syclik:
Yes, I believe that is the primary concern.

Can you clarify? Where would you put `-w`? (Does that really just silence all the boost and other warnings? If so, awesome!)

WardBrian:
> -w[ ¶](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-w)
Inhibit all warning messages

from https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

we’d want this in CXXFLAGS in releases I believe

syclik:
Thank you! 

That sounds wonderful. There isn't a reason the warnings would be useful, right? (I can't think of a single warning that conveys useful information for end users, but I might not be thinking hard enough right now.)

WardBrian:
It looks like clang and gcc accept both -w and -Wno-everything on every version we support

syclik:
@mitzimorris, I'm closing this PR for now.

I think the right move here is to either:
1. put `-w` into Math's default make flags, but not alter the include flags. Changing the include flags has unintended side-effects that are going to be hell to debug and instruct users on correct configuration
2. shift the burden to the interfaces. While that creates more work, it's definitely controllable there.

Either way, I don't think the implementation is the one proposed in this PR. Feel free to reopen if you're confident this is the right fix and we can have a longer discussion. (I'm happy to be convinced otherwise.)

PR #2928, opened 2023-08-02
Allow arena_matrix to use move semantics

SteveBronder:
## Summary

This PR does two things

1. Allows `arena_matrix` to use move semantics
  - If an `arena_matrix`' s constructor is passed an rvalue, instead of allocating memory onto the stack allocator and then doing a copy, `arena_matrix` now pushes that rvalue into a `chainable_ptr` that sits in the chainable stacks `var_alloc_stack_` and is deleted after `stan::math::recover_memory()` is called. This saves us the allocation and the copy, instead we just delay paying the deletion cost which we would have paid anyway when the temporary was deleted

2. After talking with @bob-carpenter I think I'd like to remove the hard copies we do for reverse mode functions using Eigen types. Right now our pattern is to make an arena matrix for the result, pass that to the reverse mode callback, and then when we return from the function we make a hard copy into an actual `Eigen::Matrix` type. The hard copy into the Eigen matrix is to make sure assignments done later to the matrix (i.e. `X(1, 1) = a`) do not overwrite the memory necessary for the reverse pass. 

Instead, I suggest we just do what Eigen does and include docs that tell people that `auto` is unsafe to use recklessly with the stan math library. Instead users who want to do assignment to elements of the matrices should have the assigned object's type be the appropriate Eigen matrix type. 

For example, in the below code, using `auto` for the assignment from the function and then doing an assignment to an element of the matrix would mess up the reverse mode pass of the function the matrix was generated in. However, assigning the result of the function to an actual matrix type and then doing the element assignment is safe.

```c++
Eigen::Matrix<var, -1, 1> y;
Eigen::Matrix<var, -1, -1> X;
// Bad!! Will change memory used by reverse pass callback within multiply!
auto mu = multiply(X, y);
mu(4) = 1.0;
// Good! Will not change memory used by reverse pass callback within multiply
Eigen::Matrix<var, -1, 1> mu_good = multiply(X, y);
mu_good(4) = 1.0;
```

This is a breaking change so I think we should do a major version bump. In the tests below I've found anywhere from a 5% to 15% speed increase from using this.

The graph below plots the relative % improvement of using move semantics relative to the current develop. We are testing the expression below, which just does a bunch of multiplies, adds, and then a sum.

```c++
var lp = sum(stan::math::add(stan::math::multiply(stan::math::multiply(x, y), std::move(y)), std::move(x)));
```

The variables `x` and `y` in the above are matrices of the same size. Only the forward pass of reverse mode autodiff is used in the benchmark since this should not have any effect on the reverse passes performance.

compare_plot

We can see that for small sizes this is great. As we get to larger sizes it's still good, but at that point a lot of the actual computation and fetching data for the operations is taking up enough time that the memory allocation is not a huge concern.

You can run this performance benchmark yourself using the following

```bash
git clone https://github.com/SteveBronder/stan-perf
cd stan-perf
git checkout tmp-values
cmake -S . -B "build" -DCMAKE_BUILD_TYPE=Release
cd ./build
make -j4 move_ex move_ex_orig
# WARNING: Will use a lot of memory for the large problems and take a while
taskset -c 1 ./benchmarks/move_stuff/move_ex --benchmark_out_format=csv --benchmark_out=./benchmarks/move_stuff/move_ex.csv --benchmark_report_aggregates_only=false --benchmark_repetitions=30 --benchmark_report_aggregates_only=false --benchmark_display_aggregates_only=true --benchmark_enable_random_interleaving=true
taskset -c 1 ./benchmarks/move_stuff/move_ex_orig --benchmark_out_format=csv --benchmark_out=./benchmarks/move_stuff/move_ex_orig.csv  --benchmark_report_aggregates_only=false --benchmark_display_aggregates_only=true --benchmark_repetitions=30 --benchmark_enable_random_interleaving=true
Rscript ./plots/move_stuff/plots.R
```

## Tests

Tests are added to the `arena_matrix_test` file for checking moves work. This can be run with 

```bash
python ./runTests.py ./test/unit/math/rev/core/arena_matrix_test
```

## Side Effects

Yes

In order to utilize this we need to use [perfect forwarding](https://en.cppreference.com/w/cpp/language/reference#Forwarding_references) in our reverse mode functions. See the new `multiply` and `add` signatures in this PR for an example. I think I could do this for most of our functions that can use it with a day or two of work, though now our functions will be more "modern" but also harder to understand.

## Release notes

Allows `arena_matrix` to use move semantics along with the stack allocator.

## Checklist


SteveBronder:
@t4c1 sorry but would you mind taking a quick glance at this?

@andrjohns would you feel comfortable reviewing this?

stan/math/prim/prob/normal_log.hpp
```diff
@@ -29,20 +29,18 @@ namespace math {
  * @tparam T_loc Type of location parameter.
  */
 template <bool propto, typename T_y, typename T_loc, typename T_scale>
-inline return_type_t<T_y, T_loc, T_scale> normal_log(const T_y& y,
-                                                     const T_loc& mu,
-                                                     const T_scale& sigma) {
-  return normal_lpdf<propto, T_y, T_loc, T_scale>(y, mu, sigma);
+inline return_type_t<T_y, T_loc, T_scale> normal_log(T_y&& y, T_loc&& mu,
+                                                     T_scale&& sigma) {
+  return normal_lpdf<propto>(y, mu, sigma);
```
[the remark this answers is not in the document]
SteveBronder: Yes thanks! The normal distribution changes are mostly just to show what the new signatures would look like. But I'd like to make them so that we can forward them into `operands_and_partials`.

SteveBronder:
@andrjohns would you mind taking a look at this?

syclik:
@SteveBronder, this is rad!

I'm concerned with the impact it has on our use of `auto` through the Math codebase. Will we need to go through the current uses of auto everywhere to be safe? Or does this really only affect downstream usage.

Just to be absolutely clear, this is a "concern" and it is not blocking in my mind. It seems like we should do this; I just want to be cognizant of the effort to get this properly done before embarking on it.

syclik:
> Tbc this should only effect downstream users of the math library. At the Stan language level we never use auto as a type for assignment so it won't effect Stan user's programs. Downstream users are also likely aware of the dangers of using auto with Eigen already so I don't think much code would break

That's great!


> Totally valid concern imo. I will look through out use of auto, but we already follow Eigen's rules for auto so it shouldn't be an issue.

I'd be very, very careful here. I've been looking at the codebase and I'm pretty sure we're lax about the use of `auto`. But that might be for non-Eigen uses. Either way, because we use `auto` we have to check carefully or else we'd run into real subtle bugs here.

SteveBronder:
> I'd be very, very careful here.

Yes I'm going to fine comb look this over before merge. We generally use auto for scalar types and times a function can return a `std::vector`, `Eigen::Matrix`, or `scalar` like what we often have in the lpdf functions. The only time we would have to worry about this in the case of the math library is if we did code like 

```c++
auto mat = math_function(other_mat);
mat(0,0) = 2;
```

since the mat we use in reverse pass of `math_function` would change. But idt we have this patten in many parts of the codebase

syclik (changes requested):
@SteveBronder, just marking as "request changes" so it doesn't get moved to merged immediately after a review.

I think even after we determine it's all good, we'll need to make sure we get the right messages out about how this can break. (If that's just bumping the major version, that's fine, but we still need to do at least that.)

doxygen/contributor_help_pages/common_pitfalls.md
```diff
@@ -190,6 +190,38 @@ The general rules to follow for passing values to a function are:
...
 3. In prim, if you are confident and working with larger types, use perfect forwarding to pass values that can be moved from. Otherwise simply pass values by `const&`.
 
+
+### Using auto is Dangerous With Eigen Matrix Functions in Reverse Mode
+
+Like in [Eigen](https://eigen.tuxfamily.org/dox/TopicPitfalls.html), the use of `auto` with the Stan math library should be used with care. Along with the cautions mentioned in the Eigen docs, there are also memory considerations when using reverse mode automatic differentiation. When returning from a function in the Stan math library with an Eigen matrix output with a scalar `var` type, the actual returned type will often be an `arena_matrix<Eigen::Matrix<...>>`. The `arena_matrix` class is an Eigen matrix where the underlying array of memory is located in Stan's memory arena. The `arena_matrix` that is returned by Stan functions is normally the same one resting in the callback used to calculate gradients in the reverse pass. Directly changing the elements of this matrix would also change the memory the reverse pass callback sees which would result in incorrect calculations.
```
syclik, line 196: Let's reverse the way the first sentence is written. Instead of "Like in Eigen..." I think it'd be better if we got to the punchline first: "The use of `auto` with the Stan Math library should be used with care, like in Eigen..."

andrjohns:
@SteveBronder did you still want me to take a look at this, or is it all covered by @syclik and @t4c1?

syclik:
@andrjohns, please look too. More eyes are better on this one. It introduces some difficult side effects and has the potential to break existing code in subtle ways (and introduce subtle problems in new code).

SteveBronder:
@andrjohns if you have some time to look at this it would be great! But also imo I think this would be a major version bump so we should not rush it into the next version

andrjohns:
> @andrjohns if you have some time to look at this it would be great! But also imo I think this would be a major version bump so we should not rush it into the next version

Can do! Will add to the to-do list for next week

syclik:
After the current release, I think it's a good time to go with some breaking changes. 

I have a PR up for removing deprecated functions. That'll force a bump in version, so we might as well roll up additional breaking changes into the next release.

doxygen/contributor_help_pages/common_pitfalls.md
```diff
@@ -190,6 +190,37 @@ The general rules to follow for passing values to a function are:
...
+Eigen::Matrix<var, -1, -1> X;
+// Bad!! Will change memory used by reverse pass callback within multiply!
+auto mu = multiply(X, y);
+mu(4) = 1.0;
+// Good! Will not change memory used by reverse pass callback within multiply
+Eigen::Matrix<var, -1, 1> mu_good = multiply(X, y);
+mu_good(4) = 1.0;
+```
+
+The reason we do this is for cases where functions returns are passe to other functions. An `arena_matrix` will always make a shallow copy when being constructed from another `arena_matrix`, which let's the functions avoid unnecessary copies.
```
andrjohns, line 210:
```suggestion
The reason we do this is for cases where function returns are passed to other functions. An `arena_matrix` will always make a shallow copy when being constructed from another `arena_matrix`, which lets the functions avoid unnecessary copies.
```

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -73,6 +73,50 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
+   * @tparam T A type that inherits from Eigen::DenseBase that is not an
+   * `arena_matrix`.
+   * @param other expression
+   * @note When T is both an rvalue and a plain type, the expression is moved to
+   * the object stack. However when T is an lvalue, or an rvalue that is not a
+   * plain type, the expression is copied to the memory arena.
+   */
+  template <typename T, require_eigen_t<T>* = nullptr,
+            require_not_arena_matrix_t<T>* = nullptr>
+  arena_matrix(T&& other)  // NOLINT
```
andrjohns, line 88: Would it be possible to split this to two constructors using `require_t<std::is_rvalue_reference<T>::value>` rather than having the constructor instantiate and call a lambda? Just feels a little unnecessarily complex

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -110,23 +154,32 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
-      new (this) Base(
-          ChainableStack::instance_->memalloc_.alloc_array<Scalar>(a.size()),
-          a.cols(), a.rows());
-
+  template <typename T, require_not_arena_matrix_t<T>* = nullptr>
+  arena_matrix& operator=(T&& a) {
+    using T_t = std::decay_t<T>;
+    if (std::is_rvalue_reference<T&&>::value && is_plain_type<T_t>::value) {
+      // Note: plain_type_t here does nothing since T_t is plain type
+      auto other = make_chainable_ptr(plain_type_t<MatrixType>(std::move(a)));
```
andrjohns, line 162: If the `plain_type_t` is needed for some reason even though it does nothing, can you change the comment to explain why? Otherwise better to just remove the `plain_type_t`

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -110,23 +154,32 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
-    if ((RowsAtCompileTime == 1 && T::ColsAtCompileTime == 1)
-        || (ColsAtCompileTime == 1 && T::RowsAtCompileTime == 1)) {
-      // placement new changes what data map points to - there is no allocation
-      new (this) Base(
-          ChainableStack::instance_->memalloc_.alloc_array<Scalar>(a.size()),
-          a.cols(), a.rows());
-
+  template <typename T, require_not_arena_matrix_t<T>* = nullptr>
+  arena_matrix& operator=(T&& a) {
+    using T_t = std::decay_t<T>;
```
andrjohns, line 159: Should this be `plain_type_t` instead of just decaying? I don't think all expressions will have a `ColsAtCompileTime`/`RowsAtCompileTime` that can be called

andrjohns (changes requested):
Overall looks good, just a couple of a minor changes.

To clarify for `rev` functions moving forward, they should now:

  - Take inputs via forwarding refs instead of const refs
  - Forward inputs for any `arena_t`/`ref_type_t` construction
  - Use plain types for any temporaries/variables
  - Return `arena_t` types rather than plain types

Can you update the `contributor` doc to specify/outline this recommended 'style'?

Is there any way for the tests to check that when a function is added, that there isn't any inappropriate use of `auto`? Or is that not necessary since the `mix` tests would fail as incorrect values/memory is used?

stan/math/rev/core/arena_matrix.hpp
SteveBronder: Yes that's much cleaner. Though I still kept the immediately evaluated lambda as that keeps the constructor empty which I find kind of nice

SteveBronder:
Thanks for looking this over @andrjohns! I updated the documentation to make it more clear how and why we should use perfect forwarding in rev functions. I think it's good to explain, but idt we need to make it a core requirement that all functions now use perfect forwarding. This is a nice feature, but having signatures use `const&` types won't break anything and we can always help new developers fix that in PRs

SteveBronder:
@syclik @andrjohns can you give this one more look and then we will merge it to the 5.0 branch?

syclik:
Yes, I'll give it a good read-through again. Just need a little bit of time to get in the right mindset to get into deep C++ perfect forwarding. I'll try to set aside the hours necessary this weekend.

syclik:
@SteveBronder, I'm going to edit the PR to have the base branch be `5.0-breaking-changes` instead of `develop`.

SteveBronder:
Thanks! I'll update the merge conflicts tmrw

SteveBronder:
@syclik @andrjohns fixed the merge conflict, all good for me to merge to 5.0?

andrjohns:
> @syclik @andrjohns fixed the merge conflict, all good for me to merge to 5.0?

Yep, all good from me

syclik:
I’m still reading it. I was on it and it was taking a bit more time to
really get it done.

I’ll make sure to read through it carefully.

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -49,25 +49,91 @@ class arena_matrix<MatrixType, require_eigen_dense_base_t<MatrixType>>
           ChainableStack::instance_->memalloc_.alloc_array<Scalar>(size),
           size) {}
 
+ private:
+  template <typename T>
+  constexpr auto get_rows(const T& x) {
```
syclik, line 54: These two functions are a bit hard to understand. Could you add doxygen doc to explain what the intent is here? It's not coming through directly through code.

syclik: The two function definitions are identical. Is that intentional?

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -280,7 +330,7 @@ class arena_matrix<MatrixType, require_eigen_sparse_base_t<MatrixType>>
   template <typename Expr,
             require_not_same_t<Expr, arena_matrix<MatrixType>>* = nullptr>
   arena_matrix& operator=(Expr&& expr) {
-    *this = arena_matrix(std::forward<Expr>(expr));
+    new (this) arena_matrix(std::forward<Expr>(expr));
```
syclik, line 333: Mind walking me through this? I think I know what it's doing, but I want to double check. 

(Nothing too long; even just a couple of sentences.)

stan/math/rev/fun/multiply.hpp
```diff
@@ -129,49 +129,49 @@ inline var multiply(const T1& A, const T2& B) {
  * @tparam T1 type of the scalar
  * @tparam T2 type of the matrix or expression
  *
- * @param A scalar
+ * @param a scalar
```
syclik, line 132: Awesome! Thanks!

syclik (changes requested):

syclik:
Minor comments on the doc. 

I'm more curious about some of the code changes. 

Is there any place where we might have made a mistake by using auto internally? Do we need to sweep through the code base searching for that based on the changes here, or are we ok?

stan/math/prim/meta/is_arena_matrix.hpp
```diff
@@ -23,5 +23,17 @@ template <typename T>
...
+/*! \ingroup require_eigen_types */
+/*! \defgroup arena_matrix_types arena_matrix  */
+/*! \addtogroup arena_matrix_types */
+/*! @{ */
+
+/*! \brief Require type does not satisfy @ref is_arena_matrix */
+/*! @tparam T the type to check */
+template <typename T>
+using require_not_arena_matrix_t
+    = require_t<bool_constant<!is_arena_matrix<std::decay_t<T>>::value>>;
```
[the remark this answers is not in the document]
SteveBronder: `require_t` wants back a struct with a bool `value` member so we wrap the `!struct<T>::value` in bool constant for that.

We should really just make a `negate<>` type trait so the above can become 

```cpp
require_t<negate<is_arena_matrix<std::decay_t<T>>>>
```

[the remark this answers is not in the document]
SteveBronder: But I'd like to do that in a separate PR

stan/math/rev/core/arena_matrix.hpp
SteveBronder: > The two function definitions are identical. Is that intentional?

One gets the rows in case the conditional is true and the other gets the cols if the conditional is true. I might compress this into one function

SteveBronder:
> Is there any place where we might have made a mistake by using auto internally? Do we need to sweep through the code base searching for that based on the changes here, or are we ok?

I ctrl+f'd for `auto` in the library and it all seemed fine. Since we are already careful with using it with Eigen types we have been pretty careful on its usage

SteveBronder:
@syclik and @andrjohns anymore Qs? Else I'm going to merge this to the 5.0 branch

syclik:
No, no more questions!! Please merge when you get a chance!!!

Thank you!

PR #2948, opened 2023-09-21
Add makeflag for building TBB under UCRT automatically

andrjohns:
# Summary

Windows users (and cmdstan interfaces) currently need to add `-D_UCRT` to `make/local` when building on windows with a UCRT toolchain. This PR updates the makefiles to detect whether a UCRT toolchain is being used on Windows, and to add the flag automatically if needed.

We can detect UCRT-ness by checking whether the `UCRT` macro is defined after pre-processing the `windows.h` header

UCRT Toolchain:
```
PS C:\> echo "#include <windows.h>"| C:/rtools43/ucrt64/bin/g++ -E -dM -  | grep -i _UCRT
#define _UCRT
```

Non-UCRT Toolchain:
```
PS C:\> echo "#include <windows.h>"| C:/rtools43/mingw64/bin/g++ -E -dM -  | grep -i _UCRT
PS C:\>
```

I've tested that the TBB (and Stan) successfully build under both the UCRT `mingw-w64-ucrt-x86_64-gcc` and the non-UCRT `mingw-w64-x86_64-gcc` toolchains

## Tests

N/A

## Side Effects

N/A

## Release notes

Automatically detect UCRT toolchain use on Windows

## Checklist

### DIFF
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -121,6 +121,15 @@ CXXFLAGS_LANG ?= -std=c++1y
 CXXFLAGS_SUNDIALS ?= -pipe $(CXXFLAGS_OPTIM_SUNDIALS) $(CPPFLAGS_FLTO_SUNDIALS)
 #CXXFLAGS_GTEST
 
+make/ucrt:
+  pound := \#
+  UCRT_STRING := $(shell echo '$(pound)include <windows.h>' | $(CXX) -E -dM -  | findstr _UCRT)
+  ifneq (,$(UCRT_STRING))
+    IS_UCRT ?= true
+  else
+    IS_UCRT ?= false
+  endif
+  $(shell echo "IS_UCRT ?= $(IS_UCRT)" > $(MATH)make/ucrt)
 
 ## These are the other compiler flags that can be set.
 ##
@@ -161,18 +170,7 @@ ifeq ($(OS),Windows_NT)
     CXXFLAGS_OS ?= -m64
   endif
 
-  ifeq (,$(wildcard $(MATH)make/ucrt))
-    UCRT_STRING := $(shell echo "#include <windows.h>" | $(CXX) -E -dM -  | findstr _UCRT)
-    ifneq (,$(UCRT_STRING))
-      IS_UCRT ?= true
-    else
-      IS_UCRT ?= false
-    endif
-    $(shell echo "IS_UCRT ?= $(IS_UCRT)" > $(MATH)make/ucrt)
-  else
-    -include $(MATH)make/ucrt
-  endif
-
+  include make/ucrt
   ifeq ($(IS_UCRT),true)
     CXXFLAGS_OS += -D_UCRT
   endif
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -121,15 +121,6 @@ CXXFLAGS_LANG ?= -std=c++1y
 CXXFLAGS_SUNDIALS ?= -pipe $(CXXFLAGS_OPTIM_SUNDIALS) $(CPPFLAGS_FLTO_SUNDIALS)
 #CXXFLAGS_GTEST
 
-make/ucrt:
-  pound := \#
-  UCRT_STRING := $(shell echo '$(pound)include <windows.h>' | $(CXX) -E -dM -  | findstr _UCRT)
-  ifneq (,$(UCRT_STRING))
-    IS_UCRT ?= true
-  else
-    IS_UCRT ?= false
-  endif
-  $(shell echo "IS_UCRT ?= $(IS_UCRT)" > $(MATH)make/ucrt)
 
 ## These are the other compiler flags that can be set.
 ##
@@ -169,6 +160,16 @@ ifeq ($(OS),Windows_NT)
   else
     CXXFLAGS_OS ?= -m64
   endif
+
+  make/ucrt:
+    pound := \#
+    UCRT_STRING := $(shell echo '$(pound)include <windows.h>' | $(CXX) -E -dM -  | findstr _UCRT)
+    ifneq (,$(UCRT_STRING))
+      IS_UCRT ?= true
+    else
+      IS_UCRT ?= false
+    endif
+    $(shell echo "IS_UCRT ?= $(IS_UCRT)" > $(MATH)make/ucrt)
 
   include make/ucrt
   ifeq ($(IS_UCRT),true)
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -161,8 +161,19 @@ ifeq ($(OS),Windows_NT)
     CXXFLAGS_OS ?= -m64
   endif
 
-  UCRT_STRING := $(shell echo "\#include <windows.h>" | $(CXX) -E -dM -  | grep -i _UCRT)
-  ifneq (,$(UCRT_STRING))
+  ifeq (,$(wildcard $(MATH)make/ucrt))
+    UCRT_STRING := $(shell echo "#include <windows.h>" | $(CXX) -E -dM -  | findstr _UCRT)
+    ifneq (,$(UCRT_STRING))
+      IS_UCRT ?= true
+    else
+      IS_UCRT ?= false
+    endif
+    $(shell echo "IS_UCRT ?= $(IS_UCRT)" > $(MATH)make/ucrt)
+  else
+    -include $(MATH)make/ucrt
+  endif
+
+  ifeq ($(IS_UCRT),true)
     CXXFLAGS_OS += -D_UCRT
   endif
 
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -167,11 +167,6 @@ ifeq ($(OS),Windows_NT)
     LDLIBS_OS ?= -static-libgcc -static-libstdc++
   endif
 
-  UCRT_STRING := $(shell echo "\#include <windows.h>" | $(CXX) -E -dM -  | grep -i _UCRT)
-  ifneq (,$(UCRT_STRING))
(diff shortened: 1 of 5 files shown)


WardBrian:
Isn't windows.h famously huge? How long does that command take to run?

Do we currently use `grep` on Windows?

andrjohns:
Sigh, the "works on my machine" strikes again. Will debug and update

SteveBronder (approved):
This looks good but imo it may be nice to only run that grep command if we are using windows. About how long does it take? I don't have a windows computer so I can't test it locally. Also can you use the `-m 1` flag with grep so that it just stops on the first match?

WardBrian:
I am opposed to calling grep on Windows at all - I'm not sure if it is bundled in rtools, but the standard mingw compiler distribution wouldn't have it by default AFAIK

Compiling windows.h and grepping on every single invocation also seems like a bad idea, so if we could have something that persists this to disk somehow that would be great

SteveBronder:
> I am opposed to calling grep on Windows at all - I'm not sure if it is bundled in rtools, but the standard mingw compiler distribution wouldn't have it by default AFAIK

Is it `findstr` on windows?

> Compiling windows.h and grepping on every single invocation also seems like a bad idea, so if we could have something that persists this to disk somehow that would be great

Let's wait to see how long it takes from @andrjohns. Finding the first match of `_UCRT` is different then grepping all of `windows.h`

WardBrian:
This is grepping on the result of the compiler pre-processing windows.h, not just grepping the file itself

andrjohns:
>  > I am opposed to calling grep on Windows at all - I'm not sure if it is bundled in rtools, but the standard mingw compiler distribution wouldn't have it by default AFAIK

> Is it findstr on windows?

Good call! `findstr` seems to work just the same

> Compiling windows.h and grepping on every single invocation also seems like a bad idea

It takes about a second for me, and this is running on an Windows ARM VM on an M1 Mac, with `g++` running through the windows arm-to-x64 translation-layer:

```
PS C:\Users\andrew\Desktop\math> Measure-Command { echo "#include <windows.h>" | g++ -E -dM -  | findstr -i _UCRT }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 1
Milliseconds      : 300
Ticks             : 13004542
TotalDays         : 1.50515532407407E-05
TotalHours        : 0.000361237277777778
TotalMinutes      : 0.0216742366666667
TotalSeconds      : 1.3004542
TotalMilliseconds : 1300.4542
```

andrjohns:
And it looks like the Powershell `Select-String` function is slightly faster:
```
PS C:\Users\andrew\Desktop\math> Measure-Command { echo "#include <windows.h>" | g++ -E -dM -  | Select-String _UCRT }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 838
Ticks             : 8387663
TotalDays         : 9.70794328703704E-06
TotalHours        : 0.000232990638888889
TotalMinutes      : 0.0139794383333333
TotalSeconds      : 0.8387663
TotalMilliseconds : 838.7663
```

WardBrian:
I think sticking to cmd.exe programs is probably the best for compatibility. Findstr might also be faster if you use /l to disable regex searching

andrjohns:
> I think sticking to cmd.exe programs is probably the best for compatibility. Findstr might also be faster if you use /l to disable regex searching

Not much benefit unfortunately:
```
PS C:\Users\andrew\Desktop\math> Measure-Command { echo "#include <windows.h>" | g++ -E -dM -  | findstr /l _UCRT }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 1
Milliseconds      : 241
Ticks             : 12413473
TotalDays         : 1.43674456018519E-05
TotalHours        : 0.000344818694444444
TotalMinutes      : 0.0206891216666667
TotalSeconds      : 1.2413473
TotalMilliseconds : 1241.3473
```

andrjohns:
I just did a GHA actions run using the `windows-latest` runner to check the timing, and it's ~750ms for native windows: https://github.com/stan-dev/math/actions/runs/6318530408/job/17157709773

andrjohns:
Oh wait hold on, that's 14 seconds and 750 ms. Sheesh never mind

andrjohns:
Back to the drawing board!

WardBrian:
I think it would be fine if this was saved to a file somewhere (like make/os?) as long as it was deleted by “make clean”. 14s isn’t a killer if it only happens during “make build”

andrjohns:
We might be able to skip that completely, I started chasing through the includes in `windows.h` to see how/where `_UCRT` would get defined, and it looks like MinGW uses this macro:
```
#if !defined(_UCRT) && ((__MSVCRT_VERSION__ >= 0x1400) || (__MSVCRT_VERSION__ >= 0xE00 && __MSVCRT_VERSION__ < 0x1000))
/* Allow both 0x1400 and 0xE00 to identify UCRT */
#define _UCRT
#endif
```

make/compiler_flags
```diff
@@ -161,6 +161,22 @@ ifeq ($(OS),Windows_NT)
...
+    UCRT_STRING := $(shell echo "#include <windows.h>" | $(CXX) -E -dM -  | findstr _UCRT)
+    ifneq (,$(UCRT_STRING))
+      IS_UCRT ?= true
+    else
+      IS_UCRT ?= false
+    endif
+    $(shell echo "IS_UCRT ?= $(IS_UCRT)" > $(MATH)make/ucrt)
+  else
+    -include $(MATH)make/ucrt
+  endif
```
WardBrian, line 174: I think you can make this more concise by defining a make rule who's result is `make/ucrt`, which will then be built when you `include` it. This is how the `.d` files for dependencies work

andrjohns:
>We might be able to skip that completely, I started chasing through the includes in windows.h to see how/where _UCRT would get defined, and it looks like MinGW uses this macro

Unfortunately we can't use that Makefile logic since MinGW has their own logic for setting MSVCRT_VERSION, which I'd rather not hardcode. I've instead gone back to the `windows.h` approach but with caching

make/compiler_flags
andrjohns: Oh cool, TIL

WardBrian:
This might be a silly question but does `windows.h` always exist (e.g., on a system without MSVC etc installed)

andrjohns:
> This might be a silly question but does `windows.h` always exist (e.g., on a system without MSVC etc installed)

It will also exist if MinGW is installed (they provide a version as well). But if the system needs UCRT then I'd say it's pretty rare they wouldn't have `windows.h`

make/compiler_flags
andrjohns: Tried out the make rule, but I'm not sure about the best place to put the rule. Thoughts/prefs?

WardBrian: I think putting it just above where it is included is probably fine? I think you can put rules in places that are conditionally defined

andrjohns:
Looks like all passes (hooray!). Should we ask another dev/user with a windows machine to test the basic setup locally:
```
mingw32-make clean-all
mingw32-make -f make/standalone math-libs -j2
```

Just to make sure the initial caching process isn't too obnoxious for the dev workflow (or initial cmdstan build)

make/tests
```diff
@@ -106,7 +106,7 @@ else
 endif
 
 %.hpp-test : %.hpp test/dummy.cpp
-	$(COMPILE.cpp) $(CXXFLAGS) -O0 -include $^ -o $(DEV_NULL) -Wunused-local-typedefs
+	$(COMPILE.cpp) $(CXXFLAGS) -O0 -include $^ -S -o $(DEV_NULL) -Wunused-local-typedefs
```
WardBrian, line 109: I'm not sure where this rule is used, is there a reason for this change in this PR?

makefile
```diff
@@ -125,6 +125,7 @@ clean-deps:
 	@$(RM) $(call findfiles,test,*.d.*)
 	@$(RM) $(call findfiles,lib,*.d.*)
 	@$(RM) $(call findfiles,stan,*.dSYM)
+	@$(RM) $(call findfiles,make,ucrt)
```
WardBrian, line 128: Just noting so we remember to do it: it seems like stan and cmdstan do _not_ call the clean target of their submodules, so we will need to manually add something which deletes this higher up.

WardBrian (commented):
Two small comments. I will try this branch on my Windows machine tomorrow before final approval.

Thanks @andrjohns

WardBrian:
It looks like the Windows Header Checks are failing in #2952 and #2950, so we can definitely conclude this PR is doing _something_ right by fixing them here

make/tests
andrjohns: It's a MinGW UCRT-specific bug, where the toolchain doesn't recognise the `nul` device for some reason: https://github.com/msys2/MINGW-packages/issues/10730

Otherwise it fails with:
```
Assembler messages:
Fatal error: can't create nul: Invalid argument
```

But I should have included it only for Win+UCRT, so I've fixed that now

andrjohns: I can't replicate the failure locally, which makes me think it's related to either:

 * rtools43 uses gcc 13.2 while GHA uses 12.2
 * rtools43 gcc built by MSYS2, GHA by Mingw-W64
 
 RTools43 g++:
```
g++.exe (Rev2, Built by MSYS2 project) 13.2.0
```

GHA g++:
```
g++.exe (x86_64-posix-seh-rev2, Built by MinGW-W64 project) 12.2.0
```

WardBrian (approved):
Tried with both rtools43 (UCRT) and rtools40 on this branch, both work. The caching makes it so the build time is essentially no different

Thanks @andrjohns

PR #2956, opened 2023-10-04
updating PR template: removing Math issue

syclik:
## Summary

The existing PR template asks about a Math issue number. Over the years GitHub has gotten better with linking issues and PRs which has lead to us not using the issue by default. I think it's a good time to change our template to match current practices. 

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -26,8 +26,6 @@ Replace this text with a short note on what will change if this pull request is 
 
 ## Checklist
 
-- [ ] Math issue #(issue number)
-
 - [ ] Copyright holder: (fill in copyright holder information)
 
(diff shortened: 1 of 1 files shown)


syclik:
any reviewers: feel free to say you think it's still necessary to keep around. 

I remember adding it in when it was hard to know which "issue" was either an issue or a PR. This was included so it was much easier to write release notes. We followed the closed issues within a release. I don't think that's the current procedure now.

WardBrian (approved):
Agreed.

While we are updating this, we could consider making all of the pre-populated instructions into html comments with `<-- /-->` such that they wouldn't be rendered if the submitter didn't delete them.

My understanding is that the copyright holder field is also not necessary since that same agreement is part of the github [terms of service](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license)

PR #2962, opened 2023-10-10
Add Hypergeometric 1F0 function and gradients

andrjohns:
## Summary

This PR exposes the `hypergeometric_1f0` function from Boost and adds gradients. The gradients for the 1F0 are much simpler than the other hypergeometric series and can be expressed analytically (no need for the `grad_pFq` function)

## Tests

`prim` tests for values and throwing behaviour added, `mix` tests for gradients added via the ad testing framework

## Side Effects

N/A

## Release notes

Added the Hypergeometric 1F0 function and gradients

## Checklist


stan/math/fwd/fun/hypergeometric_1F0.hpp
```diff
@@ -0,0 +1,46 @@
...
+ *    a\left(1-z\right)^{-a-1} \f$
+ *
+ * @tparam Ta Fvar or arithmetic type of 'a' argument
+ * @tparam Tz Fvar or arithmetic type of 'z' argument
+ * @param[in] a Scalar 'a' argument
+ * @param[in] z Scalar z argument
+ * @return Hypergeometric 1F0 function
+ */
+template <typename Ta, typename Tz, require_any_fvar_t<Ta, Tz>* = nullptr>
+auto hypergeometric_1f0(const Ta& a, const Tz& z) {
```
syclik, line 29: Please replace `auto` with a proper return type.

syclik: I think this is `return_type_t<Ta, Tz>`.

stan/math/fwd/fun/hypergeometric_1F0.hpp
```diff
@@ -0,0 +1,46 @@
...
+  using FvarT = return_type_t<Ta, Tz>;
+
+  auto a_val = value_of(a);
```
syclik, line 32: Can we use the proper type instead of `auto`?

stan/math/prim/fun/hypergeometric_1F0.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *    a\left(1-z\right)^{-a-1} \f$
+ *
+ * @tparam Ta Arithmetic type of 'a' argument
+ * @tparam Tz Arithmetic type of 'z' argument
+ * @param[in] a Scalar 'a' argument
+ * @param[in] z Scalar z argument
+ * @return Hypergeometric 1F0 function
+ */
+template <typename Ta, typename Tz, require_all_arithmetic_t<Ta, Tz>* = nullptr>
+auto hypergeometric_1f0(const Ta& a, const Tz& z) {
```
syclik, line 31: Please use proper return type.

syclik: Given the type assumptions being made, is `Ta` and `Tz` just `double` and `double` only? 

If I'm reading this properly, this isn't vectorized or anything, is it?

syclik (changes requested):
Requesting changes to the types for readability. Please don't overuse `auto`. It makes it hard to understand what's going on with the code.

andrjohns:
> Requesting changes to the types for readability. Please don't overuse auto. It makes it hard to understand what's going on with the code.

Thanks @syclik! Updated to more explicit typing for readability

stan/math/prim/fun/hypergeometric_1F0.hpp
andrjohns: This is to allow all combinations of `int`/`double` (as well as `size_t`,`float`, etc.)

syclik:
Thank you!!!

stan/math/prim/fun/hypergeometric_1F0.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *    a\left(1-z\right)^{-a-1} \f$
+ *
+ * @tparam Ta Arithmetic type of 'a' argument
+ * @tparam Tz Arithmetic type of 'z' argument
+ * @param[in] a Scalar 'a' argument
+ * @param[in] z Scalar z argument
+ * @return Hypergeometric 1F0 function
+ */
+template <typename Ta, typename Tz, require_all_arithmetic_t<Ta, Tz>* = nullptr>
+double hypergeometric_1f0(const Ta& a, const Tz& z) {
```
syclik, line 31: Is this the `(double, double) -> double` full template specification? (sorry... notation is tough: `Ta = double`, `Tz = double`)

If so, this implementation doesn't make sense as the function template for `hypergeomtric_1f0` and we should probably implement this as a normal function. What do you think?

Based on the other implementations, we're not actually taking the route of defining the function template in `prim` and writing template specialization in each of the folders. We're using SFINAE to knock out template matches instead. So I think that means we don't have a single function template any more.

Am I thinking about this correctly? Thoughts? 

(My concrete suggestion: change this function to not have any template parameters.)

andrjohns:
@syclik small ping to have a look at the review responses here

stan/math/prim/fun/hypergeometric_1F0.hpp
syclik: Since this is the base template, I'd suggest having the return type be `return_type<Ta, Tz>` instead of `double`. I'll add the code suggestion right now.

The reason for that is the base template should be a catch-all for everything.

syclik:
```suggestion
return_type<Ta, Tz> hypergeometric_1f0(const Ta& a, const Tz& z) {
```

syclik: Something feels slightly off: the base template has 3 template parameters, but the specializations have 4? Was that intended? (Did I miss the actual base template declaration somewhere?)

andrjohns: > Something feels slightly off: the base template has 3 template parameters, but the specializations have 4? Was that intended? (Did I miss the actual base template declaration somewhere?)

Sorry you've lost me, all of the implementations here only have two parameters?

andrjohns: Are you counting the `require_*` elements here? The number of those don't have to be consistent between the implementations

syclik (approved):
@andrjohns, thanks!! LGTM! Thanks for updating the code.

syclik:
Merging!

PR #2965, opened 2023-10-22
Expanding 5 macros `STAN_ADD_REQUIRE_*` directly into code.

syclik:
## Summary

The mixed use of macros to generate a full set of template metaprograms is harder to read than having the template metaprograms in the code.

This PR expands the macros to the minimal set of macros required by the Math and Stan libraries. 

## Detail

There are 5 macros. I'll link to the latest released version (so the links are permanent), but they can be found on current develop. They are all inside [stan/math/prim/meta/require_helpers.hpp](https://github.com/stan-dev/math/blob/v4.7.0/stan/math/prim/meta/require_helpers.hpp):
1. STAN_ADD_REQUIRE_UNARY : https://github.com/stan-dev/math/blob/v4.7.0/stan/math/prim/meta/require_helpers.hpp#L72
2. STAN_ADD_REQUIRE_UNARY_INNER: https://github.com/stan-dev/math/blob/v4.7.0/stan/math/prim/meta/require_helpers.hpp#L114
3. STAN_ADD_REQUIRE_BINARY: https://github.com/stan-dev/math/blob/v4.7.0/stan/math/prim/meta/require_helpers.hpp#L186
4. STAN_ADD_REQUIRE_BINARY_INNER: https://github.com/stan-dev/math/blob/v4.7.0/stan/math/prim/meta/require_helpers.hpp#L229
5. STAN_ADD_REQUIRE_CONTAINER: https://github.com/stan-dev/math/blob/v4.7.0/stan/math/prim/meta/require_helpers.hpp#L332

All these macros, when used for every type, expands into a lot of definitions. Practically we don't use most of them.

## Tests

None added.

## Side Effects

None.

## Release notes

Removes 5 macros and expands directly into code: STAN_ADD_REQUIRE_UNARY, STAN_ADD_REQUIRE_UNARY_INNER, STAN_ADD_REQUIRE_BINARY, STAN_ADD_REQUIRE_BINARY_INNER, STAN_ADD_REQUIRE_CONTAINER.

## Checklist


SteveBronder:
The way these were expanded deletes all the docs for requires from the stan math website 

https://mc-stan.org/math/group__rev__row__vector__types_ga04400fc27cdcef7938e1a79d14068a7b.html#ga04400fc27cdcef7938e1a79d14068a7b

SteveBronder:
imo if we want to go this route I'd first see how many requires we can delete. Is this all the requires folded out? adding ~5k lines is not the worst, but it is a lot of replicated code over and over and I worry about the maintenance of it.

syclik:
Yes, absolutely everything taken out of the macros! I’m on another branch
seeing what can be removed.

I completely agree, but I think the actual footprint is going to be much
less. Enough so that we can write real doc and add some tests for the
things we actually use.

I submitted this as a PR to kick off testing. Since it’s so much stuff, I
don’t trust “works on my machine.”

I’ll tag you once I knock out the unused requires. Let’s see how that
looks.

syclik:
@SteveBronder, I just walked through all 1108 `require_*` template metaprograms. Good news, we only use 201 of them. (I think some of those are used by the other definitions and that's it. We may be to reduce the footprint even more.)

syclik:
@SteveBronder, review please?

I went through and also included the documentation from the macros, so the doxygen should be repopulated.

WardBrian:
My concern is that the downside to defining manually exactly what we use today is it raises the barrier for future contributions. If I know the `require` I need is `require_all_not_arena_matrix` (random example), today it exists. After this PR, I would need to first write it myself, and these things are far from obvious. The macros also make it very difficult to _incorrectly_ implement something like `require_all_not_...`, since they let you focus on the simplest case.

This is partially offset to me if this has a significant impact on compile times - have you done any timing @syclik?

syclik:
@WardBrian, I hear you! But I think the example you're bringing up highlights the opposite: I think the change **lowers** the barrier for future contributions. Btw, I think both stances are not unilaterally correct or incorrect and they are opinions, so at the end we can have different views on it. I'll just try to motivate why I think it's actually going to lower the barrier to entry.

---- 

First, the (random) example you picked is actually not one of the expanded metaprograms (if the macro was called correctly)! This stuff is really tricky and not having the actual definitions sitting in code makes it very difficult to know what does and doesn't exist. (It should be `require_all_not_arena_matrix_t`, it was missing the `_t` at the end; I'm not mentioning this to be nitpicky, I'm just agreeing with you that this is really exacting work.) In the current code, the way you'd have to realize that is by looking in two different places: the `STAN_ADD_REQUIRE_*()` macro call and the macro itself. Just to look at an existing example, let's look at `require_all_not_var_t`. 

How do you know which `STAN_ADD_REQUIRE_*()` macro expands into this if you're coming into the codebase to contribute? It doesn't show up when you use tools like `git grep`:
```
$ git grep require_all_not_var_t
stan/math/opencl/prim/multiply.hpp:          require_all_not_var_t<T_a, T_b>* = nullptr>
stan/math/opencl/prim/multiply.hpp:          require_all_not_var_t<T_a, T_b>* = nullptr>
stan/math/prim/fun/add.hpp:          require_all_not_var_t<ScalarA, ScalarB>* = nullptr>
stan/math/prim/fun/dot_product.hpp:          require_all_not_var_t<Scalar1, Scalar2> * = nullptr>
stan/math/prim/fun/squared_distance.hpp:          require_all_not_var_t<Scal1, Scal2>* = nullptr>
stan/math/prim/fun/subtract.hpp:          require_all_not_var_t<ScalarA, ScalarB>* = nullptr>
test/unit/math/rev/meta/require_generics_test.cpp:  require_scal_checker<stan::require_all_not_var_t, var, var>::all_not();
```

For someone new, it'd be a lot of detective work to understand that it's here: https://github.com/stan-dev/math/blob/v4.8.0-rc1/stan/math/prim/meta/is_var.hpp#L16. Not impossible, but not straightforward. `STAN_ADD_REQUIRE_UNARY(var, is_var, require_stan_scalar_real);`

What does this expand to? You need to look at this: https://github.com/stan-dev/math/blob/v4.8.0-rc1/stan/math/prim/meta/require_helpers.hpp#L72. And I don't know about you, but I had to look up the doc in C++ preprocessor expansion to even comprehend this chunk of code (I'm pasting here for emphasis):
```
#define STAN_ADD_REQUIRE_UNARY(check_type, checker, doxygen_group)       \
  /*! \ingroup doxygen_group */                                          \
  /*! \defgroup check_type##_types check_type  */                        \
  /*! \addtogroup check_type##_types */                                  \
  /*! @{ */                                                              \
  /*! \brief Require type satisfies checker */                           \
  template <typename T>                                                  \
  using require_##check_type##_t = require_t<checker<std::decay_t<T>>>;  \
                                                                         \
  /*! \brief Require type does not satisfy checker */                    \
  template <typename T>                                                  \
  using require_not_##check_type##_t                                     \
      = require_not_t<checker<std::decay_t<T>>>;                         \
                                                                         \
  /*! \brief Require all of the types satisfy checker */                 \
  template <typename... Types>                                           \
  using require_all_##check_type##_t                                     \
      = require_all_t<checker<std::decay_t<Types>>...>;                  \
                                                                         \
  /*! \brief Require any of the types satisfy checker */                 \
  template <typename... Types>                                           \
  using require_any_##check_type##_t                                     \
      = require_any_t<checker<std::decay_t<Types>>...>;                  \
                                                                         \
  /*! \brief Require none of the types satisfy checker */                \
  template <typename... Types>                                           \
  using require_all_not_##check_type##_t                                 \
      = require_all_not_t<checker<std::decay_t<Types>>...>;              \
                                                                         \
  /*! \brief Require at least one of the types do not satisfy checker */ \
  template <typename... Types>                                           \
  using require_any_not_##check_type##_t                                 \
      = require_any_not_t<checker<std::decay_t<Types>>...>;              \
/*! @} */
```

Even if you did read C++ preprocessor macros fluently, you'd have to find the right definition inside the chunk above.

Please compare that to what's in the PR branch https://github.com/stan-dev/math/blob/feature/simplify-meta/stan/math/prim/meta/is_var.hpp#L37:

```
/*! \brief Require none of the types satisfy is_var */
template <typename... Types>
using require_all_not_var_t = require_all_not_t<is_var<std::decay_t<Types>>...>;
```

It also shows up in tools like `git grep`. I don't know about you, but for me it seems like it's a lot easier to understand what's going on and contribute when it's laid out this way.

Note: did you catch that the `std::decay_t` inside the definition is superfluous? I wouldn't have been able to catch that without the expansion.

-------------

> After this PR, I would need to first write it myself, and these things are far from obvious.

Yes, I recognize this is a downside. But here are a few counterpoints:

1. expanding the macros actually clarifies what's happening. Even for C++-fluent people, I think having to figure out how things expand through the C++ preprocessor macros makes the macros even more opaque. 
2. each of the expansions is used in one place or another. There are examples for how to expand the requires traits. And besides, we can always go and look at the macros in the older tagged version. (I honestly don't think they're all used correctly, but that's another issue.)
3. testing; expanding the macros allows for us to actually write tests. We could have written tests for these with the macros in place, but about 80% of the macro expansions aren't used at all! We're down to ~200 traits in use. We can actually write tests for those. I plan on adding them slowly, especially for the ones that are implemented incorrectly.

When I think about these traits, I think they should be as obvious as we can make them. I think the expansion makes it more obvious. Using your example of `require_all_not_arena_matrix_t`, here's how it would be implemented:

```
/*! \brief Require none of the types satisfy is_arena_matrix */
template <typename... Types>
using require_all_not_arena_matrix_t = require_all_not_t<is_arena_matrix<std::decay_t<Types>>...>;

```

I don't know about you, but that looks a lot more friendly than looking at the macro.

-----------------

> The macros also make it very difficult to incorrectly implement something like `require_all_not_...`, since they let you focus on the simplest case.

I have a different opinion here. I'm looking at the macro expansions and can now see that some of the `is_*` traits are implemented with and without decays. The macros all decay types.

To me, this seems inconsistent and it's indicating that our traits aren't implemented correctly. (They may have the correct behavior, but because we're just decaying multiple times when it's unnecessary. Or maybe it is.)

All I'm trying to say is that the macros themselves may be incorrect and it's hard to correct them the way it is. Especially without testing. So I think expanding them is actually going to improve our chances of implementing something new correctly. (But this last part is conjecture. At the very least, I'd know how to review the traits in a PR.)

----------

One last point: the subset of developers that will write new traits at this level is small. The subset of developers that use the traits is larger than those that create these traits. Almost everyone that develops in Math has to touch these traits these days. I think making it clearer what's going on is helpful to the larger group. My opinion is that it's easier to understand when the definitions of the template metaprograms are spelled out.

If a Math developer really needed a definition that didn't exist, one of us could write it. (@WardBrian, if I asked you to write `require_any_not_vt_arena_matrix`, would you be able to piece it together from the code on the branch? What if I asked you what the definition looks like in develop now? Which is easier?)

In the long run, it would help out the Math library if more developers could understand and contribute. So I think lowering the barrier to entry is helpful. (More code is bad, but templated code where the definitions don't exist directly seem harder to understand?)

WardBrian:
> How do you know which `STAN_ADD_REQUIRE_*()` macro expands into this if you're coming into the codebase to contribute? It doesn't show up when you use tools like `git grep`:

I think reading your reply I realized my different opinion is likely a result of a different workflow. I use a pretty highly configured VSCode instance for C++, so for me, I just type `require_` and wait for a suggestion list of every possible require template to show up. I can then fuzzy-find on that suggestion list, so my actual keyboard inputs are probably more like `require_ [wait] arena <arrow keys> <TAB>`.

To me, the hardest part about this was actually knowing what the names of things mean: what is a "value type" versus a "scalar type", etc. I even wrote a program to help remind me: https://gist.github.com/WardBrian/85ca3a4bfb3a23f55ef40604987c04eb

I basically never have needed to dig into where these are defined, I was making my selection purely based off the name and the online doc's general guidance. But, when I did, VSCode's "go-to definition" feature actually does manage to point you to the specific macro instantiation. 

I can see how a different workflow (say, `emacs` and `git grep`) would be much more painful with the current set up. And, VSCode would still work just as well on this branch, there are just fewer suggestions that will show up...

I ultimately agree that this is reasonable territory to 'agree to disagree' if needed. I certainly wouldn't lose any sleep if this was merged as-is, I just wanted to have this back and forth to make sure we considered both sides

---

My workflow above of course assumes correct implementations of those options I'm picking off my menu! So if they're not actually correct, that is obviously a stronger argument than anything. Testing these is a huge pain in my experience, since it's hard to tell exactly what the compiler selected. 

I'd be super interested in what tests you have in mind look like, and also if you know of any examples where the current branch yields differrent(/better) behavior than `master`

---

> @WardBrian, if I asked you to write `require_any_not_vt_arena_matrix`, would you be able to piece it together from the code on the branch?

I think I could, and I agree it would probably be easier than reverse-engineering the macros. The issue is, if I (personally) do need to write one of these, the chance is very strong that I would actually need it for work either in `stan-dev/stan` or the compiler. So it would be an extra Math PR for work really happening elsewhere. Not the end of the world, of course

---

On compilation speed, it looks like the expression tests may be faster on this branch than others? But most other things look similar, so it's hard to say

syclik:
@WardBrian, thanks for your comments! And recognizing that at least part of the difference in perceived cognitive load has to do with the tooling.

I am 100% with you on the scalar_type vs value_type. That was one of the first reasons I started digging into this. I wanted to know what one of the `_vt` was actually returning as a type and I couldn't figure it out by looking directly at the code. Even with the code expanded it's a complicated thing, but for me it's a lot harder to piece together what's going on with it inside the macros. 

If we look at this another way, with the PR we would have added one more dev that can understand these meta programs and is willing to help make them better. Hopefully this means other devs can help too. 

Worst case, it adds some code. (When it passes all tests, all functionality is identical pre and post.) If in a while we determine this is worse, we can bring all the macros back. It's not an irreversible change.

syclik:
> I'd be super interested in what tests you have in mind look like, and also if you know of any examples where the current branch yields differrent(/better) behavior than master

Yeah! I think we can just write unit tests for the meta programs.  We have some unit tests inside something like `test/unit/math/prim/is_var.hpp`, but only for `is_var`. 

We test `requires_var_t` using different types with an `std::enable_if` or something like that to have it return true type or false type. 


> if you know of any examples where the current branch yields differrent(/better) behavior than master

I don't know any off the top of my head, but I do know that I'd implement things differently. One example is `require_var_t`: 

```
template <typename T>
using require_var_t = require_t<is_var<std::decay_t<T>>>;
```

if you look at the template specialization for `is_var` for `var`, it looks like this:
```
template <typename T>
struct is_var<T,
              std::enable_if_t<internal::is_var_impl<std::decay_t<T>>::value>>
    : std::true_type {};
```

Notice there's a `std::decay_t<T>` inside there? So we've decayed the original type T twice.  An update could be:
```
template <typename T>
using require_var_t = require_t<is_var<T>>;
```

Simpler and easier to understand. There's nothing wrong with the original implementation, but for implementational correctness, you wouldn't let this go, right?
```
template <typename T>
using require_var_t = require_t<is_var<std::decay<std::decay<T>>>>;
```

It does the same thing, but I would review it and say it's not implemented correctly. (Just as I probably wouldn't take `double x = static_cast<double>(5.0);`)

Now the tricky part: why can't we just update the macro? That would break things. There are some `is_*` template metaprograms that don't decay. Are those implemented correctly? I don't know. I want to find out.t There aren't tests to back up the current behavior. We'd also have to figure out which ones are in use to test against. 

Anyway, when I see myself trying to figure out whether I can get to the bottom of whether we're implementing it correctly, I can't really figure out how to do it through the macros. (And that might just be me, but I don't think I'm unique in thinking it's hard to understand what the requires template metaprograms do without the definitions.)

WardBrian:
Performing multiple decays seems pretty expected when you're composing things like this, and should also be harmless. Same result, and no extra runtime cost since this is all compile time anyway.  Whether or not the `is_*` metaprograms that don't perform the decay themselves are correct is worth exploring, but seems orthogonal to the use of macros. 

I think asking a macro to generate exactly the code you'd write yourself is probably too high a bar, as long as it is ultimately equivalent. I would probably also raise my eyebrows at a static cast to the same type in a PR, but I could also imagine it making total sense in a macro body, depending.

syclik:
:)

I feel like we'll have different opinions here and that's ok.

At the end of the day, if you won't lose sleep over the change, I think it's worth a shot. If it turns out to be worse, I'll be the first to offer to revert.

WardBrian:
I can agree to that. 

We should also add to the page explaining these requires (or maybe a new page?) general guidelines for writing a new one: what to name it, general conventions, existing helpers you can piece together, etc.

syclik:
Thanks. This PR is ready (pending tests; I just merged `develop` post release).

This is the page I'll be updating in future PRs: https://github.com/stan-dev/math/blob/feature/simplify-meta/doxygen/contributor_help_pages/require_meta.md

That file was modified in this PR to remove the wording about the macros. It was light on doc to begin with. As I keep working with `stan/meta`, I'll definitely document and catalog the traits. (I need to stare at the definitions to really know how to document it in a consistent way.)

syclik:
@SteveBronder, thanks for the comments! 

> Just to write down the historical context...

Thanks! A lot of that resonates. I looked back through the history of the codebase to see how it was before the macros were introduced. And I completely understand the convenience of having the macros in that development cycle. And I'm with you here: "With that all being said, I think things are in a much more stable place where a PR like this makes sense." 

I'm sure as it was being developed, we had thoughts of using all the generated requires; or at least having them all were really convenient.

> My only main thing is If we are going to limit the scope down then I think we need to add docs to explain to folks the idioms better and how to add their own require. I think that should be fairly simple

Agreed. Right now, the doc is written more in line with this than the macros.

> What makes the documentation feel light to you? I tried being pretty verbose with how the system worked, how they were named, and showing how to search for the requires you need on the stan math docs. Including breaking up the requires into what I thought were meaningful categories for people to search through

Ah. Maybe I'm missing something. The main page I'm looking at is this: https://mc-stan.org/math/md_doxygen_2contributor__help__pages_2require__meta.html, specifically the "Adding a new requires" section.

In this doc, I don't see things like:
1. When should I use the different macros? Two of the macros are mentioned out of the five.
2. What does each of the macros expand out to? My best guess is that you're expecting the dev to read the macro at this point. In the section before, it explains what `STAN_ADD_REQUIRES` expands out to, but in that section it doesn't mention that macro at all.
3. Doc for something like [`scalar_type`](https://mc-stan.org/math/structstan_1_1scalar__type.html) make sense when you already know what this means. (For me, I need to get to the definitions to understand what scalar type and value types are still -- that's on me. But the doc isn't helpful for me understanding that.) It could help to just have examples in the doc (for the future).

Please let me know if I've missed some doc about the macros! I didn't see it in the code itself and I couldn't find much more in the doxygen folder. Maybe I was searching for the wrong keywords.

> Did you do this and it broke things? My guess is that the things breaking are the is_* that are not decaying. Which is also probably the reason we had the decay there in the first place.

No, I didn't. I'm still trying to limit the scope of this PR to a refactor as opposed to features. That said, what you said resonates with me.

It feels like the `requires_*_t` should pass the type without decaying and have the bool trait determine whether the condition is met based on how it deals with types. Right now if we wanted to have a bool trait make a decision on a non-decayed type, we can't do it using the macro. So it seems a little off, but I'm not trying to fix anything right now. That sort of change should come later where we review cases carefully.

syclik:
@SteveBronder, what's required in this PR to have it pass a review?

Separately, please do list any reservations! Happy to work through those and make this code base better.

WardBrian:
I am still interested in establishing if there exists a program which has different behavior on this branch than on the current state of `develop`. It was implied earlier that these exist and that when they do, this branch has the "better" behavior, but it's all been abstract so far. 

This could be as simple as a unit test added to this branch which, when cherry-picked into `develop`, fails

syclik:
> I am still interested in establishing if there exists a program which has different behavior on this branch than on the current state of develop.

Oh no! There's a misunderstanding.

There is no behavior change in this PR. All the template metaprograms that were added directly here was what the C++ preprocess would have expanded, with no modification!

Any potential modification would happen after this PR with individual unit tests to make sure we understand the difference in behaviors. That sort of change is not part of this PR.

WardBrian:
Ah, I did misunderstand. While I am generally in favor of smaller PRs, because I am sort of dubious on the merits of this change for its own sake, I would be interested in seeing at least some of the improved doc or improved behavior be realized before merging. Happy to be overruled by @SteveBronder if he disagrees

spinkney:
> In this doc, I don't see things like:
> 
> 1. When should I use the different macros? Two of the macros are mentioned out of the five.
> 2. What does each of the macros expand out to? My best guess is that you're expecting the dev to read the macro at this point. In the section before, it explains what `STAN_ADD_REQUIRES` expands out to, but in that section it doesn't mention that macro at all.
> 3. Doc for something like [`scalar_type`](https://mc-stan.org/math/structstan_1_1scalar__type.html) make sense when you already know what this means. (For me, I need to get to the definitions to understand what scalar type and value types are still -- that's on me. But the doc isn't helpful for me understanding that.) It could help to just have examples in the doc (for the future).

Hi @syclik, I agree that this lowers the barrier to entry. At least, it helps me to understand what's going on. Are these doc updates part of this PR or in a different PR. I support adding this additional information asap along with adding this pr.

syclik:
@SteveBronder, mind if we get on a call and talk through this PR?

SteveBronder (changes requested):
Looking over these, if we are going to expand out the macros then I'd like to make sure the docs are correct before we do this. If we don't do it now then I think it's going to be super hard to fix all of these after we expand them out. I went through and added what we need to put in the doc for the requires. 

For the longform docs we need to have a section on adding new requires since people will have to do that manually now. We need to be very clear on the convention and expected API used when writing these. I think it would be something like the below. Since you mentioned the docs were very unclear I'd appreciate if you modify or use the below as well as what is already there to expand on what was confusing to you

---------------------

The Stan library requires a strict API to ensure consistency for the `requires`. The below go over all of the possible API configurations a developer should use when writing a new `requires` 

For the API docs below, let `T` represent the type parameter we want to check, `is_type` is a generic type trait which will be replaced by the developer, and `InnerCheck` is a type trait used to check either the @ref value_type or @ref scalar_type of `T`.

Each requires ends in `_t`, `_vt`, or `_st`. They differ in the following ways

- `_t` uses `Check` to test the type T passed in

Ex:
```c++
// Always decay types coming into the requires
template <typename T>
require_autodiff_t = require_t<is_autodiff<std::decay_t<T>>>;
```

- `_vt` uses `Check` to test the type T passed in and uses `InnerCheck` to test the @ref value_type of `T`

```c++
template <template <class...> class TypeCheck, class... Check>   
require_std_vector_vt = require_vt<is_std_vector, TypeCheck, std::decay_t<Check>...>;

// Ex: Used to define a signature for `std::vectors` with a value type that is autodiffable
template <typename StdVec, require_std_vector_vt<is_var, StdVec>* = nullptr>
auto my_func(StdVec&& vec);
```

- `_st` uses `Check` to test the type T passed in and uses `InnerCheck` to test the @ref scalar_type of `T`

```c++
template <template <class...> class TypeCheck, class... Check>   
require_std_vector_st = require_st<is_std_vector, TypeCheck, std::decay_t<Check>...>;

// Ex: Used to define a signature for `std::vectors` with a scalar type that is autodiffable
template <typename StdVec, require_std_vector_st<is_var, StdVec>* = nullptr>
auto my_func(StdVec&& vec);
```

There variant of the requires that places the `vt` or `st` before the type trait name only checks the @ref value_type or @ref scalar_type of `T` without testing `T`.

```c++
// Require the scalar type is an std::vector
template <typename T>   
require_st_std_vector = require_t<is_std_vector<scalar_type_t<std::decay_t<T>>>>;

// Ex: Used to define a signature for `std::vectors` with a scalar type that is autodiffable
template <typename StdVec, require_std_vector_st<is_var, StdVec>* = nullptr>
auto my_func(StdVec&& vec);
```

In the below, {TYPE_TRAIT} represents the name of the trait the requires checks. Each new require *must* follow this standard API.

`require_{TYPE_TRAIT}_t`: The template parameter must return `true` when passed to the type trait

`require_not_{TYPE_TRAIT}_t`: The template parameter must return `false` when passed to the type trait

`require_all_{TYPE_TRAIT}_t`: The template parameters must all return `true` when passed to the type trait

`require_all_not_{TYPE_TRAIT}_t`: The template parameters must all return `false` when passed to the type trait

`require_any_{TYPE_TRAIT}_t`: At least one of the template parameters must return `true` when passed to the type trait

`require_any_not_{TYPE_TRAIT}_t`: At least one of the template parameters must return `false` when passed to the type trait

In addition to all the requires with an `_t` at the end, the requires also have `_st`, `_vt` variants where in addition to the logic above, the @ref value_type or @ref scalar_type must follow the same logic as the type for `T`.  The `_st_`, and `_vt_` variants must also follow the same logic but for checking only the inner @ref value_type or @scalar_type.

syclik:
Thanks, @SteveBronder. I'll work on that and ping you when it's all addressed. Thanks for leaving the detailed comments; I think they're clear enough to propagate those changes. I'll let you know if I run into any issues with it!

stan/math/prim/meta/require_helpers.hpp
```diff
@@ -322,103 +83,6 @@ using container_type_check_base
...
-  /*! \ingroup doxygen_group */                                                \
-  /*! \defgroup check_type##_types check_type  */                              \
-  /*! \addtogroup check_type##_types */                                        \
-  /*! @{ */                                                                    \
-  /*! \brief Require type satisfies checker */                                 \
-  /*! and value type satisfies `TypeCheck` */                                  \
-  /*! @tparam TypeCheck The type trait to check the value type against*/       \
-  template <template <class...> class TypeCheck, class... Check>               \
-  using require_##check_type##_vt = require_t<                                 \
-      container_type_check_base<checker, value_type_t, TypeCheck, Check...>>;  \
```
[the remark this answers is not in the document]
syclik: @SteveBronder, I tried to implement this (adding `require_vt` and `require_st`), but it's a little trickier than it looks on the surface due to the parameter pack expansion.

I think it's the right move; I'd like to change the code in a different PR. I'm not confident I'll get it right the first couple of times.

syclik:
@SteveBronder, I updated everything! 

* Please check out the updated `doxygen/contributor_help_pages/require_meta.md`. I could use your eyes on it:
    1. I kept the bulk of the original doc, but moved it towards the middle of the doc.
    2. I added an intro that I hope explains what we're trying to accomplish. At least at a high level and correct for most of the cases.
    3. I enumerated all the macro expands. I could really use your eyes there on the correct description of the `_vt` and `_st` traits with the logical qualifiers. I was using what was written in the code documentation to guide me, but I didn't dig enough to validate whether the doc was actually correct.

* I've gone through and added all the template parameter doc that was missing! Thanks for adding the doc in the PR comment. I used a version of that and it made it a whole lot easier.

* The comment about binary type traits -- we don't use any of them! Just FYI.

* Lastly, `require_vt`. I left a comment, but wanted to reiterate. I tried implementing both `require_vt` and `require_st`. It got a bit messy with the template parameter packs. I'd rather add that as a separate PR with that change isolated. I think we should do it. If you're able to implement it, happy to include it with this PR. If you're tied up, I can try to tackle it in a different pass (where I can set up some tests and make sure I'm not going to mess that up somehow).

SteveBronder:
Awesome! I will start looking at this on Friday. I'll take a crack at `require_vt` and `require_st`.

syclik:
@SteveBronder, I think the doc in `doxygen/contributor_help_pages/require_meta.md` is much better now!! (I don't think it's perfect, but I think it documents the non-type template parameter much clearer than it's been written about before.)

Mind taking a look? If you have any suggestions on how to make that doc better, happy to give it a try.

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -1,22 +1,263 @@
...
+    or the @ref stan::scalar_type (the underlying scalar type after
+    recursively walking through the container types) satisfy a
+    condition to enable a class or function.
+
+    The container `requires` template parameters have an ending at
+    their signature of `_vt` and `_st` to symbolize whether you want
+    to inspect the @ref stan::value_type or @ref stan::scalar_type.
+
+    In the next requires traits, `is_type` is used to represent any
+    boolean type trait.
```
SteveBronder, line 175: Is there a reason this is tabbed?

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
 
-- `require_not_*_t` : A template parameter `T` must *not* satisfy the `requires` template parameter in order for the overload to be availabe.
+## Developing new requires type traits {#requires_dev_guide}
 
-*NOTE:* The `not` version of the `requires` template parameters should be used sparingly.
-Often a `requires` template parameter is used to specify what types a function should accept.
-Defining a function by the types it cannot accept can make understanding what goes into a function more difficult and error prone.
+Every requires type trait is not implemented for every boolean type
+trait available. This was done intentionally to allow us to identify
+which requires type traits are currently in use (as of 2024). If you
```
SteveBronder, line 317:
```suggestion
which requires type traits are currently in use. If you
```

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
+feel free to implement the one you need and add a pull request.
 
-```cpp
- // Works for anything that is not a std::vector
- template <typename T, require_not_std_vector_t<T>* = nullptr>
- auto add(T&& x, T&& y) { return x + y; }
-```
+We expect this to happen very infrequently. If it is happening often,
+we can go back to having all possible combinations of requires traits
+available.
```
SteveBronder, line 323:
```suggestion
```

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
 
-- `require_any_*_t` : Takes a parameter pack of types to enable if any of the types satisfy the check.
+### Adding a new requires
 
-```cpp
- // Works if either T1 or T2 enherit from EigenBase
- template <typename T1, typename T2, require_any_eigen_t<T1, T2>* = nullptr>
- auto add(T1&& x, T2&& y) { return x + y; }
-```
+If you need to add a new `requires`, please submit a pull request!
```
SteveBronder, line 334:
```suggestion
```

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
- auto add(Vec1&& A, Vec2&& B) {
-   std::vector<decltype<A[0] + B[0]>> return_vec;
-   std::transform(A.begin(), A.end(), B.begin(), return_vec.begin(),
-     [](auto&& x, auto&& y) {
-         return x + y;
-     });
-   return return_vec;
- }
-```
+    // Ex: Used to define a signature for `std::vectors` with a value type that is autodiffable
```
SteveBronder, line 365:
```suggestion
    // Ex: Used to define a signature for `std::vectors` with a value type that is @ref stan::math::var
```

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
-The `requires` template parameters type traits allow Stan to have more generic types so that the library can forward Eigen expression and have better move semantics.
-For instance, the code below will accept any arbitrary Eigen expression that, if it's an rvalue, can be forwarded to another function.
+    ```
+    template <template <class...> class TypeCheck, class... Check>   
+    require_std_vector_st = require_st<is_std_vector, TypeCheck, std::decay_t<Check>...>;
+
+    // Ex: Used to define a signature for `std::vectors` with a scalar type that is autodiffable
+    template <typename StdVec, require_std_vector_st<is_var, StdVec>* = nullptr>
+    auto my_func(StdVec&& vec);
+    ```
```
SteveBronder, line 380:
```suggestion
```cpp
template <template <class...> class TypeCheck, class... Check>   
require_std_vector_st = require_st<is_std_vector, TypeCheck, std::decay_t<Check>...>;

// Ex: Used to define a signature for `std::vectors` with a scalar type that is autodiffable
template <typename StdVec, require_std_vector_st<is_var, StdVec>* = nullptr>
auto my_func(StdVec&& vec);
```
```

File: doxygen/contributor_help_pages/require_meta.md [hunk 8]
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
+
+There variant of the requires that places the `vt` or `st` before the
+type trait name only checks the @ref value_type or @ref `scalar_type`
+of `T` without testing `T`.
```
SteveBronder, line 384:
```suggestion
The variant of the requires that places the `vt` or `st` before the
type trait name only checks the @ref value_type or @ref `scalar_type`
of `T` without testing `T`.
```

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -49,135 +285,141 @@ class a_class<T, require_std_vector_st<is_var, T>> {
...
+
+6. `require_any_not_{TYPE_TRAIT}_t`: At least one of the template
+parameters must return `false` when passed to the type trait
+
+In addition to all the requires with an `_t` at the end, the requires
+also have `_st`, `_vt` variants where in addition to the logic above,
+the @ref value_type or @ref scalar_type must follow the same logic as
+the type for `T`. The `_st_`, and `_vt_` variants must also follow the
+same logic but for checking only the inner @ref value_type or
+@scalar_type.
```
SteveBronder, line 422: This is all good! Should we add a link somewhere near the top for the docs for all the currently available requires? 

https://mc-stan.org/math/group__require__meta.html

SteveBronder (changes requested):
Just a few changes for the docs but overall I think this is good!

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -1,22 +1,263 @@
...
+broadcasting, which allows users to mix scalars arguments with vector
+or array arguments. 
+
+The typical C++ method of defining a primary function template and a
+set of function template specialization is untenable for many
+functions in the Math library. For example, a single argument of the
+function may take 6 distinct C++ types: `double`, `stan::math::var`,
+`std::vector<double>`, `std::vector<stan::math::var>`,
+`Eigen::Matrix<double, -1, 1>`, or `Eigen::Matrix<stan::math::var, -1,
+1>`.  For a 3-argument function, we would need to define 108 different
```
syclik, line 49:
```suggestion
1>`.  For a 3-argument function, we would need to define 343 (7^3) different
```

doxygen/contributor_help_pages/require_meta.md
syclik: Yes, although it's a hack. I was trying to keep the numbering sequential. I think the enumeration of requires traits is easier if all if it is numbered sequentially from 1 to 18 rather than having it go from 1-6, then 1-12.

Markdown doesn't allow for starting a numbered list from something other than 1. The intent is for the tabbed writing to show up under number 6, but hopefully the context is clear enough that it is read as describing number 7 - 18. 

Maybe it's way too much of a hack and it should just be removed. Or perhaps there's a better way?

doxygen/contributor_help_pages/require_meta.md
syclik: Yes, I'll look for a good place for this.

syclik: Ah. I did link it! (using the `@ref require_meta` tag)

L69 in this doc. It looks like:
```
For convenience in using this technique, the Math library has
implemented a set of [`requires` type traits](@ref require_meta).
```

I'll look for a second place to drop the link.

syclik:
@SteveBronder, thank you for the detailed review! 

1. I accepted all the changes!
2. I fixed the remaining `::type` / `::value` doc issues. (I went and wrote a simple test to make sure I knew exactly what compiled and didn't.)
3. I added the second link to all the require meta traits.
4. I made sure the numbering stayed contiguous. I think it looks ok. Please see the attached image. This is how it renders when I run doxygen.


image

syclik:
@SteveBronder, anything else we need to change on this PR before merging?

SteveBronder:
Let me look this over one more time but I think it is good! Should this get merged into the 5.0 branch since we are getting rid of a lot of the requires? If so then I think we should wait till the other ones are merge into the 5.0 branch as this will probably be the biggest diff

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -1,22 +1,263 @@
...
+`void*` non-type template parameter with a default of `nullptr` is
+valid and the non-type template parameter is ignored by the
+compiler. Utilizing [substitution failure is not an error
+(SFNIAE)](https://en.cppreference.com/w/cpp/language/sfinae), a
+substitution failure for the non-type template parameter will result
+in that definition being removed from the possible function
+definitions. Using this technique we have to be careful not to violate
+the [One Definition
+Rule](https://en.cppreference.com/w/cpp/language/definition) and only
+provide one definition for any set of types.
```
SteveBronder, line 67: How about something like the below instead? I think it would be good to flesh out the explanation of how this works more like in the below

-----------------------------

In the Math library, we use a technique similar to C++20's `require` keyword that allows the definition of
multiple template functions where each handles a subset of allowable
types. 

When the compiler attempts to resolve which function should be called from a set of templated function signatures there must be only one possibly valid function signature available. This is called the [One Definition
Rule](https://en.cppreference.com/w/cpp/language/definition). For example, the following code would fail because the compiler is unable to differentiate between the two function signatures.

```c++
template <typename T>
T foo(T x) {
  return x;
}

template <typename K>
K foo(K x) {
  return x;
}
```

The compiler needs a way to differentiate between the two signatures to select one and satisfy the One Definition Rule. One trick to have a single valid definition is to utilize [Substitution Failure Is Not An Error
(SFNIAE)](https://en.cppreference.com/w/cpp/language/sfinae) to purposefully create conditions where only one signature is valid because all of the other conditions fail to compile. The simplest way to do this is to start with a type trait like the below `enable_if`. The `enable_if` is only defined for the case where `B` is `true` and so if `B` is ever false the compiler would throw an error saying that `enable_if` is not well defined.

```
// Forward declare enable_if
template<bool B, class T = void>
struct enable_if {};

// Only define the case where B is true 
template<class T>
struct enable_if<true, T> { typedef T type; };

template <bool B, typename T>
using enable_if_t = typename enable_if<B, T>::type;
```

Attempting to construct this `enable_if` with `B` being `false` anywhere else in the program would cause the compiler to crash. But using it in the
 template of a function signature allows SFINAE to deduce which signature we
 would like to use.

```c++
// foo only works with floating point types 
template <typename T,  enable_if_t<std::is_floating_point<T>::value>>* = nullptr>
T foo(T x) {
  return x;
}

// foo only works with integer types
template <typename K,  enable_if_t<std::is_intergral<K>::value>>* = nullptr>
K foo(K x) {
  return x;
}

// Calls the first signature
double x_dbl = 1.0;
double y_dbl = foo(x_dbl); 

// Calls the second signature
int x = 1;
int y = foo(x);

```

The second template argument is referred to as a [non-type template
parameter](https://en.cppreference.com/w/cpp/language template_parameters#Non-type_template_parameter) and has a default value of `void`. 
When the templated signature has the correct type the `enable_if_t ` produces a `void` type which is then made into a pointer and assigned a default value of `nullptr`. 
When the templated signature does not have the correct type, the compiler utilizes [Substitution Failure Is Not An Error
(SFNIAE)](https://en.cppreference.com/w/cpp/language/sfinae), to remove the offending signature from the list of possible matches while continuing to search for the correct signature.

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -1,22 +1,263 @@
...
+T foo(const T& t) {
+  // handles stan::math::var
+  return t;
+}
+
+template <typename T, requires_not_var_t<T>* = nullptr>
+T foo(const T& t) {
+  // handles primitives
+  return t;
+}
```
SteveBronder, line 97: Think we should flip these to show "general version" and "overload for var"
```suggestion
template <typename T, requires_not_var_t<T>* = nullptr>
T foo(const T& t) {
  // handles primitives
  return t;
}

template <typename T, requires_var_t<T>* = nullptr>
T foo(const T& t) {
  // handles stan::math::var
  return t;
}
```

doxygen/contributor_help_pages/require_meta.md
```diff
@@ -1,22 +1,263 @@
...
+#### Example
+
+Here's a function that would have two different template functions for
+`stan::math::var` and `double`:
 
-The [`requires` template parameters](@ref require_meta) type traits are aliases for `std::enable_if_t` that have premade conditions for turning on and off function definitions during compilation.
-These are useful for having generalized templates while still overloading a function or class.
-You can think of these as "legacy concepts".
-These are used in a very similar fashion to C++20's `requires` keyword.
+```
```
SteveBronder, line 86:
```suggestion
```c++
```

doxygen/contributor_help_pages/require_meta.md
SteveBronder: No that's fine

SteveBronder (changes requested):
Just one section of the docs to look over. Rest looks good!

doxygen/contributor_help_pages/require_meta.md
syclik: That sounds good! (I had originally written that way too, so I think it's more natural to think that way, but then thought it was easier having the "not" condition second. If it's easier to understand that way, then great!)

doxygen/contributor_help_pages/require_meta.md
syclik: That looks good. I'll put that into the doc.

syclik:
@SteveBronder, I updated the last block of text! There were a couple of very minor edits made:
* A couple words changed for grammar.
* For the code example, the `enable_if` definition isn't a "forward declaration" -- it's an actual definition of the struct. I clarified the comment there. (A forward declaration would have looked like `template<bool B, class T = void> struct enable_if;`. The `{ }` provides the definition of the struct, albeit an empty struct.) 

It really clarifies what's happening. Thanks!!!


> Should this get merged into the 5.0 branch since we are getting rid of a lot of the requires?

My slight (very slight) preference is for it to go into develop. Reasons:
1. The doc is updated!! That is valuable. (If we put in effort, we can pull that out and get it on another PR, but I'm not really looking forward to trying that out.)
2. The expansion of the macros makes it much more readable to a set of developers. (Not everyone. A set will not notice a difference, but this will help a different set.)
3. In the strictest sense, removing the availability of these macros does change what's available, but to say that these traits were meant to be part of external API is pushing the definition a bit. There are other PRs that we merge that violate this with changing function signatures, so I don't think this is really a version-bumping breaking change.
4. I think this moves the code base forward and we can start thinking about breaking changes to include in 5.0. (One thing I've been thinking about is why not have `require_*_t<T>` extend to `require_*_t<T1, T2, ...>` to replace `require_all_*_t<T1, T2, ...>`.)
5. Having this in the code base in develop will help me figure out how to pull apart some of the utils in tests. Having it on another branch is just a context switch that makes it tougher.


With all that said, I'm really not fussed if we put it on 5.0.

SteveBronder:
I'm cool with merging to develop on this. Agree that those requires are not really part of the Stan Math API in the strict sense

SteveBronder (approved):

PR #2969, opened 2023-11-07
Only build tbbmalloc[_proxy] if requested

WardBrian:
## Summary

We have a make variable called TBB_LIBRARIES which controls whether tbb is built or if tbbmalloc and tbbmalloc_proxy are as well. This variable is supposed to be set up such that these optional packages are only built on mac (see [our docs](https://mc-stan.org/math/md_doxygen_2parallelism__support_2threading__tbb.html)). However, at the moment, they are always built due to a misconfigured(?) make dependency.

This leads to https://github.com/stan-dev/cmdstan/issues/1218, which is resolved by this PR.

## Side Effects

As far as I can tell, none, but I wouldn't be shocked if we had baked in some accidental assumption about these libraries being available.

## Release notes

Fixed a build configuration issue where the optional tbbmalloc and tbbmalloc_proxy libraries were built unconditionally.

## Checklist

### DIFF
--- a/make/libraries
+++ b/make/libraries
@@ -169,7 +169,7 @@ endif
 	touch $(TBB_BIN)/tbb-make-check
 
 
-$(TBB_BIN)/tbb.def: $(TBB_BIN)/tbb-make-check $(TBB_BIN)/tbbmalloc.def
+$(TBB_BIN)/tbb.def: $(TBB_BIN)/tbb-make-check
 	@mkdir -p $(TBB_BIN)
 	touch $(TBB_BIN)/version_$(notdir $(TBB))
 	tbb_root="$(TBB_RELATIVE_PATH)" CXX="$(CXX)" CC="$(TBB_CC)" LDFLAGS='$(LDFLAGS_TBB)' '$(MAKE)' -C "$(TBB_BIN)" -r -f "$(TBB_ABSOLUTE_PATH)/build/Makefile.tbb" compiler=$(TBB_CXX_TYPE) cfg=release stdver=c++1y  CXXFLAGS="$(TBB_CXXFLAGS)"


WardBrian:
Jenkins seems to be having some unrelated config issues, the first run of this passed all the relevant things to confirm we don't accidentally rely on this

syclik:
Great! (That sucks that it fails incorrectly at times.)

WardBrian:
I'd never seen that one before, but there seem to be at least a couple sources of nondeterminism in the math pipeline that I have not been able to pin down

syclik:
Nondeterministic behavior is tough to pin down, for sure. 

Maybe an easy-ish on dev fix is retry on failure? Like 1 retry?

WardBrian:
I'm not sure if that is something we could automate or not - good question for @serban-nicusor-toptal

serban-nicusor-toptal:
I think we can just move the step that's failing inside a [retry](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/) block. [stanc3 example ](https://github.com/stan-dev/stanc3/blob/master/Jenkinsfile#L101)
I'll create a PR early tomorrow for Math to retry the headers check.

WardBrian:
The other thing that sometimes randomly fails is the thread tests sometimes don't build.

If this is due to some sort of runner incompatibility, will the retry pick up the same runner again?

serban-nicusor-toptal:
I couldn't find anything for sure but I think it will just pick a different executor based on availability. Looks like there's an agent conditional too -> https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times
I'll read a bit more about it and then try it out in a PR.

PR #2971, opened 2023-11-08
Add arena_matrix specialization for sparse matrices

SteveBronder:
## Summary

While working on a branch to let Stan math use ctest and cmake I found a memory leak in the sparse matrix impl for vari. It was a confusing bug that had to deal with how Eigen's implimentation manages dynamic memory for sparse matrices. This PR removes the use of `chainable_alloc` and directly holding sparse matrices in the `var_value<SparseMatrix>` class and instead uses a new specialization of `arena_matrix` for sparse matrices.

## Tests

Tests were added for the new `arena_matrix`'s constructor, assignment operators, and inplace operators and can be run with.

```bash
python ./runTests.py ./test/unit/math/rev/core/arena_matrix_test.cpp
```

## Side Effects

Yes! A big design decision here is to add inplace operators `+=` and `-=` to the sparse implementation of `arena_matrix`. Normally, `+=` and `-=` do not really make sense for sparse matrices since adding a scalar to all elements of a sparse matrix would just make it a dense matrix. But for Stan math, when we do the adjoint accumulation in reverse mode, we only want to accumulate over the nonzero values of the adjoints. Talking to @bob-carpenter about this, we could think of a sparse matrix as having a mix of `double` and `var` scalars, where everywhere we see a 0 there would be a `double` and everywhere else there would be a `var`. The little vector below is what I mean with this interpretation, where we have doubles for 0 values and vars for nonzero values

```
A = [double(0), double(0), var(x), double(0), double(0), var(y)]
```

If we do `cos(A)`, since `cos(0) = 1`, we end up with a dense matrix of vars like the following 

```
B = cos(A);
print(B);
B = [var(1.0), var(1.0), cos(var(x)), var(1.0), var(1.0), cos(var(y))]
```

We want a dense output of vars because later series of expressions may use those scalar values of the output matrix.

When we call grad for reverse mode, we are taking the adjoint (gradient) of the output with respect to the input. For the zero values of the sparse matrix our adjoint is with respect to a constant which means we don't want to propagate the associated values in our return matrix upwards. So we want our `operator+=(Sparse this, Dense rhs)` to ignore any zero values when we do the adjoint accumulation. This means that when we use `arena_matrix<SparseMatrix>` in our vari implementation we end up with code for the reverse pass that nicely looks like all the rest of our code 

```c++
template <typename SparseMat>
auto cos(const SparseMat& sparse_mat) {
  arena_t<SparseMat> mat_arena = sparse_mat;
  var_value<Eigen::Matrix<double, -1, -1>> ret = cos(value_of(mat_arena));
  reverse_pass_callback([mat_arena, ret]() {
    mat_arena.adj() += -sin(value_of(mat_arena)).cwiseProduct(ret.adj());
  });
  return ret;
}
```

So while it's nonstandard for a sparse matrix to have a `+=` and `-=` defined in it, for our use case it actually makes a lot of sense.

## Release notes

Adds a sparse matrix implimentation for `arena_matrix`

## Checklist


bob-carpenter:
What is the "sparse matrix impl for vari"?  Is this something that's exposed to users anywhere or used elsewhere in our code base?

Is this work toward adding sparse matrices to Stan?  If so, is there a design doc somewhere about the plans?

SteveBronder:
The sparse matrix impl for vari is the SoA pattern for autodiff on sparse matrices. 

There is a design doc below, but it needs updated and does not contain much info on the low level implementation 
https://github.com/SteveBronder/design-docs/blob/spec/sparse-matrices/designs/0004-sparse-matrices.md

bob-carpenter:
Why are you forking the design doc rather than updating?  We don't want to have to review two design docs.  We've traditionally only reviewed functional specs.

My overall suggestion is to work through what some example models would look like, such as a time series model with a sparse Cholesky precision matrix.

If you add a pull request for an updated design design doc, I'd be happy to comment.  Some quick comments:

* All of this is going to be tricky with parameters, because of autodiff.  We can't just take zeroes and drop them out of a matrix of autodiff variables.

* You don't need to mention that we want gradients---we always want gradients with our functions

* by " common optimization technique is to reorder the rows and columns such that the structure is easier for algorithms to traverse." do you mean something like CSR notation?

* I'd put the bounds in the data in that the indexes can't go below 1 or above the number of rows (or columns)

* @dpsimpson and @avehtari have a point---we have to be very strict with sparse matrices defined as parameters and most of the sparseness in regression comes through covariates;  having said that, I don't see how it's fundamentally different than our other types, which are all sized at declaration time---we're just adding sparsity and I would very strongly prefer our design to remain uniform so that the same types can be used in all of the blocks.

* The triple extraction you talk about is buggy because the indexes are integers, but a `matrix[N, 3]` has real entries.  Instead, you can make this an `N`-array of 3-tuples.

* I didn't understand the discussion in the Hard Way vs. the Simple Way---the simple way looked like one of the hard way cases

SteveBronder:
> Why are you forking the design doc rather than updating? We don't want to have to review two design docs. We've traditionally only reviewed functional specs.

This is not a fork, this just changes the underlying vari implementation to use a new `arena_matrix` used for sparse matrices. 

> My overall suggestion is to work through what some example models would look like, such as a time series model with a sparse Cholesky precision matrix.

Do you mean in the design doc?

> If you add a pull request for an updated design design doc, I'd be happy to comment. Some quick comments:

Sure, do you mind if after I copy paste your comments about the design doc into the discussion page.

That design doc needs heavily updated imo, but either way this PR should go through as this is a current bug and I'm 99% sure this is how the impl would need to look for the design we would approve

syclik:
@SteveBronder, I'm guessing this is still an issue. Do you remember what prevented the tests from passing?

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -139,6 +135,264 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
+    return ret;
+  }
+
+ public:
+  /**
+   * Constructs `arena_matrix` from an `Eigen::SparseMatrix`.
+   * @param other Eigen Sparse Matrix class
+   */
+  template <typename T, require_same_t<T, PlainObject>* = nullptr>
+  arena_matrix(T&& other)  // NOLINT
```
andrjohns, line 191: It doesn't look like any of these constructors uses the `innerNonZerosPtr` member from the inputs, is that intentional?

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -139,6 +135,264 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
+  template <typename F, typename Expr,
+            require_convertible_t<Expr&, MatrixType>* = nullptr>
+  void inplace_ops_impl(F&& f, Expr&& other) {
+    auto&& x = to_ref(other);
+    for (int k = 0; k < (*this).outerSize(); ++k) {
+      typename Base::InnerIterator it(*this, k);
+      typename std::decay_t<Expr>::InnerIterator iz(x, k);
+      for (; static_cast<bool>(it) && static_cast<bool>(iz); ++it, ++iz) {
+        f(it.valueRef(), iz.value());
+      }
```
andrjohns, line 300: Should this (and the other overloads) just loop over `(*this).innerSize()`? The compiler might be able to optimise better if it knows the length.

Otherwise, I think this is more readable as a `while` loop:
```cpp
while (static_cast<bool>(it) && static_cast<bool>(iz)) {
  f(it.valueRef(), iz.value());
  ++it;
  ++iz;
}
```

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -139,6 +135,264 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
+            require_not_convertible_t<Expr&, MatrixType>* = nullptr,
+            require_eigen_dense_base_t<Expr>* = nullptr>
+  void inplace_ops_impl(F&& f, Expr&& other) {
+    auto&& x = to_ref(other);
+    for (int k = 0; k < (*this).outerSize(); ++k) {
+      typename Base::InnerIterator it(*this, k);
+      for (; static_cast<bool>(it); ++it) {
+        f(it.valueRef(), x(it.row(), it.col()));
+      }
+    }
```
andrjohns, line 324: Are these operations something that could be implemented using Eigen's `NullaryExpr` framework? That could allow more room for Eigen to optimise/simplify ops

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -139,6 +135,264 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
+
+ public:
+  /**
+   * Inplace operator `+=`
+   * @note Caution! Inplace operators assume that either
+   *  1. The right hand side sparse matrix has the same sparcity pattern
+   *  2. You only intend to add a scalar or dense matrix coefficients to the
+   * nonzero values of `this` This is intended to be used within the reverse
+   * pass for accumulation of the adjoint and is built as such. Any other use
+   * case should be be sure the above assumptions are satisfied.
```
andrjohns, line 356:
```suggestion
   * @note Caution! Inplace operators assume that either
   *  1. The right hand side sparse matrix has the same sparsity pattern
   *  2. You only intend to add a scalar or dense matrix coefficients to the
   * nonzero values of `this`. This is intended to be used within the reverse
   * pass for accumulation of the adjoint and is built as such. Any other use
   * case should be sure that the above assumptions are satisfied.
```

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -139,6 +135,264 @@ class arena_matrix : public Eigen::Map<MatrixType> {
...
+  }
+
+  /**
+   * Inplace operator `+=`
+   * @note Caution!! Inplace operators assume that either
+   *  1. The right hand side sparse matrix has the same sparcity pattern
+   *  2. You only intend to add a scalar or dense matrix coefficients to the
+   * nonzero values of `this` This is intended to be used within the reverse
+   * pass for accumulation of the adjoint and is built as such. Any other use
+   * case should be be sure the above assumptions are satisfied.
```
andrjohns, line 379:
```suggestion
   * @note Caution! Inplace operators assume that either
   *  1. The right hand side sparse matrix has the same sparsity pattern
   *  2. You only intend to add a scalar or dense matrix coefficients to the
   * nonzero values of `this`. This is intended to be used within the reverse
   * pass for accumulation of the adjoint and is built as such. Any other use
   * case should be sure that the above assumptions are satisfied.
```

stan/math/rev/core/var_value_fwd_declare.hpp
```diff
@@ -6,6 +6,16 @@ namespace math {
 // forward declaration of var
 template <typename T, typename = void>
 class var_value;
+
+/**
+ * Equivalent to `Eigen::Matrix`, except that the data is stored on AD stack.
+ * That makes these objects triviali destructible and usable in `vari`s.
```
andrjohns, line 12:
```suggestion
 * That makes these objects trivially destructible and usable in `vari`s.
```

stan/math/rev/core/var_value_fwd_declare.hpp
```diff
@@ -6,6 +6,16 @@ namespace math {
...
+ *
+ * @tparam MatrixType Eigen matrix type this works as (`MatrixXd`, `VectorXd`
+ * ...)
```
andrjohns, line 15:
```suggestion
 * @tparam MatrixType Eigen matrix type this works as (`MatrixXd`, `VectorXd`,
 * ...)
```

andrjohns (changes requested):
Few q's and suggestions, otherwise looks good

stan/math/rev/core/arena_matrix.hpp
SteveBronder: Yes! So actually since we know for the sparse arena matrix the pointer for the values is contiguous from our memory arena we can actually just loop over the value pointer itself with the number of nonzero entries. At least for the scalar and arena matrix inplace ops. For general sparse and dense idt we have that guarantee so we should still do something like the while loop you have in the above

stan/math/rev/core/arena_matrix.hpp
SteveBronder: I think `NullaryExpr` is only available for dense types :/

If we wanted to optimize this harder we could write specific `operator+=` for each of sparse, dense, and scalar ops that under the hood will make an Eigen map to a dense eigen vector of the values. The current inplace ops for two sparse arena matrices is 

```cpp
  template <typename F, typename Expr,
            require_convertible_t<Expr&, MatrixType>* = nullptr,
            require_same_t<Expr, arena_matrix<MatrixType>>* = nullptr>
  inline void inplace_ops_impl(F&& f, Expr&& other) {
    auto&& x = to_ref(other);
    auto* val_ptr = (*this).valuePtr();
    auto* x_val_ptr = x.valuePtr();
    const auto non_zeros = (*this).nonZeros();
    for (Eigen::Index i = 0; i < non_zeros; ++i) {
      f(val_ptr[i], x_val_ptr[i]);
    }
  }
```

But we could instead for `+=` do 

```cpp
  template <typename F, typename Expr,
            require_convertible_t<Expr&, MatrixType>* = nullptr,
            require_same_t<Expr, arena_matrix<MatrixType>>* = nullptr>
  inline void operator+=(F&& f, Expr&& other) {
    auto&& x = to_ref(other);
    auto* val_ptr = (*this).valuePtr();
    auto* x_val_ptr = x.valuePtr();
    const auto non_zeros = (*this).nonZeros();
    Eigen::Map<MatrixType>(val_ptr, non_zeros) += Eigen::Map<MatrixType>(x_val_ptr, non_zeros);
  }
```

That would allow Eigen to optimize a lot more, but at the expense of losing a bit of generalization from just having `inplace_ops_impl`. If you'd like the above then I'm fine with the extra code imo

stan/math/rev/core/arena_matrix.hpp
SteveBronder: Oh! I think I left them out because elsewhere in the Eigen docs it said it was just for compatibility with other packages. But I'll put them in just in case. No reason not to

andrjohns (approved):
LGTM thanks!

Issue #2967, opened 2023-10-25
Allow cmake users to use FetchContent_Declare


SteveBronder:
## Description

It would be nice for cmake users to be able to use something like the following to include Stan math in their projects

```
include(FetchContent)

FetchContent_Declare(
  stanmath
  GIT_REPOSITORY https://github.com/stan-dev/math
  GIT_TAG        v4.7.0
)

FetchContent_MakeAvailable(stanmath)
```
  
I have part of this written [here](https://github.com/SteveBronder/stan-perf/blob/main/CMakeLists.txt) where I have stan math compiling with cmake. I'd be happy to expand this into a CMakeLists.txt file in the stan math repo, which would then allow cmake users to do something like the above to include stan math in their projects.

Do any folks have objections to this? I'd say the cmake file I have there is like 60% complete. It does all the correct includes, but does not handle some of the specific flags we use. It also does not support MPI, but that should be easy to setup.

If I set this up we could also setup our test suite to integrate with ctest

## Expected Output
Please provide the expected output.

If this is a **bug report**, show what you expected to happen. Please also show what currently happens.

If this is a **feature request**, show what you expect to happen if the feature exists.

#### Current Version:
v4.7.0

syclik:
No objections!

WardBrian:
I have my own version of this (modified from Steve’s) here: https://github.com/WardBrian/math/tree/experiment/cmake


It has a bit more flag configuration and some logic to support a “install” target for Math, which would let other tools use a find_library call

PR #2974, opened 2023-11-20
Fix support for oneTBB new inteface >= v2021.1

jeanchristopheruel:
Submitting a support for oneTBB new interface. I suggest that this should be tested with a proper CI/CD.
oneTBB library do not provide <tbb/tbb_stddef.h> anymore as can be seen [here](https://github.com/oneapi-src/oneTBB/tree/master/include/tbb). It is now replaced by <tbb/tbb.h> since v2021.1.


WardBrian:
I believe `__has_include` is only available for C++17, while Stan still supports C++14 as a minimum.

Are you defining `TBB_INTERFACE_NEW` during compilation? That is required to use Stan-math with TBB 2021

jeanchristopheruel:
Instead of relying on the makefile, I use CMake's FetchContent to integrate Stan-math into my project as a module. This approach streamlines the build process and eliminates the need to install unused components, allowing greater flexibility in managing dependencies.

I could find related interest in using FetchContent + CMakeList.txt in  #2967

```
include(FetchContent)
FetchContent_Declare(
     stanmath
     GIT_REPOSITORY https://github.com/jeanchristopheruel/math
     GIT_TAG develop
     GIT_SHALLOW TRUE
     )
FetchContent_MakeAvailable(stanmath)

find_package(Boost 1.81.0 REQUIRED)
find_package(TBB 2021 REQUIRED)
find_package(Eigen3 3.4 REQUIRED)

add_library(stan-math INTERFACE)
target_link_libraries(stan-math INTERFACE Eigen3::Eigen TBB::tbb -lstdc++ -lm)
target_include_directories(stan-math INTERFACE ${stanmath_SOURCE_DIR} ${Boost_INCLUDE_DIRS})
add_library(stan::math ALIAS stan-math)
```

WardBrian:
#2967 is definitely the right place to discuss CMake, and I'm hoping we can support it sooner rather than later.

In the mean time, you should still be using `target_compile_definitions` or just `add_compile_options` to define `TBB_INTERFACE_NEW` as the way of selecting to use newer versions

jeanchristopheruel:
Thanks for your time. I will do so. However, it still wont support TBB_VERSION_MAJOR == 2021. Current statement for `#if TBB_VERSION_MAJOR >= 2020` will fail.

WardBrian:
We use TBB 2021 for the [conda-forge](https://github.com/conda-forge/cmdstan-feedstock) builds of cmdstan with this. If you can share the error you are getting I can try to help debug it, but it is definitely possible to use newer TBBs without source changes

jeanchristopheruel:
OneApi TBB does not have tbb_stddef.h anymore. TBB does.
[Migration guide TBB -> OneApi TBB](https://github.com/oneapi-src/oneTBB/blob/a71747cd869882e51cec231d822c5f31ed70da71/doc/main/tbb_userguide/Migration_Guide/Task_API.rst)

WardBrian:
If you define `TBB_INTERFACE_NEW`, that header will not be included by Stan Math

jeanchristopheruel:
Indeed, as you previously pointed out. My apologies.
Closing issue.

PR #2982, opened 2023-12-06
[CI] Retry headers check and threading tests in case of failure.

serban-nicusor-toptal:
## Summary

Retry headers check and threading tests in case of failure during CI.

## Side Effects

No

## Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -193,9 +193,11 @@ pipeline {
                     }
 
                     steps {
-                        unstash 'MathSetup'
-                        sh "echo CXX=${CLANG_CXX} -Werror > make/local"
-                        sh "make -j${PARALLEL} test-headers"
+                        retry(1){
+                            unstash 'MathSetup'
+                            sh "echo CXX=${CLANG_CXX} -Werror > make/local"
+                            sh "make -j${PARALLEL} test-headers"
+                        }
                     }
                     post { always { deleteDir() } }
                 }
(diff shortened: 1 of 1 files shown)


serban-nicusor-toptal:
Hey @WardBrian the change for headers check is quite straightforward, just retry once.
About your earlier question:
```
The other thing that sometimes randomly fails is the thread tests sometimes don't build.
If this is due to some sort of runner incompatibility, will the retry pick up the same runner again?
```
The retry in it's simplest form seems to handle simple blocks, tho it seems to be possible to use it for a stage-wide failure and retry https://community.jenkins.io/t/how-to-retry-a-jenkins-pipeline-stage-with-an-agent-condition/3667/4
TL;DR; detecting agent issues and if that's the case, find a new agent and try again.

Now while that's possible I don't think it might be best to use in our case, let me explain.
We leverage Docker to ship our CI images with all the dependencies, so irrelevant of the agent we will always run the same thing. The only case where differences can occur is when the code is trying to read the kernel or some low-level CPU instructions (docker containers share the host kernel). Rember we had an issue at the beginning with Flatiron that an Intel CPU was not supporting some CPU instructions ? (or was it a GPU?)

Knowing this, I think it might be more simple and straightforward to detect on which host the `threading tests` fail and simply exclude it in an agent label conditional, thus always running on the valid one.
Could these threading failures occur because of two concurrent runs on the same host ? a develop building running at the same time with a PR build. 

Because of the above train of thought I put it in a simple retry block for now, please let me know what do you think about this! Thanks!

syclik (approved):
LGTM

PR #2983, opened 2023-12-06
Use less memory in `multi_normal_cholesky_lpdf`

spinkney:
## Summary

I've converted the partial matrices to vectors and looped over them to update the derivatives.

## Tests

No new tests.

I did run a benchmark against the develop branch and it shows that the speed is roughly the same.

multi_normal_cholesky_compare

## Side Effects

None.

## Release notes

Increase the memory efficiency of the multivariate normal Cholesky parameterized lpdf.

## Checklist

### DIFF
--- a/stan/math/prim/prob/multi_normal_cholesky_lpdf.hpp
+++ b/stan/math/prim/prob/multi_normal_cholesky_lpdf.hpp
@@ -53,6 +53,9 @@ return_type_t<T_y, T_loc, T_covar> multi_normal_cholesky_lpdf(
   using T_partials_return = partials_return_t<T_y, T_loc, T_covar>;
   using matrix_partials_t
       = Eigen::Matrix<T_partials_return, Eigen::Dynamic, Eigen::Dynamic>;
+  using vector_partials_t = Eigen::Matrix<T_partials_return, Eigen::Dynamic, 1>;
+    using row_vector_partials_t
+      = Eigen::Matrix<T_partials_return, 1, Eigen::Dynamic>;
   using T_y_ref = ref_type_t<T_y>;
   using T_mu_ref = ref_type_t<T_loc>;
   using T_L_ref = ref_type_t<T_covar>;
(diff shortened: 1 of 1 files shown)


stan/math/prim/prob/multi_normal_cholesky_lpdf.hpp
```diff
@@ -119,59 +122,49 @@ return_type_t<T_y, T_loc, T_covar> multi_normal_cholesky_lpdf(
...
+    T_partials_return sum_lp_vec(0.0);
 
     for (size_t i = 0; i < size_vec; i++) {
       decltype(auto) y_val = as_value_column_vector_or_scalar(y_vec[i]);
       decltype(auto) mu_val = as_value_column_vector_or_scalar(mu_vec[i]);
-      y_val_minus_mu_val.col(i) = y_val - mu_val;
+      y_val_minus_mu_val = eval(y_val - mu_val);
+      half = mdivide_left_tri<Eigen::Lower>(L_val, y_val_minus_mu_val)
+                 .transpose();
+      scaled_diff = mdivide_right_tri<Eigen::Lower>(half, L_val).transpose();
```
andrjohns, line 138: This is the part that concerns me, since it's gone from a single solve (with a matrix) to `size_vec` solves (with a vector) for `half` and `scaled_diff` each. Especially when the single larger solve can be better vectorised with SIMD & other compiler opts

Is there enough of a memory hit to justify to extra operations?

andrjohns (changes requested):

stan/math/prim/prob/multi_normal_cholesky_lpdf.hpp
spinkney: Definitely agreed. We do this sort of thing already in
* https://github.com/stan-dev/math/blob/0c16b89a2bbf0153ebf8d435c416cd626506558d/stan/math/prim/prob/multi_normal_lpdf.hpp#L94-L98
* https://github.com/stan-dev/math/blob/0c16b89a2bbf0153ebf8d435c416cd626506558d/stan/math/prim/prob/multi_student_t_lpdf.hpp#L124-L128

I'm really not sure what's best here.

andrjohns: If it's alright with you, I'd prefer not to implement this. The current implementation is likely to scale better to larger inputs, and the changes would also reduce any benefits from OpenCL accelerated-ops.

But also completely happy for you to call someone in for a tie-breaker if you feel strongly about it!

spinkney: I'm fine with closing it but I want someone to weigh in on if we should change the other distributions. I can update the mvn derivatives pr to follow the same.

andrjohns: @SteveBronder - as the Chief of Memory Police, what do you think?

SteveBronder: imo if the benchmarks are not showing a huge difference I think it's good. 

The main thing I'm thinking of is memory pressure. While this code is going to produce more instructions that need to be run for each loop, the memory it's working over will be smaller and more likely to stay in cache. 

@andrjohns I think the memory pressure is more of a focus because either way this is implemented, if a machine has large simd instructions the vector code should still use them. 

@spinkney can you run the benchmark comparing two `var` inputs (and `var<Matrix>` inputs)  like 30 or so times and just plot that? imo I think that's what we care about the most. I have a little repo below you can use to compare both branches. Also let us know what cpu you are using. I can test this on my m2 mac as well if you send me a script to do it.

https://github.com/SteveBronder/stan-perf

spinkney: I have an M1max. Is there someone who could benchmark on a windows and linux machine?

spinkney: I got the library setup but I don't have taskset. Also, how can I set up the script to run the two branches?

SteveBronder: You don't need taskset to run the benchmarks, only if you want to isolate in a single core. 

You can add another branch in your benchmarks cmake file like 

```
FetchContent_Declare(
  stanmathalt
  GIT_REPOSITORY https://github.com/stan-dev/math
  GIT_TAG  mybranch # replace with the version you want to use
)

FetchContent_GetProperties(stanmathalt)
if(NOT stanmathalt_POPULATED)
  FetchContent_Populate(stanmathalt)
endif()

```

Then you can include it in your executible build like `${stanmathalt_SOURCE_DIR}`

spinkney: Then how do I just run benchmarks for this distribution?

SteveBronder: There's an example of how to add a benchmark on the readme and an example benchmark folder below. You need to write a little Cmake file for compiling the benchmark and should be able to use that folder as an example 

https://github.com/SteveBronder/stan-perf/tree/main/benchmarks/matmul_aos_soa

andrjohns (approved):
@spinkney @SteveBronder Considering the points above, I think this is fine to go in without worrying too much about benchmarking

I'll approve the changes and leave it up to @SteveBronder to merge if he agrees

PR #2992, opened 2023-12-22
Add simplex row and column matrix constraints

SteveBronder:
## Summary

This adds the functions `simplex_row_constrain` and `simplex_column_constrain` to be used by the stan language for matrices with simplexes along the rows or columns of the matrix. This is just the constrain version which is already a lot of code. I'm going to add the free functions in a separate PR

## Tests

Tests were added in mix and prim and can be run with 

```bash
python ./runTests.py -j3 ./test/unit/math/mix/fun/simplex_column_constrain_test.cpp \
 ./test/unit/math/mix/fun/simplex_row_constrain_test.cpp \
./test/unit/math/prim/fun/simplex_row_constrain_test.cpp  \
./test/unit/math/prim/fun/simplex_column_constrain_test.cpp 
```

A prim test for `simplex_column_constrain` was not written because under the hood `simplex_column_constrain` for doubles just calls `simplex_constrain`

## Side Effects

There is one interesting thing to point out in the reverse mode versio nof `simplex_column_constrain`. I think this is one of the only places in the Stan math library we use `Eigen::RowMajor` to have row major underlying matrices. 

For each of these algorithms we want to access memory in a contiguous manner. For rows of simplex matrices we start with a (N, M) and get back a (N, M + 1) where we operate on each column of the matrix from left to right. For columns of simplex matrices we start with an (N, M) and get back a (N + 1, M) where we need to operate on the row of the matrix from top to bottom. So for the row simplex everything is nice with column major order because that's how our access happens. But for columns of simplex vectors for reverse mode, we make a hard copy of the input changing it from column major to row major so that the rest of our memory access is contiguous. That means that for the rest of the reverse mode code we access things nicely, but then pay the cost of different storage orders when copy the input to the arena matrix in forward mode and then when we copy the adjoint from the reverse mode into the input's adjoint. But reading from different storage orders twice I think is better than reading from the noncontiguous memory for the entire algorithm. 

## Release notes

Adds constrain function for creating matrices holding row or column simplexes

## Checklist

- [x] Copyright holder: Simons foundation


SteveBronder:
@andrjohns could you take a look at this?

stan/math/prim/fun/simplex_column_constrain.hpp
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * The transform is based on a centered stick-breaking process.
+ *
+ * @tparam Mat type of the Matrix
+ * @param y Free Matrix input of dimensionality (K - 1, M).
+ * @return Matrix with simplex columns of dimensionality (K, M).
+ */
+template <typename Mat, require_eigen_matrix_dynamic_t<Mat>* = nullptr,
+          require_not_st_var<Mat>* = nullptr>
+inline auto simplex_column_constrain(const Mat& y) {
```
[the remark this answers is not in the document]
syclik: Same below.

stan/math/prim/fun/simplex_row_constrain.hpp
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * The transform is based on a centered stick-breaking process.
+ *
+ * @tparam Mat type of the Matrix
+ * @param y Free Matrix input of dimensionality (N, K - 1).
+ * @return Matrix with simplexes along the rows of dimensionality (N, K).
+ */
+template <typename Mat, require_eigen_matrix_dynamic_t<Mat>* = nullptr,
+          require_not_st_var<Mat>* = nullptr>
+inline auto simplex_row_constrain(const Mat& y) {
```
syclik, line 30: Same here for being more explicit with types instead of `auto`.

stan/math/prim/fun/simplex_row_constrain.hpp
```diff
@@ -0,0 +1,132 @@
...
+  auto&& y_ref = to_ref(y);
+  const Eigen::Index N = y_ref.rows();
+  int Km1 = y_ref.cols();
+  plain_type_t<Mat> x(N, Km1 + 1);
+  using eigen_arr = Eigen::Array<scalar_type_t<Mat>, -1, 1>;
```
syclik, line 35: Question: would it now be worth writing a function that accepts a pointer and length? And then using that function instead of reimplementing the same logic.

stan/math/rev/core/var.hpp
```diff
@@ -371,10 +371,23 @@ class var_value<T, internal::require_matrix_var_value<T>> {
    * @param other the value to assign
    * @return this
    */
-  template <typename S, require_assignable_t<value_type, S>* = nullptr,
+  template <typename S,
```
syclik, line 374: What's the purpose of this change? I don't recall seeing a note in the PR.

stan/math/rev/fun/simplex_column_constrain.hpp
```diff
@@ -0,0 +1,173 @@
...
+  using ret_type = plain_type_t<T>;
+  const Eigen::Index N = y.rows();
+  const Eigen::Index M = y.cols();
+  using eigen_mat_rowmajor
+      = Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+  arena_t<eigen_mat_rowmajor> x_val(N + 1, M);
+  if (unlikely(N == 0 || M == 0)) {
+    return ret_type(x_val);
+  }
+  arena_t<internal::change_eigen_options_t<T, Eigen::RowMajor>> arena_y = y;
```
syclik, line 76: Question: is there a cleaner implementation of this?

(It's honestly a question. It seems like you're striving for memory locality by switching to row-major storage. I don't know if there's a better way to accomplish this, but maybe we should try?)

syclik (commented):
Functionality makes sense. Putting in comments to help make the code cleaner going into the code base.

stan/math/prim/fun/simplex_column_constrain.hpp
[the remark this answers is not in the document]
SteveBronder: Yeah we can make it `plain_type_t<Mat>` and that should be fine

stan/math/prim/fun/simplex_row_constrain.hpp
SteveBronder: Sorry I'm not following, where would you want to use this?

stan/math/rev/core/var.hpp
SteveBronder: Oh I need to write docs for this. This change handles `vari_type`s  where the inner type of Eigen matrices are assignable to one another, but the actual `vari_type` is not. So things like `vari_value<Matrix<double, -1, 1>>` and `vari_value<Matrix<double, 1, -1>>`. As pointers those are not assignable to one another, but their inner matrix types are. So the `var` has to make a new `vari` and assign the inner matrices to that new `vari`.

stan/math/rev/fun/simplex_column_constrain.hpp
SteveBronder: Sadly I can't think of one. The problem is that we need to support both types of matrix types, `var<Matrix>` and `Matrix<var>`. So I need some type traits to dig into the matrix and swap out options so we can get a row major matrix here.

And yes the goal here is row major memory. That lets us use simd instructions and keeps the memory we access contiguous

SteveBronder:
@syclik could you give this anotherl look?

syclik:
@SteveBronder, sorry, I just saw a comment that I missed. Thought I'd put it here so you can track it. (It's more of a question about the PR than a single line of code.)

>> @syclik
>> Question: would it now be worth writing a function that accepts a pointer and length? And then using that function instead of reimplementing the same logic.
>>
> @SteveBronder 
> Sorry I'm not following, where would you want to use this?

The question I was asking was whether both `simplex_row_constrain` and `simplex_column_constrain` could call a function that consolidates the implementation. The chunk of code that looked like duplication was:

stan/math/prim/fun/simplex_row_constrain.hpp
```
  eigen_arr stick_len = eigen_arr::Constant(N, 1.0);
  for (Eigen::Index k = 0; k < Km1; ++k) {
    auto z_k = inv_logit(y_ref.array().col(k) - log(Km1 - k));
    x.array().col(k) = stick_len * z_k;
    stick_len -= x.array().col(k);
  }
  x.array().col(Km1) = stick_len;
```

stan/math/rev/fun/simplex_column_constrain.hpp
```
  arr_vec stick_len = arr_vec::Constant(M, 1.0);
  for (Eigen::Index k = 0; k < N; ++k) {
    const double log_N_minus_k = std::log(N - k);
    arena_z.row(k)
        = inv_logit(arena_y.array().row(k).val_op() - log_N_minus_k).matrix();
    x_val.row(k) = stick_len.array() * arena_z.array().row(k);
    stick_len -= x_val.array().row(k);
  }
  x_val.row(N) = stick_len;
```

The version with the jacobian also repeats. I guess I was wondering if it's better to have a C-like function that takes the pointer to data and length and consolidates the computation. We'd still have the same functions from the outside to handle types and everything. 

It's a question, not even a suggestion. I was just reading the code and found the logic to be complicated enough and duplicated where it made me wonder. (Repeating code twice isn't that big of a deal in the long run.)

syclik (changes requested):
Thanks for updating the PR with the doc and return types.

I had a couple minor comments. I think one of the function docs had different indexes than everything else; it'd be nice to change that. Almost everything else is just using slightly more explicit return types. (It just makes it clear that the input type is the output type, which cuts out scanning the function for the return type.)

syclik:
@SteveBronder, minor changes requested. Once addressed, we can merge!

stan/math/prim/fun/simplex_column_constrain.hpp
```diff
@@ -0,0 +1,117 @@
...
+ * @tparam T A standard vector with inner type inheriting from
+ * `Eigen::DenseBase` or a `var_value` with inner type inheriting from
+ * `Eigen::DenseBase` with compile time dynamic rows and dynamic columns
+ * @param[in] y free vector
+ * @param[in, out] lp log density accumulator
+ * @return Standard vector containing matrices with simplex columns of
+ * dimensionality (K, M).
+ */
+template <bool Jacobian, typename T, require_std_vector_t<T>* = nullptr>
+inline auto simplex_column_constrain(const T& y, return_type_t<T>& lp) {
```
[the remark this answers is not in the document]
SteveBronder: This one we want to keep `auto` because it can return a vector with any amount of vectors of matrices within it.

stan/math/prim/fun/simplex_constrain.hpp
```diff
@@ -24,7 +24,7 @@ namespace math {
  * @param y Free vector input of dimensionality K - 1.
  * @return Simplex of dimensionality K.
  */
-template <typename Vec, require_eigen_col_vector_t<Vec>* = nullptr,
+template <typename Vec, require_eigen_vector_t<Vec>* = nullptr,
           require_not_st_var<Vec>* = nullptr>
 inline auto simplex_constrain(const Vec& y) {
```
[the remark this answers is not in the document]
SteveBronder: I changed all of these to plain_type_t

stan/math/prim/fun/simplex_row_constrain.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * absolute Jacobian determinant of constraining transform
+ * @tparam T A standard vector with inner type inheriting from
+ * `Eigen::DenseBase` or a `var_value` with inner type inheriting from
+ * `Eigen::DenseBase` with compile time dynamic rows and dynamic columns
+ * @param[in] y free vector with matrices of size (N, K - 1)
+ * @param[in, out] lp log density accumulator
+ * @return vector of matrices with simplex rows of dimensionality (N, K)
+ */
+template <bool Jacobian, typename T, require_std_vector_t<T>* = nullptr>
+inline auto simplex_row_constrain(const T& y, return_type_t<T>& lp) {
```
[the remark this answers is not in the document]
SteveBronder: Same as above for std vectors of std vectors of matrices. idt plain_type_t here gives anymore information

stan/math/rev/fun/simplex_column_constrain.hpp
```diff
@@ -0,0 +1,173 @@
...
+#include <stan/math/prim/fun/log1p_exp.hpp>
+#include <cmath>
+#include <tuple>
+#include <vector>
+
+namespace stan {
+namespace math {
+
+namespace internal {
+template <typename Mat, int NewOptions, typename = void>
```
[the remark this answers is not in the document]
SteveBronder: Yes I'll move this

SteveBronder:
Actually @syclik do you mind if I just add the free functions to this PR? I think I'm just going to have them call the current `simplex_free` over the rows / columns so it shouldn't be that much extra code

syclik:
@SteveBronder, go for it! I think it makes sense.

andrjohns (approved):
All requests addressed, approving and merging

PR #2999, opened 2024-01-08
Minor patches to TBB Windows build for compatibility with RTools make

andrjohns:
## Summary

This PR applies some minor changes to the TBB's build rules on Windows so that it can be compiled with the version of `make` bundled with RTools - users will no longer need to additionally install `mingw32-make` on windows.

These changes are based on [`RcppParallel`'s changes](https://github.com/RcppCore/RcppParallel/blob/master/src/tbb/build/windows.inc).

I've run a [GHA workflow](https://github.com/stan-dev/math/actions/runs/7452346567/job/20275400861) which shows that the Math library & TBB successfully build with these changes under RTools40, RTools42, and RTools43 - without needing to install any additional packages

## Tests

N/A - existing tests should still pass, [testing workflow](https://github.com/stan-dev/math/actions/runs/7452346567/job/20275400861) already run to confirm successful build

## Side Effects

N/A

## Release notes

Updated TBB Windows build rules for compatibility with RTools make

## Checklist


andrjohns:
I've also left the existing usage of `mingw32-make` in the header tests workflow, to show that the changes won't break any existing configuration

andrjohns:
@WardBrian Any chance we could sneak this through in the 2.34 release? It would help simplify the windows setup steps (only need to set PATH now)

@wds15 would you happen to remember the POSIX issues that led to the `mingw32-make` requirement? I haven't experienced any build or runtime issues yet, but want to double-check that I'm not missing anything

lib/tbb_2020.3/build/windows.inc
```diff
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-export SHELL = cmd
+ifeq (true,$(USING_RTOOLS))
+  export SHELL = sh.exe
+else
+  export SHELL = cmd
+endif
```
WardBrian, line 19: Could we do this in the code that calls the TBB makefiles, rather than needing to modify them here? (I don't mind the extra /s later on)


If so, could we provide some alternative variable like `WINDOWS_HAS_SH` or something? Would be useful for non-RTools environments like conda, but we could still set the default for that variable based on `USING_RTOOLS`

andrjohns:
> Could we do this in the code that calls the TBB makefiles, rather than needing to modify them here? (I don't mind the extra /s later on)

If so, could we provide some alternative variable like WINDOWS_HAS_SH or something?

@WardBrian good call! I've updated the makefiles to first do a check for the presence of `sh`, which can be overridden by manually setting `WINDOWS_HAS_SH`

WardBrian:
Changes look good now, thanks!

As for this in the 2.34 release, I'm torn. Calling the fact that `make` didn't work before a bug would allow it, but I also would not mind more time for testing on something like this.

@rok-cesnovar thoughts?

rok-cesnovar:
First of all, this is awesome. Not requiring mingw32-make is a huge huge win!

But cmdstan scripts still require it right now and I think we would need to clean up the Cmdstan makefiles, cleanup docs and installation instructions before the release. We could do that, but it feels like it would be a bit rushed to me.

And yeah, more testing time would be helpful as well. Asking Windows users to test the RC because of a somewhat big change would be helpful as well.

The only other option is making an RC2 and prolonging the freeze period.

WardBrian:
Agreed @rok-cesnovar, this is a great change! I had always suspected that the issues with `make` and TBB were surface-level due to the makefiles rather than something fundamental, but I never had the knowledge or motivation to dig too deeply. 

Getting the build system out of our way is also the current barrier to using the Microsoft compilers (or at least Clang-for-MSVC) for Stan, which is a goal of mine. One day

wds15:
@andrjohns I only recall that `make` simply did not do it and the mingw thing did it.... that's when I stopped looking for further "why". It would be great to now require any more the mingw thinggy.

WardBrian (approved):

WardBrian:
@andrjohns would you mind looking into what needs to be changed in the stan and cmdstan repos?

PR #3014, opened 2024-01-26
Ensure pointer is 8-byte aligned in stack_alloc::alloc

WardBrian:
## Summary

This was originally found using UBSAN on the `csr_matrix_times_vector` function: https://github.com/stan-dev/rstan/issues/1111

The issue there was that that function places `u`, which is ` std::vector<int>`, in our memory arena. If `u` contained an odd number of elements (e.g., 3 in the above issue), this resulted in a call to `alloc` with a `len` of `12`. 

The _next_ call to `alloc` would then return a pointer which was not 8-byte aligned, which is undefined behavior. On x86 this is relatively benign, but on other platforms this could be either very slow or cause processor exceptions.

## Tests

I have updated `stack_alloc_test`

## Side Effects

This will add a few unused bytes (always less than 8) to allocations of types that `sizeof(T) % 8 !=0`. Additionally, a bit of extra math is needed in the `alloc` function, but @SteveBronder and I checked that this gets compiled down to just a `and`, `sub`, and `cmov`

## Release notes
Fixed the stack_allocator being able to return non-8-byte aligned pointers

## Checklist

### DIFF
--- a/stan/math/memory/stack_alloc.hpp
+++ b/stan/math/memory/stack_alloc.hpp
@@ -167,9 +167,11 @@ class stack_alloc {
    * @return A pointer to the allocated memory.
    */
   inline void* alloc(size_t len) {
+    size_t pad = len % 8 == 0 ? 0 : 8 - len % 8;
+
     // Typically, just return and increment the next location.
     char* result = next_loc_;
-    next_loc_ += len;
+    next_loc_ += len + pad;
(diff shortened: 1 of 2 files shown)


SteveBronder (approved):
At some point I think we should come back and rewrite some of the memory things here so that we can request arbitrary alignments, but I think good!

PR #3048, opened 2024-04-15
Fix csr_matrix_times_vector linker error

SteveBronder:
## Summary

Fixes https://github.com/stan-dev/rstanarm/issues/620 and cleans up `csr_matrix_times_vector`.

The `csr_matrix_times_vector` function was erroring when linking. @WardBrian and I suspect the issue has something to do with the length of the mangled names of either the vtable elements or the windows debugger in mingw. Either way, this PR cuts down on the number of elements that need to be passed to the lamda in the reverse pass callback of `csr_matrix_times_vector` which seems to remove the error. The full list of changes are below.

1. Cleans up `csr_matrix_times_vector`, we now always cast to a SoA Sparse Matrix before setting up the reverse pass callback. 
1. Fixes several small bugs in the arena_matrix<SparseMatrix>, there was a missing const_cast for copying one arena_matrix of sparse type to another.
2. Adds a new function `to_soa_sparse_matrix` for taking objects that are used to represent a sparse matrices CSR format and returning a `var_value<SparseMatrix>`. 
3. Adds a new constructor for `var_value<SparseMatrix/Matrix>` that accepts two arena matrices for pre allocated values and adjoints. This is used in `to_soa_sparse_matrix` so that `w`'s values and adjoints are only shallow copied into the `var_value<SparseMatrix>` 
4. Adds a new `value_of()` function for getting the values of a `SparseMatrix<var>`
5. Adds an overload for `promote_scalar_type` for sparse matrix types
6. Fixes a bug in the `vari_value<SparseMatrix>(arena_matrix)` constructor. If an arena matrix was passed in previously it would be shallow copied to the values and adjoints which would share the value pointer between the two. Now the outer index pointer and inner index pointer are shared between the two, but each has their own pointer for the values.

## Tests

Tests were added for the new `to_soa_matrix function`. Tests for the changed functions can be run with

```bash
python3 ./runTests.py ./test/unit/math/rev/fun/to_soa_sparse_matrix_test.cpp ./test/unit/math/mix/fun/csr_matrix_times_vector_test.cpp
```

## Release notes

Fixes linker issue with `csr_matrix_times_vector`

## Checklist


test/unit/math/rev/fun/to_soa_sparse_matrix_test.cpp
```diff
@@ -0,0 +1,77 @@
+#include <iostream>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/rev/fun/to_soa_sparse_matrix.hpp>
+#include <stan/math/rev/core.hpp>
+#include <test/unit/math/rev/util.hpp>
+#include <vector>
+/*
```
WardBrian, line 7: These should either be uncommented (preferrably) or removed

stan/math/rev/core/arena_matrix.hpp
```diff
@@ -225,8 +224,10 @@ class arena_matrix<MatrixType, require_eigen_sparse_base_t<MatrixType>>
    */
   arena_matrix(const arena_matrix<MatrixType>& other)
       : Base::Map(other.rows(), other.cols(), other.nonZeros(),
-                  other.outerIndexPtr(), other.innerIndexPtr(),
-                  other.valuePtr(), other.innernonZeroPtr()) {}
+                  const_cast<StorageIndex*>(other.outerIndexPtr()),
+                  const_cast<StorageIndex*>(other.innerIndexPtr()),
+                  const_cast<Scalar*>(other.valuePtr()),
+                  const_cast<StorageIndex*>(other.innerNonZeroPtr())) {}
```
WardBrian, line 230: I think I need you to walk me through these const casts. Are the const copy, move, and copy constructors all identical now? That seems odd to me

stan/math/rev/fun/csr_matrix_times_vector.hpp
```diff
@@ -100,46 +67,49 @@ inline auto csr_matrix_times_vector(int m, int n, const T1& w,
...
-    arena_t<return_t> res = w_val_mat * value_of(b_arena);
-    reverse_pass_callback(
-        [m, n, w_arena, w_val_arena, v_arena, u_arena, res, b_arena]() mutable {
-          sparse_val_mat w_val_mat(m, n, w_val_arena.size(), u_arena.data(),
-                                   v_arena.data(), w_val_arena.data());
-          internal::update_w(w_arena, m, n, u_arena, v_arena, b_arena, res);
-          b_arena.adj() += w_val_mat.transpose() * res.adj();
-        });
+    sparse_var_value_t w_mat_arena
+        = to_soa_sparse_matrix<Eigen::RowMajor>(m, n, w, u_arena, v_arena);
```
WardBrian, line 75: Q: I assume bad things happen if you call `to_soa_sparse_matrix` and u and v aren't in the arena? 
If so, could we just move the code which puts them in the area to inside that constructor? This file is basically the only place in the Math lib we currently put vectors of ints in the arena, as far as I can tell, but presumably any other CSR operations would want to do the same

stan/math/rev/fun/csr_matrix_times_vector.hpp
```diff
@@ -100,46 +67,49 @@ inline auto csr_matrix_times_vector(int m, int n, const T1& w,
...
     auto b_arena = to_arena(value_of(b));
-    arena_t<return_t> res = w_val_mat * b_arena;
-    reverse_pass_callback(
-        [m, n, w_arena, v_arena, u_arena, res, b_arena]() mutable {
-          internal::update_w(w_arena, m, n, u_arena, v_arena, b_arena, res);
-        });
+    arena_t<return_t> res = w_mat_arena.val() * b_arena;
+    reverse_pass_callback([res, w_mat_arena, b_arena]() mutable {
+      for (int k = 0; k < w_mat_arena.adj().outerSize(); ++k) {
+        for (typename sparse_var_value_t::vari_type::InnerIterator it(
```
WardBrian, line 106: [optional] I find it a bit easier to read if we can keep the `update_w` function, just because it makes the link between the first case and this one a bit more obvious.

WardBrian (changes requested):

stan/math/rev/core/arena_matrix.hpp
SteveBronder: The actual structure just holds pointers to a data that sits on our stack allocator so anytime we copy or "move" we would just be copying pointers. So that makes them all the same.

We don't null out the move constructors input argument, but that's not a necessary condition for a move constructor and I don't really see a reason to do it here

stan/math/rev/fun/csr_matrix_times_vector.hpp
SteveBronder: All of u, v, and w are copied over to the arena from within `to_soa_sparse_matrix`. An object already allocated on the arena that is handed off to `to_arena()` just returns the original object.

stan/math/rev/fun/csr_matrix_times_vector.hpp
SteveBronder: Actually I found a much easier way to do this

test/unit/math/rev/fun/to_soa_sparse_matrix_test.cpp
SteveBronder: They just needed to be uncommented. Left over from debugging

WardBrian (approved):
Thanks for the changes. Looks good now!

PR #3051, opened 2024-04-21
Add support for Windows ARM64

andrjohns:
## Summary

As mentioned in [this cmdstan issue](https://github.com/stan-dev/cmdstan/issues/1232), R and RTools have released previews of their Windows ARM64 versions.

The TBB is the only part of our existing build system that breaks (across CmdStan, Stan, and Math), as far as I can find.

## Tests

I've tested that CmdStan builds and runs models with `reduce_sum` and `STAN_THREADS=true` using the RTools ARM64 toolchain.

I'm not sure if there's a way to add any CI or similar for this, since GH doesn't have windows arm64 runners yet.

## Side Effects

N/A

## Release notes

Added support for Windows ARM64 with RTools ARM64 toolchain

## Checklist


WardBrian:
Is there any downstream changes this is based on?

it needs an entry in the STAN_CHANGES file but otherwise looks reasonable

WardBrian:
Being able to use cmake would be nice if for no other reason than because it would let us use oneTBB's other improvements like this. 

Anyway, I think it's a pretty small set of people who actually have an arm64 Windows machine. As long as these changes don't actively break x86 I think they're totally fine

PR #3055, opened 2024-04-28
Allow Hessian functors to return Hessian as compressed sparse matrix

andrjohns:
## Summary

The Hessian functors currently return a full symmetric matrix for the Hessian, using double the memory needed. This PR updates the Hessian functors to return only the lower-triangular of the Hessian as a compressed sparse matrix by default, with overloads for returning the full symmetric matrix.

## Tests

N/A - Current signatures now delegate to the sparse implementation, so current tests should continue to pass

## Side Effects

Possibility of reduced memory usage and more efficient Eigen expressions for functions which need to manipulate the Hessian.

## Release notes

Added compressed sparse matrix return for Hessian functors

## Checklist

### DIFF
--- a/stan/math/fwd/functor/hessian.hpp
+++ b/stan/math/fwd/functor/hessian.hpp
@@ -14,6 +14,9 @@ namespace math {
  * mixed definition, which is faster for Hessians, is that this
  * version is itself differentiable.
  *
+ * Instead of returning the full symmetric Hessian, we return the
+ * lower-triangular only as a column-major compressed sparse matrix.
+ *
  * <p>The functor must implement
  *
  * <code>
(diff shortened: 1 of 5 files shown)


SteveBronder:
Should we be keeping the old signatures and just adding new ones?

andrjohns:
> Should we be keeping the old signatures and just adding new ones?

I thought that's what I was doing here (unless I'm misunderstanding).

The old signatures are still present, they just delegate to the sparse signatures for the implementation

SteveBronder (approved):
lgtm

andrjohns:
> lgtm once the upstream tests pass

Adding the early return with `fx = f(x);` actually help catch a bug with the existing hessian implementations. 

The other `hessian` functions already had [this early return](https://github.com/stan-dev/math/blob/e73651b991c8a08f045ad09e0c0124eeaaae8c90/stan/math/mix/functor/hessian.hpp#L50), but they were only compiling because of the templating in the tests.

The functors passed to the `hessian` calls are supposed to take and return the higher-level type ([like `fvar<T>`](https://github.com/stan-dev/math/blob/e73651b991c8a08f045ad09e0c0124eeaaae8c90/stan/math/mix/functor/hessian.hpp#L20)), while the function value is always returned as the lower-level/value type (e.g., [`double`](https://github.com/stan-dev/math/blob/e73651b991c8a08f045ad09e0c0124eeaaae8c90/stan/math/mix/functor/hessian.hpp#L43)).

So directly assigning the result of the functor to the value will only compile if the functor is templated/overloaded, and they should have been wrapped in a `value_of()` call

SteveBronder (approved):

SteveBronder:
@andrjohns actually does this save memory? Since we return a lower right triangle we will have N(N+1)/2 entries. And

PR #3060, opened 2024-05-04
Update GHA tests to run using RTools44

andrjohns:
## Summary

Our Github CI for Windows is significantly out of date with the current RTools windows toolchain, this PR updates the workflows to run using the latest RTools44 toolchain

## Tests

N/A

## Side Effects

N/A

## Release notes

Update Github CI to use RTools44 toolchain on windows

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Windows Rtools40
+name: Windows Rtools44
 
 on:
   pull_request:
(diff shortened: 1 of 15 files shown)


andrjohns:
Turns out the RTools44 GCC has a bug in its `exp(std::complex<double>)` implementation, which is the cause of the `mix` test failures.

The result of `exp(std::complex<double>(inf))` should be `{inf, 0}` [as seen on godbolt](https://godbolt.org/z/hMbT79anv), but the RTools44 gcc returns `{inf, nan}`.

I've notified the toolchain maintainer

andrjohns:
Looks like the reason why the RTools40 toolchain was passing but not Rtools44, is that the msys2-based gcc (like rtools40 and those installed via `pacman`) define `_GLIBCXX11_USE_C99_COMPLEX` by default, but the MXE-based gcc (RTools42-44) don't.

I'm still investigating whether this is an rtools-specific or a general mxe issue, but for now we can just add `-D_GLIBCXX11_USE_C99_COMPLEX` to the windows flags

andrjohns:
> Looks like the reason why the RTools40 toolchain was passing but not Rtools44, is that the msys2-based gcc (like rtools40 and those installed via `pacman`) define `_GLIBCXX11_USE_C99_COMPLEX` by default, but the MXE-based gcc (RTools42-44) don't.
> 
> I'm still investigating whether this is an rtools-specific or a general mxe issue, but for now we can just add `-D_GLIBCXX11_USE_C99_COMPLEX` to the windows flags

I've heard back from the RTools maintainer that they would prefer to wait until the next major RTools release before adding this flag, but that I should report the bug upstream to gcc/libstdc++. So we'll need to manually add the flag in the interim.

@WardBrian / @rok-cesnovar Should I keep the makefile as unconditionally adding the flag on Windows (like it is now)? Alternatives that I can think of would be to:

  - Only add if not already defined (like: `$(echo | g++ -E -dM - | findstr _GLIBCXX11_USE_C99_COMPLEX)`)
  - Only add for MXE toolchain (like: `$(gcc -v | findstr mxe)`)

Thoughts/preferences?

WardBrian:
I think it's fine being unconditional on Windows

andrjohns:
@WardBrian ok to approve and merge before the freeze? Just want to make sure this in now that the next release is recommending the stock RTools toolchain

WardBrian (approved):
Thought I already had! LGTM

PR #2693, opened 2022-03-28
C++17 prep - namespace qualifiers only

andrjohns:
## Summary

This PR implements a subset of the changes from #2641 to assess whether the test leak-sanitizers errors are still present.

This PR implements only the namespace qualification of `apply` and `size`, leaving the `Eigen::Index` and `constexpr` changes for a separate PR

## Tests

N/A - Current tests should still pass

## Side Effects

C++17 compatibility

## Release notes

Added namespace qualifiers to `size` and `apply` calls for c++17 compatibility 

## Checklist

- [x] Math issue #2474

- [x] Copyright holder: Andrew Johnson

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Code-Quality) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #3063, opened 2024-05-06
Request C++17 by default, add fallback flag for now

WardBrian:
## Summary

There have been discussions about moving to C++17 for [a while](https://github.com/stan-dev/math/issues/1300). Stan compiles fine under `-std=c++17` since https://github.com/stan-dev/math/pull/2693, and C++17 is already used by RStan. 

There so far hasn't been a push to actually use C++17 in our code base, and this is a proposal for the first step to do this:

1. We start requesting `-std=c++17` in our Makefiles.
2. We provide a flag called `STAN_USE_CPP14` to switch back to `-std=c++1y`.

That's it. 
This is essentially designed to smoke out users who can't support C++17 while providing them with something they can put in `make/local` to continue compiling for a version or two, **until** we actually add some C++17-requiring source code, at which point we delete the opt-out.

I tried to find a more clever way to do this, but automatically detecting which standards a compiler supports seems to be a difficult/brittle thing to do. 

## Tests

None

## Release notes

Stan now requests `-std=c++17`. C++14 can still be used by defining the make variable `STAN_USE_CPP14`. This variable will go away and C++17 will become a requirement in a future version.

## Checklist

### DIFF
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -123,8 +123,12 @@ CPPFLAGS_SUNDIALS ?= -DNO_FPRINTF_OUTPUT $(CPPFLAGS_OPTIM_SUNDIALS) $(CXXFLAGS_F
 #CPPFLAGS_GTEST ?=
 
 
-## setup compiler flags
-CXXFLAGS_LANG ?= -std=c++1y
+ifeq ($(STAN_USE_CPP14),)
+  CXXFLAGS_LANG ?= -std=c++17
+else
+  $(warning "Because STAN_USE_CPP14 is set C++14 standard is requested. This will not be possible in the next release!")
+  CXXFLAGS_LANG ?= -std=c++1y
+endif
 #CXXFLAGS_BOOST ?=
 CXXFLAGS_SUNDIALS ?= -pipe $(CXXFLAGS_OPTIM_SUNDIALS) $(CPPFLAGS_FLTO_SUNDIALS)
 #CXXFLAGS_GTEST
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -120,44 +120,31 @@ INC_GTEST ?= -I $(GTEST)/include -I $(GTEST)
 CPPFLAGS_BOOST ?= -DBOOST_DISABLE_ASSERTS
 CPPFLAGS_SUNDIALS ?= -DNO_FPRINTF_OUTPUT $(CPPFLAGS_OPTIM_SUNDIALS) $(CXXFLAGS_FLTO_SUNDIALS)
 #CPPFLAGS_GTEST ?=
-HAS_CXX17 := false
+STAN_HAS_CXX17 ?= false
 ifeq ($(CXX_TYPE), gcc)
   GCC_GE_73 := $(shell [ $(CXX_MAJOR) -gt 7 -o \( $(CXX_MAJOR) -eq 7 -a $(CXX_MINOR) -ge 1 \) ] && echo true)
   ifeq ($(GCC_GE_73),true)
-    CXXFLAGS_LANG ?= -std=c++17
-    CXXFLAGS_STANDARD ?= c++17
-    HAS_CXX17 := true
+    STAN_HAS_CXX17 := true
   endif
 else ifeq ($(CXX_TYPE), clang)
   CLANG_GE_5 := $(shell [ $(CXX_MAJOR) -gt 5 -o \( $(CXX_MAJOR) -eq 5 -a $(CXX_MINOR) -ge 0 \) ] && echo true)
   ifeq ($(CLANG_GE_5),true)
-    CXXFLAGS_LANG ?= -std=c++17
-    CXXFLAGS_STANDARD ?= c++17
-    HAS_CXX17 := true
+    STAN_HAS_CXX17 := true
   endif
 else ifeq ($(CXX_TYPE), mingw32-gcc)
   MINGW_GE_50 := $(shell [ $(CXX_MAJOR) -gt 5 -o \( $(CXX_MAJOR) -eq 5 -a $(CXX_MINOR) -ge 0 \) ] && echo true)
   ifeq ($(MINGW_GE_50),true)
-    CXXFLAGS_LANG ?= -std=c++17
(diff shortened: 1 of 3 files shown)


SteveBronder:
Should we put this in the 5.0 branch?

WardBrian:
In my opinion no - the real breaking change would actually be when we remove the flag and start using C++17-only constructs in the code. 
The utility of doing this beforehand would be letting users know if their environment is incompatible early, while we can still give them an escape hatch

make/compiler_flags
```diff
@@ -123,8 +123,12 @@ CPPFLAGS_SUNDIALS ?= -DNO_FPRINTF_OUTPUT $(CPPFLAGS_OPTIM_SUNDIALS) $(CXXFLAGS_F
...
 
 
-## setup compiler flags
-CXXFLAGS_LANG ?= -std=c++1y
+ifeq ($(STAN_USE_CPP14),)
+  CXXFLAGS_LANG ?= -std=c++17
+else
+  $(warning "Because STAN_USE_CPP14 is set C++14 standard is requested. This will not be possible in the next release!")
+  CXXFLAGS_LANG ?= -std=c++1y
+endif
```
SteveBronder, line 131: I don't like that this is going to cause compiler errors for users who do not have c++17. Can we do something like the below? Idt this is very portable

```make
CPP_STANDARD := $(shell $(CXX) -x c++ -E -dM - < /dev/null | grep -oP '__cplusplus \K\d+')

# Convert the CPP_STANDARD into a format usable by Make's conditionals
IS_CPP_STANDARD_GREATER_THAN_2017 := $(shell [ $(CPP_STANDARD) -gt 201700 ] && echo 1 || echo 0)

# Conditional directive based on the CPP_STANDARD
ifeq ($(IS_CPP_STANDARD_GREATER_THAN_2017),1)
CXXFLAGS_LANG ?= -std=c++17
else
$(warning "Your compiler does not support C++17 which will be required in the next release! Please upgrade your compiler.")
CXXFLAGS_LANG ?= -std=c++1y
endif
```

WardBrian: That won't actually detect if the compiler _supports_ c++17, that will detect if it is the _default_ or not.

I agree it would be better if we can detect it, but I haven't found any reasonable way of doing so besides doing something like grepping `g++ -v --help` (and I have no idea if that would work with clang)

SteveBronder: Hmmm, what if we used CXX_MAJOR and CXX_MINOR to figure that out? We know what version of gcc and clang will support c++17. If someone is using neither of those we can just still have the way to downgrade to C++14 if they have an old compiler.

Then the only users whose code fails will be people not using gcc/clang

WardBrian: I think some combination of `TBB_CXX_TYPE`, `CXX_MAJOR`, and `CXX_MINOR` could be used, yes. We'd need to know the exact values we want for each compiler

andrjohns: I think it might be a bit more robust/portable to test for [a feature macro](https://en.cppreference.com/w/cpp/feature_test) for the corresponding support level

For example:
```sh
> echo | g++ -std=c++14 -x c++ -E -dM - | grep __cpp_aggregate_bases
> echo | g++ -std=c++17 -x c++ -E -dM - | grep __cpp_aggregate_bases
#define __cpp_aggregate_bases 201603L
```

WardBrian: I think the second command there will fail if `-std=c++17` isn't a recognized standard by the compiler

andrjohns: Still usable in a makefile though. 

Given dummy makefile:

```make
CXX17_CHECK := $(shell echo | $(CXX) -std=c++17 -x c++ -E -dM - 2>&1 | grep __cpp_aggregate_bases)
CXX30_CHECK := $(shell echo | $(CXX) -std=c++30 -x c++ -E -dM - 2>&1 | grep __cpp_aggregate_bases)

std_test:
	@echo "$(CXX17_CHECK)"
	@echo "$(CXX30_CHECK)"
	@echo "end"
```

Result is:

```sh
> make std_test
#define __cpp_aggregate_bases 201603L

end
```

WardBrian: @andrjohns that seems better - could you take on writing the check?

My other concern is similar to when we added something like that for UCRT detection, which is if it will noticeably slow down compilation. If so we'd want to make a `make/cpp17` file or something that stores the result?

SteveBronder: @WardBrian could you check out the most recent change I made to this PR? I added some checks for each compiler version and if it is a version we know supports c++17 we use that as the standard

WardBrian: I think the basic idea would work, but I'm pretty sure the logic as you have it written leads to CXXFLAGS_LANG just being completely unset if the user has a compiler type we recognize but it's too old?

WardBrian: I'm also not confident that the `[ ]` program works on Windows tbh

SteveBronder: > I think the basic idea would work, but I'm pretty sure the logic as you have it written leads to CXXFLAGS_LANG just being completely unset if the user has a compiler type we recognize but it's too old?

Oh, yes!

It looks like the tests are running on windows for jenkins?

https://github.com/stan-dev/math/actions/runs/9116147777/job/25064095453?pr=3063#step:9:23

WardBrian: I think the logic now is correct but overcomplicated. I would not set CXXFLAGS_PROGRAMS inside each if block, just set the `HAS_CPP17` flag, and then at the end (pseudocode)

```
if HAS_CPP17 
  CXXFLAGS_PROGRAM ?= -std=c++17
else
  CXXFLAGS_PROGRAM ?= -std=c++1y
  if NOT STAN_USE_CPP14
    warning()

SteveBronder:
@WardBrian do we want to merge this to include in this release?

WardBrian:
@SteveBronder I think we will need another round of RCs for other changes, so we can!

Because I opened this PR I can't actually click 'approve', so if you can that would be great

SteveBronder (approved):

PR #3064, opened 2024-05-09
Revert "Allow Hessian functors to return Hessian as compressed sparse matrix"

SteveBronder:
Reverts stan-dev/math#3055

@andrjohns thinking about this the sparse matrix is only going to save more memory if nnz < (n * (n - 1) - 1) / 2. But we have a lower triangular which takes up n * (n-1) / 2 entries. So idt this is actually saving any memory


WardBrian (approved):
Especially with the release next week it seems safer to revert if there are remaining questions

PR #3069, opened 2024-05-15
[WIP] Sparse cholesky

dpsimpson:
## Summary

Decided to dive in on the sparse cholesky. This is a WIP.

- [ ] Implement prim sparse cholesky
- [ ] Implement prim sparse triangular solves
- [ ] Implement prim sparse triangular multiplies
- [ ] Other things that a sparse triangular matrix should be able to do?
- [ ] rev mode all things
- [ ] Tests
- [ ] Docs

The interesting wrinkle here is that 
```
Eigen::SimplicialLLT<SpMat>(A) llt =  m.llt()
```
 does not (without a lot of coercing) perform a Cholesky decomposition of `A`. It instead performs a Cholesky of `A.twistedBy(perm)`, where `perm` is stored in the `Eigen::SimplicialLLT<T>` class. This means that it is not enough for `cholesky_decompose` to return
```
llt.matrixL()
```
as this will not be the matrix we are looking for. This is the problem with @SteveBronder's old branch https://github.com/stan-dev/math/blob/a97bfa2f9a418bba192666133a28930651066fe7/stan/math/prim/mat/fun/cholesky_decompose.hpp#L114

The two options here are to compute perm explicitly somewhere else and carry it around (I do not like this option). The other option is to carry treat `Eigen::SimplicialLLT<T>` _as if it were the lower triangular matrix_ when performing triangular solves, multiplication, and anything else we might need. (those operations require knowledge of `perm`).

I'm not quite sure how my preferred option will work with autodiff - we might need a light specialization.

## Tests

Coming

## Side Effects

By adding a new, non-matrix Eigen type, we need to be very careful that none of the other template patterns match it. It should be fine, but care is needed.

## Release notes

## Checklist

- [x] Copyright holder: Daniel Simpson

### DIFF
--- a/stan/math/prim/err/check_pos_definite.hpp
+++ b/stan/math/prim/err/check_pos_definite.hpp
@@ -68,7 +68,6 @@ inline void check_pos_definite(const char* function, const char* name,
   }
 }
 
-
 /**
  * Check if the specified LDLT decomposition of a matrix is positive definite.
  *
--- a/stan/math/prim/err/check_pos_definite.hpp
+++ b/stan/math/prim/err/check_pos_definite.hpp
@@ -68,6 +68,44 @@ inline void check_pos_definite(const char* function, const char* name,
   }
 }
 
+
+/**
+ * Check if the specified LDLT decomposition of a matrix is positive definite.
+ *
+ * @tparam Derived type of the Eigen::LDLT decomposition
+ * @param function function name (for error messages)
+ * @param name variable name (for error messages)
+ * @param cholesky Eigen::LDLT to test, whose progenitor
+ * must not have any NaN elements
+ * @throw std::domain_error if the matrix is not positive definite
+ */
+template <typename Derived>
+inline void check_pos_definite(const char* function, const char* name,
+                               const Eigen::LDLT<Derived>& cholesky) {
+  if (cholesky.info() != Eigen::Success || !cholesky.isPositive()
+      || !(cholesky.vectorD().array() > 0.0).all()) {
+    throw_domain_error(function, "LDLT decomposition of", " failed", name);
+  }
+}
+
+/**
+ * Check if a sparse LLT decomposition of a matrix is positive definite.
+ *
+ * @tparam Derived type of the Eigen::SimplicalLLT decomposition
+ * @param function function name (for error messages)
+ * @param name variable name (for error messages)
+ * @param cholesky Eigen::SimplicalLLTT to test, whose progenitor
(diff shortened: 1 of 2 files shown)


dpsimpson:
argh sorry. didn't expect the CI to flag with a draft WIP PR.

spinkney:
If you figure out the permutation thing I think that would also apply to the LDL decomposition because it does pivoting. I've been wanting to get that into stan-lang but I don't know how to differentiate and handle that issue.

dpsimpson:
Do you think it needs to be exposed at the language level? To me, things like pivoting are just implementation details that should be abstracted away. If this thing works with sparse matrices, it should be the same there too.

spinkney:
I mean the derivatives are what's hard. At the language level no one needs to know the pivoting. I have an implementation of a cholesky factor of correlation matrices but it constructs the LD factor instead and removes the need for the square root. So users could have a ldl_factor_correlation that is a tuple object of a lower triangular matrix and a vector that holds the D diagonal. It makes sense to have the ability to construct an LDL tuple from a PD matrix too hence why I want the LDL factorization.

dpsimpson:
Permutation matrices are linear and have determinant one, so I don't think they affect the derivatives (it is 8am and I've not had coffee, so I might be wrong)

dpsimpson:
In my mind, this would not be terribly difficult to implement, but it would add the wrinkle that the `vari` type would now have a solver type for `val_` and a matrix type for `adj_`. I'm not deep enough into the autodiff system to know if that would be a disaster (maybe for higher order? afaik the only place val_ and adj_ meet is in the chain methods, which are hand-written).

SteveBronder:
> In my mind, this would not be terribly difficult to implement, but it would add the wrinkle that the vari type would now have a solver type for val_ and a matrix type for adj_. I'm not deep enough into the autodiff system to know if that would be a disaster (maybe for higher order? afaik the only place val_ and adj_ meet is in the chain methods, which are hand-written).

The return type of the function has to be a matrix or tuple of matrices for the stan language, but you can store the actual solver to use in the reverse pass like we do here for `mdivide_left_ldlt`

https://github.com/stan-dev/math/blob/develop/stan/math/rev/fun/mdivide_left_ldlt.hpp#L42

SteveBronder:
Also is the permutation matrix fixed for each iteration of the sampler? If so would we want a cholesky signature where the user passes in the permutation matrix?

dpsimpson:
I'm closing this because I've gone a different direction.

PR #3070, opened 2024-05-16
Fix ld error with TBB_LIB on macos

andrjohns:
## Summary

While [helping a user](https://discourse.mc-stan.org/t/cannot-compile-in-cmdstanr-with-compile-model-methods-true/34464/4) specify the Homebrew TBB for use with CmdStan, I discovered that the linker on macos doesn't support the `-Wl,--disable-new-dtags` argument:

```
andrew@Andrews-MacBook-Air ~ % clang -v
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
andrew@Andrews-MacBook-Air ~ % touch foo.cpp
andrew@Andrews-MacBook-Air ~ % clang++ foo.cpp -Wl,--disable-new-dtags
ld: unknown options: --disable-new-dtags 
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It's not an issue with the Linux clang though (checked in a Debian Sid docker image)

## Tests

N/A

## Side Effects

N/A

## Release notes

Fix linker error when using external TBB on MacOS

## Checklist

### DIFF
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -284,7 +284,12 @@ CXXFLAGS_TBB ?= -I $(TBB_INC)
 else
 CXXFLAGS_TBB ?= -I $(TBB)/include
 endif
-LDFLAGS_TBB ?= -Wl,-L,"$(TBB_LIB)" -Wl,--disable-new-dtags
+LDFLAGS_TBB ?= -Wl,-L,"$(TBB_LIB)"
+
+# MacOS ld does not support --disable-new-dtags
+ifneq ($(OS),Darwin)
+  LDFLAGS_TBB += -Wl,--disable-new-dtags
+endif
 
 # Windows LLVM/Clang does not support -rpath, but is not needed on Windows anyway
 ifneq ($(OS), Windows_NT)


wds15:
And we still get the right TBB being linked in on macOS? The flag was introduced on Linux to enforce that no system defaults lead to link against a system TBB rather than the supplied TBB by Stan.

andrjohns:
> And we still get the right TBB being linked in on macOS? The flag was introduced on Linux to enforce that no system defaults lead to link against a system TBB rather than the supplied TBB by Stan.

The flag is only used when the `TBB_LIB` variable is set - and a system TBB is being linked against

WardBrian (approved):
Great!

We've actually been patching this out in the conda build for some time, I've just continuously forgotten to upstream it: https://github.com/conda-forge/cmdstan-feedstock/blob/main/recipe/no-dtag-flag.patch

andrjohns:
> Great!
> 
> We've actually been patching this out in the conda build for some time, I've just continuously forgotten to upstream it: https://github.com/conda-forge/cmdstan-feedstock/blob/main/recipe/no-dtag-flag.patch

Peak downstream behaviour

WardBrian:
This change broke a build I had that was manually setting `LDFLAGS_TBB`, since now the dtags flag is +='d it doesn't respect previous settings the way `?=` does.

The build in question is _not_ on a Mac, but _does_ use clang's lld. We don't really have a good way of detecting this automatically.

I also noticed this flag is only actually being passed if you are using a custom TBB, not one vendored by Stan, so I wonder why we even need it at all?

andrjohns:
> This change broke a build I had that was manually setting LDFLAGS_TBB, since now the dtags flag is +='d it doesn't respect previous settings the way ?= does.

A that makes sense, we could just wrap the `+=` in an `ifeq ($(LDFLAGS_TBB),)`?

> I also noticed this flag is only actually being passed if you are using a custom TBB, not one vendored by Stan, so I wonder why we even need it at all?

Because if you try to use a custom TBB with Stan on Macos then you can't build the math library/cmdstan at all - because macos doesn't support the flag

WardBrian:
> Because if you try to use a custom TBB with Stan on Macos then you can't build the math library/cmdstan at all - because macos doesn't support the flag

I meant "why do we need `--disable-new-dtags` on any platform". Like what about our vendored TBB means we don't need it on linux etc, but a user-supplied ones means we do?

WardBrian:
I think we can get nice, customizable behavior with

```makefile

# MacOS ld does not support --disable-new-dtags
ifneq ($(OS),Darwin)
  LDFLAGS_TBB_DTAGS ?= -Wl,--disable-new-dtags
endif

# Windows LLVM/Clang does not support -rpath, but is not needed on Windows anyway
ifneq ($(OS), Windows_NT)
  LDFLAGS_TBB_RPATH ?= -Wl,-rpath,"$(TBB_LIB)"
endif

LDFLAGS_TBB ?= -Wl,-L,"$(TBB_LIB)" $(LDFLAGS_TBB_DTAGS) $(LDFLAGS_TBB_RPATH)
```

PR #3071, opened 2024-05-17
Update RNG tests to use mixmax RNG - declare stan::rng_t in Math library instead

andrjohns:
## Summary

As seen over in [this Stan issue](https://github.com/stan-dev/stan/issues/3285), it looks like the new `mixmax` RNG has some edge cases. This PR updates our tests to use the same RNG - just in case there are any others.

I also think we should have the common RNG declaration (`stan::rng_t`) in the Math library instead, so that any changes are implicitly tested - but also happy to revert if others disagree

## Tests

N/A

## Side Effects

N/A

## Release notes

Update tests to use mixmax RNG, declare `stan::rng_t` in Math library

## Checklist

- [] the basic tests are passing

### DIFF
--- a/test/unit/math/fwd/prob/lkj_corr_test.cpp
+++ b/test/unit/math/fwd/prob/lkj_corr_test.cpp
@@ -5,7 +5,7 @@
 
 TEST(ProbDistributionsLkjCorr, fvar_double) {
   using stan::math::fvar;
-  boost::random::mt19937 rng;
+  boost::random::mixmax rng;
   int K = 4;
   Eigen::Matrix<fvar<double>, Eigen::Dynamic, Eigen::Dynamic> Sigma(K, K);
   Sigma.setZero();
@@ -24,7 +24,7 @@ TEST(ProbDistributionsLkjCorr, fvar_double) {
 
 TEST(ProbDistributionsLkjCorrCholesky, fvar_double) {
   using stan::math::fvar;
-  boost::random::mt19937 rng;
+  boost::random::mixmax rng;
   int K = 4;
   Eigen::Matrix<fvar<double>, Eigen::Dynamic, Eigen::Dynamic> Sigma(K, K);
   Sigma.setZero();
@@ -43,7 +43,7 @@ TEST(ProbDistributionsLkjCorrCholesky, fvar_double) {
 
 TEST(ProbDistributionsLkjCorr, fvar_fvar_double) {
   using stan::math::fvar;
-  boost::random::mt19937 rng;
+  boost::random::mixmax rng;
   int K = 4;
   Eigen::Matrix<fvar<fvar<double> >, Eigen::Dynamic, Eigen::Dynamic> Sigma(K,
                                                                            K);
@@ -63,7 +63,7 @@ TEST(ProbDistributionsLkjCorr, fvar_fvar_double) {
 
 TEST(ProbDistributionsLkjCorrCholesky, fvar_fvar_double) {
   using stan::math::fvar;
-  boost::random::mt19937 rng;
+  boost::random::mixmax rng;
   int K = 4;
   Eigen::Matrix<fvar<fvar<double> >, Eigen::Dynamic, Eigen::Dynamic> Sigma(K,
                                                                            K);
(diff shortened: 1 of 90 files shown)


WardBrian:
Because the downstream usages do not directly call the constructor of the rng type directly and instead call the utility function that accepts a seed and chain ID, I don’t see the value of moving the typedef here (if anything, it might give the wrong idea?)

andrjohns:
> Because the downstream usages do not directly call the constructor of the rng type directly and instead call the utility function that accepts a seed and chain ID, I don’t see the value of moving the typedef here (if anything, it might give the wrong idea?)

It was more to make sure that we were always using the same RNG in Stan and the Math library tests, but I'm not particularly invested either way. I'll revert for now to just specifying `boost::random::mixmax` since this was more about running the tests, and the RNG typedef location can be a separate discussion

WardBrian:
I think it's fine to want to use the same RNG for testing here as is used in Stan (though historically this wasn't done, and I could also see arguments for _not_ wanting to do this)

But it's worth noting that

> I also think we should have the common RNG declaration (`stan::rng_t`) in the Math library instead, so that any changes are implicitly tested

Isn't really true, at least in the sense that it would not have given us any earlier notice on this issue or anything else that was specific to certain seeds, because as far as I can tell we either default construct the RNGs used in tests or use a pretty small list of seeds (e.g. `1234`)

test/unit/math/prim/prob/inv_wishart_cholesky_rng_test.cpp
```diff
@@ -91,7 +91,7 @@ TEST(ProbDistributionsInvWishartCholesky, SpecialRNGTest) {
   using stan::math::inv_wishart_cholesky_rng;
   using stan::math::multiply_lower_tri_self_transpose;
 
-  boost::random::mt19937 rng(92343U);
+  boost::random::mixmax rng(92343U);
```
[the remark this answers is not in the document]
spinkney: This test should have a large tolerance because it's just checking the values on average.

test/unit/math/fwd/prob/lkj_corr_test.cpp
```diff
@@ -1,11 +1,11 @@
 #include <stan/math/fwd.hpp>
 #include <gtest/gtest.h>
-#include <boost/random/mersenne_twister.hpp>
+#include <boost/random/mixmax.hpp>
 #include <boost/math/distributions.hpp>
 
 TEST(ProbDistributionsLkjCorr, fvar_double) {
   using stan::math::fvar;
-  boost::random::mt19937 rng;
+  boost::random::mixmax rng;
```
andrjohns, line 8: @spinkney another LKJ question for you. This test is now failing on the derivative comparison below:

```cpp
EXPECT_FLOAT_EQ(2.5177896, stan::math::lkj_corr_log(Sigma, eta).d_);
```

Would you know where the fixed value `2.5177896` came from? (I know this wasn't a test you wrote, but I figured you probably understand this better than I do!)

spinkney: No, this doesn't seem right. The derivative is wrt to eta, right? And eta is changing. I think the derivative should be
```r
eta <- 0.3
d <- 4

d_eta <- -(d - 2) * digamma((d - 1) / 2 + eta) + 2 * digamma(d/2 + eta - 1) - 2 * digamma(d + 2 * eta - 2) + log(4)
for (k in 2:(d-1)) {
  d_eta <- d_eta + digamma(eta + (d-1-k)/2)            
}
d_eta
```

WardBrian: Using an RNG in this test seems like isn't necessary anyway. There is no randomness used, the value of `eta` is the same every time, we just avoid writing it down explicitly I guess?

If you replace the definition for `eta` with the line ` fvar<double> eta = 1.6294473838061094;`, which is just what the result of the default-constructed mt19937 was giving before, the test passes again.

spinkney: That makes more sense and the rng should be removed.

andrjohns:
Obsoleted

PR #3076, opened 2024-05-22
Compile sundials with -fPIC to allow shared libraries to link

andrjohns:
## Summary

When exposing Stan functions or model methods (log_prob, etc) in `cmdstanr`, R compiles a shared object and links to the CmdStan-built libraries.

However this causes an error on Linux system where `-fPIC` is required for linking static libraries to shared libraries, and the user needs SUNDIALS/CVODES functions:

```
sunmatrix_band.c:(.text+0x87c): dangerous relocation: unsupported relocation
/usr/bin/ld: /root/.cmdstan/cmdstan-2.34.1/stan/lib/stan_math/lib/sundials_6.1.1/lib/libsundials_kinsol.a(sunmatrix_band.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `SUNMatClone_Band' which may bind externally can not be used when making a shared object; recompile with -fPIC
```

## Tests

N/A

## Side Effects

N/A

## Release notes

Added `-fPIC` to `SUNDIALS` compilation flags so that downstream shared libraries can link against them

## Checklist


WardBrian:
We also ran into this [in BridgeStan](https://github.com/roualdes/bridgestan/blob/fb734eaf59efd63c907bba057e91e9b7c651fc93/Makefile#L22)

I am not sure if doing it universally is the right thing to do, however, since for usages like CmdStan it is not necessary and can supposedly lead to slower assembly being generated

andrjohns:
> We also ran into this [in BridgeStan](https://github.com/roualdes/bridgestan/blob/fb734eaf59efd63c907bba057e91e9b7c651fc93/Makefile#L22)
> 
> I am not sure if doing it universally is the right thing to do, however, since for usages like CmdStan it is not necessary and can supposedly lead to slower assembly being generated

Ah yeah I see. I've updated `cmdstanr` to check for the flag and add/rebuild if necessary, so we can close this if it's preferable

WardBrian:
If that works for you I think that's how we should proceed for now

PR #3086, opened 2024-06-17
Consistently use 'long int' for return of size functions

### LINKED ISSUE #3085: Inconsistent type used for size of containers
## Description

The various functions in the Math library that deal with sizes of containers return different types than each other:

- `size()` returns `size_t` (equivalent to `unsigned long int`)
- `rows()` returns `int`
- `cols()` returns an `Eigen::Index` which at least on GCC is equivalent to `long int`
-  `num_elements()` returns `int`
- `dims()` returns a vector of `int`

In particular, `size_t` is problematic because it is an _unsized_ type, which leads to bad semantics when the result is used in a mathematical expression as seen in this [forum post](https://discourse.mc-stan.org/t/differing-behaviour-when-using-explicit-vs-implicit-ints/35526/)

We should probably use the same type for all of these functions. Because `int` is what is used in the language, that seems like a reasonable choice. If we're worried about internal overflows, we should at least used a signed type like `ssize_t`/`long int`, rather than `size_t`

#### Current Version:
v4.9.0

WardBrian:
## Summary

Closes #3085 

## Side Effects

I certainly hope we weren't relying on something like these having wrapping semantics.

## Release notes

Functions concerned with the number of entries in a container (`size`, `cols`, `rows`, ...) consistently use `long int` as their return type. 

## Checklist

### DIFF
--- a/stan/math/opencl/prim/binomial_logit_glm_lpmf.hpp
+++ b/stan/math/opencl/prim/binomial_logit_glm_lpmf.hpp
@@ -20,6 +20,7 @@
 #include <stan/math/prim/fun/value_of_rec.hpp>
 
 #include <cmath>
+#include <cstdint>
 
 namespace stan {
 namespace math {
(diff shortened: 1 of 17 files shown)


andrjohns:
This should probably be `int64_t` instead of `long int`, since (naturally) a [`long int` on Windows is only 32-bit](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)

SteveBronder (approved):
Once tests pass this looks good to me

PR #3096, opened 2024-07-19
Improve finite_diff_auto with Eigen expressions

andrjohns:
## Summary

`finite_diff_gradient_auto` currently requires copying the input parameters and iteratively replacing values in the vector, requiring serial evaluation of the loop.

Rather than replacing values, we can instead use an Eigen expression to 'mask' a value at a specified index - avoiding the initial copy

## Tests

N/A - current tests should still pass

## Side Effects

N/A

## Release notes

Improved efficiency of `finite_diff_gradient_auto` 

## Checklist

### DIFF
--- a/stan/math/prim/functor/finite_diff_gradient_auto.hpp
+++ b/stan/math/prim/functor/finite_diff_gradient_auto.hpp
@@ -46,41 +46,30 @@ namespace math {
  * @param[out] fx function applied to argument
  * @param[out] grad_fx gradient of function at argument
  */
-template <typename F, typename VectorT,
+template <typename F, typename VectorT, typename GradVectorT,
           typename ScalarT = return_type_t<VectorT>>
-void finite_diff_gradient_auto(const F& f, const VectorT& x, ScalarT& fx,
-                               VectorT& grad_fx) {
-  VectorT x_temp(x);
+void finite_diff_gradient_auto(const F& f, VectorT&& x, ScalarT& fx,
+                               GradVectorT& grad_fx) {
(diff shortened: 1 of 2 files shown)


stan/math/prim/functor/finite_diff_gradient_auto.hpp
```diff
@@ -46,41 +46,30 @@ namespace math {
...
-
-    x_temp[i] = x[i];
-    grad_fx[i] = delta_f;
-  }
+    for (int j = 0; j < 6; ++j) {
+      decltype(auto) x_temp
+          = EigT::NullaryExpr(x.size(), [&x, &i, &h, &j](Eigen::Index k) {
+              return k == i ? x[i] + h * h_scale[j] : x[k];
+            });
+      delta_f += f(std::forward<decltype(x_temp)>(x_temp)) * mults[j];
```
SteveBronder, line 69: Why is this `decltype(auto)` instead of `auto`?

Also if it is being made here then it can be done using `std::move`

SteveBronder (changes requested):
Quick Q about move semantics but rest looks good!

stan/math/prim/functor/finite_diff_gradient_auto.hpp
andrjohns: Given that it was taking a universal/forwarding reference, I wasn't sure if `decltype(auto)` was need on the return to maintain any `reference`/`const`/etc status - but happy to drop if unnecessary!

SteveBronder (approved):

PR #3097, opened 2024-07-22
[WIP] Embedded Laplace Approximation

### LINKED ISSUE #3065: Integrated Laplace approximation
## Description
Provide basic support for an integrated Laplace approximation.

The motivation is to handle hierarchical models with a latent Gaussian model of the form
$\eta \sim p(\eta)$
$\theta \sim \text{normal}(0, K(\eta))$
$y \sim p(y \mid \theta, \eta)$.
We construct a Laplace approximation of $p(\theta \mid \eta, y)$ by matching the mode and curvature, and then obtain an approximation of the log marginal likelihood $\log p(y \mid \eta)$.

A call to the function may look as follows:
```
target += laplace_marginal(theta0, ll_fn, ..., K_fn, ...);
```
where
* `theta0` initializes the solver that finds the mode of $\log p(\theta \mid \eta, y)$
* `ll_fn` is a user specified function that returns the log likelihood $\log p(y \mid \eta, \theta)$.
* `...` is a first set of variadic arguments to be passed to `ll_fn`.
* `K_fn` is a user specified function that returns the prior covariance matrix.
* `...` is a second set of variadic arguments to be passed to `K_fn`.

We can also give the user control over the underlying Newton solver.
```
target += laplace_marginal_tol(theta0, tolerance, max_num_steps, hessian_block_size, solver, max_step_linesearch, ll_fn, ..., K_fn, ...);
```
The additional arguments are:
* `tolerance`: the difference between two successive evaluations of $\log p(\theta \mid \eta, y)$ after which convergence is considered to be achieved.
* `max_num_steps`: the number of steps after which the solver gives up.
* `hessian_block_size`: typically the Hessian of $\log p(y \mid \theta, \eta)$ with respect to $\theta$ is block diagonal and this can be taken advantage to speed up computation.
* `solver`: for log concave likelihoods, use `1`. In other cases, variations of the Newton solver can be attempted using `2` and `3`. See references for details.
* `max_step_linesearch`: number of linesearch steps at each Newton iteration. If set to 0, no linesearch happens.

Going the other way, we can also provide a simplified interface where the likelihood is not user specified but set ahead of time. 
```
target += laplace_marginal_poisson_log(theta0, *, K_fn, ...);
```
where `*` is a place-holder for arguments specific to the Poisson likelihood with a log link.

Next, we need to specify an rng function to recover draws from $p(\theta \mid \eta, y)$. In practice, we may also want "out-of-sample" draws, in which case we would evaluate $\theta$ for a different set of values passed to $K$ (think GP process). The call for the functions would be:
```
vector[nObs] theta = laplace_marginal_rng(theta0, ll_fn, ..., K_fn, ...);

vector[nObs_pred] theta_pred = laplace_marginal_pred_rng(theta0, ll_fn, ..., K_fn, ..., ...);
```
where for the second function, we pass in two sets of inputs for `K_fn`. As before, we can wrap these functions for a specific likelihood, and also give users control over the underlying Newton solver.

One thing to note is that under-the-hood, the autodiff uses higher-order derivatives and so the `fwd` mode must be supported.

There may be other features worth supporting, but for now, we can start here (and even split this into multiple PRs).

This issue replaces #755.

### References

* A detailed description of the algorithms: https://arxiv.org/abs/2306.14976
* A paper demonstrating applications: https://proceedings.neurips.cc/paper/2020/hash/673de96b04fa3adcae1aacda704217ef-Abstract.html
* A notebook showcasing existing prototype code: https://htmlpreview.github.io/?https://github.com/charlesm93/StanCon2020/blob/master/notebook-2022/lgm_stan.html#inst

#### Current Version:
v4.8.1

SteveBronder:
## Summary

Code for the embedded laplace approximation. The tests are all passing but there is a few things still needed

- [x] Update docs for all functions (@charlesm93 I added the basic arguments for a lot of the functions but could use your expertise to give the definitions for each argument)
- [x] Add corresponding rngs for each specialized laplace approximation (and tests)
- [x] Add tests for negative binomial

The file for laplace have been added to the `mix` folder since it uses higher order auto diff. 

The current signature for the generalized laplace looks like the following in C++

```c++
inline auto laplace_marginal_lpdf(LFun&& L_f, LArgs&& l_args,
                                  const Theta0& theta_0, CovarFun&& K_f,
                                  std::ostream* msgs, Args&&... args)
```

which will translate to stan like the following

```stan
target += laplace_marginal_tol_lpdf(
  likelihood_functor,  make_tuple(data_arg1, data_arg2), 
  eta, theta_init,
  covariance_function, 
  1e-6, 10, 2, 1, 5, // tuning args 
  covar_fun_arg1, covar_fun_arg2);
```

Note that the first tuple used for the likelihood arguments *must* be data.

Instead of using a tuple for the first functor's inputs and variadic arguments for the covariance functors arguments I would rather have them both be tuples like the following

```stan
target += laplace_marginal_tol_lpdf(
  likelihood_functor, make_tuple(data_arg1, data_arg2), 
  covariance_function,  make_tuple(covar_fun_arg1, covar_fun_arg2)
  theta_init,
  eta,
  1e-6, 10, 2, 1, 5);
```

I think this is nice because it makes it makes the tolerance parameters always sit at the end and both functors have the same input scheme for their arguments. Does anyone have thoughts on this

------------------------------

Other additions related to this PR

- A `filter_map` function that applies a conditionally applies a lambda `f` to each input of a tuple given a `type_trait` i.e. the following code would print "fp detected" twice and increment the `double` elements of the tuple by 1.

```c++
std::tuple<double, int, double> tup{1.0, 2, 3.0};
std::tuple<double, int, double> tup2 = filter_map<std::is_floating_point>([](auto x) { 
  std::cout << "fp detected" << std::endl;
  return x + 1;
  }, tup);
```

The test_ad suite now has a compile time option for only running the tests with only prim and reverse mode with a new boolean template parameter to `expect_ad`. This is needed to use laplace with the test framework as the laplace impl here does not work with higer order autodiff (since it needs higher order autodiff)

## Tests

Since the tests all seem very related I kept them in their own folder, is that alright? Or should I distribute them across the test folders like normal? While this PR is WIP I'm going to leave them in the same folder and if we don't want that then we can move them before we merge

```bash
./runTests.py -j20 ./test/unit/math/mix/laplace/ 
```

## Checklist

- [x] Copyright holder: Simon's Foundation


SteveBronder:
I'm closing this for now. Just needed the pr for looking at the diff

stan/math/mix/functor/laplace_likelihood.hpp
```diff
@@ -0,0 +1,370 @@
...
+  const Eigen::Index theta_size = theta.size();
+  const Eigen::Index eta_size = eta.size();
+  const Eigen::Index parm_size = theta_size + eta_size;
+  Matrix<var, Dynamic, 1> theta_var = theta;
+  Matrix<var, Eta::RowsAtCompileTime, Eta::ColsAtCompileTime> eta_var = eta;
+  int n_blocks = theta_size / hessian_block_size;
+  fvar<fvar<var>> target_ffvar = 0;
+  VectorXd v(theta_size);
+  VectorXd w(theta_size);
+  for (Eigen::Index i = 0; i < hessian_block_size; ++i) {
```
SteveBronder, line 231: Note for myself. I think for a large enough Hessian block size we could run this loop using a tbb parallel for loop

stan/math/mix/prob/laplace_poisson_log_exposure_rng.hpp
```diff
@@ -0,0 +1,129 @@
...
+ * @param train_tuple
+ * @param pred_tuple
+ * @param tolerance
+ * @param max_num_steps
+ * @param hessian_block_size
+ * @param solver
+ * @param max_steps_line_search
+ * @param rng
+ * @param msgs
+ * @param args
```
SteveBronder: @charlesm93 these need doc'd

stan/math/mix/prob/laplace_marginal_neg_binomial_2_log_lpmf.hpp
```diff
@@ -0,0 +1,276 @@
...
+#include <stan/math/fwd/fun/sum.hpp>
+#include <stan/math/prim/fun/binomial_coefficient_log.hpp>
+
+namespace stan {
+namespace math {
+
+struct neg_binomial_2_log_likelihood {
+  template <typename T_theta, typename T_eta>
+  inline return_type_t<T_theta, T_eta> operator()(
+      const Eigen::Matrix<T_theta, Eigen::Dynamic, 1>& theta, const T_eta& eta,
```
WardBrian: Ok, I've spent some time trying to get this to compile in my stanc3 branch and I think the issue is this is too restrictive/needs to be a template, in order to be able to pass `data` matrices here. I'm getting 

```
stan/lib/stan_math/stan/math/mix/functor/laplace_likelihood.hpp:32: error: no match for call to ‘(stan::math::neg_binomial_2_log_likelihood) (const Eigen::Map<Eigen::Matrix<double, -1, 1>, 0, Eigen::Stride<0, 0> >&, const Eigen::Map<Eigen::Matrix<double, -1, 1>, 0, Eigen::Stride<0, 0> >&, const std::vector<int>&, const std::vector<int>&, std::basic_ostream<char>*&)’
   32 |   return std::forward<F>(f)(std::forward<Theta>(theta),
      |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   33 |                             std::forward<Args>(args)...);
      | 
stan/lib/stan_math/stan/math/mix/prob/laplace_marginal_neg_binomial_2_log_lpmf.hpp:28: note: candidate: ‘template<class T_theta, class T_eta> stan::return_type_t<T1, T2> stan::math::neg_binomial_2_log_likelihood::operator()(const Eigen::Matrix<Type, -1, 1>&, const T_eta&, const std::vector<int>&, const std::vector<int>&, std::ostream*) const’
   28 |   inline return_type_t<T_theta, T_eta> operator()(
      | 
stan/lib/stan_math/stan/math/mix/prob/laplace_marginal_neg_binomial_2_log_lpmf.hpp:28: note:   template argument deduction/substitution failed:
stan/lib/stan_math/stan/math/mix/functor/laplace_likelihood.hpp:32: note:   ‘const Eigen::Map<Eigen::Matrix<double, -1, 1>, 0, Eigen::Stride<0, 0> >’ is not derived from ‘const Eigen::Matrix<Type, -1, 1>’
   32 |   return std::forward<F>(f)(std::forward<Theta>(theta),
      |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   33 |                             std::forward<Args>(args)...);
      | 
```

I think the same might be true for `y` in `bernoulli_logit_likelihood` above.

cc @SteveBronder

stan/math/rev/core/nested_rev_autodiff.hpp
WardBrian, line 1: This is causing compilation problems on older clangs (like what we use in CI) -- https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FStanc3/detail/PR-1513/18/pipeline/

It also looks like these changes are unused?

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1148 @@
...
+   */
+  int solver{1};
+  /* Maximum number of steps in line search*/
+  int max_steps_line_search{0};
+  /* iterations end when difference in objective function is less than tolerance
+   */
+  double tolerance{1e-12};
+  /* Maximum number of steps*/
+  int64_t max_num_steps{100};
+};
```
SteveBronder: @charlesm93 can you check these defaults are correct?

WardBrian:
`laplace_marginal_neg_binomial_log_summary_lpmf_test` took ~5 hours to run (not counting compilation) on the latest push. `laplace_marginal_poisson_log_lpmf_test` took even longer. I don't think we can (unconditionally, at least) run those kinds of test in our CI pipeline

SteveBronder:
Before I cut out a lot of tests I want to try building a separate jenkins node that just compiles these. At O3 these tests run decently quickly

stan/math/mix/functor/laplace_base_rng.hpp
```diff
@@ -0,0 +1,84 @@
...
+template <
+    typename LLFunc, typename LLArgs, typename ThetaVec, typename CovarFun,
+    typename CovarArgs, typename RNG, require_all_eigen_t<ThetaVec>* = nullptr,
+    require_t<is_all_arithmetic_scalar<CovarArgs, LLArgs, ThetaVec>>* = nullptr>
+inline Eigen::VectorXd laplace_base_rng(LLFunc&& ll_fun, LLArgs&& ll_args,
+                                        ThetaVec&& theta_0,
+                                        CovarFun&& covariance_function,
+                                        CovarArgs&& covar_args,
+                                        const laplace_options& options,
+                                        RNG& rng, std::ostream* msgs) {
```
SteveBronder, line 43: @charlesm93 check the math on this function to make sure we predict on the training data correctly.

WardBrian:
Changing the namespace of is_tuple will also require changes up in Stan. Not sure it's worth doing as part of this

SteveBronder:
Oh good call I'll put that back

Jenkinsfile
```diff
@@ -322,6 +322,31 @@ pipeline {
...
+                        }
+                    }
+                    when {
+                        expression {
+                            !skipRemainingStages
+                        }
+                    }
+                    steps {
+                        unstash 'MathSetup'
+                        sh "echo CXXFLAGS += -fsanitize=address -march=native -mtune=native >> make/local"
```
WardBrian: I would move -fsanitize=address to the if below that only runs on dev/master

stan/math/mix/functor/hessian_block_diag.hpp
WardBrian, line 1: Test for this function please!

test/unit/math/prim/functor/iter_tuple_n_test.cpp
WardBrian, line 1: Empty file -- missing check in?

stan/math/prim/functor/iter_tuple_n.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * @tparam Types types of arguments to `f`
+ * @param f functor to apply
+ * @param args arguments to apply `f` to. If `args` is a tuple or std::vector,
+ * this function will nest until it finds an Eigen type or scalar then apply `f`
+ * to that value.
+ * @return void, all arguments are passed by reference and this function will
+ * only create side effects.
+ */
+template <typename F, typename... Types>
+inline auto iter_tuple_n(F&& f, Types&&... args) {
```
WardBrian: I assume `n` is for nested, in which case I would prefer just calling it nested

WardBrian: Also, return type should be `void`

stan/math/prim/functor/iter_tuple_n.hpp
```diff
@@ -0,0 +1,44 @@
...
+  constexpr bool is_vec_container
+      = (is_std_vector_v<Types> && ...)
+        && (!is_stan_scalar<value_type_t<Types>>::value && ...);
+  if constexpr ((is_tuple_v<Types> && ...)) {
+    stan::math::for_each(
+        [&f](auto&&... args_i) {
+          iter_tuple_n(f, std::forward<decltype(args_i)>(args_i)...);
+        },
+        std::forward<Types>(args)...);
+  } else if constexpr (is_vec_container) {
```
WardBrian: Verify that this check is only triggering for vectors of tuples (e.g, not vectors of matrices).

stan/math/prim/functor/apply.hpp
```diff
@@ -4,10 +4,24 @@
...
+template <typename F, typename... TupleTypes, typename... Types>
+struct is_apply_nothrow<F, std::tuple<TupleTypes...>, Types...> {
+  static constexpr bool value
+      = (std::is_nothrow_invocable_v<std::decay_t<F>, Types..., TupleTypes...>);
+};
+
+template <typename F, typename T, typename... Types>
+inline constexpr bool is_apply_nothrow_v
+    = is_apply_nothrow<std::decay_t<F>, std::decay_t<T>, Types...>::value;
+
```
WardBrian: Delete

stan/math/prim/functor/apply.hpp
```diff
@@ -4,10 +4,24 @@
 #include <functional>
 #include <tuple>
 #include <utility>
-
+#include <iostream>
```
WardBrian: Delete

stan/math/prim/functor/apply.hpp
```diff
@@ -49,14 +62,19 @@ constexpr decltype(auto) apply_impl(F&& f, Tuple&& t,
...
       std::make_index_sequence<
           std::tuple_size<std::remove_reference_t<Tuple>>{}>{},
       std::forward<PreArgs>(pre_args)...);
 }
 
+template <typename F>
+inline constexpr std::tuple<> apply(F&& /* f */) noexcept {
+  return {};
+}
+
```
WardBrian: Delete? If not, comment

stan/math/prim/meta/contains_tuple.hpp
```diff
@@ -0,0 +1,63 @@
...
+}  // namespace internal
+
+/**
+ * Check if a tuple or type contains a tuple that passes the filter.
+ * @tparam Filter a struct that accepts one template parameter and has a static
+ *  constexpr bool member named value that is true if the type should be
+ *  included in the output tuple.
+ * @tparam T type to check
+ */
+template <template <typename...> class Filter, typename T>
```
WardBrian: Move these to filter_map.hpp

stan/math/prim/functor/filter_map.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace stan {
+namespace math {
+namespace internal {
+/**
+ * Filter a tuple and apply a functor to each element that passes the filter.
+ * @note The `Filter` will only check `T` and if `T` is a tuple, it will
+ * recursively check each element of the tuple. But it will not inspect into
+ * `std::vector` elements automatically. If you want to inspect the inner
+ * element of an `std::vector` your type trait must do that itself.
```
WardBrian: If we always need it to go into the vector, we can make that the default and this would be clearer.

stan/math/prim/functor/filter_map.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * filter.
+ */
+template <template <typename...> class Filter, bool InVector = false,
+          bool InTuple = false, typename F, typename T>
+inline constexpr decltype(auto) filter_map(F&& f, T&& x) {
+  if constexpr (inspect_tuple_v<Filter, T>) {
+    if constexpr (is_tuple_v<T>) {
+      auto ret = stan::math::apply(
+          [&f](auto&&... args) {
+            return stan::math::tuple_concat(filter_map<Filter, false, true>(
```
WardBrian, line 78: I think it's worth calling out the tuple_concat here to explain why the non-IsVector branches below always return tuples

stan/math/prim/functor/filter_map.hpp
```diff
@@ -0,0 +1,122 @@
...
+    return std::make_tuple();
+  }
+}
+}  // namespace internal
+/**
+ * Filter a tuple and apply a functor to each element that passes the filter.
+ * @note The `Filter` will only check `T` and if `T` is a tuple, it will
+ * recursively check each element of the tuple. But it will not inspect into
+ * `std::vector` elements automatically. If you want to inspect the inner
+ * element of an `std::vector` your type trait must do that itself.
```
WardBrian, line 141: I think it would really help to call out an example use case of this function in this comment

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1213 @@
...
+              - (R * (K_var.val() * s2)) * md_est.theta_grad.transpose();
+        var Z = make_callback_var(0.0, [K_var, K_adj_arena](auto&& vi) mutable {
+          K_var.adj().array() += vi.adj() * K_adj_arena.array();
+        });
+        grad(Z.vi_);
+        auto covar_args_filter = stan::math::filter_map<is_any_var_scalar>(
+            [](auto&& arg) -> decltype(auto) {
+              return std::forward<decltype(arg)>(arg);
+            },
+            covar_args_copy);
```
WardBrian: This style of call to filter map is repeated a couple places, I think it makes sense to pull it into a helper function

stan/math/prim/functor/map_if.hpp
```diff
@@ -0,0 +1,102 @@
...
+template <typename F, typename... Types>
+struct c_is_nothrow_invocable<false, F, Types...> : std::true_type {};
+template <typename F, typename... Types>
+struct c_is_nothrow_invocable<true, F, Types...>
+    : std::is_nothrow_invocable<F, Types...> {};
+
+template <bool FilterOn, typename F, typename... Types>
+inline constexpr bool c_is_nothrow_invocable_v
+    = c_is_nothrow_invocable<FilterOn, F, Types...>::value;
+
```
WardBrian: Delete

stan/math/prim/functor/partially_forward_as_tuple.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace internal {
+template <typename T>
+static constexpr bool nothrow_constructible_v
+    = std::is_nothrow_constructible_v<std::decay_t<T>, T&&>;
+template <typename... Types>
+static constexpr bool is_partial_forward_nothrow_constructible
+    = ((std::is_lvalue_reference_v<
+            Types&&> || nothrow_constructible_v<Types&&>)&&...);
+}  // namespace internal
```
WardBrian: Delete?

stan/math/prim/functor/partially_forward_as_tuple.hpp
```diff
@@ -0,0 +1,82 @@
...
+ * @param args The arguments to forward into the tuple.
+ * @return A tuple containing the forwarded arguments with types deduced via
+ * deduce_cvr_t.
+ *
+ * @note The function is declared constexpr. It is noexcept when, for each input
+ * type, the type value coming in is either an lvalue reference or an rvalue
+ * type whose move constructor is nothrow
+ */
+template <typename... Types>
+inline constexpr auto partially_forward_as_tuple(Types&&... args) {
```
WardBrian: Optional: Another name would be nice, maybe `materialize` or `copy` rather than forward as the verb.

stan/math/prim/meta/is_all_arithmetic.hpp
```diff
@@ -0,0 +1,38 @@
+#ifndef STAN_MATH_PRIM_META_IS_ALL_ARITHMETIC_HPP
+#define STAN_MATH_PRIM_META_IS_ALL_ARITHMETIC_HPP
+
+#include <stan/math/prim/meta/scalar_type.hpp>
+#include <type_traits>
+#include <tuple>
+
+namespace stan {
```
WardBrian, line 8: Rather than one specific is_all_*, want to make an stan::is_arithmetic? This would be useful in the compiler, where we currently generate `std::is_arithmetic<std::decay_t<T>>`

stan/math/rev/core/accumulate_adjoints.hpp
```diff
@@ -147,6 +147,15 @@ inline double* accumulate_adjoints(double* dest, Arith&& x, Pargs&&... args) {
...
 
+template <typename Tuple, require_tuple_t<Tuple>* = nullptr>
+inline double* accumulate_adjoints(double* dest, Tuple&& tup) {
+  return apply(
+      [dest](auto&&... args) {
+        return accumulate_adjoints(dest, std::forward<decltype(args)>(args)...);
+      },
+      std::forward<Tuple>(tup));
+}
+
```
WardBrian: Delete?

stan/math/rev/core/count_vars.hpp
```diff
@@ -34,7 +34,7 @@ template <typename Arith, require_arithmetic_t<scalar_type_t<Arith>>* = nullptr,
 inline size_t count_vars_impl(size_t count, Arith& x, Pargs&&... args);
 
 inline size_t count_vars_impl(size_t count);
-
+inline size_t count_vars_impl(size_t count, std::basic_ostream<char>*&);
```
WardBrian: Delete -- maybe revert whole file?

stan/math/mix/functor/laplace_likelihood.hpp
```diff
@@ -0,0 +1,450 @@
...
+ * @param args Additional variational arguments for likelihood function.
+ */
+template <typename F, typename Theta, typename Stream, typename... Args,
+          require_eigen_vector_t<Theta>* = nullptr>
+inline auto log_likelihood(F&& f, Theta&& theta, Stream* msgs, Args&&... args) {
+  return std::forward<F>(f)(std::forward<Theta>(theta),
+                            std::forward<Args>(args)..., msgs);
+}
+
+enum class COPY_TYPE { SHALLOW = 0, DEEP = 1 };
```
WardBrian, line 38: Docstring

stan/math/mix/functor/laplace_likelihood.hpp
```diff
@@ -0,0 +1,450 @@
...
+  nested_rev_autodiff nested;
+  const Eigen::Index theta_size = theta.size();
+  Matrix<var, Dynamic, 1> theta_var = std::forward<Theta>(theta);
+  Matrix<fvar<var>, Dynamic, 1> theta_fvar(theta_size);
+  for (Eigen::Index i = 0; i < theta_size; i++) {
+    theta_fvar(i) = fvar<var>(theta_var(i), v(i));
+  }
+  auto shallow_copy_args
+      = shallow_copy_vargs<fvar<var>>(std::forward_as_tuple(args...));
+  fvar<var> f_fvar = stan::math::apply(
```
WardBrian: Hard to read

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1213 @@
...
+   * using an LU decomposition (more general, but slower)
+   */
+  int solver{1};
+  /* Maximum number of steps in line search */
+  int max_steps_line_search{0};
+  /* iterations end when difference in objective function is less than tolerance
+   */
+  double tolerance{1e-6};
+  /* Maximum number of steps*/
+  int64_t max_num_steps{100};
```
WardBrian: aside: users won't be able to exceed 32 bit ints for this anyway

stan/math/prim/fun/Eigen.hpp
```diff
@@ -24,6 +33,7 @@
 #include <Eigen/QR>
 #include <Eigen/src/Core/NumTraits.h>
 #include <Eigen/SVD>
+#include <unsupported/Eigen/MatrixFunctions>
```
WardBrian: IMO we should include these unsupported headers only in the files that need them, like we do for e.g. FFT

stan/math/prim/fun/value_of_rec.hpp
```diff
@@ -124,11 +113,56 @@ inline auto value_of_rec(T&& M) {
...
  * @return Specified matrix.
  */
-template <typename T, typename = require_st_same<T, double>,
-          typename = require_eigen_t<T>>
-inline T value_of_rec(T&& x) {
-  return std::forward<T>(x);
+template <typename T, require_st_same<T, double>* = nullptr,
+          require_eigen_t<T>* = nullptr>
+inline decltype(auto) value_of_rec(T&& x) {
+  if constexpr (is_plain_type<T>::value || is_holder_v<T>) {
```
WardBrian, line 119: Also check if expression, evaluate or wrap in holder. Same for value_of()

stan/math/rev/fun/adjoint_of.hpp
WardBrian, line 1: Check that these changes are really necessary

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1213 @@
...
+            Eigen::Map<const Eigen::Matrix<var, -1, -1>> map_x(output_i.data(),
+                                                               output_i.size());
+            map_x.adj().setZero();
+          } else if constexpr (is_eigen_v<output_i_t>) {
+            output_i.adj().setZero();
+          } else if constexpr (is_stan_scalar_v<output_i_t>) {
+            output_i.adj() = 0;
+          } else {
+            static_assert(
+                1,
```
WardBrian: This should be `static_assert(0` (or `false`) -- check other usages as well.

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1213 @@
...
+          const double B_log_determinant
+              = 2.0 * llt_B.matrixLLT().diagonal().array().log().sum();
+          // Overwrite W instead of making a new sparse matrix
+          W.diagonal() = W_r;
+          return laplace_density_estimates{
+              objective_new - 0.5 * B_log_determinant,
+              std::move(covariance),
+              std::move(theta),
+              std::move(W),
+              std::move(Eigen::MatrixXd(L)),
```
WardBrian: No need to move a temporary

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1213 @@
...
+      // Check for convergence
+      if (abs(objective_new - objective_old) < options.tolerance) {
+        const double B_log_determinant
+            = 2.0 * llt_B.matrixLLT().diagonal().array().log().sum();
+        return laplace_density_estimates{
+            objective_new - 0.5 * B_log_determinant,
+            std::move(covariance),
+            std::move(theta),
+            std::move(W),
+            std::move(Eigen::MatrixXd(L)),
```
WardBrian, line 680: Same

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -0,0 +1,1213 @@
...
+  } else if constexpr (is_var<Input>::value) {
+    return static_cast<double>(0.0);
+  }
+}
+
+/**
+ * Helper function for printing out adjoints
+ */
+template <typename Output, require_t<is_any_var_scalar<Output>>* = nullptr>
+inline void print_adjoint(Output&& output) {
```
WardBrian: Move to test file

stan/math/mix/prob/laplace_marginal.hpp
```diff
@@ -0,0 +1,79 @@
...
+ */
+template <bool propto = false, typename LFun, typename LArgs, typename CovarFun,
+          typename ThetaVec, typename CovarArgs,
+          require_all_eigen_vector_t<ThetaVec>* = nullptr>
+inline auto laplace_marginal_tol(
+    LFun&& L_f, LArgs&& l_args, const ThetaVec& theta_0,
+    CovarFun&& covariance_function, CovarArgs&& covar_args, double tolerance,
+    int64_t max_num_steps, const int hessian_block_size, const int solver,
+    const int max_steps_line_search, std::ostream* msgs) {
+  // TEST: provisional signature to agree with parser.
```
WardBrian: Delete

stan/math/mix/prob/laplace_marginal_neg_binomial_2_log_lpmf.hpp
```diff
@@ -0,0 +1,202 @@
...
+    const ThetaVec& theta_0, CovarFun&& covariance_function,
+    CovarArgs&& covar_args, std::ostream* msgs) {
+  constexpr laplace_options ops{1, 1, 0, 1e-6, 100};
+  return laplace_marginal_density(
+      neg_binomial_2_log_likelihood{}, std::forward_as_tuple(eta, y, y_index),
+      theta_0, std::forward<CovarFun>(covariance_function),
+      std::forward<CovarArgs>(covar_args), ops, msgs);
+}
+
+struct neg_binomial_2_log_likelihood_summary {
```
WardBrian, line 111: If we want to keep this, @charlesm93 needs to add it to his doc PR and I need to add it to the compiler (both should be easy)

stan/math/mix/prob/laplace_marginal_poisson_log_exposure_lpmf.hpp
WardBrian, line 1: Should we have the filename and function match? `exposure` or `2`? @charlesm93

WardBrian (changes requested):
Bunch of comments from in-person review. Should be good to go once addressed

stan/math/mix/prob/laplace_marginal_poisson_log_exposure_lpmf.hpp
SteveBronder: I made both of these `2`

WardBrian:
Without ASAN, the laplace tests only took 16 minutes? Does that sound right? @SteveBronder

SteveBronder:
Yeah locally they take about 20 minutes

SteveBronder:
Let me give this one hard look on Sunday then Monday we merge and Tuesday we release!

WardBrian:
16 minutes -> 4 hours using ASAN is one of the worst slowdowns I’ve ever seen for a sanitizer run. Wow

WardBrian (approved):

SteveBronder:
@WardBrian merged! Tuesday lets merge the stanc3 pr then we are good to release!

PR #3099, opened 2024-07-30
Add sum_to_zero constraint, free, and check

WardBrian:
## Summary

This adds the functions necessary to implement a `sum_to_zero` constraint. This is a vector with `sum(x) == 0`. The constraining transform is to add an element set to `-sum(previous elements)`, and the freeing transform just takes the first N-1 elements out. 

Because this is linear, there is no jacobian adjustment

## Tests

Added tests in the err and constraint folders, based on the existing tests for `simplex`.

## Side Effects

None

## Release notes

Added `sum_to_zero_constrain`, `sum_to_zero_free`, and `check_sum_to_zero`.

## Checklist

### DIFF
--- a/stan/math/prim/constraint.hpp
+++ b/stan/math/prim/constraint.hpp
@@ -37,6 +37,8 @@
 #include <stan/math/prim/constraint/stochastic_column_free.hpp>
 #include <stan/math/prim/constraint/stochastic_row_constrain.hpp>
 #include <stan/math/prim/constraint/stochastic_row_free.hpp>
+#include <stan/math/prim/constraint/sum_to_zero_constrain.hpp>
+#include <stan/math/prim/constraint/sum_to_zero_free.hpp>
 #include <stan/math/prim/constraint/ub_constrain.hpp>
 #include <stan/math/prim/constraint/ub_free.hpp>
 #include <stan/math/prim/constraint/unit_vector_constrain.hpp>
--- a/stan/math/prim/constraint/sum_to_zero_constrain.hpp
+++ b/stan/math/prim/constraint/sum_to_zero_constrain.hpp
@@ -1,0 +1,106 @@
+#ifndef STAN_MATH_PRIM_CONSTRAINT_SUM_TO_ZERO_CONSTRAIN_HPP
+#define STAN_MATH_PRIM_CONSTRAINT_SUM_TO_ZERO_CONSTRAIN_HPP
+
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/prim/fun/sum.hpp>
+#include <cmath>
+
+namespace stan {
+namespace math {
+
+/**
+ * Return a vector with sum zero corresponding to the specified
+ * free vector.
+ *
+ * The sum-to-zero transform is defined such that the first K-1
+ * elements are unconstrained and the last element is the negative
+ * sum of those elements.
+ *
+ * @tparam Vec type of the vector
+ * @param y Free vector input of dimensionality K - 1.
+ * @return Zero-sum vector of dimensionality K.
+ */
+template <typename Vec, require_eigen_vector_t<Vec>* = nullptr>
+inline plain_type_t<Vec> sum_to_zero_constrain(const Vec& y) {
+  using T = value_type_t<Vec>;
+
+  int Km1 = y.size();
+  plain_type_t<Vec> x(Km1 + 1);
+  // copy the first Km1 elements
+  x.head(Km1) = y;
+  // set the last element to -sum(y)
+  x.coeffRef(Km1) = -sum(y);
+  return x;
+}
+
+/**
+ * Return a vector with sum zero corresponding to the specified
+ * free vector.
+ *
+ * The sum-to-zero transform is defined such that the first K-1
+ * elements are unconstrained and the last element is the negative
+ * sum of those elements. This is a linear transform, with no
+ * Jacobian.
+ *
+ * @tparam Vec type of the vector
+ * @param y Free vector input of dimensionality K - 1.
+ * @param lp unused
+ * @return Zero-sum vector of dimensionality K.
+ */
+template <typename Vec, require_eigen_vector_t<Vec>* = nullptr>
+inline plain_type_t<Vec> sum_to_zero_constrain(const Vec& y,
+                                               value_type_t<Vec>& lp) {
+  return sum_to_zero_constrain(y);
+}
+
+/**
+ * Return a vector with sum zero corresponding to the specified
+ * free vector.
+ *
+ * The sum-to-zero transform is defined such that the first K-1
+ * elements are unconstrained and the last element is the negative
+ * sum of those elements. This is a linear transform, with no
+ * Jacobian.
+ *
+ * @tparam Jacobian unused
+ * @tparam Vec A type inheriting from `Eigen::DenseBase` or a `var_value` with
+ *  inner type inheriting from `Eigen::DenseBase` with compile time dynamic rows
+ *  and 1 column
+ * @param[in] y free vector
+ * @param[in, out] lp unused
+ * @return Zero-sum vector of dimensionality one greater than `y`
+ */
+template <bool Jacobian, typename Vec, require_not_std_vector_t<Vec>* = nullptr>
+inline plain_type_t<Vec> sum_to_zero_constrain(const Vec& y,
+                                               return_type_t<Vec>& lp) {
+  return sum_to_zero_constrain(y);
+}
+
+/**
+ * Return a vector with sum zero corresponding to the specified
+ * free vector.
+ *
+ * The sum-to-zero transform is defined such that the first K-1
+ * elements are unconstrained and the last element is the negative
+ * sum of those elements. This is a linear transform, with no
+ * Jacobian.
+ *
+ * @tparam Jacobian unused
+ * @tparam Vec A standard vector with inner type inheriting from
+ * `Eigen::DenseBase` or a `var_value` with inner type inheriting from
+ * `Eigen::DenseBase` with compile time dynamic rows and 1 column
+ * @param[in] y free vector
+ * @param[in, out] lp unused
+ * @return Zero-sum vectors of dimensionality one greater than `y`
+ */
+template <bool Jacobian, typename T, require_std_vector_t<T>* = nullptr>
+inline auto sum_to_zero_constrain(const T& y, return_type_t<T>& lp) {
+  return apply_vector_unary<T>::apply(
+      y, [](auto&& v) { return sum_to_zero_constrain(v); });
+}
+
+}  // namespace math
+}  // namespace stan
+
+#endif
--- a/stan/math/prim/constraint/sum_to_zero_constrain.hpp
+++ b/stan/math/prim/constraint/sum_to_zero_constrain.hpp
@@ -4,6 +4,7 @@
 #include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/prim/fun/sum.hpp>
+#include <stan/math/prim/functor/apply_vector_unary.hpp>
 #include <cmath>
 
 namespace stan {
@@ -23,8 +24,6 @@ namespace math {
  */
 template <typename Vec, require_eigen_vector_t<Vec>* = nullptr>
 inline plain_type_t<Vec> sum_to_zero_constrain(const Vec& y) {
-  using T = value_type_t<Vec>;
-
   int Km1 = y.size();
   plain_type_t<Vec> x(Km1 + 1);
   // copy the first Km1 elements
--- a/stan/math/prim/constraint/sum_to_zero_constrain.hpp
+++ b/stan/math/prim/constraint/sum_to_zero_constrain.hpp
@@ -26,6 +26,9 @@ template <typename Vec, require_eigen_col_vector_t<Vec>* = nullptr,
           require_not_st_var<Vec>* = nullptr>
 inline plain_type_t<Vec> sum_to_zero_constrain(const Vec& y) {
   const auto Km1 = y.size();
+  if (unlikely(Km1 == 0)) {
+    return plain_type_t<Vec>(Eigen::VectorXd{{0}});
+  }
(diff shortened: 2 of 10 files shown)


spinkney:
@WardBrian it is perhaps better if we implement the transform from https://discourse.mc-stan.org/t/test-soft-vs-hard-sum-to-zero-constrain-choosing-the-right-prior-for-soft-constrain/3884/31?u=spinkney? It's unclear if this is better than what is in this pr from the post but it does give the math to get the correct uniform marginals.

WardBrian:
@spinkney -- this just implements the transform @bob-carpenter gave me, I'm not qualified to access it versus an alternative

SteveBronder (changes requested):
few small comments but generally everything looks good

WardBrian:
@SteveBronder all addressed. I'd like to wait on the mathematical discussion above before actually merging

bob-carpenter:
I commented on the stanc3 issue and will repeat myself here:  using the QR or SVD-based decompositions break our current abstraction that just treats the constraining and unconstraining functions as standalones that are not data-dependent.  It wouldn't be absolutely impossible to shoehorn something with data into our current framework, but it'd be a lot of work.

spinkney:
> I commented on the stanc3 issue and will repeat myself here: using the QR or SVD-based decompositions break our current abstraction that just treats the constraining and unconstraining functions as standalones that are not data-dependent. It wouldn't be absolutely impossible to shoehorn something with data into our current framework, but it'd be a lot of work.

We don't need to have a data input. The only thing that needs to be known is the size of the vector, which we already get. This is similar to what is done in the [`ilr` transform](https://github.com/bob-carpenter/transforms/blob/main/simplex_transforms/stan/transforms/ILR/ILR_functions.stan). In fact, that transform is nearly the what we want since a log of a simplex sums to zero. The issue is that this is on the positive orthant. I believe a simple modification of the `ilr` simplex could give us the sum-to-zero vector we want. 

Is there an issue with making a `linspaced_vector` within the transform?

WardBrian:
Looking at the forum thread again, the thing being computed is relatively cheap. The memory overhead (2*N for autodiff compared to this method) would probably be the bigger concern with such a scheme.

The specific transform used is technically an implementation detail, so there's nothing holding us to one choice going forward if we wanted to proceed with the feature with this transform for now

spinkney:
I believe this is what I'm suggesting. It's a bit concerning that summing all the values isn't exactly 0 but I believe that's a finite precision thing. 

```stan
functions {
  vector inv_ilr_sum_to_zero_constrain_lp(vector y) {
  int N = rows(y) + 1;
  vector[N - 1] ns = linspaced_vector(N - 1, 1, N - 1);
  vector[N - 1] w = y ./ sqrt(ns .* (ns + 1));
  vector[N] z = append_row(reverse(cumulative_sum(reverse(w))), 0) - append_row(0, ns .* w);
  return z;
}
}
data {
  int<lower=0> N;
}
parameters {
  vector[N - 1] y;
}
transformed parameters {
   vector[N] x = inv_ilr_sum_to_zero_constrain_lp(y);
}
model {
  x ~ normal(0, 10);
}
generated quantities {
  real sum_x = sum(x);
  vector[N] alpha = softmax(x);
}
```

```
  variable     mean   median    sd   mad    q5   q95  rhat ess_bulk ess_tail
  <chr>       <dbl>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>    <dbl>    <dbl>
1 y[1]     -0.0819  -0.0453  10.0  10.1  -16.5  16.3  1.00   59376.   30953.
2 y[2]     -0.00142 -0.0227  10.0   9.95 -16.4  16.5  1.00   60247.   29833.
3 y[3]      0.0466   0.0560   9.94  9.90 -16.3  16.4  1.00   57789.   29357.
4 y[4]      0.0276   0.0259   9.95  9.89 -16.3  16.4  1.00   58617.   30486.
5 y[5]     -0.0619  -0.0333  10.0   9.98 -16.5  16.4  1.00   59373.   30665.
6 y[6]      0.0395  -0.00871  9.98  9.95 -16.4  16.5  1.00   61190.   31503.
7 y[7]     -0.0313   0.0543   9.95  9.97 -16.4  16.4  1.00   57335.   30404.

1 x[1]     -0.0482 -0.0709   9.37  9.34 -15.4  15.3  1.00   56797.   29396.
2 x[2]      0.0676  0.0499   9.33  9.32 -15.3  15.4  1.00   61234.   32163.
3 x[3]      0.0114  0.0165   9.36  9.42 -15.4  15.5  1.00   60462.   31845.
4 x[4]     -0.0436 -0.0135   9.30  9.23 -15.4  15.2  1.00   58376.   30831.
5 x[5]     -0.0340 -0.0618   9.31  9.31 -15.4  15.3  1.00   58947.   29524.
6 x[6]      0.0584  0.0153   9.38  9.33 -15.3  15.5  1.00   59096.   31725.
7 x[7]     -0.0408  0.00275  9.34  9.30 -15.4  15.2  1.00   60257.   31300.
8 x[8]      0.0293 -0.0508   9.31  9.32 -15.3  15.4  1.00   57335.   30404.

1 sum_x    -9.94e-18      0 2.67e-15 2.63e-15 -3.55e-15 3.55e-15  1.00   39253.   36969.
```

WardBrian:
@spinkney when freeing, we only assume a tolerance of 1e-8, so being around 1e-18 is definitely "close enough" to zero if it's just floating point issues.

Perhaps a naive question, but shouldn't the `x`s have standard deviation 10, not the `y`s? The original forum post had a correction (similar to the unit vector, I think)

bob-carpenter:
Thanks so much, @spinkney.

> It's a bit concerning that summing all the values isn't exactly 0.

No worries---you're right at double machine precision there: 1e-18.

Brian was asking about a Jacobian adjustment, but we don't need one here---this is a linear transform on this scale (the `ns` are constant and `cumulative_sum`, rearrangement, and subtraction are linear).  

We can write out the constraining transform as a function in the math library in C++ so that we can write custom autodiff for it to make it efficient.  In C++, it can be written without allocating the linearly spaced vector---it'll be an expression template or loop.

> since a log of a simplex sums to zero

I initially read that the wrong way.  We have `log(sum(simplex)) = 0`, but  `sum(log(simplex)) < 0` if the simplex is more than zero-dimensional (i.e., has more than one element).

spinkney:
> Perhaps a naive question, but shouldn't the `x`s have standard deviation 10, not the `y`s? The original forum post had a correction (similar to the unit vector, I think)
Yes! There is one potential issue and that is that in the forum post they put the prior on the unconstrained values. You're right that we can mimic this like in the `unit_vector` case.

```stan
functions {
  vector inv_ilr_sum_to_zero_constrain_lp(vector y) {
  int N = rows(y) + 1;
  vector[N - 1] ns = linspaced_vector(N - 1, 1, N - 1);
  vector[N - 1] w = y .* inv_sqrt(ns .* (ns + 1));
  vector[N] z = append_row(reverse(cumulative_sum(reverse(w))), 0) - append_row(0, ns .* w);
  
  target += normal_lpdf(y | 0, inv_sqrt(1 - inv(N)));
  
  return z;
}
}
data {
  int<lower=0> N;
}
parameters {
  vector[N - 1] y;
  real z;
}
transformed parameters {
   vector[N] x = inv_ilr_sum_to_zero_constrain_lp(y);
}
model {
  z ~ std_normal();
}
generated quantities {
  real sum_x = sum(x);
}
```

We see that the margins of `x` match that of a standard normal (`z` for reference).
```
  variable        mean     median    sd   mad    q5   q95  rhat ess_bulk ess_tail
   <chr>          <dbl>      <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>    <dbl>    <dbl>
 1 x[1]      0.00214     0.00123   0.998 0.997 -1.64  1.64  1.00  188253.   57424.
 2 x[2]     -0.00227    -0.0000932 1.00  1.00  -1.65  1.64  1.00  193341.   61674.
 3 x[3]     -0.000257   -0.00233   0.994 0.992 -1.63  1.64  1.00  194002.   59248.
 4 x[4]      0.000861    0.00334   1.01  1.01  -1.65  1.65  1.00  198643.   56607.
 5 x[5]      0.00271     0.00272   0.994 0.996 -1.63  1.63  1.00  182120.   59134.
 6 x[6]      0.00328     0.00399   1.00  1.00  -1.65  1.66  1.00  182355.   58314.
 7 x[7]     -0.00254    -0.00450   1.00  1.00  -1.64  1.65  1.00  203890.   60267.
 8 x[8]      0.00000966  0.00172   1.00  1.00  -1.65  1.65  1.00  198791.   58403.
 9 x[9]     -0.000147    0.00395   0.999 0.998 -1.65  1.65  1.00  193653.   60290.
10 x[10]     0.000846    0.00319   0.996 0.995 -1.64  1.65  1.00  195193.   57673.
11 x[11]    -0.000333    0.000160  0.997 0.993 -1.63  1.64  1.00  196003.   60003.
12 x[12]     0.00346     0.00541   1.00  1.01  -1.65  1.65  1.00  193099.   58743.
13 x[13]    -0.00376    -0.00444   0.999 0.999 -1.66  1.63  1.00  199063.   61333.
14 x[14]    -0.00325    -0.0000660 1.00  1.01  -1.66  1.65  1.00  197189.   56415.
15 x[15]    -0.000674   -0.00130   0.999 0.994 -1.64  1.64  1.00  198973.   54759.
16 x[16]    -0.0000623  -0.00202   1.01  1.02  -1.66  1.66  1.00  184838.   59584.
17 z         0.00259     0.00135   1.00  1.01  -1.64  1.65  1.00  188801.   59141.
```
The issue with the above is that a standard normal is fairly constrained. I think it would be better if we did something like a normal(0, 10) and then the user can put a more restrictive prior on `x`. However, `x` will not be marginally distributed exactly as whatever the user puts because we have 2 priors - the `normal(0, 10)` one that's in the transform and the one the user specifies.

We can make this more flexible by using the `multiplier` keyword to lengthen or shorten the variance of the normal. First we "back out" the `inv_sqrt(1 - inv(N))` standard error. Because this is a normal distribution with mean = 0 

$$
\log f(x \mid 0, \sigma) = -0.5 (\log(2) + \log(\pi) + \sigma) - \frac{x^2}{2 \sigma^2}
$$

where $\sigma = \frac{1}{\sqrt{1 - \frac{1}{N}}}$. Since $\sigma$ is just a constant we can forget the normalizing constant and focus on the second part:

$$
\frac{-x^2}{2 \sigma^2} = -0.5 x^2 (1 - \frac{1}{N}) = -0.5 x^2 \frac{N}{N - 1}.
$$

Since this is just a normal distribution the multiplier divides that expression above by the variance of the normal that one wants, i.e.,

```stan
functions {
  vector inv_ilr_sum_to_zero_constrain_lp(vector y, real val) {
  int N = rows(y) + 1;
  vector[N - 1] ns = linspaced_vector(N - 1, 1, N - 1);
  vector[N - 1] w = y .* inv_sqrt(ns .* (ns + 1));
  vector[N] z = append_row(reverse(cumulative_sum(reverse(w))), 0) - append_row(0, ns .* w);
  
  target += -0.5 * z^2 * (N - 1) * inv(N) / val^2; 

  return z;
}
}
data {
  int<lower=0> N;
  real<lower=0> val;
}
parameters {
  vector[N - 1] y;
  real z;
}
transformed parameters {
   vector[N] x = inv_ilr_sum_to_zero_constrain_lp(y, val);
}
model {
//  x ~ normal(0, val);
  z ~ normal(0, val);
}
generated quantities {
  real sum_x = sum(x);
}
```
I believe this should work with vectors in the multiplier so that one could put different standard deviations on each element of the zero sum vector.

bob-carpenter:
I find @spinkney's constraining transform easier to understand with `N = rows(y)` than `N = rows(y) + 1`.  I also unfolded the code to define some intermediates to work through values.  

```stan
vector inv_ilr_sum_to_zero_constrain_lp(vector y) {
  int N = rows(y);
  vector[N] ns = linspaced_vector(N, 1, N); 
  vector[N] w = y ./ sqrt(ns .* (ns + 1)); 
  vector[N] u = reverse(cumulative_sum(reverse(w)));
  vector[N] v = ns .* w;
  vector[N + 1] z = append_row(u, 0) - append_row(0, v);
  return z; 
}
```

Then we need to derive the unconstraining transform from `z[1:N+1]` to `y[1:N]`.   From the above code, we can derive the following relations.

1. `ns[n] = n`
2. `w[n] = y[n] / sqrt(n * (n + 1))`
3. `u[n] = w[n] + w[n + 1] + ... + w[N]`
4. `v[n] = n * y[n] / sqrt(n * (n + 1)) = y[n] * sqrt(n / (n + 1.0))`
5.  `z[1] = u[1]`;  &nbsp;  `z[N + 1] = v[N]`; &nbsp; `z[n] = u[n] + v[n - 1]` for `2 <= n <= N`

Going in the reverse direction, we know

A. `y[n] = w[n] * sqrt((n + 1.0) / n)` and
B.  `w[n] = v[n] * n`, thus
C. `y[n] = v[n] * sqrt(n * (n + 1))`
D. `v[n - 1] = z[n] - u[n]`


Starting with `n = N`, we have the base case

```
v[N] = z[N + 1]
y[N] = v[N] * sqrt(n * (n + 1))
```

and thus

```
y[N] = z[N + 1] * sqrt(n * (n + 1))
```

For the inductive case, 

```
v[n - 1] = z[n] - u[n]
```

where `u[n] = w[n] + ... + w[N]` and `w[n] = v[n] * n`, so that



and thus

```
v[n - 1] = z[n] - u[n]
```

where 

```
u[n] = v[n] * n + ... + v[N] * N
```

and hence

```
v[n-1] = z[n] - (v[n] * n + ... + v[N] * N)
y[n - 1] = v[n - 1] * sqrt(n * (n + 1))
```

Whew.  That's still not code, though---the code will have to efficiently keep the running `v[n] * n` totals.

WardBrian:
@bob-carpenter I think there's either a mistake in your code or I'm missing something in translation. 

> Starting with `n = N`, we have the base case
> 
> ```
> v[N] = z[N + 1]
> y[N] = v[N] * sqrt(n * (n + 1))
> ```

The final element seems to be too big by a factor of `-N`, and the rest are then incorrect as a result.

spinkney:
You need $Y_n = Z_{n + 1} \sqrt{n(n+1)} / n$

Issue #3006, opened 2024-01-19
stan::math and std::complex


WardBrian:
## Description

Currently, Stan's complex number support is entirely built on `std::complex`, including autodiff, which
uses `std::complex<stan::math::var>`.

This is, unfortunately, unspecified behavior in the [C++ spec [26.4.2]](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf):

> The effect of instantiating the template `complex` for any type other than `float`, `double`, or `long double` is unspecified. The specializations `complex<float>`, `complex<double>`, and `complex<long double>` are literal types

For a reminder on what ["unspecified behavior" means](https://en.cppreference.com/w/cpp/language/ub):

> unspecified behavior - the behavior of the program varies between implementations, and the conforming implementation is not required to document the effects of each behavior. Each unspecified behavior results in one of a set of valid results. 

Essentially, unspecified behavior is the same as "implementation-defined behavior" but without the requirement that implementations document what they are doing. This is also often taken to mean **there are no backwards compatibility guarantees on any specific unspecified behavior**.

## Problems

Recent versions of `clang`/`libstdc++` have made changes which they are fully within their rights to do by the spec, but have broken Stan builds.

- In libstdc++16, they changed the definition of `log(complex)` from ` complex<T>(log(abs(x)), arg(x));` to ` complex<T>(std::log(std::abs(x)), std::arg(x));`. This broke argument dependent lookup for this function.
  A similar change broke `operator*` for our complex types.
  This lead to https://github.com/stan-dev/cmdstan/issues/1158, which was the reason we needed a 2.32.1 release.
  @andrjohns provided the fix in https://github.com/stan-dev/math/pull/2892
- In libstdc++17, a similar change was made to `fabs`, which necessitated to https://github.com/stan-dev/math/pull/2991

## What to do

This is less clear to me. 

### Option 1 - walk on egg shells

So far, all of the issues that have arisen from this have been due to argument dependent lookup breaking for these types. We can fix that by being much more explicit, as we did in https://github.com/stan-dev/math/pull/2892 and https://github.com/stan-dev/math/pull/2991. This requires auditing the existing usages, which probably requires a fair amount of C++ expertise to understand how the calls are being resolved.

### Option 2 - our own type

[...]

PR #3122, opened 2024-11-08
Encapsulate std::complex functions

SteveBronder:
## Summary

For the last few releases Stan math has had to do quick fixes when clang would update their functions for 'std::complex' which can cause conflicts when looking for our own functions that work over complex. The main issue for function calls like the following where an unqualified name lookup is used to find the correct function to call.

```c++
struct sin_fun {
  template <typename T>
  static inline auto fun(const T& x) {
    using std::sin;
    return sin(x);
  }
};
```

In the above, `sin` is looked for in both the `stan::math` namespace and the `std` namespace since [unqualified name lookup](https://en.cppreference.com/w/cpp/language/unqualified_lookup) also uses [Argument Dependent Lookup](https://en.cppreference.com/w/cpp/language/adl) (ADL). During ADL the compiler looks at the namespaces of the arguments as well as the local namespace to figure out all of the valid function signatures it should use when deciding which function is the best candidate. Since we override `std::complex` the `std` namespace is then also included in the valid candidates. Sometimes this is fine because the clang standard library uses `requires` very similar to ours for only supporting floating point types inside of their complex functions.

```c++
template <class _Tp, __enable_if_t<is_floating_point<_Tp>::value, int> >
_LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX20 complex<_Tp>
operator*(const complex<_Tp>& __lhs, const complex<_Tp>& __rhs) {
  return complex<_Tp>(__from_builtin_tag(), __lhs.__builtin() * __rhs.__builtin());
}
```

But some functions, like the one below (linked [here](https://github.com/llvm/llvm-project/blob/main/libcxx/include/complex#L984)), do not restrict themselves to only floating point types.

```c++
template <class _Tp>
inline _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX20 complex<_Tp> conj(const complex<_Tp>& __c) {
  return complex<_Tp>(__c.real(), -__c.imag());
}
```

This causes issues because ADL then finds two valid candidates functions when trying to find the best candidate for `conj` (both the `stan::math` version and the `std` version).

There is no good solution to this issue, it's just how the language is defined. The only way to make `stan::math` functions a better candidate is to use templates to make the stan math function both more generic and have an extra template with our `requires` which makes it slightly better so the compiler chooses it as the best candidate.

So this PR does a few things.

1. Use templates for all of the std functions that we want to override for complex types. The generic template and parameters makes our functions a better candidate. 
    - This is very much a patch solution. In the long run we need our own custom complex scalar type. But we can't do that yet till an issue is resolved in eigen (see [here](https://gitlab.com/libeigen/eigen/-/issues/2865)) where they assume users are overriding `std::complex` for their complex types.
2. Cleans up the vectorized functions signatures. There were 4 or 5 different requires being used and they made the function rather confusing. Most of the vectorized functions work over containers containing base types of `autodiff` so we can just use one requires instead of the several that were there
3. Fixes the headers in rev so that `rev` headers are always used first and if available. This stops issues with prim being brought in before their respective `rev` or `fwd` functions are available and fixes soem header include order issues.

## Tests

No changes to tests

## Side Effects

Hopefully not

## Release notes

Template complex functions in the Stan math library to be better candidates than the standard library functions during ADL.

## Checklist

### DIFF
--- a/stan/math/rev/fun/generalized_inverse.hpp
+++ b/stan/math/rev/fun/generalized_inverse.hpp
@@ -6,7 +6,6 @@
 #include <stan/math/rev/fun/inverse.hpp>
 #include <stan/math/prim/err.hpp>
 #include <stan/math/prim/fun/add_diag.hpp>
-#include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/prim/fun/generalized_inverse.hpp>
 
 namespace stan {
--- a/stan/math/rev/fun/softmax.hpp
+++ b/stan/math/rev/fun/softmax.hpp
@@ -6,7 +6,6 @@
 #include <stan/math/rev/fun/value_of.hpp>
 #include <stan/math/rev/core/reverse_pass_callback.hpp>
 #include <stan/math/rev/core/arena_matrix.hpp>
-#include <stan/math/rev/fun/value_of.hpp>
 #include <stan/math/prim/fun/typedefs.hpp>
 #include <stan/math/prim/fun/to_ref.hpp>
 #include <stan/math/prim/fun/softmax.hpp>


SteveBronder:
Also after this release I'm going to write a script that we can use in jenkins to check that includes are being done correctly

WardBrian:
Tagging #3006 just so this PR shows up in the history of that issue -- as you say, actually resolving it would require Eigen to make some changes first

WardBrian:
Running with latest compilers: https://jenkins.flatironinstitute.org/blue/organizations/jenkins/Stan%2FBleedingEdgeCompilersMonthly/detail/BleedingEdgeCompilersMonthly/235/pipeline

WardBrian (changes requested):
A bunch of tiny questions/comments -- thanks for tackling this!

stan/math/fwd/fun/conj.hpp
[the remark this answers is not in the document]
SteveBronder: Yes I'm going to write that script to make a file for each file in prim soon for fwd and rev so it's just easier to leave this here for now

stan/math/fwd/fun/atan2.hpp
```diff
@@ -3,30 +3,27 @@
 
 #include <stan/math/fwd/meta.hpp>
 #include <stan/math/fwd/core.hpp>
-#include <stan/math/prim/fun/square.hpp>
+#include <stan/math/fwd/fun/square.hpp>
```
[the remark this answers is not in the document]
SteveBronder: Good catch!

stan/math/prim/fun/asinh.hpp
```diff
@@ -19,6 +19,29 @@
...
+/**
+ * Return the hyperbolic arc sine of the complex argument.
+ *
+ * @tparam V a complex type with inner arithmetic type.
+ * @param[in] x argument
+ * @return hyperbolic arc sine of the argument
+ */
+template <typename T, require_complex_bt<std::is_arithmetic, T>* = nullptr>
+inline auto asinh(const T x) {
+  return std::asinh(x);
```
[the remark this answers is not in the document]
SteveBronder: This version, with `require_complex_bt`, can only be called for `std::complex<Arithmetic>` types. So no need for autodiff here

stan/math/fwd/fun/hypergeometric_1F0.hpp
```diff
@@ -1,9 +1,12 @@
 #ifndef STAN_MATH_FWD_FUN_HYPERGEOMETRIC_1F0_HPP
 #define STAN_MATH_FWD_FUN_HYPERGEOMETRIC_1F0_HPP
 
-#include <stan/math/prim/meta.hpp>
-#include <stan/math/prim/fun/hypergeometric_1F0.hpp>
 #include <stan/math/fwd/core.hpp>
+#include <stan/math/fwd/meta.hpp>
+#include <stan/math/rev/fun/log1m.hpp>
+#include <stan/math/rev/fun/inv.hpp>
+#include <stan/math/rev/fun/value_of.hpp>
```
WardBrian, line 8: Did you mean for these to be from `rev`? I assume this is a typo and should be `fwd`

stan/math/fwd/fun/inv_sqrt.hpp
```diff
@@ -3,14 +3,14 @@
...
 #include <stan/math/fwd/core.hpp>
+#include <stan/math/prim/fun/inv_sqrt.hpp>
 #include <cmath>
 
 namespace stan {
 namespace math {
 
 template <typename T>
 inline fvar<T> inv_sqrt(const fvar<T>& x) {
-  using std::sqrt;
```
WardBrian, line 13: I think this file should include sqrt? or the computation should call inv_sqrt on x.val, not use sqrt?

stan/math/fwd/fun/log_mix.hpp
```diff
@@ -28,7 +28,6 @@ template <typename T_theta, typename T_lambda1, typename T_lambda2, int N>
 inline void log_mix_partial_helper(
     const T_theta& theta, const T_lambda1& lambda1, const T_lambda2& lambda2,
     promote_args_t<T_theta, T_lambda1, T_lambda2> (&partials_array)[N]) {
-  using std::exp;
```
WardBrian, line 31: This function actually calls exp, so I think we should include it from fwd?

stan/math/fwd/fun/tcrossprod.hpp
```diff
@@ -1,11 +1,12 @@
...
 #define STAN_MATH_FWD_FUN_TCROSSPROD_HPP
 
-#include <stan/math/prim/meta.hpp>
 #include <stan/math/prim/fun/Eigen.hpp>
+#include <stan/math/fwd/meta.hpp>
+#include <stan/math/fwd/fun/multiply.hpp>
 #include <stan/math/prim/fun/to_ref.hpp>
 #include <stan/math/prim/fun/transpose.hpp>
-#include <stan/math/fwd/fun/multiply.hpp>
+#include <stan/math/prim/fun/multiply.hpp>
```
WardBrian, line 9: Was this meant to be a deletion?

WardBrian (changes requested):
A few small things in the new fwd changes, plus the above comments on doc seem to be outstanding

WardBrian:
@SteveBronder I addressed all my own docstring comments, so it's just the ~6 things above that are unresolved

stan/math/fwd/fun/tcrossprod.hpp
SteveBronder: No sorry we are supposed to use `multiply` in this function.

multiply returns a scalar for `multiply(vector, row_vector)`. Should this function also return a scalar in the case of a vector?

stan/math/fwd/fun/log_mix.hpp
SteveBronder: 👍

stan/math/fwd/fun/inv_sqrt.hpp
SteveBronder: Included `fwd/fun/sqrt.hpp`

stan/math/fwd/fun/tcrossprod.hpp
SteveBronder: ^yes. What about the second part of my Q. I think it should always return a matrix

stan/math/fwd/fun/hypergeometric_1F0.hpp
SteveBronder: 👍 typo

stan/math/rev/fun/hypergeometric_pFq.hpp
```diff
@@ -3,7 +3,7 @@
 
 #include <stan/math/prim/meta.hpp>
```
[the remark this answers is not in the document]
SteveBronder: need to be `/rev/meta.hpp`

stan/math/rev/fun/exp.hpp
```diff
@@ -3,14 +3,13 @@
...
-#include <stan/math/prim/fun/exp.hpp>
+#include <stan/math/rev/fun/cos.hpp>
 #include <stan/math/rev/fun/is_inf.hpp>
-#include <stan/math/prim/fun/isinf.hpp>
-#include <stan/math/prim/fun/isfinite.hpp>
 #include <stan/math/rev/fun/is_nan.hpp>
-#include <stan/math/rev/fun/cos.hpp>
 #include <stan/math/rev/fun/sin.hpp>
+#include <stan/math/prim/fun/isinf.hpp>
+#include <stan/math/prim/fun/isfinite.hpp>
```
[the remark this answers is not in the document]
SteveBronder: typo

stan/math/rev/fun/cos.hpp
```diff
@@ -1,15 +1,14 @@
...
-#include <stan/math/prim/fun/cos.hpp>
-#include <stan/math/prim/fun/isinf.hpp>
-#include <stan/math/prim/fun/isfinite.hpp>
 #include <stan/math/rev/core.hpp>
 #include <stan/math/rev/meta.hpp>
 #include <stan/math/rev/fun/abs.hpp>
 #include <stan/math/rev/fun/cosh.hpp>
 #include <stan/math/rev/fun/sinh.hpp>
+#include <stan/math/prim/fun/isinf.hpp>
+#include <stan/math/prim/fun/isfinite.hpp>
```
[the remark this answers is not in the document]
SteveBronder: removed

stan/math/rev/fun/columns_dot_product.hpp
```diff
@@ -2,12 +2,12 @@
...
 
 #include <stan/math/rev/meta.hpp>
-#include <stan/math/prim/err.hpp>
-#include <stan/math/prim/fun/Eigen.hpp>
 #include <stan/math/rev/core.hpp>
 #include <stan/math/rev/core/typedefs.hpp>
 #include <stan/math/rev/fun/dot_product.hpp>
 #include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/err.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
```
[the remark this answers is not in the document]
SteveBronder: fixed

stan/math/rev/fun/asinh.hpp
```diff
@@ -17,6 +12,11 @@
 #include <stan/math/rev/fun/log.hpp>
 #include <stan/math/rev/fun/polar.hpp>
 #include <stan/math/rev/fun/sqrt.hpp>
+#include <stan/math/prim/core.hpp>
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/fun/asinh.hpp>
+#include <stan/math/prim/fun/isinf.hpp>
+#include <stan/math/prim/fun/is_inf.hpp>
```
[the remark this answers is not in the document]
SteveBronder: 👍

stan/math/rev/fun/atan.hpp
```diff
@@ -15,6 +11,9 @@
 #include <stan/math/rev/fun/sqrt.hpp>
 #include <stan/math/rev/fun/is_inf.hpp>
 #include <stan/math/rev/fun/is_nan.hpp>
+#include <stan/math/prim/fun/isinf.hpp>
+#include <stan/math/prim/fun/isnan.hpp>
```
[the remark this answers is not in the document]
SteveBronder: 👍

stan/math/fwd/fun/log_mix.hpp
SteveBronder: 👍

stan/math/fwd/fun/tcrossprod.hpp
WardBrian: I think this should always return a Eigen::Matrix to match prim, rev, etc. so the latest looks correct?

WardBrian:
@SteveBronder final check -- good to merge?

WardBrian (dismissed):

WardBrian (approved):

syclik:
@SteveBronder awesome!!!

PR #3033, opened 2024-03-21
sqrt(x) gradient when x=0

nhuurre:
## Summary

Quick fix for #3031 

## Tests

New test cases for sqrt() and distance().

## Side Effects

No.

## Release notes

Give `sqrt(x)` a finite gradient at `x=0`. Allows `distance(x,y)` to work for `x=y` too.

## Checklist

- [x] Copyright holder: Niko Huurre

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [ ] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms) checks (`make cpplint`)

- [ ] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #3147, opened 2025-02-13
Fix issue 3146 for multiply_log

### LINKED ISSUE #2494: multiply_log(0, 0) has infinite gradient
## Description
For more details, see https://discourse.mc-stan.org/t/non-linear-modelling-michaelis-menten/22736/8

## Example
The gradient for `poisson_lpmf(0 | lambda)` is infinite at `lambda = 0`.

## Suggested fix
@nhuurre writes on discourse:

> We should just special-case a=0.0 like we do with a=1.0

#### Current Version:
v4.0.1

### LINKED ISSUE #3146: Error compiling rep_matrix inside of offset_multiplier_constrain
Compiling the following Stan code gives an error about bad return deduction

```stan
parameters {
  vector<lower=0>[5] sigma_gamma_n;
  matrix<multiplier=rep_matrix(sigma_gamma_n', 10)>[10, 5] z_gamma_n;
}
```

The error log is below which traces down into `multiply_log` about an inconsistent deduction for the return type. I think just adding `constexpr if ` in this statement would fix this. Though I'd look to look at why  the return types are just slightly different. That looks like a bug in type deduction

```
stan/lib/stan_math/stan/math/rev/fun/multiply_log.hpp:214:29: error: inconsistent deduction for auto return type: ‘stan::math::var_value<Eigen::Matrix<double, -1, -1, 1, -1, -1>, void>’ and then ‘stan::math::var_value<Eigen::Matrix<double, -1, -1> >’
  214 |     return make_callback_var(
      |            ~~~~~~~~~~~~~~~~~^
  215 |         multiply_log(arena_a.val(), arena_b),
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  216 |         [arena_a, arena_b](const auto& res) mutable {
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  217 |           arena_a.adj()
      |           ~~~~~~~~~~~~~      
  218 |               += (res.adj().array() * arena_b.val().array().log()).sum();
```
https://gist.github.com/SteveBronder/5ba8b98ce92e7d1965fcce5a9fc091a2

SteveBronder:
## Summary

Fixes #3146 by making the deduction for branch logic be compile time using `if constexpr`. This also cleans up a lot of the code for `offset_multiply_constrain`. Instead of having all combinations of signatures for vector inputs we just use a little `if constexpr` to deduce which inputs need iterated over.

## Tests

There is 1 new test in `rev` for testing issue #3146 specifically. You can run all tests associated with this PR via

```bash
python ./runTests.py -j4 ./test/unit/math/ -f multiply_log
python ./runTests.py -j4 ./test/unit/math/ -f offset_mult
```

## Side Effects

Nope

## Release notes

Fix issue so that rep_matrix can be used inside of offset_multiplier constraints.

## Checklist


WardBrian:
Does this affect https://github.com/stan-dev/stan/issues/3330 or is it unrelated?

SteveBronder:
This is just #3146 I'm writing a separate PR for https://github.com/stan-dev/stan/issues/3330

SteveBronder:
@nhuurre or @andrjohns would either of you mind taking a look at this? @nhuurre Brian pointed out you handled #3033 which is very similar

andrjohns:
Can these scalar/matrix/etc overloads all be condensed? They all follow a similar pattern/logic (afaict).

We have `as_array_or_scalar()`  as well as the vectorised `select()`, so the functions can be scalar/container agnostic (examples of their usage [in `bernoulli_cdf`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/bernoulli_cdf.hpp))

jachymb:
Does lmultiply need to also be considered?

WardBrian:
Huh, I didn't realize lmultiply even really existed in the math library. I'm pretty sure the stan compiler still just rewrites it to `multiply_log`.

They definitely should be consistent between the two

test/unit/math/mix/fun/multiply_log2_test.cpp
```diff
@@ -1,12 +1,26 @@
...
 #include <limits>
 
+/**
+ * NOTE: We do not test values of (0.0, 0.0) as inputs.
+ *  This is because the testing framework uses
+ *  finite difference as the comparison. The small finite
+ *  purtubations lead to cases where the value of the
+ *  function inputs is either slightly negative or slightly positive,
+ *  which can lead to the function returning NaN when we would expect a
+ *  value of 0.
```
nhuurre, line 11: Some testing is still technically possible. On #3033 I added [a test that tricks](https://github.com/stan-dev/math/pull/3033/files#diff-fd26803a5d73ee93920f45bd060d44d49e1437a203862c6e4b2cf434940f1eaa) the finite difference calculation by extending the function symmetrically.
Here that would be
```cpp
return multiply_log((x1 >= 0) ? x1 : -x1, (x2 >= 0) ? x2 : -x2);
```
(i.e. `multiply_log(|x1|,|x2|)` without using `stan::math::abs` since it would overwrite the autodiff derivatives)

However, I do not like this hack. The value added by such a test is low compared to the effort required to understand what it does and why it works.

It's better to extend the test framework. `stan::test::test_ad` needs both positive and negative perturbations because it compares autodiff to central finite differences. It shouldn't be hard to work out the equations for lateral finite differences which need perturbations only in one direction (but are less accurate). I can try and add something like `stan::test::test_ad_positive` if you want.

Positive-perturbation-only test can handle `multiply_log(0, x)`. But the derivative of `multiply_log(x, 0)` is formally infinite so finite difference is going to blow up anyway. Maybe we need a `stan::test::test_ad` variant that just checks the autodiff derivatives are finite but does not compare them with any reference value.

nhuurre (commented):
> Brian pointed out you handled #3033 which is very similar

You mean similar to the part that handles `0*log(0)` corner case. Most of this PR is if-constexpr-template magic that goes way over my head.

WardBrian:
> Most of this PR is if-constexpr-template magic that goes way over my head.

This is, luckily, the part that I feel comfortable reviewing, so between the two of us it seems likely we will be able to make it work

PR #3151, opened 2025-02-21
Broaden the template bound on the lp argument to constraints

WardBrian:
## Summary

This makes the `lp` argument of the `_constrain` functions its own template parameter rather that solely relying on the other arguments. This is useful for https://github.com/stan-dev/stanc3/issues/1436, as for technical reasons it is hard to prevent the user from passing all doubles if these are exposed. 

It also changes the order of some overloads and uses some template pack tricks to allow the non-Lp versions of these functions to be vectorized as well.

## Checklist


stan/math/prim/constraint/cholesky_corr_constrain.hpp
```diff
@@ -43,9 +43,11 @@ cholesky_corr_constrain(const EigVec& y, int K) {
 }
 
 // FIXME to match above after debugged
-template <typename EigVec, require_eigen_vector_t<EigVec>* = nullptr>
+template <typename EigVec, typename Lp,
+          require_eigen_vector_t<EigVec>* = nullptr,
+          require_convertible_t<return_type_t<EigVec>, Lp>* = nullptr>
```
SteveBronder, line 48: So this will allow `lp` to be `var` and `return_type_t<EigVec>` to be `double` or `var`.

Do we ever want to increment the jacobian when y just has a scalar type of `double`?

```cpp
lp += 0.5 * log1m(sum_sqs);
```

What are the use types you want to pass into this? Is it just the `return_type_t<EigVec> = double` and `lp = var` case? Since the data is considered constant in that case shouldn't we just call the one that does not increment the jacobian?

WardBrian: > `return_type_t<EigVec> = double` and `lp = var`

Yes, that's the relevant one that comes up during code generation. Based on my conversation with @bob-carpenter, I think we still want to let the user do it, even if it's weird. The implementation in stanc3#1494 gives the user both a `_jacobian` function and a `_constrain` function, and I think it's reasonable to view these as similar to the difference between doing a target+= with a `_lpdf`  versus a `_lupdf`.

stan/math/prim/constraint/cov_matrix_constrain.hpp
```diff
@@ -88,30 +91,26 @@ cov_matrix_constrain(const T& x, Eigen::Index K, return_type_t<T>& lp) {
...
-template <bool Jacobian, typename T, require_not_std_vector_t<T>* = nullptr>
-inline auto cov_matrix_constrain(const T& x, Eigen::Index K,
-                                 return_type_t<T>& lp) {
-  if (Jacobian) {
-    return cov_matrix_constrain(x, K, lp);
-  } else {
-    return cov_matrix_constrain(x, K);
-  }
+template <typename T, typename... Lp, require_std_vector_t<T>* = nullptr>
+inline auto cov_matrix_constrain(const T& x, Eigen::Index K, Lp&... lp) {
```
SteveBronder: Just checking. Is this like this because in `transformed data` we will have `cov_matrix_constrain(x, 10)` etc? Shouldn't the other one have this as well?

WardBrian: "the other one"?

stan/math/prim/constraint/corr_matrix_constrain.hpp
```diff
@@ -79,28 +82,24 @@ corr_matrix_constrain(const T& x, Eigen::Index k, return_type_t<T>& lp) {
...
-  } else {
-    return corr_matrix_constrain(x, k);
-  }
+template <typename T, typename... Lp, require_std_vector_t<T>* = nullptr>
+inline auto corr_matrix_constrain(const T& y, int K, Lp&... lp) {
+  static_assert(sizeof...(lp) == 0 || sizeof...(lp) == 1,
+                "corr_matrix_constrain should be called with either "
+                "two or three arguments");
+  return apply_vector_unary<T>::apply(
+      y, [&lp..., K](auto&& v) { return corr_matrix_constrain(v, K, lp...); });
```
SteveBronder: This looks like a bug in our api since we only have the `lp` version for `corr_matirix_constrain(std::vector<>)`. imo I'd rather just have two functions. It's more code but I think cleaner just to have two separate function

stan/math/prim/constraint/cov_matrix_constrain.hpp
```diff
@@ -88,30 +91,26 @@ cov_matrix_constrain(const T& x, Eigen::Index K, return_type_t<T>& lp) {
...
  * Return the symmetric, positive-definite matrix of dimensions K by K resulting
- * from transforming the specified finite vector of size K plus (K choose 2). If
- * the `Jacobian` parameter is `true`, the log density accumulator is
- * incremented with the log absolute Jacobian determinant of the transform.  All
- * of the transforms are specified with their Jacobians in the *Stan Reference
- * Manual* chapter Constraint Transforms.
+ * from transforming the specified finite vector of size K plus (K choose 2).
+ * This overload handles looping over the elements of a standard vector.
  *
- * @tparam Jacobian if `true`, increment log density accumulator with log
```
SteveBronder, line 97: Add back docs for jacobian

stan/math/prim/constraint/cov_matrix_constrain.hpp
SteveBronder: nvm ignore this. Looking at the file this is an api bug

stan/math/prim/constraint/corr_matrix_constrain.hpp
WardBrian: You mean two separate overloads that both call `apply_vector_unary`? I can split them up if you want, it's just a bunch of copy-paste

WardBrian: Split!

SteveBronder (changes requested):
So imo everything is good except that I would like to have a new signature in the places where we are doing `LP...`. That just looks kind of clunky and having two signatures just makes a bit more sense.

stan/math/prim/constraint/corr_matrix_constrain.hpp
SteveBronder: Thanks! Yes it's just a bunch of copy pastes but imo I'd rather have duplicate code that is a bit easier to read here

stan/math/prim/constraint/cov_matrix_constrain.hpp
WardBrian: I think this is just github showing the diff weirdly because a signature moved around -- there is no jacobian template on this overload

WardBrian:
@SteveBronder I pushed the change to split up the `Lp...` earlier today -- is your latest review based on before that?

SteveBronder:
Did you want to also do `cov_matrix_constrain_lkj`? (Is that used?)

WardBrian:
It’s not used anywhere by the compiler, so I don’t currently plan on exposing it to the user

WardBrian:
@SteveBronder it looks like your latest push broke the accompanying stanc3 PR: https://jenkins.flatironinstitute.org/job/Stan/job/Stanc3/job/PR-1494/14/

SteveBronder (approved):

Issue #3154, opened 2025-02-26
Use eigen logistic function for inv_logit


jachymb:
There is the following comment int prim/fun/inv_logit.hpp from Jan 8 2020

```
// TODO(Tadej): Eigen is introducing their implementation logistic() of this
// in 3.4. Use that once we switch to Eigen 3.4
```

We use Eigen 3.4 since Mar 10 2023 and there is in fact a logistic function available: https://eigen.tuxfamily.org/dox/namespaceEigen.html#title47

so this would simplify code and perhaps perform a bit better?

jachymb:
Apparently solved with #3160

PR #3155, opened 2025-02-27
Call Eigen logistic function for inv_logit instead of custom implementation

jachymb:
My first time trying to contribute. I tried to resolve an old TODO - see #3154 

## Summary

Call Eigen logistic function for inv_logit instead of custom implementation.

I *did not* touch opencl/kernels/device_functions/inv_logit.hpp - not sure if it's necessary or how opencl works here. 

## Tests

Original tests pass. No new tests added.

## Side Effects

I am not aware of any. There may perhaps be some numerical stability/precision differences, I didn't look into that.

## Release notes

Call Eigen logistic function for inv_logit instead of custom implementation.

## Checklist

- [x] Copyright holder: Me.

### DIFF
--- a/makefile
+++ b/makefile
@@ -87,6 +87,7 @@ doxygen:
 clean:
 	@echo '  removing generated test files'
 	@$(RM) $(wildcard test/prob/generate_tests$(EXE))
+	@$(RM) $(EXPRESSION_TESTS) $(call findfiles,test/expressions,*_test.cpp)
 	@$(RM) $(call findfiles,test/prob,*_generated_v_test.cpp)
 	@$(RM) $(call findfiles,test/prob,*_generated_vv_test.cpp)
 	@$(RM) $(call findfiles,test/prob,*_generated_fd_test.cpp)
--- a/runTests.py
+++ b/runTests.py
@@ -383,7 +383,7 @@ def checkToolchainPathWindows():
             universal_newlines=True,
         )
         out, err = p1.communicate()
-        if re.search(" |\(|\)", out):
+        if re.search(r" |\(|\)", out):
             stopErr(
                 "The RTools toolchain is installed in a path with spaces or bracket. Please reinstall to a valid path.",
                 -1,
--- a/stan/math/prim/prob/wiener5_lpdf.hpp
+++ b/stan/math/prim/prob/wiener5_lpdf.hpp
@@ -679,12 +679,12 @@ inline auto wiener_lpdf(const T_y& y, const T_a& a, const T_t0& t0,
   if (!include_summand<propto, T_y, T_a, T_t0, T_w, T_v, T_sv>::value) {
     return ret_t(0.0);
   }
-  using T_y_ref = ref_type_if_t<!is_constant<T_y>::value, T_y>;
-  using T_a_ref = ref_type_if_t<!is_constant<T_a>::value, T_a>;
-  using T_t0_ref = ref_type_if_t<!is_constant<T_t0>::value, T_t0>;
-  using T_w_ref = ref_type_if_t<!is_constant<T_w>::value, T_w>;
-  using T_v_ref = ref_type_if_t<!is_constant<T_v>::value, T_v>;
-  using T_sv_ref = ref_type_if_t<!is_constant<T_sv>::value, T_sv>;
+  using T_y_ref = ref_type_t<T_y>;
+  using T_a_ref = ref_type_t<T_a>;
+  using T_t0_ref = ref_type_t<T_t0>;
+  using T_w_ref = ref_type_t<T_w>;
+  using T_v_ref = ref_type_t<T_v>;
+  using T_sv_ref = ref_type_t<T_sv>;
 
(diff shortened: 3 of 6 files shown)


WardBrian:
Hi @jachymb — I believe this TODO was suggesting we use the .logistic method on matrices in a specific overload for them, not for the implementation for double. 

Right now, the prim version of this function just has two overloads, one for doubles and one for all containers that uses apply_scalar (which is, essentially, a for loop)

To complete the TODO, there would be 3 overloads:

double

prim matrices (using .logistic)

std::vectors (using apply_scalar like it currently would)


The hardest part of doing this is generally writing the template bounds so that the correct version is always selected. If you’re running into trouble with that, we can try to help!

jachymb:
Thanks for the commentary!

How about something like this replacing the current vectorized version then?

```cpp
inline auto inv_logit(const T& x) {
  return x.array().logistic();
}

template <typename T, require_std_vector_t<T>* = nullptr, require_not_eigen_t<T>* = nullptr>
inline auto inv_logit(const T& x) {
  return apply_scalar_unary<inv_logit_fun, T>::apply(x);
}
```

it seems to work when I add tests for `std::vector<double>` and `stan::math::matrix_d`

does that look reasonable?

WardBrian:
That looks more-or-less correct. Because we want the autodiff overloads (which live in a different file) to also not intersect with these, I think you need to add a `require_not_var_matrix_t` to the eigen overload.

I _think_ you might also still need to keep the `require_all_not_nonscalar_prim_or_rev_kernel_expression_t` on one of those, but I'm also not super familiar with the openCL details, so it's probably best we ask @SteveBronder on that one

stan/math/prim/fun/inv_logit.hpp
```diff
@@ -75,21 +67,29 @@ struct inv_logit_fun {
...
  * @return Inverse logit applied to each value in x.
  */
-template <
-    typename T, require_not_var_matrix_t<T>* = nullptr,
-    require_all_not_nonscalar_prim_or_rev_kernel_expression_t<T>* = nullptr>
+template <typename T, require_eigen_t<T>* = nullptr>
 inline auto inv_logit(const T& x) {
-  return apply_scalar_unary<inv_logit_fun, T>::apply(x);
+  return x.array().logistic();
 }
```
SteveBronder: So right now this function works over all containers that are not `var<Matrix>` or `var<MatrixCL>` types. If we want a version for just doubles I would add three signatures. Two in this file and one in `stan/math/rev/fun/inv_logit.hpp`

```cpp
// Version that works on Eigen types which have an arithmetic value type
template <typename T, require_eigen_vt<std::is_arithmetic, T> * = nullptr>
inline auto inv_logit(T&& x);

// Version that works on std::vector types and does apply scalar unary
template <typename T, require_std_vector_t<T> * = nullptr>
inline auto inv_logit(T&& x);
```

Then in `stan/math/rev/fun/inv_logit.hpp` to add the signature

```cpp
// Version that works with Eigen types with inner `var` value types
template <typename T, require_eigen_vt<is_var, T> * = nullptr>
inline auto inv_logit(T&& x);
```

You can look at the other function in `rev/fun/inv_logit.hpp` to see how to write functions with our autodiff scheme. We also have a guide for adding new functions at the link below

https://mc-stan.org/math/md_doxygen_2contributor__help__pages_2getting__started.html

andrjohns:
We already have several functions using Eigen's vectorised calls that could be a good guide on approaching this

For example:

  - [`prim/fun/exp`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/exp.hpp)
  - [`prim/fun/inv_sqrt`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/inv_sqrt.hpp)

jachymb:
Added the signatures suggested by SteveBronder and WardBrian, but to be honest, at this point I don't feel very confident I actually know what I'm doing :( I think I may need to backtrack a few steps to study the code & theroy a bit more. If this isn't it, feel free to close the PR or reuse any of my code in some other way.

SteveBronder:
This code is 90% of the way there! I'll look at this on Monday

stan/math/prim/fun/inv_logit.hpp
```diff
@@ -49,15 +49,7 @@ namespace math {
...
-  using std::exp;
-  if (a < 0) {
-    double exp_a = exp(a);
-    if (a < LOG_EPSILON) {
-      return exp_a;
-    }
-    return exp_a / (1 + exp_a);
-  }
-  return inv(1 + exp(-a));
+  return Eigen::internal::scalar_logistic_op<double>()(a);
```
SteveBronder: Revert this and leave it as is

SteveBronder:
@jachymb I was trying to update your branch, but something odd happened and now github will not let me reopen your PR? I very much apologize, your PR was very close! I think the issue was that you commited to your forks `develop` branch. Normally its good practice to make a new branch for a feature with `git checkout -b my-feature-name`

image

Issue #3133, opened 2024-12-07
Implement the generalized normal distribution


jachymb:
## Description
Implement Generalized normal distribution https://en.wikipedia.org/wiki/Generalized_normal_distribution

### Why this is useful?

It generalizes the normal (for p=2) and double exponential (Laplace) distribution (for p=1) and in the limit case also the uniform (p → ∞) for the shape parameter p.

These correspond to the $L_p$ norms when used for regularization and in other cases.

For examples: 

```stan
y ~ normal(X*beta, sigma);
```

produces the criterion minimize $L_2$ norm of (X*beta-y) (a.k.a. least squares). But then

```stan
y ~ double_exponential(X*beta, sigma);
```

produces the criterion minimize $L_1$ norm of (X*beta-y). (a.k.a. least absolute deviations)

Similarly, in the Bayesian interpretation of ridge and LASSO, 

```stan
beta ~ normal(0, lambda);
y ~ normal(X*beta, sigma);
```

produces the $L_2$ regularized ridge and 

```stan
beta ~ double_exponential(0, lambda);
y ~ normal(X*beta, sigma);
```

produces the $L_1$ regularized LASSO. 

Using the Generalized normal distribution would allow to conveniently use an arbitrary $L_p$ norm for the optimization criterion, or to even find the suitable value of p when used as a parameter.

jachymb:
Just some notes on the corresponding formulas:

We have the density:

$f(y|\mu,\alpha, \beta) = \frac{\beta}{2\alpha \Gamma(\beta^{-1})} \exp\left(-\left(\frac{|y-\mu|}{\alpha}\right)^{\beta}\right)$ 

The log-density:

$\log f(y|\mu,\alpha, \beta) = \log(\beta) - \log\Gamma(\beta^{-1}) - \log(2) - \log(\alpha) -\left(\frac{|y-\mu|}{\alpha}\right)^{\beta}$ 

And the parital derivatives:

$\frac{\partial \log f(y|\mu,\alpha, \beta)}{\partial y} = \beta \alpha^{-\beta}|\mu-y|(\mu-y)^{-1}$ 

$\frac{\partial \log f(y|\mu,\alpha, \beta)}{\partial \mu} = \beta \alpha^{-\beta}|\mu-y|^{\beta}(y-\mu)^{-1} = -\frac{\partial \log f(y|\mu,\alpha, \beta)}{\partial y}$ 

[...]

Issue #3031, opened 2024-03-20
`pow(base, exp)` gradient inconsistency at `base=0.0`


nhuurre:
The following model fails to initialize due to bad gradient if `s = 1` or `s = 2` but initializes fine when `s = 3` and `s = 4`.
```stan
data {
  int<lower=1,upper=4> s;
}
parameters {
  vector[2] x;
}
model {
  if (s == 1)
    target += distance(x,x);
  else if (s == 2)
    target += pow(squared_distance(x,x), 0.5);
  else if (s == 3)
    target += pow(squared_distance(x,x), 0.5 + 1e-16);
  else if (s == 4)
    target += abs(x[1] - x[1]) + abs(x[2] - x[2]);
}
```
This is surprising because `s = 1` looks like a well-formed (though redundant) model and in any case `s=1`, `s=2`, and `s=3` should all be equivalent. (`s=4` is a one-dimensional version of `s=1`)

The behaviour for `s=2` and `s=3` is  different because `pow(base, exp)` skips gradient calculation at `base=0` in the general case (as if the gradient was zero) but delegates `exp=0.5` case to `sqrt()` which always computes the mathematically correct gradient (infinity at `base=0`). And zero gradient ends up being accidentally correct whereas an infinite gradient is an unrecoverable error.

The mathematically correct value for the derivative of `pow(base, exp)` with respect to `base` at `base=0.0` is
* 0.0 when exp > 1.0
* 1.0 when exp = 1.0
* positive infinity when 0.0 < exp < 1.0
* 0.0 when exp = 0.0
* negative infinity when exp < 0.0

It would be reasonable to round infinity down to a finite-but-near-maximum floating point number.
Moreover, Stan only requires gradients to be correct [almost everywhere](https://en.wikipedia.org/wiki/Almost_everywhere). Boundary points like `base=0` are relevant only in expressions like `distance(x,x)` where the gradients cancel anyway and the only thing that matters is whether the intermediate gradient values are finite. So the most numerically stable choice might be to "round" the infinity all the way down to zero.
I'm inclined to think `pow()`'s current behaviour is correct for us and `sqrt()` (and possibly `cbrt()`)  should be "fixed" to also have vanishing gradient at zero.

#### Current Version:
v4.8.1

WardBrian:
#3033

Issue #2993, opened 2023-12-29
pow applied to mixed variables


jeanchristopheruel:
~~AD of `pow<fvar<var>>(0, 0)` causes NAN adjoint. `pow<fvar<var>>(0, 0)` and `pow<var>(0, 0)` both behave as expected.~~

**Update**
`pow<var>(0, 0)` should return NAN adjoint since the partials are undefined.

## Steps to reproduce
```c++
stan::math::nested_rev_autodiff nested{};
stan::math::var t{0.0};
auto foo = stan::math::pow(t,0);
foo.grad();
assert(not std::isfinite(t.adj()));  // fails
```

#### Current Version:
v4.7.0

andrjohns:
I think this is correct, as the second derivative of `pow(x, y)` w.r.t. `x`:

$$
\frac{d^2}{dx^2}(x^y) = x^{(-2 + y)} \cdot (-1 + y) \cdot y
$$

Is not defined when [both x & y are 0](https://www.wolframalpha.com/input?i=0%5E%28-2+%2B+0%29+*%28-1+%2B+0%29+*0)

andrjohns:
Ah I see, we handle this in the `rev` header by skipping the partials update. Thanks for catching this!

jeanchristopheruel:
Yes exacly! It is handled for `rev` [here](https://github.com/stan-dev/math/blob/ae90c3edca5d3b5e5234fdd6cbd1146f1666761d/stan/math/rev/fun/pow.hpp#L91)

jeanchristopheruel:
Actually, I think pow<var> is the issue. Check out my suggestion!

jeanchristopheruel:
@andrjohns I beleve I have fixed the issue here, and I added a test with expect_ad(f,0,0) but it seems like the test is not behaving as expected. Here it the result. Would you have any suggestion? It seems like the finite differences aren't accurate in this case.
```bash
test/unit/math/mix/fun/pow_part1_test --gtest_output="xml:test/unit/math/mix/fun/pow_part1_test.xml"
Running main() from lib/benchmark_1.5.1/googletest/googletest/src/gtest_main.cc
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from mathMixScalFun
[ RUN      ] mathMixScalFun.powInstantiations
[       OK ] mathMixScalFun.powInstantiations (0 ms)
[ RUN      ] mathMixScalFun.pow
[...]

PR #3157, opened 2025-03-04
Generalized normal distribution

jachymb:
## Summary

This is my take on the request #3133 I created a while ago. 

The Generalized normal distribution a.k.a. exponential power distribution generalizes the normal, Laplace (double exponential) and in limit also uniform distribution. This makes it an interesting prior choice as it generalizes the corresponding MAP regularization coefficient to L_p norm and can also be interesting to model noise in regression problems for basically the same reasons.

This is my first attempt at implementing a distribution to STAN, I tried to follow the guide https://mc-stan.org/math/md_doxygen_2contributor__help__pages_2adding__new__distributions.html but there are likely some problems, so I welcome any comments. 

So far I implemented only the lpdf, if other files like lcdf etc are strictly required, please let me know.

## Tests

I added one test file analogical to the one for normal distribution. 

## Side Effects

I included some very minor code polishing to the related normal and double exponential distributions (I used those as a starting point) which shouldn't do anything and can be discarded if undesired.

## Release notes

Added generalized_normal_lpdf

## Checklist

- [x] Copyright holder: Me


andrjohns:
Thanks for this! Can you remove the changes to the `normal` and `double_exponential` if they're unrelated? Feel free to open a separate PR for them afterwards though!

jachymb:
I discarded any changes to the existing distributions

jachymb:
Sorry for a broken PR, I believed the tests pass, but apparently I made some mistake when running the tests locally. 

Would appreciate any help deciphering the error messages, it doesn't seem very understandable what the problem is exactly to me. But apparently it somehow has to do with the distribution being possibly non-differentiable at mu=y.

The suspect line is:

`sign(diff) * beta_val * pow(scaled_abs_diff, beta_val - 1) * inv_alpha`

In case when mu=y, we get:

`sign(0) * beta_val * pow(0, beta_val - 1) * inv_alpha`

which I think should evaluate to 0, and also when beta=1

`sign(0) * 1 * pow(0, 0) * inv_alpha`

should also evaluate to 0 I would expect? Is pow(0,0) actually undefined in STAN math? And what would be the recommended technique for special-case handling?

WardBrian:
> Sorry for a broken PR, I believed the tests pass, but apparently I made some mistake when running the tests locally.

No worries! It's often quite a miracle when these things pass first try.

You can run the exact test that is given issues locally by running `python runTests.py -j4 test/prob/generalized_normal/generalized_normal_0*_test.cpp`

> Is pow(0,0) actually undefined in STAN math?

It's defined (https://github.com/stan-dev/math/issues/3031), but it's not twice differentiable: https://github.com/stan-dev/math/issues/2993#issuecomment-1871732573, which I believe is the problem here. This second comment is by @andrjohns who can probably also give you suggestions about what to do in this case. 

It's possible that what you're doing in the code is fine and you just need to have the test call it somewhere else where it will be better behaved

jachymb:
OK, so it may be just a testing problem. But this relates to something else: The case when beta=1 where the tests are apparently breaking is maybe just a tip of the iceberg: when beta<1, I write in the code comments the partials are undefined when y=mu, because we then get `pow(0, negative_number)`. This is because the limit can diverge to both positive and negative infinity depending on whether we approach from y>mu or y<mu. I left this undefined intentionally, but is that actually an OK behavior for STAN? Or do we need to use some technique to e.g. artificially set the partial to 0 in theses cases (like sign(0)=0 is defined to be the mathematically non-existent derivative of abs(0))

stan/math/prim/prob/generalized_normal_lpdf.hpp
```diff
@@ -0,0 +1,156 @@
...
+    logp -= sum(log(alpha_val)) * (N / math::size(alpha));
+  }
+  if (include_summand<propto, T_shape>::value) {
+    logp -= sum(lgamma(inv_beta1p)) * (N / math::size(beta));
+  }
+
+  auto ops_partials
+      = make_partials_propagator(y_ref, mu_ref, alpha_ref, beta_ref);
+
+  if (!is_constant_all<T_y, T_loc>::value) {
```
SteveBronder: For any if statements with known compile time values we can use `if constexpr`

```suggestion
  if constexpr (!is_constant_all<T_y, T_loc>::value) {
```

stan/math/prim/prob/generalized_normal_lpdf.hpp
```diff
@@ -0,0 +1,156 @@
...
+    // beta < 1. The derivative limit as mu -> y goes:
+    //   to 0 from both sides if beta > 1 (defined as 0)
+    //   to +1/alpha from right but -1/alpha from left if beta == 1 (defined as
+    //   0, consistent with double_exponential_lpdf) to +∞ from right but -∞
+    //   from left as y -> mu if beta < 1 (undefined)
+    const auto& rep_deriv
+        = to_ref_if < !is_constant<T_y>::value
+          && !is_constant<T_loc>::value
+                 > (sign(diff) * beta_val * pow(scaled_abs_diff, beta_val - 1)
+                    * inv_alpha);
```
SteveBronder: Once you are in this block you know that `T_y` and `T_loc` are not constant
```suggestion
    auto rep_deriv = eval(sign(diff) * beta_val * pow(scaled_abs_diff, beta_val - 1) * inv_alpha);
```

stan/math/prim/prob/generalized_normal_lpdf.hpp
```diff
@@ -0,0 +1,156 @@
...
+  auto ops_partials
+      = make_partials_propagator(y_ref, mu_ref, alpha_ref, beta_ref);
+
+  if (!is_constant_all<T_y, T_loc>::value) {
+    // note: The partial derivatives for y, mu are undefined when y == mu &&
+    // beta < 1. The derivative limit as mu -> y goes:
+    //   to 0 from both sides if beta > 1 (defined as 0)
+    //   to +1/alpha from right but -1/alpha from left if beta == 1 (defined as
+    //   0, consistent with double_exponential_lpdf) to +∞ from right but -∞
+    //   from left as y -> mu if beta < 1 (undefined)
```
SteveBronder: I had a hard time understanding your comment here. Is this the right interpretation? Either way can you clean it up a bit so the cut points and conditions are laid out a bit nicer
```suggestion
    // note: The partial derivatives for y, mu are undefined when 
    // y == mu && beta < 1. 
    // The derivative limit as mu -> y has the following cases:
    //   beta > 1: 0 from both sides(defined as 0)
    //   beta == 1: +1/alpha from right, but -1/alpha from left (defined as
    //     0, consistent with double_exponential_lpdf) to +∞ from right
    //   beta < 1: -∞ from left as y -> mu (undefined)
```

SteveBronder (changes requested):
Just some starting Q comments. 

I'm rather unfamiliar with this distribution so will need to do some reading before I can comment on any of the math here

jachymb:
Some notes on the distribution (I made some mistakes in the #3133 comments but here should be corrected):

We have the density:

$f(y|\mu,\alpha,\beta) \propto \exp\left(-\left(\frac{|y-\mu|}{\alpha}\right)^{\beta}\right)$ with normalization constant $\frac{1}{2\alpha\Gamma(1 + \frac{1}{\beta})} = \int_{-\infty}^{\infty} \exp\left(-\left(\frac{|y-\mu|}{\alpha}\right)^{\beta}\right) \mathrm{d}y$ (note: can set $\mu = 0$ WLOG to evaluate the integral)

The parameters  $\alpha, \beta$ must be positive.

The log-likelihood is then:

$\log f(y|\mu,\alpha,\beta) = -\left(\frac{|y-\mu|}{\alpha}\right)^{\beta} - \log(2) -\log(\alpha) - \log\Gamma(1 + \frac{1}{\beta})$

Wolfram Mathematica outputs the following partials:

$\frac{\partial \log f(y|\mu,\alpha,\beta)}{\partial \mu} = \alpha^{-\beta}\beta|y-\mu|^{\beta-1} \mathrm{sign}\left(y-\mu\right)$

$\frac{\partial \log f(y|\mu,\alpha,\beta)}{\partial y} = -\alpha^{-\beta}\beta|y-\mu|^{\beta-1} \mathrm{sign}\left(y-\mu\right) = -\frac{\partial \log f(y|\mu,\alpha,\beta)}{\partial \mu}$

Using algebraic manipulations I change this to the form:

$\frac{\partial \log f(y|\mu,\alpha,\beta)}{\partial \mu} = \alpha^{-1}\beta \left(\frac{|y-\mu|}{\alpha}\right)^{\beta-1} \mathrm{sign}\left(y-\mu\right)$

This is where I believe the test errors are comming from when testing the case $\beta = 1, y = 0, \mu=0$. The lpdf is defined, the first derivative is by convention 0 (since pow(0,0)=1 and sign(0)=0), but the second derivative undefined. And I am not even testing the case when $\beta < 1$ which I kept undefined for $y = \mu$ as the there is negative power of zero.

The $\beta$ partial is:

$\frac{\partial \log f(y|\mu,\alpha,\beta)}{\partial \beta} = \frac{\mathrm{digamma}(1 + \frac{1}{\beta})}{\beta^2} - \left(\frac{|y-\mu|}{\alpha}\right)^{\beta} \log(\frac{|y-\mu|}{\alpha})$

 this one has a also potential boundary problem when $y=\mu$ because we get $\log(0)$ but then the limit goes to 0 and this should be solved by using `multiply_log`. Not sure about second derivatives tho, do they need to be considered here?


The $\alpha$ partial I think is not problematic but just for completeness, I also used wolfram plus some algebra to obtain:

$\frac{\partial \log f(y|\mu,\alpha,\beta)}{\partial \alpha} = \alpha^{-1}(\beta \left(\frac{|y-\mu|}{\alpha}\right)^{\beta} - 1 )$

I tend to believe that @WardBrian is right that this is perhaps a problem with testing the distribution at these boundaries, but idk what a good approach to this is.

PDF plot from wikipedia for $\mu = 0, \alpha = 1$ illustrates the non-differentiability at $y=0$ for $\beta \le 1$:
pdf

jachymb:
Incorporated changes requested by @SteveBronder although that doesn't solve the test errors. I think those are more like just performance+stylistic improvements. See the notes and discussion above.

jachymb:
Hi, @andrjohns! Could you please go through the above comments and provide me with any feedback on how to properly test this or resolve some other way?

jachymb:
pushed some  changes which should fix the failing tests.

PR #3158, opened 2025-03-05
Add arm64 to GHA CI and split tests across more runners

andrjohns:
## Summary

Now that arm64 runners are available in github actions, we can add them as part of our regular CI. This will help catch issues like those fixed in #3059 - it's already identified issues with `owens_t`!

The `owens_t` function was hanging indefinitely under the arm64 runner, and this was resolved by adding our `boost_policy` definition as an argument - I'm not entirely sure why (might be another precision thing?).

I've added the runners as part of the existing Windows Rtools CI, using `pwsh` to avoid changing the current workflow in any meaningful way.

## Tests

## Side Effects

N/A

## Release notes

Add arm64 linux to Github Actions CI

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,3 +1,4 @@
+# Note that the workflows below use Powershell (pwsh) as they were originally written for Windows runners.
 name: Rtools & ARM64 CI
 
 on:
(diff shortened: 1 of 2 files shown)


andrjohns:
I've also split the tests across more runners, bringing the total runtime down to ~1.5hours.

Given that the Jenkins mix unit tests take ~4 hours, maybe we could look at using the github runners for the basic unit tests instead (leaving the GPU, mpi, etc for Jenkins)? That could speed the CI process up a bit and also free up the resources for the CI in the other repos

WardBrian:
We could also split up the Jenkins jobs into more parallel executors. Especially if they end up being short-lived, I don't think that would be a problem -- but I'd look into it in a separate PR rather than at the same time as this

.github/workflows/main.yml
```diff
@@ -20,101 +20,31 @@ concurrency:
...
+          - { os: windows-latest, label: misc, tests: 'test/unit/*_test.cpp test/unit/math/*_test.cpp test/unit/math/memory' }
+          - { os: ubuntu-24.04-arm, label: misc, tests: 'test/unit/*_test.cpp test/unit/math/*_test.cpp test/unit/math/memory' }
+          - { os: windows-latest, label: prim, tests: 'test/unit/math/prim' }
+          - { os: ubuntu-24.04-arm, label: prim, tests: 'test/unit/math/prim' }
+          - { os: windows-latest, label: rev, tests: 'test/unit/math/rev' }
+          - { os: ubuntu-24.04-arm, label: rev, tests: 'test/unit/math/rev' }
+          - { os: windows-latest, label: fwd, tests: 'test/unit/math/fwd' }
+          - { os: ubuntu-24.04-arm, label: fwd, tests: 'test/unit/math/fwd' }
+          - { os: windows-latest, label: 'non-fun mix', tests: 'test/unit/math/mix/core test/unit/math/mix/meta test/unit/math/mix/*_test.cpp' }
+          - { os: ubuntu-24.04-arm, label: 'non-fun mix', tests: 'test/unit/math/mix/core test/unit/math/mix/meta test/unit/math/mix/*_test.cpp' }
```
WardBrian: I think this could be written more succintly?

```
matrix:
  os: [windows-latest, ubuntu-24.04-arm]
  tests: [{label: 'misc', paths: 'test/unit/*_test.cpp test/unit/math/*_test.cpp test/unit/math/memory'} , ... ]

File: .github/workflows/main.yml [hunk 2]
```diff
@@ -1,4 +1,4 @@
-name: Windows Rtools44
+name: Rtools & ARM64 CI
```
WardBrian, line 2: I think it's worth adding a comment somewhere (might as well be up here at the top) explaining the historical reasons for why this file is using pwsh -- I think it's fine to do so, but it will certainly look odd to someone in the future if the context is lost

WardBrian (commented):
Two minor comments but this looks good -- thanks!

.github/workflows/main.yml
andrjohns: I didn't even know this syntax was an option, that's much cleaner!

WardBrian (approved):

WardBrian:
Hey @andrjohns, I've noticed some extra flakiness in the windows tests after this -- e.g,
https://github.com/stan-dev/math/actions/runs/13723040759/attempts/1
and the re-run:
https://github.com/stan-dev/math/actions/runs/13723040759/job/38398633425 (notice it's a different test that fails here)

These passed during the CI of #3151, which was using the old action at the time. Any idea what's up?

WardBrian:
Update: Run #4 did pass. Definitely worth looking into more next week

PR #3160, opened 2025-03-05
Use logistic function from eigen (based on jachymb's PR)

SteveBronder:
## Summary

Modifies `inv_logit` to use the `logistic` function from Eigen. Also makes a new reverse mode specialization of `inv_logit` for Eigen matrices.

## Tests

No new tests. 

## Checklist

- [x] Copyright holder: Jáchym Barvínek and Steve Bronder


andrjohns:
Can this use `apply_vector_unary`? That way the Eigen implementation will also be used for `std::vector` (and nested containers)

SteveBronder:
@andrjohns do you mean something besides the definition in prim that uses `apply_scalar_unary`?

```cpp
inline auto inv_logit(T&& x) {
  return apply_scalar_unary<inv_logit_fun, std::decay_t<T>>::apply(
      std::forward<T>(x));
}
```

andrjohns:
> @andrjohns do you mean something besides the definition in prim that uses `apply_scalar_unary`?
> 
> ```c++
> template <typename T, require_std_vector_t<T>* = nullptr>
> inline auto inv_logit(T&& x) {
>   return apply_scalar_unary<inv_logit_fun, std::decay_t<T>>::apply(
>       std::forward<T>(x));
> }
> ```

Yeah the issue with the current definition is that it won't use the vectorised Eigen implementation for `std::vector` types (either `std::vector<double>` or `std::vector<Eigen::...>`). You can instead use the pattern from `prim/fun/exp` where `apply_scalar_unary` is used for non-arithmetic containers and `apply_vector_unary` is used for arithmetic containers:

```cpp
template <typename Container, require_ad_container_t<Container>* = nullptr>
inline auto inv_logit(const Container& x) {
  return apply_scalar_unary<inv_logit_fun, Container>::apply(x);
}

template <typename Container,
          require_container_bt<std::is_arithmetic, Container>* = nullptr>
inline auto inv_logit(const Container& x) {
  return apply_vector_unary<Container>::apply(
      x, [](const auto& v) { return v.array().logistic(); });
}
```

This way the Eigen logistic function will be used for all arithmetic containers (even arbitrarily nested ones)

andrjohns:
`apply_vector_unary` also handles returning an `Eigen::Matrix` vs `Eigen::Array` to match the input type, which would resolve the current unit test failures

stan/math/prim/fun/inv_logit.hpp
```diff
@@ -69,28 +68,45 @@ inline double inv_logit(double a) {
...
 };
 
 /**
- * Vectorized version of inv_logit().
+ * Vectorized version of inv_logit() for std::vector's containing ad types.
  *
- * @tparam T type of container
- * @param x container
+ * @tparam T type of std::vector
+ * @param x std::vector
```
andrjohns: This overload would also be used for Eigen types with `fvars` I think

stan/math/prim/fun/inv_logit.hpp
```diff
@@ -69,28 +68,45 @@ inline double inv_logit(double a) {
...
-// TODO(Tadej): Eigen is introducing their implementation logistic() of this
-// in 3.4. Use that once we switch to Eigen 3.4
-
+/**
+ * Vectorized version of inv_logit() for Eigen types.
+ *
+ * @tparam T A type of either `std::vector` whose inner type inherits from
+ * `Eigen::DenseBase` or a type that directly inherits from `Eigen::DenseBase`.
+ * The inner scalar type must not have a `var` scalar type.
+ * @param x Eigen expression
```
andrjohns: This overload would also be for `std::vector` with inner types of `double` or `std::vector`

andrjohns (changes requested):
Just a couple of doc changes, but otherwise LGTM!

jachymb:
Note there is also the `Eigen::internal::scalar_logistic_op` which can be used for the scalar (`double`) implementation instead of the custom implementation using `exp` calls. Not sure if there is any noticeable advantage besides shortening the code by a few lines.

SteveBronder:
@andrjohns ty those docs should be cleaned up now!

@jachymb personally I don't know what the benefit of using Eigen's logistic function for scalars would be. For the vectorized version of the code, using the Eigen `logistic` function makes sense so we can use their [SIMD backend](https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/functors/UnaryFunctors.h#L1188). But for scalars we would want to check the performance and precision before using it. I'm fine with hand waving the vectorized check because using an Eigen internal function is almost always faster for vectors.

andrjohns (approved):
LGTM!

PR #3165, opened 2025-03-18
Minor fixes for hypergeometric 1F0 and 2F1

andrjohns:
## Summary

As pointed out in [the stanc3 repo](https://github.com/stan-dev/stanc3/issues/1285), the hypergeometric 1F0 function naming is inconsistent with the others - using a lowercase 'f' (i.e., `hypergeometric_1f0` instead of `hypergeometric_1F0`)

Additionally, `hypergeometric_2F1` fails to compile if the second argument is an autodiff type but not the others

## Tests

## Side Effects

N/A

## Release notes

Renamed `hypergeometric_1f0` to `hypergeometric_1F0` for consistency, fixed compilation for `hypergeometric_2f1` with autodiff second argument

## Checklist

### DIFF
--- a/stan/math/prim/fun/hypergeometric_1F0.hpp
+++ b/stan/math/prim/fun/hypergeometric_1F0.hpp
@@ -29,7 +29,6 @@ namespace math {
  */
 template <typename Ta, typename Tz, require_all_arithmetic_t<Ta, Tz>* = nullptr>
 return_type_t<Ta, Tz> hypergeometric_1F0(const Ta& a, const Tz& z) {
-  constexpr const char* function = "hypergeometric_1F0";
   check_less("hypergeometric_1F0", "abs(z)", std::fabs(z), 1.0);
 
   return boost::math::hypergeometric_1F0(a, z, boost_policy_t<>());
--- a/stan/math/prim/fun/hypergeometric_1F0.hpp
+++ b/stan/math/prim/fun/hypergeometric_1F0.hpp
@@ -28,9 +28,9 @@ namespace math {
  * @return Hypergeometric 1F0 function
  */
 template <typename Ta, typename Tz, require_all_arithmetic_t<Ta, Tz>* = nullptr>
-return_type_t<Ta, Tz> hypergeometric_1f0(const Ta& a, const Tz& z) {
-  constexpr const char* function = "hypergeometric_1f0";
-  check_less("hypergeometric_1f0", "abs(z)", std::fabs(z), 1.0);
+return_type_t<Ta, Tz> hypergeometric_1F0(const Ta& a, const Tz& z) {
+  constexpr const char* function = "hypergeometric_1F0";
+  check_less("hypergeometric_1F0", "abs(z)", std::fabs(z), 1.0);
 
   return boost::math::hypergeometric_1F0(a, z, boost_policy_t<>());
 }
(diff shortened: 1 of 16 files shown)


stan/math/prim/fun/hypergeometric_1F0.hpp
```diff
@@ -28,9 +28,9 @@ namespace math {
  * @return Hypergeometric 1F0 function
  */
 template <typename Ta, typename Tz, require_all_arithmetic_t<Ta, Tz>* = nullptr>
-return_type_t<Ta, Tz> hypergeometric_1f0(const Ta& a, const Tz& z) {
-  constexpr const char* function = "hypergeometric_1f0";
-  check_less("hypergeometric_1f0", "abs(z)", std::fabs(z), 1.0);
+return_type_t<Ta, Tz> hypergeometric_1F0(const Ta& a, const Tz& z) {
+  constexpr const char* function = "hypergeometric_1F0";
```
WardBrian: I realize this is just a find-and-replace PR, but is this a unused variable?

WardBrian (dismissed):

andrjohns:
Just rolled the 2F1 fixes into this PR as well instead of making a separate one

WardBrian:
Not sure if you want to look at it here, but `hypergeometric_3F2` is unhappy if given a row_vector -- `vector` and `array[] real` are fine, but row_vector raises an error on the line `append_row(as_array_or_scalar(b), 1.0);`

By contrast, `hypergeometric_pFq` is happy with `vector` and `row_vector` but not `array[] real` (`no matching function call`...).

Neither of these are strictly necessary for exposing in the language, but if we want consistency they'd be nice to address

WardBrian:
@andrjohns the rev specialization of `hypergeometric_pFq` is [failing expression tests](https://jenkins.flatironinstitute.org/job/Stan/job/Stanc3/job/PR-1504/2/), I believe because the line

```c++
  auto pfq_val = hypergeometric_pFq(a.val(), b.val(), value_of(z));
```
should be
```c++
  auto pfq_val = hypergeometric_pFq(arena_a.val(), arena_b.val(), value_of(z));
```

WardBrian (dismissed):
This can go in as is (assuming you don't want to look at the vector issues mentioned above)

andrjohns:
> This can go in as is (assuming you don't want to look at the vector issues mentioned above)

The vector issues were pretty simple fixes, so I've just added them as well

WardBrian:
@andrjohns I think the (hopefully) last thing is that hypergeometric_3F2 isn't expression safe (does a sum and a to_vector, both of which will eval) and also does indexing into `a` and `b` without first checking their sizes. So it needs a little TLC, and probably also an entry in the [`special_arg_values` list](https://github.com/stan-dev/math/blob/develop/test/sig_utils.py#L71) (unless it just can't be called with length-1 vectors, in which case we need to skip it?)

WardBrian:
Would it make sense if something like 3F2 had a signature `(tuple(real, real, real), tuple(real, real)) => real` instead of/as well as the current signature which takes a vector and just assumes the length?

WardBrian:
> The vector issues were pretty simple fixes, so I've just added them as well

It looks like this change missed the `rev` overload, which also seems to require some template changes to `grad_pFq` to compile with std::vectors

andrjohns:
> @andrjohns I think the (hopefully) last thing is that hypergeometric_3F2 isn't expression safe (does a sum and a to_vector, both of which will eval) and also does indexing into `a` and `b` without first checking their sizes. So it needs a little TLC, and probably also an entry in the [`special_arg_values` list](https://github.com/stan-dev/math/blob/develop/test/sig_utils.py#L71) (unless it just can't be called with length-1 vectors, in which case we need to skip it?)

We probably don't need to expose both pFq and 3F2 tbh. The only extra thing in 3F2 is manual handling of a particular case which doesn't meet convergence criteria but numerically converges that was already in the tests, so I'd consider that more of an edge case.

WardBrian:
Fine by me, though if it’s easy enough there’s no reason to not add them I guess. Do you have a chance to look at rev/fwd for pFq with the vector changes?

WardBrian (approved):

PR #3168, opened 2025-03-25
Save some memory in simplex constrain

WardBrian:
## Summary

I was looking at `simplex_constrain`'s rev implementation and realized two things:

1. It duplicates the prim function for the forward pass in order to store the `z`s, but is otherwise the same
2. We are exactly re-computing each `z` in the reverse pass, anyway (at least for the version that takes `lp`)

So, some napkin math, we can save ~20% of the memory overhead of the rev implementation by just not storing these, and simplify the code by delegating the forward pass to prim.

The same is very nearly true for the stochastic matrices, except the forward pass inside the rev specialization is written in a more vectorized style than the prim one is, so it isn't a direct swap. It's still true that we are re-computing something we're also storing, so I still save the memory there, I just don't replace the forward with a call to prim.

If we want, I can move the vectorized stochastic code into prim and then do the other half of the change to rev for those functions as well.

## Tests

Existing tests pass

## Side Effects

None

## Release notes

Reduce the memory overhead of the simplex constraints in reverse mode.

## Checklist

### DIFF
--- a/stan/math/rev/constraint/simplex_constrain.hpp
+++ b/stan/math/rev/constraint/simplex_constrain.hpp
@@ -33,35 +33,24 @@ inline auto simplex_constrain(const T& y) {
 
   size_t N = y.size();
   arena_t<T> arena_y = y;
-  arena_t<Eigen::VectorXd> arena_z(N);
-  Eigen::VectorXd x_val(N + 1);
 
-  double stick_len(1.0);
-  for (Eigen::Index k = 0; k < N; ++k) {
-    double log_N_minus_k = std::log(N - k);
-    arena_z.coeffRef(k) = inv_logit(arena_y.val().coeff(k) - log_N_minus_k);
-    x_val.coeffRef(k) = stick_len * arena_z.coeff(k);
-    stick_len -= x_val(k);
-  }
-  x_val.coeffRef(N) = stick_len;
-
-  arena_t<ret_type> arena_x = x_val;
+  arena_t<ret_type> arena_x = simplex_constrain(arena_y.val());
 
   if (unlikely(N == 0)) {
     return ret_type(arena_x);
   }
 
-  reverse_pass_callback([arena_y, arena_x, arena_z]() mutable {
+  reverse_pass_callback([arena_y, arena_x]() mutable {
     int N = arena_y.size();
     double stick_len_val = arena_x.val().coeff(N);
     double stick_len_adj = arena_x.adj().coeff(N);
     for (Eigen::Index k = N; k-- > 0;) {
+      double z_k = inv_logit(arena_y.val().coeff(k) - log(N - k));
       arena_x.adj().coeffRef(k) -= stick_len_adj;
       stick_len_val += arena_x.val().coeff(k);
-      stick_len_adj += arena_x.adj().coeff(k) * arena_z.coeff(k);
+      stick_len_adj += arena_x.adj().coeff(k) * z_k;
       double arena_z_adj = arena_x.adj().coeff(k) * stick_len_val;
-      arena_y.adj().coeffRef(k)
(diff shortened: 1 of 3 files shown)


spinkney:
Nice find! This reminds me that we should put in the ILR based simplex instead of stick breaking. My only hold up is that I like to keep the stick breaking code as well. We haven't come to a consensus on how we add a different parameterization of the same constraint type.

WardBrian:
@bob-carpenter has stated before that he's of the opinion that we just completely replace stick-breaking with the ILR transform for simplexes, rather than provide an option

Another ~easy option is to provide some compile time define to switch, rather than deciding on a language syntax level option

spinkney:
> @bob-carpenter has stated before that he's of the opinion that we just completely replace stick-breaking with the ILR transform for simplexes, rather than provide an option

The ILR one is super easy. We construct a zero_sum_vector (just like we have already) and then softmax it. There's a jacobian adjustment for the softmax.

```
vector[N - 1] y; // free param
vector[N] z = sum_to_zero_constrain(y);
vector[N] s = softmax(z); // simplex

jacobian_adj = -N * log_sum_exp(z) + 0.5 * log(N);
```
 
> Another ~easy option is to provide some compile time define to switch, rather than deciding on a language syntax level option

Whatever is easiest. I'd like to save all the transforms in some directory. We can have the default ones be the ones we think are "best" from what we've seen and all the other ones stored separately. I think this requires a lot of repo organizing though.

bob-carpenter:
We need to choose a default and I think this should be it.  If there are cases where we'd still want to use the old one, we can leave in the constraining/unconstraining functions for it and let people do it manually.

spinkney:
> We need to choose a default and I think this should be it. If there are cases where we'd still want to use the old one, we can leave in the constraining/unconstraining functions for it and let people do it manually.

Ah yea I saw that the constraint functions are getting exposed! Cool, that's works then. 

@WardBrian we can do this in 2 loops using the online softmax https://arxiv.org/abs/1805.02867. The first loop constructs the sum to zero, get the max value of that sum to zero vec, and returns the sum of exponentials with the max subtracted. The second loop does the safe exponential of the sum to zero vector with the max subtracted and divides by the sum of exponentials output from the first loop. The jacobian is output after.

```cpp
inline plain_type_t<Vec> simplex_ilr_constrain(const Vec& y, Lp& lp) {
  const auto N = y.size();

  plain_type_t<Vec> z = Eigen::VectorXd::Zero(N + 1);
  if (unlikely(N == 0)) {
    return z;
  }

  auto&& y_ref = to_ref(y);
  value_type_t<Vec> sum_w(0);

  // new   
  double d = 0;  // sum of exponentials
  double max_val = 0;
  double max_val_old = 0;

  for (int i = N; i > 0; --i) {
    double n = static_cast<double>(i);
    auto w = y_ref(i - 1) * inv_sqrt(n * (n + 1));
    sum_w += w;

    z.coeffRef(i - 1) += sum_w;
    z.coeffRef(i) -= w * n;
    
    // new
    max_val = max(max_val_old, z.coeff(i));
    d = d * exp(max_val_old - max_val) + exp(z.coeff(i) - max_val);
    max_val_old = max_val;
  }

  // new loop
  for (int i = 0; i < N; ++i) {
   z.coeffRef(i) = exp(z.coeff(i) - max_val) / d;
  }

  lp += -N * log(d) + 0.5 * log(N);

  return z;
 }
```

WardBrian:
@spinkney mind moving that to it's own issue?

I think this PR is worth it on its own, since we might not get to the ILR simplex this release, or if we leave the existing stick breaking as an option anyway

spinkney:
> @spinkney mind moving that to it's own issue?
> 
> I think this PR is worth it on its own, since we might not get to the ILR simplex this release, or if we leave the existing stick breaking as an option anyway

done #3170. I didn't see a way to move a comment so I'll hide the above.

PR #3177, opened 2025-04-21
Use Windows ARM64 for RTools CI

andrjohns:
## Summary

Windows ARM runners are now available for public repos, so we can add them to the CI. I've just replaced the existing windows runners, but can also add them separately if we want both x86-64 & arm64 rtools.

The `setup-r` action doesn't support arm64, so I've just added a step to manually install RTools

## Tests

## Side Effects

N/A

## Release notes

Added Windows ARM64 to the Github CI

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,8 +44,8 @@ jobs:
     - name: Download and install Rtools45
       if: runner.os == 'Windows'
       run: |
-        $ARCH = if (${{ matrix.os }} == 'windows-11-arm') { 'aarch64' } else { 'x86_64' }
-        $RTOOLS = if (${{ matrix.os }} == 'windows-11-arm') { 'rtools45-$ARCH' } else { 'rtools45' }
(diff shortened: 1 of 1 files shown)


WardBrian:
There are several orders of magnitude more x86 windows users than ARM, so if we do test this at all we should do it in addition rather than as a replacement

andrjohns:
> There are several orders of magnitude more x86 windows users than ARM, so if we do test this at all we should do it in addition rather than as a replacement

Ah yeah true, will add both

WardBrian (approved):

WardBrian:
It looks like sometimes these jobs get stuck installing rtools, e.g: https://github.com/stan-dev/math/actions/runs/14649317253/job/41111049475?pr=3171

1 hour and counting just in the install step

WardBrian:
I can't tell if it's a rate limiting issue or the installer thats hanging. If its the former we could add an authorization header to Invoke-WebRequest

PR #3183, opened 2025-05-08
Makefile: Disable Make's implict suffix rules

WardBrian:
## Summary
These rules can cause some very unintuitive behavior, see: https://www.gnu.org/software/make/manual/html_node/Implicit-Rules.html

Hopefully we aren't relying on them anywhere, I'm counting on Jenkins to tell me if so...

## Tests

None

## Side Effects

None 🤞🏻 

## Release notes

Disabled Make's implicit rules in our Makefiles.

## Checklist

### DIFF
--- a/make/compiler_flags
+++ b/make/compiler_flags
@@ -75,8 +75,7 @@ CXX_MINOR := $(word 2,$(subst ., ,$(CXX_VERSION)))
 # this should make Make slightly faster, as well as avoiding some unexpected
 # behavior when there are other files lying around with the same name as
 # Stan models
-# We do need to leave some rules in place for things like googletest to function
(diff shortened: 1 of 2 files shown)


syclik:
Unfortunately, I think we rely on a lot of implicit rules. My guess is that
we’re going to have to remove the rule from .r to whatever object file
instead of the proposed fix on this PR.

WardBrian:
It appears the only implicit rule we really depend on is the one for turning a `.cpp` file into a `.o` file in our tests. We could avoid that if we wanted to, but it seems like a relatively harmless entry to keep.

syclik (approved):
LGTM

PR #3206, opened 2025-06-15
[DRAFT/RFC] Gaussian Copula Cholesky LPDF

andrjohns:
## Summary

Opening this as a draft PR for feedback on the signature/approach.

For the Gaussian copula (and other copula families), the user needs to provide both the `y` variable and a means of transforming that variable to the unit-scale.

The method I've gone with is to require a tuple of the same length as `y`, where each element is itself a tuple - with a functor for computing the LCDF as the first element, and any additional args as the remaining.

For example, if the user wanted to model the correlation between a `Gamma(2, 1)` variable and a `Exp(2)` variable, they would pass the tuple-of-tuples:

```cpp
std::make_tuple(
    std::make_tuple(gamma_lcdf, 2, 1),
    std::make_tuple(exp_lcdf, 2)
)
```

So the final signature would be:

```
real gaussian_copula_cholesky_lpdf(vector | tuple-of-tuples, cholesky_factor_corr)
```

The current framework only supports continuous outcomes, but once we settle on a good approach I can expand to discrete outcomes by requiring that a uniform parameter for data-augmentation is also included in the tuple for that outcome.

## Tests

Basic `prim` and `mix` are added, but the `fwd` components of the `mix` tests are currently failing

## Side Effects

Extended a few of the utilities (`size`, `vector_seq_view`) for tuples

## Release notes

## Checklist

### DIFF
--- a/stan/math/prim/fun/size.hpp
+++ b/stan/math/prim/fun/size.hpp
@@ -34,7 +34,6 @@ inline int64_t size(const T& m) {
   return m.size();
 }
 
-
 template <typename T, require_tuple_t<T>* = nullptr>
 inline int64_t size(const T& /*t*/) {
   return std::tuple_size<std::remove_reference_t<T>>{};
--- a/stan/math/prim/fun/size.hpp
+++ b/stan/math/prim/fun/size.hpp
@@ -34,6 +34,12 @@ inline int64_t size(const T& m) {
   return m.size();
 }
 
+
+template <typename T, require_tuple_t<T>* = nullptr>
+inline int64_t size(const T& /*t*/) {
+  return std::tuple_size<std::remove_reference_t<T>>{};
+}
+
 }  // namespace math
 }  // namespace stan
 #endif
--- a/stan/math/prim/fun/size_mvt.hpp
+++ b/stan/math/prim/fun/size_mvt.hpp
@@ -36,8 +36,18 @@ int64_t size_mvt(const MatrixT& /* unused */) {
   return 1;
 }
 
+template <typename TupleT, require_tuple_t<TupleT>* = nullptr>
+int64_t size_mvt(const TupleT& /* unused */) {
+  return 1;
+}
+
 template <typename MatrixT, require_matrix_t<MatrixT>* = nullptr>
 int64_t size_mvt(const std::vector<MatrixT>& x) {
+  return x.size();
+}
+
+template <typename TupleT, require_tuple_t<TupleT>* = nullptr>
+int64_t size_mvt(const std::vector<TupleT>& x) {
   return x.size();
 }
 
--- a/stan/math/prim/fun/vector_seq_view.hpp
+++ b/stan/math/prim/fun/vector_seq_view.hpp
@@ -8,12 +8,13 @@
 namespace stan {
 namespace internal {
 template <typename T>
-using is_matrix_or_std_vector
-    = math::disjunction<is_matrix<T>, is_std_vector<T>>;
+using is_matrix_or_std_vector_or_tuple
+    = math::disjunction<is_matrix<T>, is_std_vector<T>, math::is_tuple<T>>;
 
 template <typename T>
 using is_scalar_container = math::disjunction<
     is_matrix<T>,
+    math::is_tuple<T>,
     math::conjunction<is_std_vector<T>, is_stan_scalar<value_type_t<T>>>>;
 }  // namespace internal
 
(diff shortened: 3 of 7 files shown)


andrjohns:
@spinkney would you have a look at the general signature/idea here? Not 100% that is the way to go, but I'm not sure of a better alternative

spinkney:
@andrjohns thanks for getting to this!

In the signature

```stan
real gaussian_copula_cholesky_lpdf(vector | tuple-of-tuples, cholesky_factor_corr)
```

Does this imply that it is looped over for each observation, that is for each $n \in N$ we have a tuple-of-tuples? I believe we also want to support an array vectors for the first argument.

> @spinkney would you have a look at the general signature/idea here? Not 100% that is the way to go, but I'm not sure of a better alternative

Yea, I see that there's a few different options with different pros and cons. I like the current implementation but I think it would be nice to have additional signatures. Here's a few questions and options:

1. Can we pass in an empty functor to signal that this variate is already on the Gaussian scale? Or is there another way to accomplish this?
2. Do we also want to support passing in the lpdf as an optional arugment? I just feel like we're already looping over the lcdf so can accomplish this in one pass too.
3. I would like to be able to choose or have the option to input a user defined inverse normal cdf function. Users may prefer to have a fast approximation or use a custom algorithm. It would be nice to support this but it's not necessary for this PR to go through.

To make this a bit more concrete:
```stan
data {
  int<lower=0> N;
  int<lower=1>  K;            // K different marginals
  array[K] int<lower=0> P;    // number of parameters for each k marginal
  array[N] vector[K] y;       // continuous outcome and an array of vector
}
transformed data {
  int P_total = to_int(sum(to_vector(P)));
}
parameters {
  vector[P_total] theta_free;
  cholesky_factor_corr[K] L;
}
model {
  tuple( tuple(real, ...), ..., tuple(real, ...) ) marginal_lcdf_tuples =  
   tuple(
     tuple(k1_lcdf, theta_free[1], theta_free[2], ...),
     tuple(k2_lcdf, theta_free[p], theta_free[p + 1], ...),
     ....,
     tuple(K_lcdf, ..., theta_free[P_total])
   );
 
  // Can the above also be any container for the parameters?
  //  such as
  //
  //  tuple( tuple(real, ...), ..., tuple(real, ...) ) marginal_tuples =  
  //   tuple(
  //     tuple(k1_lcdf, vector),
  //     tuple(k2_lcdf, array[] real),
  //     ....,
  //     tuple(K_lcdf, ..., matrix)
  //   );
  //
 
  // Why not also have another input for the lpdfs?
  //
  // tuple( tuple(real, ...), ..., tuple(real, ...) ) marginal_lpdf_tuples =  
  //  tuple(
  //    tuple(k1_lpdf, theta_free[1], theta_free[2], ...),
  //    tuple(k2_lpdf, theta_free[p], theta_free[p + 1], ...),
  //    ....,
  //    tuple(K_lpdf, ..., theta_free[P_total])
  //  );
  //

  y ~ gaussian_copula_cholesky(marginal_lcdf_tuples | L);
  
  // or optionally
  // y ~ 
  //  gaussian_copula_cholesky(
  //   marginal_lcdf_tuples | marginal_lpdf_tuples, L
  // );
  //
}
```

spinkney:
Just typing that out was tiresome and made me think another way 

```stan
// lcdf functors
 tuple( real, real, ... ) marginal_lcdf_tuples =  
  tuple( k1_lcdf, k2_lcdf, ..., K_lcdf );
 
// lpdf functors, same K number of objects in the tuple as marginal_lcdf_tuples
  tuple( real, real, ... ) marginal_lpdf_tuples =  
   tuple( k1_lpdf, k2_lpdf, ..., K_lpdf ); 

// parameters
// I'm assuming that the containers in this tuple would match the
// allowable parameters in their corresponding lcdf/lpdf functor signatures
  tuple( params_k1, ..., params_K) marginal_params_tuple =  
   tuple( k1_params, ..., K_params );

 y ~ gaussian_copula_cholesky(
         marginal_lcdf_tuples | 
         marginal_params_tuple,        
         marginal_lpdf_tuples, // optional, if omitted assumed you are doing this somewhere else
         L);
```

spinkney:
Is it legal to have a user defined `_lcdf/_lpdf` input into the signature as functor? I'm not up-to-date with the use of functors in Stan functions. 

@WardBrian @SteveBronder

WardBrian:
Yes, we currently allow lpdfs in reduce_sum, and I don’t think cdfs would require any different code generation than what we already have

andrjohns:
Closing, planning different impl

PR #3209, opened 2025-06-16
Laplace: move theta0 to tolerances

WardBrian:
Discussed in https://github.com/stan-dev/math/issues/3202#issuecomment-2968011809, counterpart to https://github.com/stan-dev/stanc3/pull/1521

## Summary

Theta0 is now a tolerance parameter rather than required. When not provided, we assume a vector of 0s is sufficient. 

This required a bit of re-organization to get the size right, but otherwise wasn't too bad because we call the covariance function pretty early and can get the size from the return of that function. 

## Tests

Existing tests updated

## Checklist

- [x] Copyright holder: (fill in copyright holder information)

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #3213, opened 2025-06-30
Compile Time Theta Default

SteveBronder:
## Summary

@WardBrian how do you feel about this alternative to #3209 ? Instead of an `std::optional` for `theta_0` I added a template to `laplace_options` for the type of `theta_0`. Since we know at compile time whether we are using the default `theta_0` we can save having to make an optional here.

## Tests

No new tests 

## Checklist

### DIFF
--- a/stan/math/mix/functor/laplace_marginal_density.hpp
+++ b/stan/math/mix/functor/laplace_marginal_density.hpp
@@ -27,7 +27,6 @@ namespace math {
 /**
  * Options for the laplace sampler
  */
-template <typename Theta>
 struct laplace_options_base {
   /* Size of the blocks in block diagonal hessian*/
   int hessian_block_size{1};
@@ -46,14 +45,22 @@ struct laplace_options_base {
   double tolerance{1e-6};
   /* Maximum number of steps*/
   int max_num_steps{100};
+};
+
+template <typename Theta, typename = void>
+struct laplace_options;
+
+template <typename Theta>
+struct laplace_options<Theta, require_eigen_t<Theta>> : public laplace_options_base {
 
   /* Initial value for theta. Defaults to 0s of the correct size if nullopt */
   Theta theta_0{0};
 };
 
-using laplace_options = laplace_options_base<Eigen::VectorXd>;
-using laplace_options_default = laplace_options_base<double>;
+template <typename Theta>
+struct laplace_options<Theta, require_not_eigen_t<Theta>> : public laplace_options_base {};
 
+using laplace_options_default = laplace_options<void>;
 namespace internal {
 
 template <typename Covar, typename ThetaVec, typename WR, typename L_t,
@@ -461,7 +468,7 @@ template <typename LLFun, typename LLTupleArgs, typename CovarFun,
 inline auto laplace_marginal_density_est(LLFun&& ll_fun, LLTupleArgs&& ll_args,
                                          CovarFun&& covariance_function,
                                          CovarArgs&& covar_args,
-                                         const laplace_options_base<ThetaVec>& options,
+                                         const laplace_options<ThetaVec>& options,
                                          std::ostream* msgs) {
   using Eigen::MatrixXd;
   using Eigen::SparseMatrix;
@@ -793,7 +800,7 @@ template <
 inline double laplace_marginal_density(LLFun&& ll_fun, LLTupleArgs&& ll_args,
                                        CovarFun&& covariance_function,
                                        CovarArgs&& covar_args,
-                                       const laplace_options_base<ThetaVec>& options,
+                                       const laplace_options<ThetaVec>& options,
                                        std::ostream* msgs) {
   return internal::laplace_marginal_density_est(
              std::forward<LLFun>(ll_fun), std::forward<LLTupleArgs>(ll_args),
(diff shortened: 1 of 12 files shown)


stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -510,9 +513,13 @@ inline auto laplace_marginal_density_est(LLFun&& ll_fun, LLTupleArgs&& ll_args,
...
   auto ll_args_vals = value_of(ll_args);
-  Eigen::VectorXd theta = options.theta_0.has_value()
-                              ? *options.theta_0
-                              : Eigen::VectorXd::Zero(theta_size);
+  Eigen::VectorXd theta = [theta_size, &options]() {
+    if constexpr (is_eigen_v<ThetaVec>) {
+      return options.theta_0;
+    } else {
+      return Eigen::VectorXd::Zero(theta_size);
+    }
```
WardBrian: Thoughts on making this a member function of the options struct? so this call would just be `options.initial_theta(theta_size);`, and the function body would basically be this code (plus maybe an assert that the size is really correct and/or moving the above checks into this function)

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -47,9 +48,12 @@ struct laplace_options {
   int max_num_steps{100};
 
   /* Initial value for theta. Defaults to 0s of the correct size if nullopt */
```
WardBrian: Comment is wrong now

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -47,9 +48,12 @@ struct laplace_options {
...
-  std::optional<Eigen::VectorXd> theta_0{std::nullopt};
+  Theta theta_0{0};
 };
 
+using laplace_options = laplace_options_base<Eigen::VectorXd>;
+using laplace_options_default = laplace_options_base<double>;
```
WardBrian: I find stuffing a double in here that we never use to be pretty unfortunate. There are some [ways of making this compile away entirely](https://brevzin.github.io/c++/2021/11/21/conditional-members/#conditional-member-variables) but they're also kind of ugly. 

Could this be a case where using inheritance makes sense?

stan/math/mix/prob/laplace_marginal_poisson_log_lpmf.hpp
```diff
@@ -96,10 +96,9 @@ inline auto laplace_marginal_poisson_log_lpmf(const std::vector<int>& y,
                                               CovarFun&& covariance_function,
                                               CovarArgs&& covar_args,
                                               std::ostream* msgs) {
-  const laplace_options ops{1, 1, 0, 1e-6, 100, std::nullopt};
   return laplace_marginal_density(
       poisson_log_likelihood{}, std::forward_as_tuple(y, y_index),
-      covariance_function, std::forward<CovarArgs>(covar_args), ops, msgs);
+      covariance_function, std::forward<CovarArgs>(covar_args), laplace_options_default{}, msgs);
```
WardBrian, line 101: I like these changes, though if I'm understanding correctly we could do this same thing on the other branch, right?

WardBrian (commented):
My overall thoughts are pretty neutral. It's nice that this avoids a runtime conditional, but it's only one conditional at the start of the optimization. On the other hand, I find the `laplace_options_base` struct much harder to read now, as we really don't semantically mean for it to be a templated struct, we just want it to either have or not-have the last member.

stan/math/mix/functor/laplace_marginal_density.hpp
SteveBronder: If we used it more than once I would say yes, but it is only used here so I'd rather just have the code here

stan/math/mix/prob/laplace_marginal_poisson_log_lpmf.hpp
SteveBronder: Yes this change could be done anywhere. I just did it here as it made sense.

stan/math/mix/functor/laplace_marginal_density.hpp
SteveBronder: Let me mess around. I think I can get rid of the double

stan/math/mix/functor/laplace_marginal_density.hpp
WardBrian: IMO it's less about the fact that it's used once and more about the fact that the immediately-evaluated-lambda trick is messy. But, I could live with it here

stan/math/mix/functor/laplace_marginal_density.hpp
SteveBronder: Mkay I think I have a nicer approach. We use a base laplace options, then we have a `laplace_options` with a template type for `Theta` If theta is a vector we include the `Theta theta_0` member. Otherwise we don't include the member at all

stan/math/mix/functor/laplace_marginal_density.hpp
```diff
@@ -45,11 +45,21 @@ struct laplace_options {
...
+template <typename Theta>
+struct laplace_options<Theta, require_eigen_t<Theta>> : public laplace_options_base {
+  /* Value for user supplied initial theta  */
+  Theta theta_0{0};
 };
 
+template <typename Theta>
+struct laplace_options<Theta, require_not_eigen_t<Theta>> : public laplace_options_base {};
+
+using laplace_options_default = laplace_options<void>;
```
WardBrian: Would it be possible just to have this be `template <bool HasInitTheta>` and then the two specializations just being `false` and `true`? 

Something like 

```c++
template <bool HasInitTheta>
struct laplace_options;

template <>
struct laplace_options<false> : public laplace_options_base {};

template <>
struct laplace_options<true> : public laplace_options_base {
  /* Value for user supplied initial theta  */
  Eigen::VectorXd theta_0{0};
};


using laplace_options_default = laplace_options<false>;
using laplace_options_theta0 = laplace_options<true>;

```

seems cleaner to me than require/require_nots and a `void`

WardBrian: It looks like this does work, see https://godbolt.org/z/GM4nqM81a

SteveBronder: I like it! Just updated

PR #3215, opened 2025-06-30
Use perfect forwarding for functions that use `apply_*_*` functions

### LINKED ISSUE #3208: Undefined behavior in combination of apply_scalar_binary/make_holder/to_ref_if
This issue is to collect information from https://github.com/stan-dev/math/pull/3203 in one place.

Basic summary:

On GCC, since c71dd3e (#2642), both `./test/prob/poisson/poisson_ccdf_log_00000_generated_v_test` and `./test/prob/loglogistic/loglogistic_cdf_00001_generated_ffv_test` fail due to taking a reference to something on the stack. The first fails under normal optimization settings consistently, the second is sporadic but can be made consistent by compiling with ASAN.

ASAN specifically blames
https://github.com/stan-dev/math/blob/3a196d410415ee2cc56c0e51ec73a2e1970a08a4/stan/math/prim/prob/poisson_lccdf.hpp#L55-L57

and

https://github.com/stan-dev/math/blob/3a196d410415ee2cc56c0e51ec73a2e1970a08a4/stan/math/prim/prob/loglogistic_cdf.hpp#L116

as the problematic lines. 

The discussion in https://github.com/stan-dev/math/pull/2414#issuecomment-1853299106 seems relevant, as @bgoodri was running into segfaults with a similar line in `poisson_lcdf`. 

There are shockingly few places where we use to_ref_if on an expression which is the result of an apply_scalar_binary call, I believe it is only in:

stan/math/prim/prob/loglogistic_cdf.hpp
stan/math/prim/prob/loglogistic_lpdf.hpp
stan/math/prim/prob/pareto_type_2_lcdf.hpp
stan/math/prim/prob/poisson_cdf.hpp
stan/math/prim/prob/poisson_lccdf.hpp
stan/math/prim/prob/poisson_lcdf.hpp
stan/math/prim/prob/weibull_lpdf.hpp

@SteveBronder believes the correct fix is probably to make the `apply_scalar_` functions _and all callsites thereof_ into perfect forwarding templates. The other solution involves materializing a signficant number of copies which would have a performance impact

SteveBronder:
## Summary

This fixes #3208 by using perfect forwarding for all functions that use our underlying apply family of functions for calling functions on containers and containers and containers. The issue was that the `Holder` class, when used in the apply functors, did not have enough type information to know which arguments it should take ownership of. 

Consider the following function, where all types are passed in via constant reference.

```cpp
inline auto gamma_p(const T1& a, const T2& b) {
  return apply_scalar_binary(
      [](const auto& c, const auto& d) { return gamma_p(c, d); }, a, b);
}
```

Calling this function with an Eigen expression that has a temporary in it would not give `apply_scalar_binary` and the `Holder` inside of `apply_scalar_binary` enough information to know that the `Holder` class should own any of the input arguments. As an example we can look at a simplified version of the code used in `poisson_lccdf.hpp`.  
```cpp
auto log_Pi = log(gamma_p(n_val + 1.0, lambda_val)));
double log_sum = sum(log_Pi);
```
`gamma_p` uses `apply_scalar_binary` and `log` uses `apply_scalar_unary`. We need to make sure the inputs and results of the `gamma_p` function do not fall out of scope by the time we go through `log` and then assign to `log_Pi`. Before this PR it would be possible for the expression `n_val + 1.0` to fall out of scope as well as the result of `gamma_p` to go out of scope from `log` after `log_Pi` is assigned.

To combat this we now use perfect forwarding for all of the functions that use our internal `apply` family of functors. This should allow the `Holder` used internally by the apply functors to know which types need to be owned by it to make sure things do not fall out of scope.

## Tests

There is no new tests for this. Since it is an isue on gcc I do wonder how we should test this in our CI/CD?

## Side Effects

I'd like to think of some test we can write so that, in the future, developers do not accidentally write functions that use the apply family of functors that do not use perfect forwarding.

## Release notes

Adds perfect forwarding to all functions that use the apply family of functors.

## Checklist


stan/math/fwd/fun/log_softmax.hpp
```diff
@@ -19,36 +19,38 @@ namespace math {
...
-      for (int k = 0; k < alpha.size(); ++k) {
-        if (m == k) {
-          log_softmax_alpha(k).d_
-              += alpha_ref.coeff(m).d_
-                 + negative_alpha_m_d_times_softmax_alpha_t_m;
-        } else {
-          log_softmax_alpha(k).d_ += negative_alpha_m_d_times_softmax_alpha_t_m;
+inline auto log_softmax(T&& x) {
+  return apply_vector_unary<T>::apply(
+      std::forward<T>(x), [&](const auto& alpha) {
```
andrjohns:
```suggestion
      std::forward<T>(x), [](auto&& alpha) {
```

The `apply_vector_unary` functors themselves should probably also perfect-forwarding, since they'll be passing their inputs to `apply_*` functions as well.

Also probably best to remove the reference-capture default while we're here

stan/math/fwd/functor/apply_scalar_unary.hpp
```diff
@@ -18,20 +18,23 @@ namespace math {
...
    * Apply the function specified by F to the specified argument.
    *
    * @param x Argument variable.
    * @return Function applied to the variable.
    */
-  static inline return_t apply(const fvar<T>& x) { return F::fun(x); }
+  template <typename T2>
+  static inline auto apply(const T2& x) {
+    return F::fun(x);
+  }
```
andrjohns:
```suggestion
  static inline auto apply(T2&& x) {
    return F::fun(std::forward<T2>(x));
  }
```

Should this also be forwarding? As the downstream calls are forwarding their arguments to `apply_scalar_unary<*>::apply()`

stan/math/prim/constraint/cholesky_corr_constrain.hpp
```diff
@@ -87,9 +87,11 @@ cholesky_corr_constrain(const EigVec& y, int K, Lp& lp) {
  * @param K The size of the matrix to return
  */
 template <typename T, require_std_vector_t<T>* = nullptr>
-inline auto cholesky_corr_constrain(const T& y, int K) {
-  return apply_vector_unary<T>::apply(
-      y, [K](auto&& v) { return cholesky_corr_constrain(v, K); });
+inline auto cholesky_corr_constrain(T&& y, int K) {
+  return apply_vector_unary<std::decay_t<T>>::apply(
```
andrjohns: Not something you necessarily need to change in this PR, but it might be a bit cleaner to move the `std::decay_t<T>` handling into `apply_vector_unary` itself

stan/math/prim/constraint/corr_constrain.hpp
```diff
@@ -43,7 +43,7 @@ inline plain_type_t<T> corr_constrain(const T& x) {
  * @param[in,out] lp log density accumulator
  */
 template <typename T_x, typename T_lp>
-inline auto corr_constrain(const T_x& x, T_lp& lp) {
+inline auto corr_constrain(T_x&& x, T_lp& lp) {
   plain_type_t<T_x> tanh_x = tanh(x);
```
andrjohns:
```suggestion
  plain_type_t<T_x> tanh_x = tanh(std::forward<T_x>(x));
```

stan/math/prim/constraint/lb_constrain.hpp
```diff
@@ -34,7 +34,7 @@ namespace math {
  */
 template <typename T, typename L, require_all_stan_scalar_t<T, L>* = nullptr,
           require_all_not_st_var<T, L>* = nullptr>
-inline auto lb_constrain(const T& x, const L& lb) {
+inline auto lb_constrain(T&& x, const L& lb) {
```
andrjohns, line 37: Looks like the the `std::forward<T>(x)` is missing from this function, and there are other container overloads in the file which probably need forwarding added as well

stan/math/prim/constraint/offset_multiplier_free.hpp
```diff
@@ -81,9 +85,9 @@ inline auto offset_multiplier_free(const std::vector<T>& x, const M& mu,
  * Overload for array of x and sigma and non-array mu
  */
 template <typename T, typename M, typename S,
+          require_all_std_vector_t<T, S>* = nullptr,
           require_not_std_vector_t<M>* = nullptr>
-inline auto offset_multiplier_free(const std::vector<T>& x, const M& mu,
-                                   const std::vector<S>& sigma) {
+inline auto offset_multiplier_free(T&& x, const M& mu, S&& sigma) {
```
andrjohns: I think `mu` also needs to be forwarded here since it's passed to `to_ref` below

stan/math/prim/constraint/offset_multiplier_free.hpp
```diff
@@ -56,21 +58,23 @@ inline auto offset_multiplier_free(const T& y, const M& mu, const S& sigma) {
   check_finite("offset_multiplier_constrain", "offset", value_of(mu_ref));
   check_positive_finite("offset_multiplier_constrain", "multiplier",
                         value_of(sigma_ref));
-  return stan::math::eval(divide(subtract(y, mu_ref), sigma_ref));
+  return stan::math::eval(
+      divide(subtract(std::forward<T>(y), std::forward<M>(mu_ref)),
+             std::forward<S>(sigma_ref)));
```
andrjohns:
```suggestion
      divide(subtract(std::forward<T>(y), std::forward<decltype(mu_ref)>(mu_ref)),
             std::forward<decltype(sigma_ref)>(sigma_ref)));
```

stan/math/prim/constraint/offset_multiplier_free.hpp
```diff
@@ -56,21 +58,23 @@ inline auto offset_multiplier_free(const T& y, const M& mu, const S& sigma) {
...
-inline auto offset_multiplier_free(const std::vector<T>& x, const M& mu,
-                                   const S& sigma) {
+inline auto offset_multiplier_free(T&& x, M&& mu, S&& sigma) {
   std::vector<plain_type_t<decltype(offset_multiplier_free(x[0], mu, sigma))>>
       ret;
   ret.reserve(x.size());
-  const auto& mu_ref = to_ref(mu);
-  const auto& sigma_ref = to_ref(sigma);
+  auto&& mu_ref = to_ref(std::forward<M>(mu));
+  auto&& sigma_ref = to_ref(std::forward<S>(sigma));
```
andrjohns, line 77: Should these also be forwarded in the `offset_multiplier_free` call below?

stan/math/prim/constraint/prob_constrain.hpp
```diff
@@ -48,10 +48,10 @@ inline T prob_constrain(const T& x) {
  * @return result constrained to fall in (0, 1)
  */
 template <typename T>
-inline T prob_constrain(const T& x, T& lp) {
-  T log_inv_logit_x = log_inv_logit(x);
-  lp += log_inv_logit_x + log1m_inv_logit(x);
-  return exp(log_inv_logit_x);
+inline auto prob_constrain(T&& x, return_type_t<T>& lp) {
+  plain_type_t<T> log_inv_logit_x = log_inv_logit(x);
```
andrjohns, line 52:
```suggestion
  plain_type_t<T> log_inv_logit_x = log_inv_logit(std::forward<T>(x));
```

andrjohns (changes requested):
Overall big fan of the changes! Just a couple of general queries before a full review

stan/math/fwd/fun/log_softmax.hpp
SteveBronder: > The apply_vector_unary functors themselves should probably also perfect-forwarding, since they'll be passing their inputs to apply_* functions as well.

Can you clarify? I'm not seeing in this function how much forwarding can be done. I do the perfect forwarding in the actual code for `apply_vector_unary` etc if that is was you mean.

> Also probably best to remove the reference-capture default while we're here

Agree

WardBrian:
Re #3208, this branch fixes the issue in `test/prob/poisson/poisson_ccdf_log_00000_generated_v_test`, but I'm still getting ASAN failures in `test/prob/loglogistic/loglogistic_cdf_00001_generated_ffv_test` on line 116 of loglogistic_cdf.hpp
https://github.com/stan-dev/math/blob/f7ccc01dadbb83b2d62757904dde1f8bf3d477fa/stan/math/prim/prob/loglogistic_cdf.hpp#L115-L116

Could be related to #3147?

stan/math/prim/constraint/lb_constrain.hpp
SteveBronder: So for functions that just operate on scalars I don't think we need to worry about forwarding as much since those functions will immediately evaluate

WardBrian:
I do think we should weigh the odds that merging something of this scope during a release window would introduce more issues than it resolves (especially because the issues seem to have been present in the code for a couple years)

SteveBronder:
I agree this is a large PR. Once I found the bug in one place I realized it was kind of systematic to all of our functions that can take in an expression and then output an expression. Let's talk irl. I agree this is risky before a release, but also think the bug is kind of large and can also happen in user space

WardBrian:
Not denying it can happen to users, but as far as we know it hasn’t in the couple years it has been present. So, would it be that bad to hold the fix a couple months for the next release 

Anyway, especially if @andrjohns thinks he can review it, I’m not opposed to merging, just wanted to ask in the name of due diligence

SteveBronder:
Yes I agree that it feels risky. But this is a very odd bug and imo I don't feel great about having this in a release when we know it is there.

andrjohns:
> Not denying it can happen to users, but as far as we know it hasn’t in the couple years it has been present. So, would it be that bad to hold the fix a couple months for the next release
> 
> Anyway, especially if @andrjohns thinks he can review it, I’m not opposed to merging, just wanted to ask in the name of due diligence

Definitely happy to review! But given that the first 2.37 RC was already a month ago and we'd want another RC with pretty decent testing after this many files were touched, I'd just worry that it would push out the broader release schedule. Although I don't have a good sense of how severe/urgent the issue is, so I'm happy to go either way

WardBrian:
> But given that the first 2.37 RC was already a month ago and we'd want another RC with pretty decent testing after this many files were touched

The changes to laplace are going to require (at least) one more release candidate anyway. We got a bit caught out with the level of feedback and some travel by various developers which has stretched out the release

SteveBronder:
> Definitely happy to review! But given that the first 2.37 RC was already a month ago and we'd want another RC with pretty decent testing after this many files were touched, I'd just worry that it would push out the broader release schedule. Although I don't have a good sense of how severe/urgent the issue is, so I'm happy to go either way

I'd list the issue as severe. I don't think it is so sever we need a 1800 line PR to fix it though. This is mostly to figure out which pieces need changed. My goal here is to pinpoint the underlying issue and check and check which things need changed. Right now I think there are two main issues

1. Anything that uses the apply family of functions and can work over eigen matrices needs to use perfect forwarding
2. Anything that returns an expression needs to use perfect forwarding and return a `Holder` class.

This will touch a *lot*, but not as much as in this PR hopefully. To be able to replicate this issue I think we need to change the expression tests so that they generate use temp matrices. This is actually not that bad of a change. Right now it just generates them as a separate line like the following.

```cpp
auto matrix4 = stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>(0.4, 1);
auto matrix5 = stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>(0.4, 1);
auto result6 = stan::math::eval(stan::math::mdivide_right_spd(matrix0_expr2,matrix1_expr3));
auto result7 = stan::math::eval(stan::math::mdivide_right_spd(matrix4,matrix5));
```

But we just need to inline `matrix4/5` so that they are temporaries

```cpp
auto result6 = stan::math::eval(stan::math::mdivide_right_spd(matrix0_expr2,matrix1_expr3));
auto result7 = stan::math::eval(stan::math::mdivide_right_spd(
stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>(0.4, 1),
  stan::test::make_arg<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>>(0.4, 1)));
```

andrjohns:
Sounds good to me! I'll hold off on any reviews then until you've got 'final' PR

SteveBronder:
I think I have something smaller that fixes this. Not tiny but not as big as this

PR #3244, opened 2025-10-03
Add a few more type traits for use in stanc

WardBrian:
## Summary

This adds a couple more type traits to the meta folder:

Two for resolving https://github.com/stan-dev/stan/issues/3356:
- `tuple_elt<N, T>` is a wrapper around `std::tuple_elt` but returns void if N is too big for the tuple size of T
- `is_tuple_of_size<T, N>` is `is_tuple` with an additional check that the size is what we expect.

Two for a general cleanup of the code gen:
- `is_integral<T>` is just `std::is_integral<std::decay<T>>`
- same for `is_floating_point`

## Release notes

Added some more type traits for dealing with tuples.

## Checklist

### DIFF
--- a/stan/math/prim/meta/is_floating_point.hpp
+++ b/stan/math/prim/meta/is_floating_point.hpp
@@ -1,0 +1,29 @@
+#ifndef STAN_MATH_PRIM_META_IS_FLOATING_POINT_HPP
+#define STAN_MATH_PRIM_META_IS_FLOATING_POINT_HPP
+
+#include <type_traits>
+
+namespace stan {
+
+/**
+ * Checks if decayed type is a floating point type
+ * @tparam The type to check
+ * @ingroup type_trait
+ */
+template <typename T>
+using is_floating_point = std::is_floating_point<std::decay_t<T>>;
+
+template <typename T>
+inline constexpr bool is_floating_point_v = is_floating_point<T>::value;
+
+template <typename... Types>
+inline constexpr bool is_all_floating_point_v
+    = (is_floating_point_v<Types> && ...);
+
+template <typename... Types>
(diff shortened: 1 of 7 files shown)


WardBrian:
Lessons:
- You cannot use `std::tuple_element<N, T>` in an evaluated context unless you are confident `tuple_size_v<T> < N`, as this triggers static asserts otherwise
- You cannot use `std::tuple_size<T>::value` in an evaluated context unless you are sure `is_tuple_v<T>`, as this leads to errors about incomplete typenames.

With sufficient indirection it is possible to avoid both of these errors by moving them into unevaluated contexts guarded by the proper `requires` templates, but it is not obvious

stan/math/prim/meta/is_floating_point.hpp
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Checks if decayed type is a floating point type
+ * @tparam The type to check
+ * @ingroup type_trait
+ */
+template <typename T>
+using is_floating_point = std::is_floating_point<std::decay_t<T>>;
+
+template <typename T>
+inline constexpr bool is_floating_point_v = is_floating_point<T>::value;
```
SteveBronder: just so the compiler does not get confused about which
```suggestion
inline constexpr bool is_floating_point_v = stan::is_floating_point<T>::value;
```

SteveBronder: This goes for all the other places as well

SteveBronder (changes requested):
The only real change here is if we are going to overwrite something in the standard library, when we use it we should preface it with `stan::` so that we reduce the change the compiler does gets confused. But otherwise good!

SteveBronder (approved):

PR #3264, opened 2025-12-18
fix gamma lccdf for alpha > 30 and overflow

spinkney:
## Summary

This pr closes #3257. The `gamma_lccdf` experiences numerical instability for alpha approximately greater than 30. It’s due to gamma_q or grad_reg_inc_gamma. This pr rewrites the code to use gamma_p and grad_reg_lower_inc_gamma, which avoids potential instability in a tgamma(alpha) and digamma(alpha) call. It seems to sample a bit faster than just wrapping `gamma_lcdf` with `log1m_exp`.

## Tests
I've implemented new tests in the problematic regime.

## Release notes

The `gamma_lccdf` is made more robust for alpha values. The issue was overflow in the derivatives when alpha was greater than 30.

## Checklist

- [x] Copyright holder: Publicis Groupe
    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #3266, opened 2025-12-23
super stable gamma_lccdf

### LINKED ISSUE #3257: Fix `gamma_lccdf` actually
See https://discourse.mc-stan.org/t/possible-issue-with-gamma-lccdf/

The function `gamma_lccdf` results in numerical instability for alpha approximately greater than 30. It’s due to gamma_q or grad_reg_inc_gamma. I can rewrite this to use gamma_p and grad_reg_lower_inc_gamma, which avoids potential instability in a tgamma(alpha) and digamma(alpha) call. It seems to sample a bit faster than just wrapping gamma_lcdf with log1m_exp.

Will put the PR in this week.

spinkney:
## Summary

This pr closes #3257. The `gamma_lccdf` experiences numerical instability for alpha approximately greater than 30. It’s due to gamma_q or grad_reg_inc_gamma. This pr rewrites the code to use gamma_p and grad_reg_lower_inc_gamma, which avoids potential instability in a tgamma(alpha) and digamma(alpha) call. It seems to sample a bit faster than just wrapping `gamma_lcdf` with `log1m_exp`.

This supercedes and improves upon pr #3264.

## Tests
I've implemented new tests in the problematic regime.

## Release notes

The `gamma_lccdf` is made more robust for alpha values. The issue was overflow in the derivatives when alpha was greater than 30.

## Checklist

- [x] Copyright holder: Publicis Groupe

### DIFF
--- a/stan/math/prim/fun/log_gamma_q_dgamma.hpp
+++ b/stan/math/prim/fun/log_gamma_q_dgamma.hpp
@@ -48,8 +48,7 @@ struct log_gamma_q_result {
  */
 template <typename T_a, typename T_z>
 inline log_gamma_q_result<return_type_t<T_a, T_z>> log_gamma_q_dgamma(
-    const T_a& a, const T_z& z, int max_steps = 250,
-    double precision = 1e-16) {
+    const T_a& a, const T_z& z, int max_steps = 250, double precision = 1e-16) {
   using std::exp;
   using std::fabs;
   using std::log;
@@ -99,7 +98,8 @@ inline log_gamma_q_result<return_type_t<T_a, T_z>> log_gamma_q_dgamma(
     // For gradient, use: d/da log(Q) = (1/Q) * dQ/da
     // grad_reg_inc_gamma computes dQ/da
     const double Q_val = exp(result.log_q);
-    const double dQ_da = grad_reg_inc_gamma(a_dbl, z_dbl, tgamma(a_dbl), digamma(a_dbl));
+    const double dQ_da
+        = grad_reg_inc_gamma(a_dbl, z_dbl, tgamma(a_dbl), digamma(a_dbl));
     result.dlog_q_da = dQ_da / Q_val;
 
   } else {
@@ -111,7 +111,8 @@ inline log_gamma_q_result<return_type_t<T_a, T_z>> log_gamma_q_dgamma(
     // grad_reg_inc_gamma computes dQ/da
     const double Q_val = exp(result.log_q);
     if (Q_val > 0) {
-      const double dQ_da = grad_reg_inc_gamma(a_dbl, z_dbl, tgamma(a_dbl), digamma(a_dbl));
+      const double dQ_da
+          = grad_reg_inc_gamma(a_dbl, z_dbl, tgamma(a_dbl), digamma(a_dbl));
       result.dlog_q_da = dQ_da / Q_val;
     } else {
       // Fallback if Q rounds to zero - use asymptotic approximation
--- a/stan/math/prim/fun/log_gamma_q_dgamma.hpp
+++ b/stan/math/prim/fun/log_gamma_q_dgamma.hpp
@@ -8,6 +8,7 @@
 #include <stan/math/prim/fun/gamma_p.hpp>
 #include <stan/math/prim/fun/gamma_q.hpp>
 #include <stan/math/prim/fun/grad_reg_inc_gamma.hpp>
+#include <stan/math/prim/fun/inv.hpp>
 #include <stan/math/prim/fun/lgamma.hpp>
 #include <stan/math/prim/fun/log.hpp>
 #include <stan/math/prim/fun/log1m.hpp>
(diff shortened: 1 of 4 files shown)


stan/math/prim/prob/gamma_lccdf.hpp
```diff
@@ -4,30 +4,84 @@
...
 #include <stan/math/prim/fun/log.hpp>
+#include <stan/math/prim/fun/log1m.hpp>
 #include <stan/math/prim/fun/max_size.hpp>
 #include <stan/math/prim/fun/scalar_seq_view.hpp>
 #include <stan/math/prim/fun/size.hpp>
 #include <stan/math/prim/fun/size_zero.hpp>
-#include <stan/math/prim/fun/tgamma.hpp>
 #include <stan/math/prim/fun/value_of.hpp>
 #include <stan/math/prim/functor/partials_propagator.hpp>
+#include <stan/math/fwd/fun/lgamma.hpp>
```
syclik: The `stan/math/fwd/` headers shouldn't be included here. Only the `prim` ones.

stan/math/prim/prob/gamma_lccdf.hpp
```diff
@@ -4,30 +4,84 @@
...
+#include <stan/math/fwd/fun/value_of.hpp>
 #include <cmath>
 
 namespace stan {
 namespace math {
 
+namespace internal {
+
+template <typename T_a>
+inline auto log_q_gamma_cf(const T_a& a, const double x,
```
syclik: can we be more explicit about the return type? Is that `T_a`?

syclik (dismissed):
A few things to change. I'll try to update the branch accordingly; it's nothing about the math itself.

stan/math/prim/prob/gamma_lccdf.hpp
spinkney: You'll see what I'm doing a bit more clearly now with the doc string. If the shape parameter `a` is `fvar` we can get around a source of instability. 

It's a bit involved but the continued fraction method `log_q_gamma_cf` is used to compute `log(Q(a,x))`, in the code this is `log_Qn`, when `beta * y > alpha + 1`. Instead of deriving the derivative analytically, the code uses forward-mode automatic differentiation to compute it numerically. This derivative is computed on lines 190-193 and stored in `dlogQ_dalpha`. This is then set as the partial derivative for the alpha parameter in the reverse-mode tape on line 223 (`partials<1>`).

Computing the derivative of the continued fraction with respect to alpha analytically is complex because:

* The continued fraction has up to 250 iterations
* Each iteration depends on alpha in a non-trivial way
* The convergence criterion itself depends on alpha

I chose forward-mode AD because it's faster than reverse mode for this case since we only need the derivative with respect to one input `alpha` and the continued fraction is a scalar function of one output.

This compiles and the tests pass except for the dependencies check on having `fwd` includes. We could allow  `fwd` includes, though I'm not sure what other effects or issues this might present. If we don't do that, I'm not sure there are any other functions or things doing something like this. I'll ping @SteveBronder who might have an idea how to do this.

spinkney: Actually I think I have a fix.

WardBrian:
> This supercedes and improves upon pr https://github.com/stan-dev/math/pull/3264.

Does this mean that 3264 needs to be merged before this, or just that it should be closed?

spinkney:
> Does this mean that 3264 needs to be merged before this, or just that it should be closed?

That #3264 should be closed

stan/math/prim/fun/log_gamma_q_dgamma.hpp
```diff
@@ -0,0 +1,156 @@
...
+ * @tparam T_a Type of shape parameter a (double or fvar types)
+ * @tparam T_z Type of value parameter z (double or fvar types)
+ * @param a Shape parameter
+ * @param z Value at which to evaluate
+ * @param precision Convergence threshold
+ * @param max_steps Maximum number of continued fraction iterations
+ * @return log(Q(a,z)) with same type as T_a and T_z
+ */
+template <typename T_a, typename T_z>
+inline auto log_q_gamma_cf(const T_a& a, const T_z& z, double precision = 1e-16,
```
SteveBronder: Quick Q for clarify: Does this function only work for double types or should it be able to accept autodiff types as well? Reading the gamma_lccdf code I'm kind of confused on the branching logic.

stan/math/prim/prob/gamma_lccdf.hpp
```diff
@@ -51,61 +56,159 @@ inline return_type_t<T_y, T_shape, T_inv_scale> gamma_lccdf(
...
+    }
+
+    bool use_cf = beta_y > alpha_dbl + 1.0;
+    T_partials_return log_Qn;
+    [[maybe_unused]] T_partials_return dlogQ_dalpha = 0.0;
+    // Extract double values for the double-only continued fraction path.
+    [[maybe_unused]] const double beta_y_dbl = value_of(value_of(beta_y));
+    [[maybe_unused]] const double alpha_dbl_val = value_of(value_of(alpha_dbl));
+
+    if (use_cf) {
```
SteveBronder: I was having a hard time parsing this logic, could we do the branching logic for the autodiff types first and then do the `use_cf` on the inside? It looks like you need branching logic for both `var` and `fvar` types.

stan/math/prim/fun/log_gamma_q_dgamma.hpp
syclik: I'm seeing autodiff types being passed through. For example, L100 in `gamma_lcdf.hpp`:

```
log_Qn = internal::log_q_gamma_cf(alpha_dbl, beta_y);
```

where `const T_partials_return beta_y = beta_dbl * y_dbl;`. So it'll only be non-double when we're using higher order autodiff.

syclik:
Notes: add `stan::math::recover_memory()` to rev tests.

syclik (approved):
Thanks for putting in the PR!

WardBrian:
@spinkney this failed on the merge:
```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] AgradCcdfLogGamma_fv_16/AgradCcdfLogTestFixture/0.Function, where TypeParam = std::tuple<AgradCcdfLogGamma, std::tuple<double, stan::math::fvar<stan::math::var_value<double, void> >, double, empty, empty, empty> >
[  FAILED  ] AgradCcdfLogGamma_fv_20/AgradCcdfLogTestFixture/0.Function, where TypeParam = std::tuple<AgradCcdfLogGamma, std::tuple<double, stan::math::fvar<stan::math::var_value<double, void> >, stan::math::fvar<stan::math::var_value<double, void> >, empty, empty, empty> >

 2 FAILED TESTS

test/prob/gamma/gamma_ccdf_log_00000_generated_fv_test --gtest_output="xml:test/prob/gamma/gamma_ccdf_log_00000_generated_fv_test.xml" failed

```

If you have time to look at it shortly, great, otherwise we can revert and re-open the changes later

spinkney:
@syclik do you think this is related to any changes you made? I don't know when I can look into this.

PR #3278, opened 2026-02-07
Set thread QoS to USER_INITIATED on Apple Silicon

### LINKED ISSUE #3277: TBB worker threads scheduled to efficiency cores on Apple Silicon
# Issue: TBB worker threads scheduled to efficiency cores on Apple Silicon

## Summary

On Apple Silicon Macs, TBB worker threads are created with the default QoS (Quality of Service) class. macOS interprets this as "not user-facing work" and may schedule these threads to efficiency (E) cores even when performance (P) cores are available. This significantly degrades Stan's parallel performance.

## Description

Apple Silicon chips have heterogeneous cores:
- **Performance (P) cores**: Fast, for compute-intensive work
- **Efficiency (E) cores**: Slower (~3x), for background tasks

macOS uses QoS classes to decide core scheduling. The default QoS class signals "this work isn't urgent," allowing macOS to prefer E-cores to save power. For Stan's compute workloads, this is the wrong signal.

### Observed behavior

- Environment: macOS 26.2 (Tahoe), Apple M3 Ultra (24 P-cores, 8 E-cores)
- Stan model using `reduce_sum` with 12 threads per chain, 2 chains
- Initial CPU usage: ~800% per chain (threads on P-cores)
- After ~4 minutes: CPU usage drops to ~100-300% per chain (threads demoted to E-cores)
- P-cores sit idle while E-cores are saturated

This behavior appears more aggressive in macOS 26 (Tahoe) but affects all Apple Silicon Macs.

### Root cause

TBB creates worker threads without setting a QoS class, so they inherit the default. macOS sees long-running default-QoS threads as background work and demotes them to E-cores.

## Proposed fix

Stan Math already has a `task_scheduler_observer` in `stan/math/rev/core/init_chainablestack.hpp` that runs when TBB worker threads are created. Adding a call to `pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0)` in `on_scheduler_entry()` would signal to macOS that these are user-initiated compute threads.

This doesn't prevent E-core usage - it tells macOS to **prefer P-cores** when available. If all P-cores are busy (e.g., running more threads than P-cores), macOS can still use E-cores. The fix ensures P-cores aren't left idle while work runs slowly on E-cores.

```cpp
void on_scheduler_entry(bool worker) {
#ifdef __APPLE__
#if defined(__arm64__) || defined(__aarch64__)
    // Prefer performance cores for compute threads
    pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0);
#endif
#endif
    // ... existing AD tape initialization ...
}
```

## References

- [oneTBB issue #896: Apple Silicon QoS support](https://github.com/uxlfoundation/oneTBB/issues/896) - TBB developers acknowledged the issue but suggested users handle it via `task_scheduler_observer`
- [Apple QoS documentation](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/PrioritizeWorkAtTheTaskLevel.html)

## Environment

- macOS: 26.2 (Tahoe), also affects earlier versions
- Hardware: Apple M3 Ultra (also affects M1, M2, M4 series)
- Stan Math: 2.38.0 (TBB 2020.3)

ssp3nc3r:
## Summary

On Apple Silicon Macs, TBB worker threads are created with the default QoS class, which macOS may schedule to efficiency cores even when performance cores are available. This significantly degrades parallel performance.

This adds a `pthread_set_qos_class_self_np()` call in `on_scheduler_entry()` to set `USER_INITIATED` QoS, signaling to macOS that these are compute threads the user is waiting for. This causes macOS to prefer performance cores when available.

## Details

- Adds `#include <pthread.h>` and `#include <sys/qos.h>` on Apple
- Calls `pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0)` when TBB worker threads enter the scheduler
- Only applies to Apple Silicon (`__arm64__` or `__aarch64__`)
- No behavioral change on other platforms

## Testing

Tested on macOS 26.2 (Tahoe) with Apple M3 Ultra (24 P-cores, 8 E-cores).

Before: CPU usage drops from ~800% to ~100-300% per chain after ~4 minutes (threads demoted to E-cores)
After: CPU usage remains stable on P-cores

Fixes #3277

### DIFF
--- a/stan/math/rev/core/init_chainablestack.hpp
+++ b/stan/math/rev/core/init_chainablestack.hpp
@@ -11,6 +11,11 @@
 #include <thread>
 #include <tuple>
 
+#ifdef __APPLE__
+#include <pthread.h>
+#include <sys/qos.h>
+#endif
+
 namespace stan {
 namespace math {
 
@@ -19,6 +24,10 @@ namespace math {
  * TBB scheduler adds a new thread to the TBB managed threadpool. This
  * hook ensures that each worker thread has an initialized AD tape
  * ready for use.
+ *
+ * On Apple Silicon, this also sets the thread QoS class to
+ * USER_INITIATED so that macOS prefers scheduling compute threads
+ * on performance cores rather than efficiency cores.
  *
  * Refer to
  * https://software.intel.com/content/www/us/en/develop/documentation/tbb-documentation/top/intel-threading-building-blocks-developer-reference/task-scheduler/taskschedulerobserver.html
@@ -37,6 +46,13 @@ class ad_tape_observer final : public tbb::task_scheduler_observer {
   ~ad_tape_observer() { observe(false); }
 
   void on_scheduler_entry(bool worker) {
+#ifdef __APPLE__
+#if defined(__arm64__) || defined(__aarch64__)
+    // Set thread QoS to USER_INITIATED so macOS prefers scheduling
+    // TBB worker threads on performance cores rather than efficiency cores.
+    pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0);
+#endif
+#endif
     std::lock_guard<std::mutex> thread_tape_map_lock(thread_tape_map_mutex_);
     const std::thread::id thread_id = std::this_thread::get_id();
     if (thread_tape_map_.find(thread_id) == thread_tape_map_.end()) {


WardBrian:
Changes look good and relatively self contained to me, but I don't have a machine to test

 
@bob-carpenter is our resident Mac Silicon fan and has used some of this functionality recently, so maybe he can take a peek before we'd merge

stan/math/rev/core/init_chainablestack.hpp
```diff
@@ -37,6 +46,13 @@ class ad_tape_observer final : public tbb::task_scheduler_observer {
   ~ad_tape_observer() { observe(false); }
 
   void on_scheduler_entry(bool worker) {
+#ifdef __APPLE__
+#if defined(__arm64__) || defined(__aarch64__)
+    // Set thread QoS to USER_INITIATED so macOS prefers scheduling
+    // TBB worker threads on performance cores rather than efficiency cores.
+    pthread_set_qos_class_self_np(QOS_CLASS_USER_INITIATED, 0);
+#endif
```
SteveBronder, line 54: Leaving this as is should be fine, but `pthread_set_qos_class_self_np` should work for x86 as well.

bob-carpenter:
I only added those statements after @WardBrian or @SteveBronder told me about them.  I haven't profiled without, which I should probably do.  We `ifdef`-ed them so they would be no-ops other than on Apple OS.

WardBrian:
@bob-carpenter you might have gotten lost in the notification soup, this is about adding similar settings to the Math library that you had previously set in walnuts

bob-carpenter:
I may have wrongly assumed people reading this knew I was working on WALNUTS, but I understood the context here.  @WardBrian pinged me and I was just responding that I haven't even verified that the commands do anything useful in the place where I am using them.  You might want to ask @SteveBronder, who I believe is the one who recommended these commands for Apple Silicon where there are a combination of "performance" cores and "efficiency" cores.

SteveBronder:
This looks good to me, but I'd like to run it on my mac first to make sure all is good. Once it works locally for me I'll approve!

SteveBronder:
Hi sorry I have not merged this yet, just wanted to update. I totally forgot to run this on my mac, but Monday I will be wfh and will do that. Then this should be good to merge!

SteveBronder (approved):
lgtm, thanks!

PR #3291, opened 2026-03-07
Reuse direct reverse mode autodiff specializations for known laplace distributions

### LINKED ISSUE #3287: Reuse direct reverse mode autodiff specializations for known laplace distributions
3f310e79f58c315f148540f0938a6c3f4ea09ac7 is a commit that has the specialized reverse mode functions for the explicit laplace distributions. for instance the file [here](https://github.com/stan-dev/math/commit/3f310e79f58c315f148540f0938a6c3f4ea09ac7#diff-e009125f8c76f0f3d07ebe55b87297775b9a716a442942f24885eb15ff4b5a43) that has reverse mode specializations for `neg_binomial_2`. Right now we use embedded reverse and higher order autodiff to calculate these from the likelihood function. If we need to speedup laplace one of the first places we should look is adding these back.

ishaan-arora-1:
I've implemented a fix for the issue. I found that we can bypass the expensive nested autodiff by providing specialized `diff()` and `third_diff()` methods directly within the likelihood functors like `neg_binomial_2_log`.

I think this should be a good fix considering we've added type traits (e.g., `has_custom_diff`) to compile-time dispatch these specialized methods without breaking the generic autodiff fallback for other distributions. Computations derived from the original implementation (commit 3f310e7) were adapted to match the current API structure, which should bring back the performance gains.

Resolves #3287

### DIFF
--- a/stan/math/mix/functor/laplace_likelihood.hpp
+++ b/stan/math/mix/functor/laplace_likelihood.hpp
@@ -31,9 +31,8 @@ template <typename F, typename = void>
 struct has_custom_diff : std::false_type {};
 
 template <typename F>
-struct has_custom_diff<F,
-    std::void_t<decltype(std::declval<const F&>().diff(
-        std::declval<const Eigen::VectorXd&>(), 1))>>
+struct has_custom_diff<F, std::void_t<decltype(std::declval<const F&>().diff(
+                              std::declval<const Eigen::VectorXd&>(), 1))>>
     : std::true_type {};
 
 template <typename F>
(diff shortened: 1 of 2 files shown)


SteveBronder:
This is cool! Can you do a benchmark of this vs the old version to see what the difference in performance is?

SteveBronder:
Also, do the benchmark first, but if the benchmark works then I think I'd prefer having two overloads instead of an if constexpr here as the if statements / nested if statements can be a bit much for readability. See the example below

https://godbolt.org/z/bjhrdKo4E

I have this repo for benchmarking stan stuff that may be useful, but we also have google benchmark

https://github.com/SteveBronder/stan-perf

PR #3068, opened 2024-05-13
[WIP] Add cmake build scheme

SteveBronder:
## Summary

This adds a cmake build system for Stan Math. The main features are

1. This works with cmakes `FetchContent` module. So users can call the following to include stan math into their library that also uses cmake 

```cmake
FetchContent_Declare(
  stanmath
  GIT_REPOSITORY https://github.com/stan-dev/math
  GIT_TAG  develop
)

// later for building their executable
target_link_libraries(my_ex stanmath::stanmath)
```

Importantly, this allows stan and cmdstan etc to do the above which is nice.

2. tests can all be run with ctest. Currently only batched tests by folder are supported but I'll add individual targets for each test in this PR as well. 

It seems like this takes much less time to compile per folder. For example the `test/unit/math/mix/fun` tests all compile into one object file which took about 12 minutes on my machine using 24 cores (and at max about 65 gb of memory)

```
time make -j24 test_unit_math_mix_fun && ./test_prob_weibull
real	12m34.919s
user	247m45.300s
sys	7m40.696s
```

The prob tests also work and are broken down by folder

```
make -j28 test_prob_weibull && ./test_prob_weibull
```

I'd like to make targets for all of `unit` and all of `prob` to put into jenkins.

The expression tests are not setup yet but I'll have them done before this pr is merged.

A few other things that need to be done / checked before merge

- [ ] Opencl tests
- [ ] mpi tests
- [ ] threading tests

## Tests

I'm not sure how to test the cmake files but am open to suggestions

## Release notes

Replace this text with a short note on what will change if this pull request is merged in which case this will be included in the release notes.

## Checklist

- [x] Copyright holder: Steve Bronder

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

[...]

PR #3293, opened 2026-03-11
Adds a check to make sure AgradRev is used for all tests in test/unit/math/rev

SteveBronder:
## Summary
This PR does two things that put us on the path to getting cmake in the math library
See the comment [here](https://github.com/stan-dev/math/pull/3068#issuecomment-3335551899) for the full todo list.

1. This cleans up a few tests and makes sure that all tests in `test/unit/math/rev` use `TEST_F(AgradRev` instead of plain `TEST(` so that the autodiff stack is cleaned after each test run.
2. Extends `test/unit/multiple_translation_units` test to make sure that it is safe to merge two TUs that both link to the Stan math library.

We need (1) because the cmake makes one large binary for the folders we want to test. 
(2) ensures that when linking multiple smaller object files via cmake we satisfy the One Definition Rule (ODR). The One Definition Rule says that across translation units (.o files),  classes/structs and non-inline functions cannot have more than one definition in the entire program and templates and types cannot have more than one definition by translation unit.

## Tests

No new tests

## Side Effects

The `make test-math-dependencies` now has an extra error it can generate in CI that tells the developer that they need to use `TEST_F(AgradRev` when a new rev test is created.

## Checklist

### DIFF
--- a/make/tests
+++ b/make/tests
@@ -38,7 +38,7 @@ endif
 
 ##
 # Adding a test for multiple translation units. If this fails,
-# a new function is probably missing an inline
+# a public umbrella header likely exposes a new non-inline external definition.
 ##
 
 ifneq ($(OS),Windows_NT)
--- a/runChecks.py
+++ b/runChecks.py
@@ -142,6 +142,47 @@ def check_non_test_files_in_test():
     return errors
 
 
+def check_rev_test_fixtures():
+    test_files = [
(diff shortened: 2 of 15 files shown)


WardBrian (approved):
Looks good. We might want to extract the comment-ignoring logic in runChecks into a helper one day, but I'm fine with duplicating for the second instance of it.

PR #3285, opened 2026-02-22
add yule_simon_rng

lingium:
## Summary

With this PR the rng of Yule-Simon distribution are added.
See issue #3239 

For $$Y\sim \text{Yule-Simon}(\alpha)$$:

$$
\Pr(Y=y\mid \alpha)= \alpha \mathrm{B}(y,\alpha+1) \quad y\in\{1,2,\dots\}
$$

$$\mathrm{B}$$ is the Beta function.

It has a Beta–Geometric mixture. To sample form $$Y$$, we do:

$$
\begin{align}
Y\mid p &\sim \text{Geometric}(p)\\
p=e^{-W} &\sim \text{Beta}(\alpha,1)\\
W &\sim\text{Exp}(\alpha)
\end{align}
$$

Stan does not have Geometric yet so we use `neg_binomial_rng(1.0, odds_ratio_p)`.

## Tests

Test is written follow the [guide](https://mc-stan.org/math/md_doxygen_2contributor__help__pages_2adding__new__distributions.html).

## Side Effects

No.

## Release notes

`yule_simon_rng` is available if merged.

## Checklist

- [x] Copyright holder: Zhi Ling

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #3283, opened 2026-02-18
add yule_simon_cdf

lingium:
## Summary

With this PR the cdf of Yule Simon distribution are added.
See issue #3237 

Expressions involved:

same as #3276 

$$
\begin{align}
P(Y \leq y) = F(y, \alpha) = 1 - \frac{\Gamma(\alpha + 1) \Gamma(y + 1)}{\Gamma(\alpha + y + 1)} , y = 1 , 2 , . . . , \alpha > 0
\end{align}
$$

## Tests

Test is written follow the [guide](https://mc-stan.org/math/md_doxygen_2contributor__help__pages_2adding__new__distributions.html).

## Side Effects

No

## Release notes

`yule_simon_cdf` is available if merged.

## Checklist

- [x] Copyright holder: Zhi Ling

    The copyright holder is typically you or your assignee, such as a university or company. By submitting this pull request, the copyright holder is agreeing to the license the submitted work under the following licenses:
      - Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
      - Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

- [x] the basic tests are passing

    - unit tests pass (to run, use: `./runTests.py test/unit`)
    - header checks pass, (`make test-headers`)
    - dependencies checks pass, (`make test-math-dependencies`)
    - docs build, (`make doxygen`)
    - code passes the built in [C++ standards](https://github.com/stan-dev/stan/wiki/Coding-Style-and-Idioms) checks (`make cpplint`)

- [x] the code is written in idiomatic C++ and changes are documented in the doxygen

- [x] the new changes are tested

PR #3299, opened 2026-03-24
Add geometric distribution (lpmf, cdf, lcdf, lccdf, rng)

### LINKED ISSUE #3098: Add Geometric distribution
## Description
Add support for the geometric distribution and the corresponding GLM. It's a special case of the negative binomial distribution (when n=1 in the usual parametrization) which is already supported, so I believe it should be relatively straightforward to implement. Arguably, geometric distribution is not some exotic distribution, so users may expect to have that available.

The expected advantages of having this are:

- Prettier/better readable stan code
- Possibly some performance gain compared to using the NB distribution

yasumorishima:
## Summary

Implements the geometric distribution as requested in #3098.

- **Parameterization**: Number of failures before first success (support {0, 1, 2, ...})
- **PMF**: P(n | θ) = θ(1 − θ)^n
- **Consistent with neg_binomial**: `geometric_lpmf(n, θ) == neg_binomial_lpmf(n, 1, θ/(1−θ))` — verified numerically
- Full autodiff support for all functions
- Boundary handling for θ = 0 and θ = 1

## Files added/modified

- `stan/math/prim/prob/geometric_lpmf.hpp` — log PMF with autodiff gradients
- `stan/math/prim/prob/geometric_cdf.hpp` — CDF with autodiff gradients
- `stan/math/prim/prob/geometric_lcdf.hpp` — log CDF with autodiff gradients
- `stan/math/prim/prob/geometric_lccdf.hpp` — log CCDF with autodiff gradients
- `stan/math/prim/prob/geometric_rng.hpp` — RNG via inverse CDF method
- `stan/math/prim/prob.hpp` — include additions (alphabetical order)
- `test/unit/math/prim/prob/geometric_test.cpp` — unit tests (8 tests)
- `test/prob/geometric/geometric_cdf_test.hpp` — CDF gradient test fixture

## Tests

- 8 unit tests passing:
  - `errorCheck` — validates parameter constraints via test rig
  - `distributionCheck` — RNG samples match PMF (chi-squared test)
  - `error_check` — domain errors for invalid θ
  - `lpmf_values` — hand-calculated PMF values including θ=1 edge case
  - `cdf_values` — hand-calculated CDF values
  - `cdf_below_support` — returns 0 / -inf for n < 0
  - `lccdf_values` — hand-calculated CCDF values
  - `rng_theta_one` — θ=1 always returns 0

## Notes

- The geometric distribution is a special case of the negative binomial with r = 1, as noted in the issue
- Follows the same implementation pattern as the recently merged Yule-Simon distribution (#3283, #3285)
- Gradient computations are guarded against division by zero at θ = 0 and θ = 1 boundaries

Closes #3098

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SteveBronder:
Hi @yasumorishima ! Is there a reason we can't just have direct calls to the negative binomial function for all of these? So these would just be calls to neg_binomial that fix `alpha` to 1

yasumorishima:
Hi @SteveBronder, thanks for the review!

Good point — I've refactored all five functions (lpmf, cdf, lcdf, lccdf, rng)
to delegate to the corresponding neg_binomial functions with α = 1 and
β = θ / (1 − θ).

The only special case is θ = 1, where β diverges. Each function handles this
with an early return (e.g., lpmf returns 0 for n = 0 and −∞ for n > 0,
cdf returns 1, etc.).

I also added autodiff test HPPs for lpmf, lcdf, and lccdf — all 1,412 tests
pass (unit + v/ffv/vv for all four functions).

Let me know if you'd like any further changes!

stan/math/prim/prob/geometric_lpmf.hpp
```diff
@@ -0,0 +1,98 @@
...
+    beta.reserve(stan::math::size(theta));
+    for (size_t i = 0; i < stan::math::size(theta); i++) {
+      beta.push_back(theta_vec[i] / (1.0 - theta_vec[i]));
+    }
+    return neg_binomial_lpmf<propto>(n_ref, 1, beta);
+  } else {
+    const auto beta = elt_divide(theta_ref, subtract(1.0, theta_ref));
+    return neg_binomial_lpmf<propto>(n_ref, 1, beta);
+  }
+}
```
SteveBronder, line 119: So looking at this I would prefer we did one of two styles 

1. Completely delegate to `neg_binomial_*`. You do not need the `std::vector` here to accumulate the non 1 values. 
```cpp
template <typename T_n, typename T_prob>
inline return_type_t<T_prob> geometric_cdf(const T_n& n, const T_prob& theta) {
  using T_n_ref = ref_type_t<T_n>;
  using T_prob_ref = ref_type_t<T_prob>;
  static constexpr const char* function = "geometric_cdf";
  check_consistent_sizes(function, "Outcome variable", n,
                         "Success probability parameter", theta);
  decltype(auto) n_ref = as_array_or_scalar(to_ref(n));
  decltype(auto) theta_ref = as_array_or_scalar(to_ref(theta));
  check_bounded(function, "Success probability parameter", value_of(theta_ref),
                0.0, 1.0);
  if constexpr (is_stan_scalar_v<T_n>) {
    if (n_ref < 0) {
      return 0.0;
    }
  } else {
    const bool is_any_nonpositive = (n_ref.array() < 0).any();
    if (is_any_nonpositive) {
      return 0.0;
    }
  }
  if constexpr (is_stan_scalar_v<T_prob>) {
    if (theta_ref == 1.0) {
      return 1.0;
    } else {
      return neg_binomial_cdf(n, 1, theta_ref / (1.0 - theta_ref));
    }
  } else {
    return theta_ref.binaryExpr(n_ref, [](const auto& theta_, const auto& n_) { 
      if (theta_ == 1.0) {
        return return_type_t<T_prob>{1.0};
      } else {
        return neg_binomial_cdf(n_, 1, elt_divide(theta_, subtract(1.0, theta_)));
      }
    }).prod();
  }
}
```

2. A full autodiff impl with the adjoint calculations

```cpp
template <typename T_n, typename T_prob>
inline return_type_t<T_prob> geometric_cdf(const T_n& n,
                                           const T_prob& theta) {
  using T_partials_return = partials_return_t<T_prob>;
  using T_n_ref = ref_type_t<T_n>;
  using T_prob_ref = ref_type_t<T_prob>;
  static constexpr const char* function = "geometric_cdf";
  check_consistent_sizes(function, "Outcome variable", n,
                         "Success probability parameter", theta);
  if (size_zero(n, theta)) {
    return 1.0;
  }
  auto n_arr = as_array_or_scalar(to_ref(n));
  auto theta_arr = as_array_or_scalar(to_ref(theta));
  // Need to check theta > 0 && theta <= 1.0. Bounded checks >= and <= 
  check_bounded(function, "Success probability parameter", value_of(theta_arr),
                std::numeric_limits<double>::min(), 1.0);

  auto ops_partials = make_partials_propagator(theta_arr);
  if constexpr (is_stan_scalar_v<T_n>) {
    if (n_arr < 0) {
      return ops_partials.build(0.0);
    }
  } else {
    if (any(n_arr < 0)) {
      return ops_partials.build(0.0);
    }
  }
  auto log1m_theta = log1m(theta_arr);
  auto P_i = -expm1((n_arr + 1.0) * log1m_theta);
  auto P = prod(P_i);
  if constexpr (is_autodiff_v<T_prob>) {
    // d/dtheta [1 - (1 - theta)^(n + 1)]
    //   = (n + 1) * (1 - theta)^n
    //
    // The select handles theta == 1, n == 0, where the derivative is 1
    // and the log form would otherwise run into 0 * -inf.
    auto dP_dtheta = select(
        n_arr == 0, 1.0, (n_arr + 1.0) * exp(n_arr * log1m_theta));
    auto theta_partials = P * elt_divide(dP_dtheta, P_i);
    if constexpr (is_stan_scalar_v<T_prob>) {
      partials<0>(ops_partials) = sum(theta_partials);
    } else {
      partials<0>(ops_partials) = theta_partials;
    }
  }
  return ops_partials.build(P);
}
```

yasumorishima:
Thanks for the detailed review @SteveBronder! Went with option 2 — all four functions now use `make_partials_propagator` with `log1m` / `expm1` / `log1m_exp` and analytic partials, no `std::vector` accumulation. CDF follows your sketch exactly; lpmf/lcdf/lccdf use the same pattern.

Boundary handling: short-circuit `theta = 1` (lpmf for n>0, lccdf for n>=0), `theta = 0` (cdf/lcdf), and `n = INT_MAX` (lccdf, matching `neg_binomial_lccdf` for the upper-bound test fixture).

Tests: 180/180 autodiff distribution tests (var/vv/ffv) + 8/8 prim unit tests pass locally. Happy to iterate.

yasumorishima:
Addressed @coderabbitai's review on `geometric_lpmf.hpp` (theta=0 producing +inf partials): with `check_bounded(theta, 0.0, 1.0)` (inclusive), `theta = 0` was accepted but `inv(theta_arr)` then poisoned the autodiff partial with `+inf`.

In commit 31dcbd6ed5:
- All four distribution functions (lpmf/cdf/lcdf/lccdf) now use `check_positive_finite` + `check_less_or_equal(..., 1.0)`, matching the documented `(0, 1]` domain and the existing `geometric_rng` pattern.
- Removed the now-unreachable `if (any(theta_arr == 0.0))` short-circuits in `cdf`/`lcdf` and the dead `// For theta = 0:` comment in `lccdf`.
- Added a `theta_zero_throws` unit test covering all four functions.

Local results: prim unit test 9/9 PASS (incl. the new throw test), all 12 generated autodiff distribution test binaries PASS.

yasumorishima:
Round 2 of CodeRabbit findings + an alignment bug Opus surfaced during self-review (commit `b37a958359`):

**`geometric_lccdf.hpp`**
- Removed the `any(n_arr < 0)` early return that dropped contributions from positive `n_i` in mixed-sign vectors. `P(N > n_i) = 1` for `n_i < 0` is the multiplicative identity (`log 0`), not an absorber, so the term has to stay in the sum. Switched to a per-element `select(n < 0, 0, (n+1)*log1m(theta))` for both value and partials.
- Replaced the `any(theta_arr == 1.0)` blanket `-inf` short-circuit with a `scalar_seq_view` per-element loop, so `theta = 1` only forces `-inf` when paired with `n_i >= 0` at the *same* index. Mirrors the `lpmf` pattern. Previously `theta=[0.5, 1.0], n=[2, -1]` returned `-inf` instead of `3*log(0.5)`.

**`geometric_lpmf.hpp`**
- Added the `is_stan_scalar_v<T_prob>` guard around `partials<0>` so scalar `theta` paired with a vector `n` collapses to a scalar partial via `sum(...)`, matching the pattern already used in `cdf`/`lcdf`/`lccdf`.

**Tests (`geometric_test.cpp`)**
- `lccdf_vectorized_mixed_sign`: locks in the `n = [-1, 2], theta = 0.5 -> 3*log(0.5)` contract.
- `lccdf_vectorized_theta_one_alignment`: locks in the per-element `theta = 1` guard with `n = [2, -1], theta = [0.5, 1.0] -> 3*log(0.5)`.

prim 11/11 PASS, autodiff distribution lccdf v/vv/ffv 90/180/90 PASS.

stan/math/prim/prob/geometric_lcdf.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+  // log P(N <= n) = -inf for n < 0
+  if (any(n_arr < 0)) {
+    return ops_partials.build(NEGATIVE_INFTY);
+  }
+
+  // log_q = log((1 - theta)^(n + 1)) = (n + 1) * log1m(theta)
+  // log P_i = log(1 - q_i) = log1m_exp(log_q_i)
+  // For theta = 1: log_q = -inf, log1m_exp(-inf) = log(1 - 0) = 0
+  //   (correct: P = 1 when success is certain).
```
SteveBronder: For all the files here, it is nice to have docs for the gradients, but we normally put these as latex in the doxygen doc

yasumorishima: Moved the gradient docs into doxygen `\f[ ... \f]` blocks per the [adding_new_distributions guide](https://github.com/stan-dev/math/blob/develop/doxygen/contributor_help_pages/adding_new_distributions.md). Each of the four files now documents the distribution formula and the partial w.r.t. `theta` in the header; the inline math comments are gone. The remaining inline comments explain boundary-handling rationale (theta = 1, n = 0, INT_MAX guards) rather than the math itself.

Pushed in `f4cb872094`.

PR #3305, opened 2026-04-19
fix(reduce_sum_static): wrap Args with ref_type_t to materialize Eigen expressions

### LINKED ISSUE #3304: reduce_sum_static: Missing ref_type_t Wrapper
HIGH — Dangling References/UB  

### Files Affected

- `stan/lib/stan_math/stan/math/prim/functor/reduce_sum_static.hpp`
  - Return type call to `reduce_sum_impl`

### Root Cause Analysis

`reduce_sum` carefully wraps shared arguments with `ref_type_t<Args&&>` to materialize temporary Eigen expression templates before passing to TBB parallelization:

```cpp
// In reduce_sum (correct):
                                 ref_type_t<Args&&>...>()(...);
//                                         ^^^^^^^^^^^^^^ Wrapping!
```

But `reduce_sum_static` was missing this wrapper:

```cpp
// BEFORE (buggy):
                                 Args...>()(...);  // ← No ref_type_t!
```

**Why This Matters:**

Eigen uses expression templates — operations like `matrix.row(i)` don't create objects, they create lightweight proxy objects. These proxies hold references to the underlying matrix.

When an Eigen expression is passed as a shared argument:
```cpp
Eigen::MatrixXd M = ...;
std::vector<int> data = ...;

// This creates temporary expression: M.row(0)
auto result = reduce_sum_static<F>(
    data, grainsize, nullptr,
    M.row(0)  // ← Temporary Eigen::Block<MatrixXd> (expression template)
);
```

The `Eigen::Block` object (proxy) holds a reference to `M`. It's a temporary that will be destroyed immediately after the function call.

### The Dangling Reference Problem

```cpp
// Without ref_type_t:
void reduce_sum_impl(..., const Eigen::Block<MatrixXd>& row) {
  // row is a reference to a temporary that was destroyed!
  // Dangling reference → undefined behavior
  
  // In TBB worker threads (running asynchronously):
  for (size_t i = 0; i < num_chunks; ++i) {
    // Accessing row → reading from deallocated memory
    double val = row(0);  // ← Undefined behavior!
  }
}
```

### How ref_type_t Fixes It

```cpp
// ref_type_t materializes the expression
template<typename T>
using ref_type_t = ...;  // Specialization that:
// - For Eigen expressions: creates actual matrix/vector copy or view
// - For scalars: passes by value
// - For vectors: passes by value

// With ref_type_t<Eigen::Block<MatrixXd>&&>:
// → Deduces to actual VectorXd (materialized)
// → Copy is created, not just reference
// → Copy is valid for duration of reduce_sum_impl
```

### The Fix

```cpp
// AFTER (correct):
                                 ref_type_t<Args&&>...>()(...);
//                                         ^^^^^^^^^^^^^^ Added wrapping!
```

This ensures arguments are materialized (actual objects created, not just references) before being passed to the parallelized TBB workers.

**Before Fix:**
- Passing Eigen expression templates caused dangling references
- Error only manifests in TBB workers (asynchronous)
- Unpredictable crashes or memory corruption
- Very difficult to debug (race condition-like behavior)

**After Fix:**
- Expressions materialized safely
- No dangling references
- Deterministic behavior

tmchow:
## Summary

Closes #3304. `reduce_sum` wraps its shared arguments with `ref_type_t<Args&&>...` before handing them to TBB so that Eigen expression-template temporaries (e.g. `M.row(0)`, `matrix.col(i)`) materialize instead of being held as references to a destroyed temporary. `reduce_sum_static`'s `STAN_THREADS` branch was not updated in the same pass and still forwarded plain `Args...`.

A caller that passes a temporary Eigen expression as a shared argument to `reduce_sum_static` can therefore end up with a dangling reference once the temporary falls out of scope at the call site.

## Fix

`stan/math/prim/functor/reduce_sum_static.hpp`:

```cpp
-                                 Args...>()(std::forward<Vec>(vmapped), false,
-                                            grainsize, msgs,
-                                            std::forward<Args>(args)...);
+                                 ref_type_t<Args&&>...>()(
+    std::forward<Vec>(vmapped), false, grainsize, msgs,
+    std::forward<Args>(args)...);
```

This exactly mirrors the wrapping already in place in `reduce_sum.hpp` (see line 208).

## Notes

- `ref_type_t` is already in scope via `stan/math/prim/meta.hpp` (already `#include`d).
- No API change; the behavior only changes for callers that were relying on temporary Eigen expressions staying alive longer than the standard allows, which is exactly the UB being fixed.
- The non-threaded branch is unchanged: it either forwards directly to `ReduceFunction()` (no `reduce_sum_impl` instantiation) or returns zero.

Closes #3304

### DIFF
--- a/stan/math/prim/functor/reduce_sum.hpp
+++ b/stan/math/prim/functor/reduce_sum.hpp
@@ -46,7 +46,11 @@ struct reduce_sum_impl<ReduceFunction, require_arithmetic_t<ReturnType>,
   struct recursive_reducer {
     Vec vmapped_;
     std::stringstream msgs_;
-    std::tuple<Args...> args_tuple_;
+    // Materialize Eigen expression-template args (e.g. M.row(0)) at storage
+    // time via the implicit ref_type_t<Args> conversion. Owning the materialized
+    // form keeps the tuple safe across TBB worker splits even after callers'
+    // temporaries fall out of scope.
+    std::tuple<ref_type_t<Args>...> args_tuple_;
     return_type_t<Vec, Args...> sum_{0.0};
 
     recursive_reducer(Vec&& vmapped, std::ostream* msgs, Args&&... args)
@@ -205,7 +209,7 @@ inline auto reduce_sum(Vec&& vmapped, int grainsize, std::ostream* msgs,
 
 #ifdef STAN_THREADS
   return internal::reduce_sum_impl<ReduceFunction, void, return_type, Vec,
-                                   ref_type_t<Args&&>...>()(
+                                   Args...>()(
       std::forward<Vec>(vmapped), true, grainsize, msgs,
       std::forward<Args>(args)...);
 #else
--- a/stan/math/prim/functor/reduce_sum.hpp
+++ b/stan/math/prim/functor/reduce_sum.hpp
@@ -47,9 +47,9 @@ struct reduce_sum_impl<ReduceFunction, require_arithmetic_t<ReturnType>,
     Vec vmapped_;
     std::stringstream msgs_;
     // Materialize Eigen expression-template args (e.g. M.row(0)) at storage
-    // time via the implicit ref_type_t<Args> conversion. Owning the materialized
-    // form keeps the tuple safe across TBB worker splits even after callers'
-    // temporaries fall out of scope.
+    // time via the implicit ref_type_t<Args> conversion. Owning the
+    // materialized form keeps the tuple safe across TBB worker splits even
+    // after callers' temporaries fall out of scope.
     std::tuple<ref_type_t<Args>...> args_tuple_;
     return_type_t<Vec, Args...> sum_{0.0};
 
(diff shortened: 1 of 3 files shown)


SteveBronder:
Thanks! Though I think I would prefer if we used `ref_type_t` in the class and not in the template instantiation directly. So we would pass `Args&&` and then the class would handle any ownership logic

tmchow:
@SteveBronder Pushed f73f114c.

- moved `ref_type_t` into `reduce_sum_impl::recursive_reducer`. Both the prim arithmetic and rev var specializations now type `args_tuple_` as `std::tuple<ref_type_t<Args>...>`, so the materialization happens at storage time. 
- The `recursive_reducer` constructor still takes `Args&&...` and forwards into the tuple, where the implicit `ref_type_t<Args>` conversion materializes any Eigen expression-template temporaries. 
- Call sites in `reduce_sum.hpp` and `reduce_sum_static.hpp` now pass plain `Args...`.

Verified locally with `STAN_THREADS=true`: `test/unit/math/prim/functor/reduce_sum_test` and `test/unit/math/rev/functor/reduce_sum_test` all pass.

Jenkins i think will exercise the rest including the mix tests and the perf benchmarks.

tmchow:
@SteveBronder Pushed `4efff0c72f`. CI failure on the previous push was a real regression in the rev specialization, not an infra issue. 

Locally the failing expression test passes and the broader unit suite stays green:

```
test/expressions/tests*_test (Prim/Rev/Fwd) -> 3/3 PASS
test/unit/math/rev/functor/reduce_sum_test -> 9/9 PASS (also with STAN_THREADS=true)
test/unit/math/prim/functor/reduce_sum_test -> 32/32 PASS
```

tmchow:
@SteveBronder anything else I should do for this? Lmk

WardBrian:
Thanks @tmchow, your changes were incorporated as part of #3384

PR #3312, opened 2026-04-27
Add matrix/vector array reshaping helpers

### LINKED ISSUE #3233: FR: More reshaping functions
> It would help if we had these, which I think we should also add:
> 
> ```stan
> matrix to_matrix(array[] vector x);
> matrix to_matrix(array[] row_vector u);
> 
> array[] vector to_vector_array(matrix y);
> array[] row_vector to_row_vector_array(matrix v);
> ``` 

 _Originally posted by @bob-carpenter in [stan#3354](https://github.com/stan-dev/stan/issues/3354#issuecomment-3275384937)_

fatimmajumder:
## Summary

Fixes stan-dev/math#3233.

This PR adds the remaining reshaping helpers requested in the issue:

- `to_matrix(std::vector<Eigen::Matrix<T, Eigen::Dynamic, 1>>)` for array-of-column-vector reshaping
- `to_vector_array(matrix)`
- `to_row_vector_array(matrix)`

The existing array-of-row-vector `to_matrix` overload was already present, so this PR leaves that behavior unchanged and adds tests for it.

The implemented semantics are:

- array of column vectors -> matrix columns, with `result(i, j) = x[j](i)`
- matrix -> vector array returns one column vector per matrix column
- matrix -> row-vector array returns one row vector per matrix row

Empty top-level arrays follow existing `to_matrix` behavior and return `0 x 0`. Zero-length contained vectors preserve the inferable dimension.

## Tests

Passed locally:

```bash
./runTests.py test/unit/math/prim/fun/to_matrix_test.cpp test/unit/math/prim/fun/to_vector_array_test.cpp test/unit/math/prim/fun/to_row_vector_array_test.cpp
```

Also passed targeted header tests:

```bash
make stan/math/prim/fun/to_matrix.hpp-test
make stan/math/prim/fun/to_vector_array.hpp-test
make stan/math/prim/fun/to_row_vector_array.hpp-test
make stan/math/prim/fun.hpp-test
```

I did not run the complete `make test-headers`; I started it, but it was traversing unrelated forward-mode headers, so I stopped it and ran the targeted header isolation tests above instead.

## Side Effects

No intentional side effects.

I intentionally did not add ragged-vector validation because the existing array-of-row-vector overload does not appear to validate ragged inputs either, and I did not want to introduce a new policy for only one direction of the reshape.

## Release notes

Adds reshaping helpers for converting between matrices and arrays of Eigen vectors/row vectors.

## Checklist

- [x] Copyright holder: Fatim Majumder
- [x] The basic tests are passing
- [x] The code is written in idiomatic C++ and changes are documented in the doxygen

### DIFF
--- a/stan/math/prim/fun/to_matrix.hpp
+++ b/stan/math/prim/fun/to_matrix.hpp
@@ -41,6 +41,29 @@ inline auto to_matrix(EigVec&& matrix) {
 
 /**
  * Returns a matrix representation of a standard vector of Eigen
+ * column vectors. Each input vector becomes a column in the result.
+ *
+ * @tparam T type of the elements in the vector
+ * @param x standard vector of Eigen column vectors
+ * @return the matrix representation of the input
+ */
+template <typename T>
+inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> to_matrix(
+    const std::vector<Eigen::Matrix<T, Eigen::Dynamic, 1>>& x) {
+  int cols = x.size();
+  if (cols == 0) {
+    return {};
+  }
+  int rows = x[0].size();
+  Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> result(rows, cols);
+  for (int j = 0; j < cols; ++j) {
+    result.col(j) = x[j];
+  }
+  return result;
+}
+
+/**
+ * Returns a matrix representation of a standard vector of Eigen
  * row vectors with the same dimensions and indexing order.
  *
  * @tparam T type of the elements in the vector
--- a/stan/math/prim/fun/to_vector_array.hpp
+++ b/stan/math/prim/fun/to_vector_array.hpp
@@ -1,0 +1,33 @@
+#ifndef STAN_MATH_PRIM_FUN_TO_VECTOR_ARRAY_HPP
+#define STAN_MATH_PRIM_FUN_TO_VECTOR_ARRAY_HPP
+
+#include <stan/math/prim/meta.hpp>
+#include <stan/math/prim/fun/Eigen.hpp>
+#include <vector>
+
+namespace stan {
+namespace math {
+
+/**
+ * Returns a standard vector of Eigen column vectors from the columns of
+ * the input matrix.
+ *
+ * @tparam EigMat type of the input matrix
+ * @param matrix input matrix
+ * @return the array of column vectors representation of the input
+ */
+template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
+inline std::vector<Eigen::Matrix<value_type_t<EigMat>, Eigen::Dynamic, 1>>
+to_vector_array(const EigMat& matrix) {
+  using T = value_type_t<EigMat>;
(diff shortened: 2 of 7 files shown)


stan/math/prim/fun/to_matrix.hpp
```diff
@@ -39,6 +39,29 @@ inline auto to_matrix(EigVec&& matrix) {
...
+/**
+ * Returns a matrix representation of a standard vector of Eigen
+ * column vectors. Each input vector becomes a column in the result.
+ *
+ * @tparam T type of the elements in the vector
+ * @param x standard vector of Eigen column vectors
+ * @return the matrix representation of the input
+ */
+template <typename T>
+inline Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> to_matrix(
```
SteveBronder, line 51: @WardBrian I think for `std::vector<Eigen::Matrix<T, Eigen::Dynamic, 1>>` we need to have a `to_row_matrix` and `to_col_matrix` so that the user can tell us if the vectors in the array are the rows or columns. Do you have thoughts about this?

WardBrian: Wouldn't they just be providing `std::vector<Eigen::Matrix<T, 1, Eigen::Dynamic>>` in that case?

SteveBronder: Oh I did not think of that. So an array of row vectors would essentially be calling `to_row_matrix` and an array of column vectors would be use `to_col_matrix`

stan/math/prim/fun/to_vector_array.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * Returns a standard vector of Eigen column vectors from the columns of
+ * the input matrix.
+ *
+ * @tparam EigMat type of the input matrix
+ * @param matrix input matrix
+ * @return the array of column vectors representation of the input
+ */
+template <typename EigMat, require_eigen_t<EigMat>* = nullptr>
+inline std::vector<Eigen::Matrix<value_type_t<EigMat>, Eigen::Dynamic, 1>>
+to_vector_array(const EigMat& matrix) {
```
SteveBronder, line 21:
```suggestion
to_col_vector_array(const EigMat& matrix) {
```

stan/math/prim/fun/to_matrix.hpp
SteveBronder: @fatimmajumder so this function is currently only works for `array vector`. We want this function to take in a standard vector with row or column vectors and then use `if constexpr (...)` to dispatch to make a rowwise or columnwise matrix based on the input type.

WardBrian: @SteveBronder there already is an overload for `array row_vector` pre-existing

SteveBronder (approved):
Good thank you!

bob-carpenter:
Thanks, [fatimmajumder](https://github.com/fatimmajumder)---I really wanted this feature for multivariate priors that want arrays of vectors.

